PR #703, opened 2021-01-04
EE-1114: Rename LMDB files

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/EE-1114

Sample Output: 
Screenshot from 2021-01-04 12-37-29

### DIFF
--- a/execution_engine/src/storage/transaction_source/lmdb.rs
+++ b/execution_engine/src/storage/transaction_source/lmdb.rs
@@ -11,6 +11,9 @@ use crate::storage::{
     MAX_DBS,
 };
 
+/// Filename for the LMDB database created by the EE.
+const EE_DB_FILENAME: &str = "data.lmdb";
+
 impl<'a> Transaction for RoTransaction<'a> {
     type Error = lmdb::Error;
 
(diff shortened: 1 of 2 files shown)


execution_engine/src/storage/transaction_source/lmdb.rs
```diff
@@ -71,10 +73,12 @@ impl LmdbEnvironment {
         max_readers: u32,
     ) -> Result<Self, error::Error> {
         let env = Environment::new()
+            // Set the flag to manage our own directory like in the storage component.
+            .set_flags(EnvironmentFlags::NO_SUB_DIR)
             .set_max_dbs(MAX_DBS)
             .set_map_size(map_size)
             .set_max_readers(max_readers)
-            .open(path.as_ref())?;
+            .open(&path.as_ref().join("data.lmdb"))?;
```
Fraser999, line 81: Can you make "data.lmdb" a `const &str` please?  I checked in components/storage.rs and see why you did it this way, but could you also change that too please so that "storage.lmdb" is a `const &str` in that file too?

Fraser999 (commented):
One nitpick.

execution_engine/src/storage/transaction_source/lmdb.rs
darthsiroftardis: Will do!

darthsiroftardis: Added a `const &str`

Fraser999 (approved):

PR #709, opened 2021-01-05
NDRS-668 - block validator - validate blocks similar to block_proposer

dwerner:
With this PR I started out trying to solve [NDRS-668](https://casperlabs.atlassian.net/browse/NDRS-668):

1. add a state machine to block_validator that loads the chainspec from storage (needed to validate deploys).
2. validate deploys against chainspec values

That _should_ have been it, but it turned out that `BlockValidator` was the source of many extra events on the queue:

The `BlockValidator` would always fetch all deploys when a `BlockValidationRequest` was received by it, regardless of if it has already validated it, or if there was a previous request in-flight for that block. This PR therefore also implements fetching the deploys for a block _if it has not been successfully validated in the past_. If it has been successfully validated, we can exit early, returning the validation to the caller.

I also add a `Debug` dump of the queue items to file (piggybacking on the already written json dump) where you can send a SIGUSR1 signal to get a dump in `queue_dump_debug.txt`, which can aid in debugging the queue. There are hashes present in this debug dump, making it possible to correlate requests/responses and other messages in general.

### DIFF
--- a/node/src/components/block_proposer.rs
+++ b/node/src/components/block_proposer.rs
@@ -406,11 +406,7 @@ impl BlockProposerReady {
                 .iter()
                 .all(|dep| past_deploys.contains(dep) || self.contains_finalized(dep))
         };
-        let ttl_valid = header.ttl() <= deploy_config.max_ttl;
-        let timestamp_valid = header.timestamp() <= block_timestamp;
-        let deploy_valid = header.timestamp() + header.ttl() >= block_timestamp;
-        let num_deps_valid = header.dependencies().len() <= deploy_config.max_dependencies as usize;
-        ttl_valid && timestamp_valid && deploy_valid && num_deps_valid && all_deps_resolved()
+        header.is_valid(deploy_config, block_timestamp) && all_deps_resolved()
     }
 
     /// Returns a list of candidates for inclusion into a block.
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -26,11 +26,14 @@ use crate::{
         EffectBuilder, EffectExt, EffectOptionExt, Effects, Responder,
     },
     types::{BlockLike, Deploy, DeployHash},
-    NodeRng,
+    Chainspec, NodeRng,
 };
 use keyed_counter::KeyedCounter;
 
 use super::fetcher::FetchResult;
+use crate::effect::requests::StorageRequest;
+use semver::Version;
+use std::sync::Arc;
 
 /// Block validator component event.
 #[derive(Debug, From, Display)]
@@ -46,6 +49,9 @@ pub enum Event<T, I> {
     /// A request to find a specific deploy, potentially from a peer, failed.
     #[display(fmt = "deploy {} missing", _0)]
     DeployMissing(DeployHash),
+
+    #[display(fmt = "block validator loaded")]
+    Loaded { chainspec: Arc<Chainspec> },
 }
 
 /// State of the current process of block validation.
@@ -59,103 +65,108 @@ pub(crate) struct BlockValidationState<T> {
     responders: SmallVec<[Responder<(bool, T)>; 2]>,
 }
 
-/// Block validator.
-#[derive(DataSize, Debug, Default)]
-pub(crate) struct BlockValidator<T, I> {
+#[derive(Debug)]
+pub(crate) struct BlockValidatorReady<T, I> {
+    /// Chainspec loaded for deploy validation.
+    chainspec: Arc<Chainspec>,
     /// State of validation of a specific block.
     validation_states: HashMap<T, BlockValidationState<T>>,
-
     /// Number of requests for a specific deploy hash still in flight.
     in_flight: KeyedCounter<DeployHash>,
 
     _marker: std::marker::PhantomData<I>,
 }
 
-impl<T, I> BlockValidator<T, I> {
-    /// Creates a new block validator instance.
-    pub(crate) fn new() -> Self {
-        BlockValidator {
-            validation_states: Default::default(),
-            in_flight: Default::default(),
-            _marker: std::marker::PhantomData,
-        }
-    }
-}
-
-impl<T, I, REv> Component<REv> for BlockValidator<T, I>
-where
-    T: BlockLike + Send + Clone + 'static,
-    I: Clone + Send + 'static,
-    REv: From<Event<T, I>>
-        + From<BlockValidationRequest<T, I>>
-        + From<FetcherRequest<I, Deploy>>
-        + Send,
+impl<T, I> BlockValidatorReady<T, I>
+    where
+        T: BlockLike + Send + Clone + 'static,
+        I: Clone + Send + 'static,
 {
-    type Event = Event<T, I>;
-    type ConstructionError = Infallible;
-
-    fn handle_event(
+    fn handle_event<REv>(
         &mut self,
         effect_builder: EffectBuilder<REv>,
-        _rng: &mut NodeRng,
-        event: Self::Event,
-    ) -> Effects<Self::Event> {
+        event: Event<T, I>,
+    ) -> Effects<Event<T, I>>
+        where
+            REv: From<Event<T, I>>
+            + From<BlockValidationRequest<T, I>>
+            + From<StorageRequest>
+            + From<FetcherRequest<I, Deploy>>
+            + Send,
+    {
+        let mut effects = Effects::new();
         match event {
             Event::Request(BlockValidationRequest {
-                block,
-                sender,
-                responder,
-                block_timestamp,
-            }) => {
-                if block.deploys().is_empty() {
+                               block,
+                               sender,
+                               responder,
+                               block_timestamp,
+                           }) => {
+                let block_deploys = block
+                    .deploys()
+                    .iter()
+                    .map(|deploy_hash| **deploy_hash)
+                    .collect::<HashSet<_>>();
+                if block_deploys.is_empty() {
                     // If there are no deploys, return early.
                     let mut effects = Effects::new();
                     effects.extend(responder.respond((true, block)).ignore());
                     return effects;
                 }
-                // No matter the current state, we will request the deploys inside this protoblock
-                // for now. Duplicate requests must still be answered, but are
-                // de-duplicated by the fetcher.
-                let effects = block
-                    .deploys()
-                    .iter()
-                    .flat_map(|deploy_hash| {
-                        // For every request, increase the number of in-flight...
-                        self.in_flight.inc(*deploy_hash);
-
-                        // ...then request it.
-                        let dh_found = **deploy_hash;
-                        let dh_not_found = **deploy_hash;
-                        effect_builder
-                            .fetch_deploy(**deploy_hash, sender.clone())
-                            .map_or_else(
-                                move |result: FetchResult<Deploy>| match result {
-                                    FetchResult::FromStorage(deploy)
-                                    | FetchResult::FromPeer(deploy, _) => {
-                                        if deploy.header().timestamp() > block_timestamp {
-                                            Event::DeployMissing(dh_found)
-                                        } else {
-                                            Event::DeployFound(dh_found)
-                                        }
-                                    }
-                                },
-                                move || Event::DeployMissing(dh_not_found),
-                            )
-                    })
-                    .collect();
 
                 // TODO: Clean this up to use `or_insert_with_key` once
                 // https://github.com/rust-lang/rust/issues/71024 is stabilized.
                 match self.validation_states.entry(block) {
                     Entry::Occupied(mut entry) => {
-                        // The entry already exists. We register ourselves as someone interested in
-                        // the ultimate validation result.
-                        entry.get_mut().responders.push(responder);
+                        // The entry already exists.
+                        if entry.get().missing_deploys.is_empty() {
+                            // Block has already been validated successfully, early return to
+                            // caller.
+                            effects.extend(responder.respond((true, entry.key().clone())).ignore());
+                        } else {
+                            // We register ourselves as someone interested in the ultimate
+                            // validation result.
+                            entry.get_mut().responders.push(responder);
+                        }
                     }
                     Entry::Vacant(entry) => {
                         // Our entry is vacant - create an entry to track the state.
                         let missing_deploys: HashSet<DeployHash> =
                             entry.key().deploys().iter().map(|hash| **hash).collect();
+
+                        let in_flight = &mut self.in_flight;
+                        let chainspec = self.chainspec.clone();
+                        let fetch_effects: Effects<Event<T, I>> = block_deploys
+                            .iter()
+                            .flat_map(|deploy_hash| {
+                                let chainspec = chainspec.clone();
+                                // For every request, increase the number of in-flight...
+                                in_flight.inc(deploy_hash);
+
+                                // ...then request it.
+                                let dh_found = *deploy_hash;
+                                let dh_not_found = *deploy_hash;
+                                effect_builder
+                                    .fetch_deploy(*deploy_hash, sender.clone())
+                                    .map_or_else(
+                                        move |result: FetchResult<Deploy>| match result {
+                                            FetchResult::FromStorage(deploy)
+                                            | FetchResult::FromPeer(deploy, _) => {
+                                                if deploy
+                                                    .header()
+                                                    .is_valid(&chainspec.genesis.deploy_config, block_timestamp)
+                                                {
+                                                    Event::DeployMissing(dh_found)
+                                                } else {
+                                                    Event::DeployFound(dh_found)
+                                                }
+                                            }
+                                        },
+                                        move || Event::DeployMissing(dh_not_found),
+                                    )
+                            })
+                            .collect();
+                        effects.extend(fetch_effects);
 
                         entry.insert(BlockValidationState {
                             missing_deploys,
@@ -163,10 +174,7 @@ where
                         });
                     }
                 }
-
-                effects
             }
-
             Event::DeployFound(deploy_hash) => {
                 // We successfully found a hash. Decrease the number of outstanding requests.
                 self.in_flight.dec(&deploy_hash);
@@ -176,7 +184,6 @@ where
                     state.missing_deploys.remove(&deploy_hash);
                 }
 
-                let mut effects = Effects::new();
                 // Now we remove all states that have finished and notify the requestors.
                 self.validation_states.retain(|key, state| {
                     if state.missing_deploys.is_empty() {
@@ -189,10 +196,7 @@ where
                         true
                     }
                 });
-
-                effects
             }
-
             Event::DeployMissing(deploy_hash) => {
                 // A deploy failed to fetch. If there is still hope (i.e. other outstanding
                 // requests), we just ignore this little accident.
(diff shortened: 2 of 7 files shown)


node/src/components/block_validator.rs
```diff
@@ -59,114 +65,116 @@ pub(crate) struct BlockValidationState<T> {
...
+                                    .fetch_deploy(*deploy_hash, sender.clone())
+                                    .map_or_else(
+                                        move |result: FetchResult<Deploy>| match result {
+                                            FetchResult::FromStorage(deploy)
+                                            | FetchResult::FromPeer(deploy, _) => {
+                                                if deploy.header().is_valid(
+                                                    &chainspec.genesis.deploy_config,
+                                                    block_timestamp,
+                                                ) {
+                                                    Event::DeployFound(dh_found)
```
Fraser999, line 159: The level of nesting here is pretty grim - I'd be inclined to try and define this closure in the `flat_map` outside of here or something like that.

node/src/components/block_validator.rs
```diff
@@ -216,9 +217,98 @@ where
...
                 });
+            }
+            Event::Loaded { .. } => {}
+        }
+        effects
+    }
+}
 
-                effects
+/// Block validator.
```
Fraser999, line 227: The comment is incorrect.

node/src/components/block_validator.rs
```diff
@@ -216,9 +217,98 @@ where
...
+    Loading(Vec<Event<T, I>>),
+    Ready(BlockValidatorReady<T, I>),
+}
+
+/// Block validator.
+#[derive(DataSize, Debug)]
+pub(crate) struct BlockValidator<T, I> {
+    #[data_size(skip)]
+    state: BlockValidatorState<T, I>,
+    _marker: std::marker::PhantomData<I>,
```
Fraser999, line 239: This can probably be removed now I think.

node/src/components/block_validator.rs
```diff
@@ -216,9 +217,98 @@ where
...
+        let mut effects = Effects::new();
+        match (&mut self.state, event) {
+            (BlockValidatorState::Loading(requests), Event::Loaded { chainspec }) => {
+                let mut new_ready_state = BlockValidatorReady {
+                    chainspec,
+                    validation_states: Default::default(),
+                    in_flight: Default::default(),
+                    _marker: std::marker::PhantomData,
+                };
+                let requests = requests;
```
Fraser999, line 297: We don't need this line I think.

node/src/components/block_validator.rs
```diff
@@ -216,9 +217,98 @@ where
...
+                }
+                self.state = BlockValidatorState::Ready(new_ready_state);
+            }
+            (BlockValidatorState::Loading(requests), request @ Event::Request(_)) => {
+                requests.push(request);
+            }
+            (BlockValidatorState::Ready(ref mut ready_state), event) => {
+                effects.extend(ready_state.handle_event(effect_builder, event));
             }
+            _ => {}
```
Fraser999, line 310: I think we should at least log an error here - I know we shouldn't get to this branch, but still, if we do, it's a bug (we'd have got a `DeployFound` or `DeployMissing` event while in the loading state).

node/src/utils/round_robin.rs
```diff
@@ -14,6 +14,7 @@ use std::{
 
 use enum_iterator::IntoEnumIterator;
 use serde::{ser::SerializeMap, Serialize, Serializer};
+use std::io::Write;
```
Fraser999, line 17: This should be grouped with the other std imports.

node/src/utils/round_robin.rs
```diff
@@ -143,6 +144,42 @@ where
     }
 }
 
+impl<I, K> WeightedRoundRobin<I, K>
+where
+    I: Debug,
+    K: Copy + Clone + Eq + Hash + IntoEnumIterator + Debug,
+{
+    /// Dump the contents of the queues (`Debug` representation) to a given file.
+    pub async fn debug_dump(&self, file: &mut std::fs::File) -> Result<(), std::io::Error> {
```
Fraser999, line 153: Should import `std::fs::File`.

node/src/utils/round_robin.rs
```diff
@@ -143,6 +144,42 @@ where
...
+            locks.push((kind, queue_guard));
+        }
+
+        for (kind, guard) in locks {
+            let queue = &*guard;
+            file.write_all(format!("Queue: {:?} ({}) [\n", kind, queue.len()).as_bytes())?;
+            for (i, event) in queue.iter().enumerate() {
+                file.write_all(format!("\t{:?}\n", event).as_bytes())?;
+                if i % 1024 == 0 {
+                    file.flush()?;
```
Fraser999, line 173: It might be better to wrap `file` in a `BufWriter` which should mean you can omit the explicit `flush()` calls other than the last one on line 177.

Fraser999 (commented):
Nitpicks only from me.

This is a very welcome PR - I can comfortably throw 1000 transfers at a 20 node nctl network without any of the nodes peaking at above 80 MB memory usage, whereas without this PR I see nodes topping out at 7GB after less than 200 transfers.  Nice work tracking down the issue.

node/src/components/block_validator.rs
```diff
@@ -26,11 +26,14 @@ use crate::{
         EffectBuilder, EffectExt, EffectOptionExt, Effects, Responder,
     },
     types::{BlockLike, Deploy, DeployHash},
-    NodeRng,
+    Chainspec, NodeRng,
 };
 use keyed_counter::KeyedCounter;
 
 use super::fetcher::FetchResult;
+use crate::effect::requests::StorageRequest;
```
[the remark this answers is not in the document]
dwerner: Done.

node/src/components/block_validator.rs
```diff
@@ -26,11 +26,14 @@ use crate::{
...
+use semver::Version;
```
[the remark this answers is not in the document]
dwerner: Done.

node/src/components/block_validator.rs
```diff
@@ -26,11 +26,14 @@ use crate::{
...
+use std::sync::Arc;
```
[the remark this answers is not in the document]
dwerner: Done.

node/src/components/block_validator.rs
```diff
@@ -46,6 +49,9 @@ pub enum Event<T, I> {
     /// A request to find a specific deploy, potentially from a peer, failed.
     #[display(fmt = "deploy {} missing", _0)]
     DeployMissing(DeployHash),
+
+    #[display(fmt = "block validator loaded")]
```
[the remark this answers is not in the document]
dwerner: Added one.

node/src/components/block_validator.rs
```diff
@@ -59,114 +65,116 @@ pub(crate) struct BlockValidationState<T> {
...
+pub(crate) struct BlockValidatorReady<T, I> {
+    /// Chainspec loaded for deploy validation.
+    chainspec: Arc<Chainspec>,
     /// State of validation of a specific block.
     validation_states: HashMap<T, BlockValidationState<T>>,
-
     /// Number of requests for a specific deploy hash still in flight.
     in_flight: KeyedCounter<DeployHash>,
 
     _marker: std::marker::PhantomData<I>,
```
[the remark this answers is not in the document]
dwerner: Done.

node/src/components/block_validator.rs
```diff
@@ -59,114 +65,116 @@ pub(crate) struct BlockValidationState<T> {
...
+                let block_deploys = block
+                    .deploys()
+                    .iter()
+                    .map(|deploy_hash| **deploy_hash)
+                    .collect::<HashSet<_>>();
+                if block_deploys.is_empty() {
                     // If there are no deploys, return early.
                     let mut effects = Effects::new();
                     effects.extend(responder.respond((true, block)).ignore());
                     return effects;
```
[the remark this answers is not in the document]
dwerner: Done.

node/src/components/block_validator.rs
```diff
@@ -59,114 +65,116 @@ pub(crate) struct BlockValidationState<T> {
...
+                            entry.get_mut().responders.push(responder);
+                        }
                     }
                     Entry::Vacant(entry) => {
                         // Our entry is vacant - create an entry to track the state.
                         let missing_deploys: HashSet<DeployHash> =
                             entry.key().deploys().iter().map(|hash| **hash).collect();
 
+                        let in_flight = &mut self.in_flight;
+                        let chainspec = self.chainspec.clone();
```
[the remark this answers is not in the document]
dwerner: Done.

node/src/components/block_validator.rs
```diff
@@ -59,114 +65,116 @@ pub(crate) struct BlockValidationState<T> {
...
+                        let fetch_effects: Effects<Event<T, I>> = block_deploys
+                            .iter()
+                            .flat_map(|deploy_hash| {
+                                let chainspec = chainspec.clone();
+                                // For every request, increase the number of in-flight...
+                                in_flight.inc(deploy_hash);
+
+                                // ...then request it.
+                                let dh_found = *deploy_hash;
+                                let dh_not_found = *deploy_hash;
```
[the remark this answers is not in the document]
dwerner: Done.

node/src/components/block_validator.rs
dwerner: Done.

node/src/components/block_validator.rs
dwerner: Fixed.

node/src/components/block_validator.rs
dwerner: Yep, thanks.

node/src/components/block_validator.rs
dwerner: Removed.

node/src/components/block_validator.rs
dwerner: Made this more explicit as to what match arm was being hit, now logging an error.

node/src/utils/round_robin.rs
dwerner: Done.

node/src/utils/round_robin.rs
dwerner: Done.

node/src/utils/round_robin.rs
dwerner: Done.

node/src/utils/round_robin.rs
```diff
@@ -143,6 +146,39 @@ where
...
+        let mut writer = BufWriter::new(file);
+        for (kind, guard) in locks {
+            let queue = &*guard;
+            writer.write_all(format!("Queue: {:?} ({}) [\n", kind, queue.len()).as_bytes())?;
+            for event in queue.iter() {
+                writer.write_all(format!("\t{:?}\n", event).as_bytes())?;
+            }
+            writer.write_all(b"]\n")?;
+        }
+        Ok(())
```
Fraser999, line 178: We should have kept the final `flush()` call.

From [the docs](https://doc.rust-lang.org/std/io/struct.BufWriter.html):

> It is critical to call flush before BufWriter<W> is dropped. Though dropping will attempt to flush the contents of the buffer, any errors that happen in the process of dropping will be ignored. Calling flush ensures that the buffer is empty and thus dropping will not even attempt file operations.

node/src/utils/round_robin.rs
```diff
@@ -7,13 +7,16 @@
...
+    fs::File,
     hash::Hash,
+    io::{self, Write},
     num::NonZeroUsize,
     sync::atomic::{AtomicUsize, Ordering},
 };
 
 use enum_iterator::IntoEnumIterator;
 use serde::{ser::SerializeMap, Serialize, Serializer};
+use std::io::BufWriter;
```
Fraser999, line 19: This should be grouped with the other std imports

node/src/utils/round_robin.rs
dwerner: Done

node/src/utils/round_robin.rs
dwerner: Done.

Fraser999 (approved):

node/src/components/block_validator.rs
```diff
@@ -46,6 +50,11 @@ pub enum Event<T, I> {
     /// A request to find a specific deploy, potentially from a peer, failed.
     #[display(fmt = "deploy {} missing", _0)]
     DeployMissing(DeployHash),
+
+    /// An event changing the current state to BlockValidatorReady, once the chainspec has been
+    /// loaded.
+    #[display(fmt = "block validator loaded")]
+    Loaded { chainspec: Arc<Chainspec> },
```
goral09, line 57: We could also use the approach that @marc-casperlabs used in the `BlockProposer` – basically turning it into an actor-like component that loads its state prior to launch.

node/src/components/block_validator.rs
```diff
@@ -59,114 +68,114 @@ pub(crate) struct BlockValidationState<T> {
     responders: SmallVec<[Responder<(bool, T)>; 2]>,
 }
 
-/// Block validator.
-#[derive(DataSize, Debug, Default)]
-pub(crate) struct BlockValidator<T, I> {
+#[derive(Debug)]
+pub(crate) struct BlockValidatorReady<T, I> {
+    /// Chainspec loaded for deploy validation.
+    chainspec: Arc<Chainspec>,
```
goral09, line 74: `BlockValidator` shouldn't need the whole chainspec, it has things like Genesis contracts (those are byte arrays). It should have a smaller, subset of configuration parameters that it needs. Can be ticketed and done separately.

goral09: Another thing, this configuration will change based on the protocol version so it cannot be set in stone. Block Validator may have to use different configurations for block #11, and later block #12 and may even have to go back to the configuration from the block #11 if the messages arrive out of order (from other nodes, not internal ones).

node/src/components/block_validator.rs
dwerner: I largely followed what had been done in the `block_proposer` - what improvements do you think could be made, or did I miss something?

node/src/components/block_validator.rs
dwerner: If I'm not mistaken, this is just an Arc reference to a shared value, so we aren't copying those byte arrays around. It seems like the same is done in the `block_proposer` as well.

goral09: Might be, we should change it in both places then.

dwerner: > Another thing, this configuration will change based on the protocol version so it cannot be set in stone. Block Validator may have to use different configurations for `block #11`, and later `block #12` and may even have to go back to the configuration from the `block #11` if the messages arrive out of order (from other nodes, not internal ones).

Yes, this should be handled when we work on chainspec upgrades.

node/src/components/block_validator.rs
```diff
@@ -59,114 +68,114 @@ pub(crate) struct BlockValidationState<T> {
...
+                    .map(|deploy_hash| **deploy_hash)
+                    .collect::<HashSet<_>>();
+                if block_deploys.is_empty() {
+                    // If there are no deploys, return early.
+                    return responder.respond((true, block)).ignore();
+                }
 
                 // TODO: Clean this up to use `or_insert_with_key` once
                 // https://github.com/rust-lang/rust/issues/71024 is stabilized.
                 match self.validation_states.entry(block) {
```
goral09, line 120: How long does the block validator store the information about the block's validation state? Once a block is validated, is the entry removed from `validation_states`?

node/src/components/block_validator.rs
```diff
@@ -59,114 +68,114 @@ pub(crate) struct BlockValidationState<T> {
...
                             entry.key().deploys().iter().map(|hash| **hash).collect();
 
+                        let in_flight = &mut self.in_flight;
+                        let chainspec = Arc::clone(&self.chainspec);
+                        let fetch_effects: Effects<Event<T, I>> = block_deploys
+                            .iter()
+                            .flat_map(|deploy_hash| {
+                                let chainspec = Arc::clone(&chainspec);
+                                // For every request, increase the number of in-flight...
+                                in_flight.inc(deploy_hash);
```
goral09, line 145: `in_flight` uses a `HashMap` internally, this might end up doing a lot of `hash`-ing when we increment the number of in-flight requests for every deploy in a block.

node/src/components/block_validator.rs
goral09: Is there a ticket for that so that we don't lose the track of it?

node/src/components/block_validator.rs
dwerner: It is only dropped if a block _fails_ validation, and that behaviour existed before my changeset.

node/src/components/block_validator.rs
```diff
@@ -59,114 +68,114 @@ pub(crate) struct BlockValidationState<T> {
...
+                                    move |result: FetchResult<Deploy>| match result {
+                                        FetchResult::FromStorage(deploy)
+                                        | FetchResult::FromPeer(deploy, _) => {
+                                            if deploy.header().is_valid(
+                                                &chainspec.genesis.deploy_config,
+                                                block_timestamp,
+                                            ) {
+                                                Event::DeployFound(deploy_hash)
+                                            } else {
+                                                Event::DeployMissing(deploy_hash)
```
goral09, line 159: I am worried that this is swallowing the real reason this block cannot be validated – it turns an _invalid_ deploy into a missing one. It's probably fine for now , I just want to put it out there.

dwerner: True, after all the work of fetching the deploys, we are returning a `bool`. I would like to rethink a few aspects of the `block_validator` at some point - for instance, batching the deploy fetching.

goral09 (approved):
Looks good, no blockers from me.

node/src/components/block_validator.rs
goral09: > batching the deploy fetching.
Oh yes, definitely that :+1: :100:

node/src/components/block_validator.rs
dwerner: I mentioned this to @Fraser999, he's working on designing our upgrade approach.

goral09: Consensus has [`ProtocolConfig`](https://github.com/CasperLabs/casper-node/blob/master/node/src/components/consensus/config.rs#L27) that uses a subset of the original `Chainspec` and is actually [created from one](https://github.com/CasperLabs/casper-node/blob/master/node/src/components/consensus/config.rs#L41).

node/src/components/block_proposer.rs
```diff
@@ -406,11 +406,7 @@ impl BlockProposerReady {
                 .iter()
                 .all(|dep| past_deploys.contains(dep) || self.contains_finalized(dep))
         };
-        let ttl_valid = header.ttl() <= deploy_config.max_ttl;
-        let timestamp_valid = header.timestamp() <= block_timestamp;
-        let deploy_valid = header.timestamp() + header.ttl() >= block_timestamp;
-        let num_deps_valid = header.dependencies().len() <= deploy_config.max_dependencies as usize;
-        ttl_valid && timestamp_valid && deploy_valid && num_deps_valid && all_deps_resolved()
+        header.is_valid(deploy_config, block_timestamp) && all_deps_resolved()
```
afck, line 409: This seems to be `self.deploy_config` currently. Will this be upgradeable? Then we will need to use the deploy config version appropriate for that block height/era. (Is there a ticket for that?)

node/src/components/block_validator.rs
marc-casperlabs: Side-note: The `Arc` is mostly there for transfering the chainspec around. I just did not feel like cloning it each time it was requested. I used a `ProtocolConfig` because that is what the code previously used, so I am currently neither for nor against keeping the actual `Arc` around :)

PR #715, opened 2021-01-06
NO TICKET: Tidier error conversion

xcthulhu:


xcthulhu:
I made this change in a branch I was working on.  Wasn't necessary for what I was doing but I figured I'd make a little PR for it.

henrytill (approved):

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -2371,10 +2371,9 @@ where
     where
         Error: From<S::Error>,
     {
-        match self.state.commit(correlation_id, pre_state_hash, effects)? {
-            CommitResult::Success { state_root, .. } => Ok(CommitResult::Success { state_root }),
-            commit_result => Ok(commit_result),
-        }
+        self.state
+            .commit(correlation_id, pre_state_hash, effects)
+            .map_err(Error::from)
```
goral09, line 2376: How does this even work? How is compiler able to figure out that it needs to drop everything but `state_root` when the result is `CommitResult::Success` ? Is there a mapper defined already that turns the result of `commit` call into the `CommitResult`?

xcthulhu:
bors merge

PR #716, opened 2021-01-06
NCTL :: Consensus storage sub-folder

ghost:
Consensus is now persisting some transient data to disk for use in restart scenarios.  This PR covers this situation and adds a new command: `view-node-storage-consensus`.


linclelinkpart5 (approved):
Looks good!

PR #717, opened 2021-01-06
NDRS-672: Track validator weights when syncing

xcthulhu:


xcthulhu:
Tracking validator weights when syncing is close to done, but there's a few dangling things:

 1. @henrytill @EdHastingsCasperLabs I need to change the protocol buffer for `StepResult`, but I am not sure what that entails.
 2. @goral09 @fizyk20 @afck When a blockheader comes over the wire, we should reject it if it has the wrong era validators.  I'm not sure how to do this and maybe this is another ticket.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -2476,7 +2476,7 @@ where
         &self,
         correlation_id: CorrelationId,
         step_request: StepRequest,
-    ) -> Result<StepResult, Error> {
+    ) -> Result<StepResult, GetEraValidatorsError> {
```
xcthulhu, line 2479: @EdHastingsCasperLabs It ended up being easiest to have this become a `GetEraValidatorsError`, but it's a bit odd and I'm not sure if there's a better way here.

grpc/server/src/engine_server/mod.rs
```diff
@@ -451,7 +451,11 @@ where
         };
 
         let response = match self.commit_step(correlation_id, request) {
-            Ok(StepResult::Success { post_state_hash }) => {
+            // TODO: Update protobufs for StepResult to use GetEraValidatorsError and have the
+            // validator weights
```
xcthulhu, line 455: @henrytill @EdHastingsCasperLabs including `next_era_validators` in the protocol buffers drastically changes their format.  Not including them doesn't seem to break any tests... what should I expect to break here?

node/src/components/block_executor.rs
```diff
@@ -528,9 +542,23 @@ impl<REv: ReactorEventT> Component<REv> for BlockExecutor {
...
+                    Ok(StepResult::Success {
+                        post_state_hash,
+                        next_era_validators,
+                    }) => {
                         state.state_root_hash = post_state_hash.into();
-                        self.finalize_block_execution(effect_builder, state)
+                        let next_era_validators: BTreeMap<PublicKey, U512> = next_era_validators
+                            .into_iter()
+                            .map(|(casper_types_public_key, weight)| {
+                                let public_key = PublicKey::try_from(casper_types_public_key).expect("Could not convert casper_types public key into node public key");
```
xcthulhu, line 553: @afck @fizyk20 @goral09 Feels bad to do an `expect` here... is there something better I can do?

xcthulhu: Perhaps a `filter_map` would be better here...?

grpc/server/src/engine_server/mod.rs
xcthulhu: I've updated the protocol buffers in https://github.com/CasperLabs/casper-node/pull/717/commits/67140cb59f3dcf950bb8a6402196ee6d72f03341

xcthulhu:
>     1. @henrytill @EdHastingsCasperLabs I need to change the protocol buffer for `StepResult`, but I am not sure what that entails.

To answer this question, the protocol buffers are used in the tests.  I have updated the protocol buffers here: https://github.com/CasperLabs/casper-node/pull/717/commits/67140cb59f3dcf950bb8a6402196ee6d72f03341

node/src/components/block_executor.rs
afck: That can only fail if a validator bonds with an invalid public key, can it?
I think we have the same problem in `EraSupervisor` somewhere; we just need to make sure we do the same thing in both places.

(Also: formatting! :grin:)

xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/717/commits/86eff05cf00063166109268eaaaf6e1f5436cc06

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -2543,7 +2543,12 @@ where
         let gas_limit = Gas::new(U512::from(std::u64::MAX));
         let deploy_hash = {
             // seeds address generator w/ protocol version
-            let bytes: Vec<u8> = step_request.protocol_version.value().into_bytes()?.to_vec();
+            let bytes: Vec<u8> = step_request
+                .protocol_version
+                .value()
+                .into_bytes()
+                .map_err(Error::from)?
```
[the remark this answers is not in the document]
xcthulhu: `into_bytes` gives a `Result<_, bytesrepr::Error>`.  But `GetEraValidatorsError` doesn't have a variant for `bytesrepr::Error`.  There is a variant for `engine_state::error::Error`, and `engine_state::error::Error` has a variant for `bytesrepr::Error`, so we do two coercions.

I could modify `GetEraValidatorsError` but it was simpler to do this.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -2678,20 +2683,32 @@ where
                 step_request.pre_state_hash,
                 effects.transforms,
             )
-            .map_err(Into::into)?;
+            .map_err(|err| Error::Exec(Into::into(err)))?;
```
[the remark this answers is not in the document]
xcthulhu: Again, I did this to avoid extending the `GetEraValidators` error enum.

grpc/server/src/engine_server/mod.rs
```diff
@@ -450,19 +448,39 @@ where
...
+                (post_state_hash, next_era_validators)
             }
             Ok(result) => {
                 let err_msg = result.to_string();
                 warn!("{}", err_msg);
                 let mut ret = ipc::StepResponse::new();
                 ret.mut_step_result().mut_error().set_message(err_msg);
+                return SingleResponse::completed(ret);
+            }
+            Err(err) => {
```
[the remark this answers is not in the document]
xcthulhu: So in this leg here:

```rust
            Ok(result) => {
                let err_msg = result.to_string();
                warn!("{}", err_msg);
                let mut ret = ipc::StepResponse::new();
                ret.mut_step_result().mut_error().set_message(err_msg);
                return SingleResponse::completed(ret);
            }
```

Here `result` is a `StepResult`. But it is not `StepResult::Success` which is handled as the first `match` case.  When we convert to protobuf we make it an error.

The other path for dealing with `GetErrorValidatorsErrors` has the same logic but the types are different.  Not sure how to dry this code out exactly.

afck (approved):
Looks good to me, but let's also wait for someone more familiar with the Contract Runtime to review it.

henrytill:
As discussed offline, you should restore the broken step tests so that @EdHastingsCasperLabs can take a look.

EdHastingsCasperLabs (commented):
I'm digging into this now; should have a resolution soon.

xcthulhu:
Thanks @EdHastingsCasperLabs for the fixup... I want to wait for your approval before going ahead with the merge.

node/src/types/block.rs
```diff
@@ -537,6 +558,7 @@ pub struct BlockHeader {
     era_id: EraId,
     height: u64,
     proposer: PublicKey,
+    next_era_validator_weights: Option<BTreeMap<PublicKey, U512>>,
```
goral09, line 561: Why is it an `Option`? We should be able to always populate it.

goral09:
> 2. @goral09 @fizyk20 @afck When a blockheader comes over the wire, we should reject it if it has the wrong era validators.  I'm not sure how to do this and maybe this is another ticket.

We don't send linear chain blocks over the wire so we don't have to validate them. We only create finality signatures for them, which are gossiped, and they are validated.

node/src/components/linear_chain_sync.rs
```diff
@@ -265,40 +285,60 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
...
+            } if highest_block_seen != block_height => {
+                if let Some(validator_weights_for_new_era) =
+                    block_header.next_era_validator_weights()
+                {
+                    *validator_weights = validator_weights_for_new_era.clone();
+                }
+                self.state = curr_state;
+                self.fetch_next_block_deploys(effect_builder)
+            }
+            // Otherwise transition to State::SyncingDescendants
```
goral09, line 305: I think the approach with conditions on match patterns is dangerous as it's error-prone. If this and the next arms were switched places then the logic changes. That's why I put it under one match case. It was deliberate.

PR #722, opened 2021-01-07
NDRS-713: update types' PublicKey in the fashion of node

henrytill:
Ref: https://casperlabs.atlassian.net/browse/NDRS-713

This PR replaces the `PublicKey` type with the one found in node's `crypto` modules (adding the `SecretKey` and `Signature` types alongside it).  

PEM/DER encoding/decoding and file operations are kept in node using an `AsymmetricKeyExt` extension trait.


node/src/components/consensus.rs
```diff
@@ -288,16 +288,7 @@ where
...
-                    Ok(Some(validator_weights)) => validator_weights
-                        .into_iter()
-                        .filter_map(|(key, stake)| match key.try_into() {
-                            Ok(key) => Some((key, stake)),
-                            Err(error) => {
-                                error!(%error, "error converting the bonded key");
-                                None
-                            }
-                        })
-                        .collect(),
```
xcthulhu, line 300: :champagne:

types/src/crypto/asymmetric_key/tests.rs
```diff
@@ -0,0 +1,42 @@
+use crate::{crypto::SecretKey, PublicKey};
+
+#[test]
+fn can_construct_ed25519_keypair_from_zeroes() {
+    let bytes = [0; SecretKey::ED25519_LENGTH];
+    let secret_key = SecretKey::ed25519(bytes);
+    let _public_key: PublicKey = secret_key.into();
+}
+
+// is this correct?
```
Fraser999, line 10: Yes, it seems so.  Setting the secret key to all zeros hits [this code](https://github.com/RustCrypto/traits/blob/elliptic-curve-v0.5.0/elliptic-curve/src/scalar.rs#L32-L38), and from there we get to [this code](https://github.com/CasperLabs/casper-node/blob/a31d1edd4a1067993f51eb40027c3a504dee00b8/types/src/crypto/asymmetric_key.rs#L365).

The first linked code block is the only way `k256::PublicKey::from_secret_key` can fail I believe, although possibly the `.expect` warrants a comment to explain why it's safe.

I believe it's not a problem in production as there should be no reasonable way for a non-malicious user to have a secret key which is all zeros.  They can't construct one, and they won't ever receive one from a malicious peer.

types/src/lib.rs
```diff
@@ -23,38 +23,6 @@ extern crate alloc;
 #[macro_use]
 extern crate std;
 
-mod access_rights;
```
Fraser999, line 26: I thought the order was mods then imports?

types/Cargo.toml
```diff
@@ -12,12 +12,16 @@ license-file = "../LICENSE"
...
 [dependencies]
 base16 = { version = "0.2.1", default-features = false }
+base64 = { version = "0.13.0", default-features = false }
 bitflags = "1"
 blake2 = { version = "0.9.0", default-features = false }
 # TODO: Replace failure with thiserror once no_std support is landed https://github.com/dtolnay/thiserror/pull/64
 datasize = { version = "0.2.0", default-features = false }
+ed25519-dalek = { version = "1.0.0", default-features = false, features = ["rand", "u64_backend"] }
 failure = { version = "0.1.8", default-features = false, features = ["failure_derive"] }
+hex = { version = "0.4.2", default-features = false }
```
Fraser999, line 22: We can probably avoid importing both `hex` and `base16`, but I'd leave that as a tech debt ticket to keep this already-large PR manageable.

Fraser999 (approved):
Looks good to me - very happy to see this change!

If we find users of our casper-types crate want to make use of the sign/verify functionality it should be an easy update to move that code I guess.

No blockers from me, only a nitpick which can be ignored if wanted.

types/src/crypto/asymmetric_key/tests.rs
henrytill: Agreed.  Had also verified this after I wrote that, just forgot to remove the comment.

types/src/lib.rs
henrytill: Whoops, CLion did this.  Will fix.

PR #724, opened 2021-01-08
Op 1750 casper-node redirect stderr into casper-node.stderr.log

rob-casper:
- seperate stderr logging as it outputs non-JSON stack traces (without redirection ends up in journald)
- logrotate applied to /var/log/casper/casper-node* (glob)


sacherjj (approved):
LGTM

PR #728, opened 2021-01-10
HOT-FIX: Recover from `EraValidatorsMissing` error.

goral09:


afck (approved):

afck:
Clippy is right, though.
(Maybe it's even simpler to remove the new method, and just use `matches!` directly?)

goral09:
> Clippy is right, though.
> (Maybe it's even simpler to remove the new method, and just use `matches!` directly?)

I didn't want to import EE error type into the `linear_chain` component.

Interestingly, _my_ clippy doesn't complain at all.

PR #730, opened 2021-01-10
NDRS-698: if any HTTP server fails to start, causes the node to stop

Fraser999:
Completes [NDRS-698](https://casperlabs.atlassian.net/browse/NDRS-698).

Tagging @momipsl since this will be a new reason for nodes to fail to start in nctl: they need to be able to bind to not just the specified address for the small_network, but also the three addresses for the three HTTP servers.

### DIFF
--- a/node/src/components/event_stream_server.rs
+++ b/node/src/components/event_stream_server.rs
@@ -33,6 +33,7 @@ use tokio::sync::mpsc::{self, UnboundedSender};
 use super::Component;
 use crate::{
     effect::{EffectBuilder, Effects},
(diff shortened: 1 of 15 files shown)


marc-casperlabs:
I'd like to give a little context why the code ended up in its current state; originally it was set up this way to make it easier to run multiple nodes on a single machine - one would have to allocate only one unique port for each, not the HTTP server any (first node "won", others would simply fail to start their HTTP server but not care).

Hopefully this won't break any old testing scripts.

marc-casperlabs (approved):
LGTM

PR #733, opened 2021-01-11
HWY-221: Add a test, remove two TODOs.

afck:
https://casperlabs.atlassian.net/browse/HWY-221

### DIFF
--- a/node/src/components/consensus/highway_core/highway.rs
+++ b/node/src/components/consensus/highway_core/highway.rs
@@ -690,8 +690,8 @@ pub(crate) mod tests {
         let mut rng = crate::new_rng();
         let now: Timestamp = 500.into();
 
-        let _ = add_unit!(state, rng, CAROL, 0xC0; N, N, N)?;
-        let _ = add_unit!(state, rng, CAROL, 0xC1; N, N, N)?;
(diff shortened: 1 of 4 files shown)


node/src/components/consensus/highway_core/highway.rs
```diff
@@ -689,6 +684,74 @@ pub(crate) mod tests {
...
     }
 
+    #[test]
+    fn missing_dependency() -> Result<(), AddUnitError<TestContext>> {
+        let mut state = State::new_test(WEIGHTS, 0);
+        let mut rng = crate::new_rng();
+        let now: Timestamp = 500.into();
+
+        let _ = add_unit!(state, rng, CAROL, 0xC0; N, N, N)?;
+        let _ = add_unit!(state, rng, CAROL, 0xC1; N, N, N)?;
```
xcthulhu, line 694: Why do we need to bind these?  Aren't we just calling them for effect?

afck: You're right, I'll remove the `let _ =`.

xcthulhu (approved):

afck:
bors merge

PR #734, opened 2021-01-11
Don't slow down with fewer than 50% failed rounds.

afck:
Since this is hard-coded and it's reasonable to set any fault tolerance threshold up to 50%, let's make sure that fewer than 50% of the validators cannot force the others to increase their round length.


goral09 (approved):

afck:
bors merge

PR #739, opened 2021-01-12
NDRS-619: Remove FinalizedBlock Event from event stream

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-619

Removed the `FinalizedBlock` event from the event stream, "actual block" as mentioned in the description is already part of the event stream as the `BlockHeader`

### DIFF
--- a/node/src/reactor/joiner.rs
+++ b/node/src/reactor/joiner.rs
@@ -727,7 +727,7 @@ impl reactor::Reactor for Reactor {
                     ),
                 ),
                 ConsensusAnnouncement::Finalized(_) => {
-                    info!("A block was finalized");
+                    // A block was finalized.
                     Effects::new()
                 }
                 ConsensusAnnouncement::Fault {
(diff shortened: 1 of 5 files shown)


node/src/reactor/joiner.rs
```diff
@@ -726,14 +726,10 @@ impl reactor::Reactor for Reactor {
...
-                ConsensusAnnouncement::Finalized(block) => reactor::wrap_effects(
-                    Event::EventStreamServer,
-                    self.event_stream_server.handle_event(
-                        effect_builder,
-                        rng,
-                        event_stream_server::Event::BlockFinalized(block),
-                    ),
-                ),
+                ConsensusAnnouncement::Finalized(_) => {
+                    info!("A block was finalized");
```
goral09, line 730: We don't need this log. It gets logged when it is finalized in consensus already.

darthsiroftardis: Removed log statement

Fraser999 (approved):

goral09 (approved):

PR #747, opened 2021-01-13
NO-TICKET: fix issue in network component

Fraser999:


marc-casperlabs (approved):
That should fix the problem. Maybe there's a case for not using a `watch`, but `oneshot` here though? Regardless, this will work.

PR #753, opened 2021-01-14
NO-TICKET: fix a missing wireup in the joiner reactor

Fraser999:


sacherjj (approved):
Tricky one.

node/src/reactor/joiner.rs
```diff
@@ -213,7 +213,11 @@ impl From<StorageRequest> for Event {
 
 impl From<NetworkRequest<NodeId, Message>> for Event {
     fn from(request: NetworkRequest<NodeId, Message>) -> Self {
-        Event::SmallNetwork(small_network::Event::from(request))
+        if env::var(ENABLE_LIBP2P_ENV_VAR).is_ok() {
+            Event::Network(network::Event::from(request))
+        } else {
+            Event::SmallNetwork(small_network::Event::from(request))
+        }
```
goral09, line 220: Should we unify approaches in `validator.rs` and `joiner.rs`? IIRC, in `validator.rs` it's explicitly matched on inside `handle_message`.

Fraser999: I briefly tried that for this particular case, but immediately hit compiler errors.  I agree it's probably worth doing, but I'd prioritise the work towards providing a macro for reactors, and probably also merging the three reactor types before taking this on personally.  That other work would make unifying these redundant.

goral09: I doubt that switching to a macro-based approach or merging reactors will happen anytime before the mainnet release. We've had macros for months now and we don't use it for either of the 3 reactors.

goral09 (approved):

PR #754, opened 2021-01-14
EE-1141: Hide purse APIs from docs.

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1141

By the way I renamed "getPurseBalance" into "getBalance" in AssemblyScript API, but I didn't hide it in Rust API


henrytill (changes requested):
Looks good.

I'm creeping the scope here, but I think we probably need to address these other places right now.

1. I think we need to address this so I'll request the change here: rename `get_balance` to `get_purse_balance` and hide it.  Then write a new `get_balance` that doesn't take any arguments and gets the balance of the caller's main purse.

2. Remove transfer to purse from the client.

henrytill:
Also: hide `system::create_purse` and the AS equivalent.

henrytill (approved):
Good here.

client/tests/integration_test.rs
```diff
@@ -776,41 +768,19 @@ mod transfer {
...
-                "",
                 "",
                 deploy_params::test_data_valid(),
                 payment_params::test_data_with_name()
             ),
             Err(Error::InvalidArgument(
-                "target_account | target_purse",
-                "Invalid arguments to get_transfer_target - must provide either a target account or purse. account=, purse=".to_string()).into())
+                "target_account",
+                "Invalid arguments to get_transfer_target - must provide either a target account. account=".to_string()).into())
```
dwerner, line 783:
```suggestion
                "Invalid arguments to get_transfer_target - must provide a target account. account=".to_string()).into())
```

dwerner (approved):

PR #755, opened 2021-01-14
HWY-229: Use the same numbers for adaptive round length as in the paper.

afck:
This makes us a bit less eager to slow down, and a bit more eager to speed up, which is appropriate if we set the minimum round length to a reasonable value anyway.

https://casperlabs.atlassian.net/browse/HWY-229

### DIFF
--- a/node/src/components/consensus/protocols/highway/round_success_meter.rs
+++ b/node/src/components/consensus/protocols/highway/round_success_meter.rs
@@ -11,22 +11,29 @@ use crate::{
     types::Timestamp,
 };
 
-// The number of most recent rounds we will be keeping track of.
+/// The number of most recent rounds we will be keeping track of.
 const NUM_ROUNDS_TO_CONSIDER: usize = 40;
-// The maximum number of failures allowed among NUM_ROUNDS_TO_CONSIDER latest rounds, with which we
-// won't increase our round length. Exceeding this threshold will mean that we should slow down.
-const MAX_FAILED_ROUNDS: usize = 20;
-// We will try to accelerate (decrease our round exponent) every `ACCELERATION_PARAMETER` rounds if
-// we have few enough failures.
+/// The number of successful rounds that triggers us to slow down: With this many or fewer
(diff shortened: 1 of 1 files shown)


fizyk20 (approved):
LGTM :+1:

node/src/components/consensus/protocols/highway/round_success_meter.rs
```diff
@@ -11,23 +11,30 @@ use crate::{
...
+/// The required quorum in a summit we will look for to check if a round was successful is
+/// determined by this FTT.
 const THRESHOLD: u64 = 1;
 
+/// The maximum number of failures allowed among NUM_ROUNDS_TO_CONSIDER latest rounds, with which we
+/// won't increase our round length. Exceeding this threshold will mean that we should slow down.
+const MAX_FAILED_ROUNDS: usize = NUM_ROUNDS_TO_CONSIDER - NUM_ROUNDS_SLOWDOWN - 1;
+/// The maximum number of failures with which we will attempt to accelerate (decrease the round
+/// exponent).
+const MAX_FAILURES_FOR_ACCELERATION: usize = NUM_ROUNDS_TO_CONSIDER - NUM_ROUNDS_SPEEDUP;
```
goral09, line 36: Isn't the round either successful or failed? Do we need these two additional constants ? Wouldn't `NUM_ROUNDS_SLOWDOWN` and `NUM_ROUNDS_SPEEDUP` suffice? 

If the number of successful rounds is less than `NUM_ROUNDS_SLOWDOWN` – we slow down. 
If the number of successful rounds is greater than `NUM_ROUNDS_SPEEDUP` – we speed up.

afck: Yes, but I noticed that the code behaves a bit differently:
It'll slow down even before we have run `NUM_ROUNDS_TO_CONSIDER` (currently 40) in total. So if we're only in round number 20 and have `MAX_FAILED_ROUNDS` failures, we slow down. The number of successful rounds at that point is not meaningful.

afck:
bors merge

PR #759, opened 2021-01-14
NDRS-715 block-proposer: gas price prioritization

dwerner:
This PR implements NDRS-715 - proper gas price prioritization by the block-proposer, as per CEP https://github.com/CasperLabs/ceps/pull/22

This also includes the related https://casperlabs.atlassian.net/browse/NDRS-669 (use wasm-less transfer gas cost from chainspec)


node/src/components/block_proposer/deploy_sets.rs
```diff
@@ -11,12 +11,21 @@ use crate::{
...
 /// Stores the internal state of the BlockProposer.
 #[derive(Clone, DataSize, Debug, Deserialize, Serialize)]
 pub struct BlockProposerDeploySets {
     /// The collection of deploys pending for inclusion in a block.
-    pub(super) pending: HashMap<DeployHash, DeployType>,
+    /// The key of this map is used as a method of keeping the map sorted first by `gas_price`
+    /// (willingness to pay), then by `payment_amount` and finally include `deploy_hash` to
+    /// facilitate map operations.
+    /// https://github.com/CasperLabs/ceps/blob/Gas_spot_market/text/0022-gas-spot-market.md#ordering
+    pub(super) pending: BTreeMap<PendingKey, DeployType>,
```
afck, line 28: A `BinaryHeap` might be slightly faster, and would avoid the need to duplicate the data in `PendingKey`. It also has a `retain` method that would simplify `prune_pending_deploys`.

node/src/components/block_proposer.rs
```diff
@@ -332,7 +336,26 @@ impl BlockProposerReady {
...
+                    Some(value) => value,
+                    None => {
+                        info!(
+                            "could not convert motes to gas for wasm deploy {} at gas price {}",
+                            hash, gas_price
+                        );
+                        return;
+                    }
+                }
+            };
```
afck, line 353: So `wasmless_transfer_cost` is currently hard-coded in _motes_, not _gas_? I guess we'll change that to be in gas, in the future, and multiply that by `gas_price` as well?

But for now, if we pay the same total amount for every wasmless transfer, shouldn't the gas price also be the same for all of them?

node/src/components/block_proposer.rs
```diff
@@ -343,17 +366,27 @@ impl BlockProposerReady {
...
+                .map(|(key, _)| *key);
+            match existing {
+                Some(key) => match self.sets.pending.remove(&key) {
+                    Some(deploy_type) => {
+                        self.sets
+                            .finalized_deploys
+                            .insert(deploy_hash, deploy_type.take_header());
+                    }
+                    None => {
+                        panic!("should be unreachable");
```
afck, line 383: You could use the `unreachable!` macro then, or just `expect`.

node/src/components/block_proposer/tests.rs
```diff
@@ -218,6 +260,47 @@ fn should_add_and_take_deploys() {
...
+fn should_not_add_wasm_deploy_with_zero_gas_price() {
+    let creation_time = Timestamp::from(100);
+    let ttl = TimeDiff::from(Duration::from_millis(100));
+
+    let mut rng = crate::new_rng();
+    let deploy1 = generate_deploy(&mut rng, creation_time, ttl, default_gas_payment(), 0);
+    let mut proposer = create_test_proposer(0);
+    // pending
+    proposer.add_deploy_or_transfer(creation_time, *deploy1.id(), deploy1.deploy_type().unwrap());
+    assert_eq!(proposer.sets.pending.len(), 0, "there should be 0 pending deploys - the one added has a gas price (willingness to pay) of 0");
```
afck, line 273: You can break string literals with a backslash:
```suggestion
    assert_eq!(proposer.sets.pending.len(), 0, "there should be 0 pending deploys - \
        the one added has a gas price (willingness to pay) of 0");
```

node/src/components/block_proposer/tests.rs
```diff
@@ -11,31 +13,36 @@ use crate::{
...
 
 use super::*;
-use casper_types::standard_payment::ARG_AMOUNT;
 
 fn default_gas_payment() -> Gas {
     Gas::from(1u32)
 }
 
+// gas_price used for test deploys
+const TEST_GAS_PRICE: u64 = 1;
```
afck, line 22: Might be better to use a gas price other than 1 in all the tests, to make it more likely that we'd detect if we forgot to multiply or divide somewhere.

node/src/components/block_proposer/tests.rs
```diff
@@ -430,50 +646,104 @@ fn should_not_propose_deploy_if_block_size_limit_passed() {
...
+    expected: TestValues,
+}
+
+impl Default for TestArgs {
+    fn default() -> Self {
+        TestArgs {
+            transfer_count: 0,
+            deploy_count: 0,
+            wasmless_transfer_cost: 1,
+            gas_price: 1u64,
```
afck, line 689: I'd also use a default other than 1 here.

node/src/components/block_proposer/tests.rs
```diff
@@ -483,112 +753,200 @@ fn test_proposer_with(
...
-        let deploy = generate_deploy(&mut rng, creation_time, ttl, vec![], payment_amount);
-        println!("generated deploy with size {}", deploy.serialized_length());
+        let deploy = generate_deploy(
+            &mut rng,
+            creation_time,
+            ttl,
+            deploy_payment_amount,
+            gas_price,
+        );
+        // println!("generated deploy with size {}", deploy.serialized_length());
```
afck, line 775: Are these println comments kept deliberately?

node/src/components/block_proposer.rs
dwerner: So the chainspec value for `wasmless_transfer_cost` is in fact gas, however the `payment_amount` here has to be stored in `Motes` by the `DeployType` such that it can then be scaled/converted from `Motes` by the specified `gas_price`.

afck: But which gas price is used to convert the gas to motes then?
It should be _this_ transfer's `gas_price`, shouldn't it?

node/src/components/block_proposer/deploy_sets.rs
dwerner: I guess it's a question of optimizing for deletion or for pruning. I could see how this could be done with an `Ord` impl and `BinaryHeap`, but one wrinkle in that plan is that `std::collections::BinaryHeap::retain` is still nightly-only. There's a `BinaryHeapPlus` crate that I was looking at, but I opted to go for the `BTreeMap` approach because it more resembled the existing `HashMap` interface that was here before.

node/src/components/block_proposer.rs
AlexanderLimonov: @afck, if payment_amount did what it's supposed to do (store a total payment amount...in motes), you could just sort it by that, instead of the _implied_ gas price. The problem seems to be that payment_amount attached to wasmless transfer deploys is in _gas_?

afck: It looks like wasmless transfers have an attached `gas_price` and a constant transfer cost (originally in gas), and the payment amount is their product.
The weird thing is that the `wasmless_transfer_cost` here is in motes, not gas, because it has been multiplied earlier (elsewhere in the code) by the formerly constant conversion rate?

node/src/components/block_proposer.rs
```diff
@@ -332,7 +336,26 @@ impl BlockProposerReady {
         if self.sets.finalized_deploys.contains_key(&hash) {
             info!(%hash, "deploy rejected from the buffer");
         } else {
-            self.sets.pending.insert(hash, deploy_or_transfer);
+            let gas_price = deploy_or_transfer.header().gas_price();
+            let payment_amount_gas = if deploy_or_transfer.is_transfer() {
+                Gas::from(self.wasmless_transfer_cost)
+            } else {
+                match Gas::from_motes(deploy_or_transfer.payment_amount_motes(), gas_price) {
+                    Some(value) => value,
```
AlexanderLimonov, line 344: @afck, everything seems right here, however?

- we get the gas price from the header
- IF a transfer, we convert u64 wasmless_transfer_cost we got from the chainspec to Gas (note that we don't really need to do that for the sort itself @dwerner)
- ELSE (IF regular deploy) we convert payment amount specified in the deploy to gas amount using the per unit price

So, now I'm really confused what the problem is supposed to be

node/src/components/block_proposer.rs
AlexanderLimonov: See comment on line 344 of block_proposer.rs

node/src/components/block_proposer.rs
afck: Ah, you're right, the sorting should actually be fine.
It's still kind of wrong that there is a constant `self.wasmless_transfer_cost` in motes, but it doesn't matter here.

AlexanderLimonov: @afck, it is not in motes. It's read from chainspec (which intends it to be gas, as stated there) as a u64, then, _for transfers_ converted to gas using `from`

`impl From<u32> for Gas {
    fn from(gas: u32) -> Self {
        let gas_u512: U512 = gas.into();
        Gas::new(gas_u512)
    }
}`

AlexanderLimonov: Note that the regular deploy conversion to Gas used `from_motes`, rather than `from`

afck: Ah, right! Sorry for the confusion!

PR #760, opened 2021-01-14
NDRS-735 - separate deploys and transfers in BlockHeader

dwerner:
This PR implements NDRS-735 - separate transfers and deploys in the block header.

### DIFF
--- a/node/src/types/block.rs
+++ b/node/src/types/block.rs
@@ -1154,6 +1154,7 @@ pub(crate) mod json_compatibility {
             &self.header.deploy_hashes
         }
 
+
         /// Returns the hashes of the transfer `Deploy`s included in the `Block`.
         pub fn transfer_hashes(&self) -> &Vec<DeployHash> {
             &self.header.deploy_hashes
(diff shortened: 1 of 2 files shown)


node/src/types/block.rs
```diff
@@ -560,6 +561,11 @@ impl BlockHeader {
         &self.deploy_hashes
     }
 
+    /// The list of deploy hashes included in the block.
```
Fraser999, line 564:
```suggestion
    /// The list of transfer hashes included in the block.
```

node/src/types/block.rs
```diff
@@ -1142,6 +1153,10 @@ pub(crate) mod json_compatibility {
         pub fn deploy_hashes(&self) -> &Vec<DeployHash> {
             &self.header.deploy_hashes
         }
+        /// Returns the hashes of the transfer `Deploy`s included in the `Block`.
```
Fraser999, line 1156:
```suggestion

        /// Returns the hashes of the transfer `Deploy`s included in the `Block`.
```

Fraser999 (commented):
Couple of nitpicks, otherwise LGTM.

Fraser999 (approved):

goral09:
This breaks clarity right? Since the serialization is changed.

dwerner:
> This breaks clarity right? Since the serialization is changed.

Good point @zie1ony - this may affect the event store.

PR #762, opened 2021-01-15
OP-1748: Support DNS resolution in new networking component

marc-casperlabs:
Tested this in a small local testnet, this lets me use DNS names in bind_address, public_address and known_addresses without issues here.

### DIFF
--- a/node/src/components/network.rs
+++ b/node/src/components/network.rs
@@ -700,13 +700,30 @@ fn handle_identify_event(swarm: &mut Swarm<Behavior>, event: IdentifyEvent) {
 /// "/ip4/127.0.0.1/tcp/34553".
 fn address_str_to_multiaddr(address: &str) -> Multiaddr {
     let mut parts_itr = address.split(':');
-    let multiaddr_str = format!(
-        "/ip4/{}/tcp/{}",
-        parts_itr.next().expect("address should contain IP segment"),
-        parts_itr
-            .next()
-            .expect("address should contain port segment")
-    );
+    let multiaddr_str = if address
+        .chars()
+        .next()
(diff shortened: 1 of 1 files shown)


rob-casper (approved):

node/src/components/network.rs
```diff
@@ -700,13 +700,30 @@ fn handle_identify_event(swarm: &mut Swarm<Behavior>, event: IdentifyEvent) {
...
-        "/ip4/{}/tcp/{}",
-        parts_itr.next().expect("address should contain IP segment"),
-        parts_itr
-            .next()
-            .expect("address should contain port segment")
-    );
+    let multiaddr_str = if address
+        .chars()
+        .next()
+        .expect("cannot convert emptry address")
```
Fraser999, line 706:
```suggestion
        .expect("cannot convert empty address")
```

node/src/components/network.rs
```diff
@@ -700,13 +700,30 @@ fn handle_identify_event(swarm: &mut Swarm<Behavior>, event: IdentifyEvent) {
...
+        .is_numeric()
```
Fraser999, line 707: This isn't robust: hostnames could begin with a digit.  I'm expecting this whole function to become obsolete once we remove the small_network component though and start using mutiaddr format in our config.  In which case, I'm ok with this for now.

Fraser999 (approved):
Optimistic approval given that we agree that `address_str_to_multiaddr` should be removed soon.

node/src/components/network.rs
marc-casperlabs: I've had the same thought a few hours after submitting this. I'm also in the "this should be gone soon" camp. Thanks!

PR #763, opened 2021-01-15
NO-TICKET: Initial kubernetes support

marc-casperlabs:
This has no impact on our core offering (the node) and can be gradually improved/replaced, so to avoid having a long running branch go stale, I propose to merge this to `master` and gradually phase pieces into our regular workflow, should we desire to do so.


xcthulhu (changes requested):
While this is a great way to ensure we have more determinism in our deploys, I suggest this be in a separate repo.

nix/shell.nix
```diff
@@ -0,0 +1,25 @@
...
+      inherit pname version;
+      sha256 = "1lri7a6pmlx9ghbrsgd702c3n862glwy0p8idh0lwdg313anmqwv";
+    };
+    doCheck = false;
+  };
+  python = pkgs.python3.withPackages
+    (python-packages: with python-packages; [ click kubernetes volatile ]);
+in pkgs.mkShell {
+
+  buildInputs = with pkgs; [ skopeo kubectl python black ];
```
xcthulhu, line 17: Why do we need `black`?  It's just a formatter, right?

nix/shell.nix
```diff
@@ -0,0 +1,25 @@
...
+    pname = "volatile";
+    version = "2.1.0";
+    src = pkgs.python38Packages.fetchPypi {
+      inherit pname version;
+      sha256 = "1lri7a6pmlx9ghbrsgd702c3n862glwy0p8idh0lwdg313anmqwv";
+    };
+    doCheck = false;
+  };
+  python = pkgs.python3.withPackages
+    (python-packages: with python-packages; [ click kubernetes volatile ]);
```
xcthulhu, line 14: It feels like this stuff should be in `default.nix`...

nix/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Building a docker image of a node
+
+To build your current source into a container image, enter a nix-shell in the same folder as this `README.md`, then run `build-node.sh`.
+
+```console
+$ nix-shell
+$ ./build-node.sh
+[...]
+Created new docker image casper-node:f2b9cd7a-dirty.
```
xcthulhu, line 17: Using a `nix-shell` here doesn't really follow best practices.  Is it a big lift to have `nix-build` create the local docker in `result`?

nix/README.md
```diff
@@ -0,0 +1,63 @@
...
+```console
+$ curl -L https://nixos.org/nix/install | sh
+```
+
+## Building a docker image of a node
+
+To build your current source into a container image, enter a nix-shell in the same folder as this `README.md`, then run `build-node.sh`.
+
+```console
+$ nix-shell
```
xcthulhu, line 14: Probably should have `nix-shell --pure` here, although see my comment below.

nix/README.md
```diff
@@ -0,0 +1,63 @@
...
+$ ./build-node.sh
+[...]
+Created new docker image casper-node:f2b9cd7a-dirty.
+
+Load into local docker:
+docker load -i ./result
+
+Publish image
+skopeo --insecure-policy copy docker-archive:./result docker://clmarc/casper-node:f2b9cd7a-dirty
```
xcthulhu, line 23: A few things:

  - Should we really be using `clmarc` on docker-hub?
  - Why `insecure-policy`?  Maybe there's no way around it...
  - Rather than have `skopeo` in `shell.nix`,  I think it's better to say:

```console
nix-shell --pure -p skopeo --command "skopeo --insecure-policy copy docker-archive:./result docker://clmarc/casper-node:f2b9cd7a-dirty"
```

default.nix
```diff
@@ -0,0 +1,39 @@
...
+  name = "rustenv";
+  buildInputs = with pkgs; [
+    zlib.dev
+    nodejs
+    cmake
+    pkg-config
+    openssl.dev
+    protobuf
+
+    # Dependencies just used in makefile
```
xcthulhu, line 16: I don't see a `Makefile`... is this TBD?

nix/shell.nix
marc-casperlabs: The nix scripts in general need a bit of cleanup, but here it is almost justifiable. My general idea is that `shell.nix` == convenient development environment, while `default.nix` (or in this case, `node.nix`) is the cleaner packaging.

I have "misused" the `shell.nix` to include tooling that is used to develop `casper-tool` as well. Short to midterm, I would like to remove some of these scripts altogether and have a unified "development environment for the whole project" `shell.nix` on the root, like I have on my local machine.

nix/shell.nix
marc-casperlabs: There is no `default.nix` right now. We're pending a decision on how to tackle the nix setup in general, especially regarding the container.

nix/README.md
marc-casperlabs: I usually don't use `--pure` for dev shells, i.e. when not testing packaging, but using nix-shell to make tools more available, since it removes my personal development tools from the path.

nix/README.md
marc-casperlabs: > Should we really be using clmarc on docker-hub?

Definitely not. This is due change soon.

> Why insecure-policy? Maybe there's no way around it...

It has zero effect, as far as I know. We're uploading a local file to a remote repo, no docker policies are involved.

> Rather than have skopeo in shell.nix, I think it's better to say: [...]

Possibly, but a lot less newbie friendly :) Arguably it's not even good to have these commands **echo**'d instead of offering a more integrated tool that does with a simple `upload` or even an `--upload` flag to `build-image.sh`.

This is all due to the fact that this tooling is currently in a lot of flux still.

default.nix
marc-casperlabs: The `default.nix` is my rough working cut of a nix file for the whole project. It could be removed instead. The `Makefile` here is thus the projects root `Makefile`.

marc-casperlabs:
Regarding @xcthulhu , this is a rough cut. Mainly the idea is to get the current state into master, albeit separate from everything else in `/nix` and `/utils/casper-tools`. There are roughly five different "sections" of the process that need to be worked upon separately, so this serves to set up an initial scaffolding only.

I could see removing `default.nix` though, that one slipped in by accident. On the other hand, I also don't mind keeping it in and improving on it.

marc-casperlabs:
I removed `default.nix`, leaving only an unsupported `shell.nix` at the root in. The remaining nix derivations (`node.nix`, `deps.nix` and `node-container.nix`) are soon to be replaced by SRE when the regular Docker tooling comes up.

rob-casper (approved):
Approving keeping in mind reasons already outlined (avoiding stale branch, to be gradually improved/replaced)

eg. hardcoded values such as network name and ports to be replaced with vars/script args

All looks self contained ✅

xcthulhu (approved):
We talked it all over and it appears we are going to go with a mono repo, so I'm marking this as approved.

nix/demo-cluster.py
```diff
@@ -0,0 +1,165 @@
...
+import time
+import tarfile
+
+import click
+import kubernetes
+import toml
+from kubernetes.client.rest import ApiException
+import volatile
+
+TAG = "9bddd925"
```
rob-casper, line 14: hardcoded tag

nix/forward-ports.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+kubectl -n casper-mynet port-forward casper-node-2 7777 &
```
rob-casper, line 3: hardcoded network name, ports

momipsl:
So there is history in the use of k8s and casper labs.  12 months ago when i started the stests software I took certification on the assumption that it would be a big part of the solution.  However SRE team at that time did not have any k8s skills and stonewalled it's usage.  Furthermore I proposed (twice) a k8s expert friend of mine as a prospective candidate to focus on this specific task but he was not engaged as SRE was in 'transition'.  So now that SRE has a deeper skill set I am glad to be seeing this kindof PR.

Upon reviewing the PR there is work to be done to be able leverage this to perform wider scenario testing as the stests software  attempts to do.  I would suggest the following:

1. Move this code into utils/ktests  (kubernetes tests)
2. Renamed utils/nctl -> utils/ltests (loca tests)
3. Fold the stests repo into utils/stests.

Then begin to retroactively integrate ktests into stests.  I.E. extend stests simulation workflow engine to mutate target nodeset whilst simaltaneously dispatching.  It should be able to do this against the terraform & k8s 'infra backends'.

marc-casperlabs:
Sounds good, let's make sure that that's capture. I'll be merging and improving upon this code in the meantime.

bors r+

PR #764, opened 2021-01-15
NO-TICKET: Better reporting of DNS failures

marc-casperlabs:
Improves the error reporting in case of DNS failures by ensuring that the failed lookup error includes the address in question.


Fraser999 (approved):
LGTM

PR #766, opened 2021-01-15
EE-1143: Use real gas price

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1143

### DIFF
--- a/node/src/components/block_proposer/tests.rs
+++ b/node/src/components/block_proposer/tests.rs
@@ -17,6 +17,10 @@ fn default_gas_payment() -> Gas {
     Gas::from(1u32)
 }
 
+fn default_gas_price() -> u64 {
+    1
+}
+
 fn generate_transfer(
     rng: &mut TestRng,
     timestamp: Timestamp,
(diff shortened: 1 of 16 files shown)


AlexanderLimonov:
Please see economics channel - without refunds, EE might not need to use gas price for anything, since the deploy header supplies the total charge. The tricky part is the payment code.

node/src/components/block_proposer/tests.rs
```diff
@@ -17,14 +17,18 @@ fn default_gas_payment() -> Gas {
     Gas::from(1u32)
 }
 
+fn default_gas_price() -> u64 {
```
Fraser999, line 20: Wouldn't it be normal to have this as a const?

Fraser999 (commented):
Now that we're actually using the gas price from the deploy, it might be worthwhile updating [the default value for it in the client](https://github.com/CasperLabs/casper-node/blob/master/client/src/deploy/creation_common.rs#L242) too?

dwerner (approved):

mpapierski:
> 
> 
> Now that we're actually using the gas price from the deploy, it might be worthwhile updating [the default value for it in the client](https://github.com/CasperLabs/casper-node/blob/master/client/src/deploy/creation_common.rs#L242) too?

Changed from 10 to 1 so people won't suddenly pay 10x times the cost since we're now using real gas price.

PR #768, opened 2021-01-15
NDRS-751 - Fetcher for ReadTrieResults

xcthulhu:


execution_engine/proptest-regressions/storage/trie/tests.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc 0a2fb8941a2bd26bca22e4f749e34f10cab194d7361bb1dd5e68ab6a1351145f # shrinks to trie = Leaf { key: Key::Account(0000000000000000000000000000000000000000000000000000000000000000), value: ContractPackage(ContractPackage { access_key: URef(0000000000000000000000000000000000000000000000000000000000000000, NONE), versions: {ContractVersionKey(1, 1): [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]}, disabled_versions: {}, groups: {} }) }
```
xcthulhu, line 7: @henrytill This is from a proptest for serializing `Trie<Key,StoredValue>`s with serde... should I delete this?

execution_engine/src/storage/global_state/mod.rs
```diff
@@ -47,7 +49,7 @@ pub enum CommitResult {
     Serialization(bytesrepr::Error),
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ReadTrieResult {
     pub trie_key: Blake2bHash,
     pub maybe_trie: Option<Trie<Key, StoredValue>>,
```
xcthulhu, line 55: @EdHastingsCasperLabs I wrote this prior to understanding how the fetcher works.  I need to know when I ask for a `trie_key` if I got the right value for that key.  If I got the wrong value, we should disconnect and if I got no value we should query another peer.

I'm thinking I can just use the `fetcher` Request/Response types instead of what I've done here.

execution_engine/proptest-regressions/storage/trie/tests.txt
henrytill: Yes

execution_engine/src/storage/global_state/mod.rs
EdHastingsCasperLabs: T could be Trie<Key, StoredValue> in which case the existing event set up for the generic fetcher should work as is.

henrytill: Return an `Option<Trie<Key, StoredValue>>` and put that directly into an `Event::GetFromStorageResult`.

xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/768/commits/0b5a8caa002d16b36f742017215a27919ce9314f

execution_engine/proptest-regressions/storage/trie/tests.txt
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/768/commits/1955493d98cba63ae656a53092aa56aa01264048

henrytill (approved):
Looks good here.

EdHastingsCasperLabs (approved):
double plus good

PR #769, opened 2021-01-16
OP-1650 Enabling cargo publish to crates.io in CI

sacherjj:
Adding cargo publish to .drone.yml and disabling non-tag workflows on tag.
Correcting grpc/server Cargo.toml to allow correct cargo publish
Adding explicit --token to publish.sh as it didn't honor cargo login in CI.

@Fraser999 Needed to use `--no-verify` with client publish.  Is this an issue?

Tested these changes in release-0.6.3 tagging and reworked until successful.

https://casperlabs.atlassian.net/browse/OP-1650

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -357,7 +357,7 @@ steps:
 - name: publish-crate
   image: casperlabs/node-build-u1804
   commands:
-    - "./ci/publish.sh"
+    - "./ci/publish_to_crates_io.sh"
   environment:
     CARGO_TOKEN:
       from_secret: crates_io_token
--- a/.drone.yml
+++ b/.drone.yml
@@ -42,6 +42,9 @@ trigger:
     - trying
     - staging
     - "release-*"
+  event:
+    exclude:
+      - tag
 
 ---
 # Failure state from pre-checks pipeline
@@ -72,6 +75,9 @@ trigger:
     - trying
     - staging
     - "release-*"
+  event:
+    exclude:
+      - tag
 
 depends_on:
   - pre-checks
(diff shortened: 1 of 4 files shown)


thamps-casp (approved):

rob-casper (approved):

Fraser999:
> Needed to use --no-verify with client publish. Is this an issue?

I'm a bit concerned about that, yes.  If you have a copy of the build failure which required the use of `--no-verify`, we could maybe try and fix the problem?

sacherjj:
@Fraser999 This is what I received:
```

Finished dev [unoptimized + debuginfo] target(s) in 2m 10s
--
674 | error: failed to verify package tarball
675 |  
676 | Caused by:
677 | Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
678 | Added: /drone/src/target/package/casper-client-0.6.3/headers
679 | /drone/src/target/package/casper-client-0.6.3/headers/casper_client.h
680 |  
681 | To proceed despite this, pass the `--no-verify` flag.

```

Fraser999:
@sacherjj Thanks.  It was indeed FFI related.  Our build script invoked cbindgen which generated our C header inside our source tree (at `client/headers/casper_client.h`).  `cargo publish` saw this as pollution.

[This PR](https://github.com/sacherjj/casper-node/pull/1) changes the build script to generate the header at `target/headers/casper_client.h`, and while it's not strictly inside `OUT_DIR` as the error message recommends, it is at least in our build tree rather than source tree, and `cargo publish` seems happy with this.

sacherjj:
I'll pull this out before merging and we will test it with next release branch.

PR #771, opened 2021-01-17
NO-TICKET: add wasm-strip as an optional post-build step to each contract in the Makefile

Fraser999:
This adds a call to [wasm-strip](https://webassembly.github.io/wabt/doc/wasm-strip.1.html) after each contract is built.  The call is ignored if it fails, meaning that wasm-strip is not required to be available on the build machine.

Examining the total size of wasm files in my build folder using
```
find target/wasm32-unknown-unknown/release -type f -name '*.wasm' -exec du -ch {} + | grep total$
```
I see the total drop by ~92.5% from 75M to 5.6M.


dwerner (approved):
Should we add a note to the readme indicating wasm-strip as an optional build dependency?

henrytill (approved):
LGTM

PR #776, opened 2021-01-19
NDRS-758: don't hold ExecutableDeployItem args in serialized form

Fraser999:
This updates the `args` fields of all `ExecutableDeployItem` variants to be of type `RuntimeArgs` rather than a serialized form of that.

This is a more appropriate way to hold the args in general, but it also makes for a more legible JSON-encoded form.  Rather than e.g.
```
"args": "02000000070000006e616d655f3031010000000000070000006e616d655f3032040000002a00000001"
```
we now have:
```
"args": [
  [
    "name_01",
    {
      "cl_type": "Bool",
      "bytes": "00",
      "parsed": false
    }
  ],
  [
    "name_02",
    {
      "cl_type": "I32",
      "bytes": "2a000000",
      "parsed": 42
    }
  ]
]
```

Tagging @zie1ony, @AbnerZheng and @momipsl as an alert to the minor API change.  Please also note that `serialized_bytes` has been renamed to `bytes` and `parsed_to_json` renamed to `parsed`.


client/lib/deploy.rs
```diff
@@ -202,46 +202,80 @@ mod tests {
...
+          "hash": "09dcee4b212cfd53642ab323fbef07dafafc6f945a80a00147f62910a915c4e6",
+          "version": null,
+          "entry_point": "entrypoint",
+          "args": [
+            [
+              "name_01",
+              {
+                "cl_type": "Bool",
+                "serialized_bytes": "00",
+                "parsed_to_json": false
```
henrytill, line 229: Seems oddly self-referential in name.  Maybe just `parsed`?

client/lib/deploy.rs
```diff
@@ -202,46 +202,80 @@ mod tests {
...
+        "StoredVersionedContractByHash": {
+          "hash": "09dcee4b212cfd53642ab323fbef07dafafc6f945a80a00147f62910a915c4e6",
+          "version": null,
+          "entry_point": "entrypoint",
+          "args": [
+            [
+              "name_01",
+              {
+                "cl_type": "Bool",
+                "serialized_bytes": "00",
```
henrytill, line 228: Maybe just `bytes`?

henrytill (approved):
Excellent.  A couple non-blocking opinions - curious what @EdHastingsCasperLabs thinks.

EdHastingsCasperLabs (approved):
All in all this seems like a smart change to the system for a lot of reasons. Gold star.

client/lib/deploy.rs
EdHastingsCasperLabs: in this context, I wouldn't blink if you just called it "value".

Fraser999: Renamed.

client/lib/deploy.rs
Fraser999: Renamed.

PR #777, opened 2021-01-19
NO-TICKET: CLValue to JSON

zie1ony:
Full implementation of CLValue to JSON parser. It is used to produce `parsed_to_json` field when querying RPC for a CLValue.


Fraser999 (changes requested):
I created https://github.com/zie1ony/casper-node/pull/1 rather than doing a review to speed up the process.

This PR is very welcome - I wish I hadn't written off the possibility to convert complex CLTypes to JSON so readily!  Nice one :)

Fraser999 (approved):

PR #778, opened 2021-01-19
NO TICKET - direct queues to dump to /tmp with timestamp

dwerner:
This PR directs the queue dump debug mechanism to `/tmp` because, when `cwd` is not set to a writable directory, we simply crash. This also appends a timestamp on the filename.

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -46,18 +46,19 @@ use futures::{future::BoxFuture, FutureExt};
 use once_cell::sync::Lazy;
 use prometheus::{self, Histogram, HistogramOpts, IntCounter, Registry};
 use quanta::IntoNanoseconds;
+use serde::Serialize;
+use tokio::time::{Duration, Instant};
 use tracing::{debug, debug_span, info, trace, warn};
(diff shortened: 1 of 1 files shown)


Fraser999 (approved):

node/src/reactor.rs
```diff
@@ -406,8 +408,8 @@ where
                 .await
                 .expect("could not serialize snapshot");
 
-            let mut file =
-                File::create("queue_dump_debug.txt").expect("could not create dump file");
```
marc-casperlabs, line 410: If queue dumps are large, they might fill up `/tmp`, which may be limited to 1-2 GB in size on many distros using tmpfs.

Should we maybe try an `or_else` here? We try to dump to the local file first, if that fails try dumping to `/tmp`, if that fails panic?

We could also just log our failure to dump as an error, allowing the user to clear up space/whatever and send the SIGUSR1 again.

marc-casperlabs (commented):
Offering some feedback.

marc-casperlabs (approved):

PR #781, opened 2021-01-20
NDRS-761: Parameterize `FetchResult` with peer type.

goral09:
`LinearChainSync` component uses `FetchResult`'s `peer` but it was hardcoded to `NodeId` while the whole component operates with abstract type `I`. This is a vestigial element – everywhere else peer type is parameterized.


afck (approved):
How did that even work before? Was `FetchResult` never used in any context where `I` wasn't `NodeId`?

goral09:
bors merge

PR #784, opened 2021-01-20
NDRS-760: Track which peers had the data and prioritize them.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-760

Track which peers we successfully have downloaded data from and prioritize them when we request the next piece of data.

Every 5 successful attempts we will try 1 new peer (which is not in the set of the successful peer). If it has the data, it will be added to the set.

### DIFF
--- a/node/src/components/linear_chain_sync/peers.rs
+++ b/node/src/components/linear_chain_sync/peers.rs
@@ -1,3 +1,5 @@
+use std::collections::VecDeque;
+
 use datasize::DataSize;
 use rand::{seq::SliceRandom, Rng};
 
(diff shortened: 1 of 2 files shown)


fizyk20 (approved):

node/src/components/linear_chain_sync/peers.rs
```diff
@@ -26,7 +36,12 @@ impl<I: Clone + PartialEq + 'static> PeersState<I> {
 
     /// Returns a random peer.
     pub(crate) fn random(&mut self) -> Option<I> {
-        self.peers_to_try.pop()
+        if self.succ_attempts < self.succ_attempts_max {
+            self.next_succ().or_else(|| self.peers_to_try.pop())
+        } else {
+            self.succ_attempts = 0;
+            self.peers_to_try.pop()
```
afck, line 43: If there are no more `peers_to_try`, shouldn't we use one from `succ_peers` anyway?

goral09: Oh, right. Good catch.

goral09: Done in [`8b25d5d` (#784)](https://github.com/CasperLabs/casper-node/pull/784/commits/8b25d5d0e84fc319c6f387104c1885d09e610158)

afck (approved):

goral09:
bors merge

PR #785, opened 2021-01-20
HWY-{209,230,231}: Remove blocking loops in the synchronizer, add timeout.

afck:
This breaks up the blocking loop that recursively adds all vertices with satisfied dependencies. Instead, only one vertex is added per event handler call, and a new event is scheduled — as a timer event with 0 delay — to add the next one.

It also adds a timeout to entries in the synchronizer queues, and periodically removes entries older than the timeout.

Finally, it makes the synchronizer more conservative when an invalid vertex is detected: We shouldn't remove a pending vertex just because _one_ of its senders is faulty.

https://casperlabs.atlassian.net/browse/HWY-209
https://casperlabs.atlassian.net/browse/HWY-230
https://casperlabs.atlassian.net/browse/HWY-231


goral09:
You could shortly describe, in the PR's description, what's the approach you took to solve this problem.

xcthulhu:
It's not obvious why we need to schedule an event.  There are events which don't have timers on them, right?

afck:
@xcthulhu: Yes, that's true. It would add another `ProtocolOutcome`, but maybe it would be cleaner.
In this PR, I just abused the timer and set the timestamp to "now" to get the same result.
I'll look into it, maybe I can fix it today.

goral09:
That was my idea – to add a `ProtocolOutcome` variant, like `Requeue(Vertex<C>)`.

afck:
@goral09, @xcthulhu: I pushed a commit that separates immediate actions from timers.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -42,6 +43,14 @@ use crate::{
 /// even if eras are longer than this.
 const MAX_ENDORSEMENT_EVIDENCE_LIMIT: u64 = 10000;
 
+/// The timer for creating new units, as a validator actively participating in consensus.
+const TIMER_ID_ACTIVE_VALIDATOR: TimerId = TimerId(0);
+/// The timer for adding a vertex with a future timestamp.
+const TIMER_ID_VERTEX_WITH_FUTURE_TIMESTAMP: TimerId = TimerId(1);
```
xcthulhu, line 49: I think I am missing something.  Why can't we have `enum TimerId = { ActiveValidator, VertexWithFutureTimestamp }`?  This way the compiler can tell us if we haven't exhaustively handled all of the cases in our code.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -42,6 +43,14 @@ use crate::{
...
+
+/// The action of adding a vertex from the `vertices_to_be_added` queue.
+const ACTION_ID_VERTEX: ActionId = ActionId(0);
```
xcthulhu, line 52: If we decide to have `TimerId` be an enum, maybe it makes sense for this to be an enum too?

node/src/components/consensus/protocols/highway.rs
afck: It would make the `ConsensusProtocol` trait Highway-specific.
I guess we don't care much about that anymore, though, so I could do that if you prefer.

goral09: I don't think we should be turning it into an enum – like you said, `ConsensusProtocol` is still unaware of Highway details. Let's leave it that way.

afck:
This is only half of HWY-209 — the timeout for the pending vertices, but not the maximum count. I'm marking it "ready for review" for now anyway, because I won't be able to add the other half before Monday, and it could be done in a separate PR.

afck:
I also extended `is_empty` to take into account the other collections as well, not just `vertex_deps` in [this commit](https://github.com/CasperLabs/casper-node/pull/785/commits/1d5ceb87f4ecce728dac2c35062b671dd150c119#diff-1accc7f900f13303b5cf3a898818d9de5dac9ad6796476c57415e32d072f0868R218-R222). (Sorry, should probably have been a separate commit.)

resources/local/config.toml
```diff
@@ -37,6 +37,8 @@ secret_key_path = 'secret_key.pem'
 # The folder in which the files with per-era latest unit hashes will be stored.
 unit_hashes_folder = "../node-storage"
 
+# The duration for which incoming vertices with missing dependencies should be kept in a queue.
+pending_vertex_timeout = '1day'
```
goral09, line 41: Really, that long? Even if eras are supposed to be 30minutes?

afck: That's a good point.
Do we have test data on how long it will take to fully sync a complete era's protocol state?
Should I just make it `'30min'`?

goral09: Why not use `era_duration` for that?

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -0,0 +1,258 @@
...
+            vertices_to_be_added_later: BTreeMap::new(),
+            vertices_to_be_added: Vec::new(),
+            pending_vertex_timeout,
+        }
+    }
+
+    /// Removes expired pending vertices from the queues, and schedules the next purge.
+    pub(crate) fn purge_vertices(&mut self) {
+        let timeout = self.pending_vertex_timeout;
+        self.vertices_to_be_added.retain(|pv| !pv.expired(timeout));
```
goral09, line 99: Do we need to iterate through this collection? The comment says:
```
Vertices that might be ready to add to the protocol state: We are not currently waiting for a requested dependency.
```
which means that we're not blocked by a spammy peer and we have all the information necessary to add it.

afck: The vertices in that queue could still be missing dependencies: It's just that the last dependency we knew of was satisfied, so we'll try again when the event fires. That might reveal more missing dependencies and add them back to `vertex_deps`.

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -0,0 +1,258 @@
...
+        }
+        let keys = self
+            .vertices_to_be_added_later
+            .iter()
+            .filter(|(_, pvs)| pvs.is_empty())
+            .map(|(key, _)| *key)
+            .collect_vec();
+        for key in keys {
+            self.vertices_to_be_added_later.remove(&key);
+        }
```
goral09, line 115: Can you extract this to a method, something like `remove_empty`?

resources/local/config.toml
afck: Last week we had discussed that we'd make it configurable.
I don't think it should always just be the `era_duration`. It would be better if we could just measure how long syncing a full era takes in practice, and use that value.

node/src/components/consensus/protocols/highway/synchronizer.rs
goral09: Ah, ok. I haven't really checked how it gets populated.

node/src/components/consensus/protocols/highway/synchronizer.rs
afck: Done in https://github.com/CasperLabs/casper-node/pull/785/commits/a15f38b782047d23687b10e79743c7e136d92b15.

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub(crate) fn purge_vertices(&mut self) {
+        let timeout = self.pending_vertex_timeout;
+        self.vertices_to_be_added.retain(|pv| !pv.expired(timeout));
+        for pvs in self
+            .vertices_to_be_added_later
+            .values_mut()
+            .chain(self.vertex_deps.values_mut())
+        {
+            pvs.retain(|pv| !pv.expired(timeout));
+        }
```
goral09, line 106: What do you think about a method that removes expired entries from a collection of `PendingVertex`? It would replace those two and will reduce `purge_vertices` to:
```rust
Self::remove_expired(self.vertices_to_be_added);
self
  .vertices_to_be_added_later
  .values_mut()
  .chain(self.vertex_deps.values_mut())
  .for_each(Self::remove_expired)
```

afck: Doesn't work quite like that; I think the most readable solution might be to just extract the whole logic of removing expired entries _and_ then removing empty lists from _maps_. I'll give that a try.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -538,27 +442,41 @@ where
...
+                let now = Timestamp::now();
                 match pvv.timestamp() {
-                    Some(timestamp) if timestamp > Timestamp::now() => {
+                    Some(timestamp)
+                        if timestamp > now + self.synchronizer.pending_vertex_timeout() =>
+                    {
+                        trace!("received a vertex with a timestamp far in the future; dropping");
+                        vec![]
+                    }
+                    Some(timestamp) if timestamp > now && is_faulty => {
```
goral09, line 453: Why does it matter that a message is from the future? If a validator is faulty but the message is from "now", we would add it?

goral09: Apparently we wouldn't: https://github.com/CasperLabs/casper-node/pull/785/files#diff-a2c951f6efd58f4890ffbf04a54d3527c0ebbe0d985089ed8e5df0b6d2c61688R475-R478 

If that's true then I would remove the `timestamp > now` condition as it makes it look like it matters.

afck: I didn't actually change that part of the logic in this PR.
I think the reason may be that we had always added all incoming vertices, and then added logic for handling vertices with future timestamp. When we did that, we decided that faulty validators don't deserve putting vertices with future timestamps into the queue.

But you're right, it would make sense to me to just always reject vertices from faulty validators, except if they are dependencies of other queue entries. I'll do that in this PR, if that's okay.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -774,7 +710,9 @@ where
     }
 
     fn has_received_messages(&self) -> bool {
-        !self.highway.state().is_empty() || !self.vertex_deps.is_empty()
+        !self.highway.state().is_empty()
+            || !self.synchronizer.is_empty()
+            || !self.pending_values.is_empty()
```
goral09, line 715: In which scenario we would have `pending_values` but nothing in the `synchronizer`?

afck: I kept `pending_values` in `HighwayProtocol` itself. So:
If someone sends us a vertex with a block, it first goes into the synchronizer, until all dependencies are satisfied. Then it goes out of the synchronizer and into `pending_vertices` until the block validator responds.

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -229,11 +229,8 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
         self.pending_vertex_timeout
     }
 
-    pub(crate) fn drop_dependent_vertices(
-        &mut self,
-        mut vertices: Vec<Dependency<C>>,
-    ) -> HashSet<I> {
-        let mut senders = HashSet::new();
+    pub(crate) fn drop_dependent_vertices(&mut self, mut vertices: Vec<Dependency<C>>) -> Vec<I> {
```
goral09, line 209: Can you add a comment to this method? It accepts a vector of vertices but returns a vector of peers. I actually forgot why and it's not obvious from the method body itself.

goral09 (approved):
LGTM, although it got even more complex and given that we don't have much tests, it may be difficult to catch subtler problems.

afck:
> although it got even more complex and given that we don't have much tests, it may be difficult to catch subtler problems

I agree! I'm hesitant to also add the entry count limit now. ([HWY-237](https://casperlabs.atlassian.net/browse/HWY-237))
I'd rather work on the design first ([HWY-151](https://casperlabs.atlassian.net/browse/HWY-151)) and maybe add more `HighwayProtocol` tests. I added [HWY-238](https://casperlabs.atlassian.net/browse/HWY-238) for it, but I'd rather do it in a separate PR, because it will take a few days.

afck:
bors merge

PR #789, opened 2021-01-20
NO-TICKET: Verify hashes for blocks from the wire

xcthulhu:


xcthulhu:
As far as I can tell, it's possible that the block coming over the wire has the wrong hash.  This enforces that the header has the hash we requested and the block has the correct hash listed.

goral09 (approved):
Good catch!

afck (approved):

PR #790, opened 2021-01-21
NO-TICKET: Logging erroneous synchronizer starts

xcthulhu:


goral09 (commented):
I don't think this is necessary. Previous version was effectively the same. It's incorrect to receive `Start` event in all 3 states. You can change the `trace` to `error` if you prefer.

xcthulhu:
> You can change the `trace` to `error` if you prefer.

Is it an error in the `None` case?

goral09:
> > You can change the `trace` to `error` if you prefer.
> 
> Is it an error in the `None` case?

Hmm, rather unlikely. That could happen if between the moment `LinearChainSync` is initialized and the moment `joiner` reactor transitions to `validator`, `LinearChainSync` component receives a `NewPeer` event and then sends `Start` to itself…

OK, even though it sounds as an indication of some problem it's not strictly speaking an error. You're right.

goral09 (approved):

afck (approved):

PR #791, opened 2021-01-21
NO-TICKET: Create all parents of unit_hash_file

xcthulhu:


xcthulhu:
@henrytill I bumped into this bug when I was running the `run-dev-tmux.sh` script.  I think you might have run into it too?  I've tested this locally and it appears to work again.

fizyk20 (approved):
I really should have done that when I was adding this functionality :+1: I haven't because this was meant to be kept in the same directory as the storage, and storage creates its own directories already - could be useful in other cases, though.

xcthulhu:
bors merge

PR #794, opened 2021-01-21
EE-1155: Auction use account's main purse

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1155

- Removes purse arguments from the API (bonding purse, target purse, unbonding purse, etc)
- Explicit "get_main_purse" call is limited to minimum only in add_bid/delegate as that still has to take funds from the user and put it in separate purse.
- "unbonding_purse" is removed from `UnbondingPurse` data structure and instead we're transferring funds straight to unbonder's account.


types/src/auction/detail.rs
```diff
@@ -146,14 +147,19 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
...
             if current_era_id >= unbonding_purse.era_of_creation() + unbonding_delay {
+                let account_hash =
+                    AccountHash::from_public_key(unbonding_purse.unbonder_public_key(), |x| {
+                        provider.blake2b(x)
+                    });
+
                 // Move funds from bid purse to unbonding purse
-                provider
-                    .transfer_from_purse_to_purse(
+                let _transferred_to = provider
```
henrytill, line 156: It is more idiomatic not to assign this when we are calling for effect.

smart_contracts/contracts/test/auction-bids/src/main.rs
```diff
@@ -57,13 +53,10 @@ fn delegate() {
...
         ARG_DELEGATOR => delegator,
         ARG_VALIDATOR => validator,
-        ARG_SOURCE_PURSE => account::get_main_purse(),
         ARG_AMOUNT => amount,
     };
 
-    let (purse, _amount): (URef, U512) = runtime::call_contract(auction, METHOD_DELEGATE, args);
-
-    runtime::put_key(DELEGATE_PURSE, purse.into())
+    let (_purse, _amount): (URef, U512) = runtime::call_contract(auction, METHOD_DELEGATE, args);
```
henrytill, line 59: Let's also remove these values, as we are just calling for effect.

(and below)

smart_contracts/contracts/client/add-bid/src/main.rs
```diff
@@ -4,26 +4,17 @@
 extern crate alloc;
 
 use auction::DelegationRate;
-use casper_contract::{
-    contract_api::{account, runtime, system},
-    unwrap_or_revert::UnwrapOrRevert,
-};
-use casper_types::{auction, runtime_args, PublicKey, RuntimeArgs, URef, U512};
+use casper_contract::contract_api::{runtime, system};
+use casper_types::{auction, runtime_args, PublicKey, RuntimeArgs, U512};
```
henrytill, line 8: While we are here, can we put `auction::DelegationRate` in the `casper_types` stanza?

henrytill (approved):
LGTM.  Please address comments before merging.

PR #796, opened 2021-01-22
EE-1157: Transfer to account support in AS.

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1157
Ref: https://casperlabs.atlassian.net/browse/EE-1113


smart_contracts/contract_as/assembly/purse.ts
```diff
@@ -200,3 +219,53 @@ export function transferFromPurseToPurse(sourcePurse: URef, targetPurse: URef, a
...
+
+/**
+ * Transfers `amount` of motes from main purse purse to `target` account.
+ * If `target` does not exist it will be created.
+ *
+ * @param amount Amount is denominated in motes
+ * @returns This function will return a [[TransferredTo.TransferError]] in
+ * case of transfer error, in case of any other variant the transfer itself
+ * can be considered successful.
+ * @hidden
```
henrytill, line 231: Should this public?

henrytill (approved):
LGTM, but perhaps `transferToAccount` should be public?

smart_contracts/contract_as/assembly/purse.ts
mpapierski: Done

PR #797, opened 2021-01-22
NDRS-766: introducing joiner-dev-tmux.sh

xcthulhu:


xcthulhu:
@EdHastingsCasperLabs did some follow-up with @marc-casperlabs so that we can have proper unit tests here, but this is sufficient for now for my purposes.

EdHastingsCasperLabs (approved):

henrytill (changes requested):
I we should remove the generated joiner `chainspec.toml` and add it to `.gitignore`.

goral09:
You should really try using `nctl`. It's already in this repo, it doesn't add any overhead since it's mostly bash scripts. It already has features like starting the network (`nctl-start`), stopping the network (`nctl-stop`), start/stop single node (`nctl-start|stop node=X`), joining the network using latest available hash (`nctl-join node=6`), joining the network using specific hash (`nctl-start node=6 hash=<hash>`) and much more.

henrytill (approved):
LGTM

PR #798, opened 2021-01-22
NO-TICKET: Log the era supervisor's public key

xcthulhu:


xcthulhu:
I found this helpful in debugging logs for the local network runners.

afck (commented):
We could add it to [this](https://github.com/CasperLabs/casper-node/blob/7efb351a0384b7d6d8e2b1ac556fc22ee1e7fffc/node/src/components/consensus/era_supervisor.rs#L263) and maybe [this](https://github.com/CasperLabs/casper-node/blob/7efb351a0384b7d6d8e2b1ac556fc22ee1e7fffc/node/src/components/consensus/era_supervisor.rs#L260) log message instead, when activating as a validator. (Because for read-only nodes — which never activate — the public signing key is not relevant.)

It would also be good to have a consistent terminology: It's called `our_id` in [other log messages](https://github.com/CasperLabs/casper-node/blob/7efb351a0384b7d6d8e2b1ac556fc22ee1e7fffc/node/src/components/consensus/era_supervisor.rs#L257).

Fraser999 (approved):

afck (approved):

xcthulhu:
> We could add it to [this](https://github.com/CasperLabs/casper-node/blob/7efb351a0384b7d6d8e2b1ac556fc22ee1e7fffc/node/src/components/consensus/era_supervisor.rs#L263) and maybe [this](https://github.com/CasperLabs/casper-node/blob/7efb351a0384b7d6d8e2b1ac556fc22ee1e7fffc/node/src/components/consensus/era_supervisor.rs#L260) log message instead, when activating as a validator. (Because for read-only nodes — which never activate — the public signing key is not relevant.)
> 
> It would also be good to have a consistent terminology: It's called `our_id` in [other log messages](https://github.com/CasperLabs/casper-node/blob/7efb351a0384b7d6d8e2b1ac556fc22ee1e7fffc/node/src/components/consensus/era_supervisor.rs#L257).

I went ahead and made these changes in https://github.com/CasperLabs/casper-node/pull/798/commits/82574cf2135bd83cba544d9ee739224a29e0c81d and https://github.com/CasperLabs/casper-node/pull/798/commits/252c09c0c76450d7fbac69feab3c0112d576aa14

xcthulhu:
bors merge

PR #800, opened 2021-01-22
NO_TICKET - panic when block-proposer state is corrupted

dwerner:
This PR causes the node to panic if we load a corrupted state for the block-proposer. This is a short-term fix to prevent a node from starting in a bad state.

### DIFF
--- a/node/src/effect.rs
+++ b/node/src/effect.rs
@@ -1106,7 +1106,10 @@ impl<REv> EffectBuilder<REv> {
         .transpose()
         .unwrap_or_else(|err| {
             let type_name = type_name::<T>();
-            panic!("could not deserialize state from storage type name {:?} err {:?}", type_name, err);
+            panic!(
+                "could not deserialize state from storage type name {:?} err {:?}",
+                type_name, err
+            );
         })
     }
 
--- a/node/src/effect.rs
+++ b/node/src/effect.rs
@@ -1106,8 +1106,7 @@ impl<REv> EffectBuilder<REv> {
         .transpose()
         .unwrap_or_else(|err| {
             let type_name = type_name::<T>();
-            warn!(%type_name, %err, "could not deserialize state from storage");
-            None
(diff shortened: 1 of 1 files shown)


goral09 (approved):
LGTM but make sure @marc-casperlabs checks it as well :+1:

xcthulhu (approved):

node/src/effect.rs
```diff
@@ -1106,8 +1106,10 @@ impl<REv> EffectBuilder<REv> {
         .transpose()
         .unwrap_or_else(|err| {
             let type_name = type_name::<T>();
-            warn!(%type_name, %err, "could not deserialize state from storage");
-            None
+            panic!(
```
marc-casperlabs, line 1109: This is not logic limited to the block proposer (well, right now it is, but it is in a generic function).

I think we have two cases in general here, components that are fine with not having state restored (because it is cache anyway, or they can recover from it), and those that absolutely need it.

If we want to support both, we need to return a `Result<Option<_>>` instead of an `Option<_>` in the effect instead and introduce a proper error type.

On the other hand, we can say that data corruption is no joke and we will always crash if it happens. In this case, this is okay.

I would suggest we keep the log  though and panic with a slightly redundant message after, just to have an entry in the JSON log? Unless I am out of the loop and we implemented JSON-panics.

Finally is it possible to call `.fatal` instead of panic'ing here? That one is minor though.

dwerner: > On the other hand, we can say that data corruption is no joke and we will always crash if it happens. In this case, this is okay.

@fizyk20 or @goral09 I'll let you decide but I think this is the point of the PR.

If this were more than a quick patch I'd be fine with going to greater lengths, but I think this is extraneous for a quick fix.

goral09: If we ever decide that we don't have to panic when deserializing fails then we can improve/change it.

goral09:
bors merge

PR #803, opened 2021-01-24
NDRS-785: abbreviate large string JSON fields

Fraser999:
This provides a `json_pretty_print()` function in casper-types which replaces large (over 100 char) strings with `x chars` where `x` is the number of chars in the field.

The client has been updated to use this function by default.  However, if the user specifies extra verbosity by passing the `-v` arg twice (i.e. `-vv` or `-v -v`) then the full hex-encoded bytes of all values will be shown, as they are currently.

Example output of a `put-deploy` request using `-v`:

```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "account_put_deploy",
  "params": {
    "deploy": {
      "approvals": [
        {
          "signature": "130 chars",
          "signer": "01f60bce2bb1059c41910eac1e7ee6c3ef4c8fcc63a901eb9603c1524cadfb0c18"
        }
      ],
      "hash": "5c61b4096243a8bfe89f9337c60c8276a598dfdad94ccbb0eb9497bd614a0e14",
      "header": {
        "account": "01f60bce2bb1059c41910eac1e7ee6c3ef4c8fcc63a901eb9603c1524cadfb0c18",
        "body_hash": "89f13f2b89fc304b849ed047e3d5c6c1fc2e7ad81619e02bd9558028451eae3a",
        "chain_name": "net",
        "dependencies": [],
        "gas_price": 1,
        "timestamp": "2021-01-24T15:35:26.295Z",
        "ttl": "1h"
      },
      "payment": {
        "ModuleBytes": {
          "args": [],
          "module_bytes": "1308110 chars"
        }
      },
      "session": {
        "ModuleBytes": {
          "args": [],
          "module_bytes": "176328 chars"
        }
      }
    }
  }
}
```

### DIFF
--- a/client/src/get_auction_info.rs
+++ b/client/src/get_auction_info.rs
@@ -32,13 +32,14 @@ impl<'a, 'b> ClientCommand<'a, 'b> for GetAuctionInfo {
     fn run(matches: &ArgMatches<'_>) {
         let maybe_rpc_id = common::rpc_id::get(matches);
         let node_address = common::node_address::get(matches);
-        let verbose = common::verbose::get(matches);
+        let mut verbosity_level = common::verbose::get(matches);
 
-        let response = casper_client::get_auction_info(maybe_rpc_id, node_address, verbose)
+        let response = casper_client::get_auction_info(maybe_rpc_id, node_address, verbosity_level)
             .unwrap_or_else(|error| panic!("response error: {}", error));
-        println!(
-            "{}",
-            serde_json::to_string_pretty(&response).expect("should encode to JSON")
-        );
+
+        if verbosity_level == 0 {
+            verbosity_level += 1
+        }
+        casper_client::pretty_print_at_level(&response, verbosity_level);
     }
 }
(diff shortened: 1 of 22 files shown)


client/src/get_auction_info.rs
```diff
@@ -32,13 +32,14 @@ impl<'a, 'b> ClientCommand<'a, 'b> for GetAuctionInfo {
...
+        let response = casper_client::get_auction_info(maybe_rpc_id, node_address, verbosity_level)
             .unwrap_or_else(|error| panic!("response error: {}", error));
-        println!(
-            "{}",
-            serde_json::to_string_pretty(&response).expect("should encode to JSON")
-        );
+
+        if verbosity_level == 0 {
+            verbosity_level += 1
+        }
```
goral09, line 42: Here, and all other places, why do you increment the verbosity level?

goral09 (approved):

client/src/get_auction_info.rs
Fraser999: The client prints the response regardless of what verbosity level the user requests - it's pretty much all it ever does with the responses.  So the verbosity level of `0` lets the user specify not to print the request, but we need to set it to `1` to get the response printed.

There might be a neater way to do this - I didn't like mutating a user-set config value either, but I didn't want to make it overly complex.

PR #804, opened 2021-01-24
NO-TICKET: Memory usage of `BlockValidator`

goral09:
I've added `data_size(skip)` annotation on `Loading` variant of the enum, mostly because we shouldn't be in the `Loading` state for long enough for this metric to be necessary but also because it is a pain to write an implementation of `DataSize`  for the `Event` enum.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -34,7 +34,7 @@ steps:
     commands:
       - cargo install cargo-audit
       - cargo generate-lockfile
-      - cargo audit
+      - cargo audit --ignore RUSTSEC-2021-0013
 
 trigger:
   branch:
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -68,9 +68,10 @@ pub(crate) struct BlockValidationState<T> {
     responders: SmallVec<[Responder<(bool, T)>; 2]>,
 }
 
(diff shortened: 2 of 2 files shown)


goral09:
@Fraser999 @marc-casperlabs I am having troubles with `cargo-audit` steps in this PR: https://drone-auto.casperlabs.io/CasperLabs/casper-node/3992/1/5 . Can you help me?

Fraser999:
It looks like our dependency on `raw-cpuid` v8.1.2 via `quanta` in casper-node is to blame here.  There's [a dependabot PR](https://github.com/metrics-rs/quanta/pull/37) to quanta which updates its dependency and would probably fix this, but that PR is failing CI currently.

Until that PR is merged, we might be best to change the audit step in .drone.yml to
```
cargo audit --ignore RUSTSEC-2021-0013
```
although we need to remember to re-enable it as soon as we can.

goral09:
Thanks @Fraser999 for taking care of it :) 

I will update the drone's instruction and create a follow-up ticket.

goral09:
JIRA ticket https://casperlabs.atlassian.net/browse/NDRS-788

afck (approved):

goral09:
bors merge

PR #805, opened 2021-01-25
NDRS-773/yamux buffer size, cannot reproduce

marc-casperlabs:
This contains more than one thing actually:

* Changes the log level of libp2p by default (NDRS-787)
* Adds tests simulating what was _suspected_ to be the issue with the networking component (large payloads), but was not.
* Fixes a bug in the reactor macro that prevented it from instantiating components with construction effects.

Most importantly, it changes the behavior of the receive buffer negotation from the default of `WindowUpdateMode::OnReceive` to `WindowUpdateMode::OnRead`


node/src/components/network.rs
```diff
@@ -8,10 +8,11 @@ mod peer_discovery;
 mod protocol_id;
 #[cfg(test)]
 mod tests;
+#[cfg(test)]
+mod tests2;
```
Fraser999: I guess a more meaningful name would be good here.  Or would renaming the old module be more appropriate?

node/src/components/network.rs
```diff
@@ -104,10 +106,14 @@ enum ConnectionState {
 }
 
 #[derive(DataSize)]
-pub(crate) struct Network<REv, P> {
+pub struct Network<REv, P> {
```
Fraser999, line 109: Do we _really_ need this exposed in our API?

node/src/components/network.rs
```diff
@@ -104,10 +106,14 @@ enum ConnectionState {
...
     our_id: NodeId,
     #[data_size(skip)]
     peers: HashMap<NodeId, ConnectedPoint>,
+    /// A set of peers whose address we currently know. Kept in-sync with the internal Kademlia
+    /// routing table.
+    #[data_size(skip)]
+    seen_peers: HashSet<PeerId>,
```
Fraser999, line 116: This could use more documentation.  I'm not seeing why we need this and also the `peers` member.

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,238 @@
+use std::{
```
EdHastingsCasperLabs, line 1: nitpick...I'm not loving this trend I've noticed of suffixing filenames vs giving them more verbose names.

marc-casperlabs: Good point. I'll rename it. I actually would want to potentially `test2` with `test`, as they are similar enough, the separate implementation was done in the name of expediency.

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,240 @@
...
+impl Collectable for DummyPayload {
+    type CollectedType = DummyPayloadId;
+
+    fn into_collectable(self) -> Self::CollectedType {
+        self.id()
+    }
+}
+
+impl Debug for DummyPayload {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
Fraser999: Let's import `Formatter` please.

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,240 @@
...
+    // and connecting.
+    let timeout = Duration::from_secs(60);
+    let large_size: usize = 1024 * 1024 * 4;
+
+    let mut rng = crate::new_rng();
+
+    // Port for first node, other will connect to it.
+    let first_node_port = testing::unused_port_on_localhost() + 1;
+
+    let mut net = testing::network::Network::<LoadTestingReactor>::new();
```
Fraser999: Should import `Network`.

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,240 @@
...
+    let chainspec = Chainspec::random(&mut rng);
+
+    // Create the root node.
+    let cfg = TestReactorConfig {
+        chainspec: chainspec.clone(),
+        network_config: crate::components::network::Config::default_local_net_first_node(
```
Fraser999: Should import `network::self` so we can call `network::Config`, or import `Config as NetworkConfig`.

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        net.add_node_with_config(cfg, &mut rng).await.unwrap();
+    }
+
+    info!("Network setup, waiting for discovery to complete");
+    net.settle_on(&mut rng, network_online, timeout).await;
+    info!("Discovery complete");
+
+    // At this point each node has at least one other peer. Assuming no split, we can now start
+    // gossiping a large payloads. We gossip one on each node.
```
Fraser999:
```suggestion
    // gossiping large payloads. We gossip one on each node.
```

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,240 @@
...
+            others_received(dummy_payload.id(), sender.clone()),
+            timeout,
+        )
+        .await;
+        info!(?sender, "Completed gossip test for sender")
+    }
+}
+
+/// Checks if all nodes are connected to at least one other node.
+pub fn network_online(
```
Fraser999: Does this need to be `pub`?

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,240 @@
...
+    let known_nodes_target = (k_value / 2).min(nodes.len() - 1);
+
+    // Checks if all nodes have reached the known nodes target.
+    nodes
+        .values()
+        .all(|runner| runner.reactor().inner().net.seen_peers().len() >= known_nodes_target)
+}
+
+/// Checks whether or not every node except `sender` on the network received the given payload.
+pub fn others_received(
```
Fraser999: Does this need to be `pub`?

node/src/components/network/tests2.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Sanity check of K_VALUE.
+    assert!(
+        k_value >= 7,
+        "K_VALUE is really small, expected it to be at least 7"
+    );
+
+    // The target of known nodes to go for. This has a hard bound of `K_VALUE`, since if all nodes
+    // end up in the same bucket, we will start evicting them. In general, we go for K_VALUE/2 for
+    // reasonable interconnection, or the network size - 1, which is another bound.
+    let known_nodes_target = (k_value / 2).min(nodes.len() - 1);
```
Fraser999: For larger networks, this number seems pretty small.  As you mentioned - in a very badly imbalanced network, all nodes could end up in a single k-bucket's address space (although this should be _highly_ improbable).  But in that case, every peer should eventually be connected to `k` peers I believe.

So, I'd suggest at least changing `(k_value / 2)` to just `k_value`.

Fraser999 (commented):
I was under the impression that the yamux buffer was potentially overwhelmed not so much by a small number of large messages (as the test here replicates) but more by a large number of messages.

Perhaps the test could be refactored to send many messages in a tight loop to see if we can reproduce the observed error?

node/src/components/network.rs
marc-casperlabs: Been renamed!

node/src/components/network.rs
marc-casperlabs: Yes. The macro does not work for components that are just `pub(crate)` due to a shortcoming in the Rust compiler that requires a workaround.

node/src/components/network/tests2.rs
marc-casperlabs: Highly disagree. The component being named `Network` and the `testing::network::Network` being named, well `Network`, is extremely confusing and renaming it to `TestNetwork` is even worse.

These is only a single place where this is type is used as well!

node/src/components/network/tests2.rs
marc-casperlabs: See above.

node/src/components/network/tests2.rs
marc-casperlabs: Fixed!

node/src/components/network/tests2.rs
marc-casperlabs: Fixed!

node/src/components/network.rs
marc-casperlabs: `peers` tracks connections, which can be arbitrarily opened and closed by libp2p. `seen_peers` tracks peers we know about, regardless if we have ever been or are connected to them. The latter is used by the test to check if node discovery works. This would also be a worthwhile metric to expose.

marc-casperlabs:
> I was under the impression that the yamux buffer was potentially overwhelmed not so much by a small number of large messages (as the test here replicates) but more by a large number of messages.

Yamux operates on a bytestream level, so I would assume that 100 messages of 10 bytes each would be the same as a single message of 1000 bytes? I did not check, but I think framing happens on a higher level, as yamux only knows streams.

> Perhaps the test could be refactored to send many messages in a tight loop to see if we can reproduce the observed error?

It could, the same reactor and primitives could be used. There's other conceivable scenarios, at this point, I see a value but not a burning need for it though. Did you review already include the commit which contained https://github.com/CasperLabs/casper-node/pull/805/files#diff-ced0e709fad62f0c013bb3d691fb1082aface83249abccb628224b31e62bec65R205-R228 ?

marc-casperlabs:
I should add, I successfully ran the test on my machine with 50 nodes, 4 MB payloads, using both the old and new `WindowUpdateMode` setting.

node/src/components/network.rs
Fraser999: Could you add the critical distinction to the comment then please?  (As in, we may never have been connected to these peers, whereas `peers` represents the currently-connected list of peers)

node/src/components/network/tests_bulk_gossip.rs
```diff
@@ -0,0 +1,240 @@
...
+    // and connecting.
+    let timeout = Duration::from_secs(60);
+    let large_size: usize = 1024 * 1024 * 4;
+
+    let mut rng = crate::new_rng();
+
+    // Port for first node, other will connect to it.
+    let first_node_port = testing::unused_port_on_localhost() + 1;
+
+    let mut net = testing::network::Network::<LoadTestingReactor>::new();
```
Fraser999, line 144: I understand that you highly disagree, but we've already discussed this and we have a style guide to avoid wasting time on issues like this.  Please amend as per the coding standards and the rest of the codebase.

node/src/components/network/tests_bulk_gossip.rs
```diff
@@ -0,0 +1,240 @@
...
+    let chainspec = Chainspec::random(&mut rng);
+
+    // Create the root node.
+    let cfg = TestReactorConfig {
+        chainspec: chainspec.clone(),
+        network_config: crate::components::network::Config::default_local_net_first_node(
```
Fraser999, line 150: Similarly here.

node/src/components/network/tests_bulk_gossip.rs
```diff
@@ -0,0 +1,240 @@
...
+        ),
+    };
+
+    net.add_node_with_config(cfg, &mut rng).await.unwrap();
+
+    // Create `node_count-1` additional node instances.
+    for _ in 1..node_count {
+        let cfg = TestReactorConfig {
+            chainspec: chainspec.clone(),
+            network_config: crate::components::network::Config::default_local_net(first_node_port),
```
Fraser999, line 161: And here.

node/src/components/network/tests_bulk_gossip.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Sanity check of K_VALUE.
+    assert!(
+        k_value >= 7,
+        "K_VALUE is really small, expected it to be at least 7"
+    );
+
+    // The target of known nodes to go for. This has a hard bound of `K_VALUE`, since if all nodes
+    // end up in the same bucket, we will start evicting them. In general, we go for K_VALUE/2 for
+    // reasonable interconnection, or the network size - 1, which is another bound.
+    let known_nodes_target = (k_value / 2).min(nodes.len() - 1);
```
Fraser999, line 217: There's an unanswered comment on this line in the old file.  Basically, I feel this `(k_value / 2)` might be better as `k_value`, or did you already try that?

Fraser999 (commented):
> Yamux operates on a bytestream level, so I would assume that 100 messages of 10 bytes each would be the same as a single message of 1000 bytes? I did not check, but I think framing happens on a higher level, as yamux only knows streams.

Well, I take your point.  But on the other hand, we can't reproduce the issue, and we don't really know if this test is replicating the scenario in which the log messages appeared or not.  We do definitively know that we weren't sending 4MB messages there though.

node/src/components/network/tests_bulk_gossip.rs
marc-casperlabs: I did not, yet. I'll follow up on this.

node/src/components/network/tests_bulk_gossip.rs
marc-casperlabs: "Fixed".

node/src/components/network/tests2.rs
marc-casperlabs: I'm following up on this in a separate PR.

PR #806, opened 2021-01-25
NDRS-783: Track Trusted Block Header in Fast Sync

xcthulhu:


node/src/components/linear_chain_fast_sync.rs
```diff
@@ -131,28 +138,22 @@ impl<I: Clone + PartialEq + 'static> LinearChainFastSync<I> {
...
                 highest_block_seen,
                 trusted_hash,
+                trusted_header,
                 ref latest_block,
                 validator_weights,
                 ..
-            } => {
+            } if highest_block_seen == block_height => {
+                // TODO: Fail gracefully in these cases
                 assert_eq!(highest_block_seen, block_height);
```
afck, line 157: That assertion is redundant now, since it's in the `match` condition itself.

afck (approved):

node/src/components/linear_chain_fast_sync.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/806/commits/240258f15750eccd6ad0a95086463f1d75674af8

PR #807, opened 2021-01-25
NDRS-793 - use jemallocator and dump queues when under memory pressure

dwerner:
This PR is intended to implement NDRS-793 - we will dump the contents of the message queues when memory utilization reaches an optional value set by and env var `CL_MEM_DUMP_THRESHOLD_MB` (in megabytes).

This is done in-process and brings in `jemalloc` as the global allocator, from which we extract the amount of memory that the node has allocated up to that point. We get the `total` memory available by bringing in the `sys-info` crate. This PR sets the groundwork upon which we can build out more robust memory metrics and tuning using `jemalloc`.


node/src/reactor.rs
```diff
@@ -223,6 +227,9 @@ where
 
     /// An accurate, possible TSC-supporting clock.
     clock: Clock,
+
+    /// Last queue dump timestamp
+    last_queue_dump: Option<Timestamp>,
```
goral09, line 232: Since you don't make a use of the value of the last queue dump, this _could_ be just a boolean, but I understand that having the actual timestamp may be more useful in the future.

node/src/reactor.rs
```diff
@@ -443,9 +464,42 @@ where
...
+
+        match epoch::mib() {
+            Ok(mib) => {
+                mib.advance().unwrap();
+            }
+            Err(error) => {
+                tracing::error!(%error, "unable to get epoch::mib from jemalloc");
+                return None;
+            }
+        }
```
goral09, line 487: Can you add a comment why we have to do that?

node/src/reactor.rs
```diff
@@ -443,9 +464,42 @@ where
...
+        let allocated = match stats::allocated::mib() {
+            Ok(allocated_mib) => allocated_mib.read().unwrap(),
```
goral09, line 489: So, `sys_info` wouldn't give us precise measurements of the used memory?

dwerner: We could get stats from sys-info, and that could be enough in this case. My desire is to build out a more robust set of metrics on top of jemalloc going forward as well.

node/src/reactor.rs
goral09: I found the answer in the documentation of `jemalloc-ctl`:
> many statistics are cached and only updated when the epoch is advanced.

goral09:
Looking at the [Examples](https://docs.rs/jemalloc-ctl/0.3.3/jemalloc_ctl/#examples) section in the documentation of `jemalloc-ctl`, it looks like they're defining `Jemalloc` to be a global allocator before taking these measurements:
```rust
#[global_allocator]
static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
```

Will it work if we don't do that?

node/src/reactor.rs
dwerner: This seems to be an artifact of the internal implementation in `jemalloc_ctl` [here](https://docs.rs/jemalloc-ctl/0.3.3/jemalloc_ctl/stats/struct.allocated.html#method.mib) - I really just wanted to play nice and prevent a crash from `unwrap()`.

dwerner: Oh I understand, you mean the `epoch::advance()` call. Sorry

goral09: I meant the `mib.advance()` part. It wasn't clear to me why we have to do that before reading the allocated memory. The documentation had an answer, can you add it? I  mean the part about caching statistics.

dwerner:
> Looking at the [Examples](https://docs.rs/jemalloc-ctl/0.3.3/jemalloc_ctl/#examples) section in the documentation of `jemalloc-ctl`, it looks like they're defining `Jemalloc` to be a global allocator before taking these measurements:
> 
> ```rust
> #[global_allocator]
> static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
> ```
> 
> Will it work if we don't do that?

This is done somewhat implicitly by the `jemallocator-global` crate.

goral09:
> > Looking at the [Examples](https://docs.rs/jemalloc-ctl/0.3.3/jemalloc_ctl/#examples) section in the documentation of `jemalloc-ctl`, it looks like they're defining `Jemalloc` to be a global allocator before taking these measurements:
> > ```rust
> > #[global_allocator]
> > static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
> > ```
> > 
> > 
> > Will it work if we don't do that?
> 
> This is done somewhat implicitly by the `jemallocator-global` crate.

Are there any disadvantages when using `jemalloc` as a default allocator?

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
...
+                tracing::error!(%error, "unable to get mem_info using sys-info");
+                return None;
+            }
+        };
+
+        let total = mem_info.total;
+
+        match epoch::mib() {
+            Ok(mib) => {
+                // jemalloc_ctl requires you to advance the epoch to update it's stats
```
goral09, line 481:
```suggestion
                // jemalloc_ctl requires you to advance the epoch to update its stats
```

goral09 (approved):
The code looks good to me but I can't tell if using `jemalloc` as a default allocator is a good thing or not. I can read about it tomorrow but if @marc-casperlabs or @Fraser999 approve it, given their bigger knowledge & experience in this than mine, then you can go ahead and merge it :)

dwerner:
> Are there any disadvantages when using this allocator?

For quite some time, `jemalloc` was rust's default allocator (see the [docs](https://doc.rust-lang.org/edition-guide/rust-2018/platform-and-target-support/global-allocators.html) discussing that here). In my own personal experience it's better at handling highly fragmented memory scenarios, and can be instrumented and tuned pretty extensively. FreeBSD uses it as libc's `malloc`.

I'm not aware of any drawbacks of using it over the system allocator.

node/src/reactor.rs
```diff
@@ -392,6 +400,19 @@ where
                 self.reactor.update_metrics(event_queue);
                 self.last_metrics = now;
             }
+
+            if self.last_queue_dump.is_none() {
+                if let Some((allocated, total)) = Self::get_allocated_memory() {
+                    if allocated >= (total as f32 * MEM_DUMP_THRESHOLD) as u64 {
+                        tracing::info!(
```
Fraser999, line 407:
```suggestion
                        info!(
```

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
         self.event_count += 1;
     }
 
+    /// Get both the allocated and total memory from sys-info + jemalloc
```
Fraser999, line 467:
```suggestion
    /// Gets both the allocated and total memory from sys-info + jemalloc
```

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
...
+    fn get_allocated_memory() -> Option<(u64, u64)> {
+        let mem_info = match sys_info::mem_info() {
+            Ok(mem_info) => mem_info,
+            Err(error) => {
+                tracing::error!(%error, "unable to get mem_info using sys-info");
```
Fraser999, line 472:
```suggestion
                error!(%error, "unable to get mem_info using sys-info");
```

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
         self.event_count += 1;
     }
 
+    /// Get both the allocated and total memory from sys-info + jemalloc
+    fn get_allocated_memory() -> Option<(u64, u64)> {
```
Fraser999, line 468: I'd rather see a small struct holding these two values rather than a tuple, especially since the two values are of the same type.

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
...
+
+        let total = mem_info.total;
+
+        match epoch::mib() {
+            Ok(mib) => {
+                // jemalloc_ctl requires you to advance the epoch to update its stats
+                mib.advance().unwrap();
+            }
+            Err(error) => {
+                tracing::error!(%error, "unable to get epoch::mib from jemalloc");
```
Fraser999, line 485:
```suggestion
                error!(%error, "unable to get epoch::mib from jemalloc");
```

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
...
+                return None;
+            }
+        }
+        let allocated = match stats::allocated::mib() {
```
Fraser999, line 489: Let's import `jemalloc_ctl::stats::allocated` please.

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
...
+            Ok(allocated_mib) => allocated_mib.read().unwrap(),
```
Fraser999, line 490: Should we really be unwrapping here?

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
...
+            Err(error) => {
+                tracing::error!(%error, "unable to get active mib using jemalloc");
```
Fraser999, line 492:
```suggestion
                error!(%error, "unable to get active mib using jemalloc");
```

node/src/reactor.rs
```diff
@@ -443,9 +464,43 @@ where
...
+                return None;
+            }
+        };
+
+        Some((allocated as u64, total))
+    }
+
     // Handles dumping queue contents to files in /tmp.
```
Fraser999, line 500:
```suggestion
    /// Handles dumping queue contents to files in /tmp.
```

node/src/reactor.rs
```diff
@@ -60,6 +61,9 @@ use crate::{
 use quanta::Clock;
 pub use queue_kind::QueueKind;
 
+/// Default threshold for amount of total ram remaining before dumping queues to disk.
+const MEM_DUMP_THRESHOLD: f32 = 0.25;
```
Fraser999, line 65: Would it be useful to have this as a config value?  I can see a need to tweak this value somewhat as we get more experience.

node/src/reactor.rs
```diff
@@ -60,6 +61,9 @@ use crate::{
 use quanta::Clock;
 pub use queue_kind::QueueKind;
 
+/// Default threshold for amount of total ram remaining before dumping queues to disk.
```
Fraser999, line 64:
```suggestion
/// Lower threshold for fraction of total RAM remaining before dumping queues to disk.
```

Fraser999 (commented):
I don't have a lot of experience in tinkering with global allocators, so I have similar concerns as @goral09, but I am good with your answers.

I would like to know that this PR actually does as advertised though.  Have you tested it manually?  You can presumably test this by setting the threshold value to something close to 1.

node/src/reactor.rs
dwerner: I'm not clear on how I can access the R::Config value from this point of view - perhaps I can work on this in a follow-up PR. I also tried plumbing through an environment variable, similar to `DISPATCH_EVENT_THRESHOLD`  but was stuck on getting nctl to pick up on it.

goral09: >  was stuck on getting nctl to pick up on it.

Instead of exporting the value, try supplying it in the same line as the nctl command, like this `MEM_DUMP_THRESHOLD=0.1 nctl-start node=6`

Fraser999: Right - I see the issue now.  I'd be fine with making it an env var for now, and we can look to make it a config value proper in a future PR if required.

node/src/reactor.rs
```diff
@@ -60,6 +61,9 @@ use crate::{
 use quanta::Clock;
 pub use queue_kind::QueueKind;
 
+/// Lower threshold for percentage of total RAM allocated before dumping queues to disk.
```
Fraser999, line 64:
```suggestion
/// Upper threshold for percentage of total RAM allocated before dumping queues to disk.
```

node/src/reactor.rs
```diff
@@ -443,33 +470,76 @@ where
...
+    fn get_allocated_memory() -> Option<AllocatedMem> {
+        let mem_info = match sys_info::mem_info() {
+            Ok(mem_info) => mem_info,
+            Err(error) => {
+                error!(%error, "unable to get mem_info using sys-info");
+                return None;
+            }
+        };
+
+        // mem_info gives us kb
```
Fraser999, line 483:
```suggestion
        // mem_info gives us kB
```

node/src/reactor.rs
```diff
@@ -443,33 +470,76 @@ where
...
+                if let Err(advance_error) = mib.advance() {
+                    error!(%advance_error, "unable to advance jemalloc epoch");
+                }
+            }
+            Err(error) => {
+                error!(%error, "unable to get epoch::mib from jemalloc");
+                return None;
+            }
+        }
+        let allocated = match jemalloc_allocated::mib() {
```
Fraser999, line 498: Does this value represent kB rather than bytes too?  If so, we need to multiply it by 1024 too.

Either way, it'd be good to at least document the units used for each field of the `AllocatedMem` struct I guess.

Fraser999 (commented):
I'm still not clear on whether you actually saw this code triggering an event queue dump.

node/src/reactor.rs
```diff
@@ -443,33 +470,76 @@ where
...
+    /// Handles dumping queue contents to files in /tmp.
+    async fn dump_queues(&mut self) {
         let timestamp = Timestamp::now();
+        self.last_queue_dump = Some(timestamp);
         let output_fn = format!("/tmp/queue_dump-{}.json", timestamp);
         let mut serializer = serde_json::Serializer::pretty(match File::create(&output_fn) {
             Ok(file) => file,
             Err(error) => {
-                tracing::error!(%error, "could not create output file ({}) for queue snapshot", output_fn);
+                error!(%error, "could not create output file ({}) for queue snapshot", output_fn);
```
marc-casperlabs, line 523: This should be a warning, not an error.

node/src/reactor.rs
```diff
@@ -443,33 +470,76 @@ where
...
                 return;
             }
         });
 
         if let Err(error) = self.scheduler.snapshot(&mut serializer).await {
-            tracing::error!(%error, "could not serialize snapshot to {}", output_fn);
+            error!(%error, "could not serialize snapshot to {}", output_fn);
```
marc-casperlabs, line 529: This _could_ be an error, given it is not likely to ever work if it failed once, but even here I'd consider a warning.

node/src/reactor.rs
```diff
@@ -443,33 +470,76 @@ where
...
             return;
         }
 
         let debug_dump_filename = format!("/tmp/queue_dump_debug-{}.txt", timestamp);
         let mut file = match File::create(&debug_dump_filename) {
             Ok(file) => file,
             Err(error) => {
-                tracing::error!(%error, "could not create debug output file ({}) for queue snapshot", debug_dump_filename);
+                error!(%error, "could not create debug output file ({}) for queue snapshot", debug_dump_filename);
```
marc-casperlabs, line 537: Also a warning at most.

node/src/reactor.rs
```diff
@@ -443,33 +470,76 @@ where
...
                 return;
             }
         };
         if let Err(error) = self.scheduler.debug_dump(&mut file).await {
-            tracing::error!(%error, "could not serialize debug snapshot to {}", debug_dump_filename);
+            error!(%error, "could not serialize debug snapshot to {}", debug_dump_filename);
```
marc-casperlabs, line 542: See above :)

node/src/reactor.rs
goral09: I actually think it's better as an env var. If we put it in the chainspec then it will appear in validators' configs and they will start asking, tweaking etc. This is for debugging purposes mostly. Let's make it an env var and default to some reasonable value.

node/src/reactor.rs
```diff
@@ -191,6 +195,11 @@ pub trait Finalize: Sized {
     }
 }
 
+struct AllocatedMem {
```
marc-casperlabs, line 198: This struct needs documentation, badly. Especially since the units aren't in the member names.

node/src/reactor.rs
```diff
@@ -60,6 +61,9 @@ use crate::{
 use quanta::Clock;
 pub use queue_kind::QueueKind;
 
+/// Lower threshold for percentage of total RAM allocated before dumping queues to disk.
+const MEM_DUMP_THRESHOLD: f32 = 75.0;
```
marc-casperlabs, line 65: An absolute, configurable value would be better. When running on a cluster, to the best of my knowledge, total memory might appear a lot larger than the allocated/limited memory, e.g. given an allocation and hard limit of 500M for a node, on a machine that has 16G of memory, it should show 16G of total mem even though the process will be killed at 500.1M of memory usage.

node/src/app/cli.rs
```diff
@@ -21,6 +21,9 @@ use casper_node::{
 };
 use prometheus::Registry;
 
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
```
marc-casperlabs, line 25: It would be nice to have a little comment on why we decided to swap out the allocator :)

marc-casperlabs (commented):
There are some changes I'd like to see.

dwerner:
> I'm still not clear on whether you actually saw this code triggering an event queue dump.

Sorry, I forgot to reply to that. I tested this manually (in it's current form for % of total memory) with a MEM_DUMP_THRESHOLD such that dumps would trigger when the allocated usage hit 32mb.

node/src/reactor.rs
dwerner: Fair point, perhaps this should be an absolute value, but optional? @goral09

node/src/reactor.rs
dwerner: It's bytes, but I've documented that here and in the struct definition now.

node/src/reactor.rs
goral09: Maybe, I think that absolute value will be a pain to set for node operators or debugging. BUT if we can't have reliable readings with percentage then so it be. It has to be correct first and foremost.

dwerner: Well another issue with percentage is that the total is -system- total, so there may be other processes that add to memory pressure on a single box... leading to the OOM killer stepping in before we actually hit the percentage limit.

goral09: I thought that using different allocator will help us measure our own usage better and the total is the same regardless of the load on the machine.

dwerner:
> > I'm still not clear on whether you actually saw this code triggering an event queue dump.

Manually tested and verified that the dumps happen when the optional CL_MEM_DUMP_THRESHOLD_MB is hit, and are not happening when the env var isn't defined.

node/src/reactor.rs
dwerner: That's correct, my point was just that other processes may also be consuming memory, so acting on a ratio of `allocated/total` would ignore those other processes, and potentially you could OOM without dumping.

goral09: Ah, right. Yes I was afraid of that as well. My "solution" was that we should run 1 binary per machine and not put more load on it that could skew the results.

goral09 (approved):

Fraser999 (approved):

PR #809, opened 2021-01-26
NDRS-801: Verify that the execution matches expected during sync.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-801

The bug was introduced in #760 , where wasmless transfers got separated from the deploys but this change was not propagated when constructing a `ProtoBlock`. Which lead to a consensus not requesting wasmless transfers when fetching dependencies and, in a result, executing an incomplete block (hence different post-state hash).


afck (approved):
Good catch!

goral09:
bors merge

dwerner (approved):

PR #812, opened 2021-01-26
NO-TICKET: Fix raw gas/motes conversions.

mpapierski:
There are some checks in the node (i.e. https://github.com/CasperLabs/casperlabs-node/blob/b992ae0f72fa7a5c72c101dca03e0271a5736481/node/src/components/block_proposer.rs#L454-L463) to avoid overflows so chances are bogus deploys with `gas_price == u64::max_value()` or similar won't reach EE but after merging #766 those are better to be matched rather than unwrap/expected.


henrytill (approved):
Good catch.

As discussed offline, lets drop some TODOs in to remind us to eventually put in an overflow error variant.

PR #813, opened 2021-01-27
NO TICKET: Qualifying `Error` in `effects.rs`

xcthulhu:


xcthulhu:
This unqualified `Error` was a head scratcher because there's a few other errors floating around in this file.

@Fraser999 I've tagged you because I'm not sure what the convention is.  Rest of this file does qualified errors like `execution_state::Error`.  Is that right?

Fraser999 (approved):
Yes, this is the convention where there is ambiguity about a type.  `Error` is probably the most common such case.

PR #816, opened 2021-01-27
NDRS-791: Introducing effect for putting a trie into the trie store

xcthulhu:


xcthulhu:
@henrytill we're going to need to keep track of the key tries are inserted under, since fast sync can only know its done if there's no pending tries we are waiting to be inserted.

I had the EE also return any trie keys missing beneath the trie we insert, since we'll need to queue them up in fast sync to be fetched from over the wire.

execution_engine/src/core/engine_state/put_trie.rs
```diff
@@ -0,0 +1,33 @@
+use crate::shared::newtypes::Blake2bHash;
+
+/// Result from inserting a [Trie][storage::trie::Trie] into the trie store.  Contains:
+///
+/// - The trie key the inserted [Trie][storage::trie::Trie] (which is its [Blake2bHash])
+///
+/// - The missing descendant keys of that trie transitively refers to. Found using
+///   [storage::global_state::StateProvider::missing_trie_keys].
+#[derive(Debug)]
+pub struct InsertedTrieKeyAndMissingDescendants {
```
henrytill, line 10: LMAO

henrytill (approved):

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1635,17 +1637,26 @@ where
             .map_err(Error::from)
     }
 
-    pub fn put_trie(
+    pub fn put_trie_and_find_missing_descendant_trie_keys(
```
afck, line 1640: I wonder whether `put_trie_and_find_missing_descendant_keys` or `put_trie_and_find_missing_descendants` would be any less clear.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1635,17 +1637,26 @@ where
...
+    ) -> Result<InsertedTrieKeyAndMissingDescendants, Error>
     where
         Error: From<S::Error>,
     {
-        self.state
+        let inserted_trie_key = self
+            .state
             .put_trie(correlation_id, trie)
-            .map_err(Error::from)
+            .map_err(Error::from)?;
```
afck, line 1651: Is `map_err(Error::from)` necessary? Doesn't `?` do that automatically? (Also below.)

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1635,17 +1637,26 @@ where
...
+        let missing_descendant_trie_keys = self
+            .state
+            .missing_trie_keys(correlation_id, inserted_trie_key)
+            .map_err(Error::from)?;
```
afck, line 1655: Should it even be possible to insert a trie with missing descendants? Or should that wait in some queue until we have all descendants?

execution_engine/src/core/engine_state/put_trie.rs
```diff
@@ -0,0 +1,33 @@
+use crate::shared::newtypes::Blake2bHash;
+
+/// Result from inserting a [Trie][storage::trie::Trie] into the trie store.  Contains:
+///
+/// - The trie key the inserted [Trie][storage::trie::Trie] (which is its [Blake2bHash])
+///
+/// - The missing descendant keys of that trie transitively refers to. Found using
```
afck, line 7:
```suggestion
/// - The missing descendant keys which that trie transitively refers to. Found using
```

execution_engine/src/core/engine_state/put_trie.rs
afck: That sounds like it's the results from already _inserting_ the missing descendants.

execution_engine/src/core/engine_state/mod.rs
xcthulhu: > Should it even be possible to insert a trie with missing descendants?

The `put_trie` methods makes this possible yeah.

> Or should that wait in some queue until we have all descendants?

The idea is to have a queue.

[This unit test in the EE][1] demonstrates the loop for how fast sync is going to use.

[1]: http://github.com/xcthulhu/casper-node/blob/master/execution_engine/src/storage/trie_store/operations/tests/synchronize.rs#L52-L80

execution_engine/src/core/engine_state/mod.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/816/commits/262770cabe6535ee8aa7fc93217a86d562039a4e

execution_engine/src/core/engine_state/put_trie.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/816/commits/c8693d3a44055a74062b62864eafbfd2eea7190e

xcthulhu:
bors merge

PR #825, opened 2021-01-28
NDRS-797: Initial work on casper-node upgrades

Fraser999:
This PR lays the groundwork for adapting the casper-node to properly handle protocol upgrades.

The first commit adds new two new subcommands to the node: `migrate-config` and `migrate-data`.  The former will be run post-install to migrate any config changes.  The latter will be run by the casper-node-launcher just before running the node in normal mode to migrate stored data as required.

The second commit refactors storage to handle providing switch blocks by era ID.  This is required as part of the upgrade story as a newly-started node will need to be able to determine the current list of validators and their key weights.

However, this commit is also required as part of the task which @fizyk20 is currently working on, hence this PR only covers part of the changeset required to handle node upgrades.

### DIFF
--- a/node/src/app/cli.rs
+++ b/node/src/app/cli.rs
@@ -4,7 +4,11 @@
 
 pub mod arglang;
 
-use std::{env, fs, path::PathBuf, str::FromStr};
+use std::{
+    env, fs,
+    path::{Path, PathBuf},
+    str::FromStr,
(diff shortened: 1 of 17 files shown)


fizyk20 (approved):

Fraser999:
bors merge

PR #826, opened 2021-01-28
NO-TICKET: enable libp2p component by default

Fraser999:
This PR reverses the behaviour of which networking component is chosen by default: now libp2p is the default, and users need to set the env var `CASPER_ENABLE_LEGACY_NET` to enable the old `small_network` component.

### DIFF
--- a/node/src/components/network.rs
+++ b/node/src/components/network.rs
@@ -67,8 +67,8 @@ use crate::{
     NodeRng,
 };
 
-/// Env var which, if it's defined at runtime, enables the libp2p server.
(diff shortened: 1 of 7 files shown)


goral09 (approved):
I assume you tested it :+1:  :)

Fraser999:
> I assume you tested it +1 :)

Yes - I was doing that while CI ran.  Tested both with and without the env var set.

Fraser999:
bors merge

Fraser999:
bors merge

PR #827, opened 2021-01-28
NDRS-803: Add an integration test for a new node joining the network.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-803

Adds an integration test that exercises a scenario where a new node is joining an already running network.

Test scenario:
1. Send a batch of Wasm transfers.
2. Send a batch of native transfers.
3. Take a note of the LFB hash.
4. Send a batch of Wasm transfers.
5. Send a batch of native transfers.
6. Start a new node and join the network using LFB hash from **3)**
7. Verify that eventually, the LFB hash on the new node matches the hash on one of the old nodes.

This PR also adds new nctl command `nctl-view-chain-lfb` which will query nodes for the hash of the last finalized block.

Note that this requires `nctl` set up (as described in the `nctl` directory) and that there's an already running network.

Example usage:
```shell
nctl-assets-setup # generates chainspec, accounts.csv etc.
nctl-start # by default starts 5 nodes
source sync_test.sh node=6 timeout=500
```


momipsl (approved):
Nice work.  Good to see the tool being useful.

PR #828, opened 2021-01-28
NDRS-823: Factor out next era validators from EraSupervisor::handle_create_new_era

xcthulhu:


node/src/components/consensus/era_supervisor.rs
```diff
@@ -582,21 +572,31 @@ where
...
+        ) {
+            (Some(era_end), Some(next_era_validator_weights)) => {
+                (era_end, next_era_validator_weights)
+            }
+            _ => {
+                return fatal!(
+                    self.effect_builder,
+                    "attempted to create a new era with a non-switch block header: {}",
+                    block_header
+                )
```
xcthulhu, line 588: @marc-casperlabs Here we took the opportunity to do the right thing

xcthulhu:
This has been collaborative work with @afck

goral09 (approved):

PR #829, opened 2021-01-28
NO TICKET: joiner-dev-tmux.sh now shares assets with run-dev-tmux.sh

xcthulhu:


xcthulhu:
A recent change to the `config.toml` format that @afck made wasn't propagated to my joiner tests, so I've changed the joiner tests to just have symbolic links to the other local test config used in the test scripts.

afck (commented):
Why not delete the whole `resources/joiner` directory then, and just use `resources/local`?

xcthulhu:
> Why not delete the whole resources/joiner directory then, and just use resources/local?

I need a custom [accounts.csv][1] with just one bonded validator.

[1]: https://github.com/CasperLabs/casper-node/blob/630a446002ea574bf011168c944cca94de0bf2d2/resources/joiner/accounts.csv

afck (approved):

xcthulhu:
bors +r

xcthulhu:
bors merge

PR #830, opened 2021-01-29
EE-1140: remove grpc layer

EdHastingsCasperLabs:
This PR removes the legacy grpc layer.

https://casperlabs.atlassian.net/browse/EE-1140

Basically mechanical, except for the concurrent_executor test deprecation.


Makefile
```diff
@@ -153,7 +152,6 @@ check: \
 	build \
 	check-format \
 	lint \
-	audit \
```
Fraser999, line 156: I missed this when I committed the workaround - I only changed the drone file.

If you want to leave audit in with just that single warning ignored, you can change the audit target to do
```
$(CARGO) audit --ignore RUSTSEC-2020-0123
```

Fraser999 (approved):
Great to see this one land!

One minor non-blocker, but otherwise looks grand.

Makefile
EdHastingsCasperLabs: k

mpapierski (approved):

henrytill (approved):

PR #832, opened 2021-01-29
NDRS-767: Remove finality signatures from the block. 

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-767

ChangeLog:

- Removed `proofs` member, and associated `append_proof` function from DTO Block.
- Created a new `BlockSignatures` object to hold proofs for a block tagged by its hash
- Replaced `BlockCache` in `linear_chain` with `SignatureCache` 
- Made block storage a more immutable process, once blocks are stored into storage and they cannot be modified further.
- Added `get_block_with_metadata` function to retrieve both block and its associated metadata(signatures)
- Modified RPC process to query both block and signatures to send JSON conforming block representations to external consumers via RPC calls.


node/src/types/block.rs
```diff
@@ -1250,7 +1248,7 @@ pub(crate) mod json_compatibility {
                 hash: block.hash,
                 header: JsonBlockHeader::from(block.header),
                 body: block.body,
-                proofs: block.proofs.into_iter().map(JsonProof::from).collect(),
+                proofs: vec![],
```
EdHastingsCasperLabs, line 1251: Why are you returning an empty vec?

darthsiroftardis: Well, the JSON block contains proofs but not the actual block, so I generate a JSON block with empty proofs, and then append the signatures to the JSON block separately

node/src/components/linear_chain.rs
```diff
@@ -67,13 +67,13 @@ pub enum Event<I> {
         /// The deploys' execution results.
         execution_results: HashMap<DeployHash, ExecutionResult>,
     },
-    /// The result of requesting a block from storage to add a finality signature to it.
-    GetBlockForFinalitySignaturesResult(Box<FinalitySignature>, Option<Box<Block>>),
+    /// The result of requesting finality signatures from storage to add pending signatures.
+    GetStoredFinalitySignaturesResult(Box<FinalitySignature>, Option<Box<BlockSignatures>>),
```
xcthulhu, line 71: I am not sure we want this to be a `Box<FinalitySignature>`.  It might be more appropriate to have:

```rust
GetStoredFinalitySignaturesResult {
  block_hash: BlockHash, 
  era_id: u64,
  block_signatures: Box<BlockSignatures> 
}
```

that might be a lift, however.

node/src/effect.rs
```diff
@@ -842,6 +876,54 @@ impl<REv> EffectBuilder<REv> {
         .await
     }
 
+    /// Gets the requested block and its associated metadata.
+    pub(crate) async fn get_block_at_height_with_metadata(
```
xcthulhu, line 880:
```suggestion
    pub(crate) async fn get_block_at_height_with_metadata_from_storage(
```

node/src/effect.rs
```diff
@@ -842,6 +876,54 @@ impl<REv> EffectBuilder<REv> {
...
+                block_height,
+                responder,
+            },
+            QueueKind::Regular,
+        )
+        .await
+    }
+
+    /// Gets the requested block by hash with its associated metadata.
+    pub(crate) async fn get_block_with_metadata(
```
xcthulhu, line 898:
```suggestion
    pub(crate) async fn get_block_with_metadata_from_storage(
```

node/src/effect.rs
```diff
@@ -842,6 +876,54 @@ impl<REv> EffectBuilder<REv> {
...
+                block_hash,
+                responder,
+            },
+            QueueKind::Regular,
+        )
+        .await
+    }
+
+    /// Get the highest block with its associated metadata.
+    pub(crate) async fn get_highest_block_with_metadata(self) -> Option<(Block, BlockSignatures)>
```
xcthulhu, line 916:
```suggestion
    pub(crate) async fn get_highest_block_with_metadata_from_storage(self) -> Option<(Block, BlockSignatures)>
```

xcthulhu:
@darthsiroftardis I know it's pretty verbose, but we should have functions be named like `get_block_at_height_with_metadata_from_storage` and so on.

Soon we are going to need an effect/fetcher `get_block_at_height_with_metadata` that queries the gossip network. This is for joining securely using finality signatures. But almost always you want to use storage for security purposes.

EdHastingsCasperLabs (approved):
Overall, good job. Please apply name changes as requested by @xcthulhu before merging.

node/src/components/linear_chain.rs
```diff
@@ -534,19 +532,34 @@ where
...
                 }
             }
-            Event::IsBonded(None, _, true) => {
+            Event::IsBonded(None, fs, true) => {
                 // Unknown block but validator is bonded.
                 // We should finalize the same block eventually. Either in this or in the
                 // next era.
-                Effects::new()
+                // Store the signatures.
+                let mut effects = Effects::new();
```
Fraser999, line 553: We don't need a `mut effects` here, we can just construct them immutably below.

node/src/components/storage.rs
```diff
@@ -521,6 +526,68 @@ impl Storage {
...
+                let mut txn = self.env.begin_ro_txn()?;
+
+                let block: Block = if let Some(block) =
+                    self.get_single_block(&mut self.env.begin_ro_txn()?, &block_hash)?
+                {
+                    block
+                } else {
+                    return Ok(responder.respond(None).ignore());
+                };
+                let hash = block.hash();
```
Fraser999, line 542: Can't we just use `block_hash` rather than this?

node/src/effect.rs
```diff
@@ -686,6 +687,39 @@ impl<REv> EffectBuilder<REv> {
         .await
     }
 
+    /// Gets the requested signatures for a give block hash.
```
Fraser999, line 690:
```suggestion
    /// Gets the requested signatures for a given block hash.
```

node/src/effect.rs
```diff
@@ -686,6 +687,39 @@ impl<REv> EffectBuilder<REv> {
...
+            |responder| StorageRequest::GetBlockSignatures {
+                block_hash,
+                responder,
+            },
+            QueueKind::Regular,
+        )
+        .await
+    }
+
+    /// Store the requested finality signatures into storage.
```
Fraser999, line 708:
```suggestion
    /// Puts the requested finality signatures into storage.
```

node/src/types/block.rs
Fraser999, line 1251: I'd be in favour of removing this `From` impl since we can't _really_ construct a `JsonBlock` properly from just a `Block` any more.  The `JsonBlock` represents a block and its signatures, so I feel it's a little dangerous to have a public constructor of `JsonBlock` which doesn't properly construct it.

I think we should either rely on your `new()` constructor, or we could do a `From<(Block, BlockSignatures)>` if you really want.

node/src/types/block.rs
```diff
@@ -1233,6 +1220,17 @@ pub(crate) mod json_compatibility {
     }
 
     impl JsonBlock {
+        /// Create a new JSON Block with a Linear chain block and its associated signatures.
+        pub fn new(block: Block, signatures: BlockSignatures) -> Self {
+            let mut proto_block = JsonBlock::from(block);
```
Fraser999, line 1225: I think the variable would be better named `json_block`, since `proto_block` implies it's a `ProtoBlock` as defined elsewhere in this module.

node/src/types/block.rs
```diff
@@ -1233,6 +1220,17 @@ pub(crate) mod json_compatibility {
...
+            proto_block.proofs = signatures
+                .proofs
+                .iter()
```
Fraser999, line 1228: If we use `into_iter` there would be no need to copy the values in `JsonProof::new()`. 

In fact we could simplify this to just `signatures.proofs.into_iter().map(JsonProof::from).collect();`.

node/src/types/block.rs
```diff
@@ -1274,6 +1271,15 @@ pub(crate) mod json_compatibility {
...
 
+    impl JsonProof {
+        pub fn new(public_key: PublicKey, signature: Signature) -> Self {
+            JsonProof {
+                public_key,
+                signature,
+            }
+        }
+    }
+
```
Fraser999, line 1282: This could be removed.

node/src/types/block.rs
```diff
@@ -783,14 +778,54 @@ impl From<bytesrepr::Error> for BlockValidationError {
...
 
+/// A storage representation of finality signatures with the associated block hash.
+#[derive(Debug, Serialize, Deserialize, Clone, DataSize)]
+pub struct BlockSignatures {
+    /// The block hash for a given block.
+    pub block_hash: BlockHash,
+    /// The era id for the given set of finality signatures
+    pub era_id: EraId,
+    /// The signatures associated with the block hash.
+    pub proofs: BTreeMap<PublicKey, Signature>,
```
Fraser999, line 789: These fields should ideally not be `pub`.  Instead there should be a `pub(crate)` getter for each so that we can be sure the only way a `BlockSignatures` struct can be modified by a caller is via its one mutable method `append_proof()`.

node/src/types/block.rs
```diff
@@ -783,14 +778,54 @@ impl From<bytesrepr::Error> for BlockValidationError {
     }
 }
 
+/// A storage representation of finality signatures with the associated block hash.
+#[derive(Debug, Serialize, Deserialize, Clone, DataSize)]
+pub struct BlockSignatures {
+    /// The block hash for a given block.
+    pub block_hash: BlockHash,
+    /// The era id for the given set of finality signatures
```
Fraser999, line 786:
```suggestion
    /// The era id for the given set of finality signatures.
```

node/src/types/block.rs
```diff
@@ -783,14 +778,54 @@ impl From<bytesrepr::Error> for BlockValidationError {
...
+    ) -> Option<Signature> {
+        self.proofs.insert(public_key, signature)
+    }
+}
+
+impl Display for BlockSignatures {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(
+            formatter,
+            "Block signature for Hash: {} in era_id: {} with proofs of lenght: {}",
```
Fraser999, line 814:
```suggestion
            "block signatures for hash: {} in era_id: {} with {} proofs",
```

Fraser999 (changes requested):

goral09:
This change would have been a bit easier to review if you did as I suggested – split this work into two parts: first, add the new functionality to the storage, and second replace currently used proofs with the new scheme.

node/src/components/linear_chain.rs
```diff
@@ -438,52 +439,49 @@ where
...
+                if let Some(signatures) = &maybe_signatures {
+                    if signatures.era_id != fs.era_id {
                         warn!(public_key=%fs.public_key, "Finality signature with invalid era id.");
                         // TODO: Disconnect from the sender.
                         return Effects::new();
                     }
-                    // Populate cache so that next finality signatures don't have to read from the
-                    // storage. If block is already from cache then this will be a noop.
-                    self.block_cache.insert(*block.clone())
+                    // Populate cache
```
goral09, line 461: Please bring back the full comment and update if necessary.

node/src/components/linear_chain.rs
```diff
@@ -534,19 +532,34 @@ where
...
+                // Store the signatures.
+                let mut effects = Effects::new();
+                let mut signatures = BlockSignatures::new(fs.block_hash, fs.era_id);
+                signatures.append_proof(fs.public_key, fs.signature);
+                effects.extend(
+                    effect_builder
+                        .put_signatures_to_storage(signatures.clone())
+                        .ignore(),
+                );
+                self.signature_cache.insert(signatures);
```
goral09, line 561: This is changing the logic and is not correct. `signature_cache` should only contain signatures that we have validated and stored. You should remove the whole bit that you added to this match arm.

goral09 (changes requested):
There are a couple of blockers so please address all of my comments.

node/src/components/linear_chain.rs
```diff
@@ -534,19 +532,34 @@ where
...
                             .ignore(),
                     );
-                    block.append_proof(fs.public_key, fs.signature);
+                    signature.append_proof(fs.public_key, fs.signature);
                     // Cache the results in case we receive the same finality signature before we
                     // manage to store it in the database.
-                    self.block_cache.insert(*block.clone());
-                    effects.extend(effect_builder.put_block_to_storage(block).ignore());
+                    self.signature_cache.insert(*signature.clone());
+                    info!(hash=%signature.block_hash, "storing finality signatures");
```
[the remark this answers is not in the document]
darthsiroftardis: Yeah, I think the inconsistency is due to my attempts to debug some unintended behavior that had arisen due to mistakes

node/src/components/linear_chain.rs
```diff
@@ -242,32 +242,31 @@ impl<I> LinearChain<I> {
         I: Display + Send + 'static,
     {
         let mut effects = Effects::new();
-        let block_hash = block.hash();
+        let mut known_signatures = self.signature_cache.get_signatures(block_hash);
```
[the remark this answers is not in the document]
darthsiroftardis: I tried to follow the previous steps as closely as possible, I thought it made sense to query the cache whilst collecting the pending signatures and then update them for a particular block hash as I thought we were doing previously, and return the updated DTO BlockSignatures object back to be processed the same way we returned an updated block

node/src/components/rpc_server/event.rs
```diff
@@ -23,8 +23,8 @@ pub enum Event {
     RpcRequest(RpcRequest<NodeId>),
     GetBlockResult {
         maybe_id: Option<BlockIdentifier>,
-        result: Box<Option<Block>>,
-        main_responder: Responder<Option<Block>>,
```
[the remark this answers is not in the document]
darthsiroftardis: No it can remain as `GetBlockResult` as from the perspective of the RPC event, it really cares about retrieving a block to send to external callers which includes the signatures as metadata. Its also the convention we have with `Deploy` even though there is associated metadata there too

node/src/components/rpc_server/rpcs/state.rs
```diff
@@ -341,7 +341,7 @@ impl RpcWithoutParamsExt for GetAuctionInfo {
                             error_msg,
                         ))?);
                     }
-                    Some(block) => block,
+                    Some((block, _)) => block,
```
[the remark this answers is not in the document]
darthsiroftardis: Personally, I think the effect of querying a block should include the signatures always, but these requests are for information contained within the block and depending on the RPC request, extraneous information can be discarded as needed

node/src/components/linear_chain.rs
```diff
@@ -534,19 +532,34 @@ where
...
-                Effects::new()
+                // Store the signatures.
+                let mut effects = Effects::new();
+                let mut signatures = BlockSignatures::new(fs.block_hash, fs.era_id);
+                signatures.append_proof(fs.public_key, fs.signature);
+                effects.extend(
+                    effect_builder
+                        .put_signatures_to_storage(signatures.clone())
+                        .ignore(),
+                );
```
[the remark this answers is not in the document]
darthsiroftardis: Right, that was a mistake I made, I am removing it

node/src/components/linear_chain.rs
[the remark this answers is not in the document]
goral09: Signatures from `signature_cache` have already been gossiped, sent to the event stream, and stored (or at least they should have been). If we include them here again,  they will go through that same process again. I am pretty sure it didn't work that way before (i.e. before this PR).

node/src/effect.rs
darthsiroftardis: Applied suggestion

node/src/effect.rs
darthsiroftardis: Applied suggestion

node/src/components/linear_chain.rs
```diff
@@ -135,54 +135,53 @@ impl<I: Display> Display for Event<I> {
...
+            for (pub_key, sig) in &block_signature.proofs {
+                entry.insert(*pub_key, *sig);
             }
         }
-        self.blocks.insert(*block.hash(), block.take_header());
+        self.signatures
+            .insert(block_signature.block_hash, block_signature);
+    }
+
+    fn get_signatures(&self, block_hash: &BlockHash) -> BlockSignatures {
```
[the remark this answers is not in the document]
darthsiroftardis: Renamed

node/src/components/linear_chain.rs
```diff
@@ -438,52 +439,49 @@ where
...
                         return Effects::new();
                     }
-                    // Populate cache so that next finality signatures don't have to read from the
-                    // storage. If block is already from cache then this will be a noop.
-                    self.block_cache.insert(*block.clone())
+                    // Populate cache
+                    self.signature_cache.insert(*signatures.clone());
                 }
-                // Check if validator is bonded in the era in which the block was created.
+                // Check if the validator is bonded.
```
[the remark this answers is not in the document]
darthsiroftardis: Modified comment as requested

node/src/components/linear_chain.rs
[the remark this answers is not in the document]
darthsiroftardis: Changed log level

node/src/types/block.rs
darthsiroftardis: Used suggestiong

node/src/types/block.rs
darthsiroftardis: Removed

node/src/types/block.rs
darthsiroftardis: Changed to `pub(crate)`

node/src/types/block.rs
darthsiroftardis: Applied grammar

node/src/types/block.rs
darthsiroftardis: Applied suggestion

node/src/components/linear_chain.rs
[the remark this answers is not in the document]
darthsiroftardis: Removed

node/src/types/block.rs
```diff
@@ -783,14 +778,54 @@ impl From<bytesrepr::Error> for BlockValidationError {
...
+impl BlockSignatures {
+    pub(crate) fn new(block_hash: BlockHash, era_id: EraId) -> Self {
+        BlockSignatures {
+            block_hash,
+            era_id,
+            proofs: BTreeMap::new(),
+        }
+    }
+
+    pub(crate) fn append_proof(
```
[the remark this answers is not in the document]
darthsiroftardis: Changed to `insert_proof`

node/src/components/linear_chain.rs
darthsiroftardis: Brought back full comment

node/src/components/linear_chain.rs
darthsiroftardis: Removed incorrect logic

node/src/components/linear_chain.rs
```diff
@@ -135,54 +135,53 @@ impl<I: Display> Display for Event<I> {
...
-    /// Inserts new block to the cache.
-    /// Caches block's finality signatures as well.
-    /// If we receive a block from a new era, old era blocks are removed.
-    fn insert(&mut self, block: Block) {
-        let new_block_era = block.header().era_id();
-        if self.curr_era < new_block_era {
-            // We optimistically assume that most of the signatures that arrive in close temporal
-            // proximity refer to the same era.
-            self.blocks.clear();
+    fn insert(&mut self, block_signature: BlockSignatures) {
```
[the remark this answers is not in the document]
darthsiroftardis: Brought back comment

node/src/components/linear_chain.rs
```diff
@@ -135,54 +135,53 @@ impl<I: Display> Display for Event<I> {
...
+        if self.curr_era < block_signature.era_id {
+            self.signatures.clear();
```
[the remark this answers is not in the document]
darthsiroftardis: Brought back comment

node/src/components/linear_chain.rs
```diff
@@ -242,32 +246,31 @@ impl<I> LinearChain<I> {
         I: Display + Send + 'static,
     {
         let mut effects = Effects::new();
-        let block_hash = block.hash();
+        let mut known_signatures = self.signature_cache.get_known_signatures(block_hash);
```
goral09, line 249: This is still wrong. We must not collect any signature we already know of but only those that are still pending.

node/src/components/linear_chain.rs
```diff
@@ -438,52 +439,49 @@ where
...
-                            // future era.
-                            Err((maybe_block, fs))
+                            Err((maybe_signatures, fs))
                         }
                     })
             }
             .result(
-                |(maybe_block, fs, is_bonded)| Event::IsBonded(maybe_block, fs, is_bonded),
-                |(maybe_block, fs)| Event::IsBondedFutureEra(maybe_block, fs),
+                |(maybe_signature, fs, is_bonded)| Event::IsBonded(maybe_signature, fs, is_bonded),
```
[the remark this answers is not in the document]
darthsiroftardis: Renamed to signatures

node/src/effect.rs
darthsiroftardis: Applied suggestion

node/src/components/linear_chain.rs
```diff
@@ -135,54 +135,53 @@ impl<I: Display> Display for Event<I> {
...
 }
 
-// Cache for blocks.
-// Blocks are invalidated when a block for newer era is inserted.
 #[derive(DataSize, Debug)]
-struct BlockCache {
+struct SignatureCache {
     curr_era: EraId,
-    blocks: HashMap<BlockHash, BlockHeader>,
+    signatures: HashMap<BlockHash, BlockSignatures>,
```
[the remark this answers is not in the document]
goral09: You were right. It was not needed in the end.

goral09 (approved):
Looks good to me. Thanks for working on that. Sorry for not providing unit tests.

node/src/types/block.rs
Fraser999: I meant that we should have these member variables private, and have `pub(crate)` methods like `pub(crate) fn block_hash(&self) -> &BlockHash`

node/src/components/linear_chain.rs
```diff
@@ -515,16 +534,17 @@ where
                     }
                 }
             }
-            Event::IsBonded(Some(mut block), fs, true) => {
+            Event::IsBonded(Some(mut signature), fs, true) => {
```
Fraser999, line 537:
```suggestion
            Event::IsBonded(Some(mut signatures), fs, true) => {
```

node/src/components/linear_chain.rs
```diff
@@ -515,16 +534,17 @@ where
...
                 // Known block and signature from a bonded validator.
                 // Check if we had already seen this signature before.
-                let signature_known = block
-                    .proofs()
+                let signature_known = signature
```
Fraser999, line 540:
```suggestion
                let signatures_known = signatures
```

Fraser999 (commented):
A couple of missed comments, but this is looking good now.

Fraser999 (approved):
Unblocking this PR, in the expectation that the outstanding comments are addressed.

goral09:
bors merge

PR #833, opened 2021-01-29
EE-1163: Refactor `engine_state::mod.rs` file. [NFC]

goral09:
This PR deduplicates some of the code between `deploy` and `transfer` methods by extracting the code to macros. It couldn't be a function b/c we need to return early in error cases so we need to "expand" the `return` logic on the call side.


henrytill (changes requested):
I don't want to introduce macros here.

PR #834, opened 2021-01-29
EE-1152: DelegatorNotFound while updating delegator rewards

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1152

Node while running a step function does distribute rewards using trusted validators passed from node, but between step calls, a delegator can undelegate funds from one of those trusted validators. While updating delegator rewards in a step call, a delegator can be already missing from current set of bids, but node does not know that through its trusted validators.

Original title of the ticked assumed MismatchedEraValidators error, but that was not observed during nctl testing, and arguments in the step function are assumed to match the global state information.

# Testing done

```
nctl-assets-setup && nctl-start
sleep 30
nctl-auction-delegate delegator=1
nctl-auction-delegate delegator=2
nctl-auction-delegate delegator=3
nctl-auction-delegate delegator=4
nctl-auction-delegate delegator=5
nctl-await-n-eras offset=1
nctl-auction-undelegate delegator=1
nctl-auction-undelegate delegator=2
nctl-auction-undelegate delegator=3
nctl-auction-undelegate delegator=4
nctl-auction-undelegate delegator=5
sleep 60
nctl-status 
```

^ Without fix network dies before nctl-status is executed. With the fix the network is still alive past the test scenario


EdHastingsCasperLabs (approved):

goral09:
Why not simply disallowing undelegating mid-era? Wouldn't that make it simpler? Eras are supposed to be really short after all – 30 minutes or so. Validator cannot unbond mid-era but delegaters can undelegate?

MParlikar:
I think it's fine to not 'commit' an undelegate until the step function runs at the end of an era.  This way there is no confusion about partial rewards, etc...Also we do not have to worry about validator weights fluctuating mid-era - in the event of a large number of undelegate calls for a given validator.

EdHastingsCasperLabs:
I suggest we allow this severe bug fix PR merge, and we can discuss altering the auction behavior as a separate piece of work.

goral09:
> I suggest we allow this severe bug fix PR merge, and we can discuss altering the auction behavior as a separate piece of work.

Sure, I didn't mean to block the fix. I was simply wondering why not go with something simpler.

PR #835, opened 2021-01-29
NO_TICKET ensure transfers are included in block limited only by bloc…

dwerner:
This PR is intended to ensure that transfers are included in block limited only by `block_max_transfer_limit`. @EdHastingsCasperLabs asked for this change in our testing meeting today.

### DIFF
--- a/node/src/components/block_proposer/event.rs
+++ b/node/src/components/block_proposer/event.rs
@@ -72,7 +72,7 @@ impl DeployType {
     }
 
     /// Asks if the variant is Wasm.
-    pub fn is_wasm(&self) -> bool {
+    pub fn is_other(&self) -> bool {
         matches!(self, DeployType::Other { .. })
     }
 }
--- a/node/src/components/block_proposer/event.rs
+++ b/node/src/components/block_proposer/event.rs
@@ -72,7 +72,7 @@ impl DeployType {
     }
 
     /// Asks if the variant is Wasm.
-    pub fn is_other(&self) -> bool {
+    pub fn is_wasm(&self) -> bool {
         matches!(self, DeployType::Other { .. })
     }
 }
--- a/node/src/components/block_proposer/tests.rs
+++ b/node/src/components/block_proposer/tests.rs
@@ -487,7 +487,6 @@ fn should_not_propose_deploy_if_block_size_limit_passed() {
         proposed_count: 4,
         remaining_pending_count: 1,
         max_block_size: Some(2 * DEPLOY_APPROX_MIN_SIZE),
-        ..Default::default()
     });
 }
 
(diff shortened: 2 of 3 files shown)


goral09:
What's the size native transfer deploy?

node/src/components/block_proposer/tests.rs
```diff
@@ -585,6 +567,8 @@ fn test_proposer_with(
     let block = proposer.propose_proto_block(config, test_time, past_deploys, true);
     let all_deploys = BlockLike::deploys(&block);
     proposer.finalized_deploys(all_deploys.iter().map(|hash| **hash));
+    println!("proposed deploys {}", block.wasm_deploys().len());
+    println!("proposed transfers {}", block.transfers().len());
```
goral09, line 571: Leftover?

node/src/components/block_proposer/tests.rs
```diff
@@ -483,24 +469,24 @@ fn should_not_propose_deploy_if_block_size_limit_within_threshold() {
...
     });
 }
 
 #[test]
 fn should_not_propose_deploy_if_block_size_limit_passed() {
     test_proposer_with(TestArgs {
-        deploy_count: 0,
-        transfer_count: 1,
+        deploy_count: 3,
+        transfer_count: 2, // transfers should -not- count towards the block size limit
```
goral09, line 482: Can you add a test that sets a block size limit lower than `max_transfer_count` occupies? i.e. show that transfers can exceed block size limit.

node/src/components/block_proposer/tests.rs
dwerner: I'd like to keep these, as they are nice to have as output in the tests if they fail.

dwerner:
> What's the size of native transfer deploy?

There is a constant in `block_proposer` I added:
```
/// Experimentally, deploys are in the range of 270-280 bytes, we use this to determine if we are
/// within a threshold to break iteration of `pending` early.
const DEPLOY_APPROX_MIN_SIZE: usize = 300;
```

node/src/components/block_proposer/event.rs
```diff
@@ -72,7 +72,7 @@ impl DeployType {
     }
 
     /// Asks if the variant is Wasm.
-    pub fn is_wasm(&self) -> bool {
+    pub fn is_other(&self) -> bool {
```
goral09, line 75: `is_wasm` was a better name IMO. I understand that the Wasm variants is called `Other` right now but in our domain it's either Wasm or native transfer and I would stick to these names for now. When we get to add other variants w can change names, add other methods.

dwerner: Fair enough, I'll change it back. :+1:

node/src/components/block_proposer/tests.rs
EdHastingsCasperLabs: Then change the logic to only print them when it fails, please.

EdHastingsCasperLabs (approved):
nitpick should not block merge as we need this urgently

PR #836, opened 2021-01-30
NDRS-447 - implement rate limiting on http endpoints using tower

dwerner:
This PR implements rate limiting on all our http endpoints using tower.


goral09 (approved):

resources/local/config.toml
```diff
@@ -89,6 +89,9 @@ systemd_support = false
 # The actual bound address will be reported via a log line if logging is enabled.
 address = '0.0.0.0:7777'
 
+# The max rate of requests (per second) before they are limited.
+qps_limit = 1000
```
EdHastingsCasperLabs, line 93: this seems like a very high limit per second to me

resources/local/config.toml
```diff
@@ -121,6 +127,9 @@ event_stream_buffer_length = 100
 # The capacity of the broadcast channel size.
 broadcast_channel_size = 100
 
+# The max rate of requests (per second) before they are limited.
+qps_limit = 1000
```
EdHastingsCasperLabs, line 131: is this even relevant to sse?

client/tests/integration_test.rs
```diff
@@ -747,6 +774,48 @@ mod put_deploy {
     }
 }
 
+mod rate_limit {
```
Fraser999, line 777: I'm not sure this test really makes sense here.  This is proving that we can rate-limit our mock server, not our production node, and it doesn't test the client in any way (unless I'm misunderstanding it?)

node/src/components/rpc_server/http_server.rs
```diff
@@ -11,11 +12,13 @@ use super::{
     ReactorEventT,
 };
 use crate::effect::EffectBuilder;
+use std::time::Duration;
```
Fraser999, line 15: This needs moved up.

node/src/types/timestamp.rs
```diff
@@ -300,7 +300,6 @@ impl FromBytes for TimeDiff {
     }
 }
 
-#[cfg(test)]
```
Fraser999, line 303: Does this need to be removed?

resources/local/config.toml
```diff
@@ -89,6 +89,9 @@ systemd_support = false
 # The actual bound address will be reported via a log line if logging is enabled.
 address = '0.0.0.0:7777'
 
+# The max rate of requests (per second) before they are limited.
```
Fraser999, line 92: I think the comment should also state if this is a per-client limit or not, and also what happens when the limit is hit.  (I'm guessing the client sending connecting is temporarily blocked rather than being dropped?)

We need to duplicate that info in all 6 places I guess.

Fraser999 (commented):
This looks like a good improvement.  I'm keen to know whether the rate limiting is per client or not, and what effect rate-limiting has on a client.

A few minor changes requested, but basically looking good to me.

resources/local/config.toml
dwerner: Any http endpoint could be overwhelmed by too many requests. My thinking at the time was to apply the rate limiting across all endpoints, even if they are used for long lived connections like in the case of SSE.

dwerner:
> This looks like a good improvement. I'm keen to know whether the rate limiting is per client or not, and what effect rate-limiting has on a client.
> 
> A few minor changes requested, but basically looking good to me.

In my testing (in client integration_test.rs) the effect is stalling out the response (as in keeping tcp connection alive, but waiting to pass on the event to the underlying warp filter), rather than a 429. The basic rate limiting provided by the `https://docs.rs/tower/0.3.1/tower/limit/rate/struct.RateLimit.html` implementation is very generic, and not http or client-specific.

client/tests/integration_test.rs
dwerner: I thought it was worthwhile to add here because while it's not a real test of our production code, it does actually instantiate a client and make http requests against it.

In doing that I found a bug in the client where if it's executed within a tokio runtime it can overwhelm the executor and block all task processing threads.

node/src/types/timestamp.rs
dwerner: The `From<Duration> for TimeDiff` impl isn't exposed to the test, and I think this is a worthwhile implementation to have. Is there a reason why it should stay as `#[cfg(test)]`?

client/tests/integration_test.rs
```diff
@@ -747,6 +774,48 @@ mod put_deploy {
...
+
+    #[tokio::test(threaded_scheduler)]
+    async fn client_should_should_be_rate_limited_to_approx_1_qps() {
+        // Transfer uses PutDeployParams + PutDeploy
+        let server_handle = Arc::new(MockServerHandle::spawn::<PutDeployParams>(
+            PutDeploy::METHOD,
+        ));
+
+        let now = Timestamp::now();
+        // Our default is 100 req/min, so this will time out
```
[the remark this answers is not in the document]
dwerner: Fixed this.

client/tests/integration_test.rs
```diff
@@ -747,6 +774,48 @@ mod put_deploy {
...
+        for i in 0..10u32 {
+            println!("request {}", i);
+            let amount = "100";
+            let maybe_target_account =
+                "01522ef6c89038019cb7af05c340623804392dd2bb1f4dab5e4a9c3ab752fc0179";
+
+            // block_in_place is needed to prevent deadlock on the futures::executor::block_on
+            // call inside transfer
```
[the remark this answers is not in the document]
dwerner: If you remove the `tokio::task::spawn_blocking` call wrapping the call to `transfer`, the client will eventually eat all executor threads and deadlock (at 64 consecutive requests).

sacherjj:
Run a build from this PR against release-test-16 and works well for RPC limiting.  Running looping add_bid deploy and looping native transfer in two terminals with rpc 5/s limit.  Seeing 4 transfers get in and 1 deploy per second and working without issue.

client/tests/integration_test.rs
```diff
@@ -8,10 +8,8 @@ use tower::builder::ServiceBuilder;
 use warp::{Filter, Rejection};
 use warp_json_rpc::Builder;
 
-use casper_node::crypto::Error::*;
-use hex::FromHexError::*;
-use std::time::Duration;
-use Error::*;
+use casper_node::crypto::Error as CryptoError;
+use hex::FromHexError;
```
Fraser999, line 12: This should be with the other 3rd party imports.

client/tests/integration_test.rs
```diff
@@ -15,12 +21,17 @@ use casper_node::rpcs::{
     state::{GetBalance, GetBalanceParams},
     RpcWithOptionalParams, RpcWithParams,
 };
+use std::sync::Arc;
+use tokio::sync::Mutex;
```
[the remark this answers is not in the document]
Fraser999: These haven't been moved.

client/tests/integration_test.rs
goral09: OK, I have to ask about it – why are we introducing our own import format rules that we have to apply manually? What's the point really? When it comes to formatting, the rules should be that it's uniform and automatic. I really don't are how they're sorted, I never look at them.

client/tests/integration_test.rs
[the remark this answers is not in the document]
Fraser999: OK - thanks - makes sense.  It might be worth expanding the comment and referring to `spawn_blocking` rather than `block_in_place`?

client/tests/integration_test.rs
```diff
@@ -747,6 +774,48 @@ mod put_deploy {
...
+        for i in 0..10u32 {
+            println!("request {}", i);
+            let amount = "100";
+            let maybe_target_account =
+                "01522ef6c89038019cb7af05c340623804392dd2bb1f4dab5e4a9c3ab752fc0179";
+
+            // block_in_place is needed to prevent deadlock on the futures::executor::block_on
+            // call inside transfer
+            let server_handle = server_handle.clone();
+            let join_handle = tokio::task::spawn_blocking(move || {
```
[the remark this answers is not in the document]
Fraser999: This has been missed too.

client/tests/integration_test.rs
```diff
@@ -786,9 +786,8 @@ mod rate_limit {
         ));
 
         let now = Timestamp::now();
-        // Our default is 100 req/min, so this will time out
-        for i in 0..10u32 {
-            println!("request {}", i);
+        // Our default is 10 req/s, so this will hit the threshold
```
Fraser999, line 789: Isn't the mock server defaulting to 1 QPS?

client/tests/integration_test.rs
Fraser999: It's been a rule we've applied (reasonably strictly) since before I started here.  It makes sense when you have large groups of imports IMO.

client/tests/integration_test.rs
Fraser999: Hmm - well, it does instantiate a client, but it's not _testing_ the client (at least not any more than the existing tests do).

And I wouldn't agree that that's a bug - it's just a misuse of the client inside the test really?  There wasn't a bugfix applied to the client as a result of this.

This feels more like it belongs as an example than a test.  However, I don't want to block this PR over this.

Could we at least reduce the test run time down from 10 seconds?  That seems a bit excessive to me.

node/src/types/timestamp.rs
Fraser999: I'm guessing it's to keep the API as small as possible for the production case, but if we need it then fair enough.  I don't see a problem exposing it - I just wanted to check we did need the change.

client/tests/integration_test.rs
goral09: I know it's been a rule but I have never understood it.

>  It makes sense when you have large groups of imports IMO.

Why does it make sense? Do you ever read the imports?

node/src/types/timestamp.rs
dwerner: Yeah, I'm using a Duration to construct a TimeDiff, and I think it's a worthwhile implementation to have for interop.

client/tests/integration_test.rs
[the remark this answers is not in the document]
dwerner: Sorry, thought I had.

client/tests/integration_test.rs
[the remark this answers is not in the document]
dwerner: I agree I completely missed this, but I wanted to say, I don't personally find much value in this kind of change, in fact it hides meaning. It makes `tokio::task` -> `task` and there are other libraries that implement the same meaning of `task`, like `async_std`. Same applies for `futures::executor` back then `tokio::executor` was still a thing. I'm happy with adjusting to your preference on this case, but I'd like to better understand your reasoning for it.

Fraser999 (approved):

dwerner:
Oops.

client/tests/integration_test.rs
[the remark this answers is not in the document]
Fraser999: It's purely due to that being the code style since I started here.  My own preference is almost always for consistency, so since we have this style elsewhere, I prefer to see it applied everywhere.

If you fancy changing the style guide, we should probably discuss outside of a PR (I think you'd likely get some support for your POV).

client/tests/integration_test.rs
Fraser999: I don't tend to read the full list of imports, but I sometimes look through them for a specific crate's.  I find that simpler if they're categorised like this.

I also don't overly care about this in terms of the rule itself - I'd be fine with all imports grouped together.  I do care about consistently applying the current style guide though :)

PR #842, opened 2021-02-01
HWY-239: Pause if too much stake is offline

afck:
This prevents the remaining validators from creating more and more units (increasing the protocol state size, and thus memory usage) while too many validators are offline to finalize anything. E.g. if there's a 34% fault tolerance threshold, we need more than 67% = 50% + 34% / 2 online for liveness, so if 33% or more are offline, the other validators will pause.

During a pause, instead of units, they gossip new `Ping` messages, so that others can keep track of who's still online. Once enough validators are back online again, they resume.

https://casperlabs.atlassian.net/browse/HWY-239

### DIFF
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -213,13 +213,17 @@ impl<C: Context> ActiveValidator<C> {
 
     /// Returns whether enough validators are online to finalize values with the target fault
     /// tolerance threshold, always counting this validator as online.
-    fn enough_validators_online(&self, state: &State<C>, timestamp: Timestamp) -> bool {
+    fn enough_validators_online(&self, state: &State<C>, now: Timestamp) -> bool {
         let target_quorum = (state.total_weight() + self.target_ftt) / 2;
-        let mut weight = state.online_and_honest_weight(timestamp);
-        if !state.is_online_and_honest(self.vidx, timestamp) {
-            weight += state.weight(self.vidx); // We know we are online.
-        }
-        weight > target_quorum
+        let online_weight: Weight = state
+            .weights()
+            .enumerate()
+            .filter(|(vidx, _)| {
+                self.vidx == *vidx || (!state.is_faulty(*vidx) && state.is_online(*vidx, now))
+            })
+            .map(|(_, w)| *w)
+            .sum();
+        online_weight > target_quorum
     }
 
     /// Returns actions a validator needs to take upon receiving a new unit.
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -10,8 +10,8 @@ use tracing::{error, info, trace, warn};
 use super::{
     endorsement::{Endorsement, SignedEndorsement},
     evidence::Evidence,
-    highway::{Endorsements, ValidVertex, Vertex, WireUnit},
-    state::{self, Panorama, State, Unit},
+    highway::{Endorsements, Ping, ValidVertex, Vertex, WireUnit},
+    state::{self, Panorama, State, Unit, Weight},
     validators::ValidatorIndex,
 };
 
@@ -73,6 +73,9 @@ pub(crate) struct ActiveValidator<C: Context> {
     unit_hash_file: Option<PathBuf>,
     /// The hash of the last known unit created by us.
     own_last_unit: Option<C::Hash>,
+    /// The target fault tolerance threshold. The validator pauses (i.e. doesn't create new units)
+    /// if not enough validators are online to finalize values at this FTT.
+    target_ftt: Weight,
 }
 
 impl<C: Context> Debug for ActiveValidator<C> {
@@ -93,6 +96,7 @@ impl<C: Context> ActiveValidator<C> {
         start_time: Timestamp,
         state: &State<C>,
         unit_hash_file: Option<PathBuf>,
+        target_ftt: Weight,
     ) -> (Self, Vec<Effect<C>>) {
         let own_last_unit = unit_hash_file
             .as_ref()
@@ -115,6 +119,7 @@ impl<C: Context> ActiveValidator<C> {
             next_proposal: None,
             unit_hash_file,
             own_last_unit,
+            target_ftt,
         };
         let effects = av.schedule_timer(start_time, state);
         (av, effects)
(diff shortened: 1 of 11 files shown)


node/src/components/consensus/highway_core/active_validator.rs
```diff
@@ -180,21 +185,43 @@ impl<C: Context> ActiveValidator<C> {
...
+        }
         effects
     }
 
+    /// Returns whether enough validators are online to finalize values with the target fault
+    /// tolerance threshold, always counting this validator as online.
+    fn enough_validators_online(&self, state: &State<C>, timestamp: Timestamp) -> bool {
+        let target_quorum = (state.total_weight() + self.target_ftt) / 2;
+        let mut weight = state.online_and_honest_weight(timestamp);
+        if !state.is_online_and_honest(self.vidx, timestamp) {
```
goral09, line 219: Do we really have to check if we're online ourselves?

node/src/components/consensus/highway_core/active_validator.rs
```diff
@@ -556,6 +583,9 @@ impl<C: Context> ActiveValidator<C> {
                 endorsements.endorsers.iter().any(is_ours)
                     && !state.has_endorsement(endorsements.unit(), self.vidx)
             }
+            Vertex::Ping(ping) => {
+                ping.creator() == self.vidx && !state.has_ping(self.vidx, ping.timestamp())
```
goral09, line 587: Why do we have to check if we have the ping message in the state? Is it being gossiped?

node/src/components/consensus/highway_core/highway.rs
```diff
@@ -304,6 +324,7 @@ impl<C: Context> Highway<C> {
             Dependency::Unit(hash) => self.state.has_unit(hash),
             Dependency::Evidence(idx) => self.state.is_faulty(*idx),
             Dependency::Endorsement(hash) => self.state.is_endorsed(hash),
+            Dependency::Ping(_, _) => false, // We don't store signatures; nothing depends on pings.
```
goral09, line 327: I think the comment should say that `Ping` does not depend on anything.

node/src/components/consensus/highway_core/highway/vertex.rs
```diff
@@ -59,32 +62,29 @@ impl<C: Context> Vertex<C> {
...
+        matches!(self, Vertex::Evidence(_))
     }
 
     /// Returns a `Timestamp` provided the vertex is a `Vertex::Unit`
     pub(crate) fn timestamp(&self) -> Option<Timestamp> {
         match self {
             Vertex::Unit(signed_wire_unit) => Some(signed_wire_unit.wire_unit().timestamp),
-            Vertex::Evidence(_) => None,
-            Vertex::Endorsements(_) => None,
+            Vertex::Ping(ping) => Some(ping.timestamp()),
```
goral09, line 86: The comment could be updated. Now it only mentions `Vertex::Unit`.

node/src/components/consensus/highway_core/highway/vertex.rs
```diff
@@ -277,3 +278,64 @@ impl<C: Context> Endorsements<C> {
...
+        &self,
+        validators: &Validators<C::ValidatorId>,
+    ) -> Result<(), VertexError> {
+        let v_id = validators.id(self.creator).ok_or(PingError::Creator)?;
+        let hash = Self::hash(self.creator, self.timestamp);
+        if !C::verify_signature(&hash, v_id, &self.signature) {
+            return Err(PingError::Signature.into());
+        }
+        Ok(())
+    }
```
goral09, line 334: I don't think validity of a model should be in the model itself.

node/src/components/consensus/highway_core/state.rs
```diff
@@ -479,6 +490,37 @@ impl<C: Context> State<C> {
...
+    pub(crate) fn has_ping(&self, creator: ValidatorIndex, timestamp: Timestamp) -> bool {
+        let max_round_len = round_len(self.params.max_round_exp());
+        self.pings[creator] + max_round_len > timestamp
+    }
+
+    /// Returns whether the validator is honest and the latest unit or ping is at most
+    /// `PING_TIMEOUT` maximum round lengths old.
+    pub(crate) fn is_online_and_honest(&self, vidx: ValidatorIndex, now: Timestamp) -> bool {
+        let max_round_len = round_len(self.params.max_round_exp());
+        !self.is_faulty(vidx) && self.pings[vidx] + max_round_len * PING_TIMEOUT >= now
```
goral09, line 512: Can you split it up in to `is_honest` and `is_online` parts?

Personally, I find this version more readable: `self.pings[vidx] >= now - max_round_len * PING_TIMEOUT`

node/src/components/consensus/highway_core/highway_testing.rs
```diff
@@ -215,20 +215,16 @@ impl HighwayValidator {
...
-                    | HighwayMessage::Timer(_)
-                    | HighwayMessage::RequestBlock(_) => vec![msg],
-                    HighwayMessage::WeAreFaulty(ev) => {
-                        panic!("validator equivocated unexpectedly: {:?}", ev);
-                    }
-                }
-            }
-            Some(DesFault::Mute) => {
+            Some(DesFault::Mute(from, till))
+                if *from <= delivery_time && delivery_time <= *till =>
```
goral09, line 226: Please add a comment explaining what this condition means, and why it has to have a special handling.

node/src/components/consensus/highway_core/highway_testing.rs
```diff
@@ -353,8 +366,11 @@ where
         Ok(())
     }
 
-    fn next_consensus_value(&mut self) -> ConsensusValue {
-        self.consensus_values.pop_front().unwrap_or_default()
+    fn next_consensus_value(&mut self, height: u64) -> ConsensusValue {
+        self.consensus_values
+            .get(height as usize)
+            .cloned()
+            .unwrap_or_default()
```
goral09, line 373: This is not _next_ consensus value anymore. Also, does it remove the value from the `consensus_value` queue? If not, does it mean we might propose the same value more than once?

node/src/components/consensus/highway_core/active_validator.rs
afck: No, we don't; that's the point of this line: We should always count ourselves as online, even if we just restarted and the `state` doesn't know about it yet. That's why, if the state doesn't know we're online, we add our own weight anyway.

node/src/components/consensus/tests/consensus_des_testing.rs
```diff
@@ -62,7 +62,7 @@ impl Display for ValidatorId {
 
 #[derive(Debug, Clone, Copy)]
 pub(crate) enum Fault {
-    Mute,
+    Mute(Timestamp, Timestamp),
```
goral09, line 65: Can you refactor it to a proper struct with fields and use meaningful names (possible comments).

node/src/components/consensus/highway_core/active_validator.rs
afck: Yes, it is.
`has_ping` returns `true` even if we have a ping or unit that's only slightly older, to prevent spam. So: If you send me two pings with less than one maximum round length in between, I won't gossip the second one.
And honest nodes will only produce a ping if their most recent ping or unit is old enough.

`has_ping` is a misnomer… maybe I should call it `needs_ping`, and negate it?

node/src/components/consensus/highway_core/highway.rs
afck: No, `has_dependency` is about asking whether we have a unit with a particular hash, etc.
If it asks us whether we have a particular ping, it just always returns false, because we don't even store the signatures and pings are never requested as dependencies.

All of this slightly abuses the vertices-and-dependencies system, but it avoids a lot of `Option`s or a whole new message type.

goral09 (commented):
Looks good, I can't see any blockers but I will give it another pass after the comments are addressed.

node/src/components/consensus/highway_core/highway/vertex.rs
afck: [It's the same for evidence](https://github.com/CasperLabs/casper-node/blob/b2e4b0485325896447c7d3e60477ba83ecd7b260/node/src/components/consensus/highway_core/evidence.rs#L81).

Where do you think it should be? Should all vertex type validation functions be `Highway` methods? Isn't it cleaner to have them in the module belonging to their type?

node/src/components/consensus/highway_core/active_validator.rs
goral09: That's fine, `has_ping` sounds good to me.

node/src/components/consensus/highway_core/highway/vertex.rs
goral09: Maybe it _looks_ cleaner but it doesn't seem right. I'm OK with verifying signature in the model but testing whether the creator is a validator – this should be part of the `Highway`. It's up to a specific controller how to interpret validity. 

Anyway, not a blocker and I certainly won't ask you to change it in this PR. Just something I wanted to share.

node/src/components/consensus/highway_core/highway_testing.rs
afck: But that was the problem: That we removed it from the queue and proposed it only once, even if the block got orphaned. The existing tests only worked because they never orphan a block.

This way, it is in fact the _next_ consensus value: On height H it will always return entry number H, which is the one that comes after its parent.

goral09: Ah, I see! The alternative implementation would detect orphaned consensus values and put it back in the queue.

node/src/components/consensus/highway_core/active_validator.rs
goral09: That confused me – that we first go through `state` and return the weight of all honest and online validators but we check if we are in that set ourselves.

node/src/components/consensus/tests/consensus_des_testing.rs
goral09: Actually, how do you feel about adding a separate variant? Something like `TemporarilyMute`, `PermanentlyMute`? I imagine it would be difficult to figure out the `end_mute` timestamp if I wanted to have the validator be mute/offline permanently.

node/src/components/consensus/highway_core/active_validator.rs
afck: Good point, I'll just move that weight calculation to `active_validator`.

node/src/components/consensus/highway_core/highway/vertex.rs
afck: OK, I'll look into it; maybe I'll create a follow-up PR.

node/src/components/consensus/highway_core/state.rs
afck: As discussed, I'll leave the inequality as it is, to avoid underflows in tests.

goral09 (approved):
Nice.

node/src/components/consensus/highway_core/highway/vertex.rs
goral09: I don't think the comment has been updated.

afck: Yes it has. GitHub just shows the snippet of the version you commented on.

PR #845, opened 2021-02-02
NDRS-839: Make consensus component `Send` by adding trait bounds and swapping `Arc` in for `Rc`

xcthulhu:


afck (approved):

xcthulhu:
bors merge

goral09:
Why do we need this? Every component should be accessed by a single thread at a time. There should be no need to send it across threads.

PR #846, opened 2021-02-02
NDRS-799: Introducing effect for missing_trie_keys

xcthulhu:


node/src/effect.rs
```diff
@@ -789,6 +789,25 @@ impl<REv> EffectBuilder<REv> {
         .await
     }
 
+    /// Gets missing descendant trie keys for an ancestral trie key.
+    #[allow(unused)]
```
EdHastingsCasperLabs, line 793: eh?

xcthulhu: The other function for doing `put_trie` is also unused (temporarily).

This effect is less involved, so I'll just shelve this PR for now until I use this function.

PR #848, opened 2021-02-03
EE-1163: Native transfer error charge

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1163

- This PR will now use a precomputed finalization result for user errors (https://github.com/CasperLabs/casperlabs-node/blob/4ad2e250ac839c41c9761e1aaba3281b17e34bb2/execution_engine/src/core/engine_state/execution_result.rs#L254-L275) in native transfer instead of early exit with 0 cost.
- During refactor I noticed I used wrong field for charging cost of wasmless transfer
- I also verified other test correctness in the meantime by changing default gas price to >1 and corrected failing tests.

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -574,6 +574,15 @@ where
             Err(e) => return Ok(ExecutionResult::precondition_failure(e)),
         };
 
+        let proposer_addr = proposer.to_account_hash();
+        let proposer_account = match tracking_copy
+            .borrow_mut()
+            .get_account(correlation_id, proposer_addr)
+        {
+            Ok(proposer) => proposer,
+            Err(e) => return Ok(ExecutionResult::precondition_failure(Error::Exec(e))),
+        };
+
         let mint_contract = match tracking_copy
             .borrow_mut()
             .get_contract(correlation_id, protocol_data.mint())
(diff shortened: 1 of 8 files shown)


execution_engine/src/core/engine_state/mod.rs
```diff
@@ -574,6 +578,15 @@ where
             Err(e) => return Ok(ExecutionResult::precondition_failure(e)),
         };
 
+        let proposer_addr = proposer.to_account_hash();
+        let proposer_account = match tracking_copy
+            .borrow_mut()
+            .get_account(correlation_id, proposer_addr)
+        {
+            Ok(proposer) => proposer,
+            Err(e) => return Ok(ExecutionResult::precondition_failure(Error::Exec(e))),
```
EdHastingsCasperLabs, line 587: nitpick: `Err(error)`

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -604,8 +617,96 @@ where
...
         let gas_limit = Gas::new(U512::from(std::u64::MAX));
 
+        let wasmless_transfer_gas_cost = Gas::new(U512::from(
+            protocol_data.system_config().wasmless_transfer_cost(),
+        ));
+
+        let wasmless_transfer_cost = match Motes::from_gas(
+            wasmless_transfer_gas_cost,
+            WASMLESS_TRANSFER_FIXED_GAS_PRICE,
+        ) {
```
EdHastingsCasperLabs, line 627: > let wasmless_transfer_cost = match Motes::from_gas
recommend renaming to `wasmless_transfer_motes`

Even though, for all intents & purposes they have the same underlying value, it seems needlessly confusing IMO to create a Motes instance w/ this name when the config'd value (_protocol_data.system_config().wasmless_transfer_cost()_), is also named wasmless_transfer_cost but is expressed as Gas.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -604,8 +617,96 @@ where
...
+            let proposer_main_purse = proposer_account.main_purse();
+
+            match tracking_copy
+                .borrow_mut()
+                .get_purse_balance_key(correlation_id, proposer_main_purse.into())
+            {
+                Ok(balance_key) => balance_key,
+                Err(_) => {
+                    return Ok(ExecutionResult::precondition_failure(
+                        Error::InsufficientPayment,
```
EdHastingsCasperLabs, line 646: This isn't actually `InsufficientPayment`, is it? Why not return the actual error instead of swallowing it?

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -604,8 +617,96 @@ where
...
+        let proposer_purse = proposer_account.main_purse();
+
+        let account_main_purse = account.main_purse();
+
+        let account_main_purse_balance_key = match tracking_copy
+            .borrow_mut()
+            .get_purse_balance_key(correlation_id, account_main_purse.into())
+        {
+            Ok(balance_key) => balance_key,
+            Err(_) => {
```
EdHastingsCasperLabs, line 661: Same question as above...is there a reason to return `InsufficientPayment` instead of the actual error?

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -604,8 +617,96 @@ where
...
+                    Error::InsufficientPayment,
+                ))
+            }
+        };
+
+        if account_main_purse_balance < wasmless_transfer_cost {
+            // We don't have minimum balance to operate and therefore we can't charge for user
+            // errors.
+            return Ok(ExecutionResult::precondition_failure(
+                Error::InsufficientPayment,
```
EdHastingsCasperLabs, line 684: This is the correct use of `InsufficientPayment`

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -641,104 +742,55 @@ where
                                 .write(Key::Account(public_key), StoredValue::Account(new_account))
                         }
                         None => {
+                            // Implies that execution_result is a failure.
                             return Ok(execution_result);
```
EdHastingsCasperLabs, line 746: I think we need to charge for this.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -641,104 +742,55 @@ where
...
-                    return Ok(ExecutionResult::Failure {
-                        error: Error::Exec(error),
-                        effect: Default::default(),
-                        transfers: Vec::default(),
-                        cost: Gas::default(),
-                    });
-                }
-            };
+            let source_purse_balance = {
+                let source_purse_balance_key = match tracking_copy
```
EdHastingsCasperLabs, line 766: if source purse is main purse, you already have it.

EdHastingsCasperLabs (approved):

execution_engine/src/core/engine_state/mod.rs
mpapierski: Good point! Fixed

PR #849, opened 2021-02-03
NO-TICKET: update RPC server

Fraser999:
This PR brings the JSON-RPC server more in line with [the error object as defined in the specs](https://www.jsonrpc.org/specification#error_object).

* -32601, "Method not found" is now handled correctly
* -32602, "Invalid params" is now handled correctly
* -32603, "Internal error" is not required
* our server error codes have been negated to fit the range -32000 to -32099

Due to shortcomings in a third-party crate, we're not able to distinguish between an error due to the request containing invalid JSON, or the request containing valid JSON, but not a valid JSON-RPC request.  As such our server returns -32700, "Parse error" for both cases.

### DIFF
--- a/node/src/components/rpc_server/http_server.rs
+++ b/node/src/components/rpc_server/http_server.rs
@@ -1,17 +1,46 @@
 use std::{convert::Infallible, time::Duration};
 
-use futures::future::{self};
-use hyper::server::{conn::AddrIncoming, Builder};
+use futures::future;
+use http::{Response, StatusCode};
+use hyper::{
+    server::{conn::AddrIncoming, Builder},
+    Body,
+};
+use serde::Serialize;
 use tokio::sync::oneshot;
 use tower::builder::ServiceBuilder;
 use tracing::{info, trace};
-use warp::Filter;
+use warp::{Filter, Rejection};
 
 use super::{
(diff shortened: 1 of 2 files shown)


node/src/components/rpc_server/http_server.rs
```diff
@@ -33,6 +62,24 @@ pub(super) async fn run<REv: ReactorEventT>(
...
+            response_builder
+                .error(warp_json_rpc::Error::METHOD_NOT_FOUND)
+                .map_err(|_| warp::reject())
+        });
+
+    // Catch requests which don't parse as JSON.
+    let parse_failure = warp::path(RPC_API_PATH).and_then(move || async move {
+        let error_response = new_error_response(warp_json_rpc::Error::PARSE_ERROR);
+        Ok::<_, Rejection>(error_response)
+    });
```
goral09, line 78: We won't know exactly _what_ error that was, right?

Fraser999: Yes, that's correct.  The response contains no hints as to why the request couldn't be parsed.

EdHastingsCasperLabs (approved):

Fraser999:
bors merge

PR #853, opened 2021-02-04
libp2p gossiper

marc-casperlabs:
This PR rips out the old gossiper and replaces it with libp2p's gossiper. It essentially renders the `small_network` component inoperable.

I have tried to keep the refactoring to a minimum and reuse as much of the old structure as possible, to not balloon the scope of this.


node/src/components/network/behavior.rs
```diff
@@ -122,12 +124,54 @@ impl Behavior {
...
     /// Initiates gossiping the given message.
-    pub(super) fn gossip(&mut self, message: GossipMessage) {
-        if let Err(error) = self.gossip_behavior.publish(&*TOPIC, message) {
+    pub(super) fn gossip(&mut self, message: GossipObject, max_gossip_message_size: u32) {
+        let data = match message.serialize(max_gossip_message_size as usize) {
+            Ok(encoded) => encoded,
+            Err(err) => {
+                warn!(?err, "{}: failed to serialize gossip message", self.our_id);
+                return;
+            }
```
goral09, line 133: I don't think it's a good idea to fail to gossip silently. Consensus, for example, requires for gossiping to work, if we fail to deliver consensus messages silently then the network will stall.

dwerner (changes requested):
I want to block this because it kills the small_network component.

sacherjj:
bors ping

sacherjj:
bors ping

sacherjj:
bors ping

sacherjj:
bors ping

sacherjj:
bors ping

PR #855, opened 2021-02-04
HWY-242: Fix DataSize implementation for HighwayProtocol

afck:
`estimate_heap_size` was called on `&HighwayProtocol` instead of `HighwayProtocol` and returned `0`.

https://casperlabs.atlassian.net/browse/HWY-242

### DIFF
--- a/node/src/components/consensus/cl_context.rs
+++ b/node/src/components/consensus/cl_context.rs
@@ -17,6 +17,7 @@ use crate::{
     NodeRng,
 };
 
+#[derive(DataSize)]
 pub(crate) struct Keypair {
     secret_key: Arc<SecretKey>,
     public_key: PublicKey,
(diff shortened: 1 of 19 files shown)


marc-casperlabs (approved):
It does look good to me, I would suggest we fix the remaining `impl`s in a separate PR once `datasize` has caught up. The `BTreeSet` one looks important at least.

afck:
It only matters if there are equivocations and endorsements, so at least we'll get useful data from our tests with this PR. But I'll work on the `BTreeSet` impl, too.

goral09 (approved):

afck:
https://github.com/CasperLabs/datasize-rs/pull/5 adds support for `BTreeSet`. (And I was wrong: It's already implemented for `PathBuf`.)

PR #857, opened 2021-02-04
NDRS-842: Have network retry connecting

xcthulhu:
https://casperlabs.atlassian.net/browse/NDRS-842

### DIFF
--- a/node/src/components/network.rs
+++ b/node/src/components/network.rs
@@ -17,6 +17,7 @@ use std::{
     fmt::{self, Debug, Display, Formatter},
     marker::PhantomData,
     num::NonZeroU32,
+    sync::{Arc, Mutex},
 };
 
 use datasize::DataSize;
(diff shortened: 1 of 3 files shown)


dwerner (commented):
I'm wondering if we should do some kind of back-off/delay between retries?

dwerner (approved):

dwerner (approved):
Cool, looks good to me.

xcthulhu:
Chatted with @dwerner; we should be using `tokio::sync::Mutex` instead of the standard library one.

xcthulhu:
> Chatted with @dwerner; we should be using tokio::sync::Mutex instead of the standard library one.

Okay, so this is tricky to do since we are using this mutex both syncronously and asynchronously now... we'll have to make a follow up ticket for this.

xcthulhu:
bors merge

EdHastingsCasperLabs (approved):

PR #865, opened 2021-02-05
NO-TICK: Store new finality signatures in the storage.

goral09:
As part of the work on NDRS-833 I've noticed that we don't store new finality signatures in the persistent storage. This PR fixes that problem.


afck (approved):

goral09:
bors merge

PR #869, opened 2021-02-05
NO-TICKET: Revert changes in finality signatures

darthsiroftardis:
Reverted changes made by NDRS-767: Remove Finality Signatures from Block


EdHastingsCasperLabs (changes requested):
@goral09 may have found the actual bug; blocking this for now.

PR #876, opened 2021-02-06
NO-TICKET/networking metrics

marc-casperlabs:
This adds generic networking metrics, counting queued message and requests for outgoing broadcasts, as well as direct messages.

Currently only implemented for `small_net` due to some time constraints, but the same structure should be used for the libp2p `networking` component. Having the same metrics on both will allow for some comparability.

These metrics should be especially useful when tweaking the chattiness of consensus.

### DIFF
--- a/node/src/components/networking_metrics.rs
+++ b/node/src/components/networking_metrics.rs
@@ -7,7 +7,7 @@ pub(crate) struct NetworkingMetrics {
     /// How often a request to send a message directly to a peer was made.
     pub(crate) direct_message_requests: IntCounter,
     /// Current number of open connections.
-    pub(crate) open_connnections: IntGauge,
+    pub(crate) open_connections: IntGauge,
     /// Number of messages still waiting to be broadcast.
     pub(crate) queued_broadcast_messages: IntGauge,
     /// Number of messages still waiting to be sent out.
@@ -48,7 +48,7 @@ impl NetworkingMetrics {
         Ok(NetworkingMetrics {
             broadcast_requests,
             direct_message_requests,
-            open_connnections,
+            open_connections: open_connnections,
             queued_broadcast_messages,
             queued_direct_messages,
             registry: registry.clone(),
@@ -65,7 +65,7 @@ impl Drop for NetworkingMetrics {
             .unregister(Box::new(self.direct_message_requests.clone()))
             .expect("did not expect deregistering direct_message_requests to fail");
         self.registry
-            .unregister(Box::new(self.open_connnections.clone()))
+            .unregister(Box::new(self.open_connections.clone()))
             .expect("did not expect deregistering open_connnections to fail");
         self.registry
             .unregister(Box::new(self.queued_broadcast_messages.clone()))
--- a/node/src/components/networking_metrics.rs
+++ b/node/src/components/networking_metrics.rs
@@ -1,0 +1,99 @@
+use prometheus::{IntCounter, IntGauge, Registry};
+
+/// Network-type agnostic networking metrics.
+pub(crate) struct NetworkingMetrics {
+    /// How often a request was made a component to broadcast.
+    pub(crate) broadcast_requests: IntCounter,
+    /// How often a request to send a message directly to a peer was made.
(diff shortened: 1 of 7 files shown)


node/src/components/networking_metrics.rs
```diff
@@ -0,0 +1,66 @@
+use prometheus::{IntCounter, IntGauge, Registry};
+
+/// Network-type agnostic networking metrics.
+pub(crate) struct NetworkingMetrics {
+    /// How often a request was made a component to broadcast.
```
afck, line 5:
```suggestion
    /// How often a request was made by a component to broadcast.
```

node/src/components/networking_metrics.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl NetworkingMetrics {
+    /// Creates a new instance of networking metrics.
+    pub fn new(registry: &Registry) -> Result<Self, prometheus::Error> {
+        let broadcast_requests =
+            IntCounter::new("net_broadcast_requests", "number of broadcasting requests")?;
+        let direct_message_requests = IntCounter::new(
+            "net_direct_message_requests",
+            "How often a request to send a message directly to a peer was made.",
```
afck, line 25: Maybe for consistency with `broadcast_requests`:
```suggestion
            "number of requests to send a message directly to a peer",
```
(Also for the other ones.)

node/src/components/small_network.rs
```diff
@@ -435,6 +449,9 @@ where
                         times_seen_asymmetric: 0,
                     },
                 );
+                self.net_metrics
+                    .open_connections
+                    .set(self.incoming.len() as i64);
```
afck, line 454: What about outgoing connections? Should `self.outgoing.len()` be added here?
(Sorry if my questions don't make sense; I'm not familiar with the `small_network` code.)

afck (approved):
Just some nit-picks. But someone more familiar with `small_network` should review this.

node/src/components/networking_metrics.rs
```diff
@@ -0,0 +1,66 @@
...
+            registry: registry.clone(),
+        })
+    }
+}
+
+impl Drop for NetworkingMetrics {
+    fn drop(&mut self) {
+        self.registry
+            .unregister(Box::new(self.broadcast_requests.clone()))
+            .expect("did not expect deregistering broadcast_requests to fail");
```
Fraser999, line 55: I'm not a fan of panicking code in drop impls.  Do we really need to `expect` here?

node/src/components/small_network.rs
marc-casperlabs: `small_network` currently has a 1:1 ratio of incoming to outgoing - if there is no matching outgoing connection for every incoming or vice verse, we attempt to rectify this. If it fails I believe we terminate both connections.

Fraser999: I think it's probably ok as is.  We don't retain asymmetric connections for too long, so this will give a reasonable representation of the number of notional connections where each is comprised of two real connections.

Eventually I believe we'll either change to libp2p (where each connection is bidirectional) or we'll update small_network to have bidirectional connections, so this metric will remain consistent after we make that transition.

Fraser999 (approved):
You don't feel a need to add a counter for the number of gossip requests made by the component?

Other than that, LGTM.

dwerner (approved):
No new stacking nitpicks on the others, looks good to me. Good to have more metrics around this.

EdHastingsCasperLabs (approved):
please put up a followup pr that addresses the nitpicks when you get a chance; merging this so we can get some use out of it in the meantime.

node/src/components/networking_metrics.rs
marc-casperlabs: This is currently in line with all of the other drop impls on metrics. If we want to change this, we should ticket this and establish a guideline.

My reasoning for panicking is that this will never happen and if it does, an assumption made was very incorrect. I'd not be confident to keep on running in this case and would want to make a lot of noise.

PR #879, opened 2021-02-07
HWY-249: Add the timestamp to `CandidateBlock`.

afck:
Validity of a proto block depends on the timestamp. If there are two different candidate blocks (belonging to different units) with the same proto block, we must not apply the validity result to both.

https://casperlabs.atlassian.net/browse/HWY-249

### DIFF
--- a/node/src/components/consensus/candidate_block.rs
+++ b/node/src/components/consensus/candidate_block.rs
@@ -3,21 +3,30 @@ use serde::{Deserialize, Serialize};
 
 use casper_types::PublicKey;
 
-use crate::{components::consensus::traits::ConsensusValueT, types::ProtoBlock};
+use crate::{
+    components::consensus::traits::ConsensusValueT,
+    types::{ProtoBlock, Timestamp},
+};
 
 /// A proposed block. Once the consensus protocol reaches agreement on it, it will be converted to
 /// a `FinalizedBlock`.
 #[derive(Clone, DataSize, Debug, PartialOrd, Ord, PartialEq, Eq, Hash, Serialize, Deserialize)]
 pub(crate) struct CandidateBlock {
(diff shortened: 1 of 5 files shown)


node/src/components/consensus/era_supervisor/era.rs
```diff
@@ -192,20 +192,25 @@ impl<I> Era<I> {
     pub(crate) fn resolve_validity(
         &mut self,
         proto_block: &ProtoBlock,
+        timestamp: Timestamp,
```
goral09, line 195: This is a timestamp of the unit that proposed the `ProtoBlock`?

afck: Yes

afck: …and it will be the timestamp of the `FinalizedBlock` and `Block` itself.

goral09 (approved):

node/src/components/consensus/era_supervisor/era.rs
goral09: Can you add a comment or improve the name? In order to figure out what the `timestamp` is here one has to go up a couple of levels and then down into another branch where it was created.

node/src/components/consensus/candidate_block.rs
```diff
@@ -27,6 +36,14 @@ impl CandidateBlock {
         &self.proto_block
     }
 
+    /// Returns the candidate block's timestamp, i.e. when the block was proposed.
+    ///
+    /// This is idential to the timestamp of the Highway unit, and the timestamp of the `Block`,
```
goral09, line 41:
```suggestion
    /// This is identical to the timestamp of the Highway unit, and the timestamp of the `Block`,
```

PR #880, opened 2021-02-07
HWY-246: Deduplicate synchronizer entries.

afck:
We are seeing high memory usage in tests, due to pending vertices in the Highway synchronizer. Currently a new entry in a synchronizer queue is created whenever we receive a vertex that we can't add to the protocol state yet, e.g. due to missing dependencies. If we receive the same vertex multiple times, it will be duplicated in multiple queue entries.

This PR turns the queues into maps, keyed by the vertex and the sending peer, and when we receive a duplicate it only updates the timestamp of the existing entry instead of creating a duplicate.

https://casperlabs.atlassian.net/browse/HWY-246

### DIFF
--- a/node/src/components/consensus/protocols/highway/synchronizer.rs
+++ b/node/src/components/consensus/protocols/highway/synchronizer.rs
@@ -24,7 +24,7 @@ mod tests;
 /// Incoming pre-validated vertices that we haven't added to the protocol state yet, and the
 /// timestamp when we received them.
 #[derive(DataSize, Debug)]
-pub(crate) struct PendingVertices<I, C>(HashMap<(I, PreValidatedVertex<C>), Timestamp>)
+pub(crate) struct PendingVertices<I, C>(HashMap<PreValidatedVertex<C>, HashMap<I, Timestamp>>)
 where
     C: Context;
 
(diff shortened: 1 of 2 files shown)


node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -21,6 +21,66 @@ use super::{HighwayMessage, ProtocolOutcomes, ACTION_ID_VERTEX};
...
+
+impl<I, C: Context> Default for PendingVertices<I, C> {
+    fn default() -> Self {
+        PendingVertices(Default::default())
+    }
+}
+
+impl<I: NodeIdT, C: Context> PendingVertices<I, C> {
+    /// Removes expired vertices and returns `true` if it is now empty.
+    fn remove_expired(&mut self, oldest: Timestamp) {
```
goral09, line 39: The comment says _returns `true` if it is now empty_ but the method does not return anything.

goral09: I don't know what `oldest` means in the context of this method. Can you give it a better name?

goral09:
Can you add a bit more to the description of your PR? We want to use it for the changelog later. It will make our life easier.

goral09 (approved):
Looks good to me. Have you tested it?

afck:
Only with the unit tests. I'm hoping we can test a bunch of small fixes like this one in a bigger test at once.

PR #881, opened 2021-02-07
HWY-250: Use `ConsensusValue`'s hash as a key in map.

goral09:
Instead of storing the whole `ConsensusValue` as a key in the map (where `ConsensusValue` is a set of all deploys that the block includes), we will store a hash of `ConsensusValue`. This will improve memory management.

https://casperlabs.atlassian.net/browse/HWY-250

### DIFF
--- a/node/src/components/consensus/candidate_block.rs
+++ b/node/src/components/consensus/candidate_block.rs
@@ -9,7 +9,7 @@ use casper_types::PublicKey;
 
 use crate::{
     components::consensus::traits::ConsensusValueT,
-    crypto::hash::Digest, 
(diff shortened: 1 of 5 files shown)


afck (commented):
That solves only the smaller part of the problem, and reduces the size of the synchronizer but not the protocol state.

Can we make `ConsensusValue` itself be a candidate block hash, so it's also just a hash in the protocol state?

afck (approved):
You're right, removing it from the protocol state, too, will be a bigger issure we can do later. This should improve memory usage.

goral09:
bors merge

PR #882, opened 2021-02-07
HWY-251: BlockValidator accepts alternative sources of data.

goral09:
https://casperlabs.atlassian.net/browse/HWY-251

After this PR, if fetching a `Deploy` fails due to a timeout then `BlockValidator` will retry downloading a deploy from an alternative source.

### DIFF
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -232,7 +232,7 @@ where
                 self.validation_states.retain(|key, state| {
                     if state.missing_deploys.contains(&deploy_hash) {
                         match state.source() {
-                            Some(peer) => { 
+                            Some(peer) => {
                                 info!(%deploy_hash, ?peer, "trying the next peer");
                                 // There's still hope to download the deploy.
                                 let (chainspec, block_timestamp) = &state.context;
@@ -255,7 +255,7 @@ where
                         true
                     }
                 });
-                
+
                 if retried {
                     // If we retried, we need to increase this counter.
                     self.in_flight.inc(&deploy_hash);
@@ -263,7 +263,8 @@ where
             }
             Event::DeployInvalid(deploy_hash) => {
                 info!(%deploy_hash, "deploy invalid");
-                // Deploy is invalid. There's no point waiting for other in-flight requests to finish.
+                // Deploy is invalid. There's no point waiting for other in-flight requests to
+                // finish.
                 self.in_flight.dec(&deploy_hash);
 
                 self.validation_states.retain(|key, state| {
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -48,8 +48,12 @@ pub enum Event<T, I> {
     DeployFound(DeployHash),
 
     /// A request to find a specific deploy, potentially from a peer, failed.
-    #[display(fmt = "deploy {} missing/invalid {}", _0, _1)]
-    DeployMissing(DeployHash, bool),
+    #[display(fmt = "deploy {} missing", _0)]
+    DeployMissing(DeployHash),
+
+    /// Deploy was invalid. Failed the chainspec test.
+    #[display(fmt = "deploy {} invalid", _0)]
+    DeployInvalid(DeployHash),
 
     /// An event changing the current state to BlockValidatorReady, once the chainspec has been
     /// loaded.
@@ -108,7 +112,7 @@ pub(crate) struct BlockValidatorReady<T, I> {
 impl<T, I> BlockValidatorReady<T, I>
 where
     T: BlockLike + Debug + Send + Clone + 'static,
-    I: Clone + Send + PartialEq + Eq + 'static,
+    I: Clone + Debug + Send + PartialEq + Eq + 'static,
 {
     fn handle_event<REv>(
         &mut self,
@@ -214,7 +218,8 @@ where
                     }
                 });
             }
-            Event::DeployMissing(deploy_hash, timeout) => {
+            Event::DeployMissing(deploy_hash) => {
+                info!(%deploy_hash, "request to download deploy timed out");
                 // A deploy failed to fetch. If there is still hope (i.e. other outstanding
                 // requests), we just ignore this little accident.
                 if self.in_flight.dec(&deploy_hash) != 0 {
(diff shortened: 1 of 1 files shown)


node/src/components/block_validator.rs
```diff
@@ -48,8 +48,8 @@ pub enum Event<T, I> {
     DeployFound(DeployHash),
 
     /// A request to find a specific deploy, potentially from a peer, failed.
-    #[display(fmt = "deploy {} missing", _0)]
-    DeployMissing(DeployHash),
+    #[display(fmt = "deploy {} missing/invalid {}", _0, _1)]
+    DeployMissing(DeployHash, bool),
```
dwerner, line 52: Why not have a new `DeployInvalid` variant?

node/src/components/block_validator.rs
```diff
@@ -229,6 +254,45 @@ where
...
+    let validate_deploy = move |result: FetchResult<Deploy, I>| match result {
+        FetchResult::FromStorage(deploy) | FetchResult::FromPeer(deploy, _) => {
+            if deploy
+                .header()
+                .is_valid(&chainspec.deploy_config, block_timestamp)
+            {
+                Event::DeployFound(deploy_hash)
+            } else {
+                info!(%deploy_hash, "deploy is invalid");
+                Event::DeployMissing(deploy_hash, false)
```
dwerner, line 283: Isn't this "invalid" rather than missing?

node/src/components/block_validator.rs
goral09: I can do that, sure :) I tried to have a minimal set of changes too much.

dwerner (approved):
Other than my nitpick, looks good to me.

node/src/components/block_validator.rs
goral09: There's just one event variant `DeployMissing` for both invalid deploys and timed out requests. In this case, deploy failed the test against the chainspec so it is invalid. The boolean is interpreted as `timed_out` so `false` here means it's invalid.

dwerner: Sure, I'm just thinking it would be an improvement to have an "invalid" variant. I won't block the pr on that though.

goral09: Done in [`37fd3f7` (#882)](https://github.com/CasperLabs/casper-node/pull/882/commits/37fd3f7ead882e9da4856cdb8e930d2a3444f674)

node/src/components/block_validator.rs
goral09: Done in [`37fd3f7` (#882)](https://github.com/CasperLabs/casper-node/pull/882/commits/37fd3f7ead882e9da4856cdb8e930d2a3444f674)

goral09:
Turning it to a draft for now as I'm seeing fatal errors (panics).

dwerner (approved):
:+1: looks good.

node/src/components/block_validator.rs
```diff
@@ -141,40 +171,27 @@ where
...
                             })
                             .collect();
                         effects.extend(fetch_effects);
 
                         entry.insert(BlockValidationState {
                             missing_deploys,
                             responders: smallvec![responder],
+                            sources: VecDeque::new(), /* This is empty b/c we create the first
+                                                       * request
+                                                       * using `sender`. */
```
afck, line 193:
```suggestion
                            sources: VecDeque::new(), /* This is empty b/c we create the first
                                                       * request using `sender`. */
```
Or maybe even better:
```suggestion
                            // This is empty b/c we create the first request using `sender`.
                            sources: VecDeque::new(),
```

node/src/components/block_validator.rs
```diff
@@ -202,15 +219,57 @@ where
...
+                let mut retried = false;
+
                 self.validation_states.retain(|key, state| {
                     if state.missing_deploys.contains(&deploy_hash) {
+                        match state.source() {
+                            Some(peer) => {
+                                info!(%deploy_hash, ?peer, "trying the next peer");
+                                // There's still hope to download the deploy.
+                                let (chainspec, block_timestamp) = &state.context;
+                                effects.extend(fetch_deploy(effect_builder, Arc::clone(chainspec), *block_timestamp, deploy_hash, peer));
```
afck, line 239: Line too long.

node/src/components/block_validator.rs
```diff
@@ -202,15 +219,57 @@ where
...
+                                // This validation state contains a failed deploy hash, it can never
+                                // succeed.
+                                state.responders.drain(..).for_each(|responder| {
+                                    effects.extend(responder.respond((false, key.clone())).ignore());
+                                });
+                                false
+                            }
+                        }
+                    } else {
+                        true
```
afck, line 255: This would be much easier to read as an early return.

afck (approved):

node/src/components/block_validator.rs
goral09: How would that look like? We're iterating over all collection of `self.validation_states`.

node/src/components/block_validator.rs
goral09: That's what `cargo fmt` does, not me.

node/src/components/block_validator.rs
goral09: `cargo fmt` ignored this.

node/src/components/block_validator.rs
afck: But we're inside of a closure here, so you can just:
```rust
if !state.missing_deploys.contains(&deploy_hash) {
    return true;
}
```

node/src/components/block_validator.rs
afck: I know, it sometimes gives up on a whole file, if it doesn't know how to reformat some long lines.

node/src/components/block_validator.rs
goral09: I will revert the boolean and change the order of `if { … } else { … }` but I am not a fan of early return.

afck (approved):

EdHastingsCasperLabs (approved):

node/src/components/block_validator.rs
goral09: I had to add yet another `if` so I added an early return this time.

node/src/components/block_validator.rs
```diff
@@ -230,33 +230,36 @@ where
...
-                                // succeed.
-                                state.responders.drain(..).for_each(|responder| {
-                                    effects.extend(responder.respond((false, key.clone())).ignore());
-                                });
-                                false
-                            }
+                        return true
+                    }
+                    if retried {
+                        // We don't want to retry downloading the same element more than once.
```
afck, line 236: Not sure I'm following. So if we have three sources we'd only try two of them?

goral09: If there are multiple blocks that depend on that particular deploy (multiple entries in `self.validation_states` where `state.missing_deploys.contains(&deploy_hash) = true`), we don't want to send a request to download that particular `deploy_hash` for each of them. It's enough to send one, and resolve all of the dependencies when it's downloaded.

afck: Ah, makes sense; I misunderstood, sorry! :+1:

goral09:
bors merge

PR #885, opened 2021-02-08
EE-1165: Reinvest rewards

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1165

This PR changes the reward distribution logic to reinvest validator/delegator rewards by increasing their stakes. To obtain the reward one has to withdraw or undelegate their bid.


henrytill (changes requested):
We need tests specifically validating the restaking behavior,

types/src/auction.rs
```diff
@@ -456,140 +458,66 @@ pub trait Auction:
...
-        }
-
-        let mut bids = detail::get_bids(self)?;
-
-        let bid = match bids.get_mut(&validator_public_key) {
-            Some(bid) => bid,
-            None => return Err(Error::ValidatorNotFound),
-        };
+            // Update stakes for a validator that has not been slashed
+            if let Some(bid) = bids.get_mut(&public_key) {
```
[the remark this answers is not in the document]
mpapierski: I figured since this condition is present in both update_* functions (originally it was also reading from global state on each call), so why not simplify them and avoid extra branching there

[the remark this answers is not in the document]
henrytill: I would prefer that we keep it as-is and continue to do this check here.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -652,6 +652,767 @@ fn should_withdraw_bids_after_distribute() {
...
+    builder.exec(distribute_request_2).commit().expect_success();
+
+    let total_supply_2 = builder.total_supply(None);
+    assert!(total_supply_2 > initial_supply_1);
+
+    let expected_total_reward_2 = *GENESIS_ROUND_SEIGNIORAGE_RATE
+        * (initial_supply_1
+            + expected_validator_1_payout_1
+            + expected_delegator_1_payout_1
+            + expected_delegator_2_payout_1);
```
henrytill, line 892: Maybe
```rust
    let expected_total_reward_2 = *expected_total_reward_1
        + expected_validator_1_payout_1
        + expected_delegator_1_payout_1
        + expected_delegator_2_payout_1;
```

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -652,6 +652,767 @@ fn should_withdraw_bids_after_distribute() {
...
+
+    let mut timestamp_millis =
+        DEFAULT_GENESIS_TIMESTAMP_MILLIS + DEFAULT_LOCKED_FUNDS_PERIOD_MILLIS;
+
+    let mut builder = InMemoryWasmTestBuilder::default();
+
+    builder.run_genesis(&DEFAULT_RUN_GENESIS_REQUEST);
+
+    // initial token supply
+    let initial_supply_1 = builder.total_supply(None);
```
henrytill, line 758: Let's just call it `initial_supply`.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -652,6 +652,767 @@ fn should_withdraw_bids_after_distribute() {
...
+    )
+    .build();
+
+    builder.exec(distribute_request_2).commit().expect_success();
+
+    let expected_total_reward_2 = *GENESIS_ROUND_SEIGNIORAGE_RATE
+        * (initial_supply
+            + validator_1_staked_amount_1
+            + validator_2_staked_amount_1
+            + validator_3_staked_amount_1);
```
henrytill, line 1281: (as above)

henrytill (commented):
Looking good, a couple small requests

types/src/auction.rs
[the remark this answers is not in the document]
mpapierski: Sure, I reverted the logic as much as I could incl. this check

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
mpapierski: No it won't work unfortunately without `GENESIS_ROUND_SEIGNIORAGE_RATE` multiplier but I realized I could simplify this a bit and have `let expected_total_reward_2 = *GENESIS_ROUND_SEIGNIORAGE_RATE * total_supply_2` assuming that `total_supply_2` would be moved before 2nd distribute call

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
mpapierski: Fixed

henrytill (approved):
Great work

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
henrytill: Gotcha

PR #888, opened 2021-02-09
NO-TICKET: generate node version consistently

Fraser999:
When building from a repo where the git tags were not up to date, the version the node reported was highly misleading, using the latest git tag rather than the version specified in `Cargo.toml`.  This misleading version is used to populate the output when running the node with `-h` or `-V` and is also reported via the status endpoints.  However, when the node first starts, a further version is output which is based upon the value in `Cargo.toml`.

This PR ensures the proper crate version as specified in the `Cargo.toml` is used in all these cases.

### DIFF
--- a/node/build.rs
+++ b/node/build.rs
@@ -4,7 +4,6 @@ use vergen::ConstantsFlags;
 
 fn main() {
     let mut flags = ConstantsFlags::empty();
-    flags.toggle(ConstantsFlags::SEMVER_LIGHTWEIGHT);
     flags.toggle(ConstantsFlags::SHA_SHORT);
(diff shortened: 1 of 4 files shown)


node/build.rs
```diff
@@ -4,7 +4,6 @@ use vergen::ConstantsFlags;
 
 fn main() {
     let mut flags = ConstantsFlags::empty();
-    flags.toggle(ConstantsFlags::SEMVER_LIGHTWEIGHT);
```
goral09, line 7: We won't be using semver anymore?

Fraser999: We will - it's just that the semver will be drawn from the `package.version` in node's Cargo.toml.

goral09 (approved):

Fraser999:
bors merge

PR #890, opened 2021-02-09
NO-TICKET: Reject pings by faulty validators.

afck:
We don't count faulty validators towards the "online" ones anyway, so there's no point processing their pings.


xcthulhu:
LGTM

xcthulhu (approved):

PR #891, opened 2021-02-09
NDRS-865: Syncing node shuts down when upgrade is due.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-865

In order for the joining node (using `linear_chain_sync`) component to synchronize with an upgraded network, it has to be able to detect when the upgrade is due (based on the locally installed chainspecs) and shut down. 

After this PR is merged, syncing node will detect that a new chainspec has been installed and shut down after finalizing a linear chain block that should be the last one before the upgrade.


node/src/components/linear_chain_sync.rs
```diff
@@ -144,6 +156,11 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
         let height = block_header.height();
         let hash = block_header.hash();
         trace!(%hash, %height, "downloaded linear chain block.");
+        if self.should_upgrade(block_header.era_id()) {
+            info!(era = block_header.era_id().0, "shutting down for upgrade");
+            self.stop_for_upgrade = true;
```
afck, line 161: Would it be cleaner to make this a new `State::StopForUpgrade` instead of a boolean?

afck (approved):
No blockers from my side, but another pair of eyes would be good.

node/src/app/cli.rs
```diff
@@ -170,7 +170,10 @@ impl Cli {
                 // .await?;
                 // initializer2_runner.run(&mut rng).await;
 
-                initializer_runner.run(&mut rng).await;
+                let initializer_exit = initializer_runner.run(&mut rng).await;
+                if initializer_exit.is_err() {
+                    return Result::Ok(());
```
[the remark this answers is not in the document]
goral09: I will add log entries.

node/src/reactor.rs
```diff
@@ -324,6 +331,26 @@ impl Drop for RunnerMetrics {
     }
 }
 
+/// Exit status of a runner instance.
+pub enum RunnerExitStatus {
+    /// Runner finished work without errors.
+    Ok,
+    /// Runner finished with an error.
+    Err,
+}
```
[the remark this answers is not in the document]
goral09: Initially, there were no exit codes but you're right now.

node/src/components/linear_chain_sync.rs
goral09: I followed the pattern from the `EraSupervisor` but you're right.

node/src/app/cli.rs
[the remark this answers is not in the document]
goral09: Actually, I can't do that since with `type RunnerExitStatus = Result<(), u8>` I can't easily add custom methods to `RunnerExitStatus` (it's a type alias for `Result` that is defined outside of my crate). I will rely on the previously logged messages.

node/src/components/linear_chain_sync.rs
goral09: Actually, this seems like a lot of code changes to only use for `linear_chain_sync` since it is all ignored in `linear_chain_fast_sync`.

node/src/reactor.rs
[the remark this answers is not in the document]
goral09: Done in [`2c15046` (#891)](https://github.com/CasperLabs/casper-node/pull/891/commits/2c15046eedb61aae6a831659179518023331ca70).

node/src/components/consensus/era_supervisor.rs
```diff
@@ -920,6 +916,13 @@ where
             .announce_disconnect_from_peer(sender)
             .ignore()
     }
+
+    pub(super) fn should_upgrade(&self, era_id: &EraId) -> bool {
+        match self.era_supervisor.next_upgrade_activation_point {
+            None => false,
+            Some(upgrade_point) => upgrade_point.should_upgrade(&era_id),
```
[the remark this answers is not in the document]
goral09: Done in [`2c15046` (#891)](https://github.com/CasperLabs/casper-node/pull/891/commits/2c15046eedb61aae6a831659179518023331ca70).

node/src/reactor/joiner.rs
```diff
@@ -762,14 +762,19 @@ impl reactor::Reactor for Reactor {
...
-                    Event::LinearChainSync,
-                    self.linear_chain_sync.handle_event(
-                        effect_builder,
-                        rng,
+                ConsensusAnnouncement::Handled(block_header) => {
+                    let mut effects = Effects::new();
+                    if block_header.switch_block() {
+                        let reactor_event =
+                            Event::ChainspecLoader(chainspec_loader::Event::CheckForNextUpgrade);
+                        effects.extend(self.dispatch_event(effect_builder, rng, reactor_event));
```
afck, line 770: Could this be done in `linear_chain_sync` instead of the reactor itself?

goral09: I have just removed it as that's wrong :grimacing:

goral09: We shouldn't be checking for the next upgrade on switch blocks but check if we should shut down on switch block! And that's going to happen in the `linear_chain_sync`.

node/src/reactor.rs
```diff
@@ -324,6 +331,12 @@ impl Drop for RunnerMetrics {
     }
 }
 
+/// Exit status of a runner instance.
+pub type RunnerExitStatus = Result<(), u8>;
+
+/// Exit code indicating that node should shut down for a scheduled upgrade.
+pub const UPGRADE_EXIT_CODE: u8 = 0;
```
Fraser999, line 338: I think it might be better to have an enum here, but this already conflicts with changes I have locally, so I'll make that change in my PR if you don't mind?  I assume your planned next steps wouldn't need this changed, so we'll not end up conflicting again here?

node/src/reactor/joiner.rs
```diff
@@ -762,14 +762,17 @@ impl reactor::Reactor for Reactor {
...
-                        effect_builder,
-                        rng,
+                ConsensusAnnouncement::Handled(block_header) => {
+                    let mut effects = Effects::new();
+                    let reactor_event = Event::LinearChainSync(
                         linear_chain_sync::Event::BlockHandled(block_header),
-                    ),
-                ),
+                    );
+                    effects.extend(self.dispatch_event(effect_builder, rng, reactor_event));
```
Fraser999, line 770: You could just construct `effects` here rather than extending an empty one?

node/src/effect/announcements.rs
```diff
@@ -241,7 +241,7 @@ impl Display for LinearChainAnnouncement {
 /// A chainspec loader announcement.
 #[derive(Debug, Serialize)]
 pub enum ChainspecLoaderAnnouncement {
-    /// New finality signature received.
+    /// New upgrade recognized.
```
Fraser999, line 244: Oops - thanks! :)

node/src/components/small_network.rs
```diff
@@ -154,9 +153,9 @@ where
     pending: HashSet<SocketAddr>,
     /// The interval between each fresh round of gossiping the node's public listening address.
     gossip_interval: Duration,
-    /// The hash of the chainspec.  We only remain connected to peers with the same
+    /// Name of the network we participate in. We only remain connected to peers with the same
     /// `genesis_config_hash` as us.
```
Fraser999, line 157:
```suggestion
    /// network name as us.
```

Fraser999 (approved):
A couple of non-blocking nitpicks only.

node/src/reactor.rs
goral09: I didn't use enum on purpose as that would tie, otherwise generic `Reactor/Runner`, to the domain. I thought that generic exit codes were a good compromise.

If you need to make it an enum I won't oppose.

goral09:
bors merge

PR #892, opened 2021-02-09
NDRS-807: Remove deploy/transfer hashes and proposer out of header

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-807

CHANGELOG:

- Moved three fields: `deploy_hashes`, `transfer_hashes` and `proposer` out of `BlockHeader` into new stuct `BlockBody`
- Refactored `linear_chain_sync` and `linear_chain_fast_sync` to have `Block` instead of `BlockHeader`
- Refactored `era_supervisor` to have `Block` when announcing handling of linear chain blocks.
- Refactored JsonBlock to account for the change in DTO Block structure.


node/src/reactor/joiner.rs
```diff
@@ -823,7 +823,7 @@ impl reactor::Reactor for Reactor {
 
             Event::LinearChainAnnouncement(LinearChainAnnouncement::BlockAdded {
                 block_hash,
-                block_header,
+                block: block_header,
```
dwerner, line 826: This is not clear: is it a block or a block_header?

node/src/reactor/validator.rs
```diff
@@ -906,7 +906,7 @@ impl reactor::Reactor for Reactor {
             }
             Event::LinearChainAnnouncement(LinearChainAnnouncement::BlockAdded {
                 block_hash,
-                block_header,
+                block: block_header,
```
dwerner, line 909: Same as above.

dwerner (approved):
Looks good to me, I had one nitpick, but assuming that's addressed, I think we need to let @zie1ony know that this affects the event stream serialized type.

node/src/reactor/joiner.rs
darthsiroftardis: It should be block_header

node/src/reactor/validator.rs
darthsiroftardis: `block_header`

node/src/components/consensus/era_supervisor.rs
```diff
@@ -493,16 +493,16 @@ where
 
     pub(super) fn handle_linear_chain_block(
         &mut self,
-        block_header: BlockHeader,
+        block: Block,
```
goral09, line 496: Why do we need a whole `Block` here if all you ever use are header's fields?

afck: I think it was because of `announce_block_handled` below.

node/src/types/block.rs
```diff
@@ -743,36 +717,98 @@ impl FromBytes for BlockHeader {
...
+/// The body portion of a block.
+#[derive(Clone, DataSize, Ord, PartialOrd, Eq, PartialEq, Hash, Serialize, Deserialize, Debug)]
+pub struct BlockBody {
+    proposer: PublicKey,
+    deploy_hashes: Vec<DeployHash>,
+    transfer_hashes: Vec<DeployHash>,
+}
+
+impl BlockBody {
+    /// Create a new body from deploy and transfer hashes.
```
afck, line 753:
```suggestion
    /// Creates a new body from deploy and transfer hashes.
```
(Method doc comments are usually in the third person; also below.)

node/src/types/block.rs
```diff
@@ -859,7 +895,7 @@ impl Display for BlockSignatures {
 pub struct Block {
     hash: BlockHash,
     header: BlockHeader,
-    body: (), // TODO: implement body of block
+    body: BlockBody,
```
afck, line 898: Might be easier to just put the three fields directly into `Block`; I imagine we won't be using the `BlockBody` without the header anywhere?
(Just a thought; I don't have a strong opinion on that.)

afck (approved):

node/src/components/consensus/era_supervisor.rs
goral09: I thought it's needed by the `linear_chain` but it's for `linear_chain_sync`. Makes sense.

xcthulhu (changes requested):
Block bodies and block headers need to stored separately now.

Looks like we have to make the appropriate changes to `components::storage::handle_storage_request` to enable this.

goral09:
> Block bodies and block headers need to stored separately now.
> 
> Looks like we have to make the appropriate changes to `components::storage::handle_storage_request` to enable this.

Can we make this in a separate PR?

xcthulhu:
> Can we make this in a separate PR?

It can't slip, so I'd rather not.

goral09:
> > Can we make this in a separate PR?
> 
> It can't slip, so I'd rather not.

Not that it would slip but this is something that I will need/work on as part of my work. I would like to avoid the potential conflicts. 

It that part of work doesn't have to be made in this PR then I would prefer to leave it out.

node/src/components/storage.rs
```diff
@@ -351,14 +354,33 @@ impl Storage {
         Ok(match req {
             StorageRequest::PutBlock { block, responder } => {
                 let mut txn = self.env.begin_rw_txn()?;
-                let outcome = txn.put_value(self.block_db, block.hash(), &block, true)?;
+                if !txn.put_value(
+                    // It's not really appropriate to put block bodes in the metadata database,
+                    // but we have too many databases in storage and we can't make any more.
```
goral09, line 359: Why we can't have more?

node/src/components/storage.rs
```diff
@@ -351,14 +354,33 @@ impl Storage {
...
+
+                    // We can't put this in the block_header_db because we scan it to make our
+                    // indices at startup, so everything in that database must be a block header.
+                    self.block_metadata_db,
```
goral09, line 363: Aren't we overwriting finality signatures with this?

goral09: Yes, we are. We can't do that.

goral09 (changes requested):
We cannot reuse `block_metadata_db` b/c we will be overwriting finality signatures.

node/src/components/storage.rs
xcthulhu: Fixed in 8039fb9ccf0e55a813689f9d44c7fb190e1f42b5

node/src/components/storage.rs
xcthulhu: Fixed in 8039fb9ccf0e55a813689f9d44c7fb190e1f42b5

afck (approved):
Looks good to me, but method comments should be third person, and I'd wait for another approval from someone more familiar with storage.

Fraser999 (approved):
Approved assuming the multiple doc comment changes Andreas asked for are implemented.

node/src/components/storage.rs
```diff
@@ -351,14 +359,28 @@ impl Storage {
...
+                )? {
+                    error!("Could not insert block body for block: {}", block);
+                    txn.abort();
+                    return Ok(responder.respond(false).ignore());
+                }
+                if !txn.put_value(self.block_header_db, block.hash(), block.header(), true)? {
+                    error!("Could not insert block header for block: {}", block);
+                    txn.abort();
+                    return Ok(responder.respond(false).ignore());
+                }
```
goral09, line 376: We should try to remove the body if we failed to insert the header.

darthsiroftardis:
@xcthulhu @goral09 Can you please re-view to see if the changes are satisfactory

node/src/components/storage.rs
darthsiroftardis: Right but this is in the case of a failure or some error to store the the block

goral09: Yes, I understand that. I just want to make sure that we're not leaving any leftover, unnecessary state.

goral09: I checked. It does revert it.

darthsiroftardis: Aaaah okay, sorry my mistake

goral09 (approved):

xcthulhu (approved):

PR #893, opened 2021-02-09
NDRS-808: Merge era_end and next_era_validator_weights.

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-808

CHANGELOG:

- Renamed `EraEnd` to `EraReport`
- Created a new umbrella struct titled `EraEnd` to contain `EraReport` and moved `next_era_validator_weights` into `EraEnd`
- Refactored `FinalizedBlock` to contain `EraReport`
- Refactored `BlockHeader` to contain the *new* `EraEnd`
- Added a unit test to check serialization for the *new* `EraEnd` struct

Sample Output
Screenshot from 2021-02-09 15-33-15


node/src/types/block.rs
```diff
@@ -658,7 +678,13 @@ impl BlockHeader {
...
     /// The validators for the upcoming era and their respective weights.
     pub fn next_era_validator_weights(&self) -> Option<&BTreeMap<PublicKey, U512>> {
-        self.next_era_validator_weights.as_ref()
+        match &self.era_end {
+            Some(era_end) => {
+                let validator_weights = &era_end.next_era_validator_weights;
+                Some(validator_weights)
+            }
+            None => None,
+        }
```
goral09, line 687:
```rust
self.era_end.as_ref().map(|era_end| era_end.next_era_valdator_weights)
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -765,6 +765,7 @@ where
                     rewards: tbd.rewards,
                     equivocators: era.accusations(),
                     inactive_validators: tbd.inactive_validators,
+                    next_era_validator_weights: BTreeMap::default(),
```
afck, line 768: I'd prefer using a different type in `FinalizedBlock` instead of just putting an empty map in there.

I'd actually keep `EraEnd` without `next_era_validator_weights`, and call the new structure _with_ that field `Switch` or `SwitchBlockData` or something maybe?

node/src/types/block.rs
```diff
@@ -978,13 +1006,17 @@ impl Block {
         let state_root_hash = Digest::random(rng);
         let finalized_block = FinalizedBlock::random(rng);
         let parent_seed = Digest::random(rng);
+        let next_era_validator_weights = match finalized_block.era_end {
+            Some(_) => Some(BTreeMap::<PublicKey, U512>::default()),
```
afck, line 1010: I'd rather use a non-empty random map. An empty set of validators makes no sense.

xcthulhu: `next_era_validator_weights` could naturally contain `finalized_block.proposer`, and you could generate some other ones along with that.

afck: It doesn't have to contain the proposer, though; they could have unbonded, and be out in the next era. Would be good to test both cases. (Although probably at the moment we don't have any tests where that matters.)

node/src/components/consensus/era_supervisor.rs
darthsiroftardis: Created a new struct called `SwitchBlockData` and inserted into `BlockHeader`

node/src/types/block.rs
darthsiroftardis: Removed default scenario

darthsiroftardis:
@afck is this something more appropriate:
Screenshot from 2021-02-11 13-37-14

node/src/types/block.rs
```diff
@@ -580,11 +666,10 @@ pub struct BlockHeader {
     body_hash: Digest,
     random_bit: bool,
     accumulated_seed: Digest,
-    era_end: Option<EraEnd>,
+    switch_block_data: Option<SwitchBlockData>,
```
EdHastingsCasperLabs, line 669: I'm not a fan of suffixing things with "_data" ... it's _all_ data. Also, now we have to describe another concept in addition to the era concept. Also, this is going to affect downstream consumers which it would be better to avoid if possible. 

I realize @afck asked for this but I would prefer that this be returned to era_end / EraEnd.

EdHastingsCasperLabs (changes requested):
See naming comment...

node/src/types/block.rs
afck: Makes sense.

We just need a name for the other type then, the one without the validator set, containing equivocators, rewards and inactive validators. I'm really out of ideas there. `ValidatorPerformance`? :grimacing:

afck: `ValidatorRating`? `EraEndJudgement`? `EraReport`?

darthsiroftardis: Renamed `SwitchBlockData` to `EraEnd` and the OG `EraEnd` to `EraReport`

node/src/types/block.rs
```diff
@@ -63,18 +63,38 @@ static ERA_END: Lazy<EraEnd> = Lazy::new(|| {
...
     let inactive_validators = vec![public_key_3];
 
-    EraEnd {
+    EraReport {
         equivocators,
         rewards,
         inactive_validators,
     }
 });
+static ERA_END: Lazy<EraEnd> = Lazy::new(|| {
```
EdHastingsCasperLabs, line 72: If "era report" is the new thing, this should be ERA_REPORT

node/src/types/block.rs
```diff
@@ -63,18 +63,38 @@ static ERA_END: Lazy<EraEnd> = Lazy::new(|| {
...
+            U512::from(456),
+        );
+        next_era_validator_weights.insert(
+            PublicKey::from(&SecretKey::ed25519([6u8; SecretKey::ED25519_LENGTH])),
+            U512::from(789),
+        );
+        next_era_validator_weights
+    };
+
+    let era_end = EraReport::doc_example().clone();
```
EdHastingsCasperLabs, line 89: era_report

node/src/types/block.rs
```diff
@@ -63,18 +63,38 @@ static ERA_END: Lazy<EraEnd> = Lazy::new(|| {
...
+    EraEnd::new(era_end, next_era_validator_weights)
+});
 static FINALIZED_BLOCK: Lazy<FinalizedBlock> = Lazy::new(|| {
     let deploy_hashes = vec![*Deploy::doc_example().id()];
     let random_bit = true;
     let proto_block = ProtoBlock::new(deploy_hashes, vec![], random_bit);
     let timestamp = *Timestamp::doc_example();
-    let era_end = Some(EraEnd::doc_example().clone());
+    let era_end = Some(EraReport::doc_example().clone());
```
EdHastingsCasperLabs, line 97: era_report

EdHastingsCasperLabs (approved):

afck (approved):

PR #895, opened 2021-02-09
NDRS-875 - track metric for connected peers

dwerner:
Add a metric for the total count of connected peers.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -453,6 +453,8 @@ where
                     .open_connections
                     .set(self.incoming.len() as i64);
 
+                self.net_metrics.peers.set(self.peers().len() as i64);
+
                 // If the connection is now complete, announce the new peer before starting reader.
                 effects.extend(self.check_connection_complete(effect_builder, peer_id.clone()));
 
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -686,7 +686,9 @@ where
                         their_network=?network_name,
                         "dropping connection due to network name mismatch"
                     );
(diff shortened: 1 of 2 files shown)


goral09:
We already have it under `/status` endpoint. Why do we need a metrics for that as well?

dwerner:
> We already have it under `/status` endpoint. Why do we need a metrics for that as well?

SRE is asking for it, so that we can build charts based on this.

goral09 (approved):
OK.

node/src/components/small_network.rs
```diff
@@ -453,6 +453,8 @@ where
                     .open_connections
                     .set(self.incoming.len() as i64);
 
+                self.net_metrics.peers.set(self.peers().len() as i64);
```
Fraser999, line 456: I think this might be best moved to [`handle_message()`](https://github.com/CasperLabs/casper-node/blob/23da6ee23c1f855dbe249dc0448e0a7effbb2661/node/src/components/small_network.rs#L682-L701) where we decide whether to keep a new connection or not based on the handshake message.  We'd probably want it called for the case where we remove a peer and where we keep a peer.

Also, I think we should add this metric to the new `network` component in this PR so it doesn't get forgotten about.  That should go in where we handle `Event::ConnectionEstablished` and `Event::ConnectionClosed` I suppose.

dwerner: It would appear that `NetworkingMetrics` are not yet implemented for the new `network` component. I think @marc-casperlabs mentioned that there should be a todo for that?

Fraser999 (approved):

PR #896, opened 2021-02-09
NDRS-876 - track metric for last block time

dwerner:
Adds metrics for `time_of_last_finalized_block`, `block_completion_duration` and fixes an unregistered metric `time_of_last_proposed_block`. 

Talking with @EdHastingsCasperLabs `time_of_last_added_block` wasn't going to be meaningful, so we pivoted to instead record `block_completion_duration.`

### DIFF
--- a/node/src/components/consensus/metrics.rs
+++ b/node/src/components/consensus/metrics.rs
@@ -10,7 +10,9 @@ pub struct ConsensusMetrics {
     /// Amount of finalized blocks.
     finalized_block_count: IntCounter,
     /// Timestamp of the most recently accepted proto block.
-    time_of_last_proposed_block: Gauge,
+    time_of_last_proposed_block: IntGauge,
+    /// Timestamp of the most recently finalized block.
+    time_of_last_finalized_block: IntGauge,
     /// The Current era.
     pub current_era: IntGauge,
     /// registry component.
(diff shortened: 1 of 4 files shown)


node/src/components/linear_chain.rs
```diff
@@ -594,3 +601,34 @@ where
         }
     }
 }
+
+#[derive(Debug)]
+struct LinearChainMetrics {
+    /// Time in milliseconds since the unix epoch that the last block was added.
+    time_of_last_added_block: IntGauge,
```
afck, line 608: Where does this get updated?

node/src/components/consensus/metrics.rs
```diff
@@ -46,13 +55,15 @@ impl ConsensusMetrics {
     pub(crate) fn finalized_block(&mut self, finalized_block: &FinalizedBlock) {
         let time_since_proto_block = finalized_block.timestamp().elapsed().millis() as f64;
         self.finalization_time.set(time_since_proto_block);
+        self.time_of_last_finalized_block
+            .set(finalized_block.timestamp().millis() as i64);
         self.finalized_block_count.inc();
```
goral09, line 60: While you're at it – can you change `finalized_block_count` to `IntGauge` and use the `.finalized_block.height()`? Otherwise, this metric does not survive restart.

node/src/components/linear_chain.rs
goral09, line 608: Where is this metric updated? Also, we already have very similar metrics in `BlockExecutor` and `Consensus` so why do we need another one?

dwerner: I had updated the description but I neglected to push the last commit.

node/src/components/consensus/metrics.rs
goral09: ping.

dwerner: Will do.

goral09: Thanks

node/src/components/linear_chain.rs
dwerner: Sorry, stale commit.

afck (approved):

EdHastingsCasperLabs (approved):

PR #898, opened 2021-02-10
NDRS-837: Add account-address subcommand

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-837

CHANGELOG:

- Added a new sub-command called `account-address` with a single arg `--public-key` which takes a hex formatted public key or a path to a key and outputs the associated account-hash for the public key

Sample Output:
Screenshot from 2021-02-10 10-56-44

### DIFF
--- a/client/src/account_address.rs
+++ b/client/src/account_address.rs
@@ -13,16 +13,18 @@ enum DisplayOrder {
 }
 
 /// Handles providing the arg for and retrieval of the key.
-mod key {
+mod public_key {
     use casper_node::crypto::AsymmetricKeyExt;
     use casper_types::AsymmetricType;
 
     use super::*;
 
     const ARG_NAME: &str = "public-key";
-    const ARG_SHORT: &str = "pk";
+    const ARG_SHORT: &str = "p";
     const ARG_VALUE_NAME: &str = "FORMATTED STRING or PATH";
-    const ARG_HELP: &str = "Path to the public key or the formatted hex string";
+    const ARG_HELP: &str = "This must be a properly formatted public key. The public key may instead be read in from a file, in which case \
+        enter the path to the file as the --public-key argument. The file should be one of the two public \
+        key files generated via the `keygen` subcommand; \"public_key_hex\" or \"public_key.pem\"";
 
     pub(super) fn arg() -> Arg<'static, 'static> {
         Arg::with_name(ARG_NAME)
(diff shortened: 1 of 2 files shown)


dwerner (approved):

client/src/account_address.rs
```diff
@@ -0,0 +1,89 @@
...
+use crate::{command::ClientCommand, common};
+
+/// This struct defines the order in which the args are shown for this subcommand's help message.
+enum DisplayOrder {
+    Verbose,
+    Key,
+}
+
+/// Handles providing the arg for and retrieval of the key.
+mod key {
```
Fraser999, line 16:
```suggestion
/// Handles providing the arg for and retrieval of the public key.
mod public_key {
```

client/src/account_address.rs
```diff
@@ -0,0 +1,89 @@
...
+    use casper_node::crypto::AsymmetricKeyExt;
+    use casper_types::AsymmetricType;
+
+    use super::*;
+
+    const ARG_NAME: &str = "public-key";
+    const ARG_SHORT: &str = "pk";
```
Fraser999, line 23: I don't think this will work - did you try it?  The short args are always single chars afaik.

Either way, could you change it to `p` as per the task description please?

client/src/account_address.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        value.to_string()
+    }
+}
+
+pub struct GenerateAccountHash {}
+
+impl<'a, 'b> ClientCommand<'a, 'b> for GenerateAccountHash {
+    const NAME: &'static str = "account-address";
+    const ABOUT: &'static str = "Generate an account hash from a given public key";
```
Fraser999, line 66:
```suggestion
    const ABOUT: &'static str = "Generates an account hash from a given public key";
```

client/src/account_address.rs
```diff
@@ -0,0 +1,89 @@
...
+mod key {
+    use casper_node::crypto::AsymmetricKeyExt;
+    use casper_types::AsymmetricType;
+
+    use super::*;
+
+    const ARG_NAME: &str = "public-key";
+    const ARG_SHORT: &str = "pk";
+    const ARG_VALUE_NAME: &str = "FORMATTED STRING or PATH";
+    const ARG_HELP: &str = "Path to the public key or the formatted hex string";
```
Fraser999, line 25: This should be way more helpful.  It should be more like the help string for the `query-state` subcommand, as specified in the task description.

Fraser999 (changes requested):

client/src/account_address.rs
```diff
@@ -0,0 +1,89 @@
...
+        let key_string = key::get(matches);
+        let public_key = PublicKey::from_hex(key_string)
+            .unwrap_or_else(|error| panic!("error in retrieving public key: {}", error));
+        let account_hash = public_key.to_account_hash();
+
+        if verbosity_level == 0 {
+            verbosity_level += 1
+        }
+
+        casper_client::pretty_print_at_level(&account_hash, verbosity_level)
```
Fraser999, line 87: I think it's safe to replace this with just a simple `println!()` of the account hash - no need for verbosity levels here, nor json-encoding.

client/src/account_address.rs
darthsiroftardis: Applied suggestion

client/src/account_address.rs
darthsiroftardis: Change to simply `p`

client/src/account_address.rs
darthsiroftardis: Applied suggestion

client/src/account_address.rs
darthsiroftardis: Changed description to more closely reflect that in `query-state`

client/src/account_address.rs
darthsiroftardis: Replaced with a simple `println` statement

EdHastingsCasperLabs (approved):

Fraser999 (approved):

PR #900, opened 2021-02-10
NDRS-879: Log active/inactive stake.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-879

### DIFF
--- a/node/src/components/consensus/highway_core/state.rs
+++ b/node/src/components/consensus/highway_core/state.rs
@@ -12,7 +12,7 @@ pub(crate) use params::Params;
 use quanta::Clock;
 pub(crate) use weight::Weight;
 
-pub(crate) use panorama::{Observation, Panorama};
+pub(crate) use panorama::{Observation, Panorama, PanoramaStakes};
 pub(super) use unit::Unit;
 
 use std::{
@@ -405,9 +405,9 @@ impl<C: Context> State<C> {
             let block = Block::new(fork_choice, value, self);
             self.blocks.insert(hash, block);
         }
+        trace!(?unit, "adding valid unit to the protocol state");
         self.add_ping(unit.creator, unit.timestamp);
         self.units.insert(hash, unit);
-
         // Update the panorama.
         let unit = self.unit(&hash);
         let creator = unit.creator;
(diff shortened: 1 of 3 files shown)


node/src/components/consensus/highway_core/state.rs
```diff
@@ -429,6 +429,8 @@ impl<C: Context> State<C> {
             }
         };
         self.panorama[creator] = new_obs;
+        let panorama_stakes = PanoramaStakes::new(&self.panorama, &self.weights);
+        info!(?panorama_stakes, "panorama stakes");
```
afck, line 433: With 100 validators that will be 100 INFO-level logs per round; isn't that a bit spammy? Wouldn't it be enough to print it once per round?

node/src/components/consensus/highway_core/state/panorama.rs
```diff
@@ -236,3 +237,63 @@ impl<C: Context> Panorama<C> {
...
+                    faulty_weight += *weights.get(idx);
+                    ParticipationStatus::Faulty
+                }
+                Observation::None => {
+                    passive_weight += *weights.get(idx);
+                    ParticipationStatus::None
+                }
+            })
+            .collect();
+        let validators_participation = ValidatorMap::from(panorama_voted_stake);
```
afck, line 289: I think this can be [`collect`ed directly as a `ValidatorMap`](https://github.com/CasperLabs/casper-node/blob/9056ac7e8d93d84a5e3e61825a29eb2f0e22440a/node/src/components/consensus/highway_core/validators.rs#L185), instead of allocating a `Vec` first.

afck (commented):
I'd prefer to avoid adding more logic to the `state` module, and move this kind of overview one or two levels higher, to `highway_core` or even to `protocols::highway`. `PanoramaStakes` doesn't need to be inside the `panorama` module.

We could print it in `Highway::add_valid_unit` instead. But I think it would be less spammy if we just set a timer in `HighwayProtocol` and logged the participation status e.g. once per minute, or once per minimum round length?

We should also add the instance ID, so it's clear which era it belongs to.

afck:
If we move it out of `state`, we also have access to the actual public keys instead of just indices.
Would it make sense to, instead of printing `A`, `N` or `F` for every validator, just print the public keys of the inactive ones and the faulty ones?

goral09:
@afck It's the second time when you're commenting on my approach to this. I think it will be quicker if you do it yourself.

afck:
OK, I'll give it a try. :+1:

PR #901, opened 2021-02-10
EE-1170: evict non-participating validators

henrytill:
This PR:
- adds an `inactive` field to the `auction::Bid` structure
- passes eviction information to `Auction::run_auction`, where evicted validators have their bids updated to set `inactive` to `true`.
- adds an `Auction::activate_bid` method to activate an inactive validator.

cc @fizyk20: Turning on eviction breaks the two `run_equivocator_network` tests.  I have marked them with `#[ignore]`, but I will need your help to address this issue before merging.


EdHastingsCasperLabs (approved):

types/src/auction.rs
```diff
@@ -324,15 +328,19 @@ pub trait Auction:
 
         // Process bids
         let mut bids_modified = false;
-        for bid in bids.values_mut() {
-            bids_modified = bid.process(era_end_timestamp_millis)
+        for (validator_public_key, bid) in bids.iter_mut() {
+            bids_modified = bid.process(era_end_timestamp_millis);
+
+            if evicted_validators.contains(validator_public_key) {
+                bids_modified = bid.deactivate()
```
mpapierski, line 335: I believe it should be more like:

```suggestion
                if bid.deactivate() {
                    bids_modified = true;
                }
```
this way bids_modified would be set to true when one of the bids are deactivated. Otherwise if (lets say) last bid.deactivate would return false, all previous modifications would be lost?

types/src/auction/bid/mod.rs
```diff
@@ -196,6 +207,18 @@ impl Bid {
         vesting_schedule.initialize(staked_amount)
     }
 
+    /// Sets given bid's `inactive` field to `false`
+    pub fn activate(&mut self) -> bool {
+        self.inactive = false;
```
mpapierski, line 212: Personal preference nitpick: Unless I'm missing something special about this name I'd prefer "active" name over "inactive" as with "inactive" the meaning is already negated:

```rust
pub fn activate(&mut self) {
        self.active = true;
}
```

And also in logic above doesn't need negation

```diff
- .filter(|(_public_key, bid)| bid.vesting_schedule().is_some() && !bid.inactive())
+ .filter(|(_public_key, bid)| bid.vesting_schedule().is_some() && bid.is_active())
```

henrytill: I chose `inactive` because the default value, `false`, should be the default case: all bids are active by default.

EdHastingsCasperLabs: I prefer it the way henry has it, as the starting and most common state is `active`; thus defaulting to inactive: false (default value of bool) makes the most logical sense to me.

types/src/auction.rs
henrytill: `deactivate` always returns true though

types/src/auction/bid/mod.rs
EdHastingsCasperLabs: ninja'd

henrytill:
bors retry

PR #902, opened 2021-02-11
HWY-252: Pause consensus if execution lags behind finalization.

afck:
In a test with 75 validators and 1000 transactions per block we observed that block execution was much slower than consensus. That didn't stop consensus from finalizing all the era-0 blocks, though, which ended up in the block executor queue. Then there was a gap that would have lasted several hours, where consensus waited for the switch block to be executed, so it could start era 1.

This PR makes consensus wait if the block executor falls behind by more than 3 blocks (configurable): Whenever the height of the latest executed block is more than 3 below the height of the latest finalized block, consensus will switch to "paused" mode, using the mechanism introduced in https://github.com/CasperLabs/casper-node/pull/842.

https://casperlabs.atlassian.net/browse/HWY-252


node/src/components/consensus/era_supervisor.rs
```diff
@@ -370,6 +373,22 @@ where
     pub(crate) fn stop_for_upgrade(&self) -> bool {
         self.stop_for_upgrade
     }
+
+    fn executed_block(&mut self, block_header: &BlockHeader) {
+        self.next_executed_height = self.next_executed_height.max(block_header.height() + 1);
```
goral09, line 378: Is it possible for the `block_header.height()` to be lower than `self.next_executed_height`?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -370,6 +373,22 @@ where
     pub(crate) fn stop_for_upgrade(&self) -> bool {
         self.stop_for_upgrade
     }
+
+    fn executed_block(&mut self, block_header: &BlockHeader) {
```
goral09, line 377: Can you add a comment? I got a bit lost between `next_executed_block`, `next_executed_height` :dizzy_face:

goral09 (approved):

node/src/components/consensus/era_supervisor.rs
afck: I think it might happen, during joining.
But whenever you're an active, fully synced, participating validator that is in a position to actually contribute to consensus, the latest executed height will be slightly behind the latest finalized one.

afck:
(Waiting for the integration test before merging.)

node/src/components/consensus/era_supervisor.rs
afck: Done.

EdHastingsCasperLabs (approved):

PR #904, opened 2021-02-11
NDRS-{866|867|868}: Joining node synchronizes with the upgraded chain.

goral09:
## Description
While executing linear chain blocks (from Genesis up until fully synchronized), node will check on every switch block if it matches the scheduled upgrade. If it does, then it will persist its current state and shut down. After the restart, it will read its last state and continue where it left off. The process repeats until fully synchronized. Once the node is synchronized it will clean the snapshot state of the syncing component.

## JIRA tickets:
https://casperlabs.atlassian.net/browse/NDRS-866
https://casperlabs.atlassian.net/browse/NDRS-867
https://casperlabs.atlassian.net/browse/NDRS-868

## How to test:
1. Install `nctl`.
2. Build packages: `nctl-compile`.
3. Start the network: `nctl-start` (starts 5 nodes by default).
4. Wait for a couple of eras: `nctl-view-chain-era`
5. In the root directory of idle node (when using defaults, `node-6` is idle), create a new directory `2_0_0`.
6. Copy the chainspec to it (`cp config/chainspec.toml 2_0_0`) and modify it to reflect a scheduled upgrade: `protocol_version` needs to be set to the same value as the folder (`2_0_0` in this case) and `protocol.activation_point.era_id` to some non-genesis era that is lower than the current era of the network.
7. Take a note of trusted hash (`nctl-view-chain-lfb node=1`) and put it in the node config of the new node (`config/node-config.toml` under `node.trusted_hash`). 
8. Start the new node: `nctl-start node=6`.
9. Observe that the new node synchronizes until the end of an era that is 1 lower than the value set for `protocol.activation_point.era_id` in the `2_0_0` chainspec. `node-6` should shut down.

Since we also want to test that the node resumes after restart, we need to simulate that the upgrade happened and we're ready to continue. To do that, we will remove the `2_0_0` directory. `node-6` should pick up where it left and finish synchronization.

10. Set `node.trusted_hash`, in the genesis chainspec of `node-6`, to a trusted hash (can be a new one).
11. Start the node: `nctl-start node=6`.
12. Check that the node reuses previous state: `cat <node-6 log file> | grep "reusing"` should log the pre-shutdown state.
13. Observe that the node finishes synchronization: `tail -f <node-6 log file> | grep "finished joining"`.
14. Check that all nodes are at the same LFB: `nctl-view-chain-lfb`.

### Notes
I also tested that the state is cleared once the node is synchronized. 

To reproduce:

Once the node is synchronized, simply shut it down and restart with a new trusted hash. It won't reuse the previous state and start the joining process from scratch.


node/src/testing/multi_stage_test_reactor.rs
```diff
@@ -400,7 +400,9 @@ impl Reactor for MultiStageTestReactor {
 
                     let node_id = joiner_reactor.node_id();
                     effects.extend(joiner_reactor.into_validator_config().boxed().event(
-                        |validator_init_config| {
+                        |res_validator_init_config| {
+                            let validator_init_config = res_validator_init_config
+                                .unwrap_or_else(|err| panic!("failed {:?}", err));
```
afck, line 405: Might as well just `unwrap()`; "failed" doesn't add any information.

node/src/components/storage.rs
```diff
@@ -340,6 +340,38 @@ impl Storage {
...
+    /// If key is non-empty, returns bytes from under the key. Otherwise returns `Ok(None)`.
+    /// May also fail with storage errors.
+    #[cfg(not(feature = "fast-sync"))]
+    pub(crate) fn read_state_store<K>(&self, key: &K) -> Result<Option<Vec<u8>>, Error>
+    where
+        K: AsRef<[u8]>,
+    {
+        let txn = self.env.begin_ro_txn()?;
+        let bytes = match txn.get(self.state_store_db, &key) {
+            Ok(slice) if slice.is_empty() => None,
```
afck, line 353: I think matching slices is supported now?
```suggestion
            Ok(&[]) => None,
```

node/src/components/linear_chain_sync.rs
```diff
@@ -63,23 +68,68 @@ pub(crate) struct LinearChainSync<I> {
...
+            } => format!(
+                "syncing descendants of {:?}, latest block hash={:?}, height={:?}, era={:?}",
+                trusted_hash,
+                latest_block.hash(),
+                latest_block.height(),
+                latest_block.era_id(),
+            ),
+            _ => panic!("unexpected initial state: {:?}", &state),
+        };
+        info!(?state_status, "reusing previous state");
```
afck, line 125: Would it be cleaner to `impl Display for State`?
```suggestion
        info!(state_status = %state, "reusing previous state");
```

node/src/components/linear_chain_sync.rs
```diff
@@ -622,3 +708,45 @@ where
             move || Event::GetBlockHeightResult(block_height, BlockByHeightResult::Absent(cloned)),
         )
 }
+
+/// Returns key in the Global State under which the LinearChainSync's state is stored.
```
afck, line 712: Global state? Do you mean database?

node/src/components/linear_chain_sync.rs
```diff
@@ -622,3 +708,45 @@ where
...
+fn state_key(chainspec: &Chainspec) -> Vec<u8> {
+    chainspec.network_config.name.clone().into()
+}
+
+/// Deserialized vector of bytes into `LinearChainSync::State`.
+/// Panics on deserialization errors.
+fn deserialize_state(serialized_state: &[u8]) -> Option<State> {
+    bincode::deserialize(&serialized_state).unwrap_or_else(|error| {
+        panic!(
+            "could not deserialize state from storage type name err {:?}",
```
afck, line 722:
```suggestion
            "could not deserialize state from storage type name, error: {:?}",
```
Or just:
```suggestion
            "could not deserialize state from storage type name: {:?}",
```

afck (approved):
No blockers from my side!

node/src/components/linear_chain_sync.rs
goral09: Good idea.

node/src/components/linear_chain_sync.rs
goral09: Done in [`6d6256a` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/6d6256afbfb52e3c3bcc4ea817550a3ce25cea36).

node/src/components/linear_chain_sync.rs
goral09: Done in [`6d6256a` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/6d6256afbfb52e3c3bcc4ea817550a3ce25cea36).

node/src/components/linear_chain_sync.rs
goral09: Done in [`6d6256a` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/6d6256afbfb52e3c3bcc4ea817550a3ce25cea36).

node/src/components/storage.rs
goral09: Done in [`6d6256a` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/6d6256afbfb52e3c3bcc4ea817550a3ce25cea36).

node/src/testing/multi_stage_test_reactor.rs
goral09: Done in [`6d6256a` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/6d6256afbfb52e3c3bcc4ea817550a3ce25cea36).

node/src/reactor/joiner.rs
```diff
@@ -485,8 +490,20 @@ impl reactor::Reactor for Reactor {
             .map(|(pk, motes)| (pk, motes.value()))
             .collect();
 
-        let linear_chain_sync =
-            LinearChainSync::new(registry, init_hash, validator_weights.clone())?;
+        let linear_chain_sync = {
+            let maybe_state =
+                linear_chain_sync::read_init_state(&storage, chainspec_loader.chainspec())?;
+            if let Some(state) = maybe_state {
+                LinearChainSync::from_state(registry, chainspec_loader.chainspec(), state)?
```
xcthulhu, line 497: Cleaner to just have `LinearChainSync::new` take the chainspec and storage and do this there.

Fast sync needs storage anyway, it's better if we keep the APIs of the two implementations as close as is reasonably possible.

node/src/components/linear_chain_fast_sync.rs
```diff
@@ -53,6 +58,15 @@ impl<I: Clone + PartialEq + 'static> LinearChainFastSync<I> {
         })
     }
 
+    /// Initialize `LinearChainSync` component from preloaded `State`.
+    pub fn from_state(
+        _registry: &Registry,
+        _chainspec: &Chainspec,
+        _state: State,
+    ) -> Result<Self, prometheus::Error> {
+        panic!("not implemented yet for fast sync")
```
xcthulhu, line 67: Rather than a static method, we can roll this up in to the `new` constructor for `LinearChainSync` and then we don't need this sentinel.

node/src/reactor/joiner.rs
```diff
@@ -909,7 +929,12 @@ impl Reactor {
     /// Deconstructs the reactor into config useful for creating a Validator reactor. Shuts down
     /// the network, closing all incoming and outgoing connections, and frees up the listening
     /// socket.
-    pub async fn into_validator_config(self) -> ValidatorInitConfig {
+    pub async fn into_validator_config(self) -> Result<ValidatorInitConfig, Error> {
+        // Clean the state of the linear_chain_sync before shutting it down.
+        linear_chain_sync::clean_linear_chain_state(
+            &self.storage,
+            self.chainspec_loader.chainspec(),
+        )?;
```
xcthulhu, line 937:
```suggestion
        #[cfg(not(feature = "fast-sync"))]
        linear_chain_sync::clean_linear_chain_state(
            &self.storage,
            self.chainspec_loader.chainspec(),
        )?;
```

node/src/components/linear_chain_fast_sync.rs
```diff
@@ -602,3 +616,21 @@ where
...
+    panic!("not yet implemented for fast sync")
+}
+
+/// Cleans the linear chain state storage.
+/// May fail with storage error.
+pub(crate) fn clean_linear_chain_state(
+    _storage: &Storage,
+    _chainspec: &Chainspec,
+) -> Result<(), storage::Error> {
+    panic!("not yet implemented for fast sync")
```
xcthulhu, line 635: Let's just have the callsite for this be behind `#[cfg(not(feature = "fast-sync"))]` so we can get rid of this stub

node/src/reactor/joiner.rs
goral09: This feels dirty but I will do that :)

node/src/components/linear_chain_fast_sync.rs
goral09: Good idea.

node/src/components/linear_chain_fast_sync.rs
goral09: Done in [`087dddd` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/087dddd78b6b23012eca2d7f765adf874623c248).

node/src/reactor/joiner.rs
goral09: Done in [`087dddd` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/087dddd78b6b23012eca2d7f765adf874623c248).

node/src/components/linear_chain_fast_sync.rs
goral09: Done in [`087dddd` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/087dddd78b6b23012eca2d7f765adf874623c248).

node/src/reactor/joiner.rs
goral09: Done in [`087dddd` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/087dddd78b6b23012eca2d7f765adf874623c248).

goral09: > Fast sync needs storage anyway, it's better if we keep the APIs of the two implementations as close as is reasonably possible.


Actually, I didn't want to change the common API precisely b/c I don't know if fast sync needs the same arguments :+1:

xcthulhu (approved):
LGTM, but we should wait for @Fraser999

node/src/components/linear_chain_sync.rs
```diff
@@ -622,3 +694,45 @@ where
...
+            "could not deserialize state from storage, error {:?}",
+            error
+        )
+    })
+}
+
+/// Reads the `LinearChainSync's` state from storage, if any.
+/// Panics on deserialization errors.
+pub(crate) fn read_init_state(
+    storage: &Storage,
```
Fraser999, line 717: Passing a ref to one component into another component's function is very much against the spirit of the reactor pattern we're using.  I'm OK with this as a stopgap measure, but we should see this as a high priority item to be refactored.

node/src/components/storage.rs
```diff
@@ -340,6 +340,38 @@ impl Storage {
...
+    /// If key is non-empty, returns bytes from under the key. Otherwise returns `Ok(None)`.
+    /// May also fail with storage errors.
+    #[cfg(not(feature = "fast-sync"))]
+    pub(crate) fn read_state_store<K>(&self, key: &K) -> Result<Option<Vec<u8>>, Error>
+    where
+        K: AsRef<[u8]>,
+    {
+        let txn = self.env.begin_ro_txn()?;
+        let bytes = match txn.get(self.state_store_db, &key) {
+            Ok(&[]) => None,
```
Fraser999, line 353: It seems like `deserialize_state()` in linear_chain_sync should be handling this case.  How would we even get an empty slice stored there?

Fraser999 (commented):
The only point I'm concerned about is the `state_key()` issue.  Passing the storage ref into the other components' functions is definitely something we wanted to avoid, but it shouldn't cause anything to break I believe.

node/src/components/linear_chain_sync.rs
```diff
@@ -622,3 +694,45 @@ where
             move || Event::GetBlockHeightResult(block_height, BlockByHeightResult::Absent(cloned)),
         )
 }
+
+/// Returns key in the database, under which the LinearChainSync's state is stored.
+fn state_key(chainspec: &Chainspec) -> Vec<u8> {
+    chainspec.network_config.name.clone().into()
```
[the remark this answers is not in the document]
goral09: Good point. Will do.

node/src/components/linear_chain_sync.rs
goral09: I thought about this but this isn't a reactor yet, it's still an initialization/setup phase and we're passing reference, not moving the value.

node/src/components/storage.rs
goral09: Maybe I can remove it now. My idea was that if we store an empty vector then we would be effectively removing it from the DB. I never checked that there's a `del` function available. I should use it instead.

EdHastingsCasperLabs (approved):

Fraser999 (approved):

node/src/components/storage.rs
goral09: Removed in [`e0688b5` (#904)](https://github.com/CasperLabs/casper-node/pull/904/commits/e0688b51ed7c2aaff51770152796b4578e1e3268).

node/src/components/linear_chain_sync.rs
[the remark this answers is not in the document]
goral09: Fixed in https://github.com/CasperLabs/casper-node/pull/904/commits/e0688b51ed7c2aaff51770152796b4578e1e3268.

node/src/components/linear_chain_sync.rs
goral09: Does this sound good to you? cc @marc-casperlabs

goral09:
bors merge

PR #906, opened 2021-02-11
NDRS-851: Enforce deploy's runtime args len.

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/NDRS-851

This PR adds two configuration switches to a chainspec

```diff
+ [deploys]
+ payment_args_max_length = 1024
+ session_args_max_length = 1024
```

which makes deploy acceptor verify that both payment runtime args and session runtime args are not larger than the configured values.

### DIFF
--- a/node/src/types/chainspec/deploy_config.rs
+++ b/node/src/types/chainspec/deploy_config.rs
@@ -29,6 +29,8 @@ pub struct DeployConfig {
     pub(crate) block_max_deploy_count: u32,
     pub(crate) block_max_transfer_count: u32,
     pub(crate) block_gas_limit: u64,
+    pub(crate) payment_args_max_length: u32,
+    pub(crate) session_args_max_length: u32,
 }
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


node/src/types/chainspec/deploy_config.rs
```diff
@@ -135,6 +151,7 @@ mod tests {
         let mut rng = crate::new_rng();
         let config = DeployConfig::random(&mut rng);
         let encoded = toml::to_string_pretty(&config).unwrap();
+        eprintln!("{}", encoded);
```
EdHastingsCasperLabs, line 154: why do we want to eprint this?

node/src/types/deploy.rs
```diff
@@ -632,6 +650,32 @@ impl Deploy {
                 got: header.ttl(),
             });
         }
+        
+        let payment_args_length = self.payment().args().serialized_length();
+        if payment_args_length > config.payment_args_max_length as usize {
+            warn!(
```
EdHastingsCasperLabs, line 656: You should not log a warning for this (or below), it is a user error. Info or debug would be acceptable.

EdHastingsCasperLabs (approved):

PR #907, opened 2021-02-11
NCTL :: view node peer count command

ghost:
New  command  `nctl-view-node-peer-count`.


linclelinkpart5 (approved):
Awesome, the peer count command seems very useful!

PR #908, opened 2021-02-11
NO_TICKET small_network blocklist tweak

dwerner:
Add decay to blocklist in small_network, along with an exclusion for anything already in `known_addresses`.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -84,18 +84,26 @@ use tracing::{debug, error, info, trace, warn};
 
 use self::error::Result;
 pub(crate) use self::{event::Event, gossiped_address::GossipedAddress, message::Message};
-use crate::{NodeRng, components::{
+use crate::{
+    components::{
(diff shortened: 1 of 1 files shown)


EdHastingsCasperLabs (approved):

Fraser999 (approved):

node/src/components/small_network.rs
```diff
@@ -704,8 +714,10 @@ where
     }
 
     fn connect_to_peer_if_required(&mut self, peer_address: SocketAddr) -> Effects<Event<P>> {
+        self.blocklist
+            .retain(|_, ts| *ts > Timestamp::now() + *BLOCKLIST_RETAIN_DURATION);
```
goral09, line 718: Won't the right-hand side will always be bigger? `ts` is the time when the entry was added, which has to be in the past, while the right side is `now() + RETAIN_DURATION`.  I guess we should remove `BLOCKLIST_RETAIN_DURATION` from `now()`, rather than add it.

dwerner: Thanks, @goral09, fixing this here https://github.com/CasperLabs/casper-node/pull/910

PR #912, opened 2021-02-12
NO-TICKET: Don't evict validators if era was too short to produce a unit.

afck:
In practice, with reasonably long eras, this won't make any difference, but in some tests with single-block eras we observed that everyone got evicted because they did not have a chance to create a unit before the last (and only) block.

This PR makes Highway only report a validator as inactive if there was enough time before the first and last block of an era for them to create a unit.

### DIFF
--- a/node/src/components/consensus/highway_core/finality_detector.rs
+++ b/node/src/components/consensus/highway_core/finality_detector.rs
@@ -11,7 +11,7 @@ use crate::{
         consensus_protocol::{FinalizedBlock, TerminalBlockData},
         highway_core::{
             highway::Highway,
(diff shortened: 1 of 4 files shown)


afck:
(This is an alternative to #911.)

fizyk20 (approved):
Looks good, but should we also reduce the era length in the tests again? (I think it is increased to 4 now.)

EdHastingsCasperLabs (approved):

afck:
(I'm getting the test failure locally, too, but only after several attempts. Trying the same test on `master` now…)

PR #914, opened 2021-02-12
NDRS-883: nctl memory metrics

marc-casperlabs:
Makes it convenient and easy to run Prometheus with an nctl network, getting access to all the metrics, as well as local process memory usage.


momipsl (approved):
Looks good - thx for taking initiative.  Perhaps it can be folded into nctl directly in. due course.

PR #918, opened 2021-02-14
NDRS-840: preparation to support upgrades

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-840

This PR paves the way for node upgrades.  It contains the following high-level changes:

* The chainspec is now loaded on startup and a single version passed to all components requiring it.  It is no longer held in storage.
* The chainspec loader retrieves the highest block from storage and makes a decision on how to proceed based on the block's era ID, whether it's a switch block, the chainspec's protocol version, and the next upgrade's activation point.  The choice may result in the node exiting with a code 0 (an indication to the launcher to run the next installed version), a code 102 (the launcher should run the previous installed version), or a code 101 (the launcher and node should both exit).  Alternatively, the chainspec loader can indicate that the process should continue to run using the next reactor.
* The chainspec loader now calls the contract runtime methods `commit_genesis()` or `upgrade()` or neither as appropriate (it currently always calls `commit_genesis()`).
* The JSON-RPC server and event-stream server now use the current protocol version as the API version provided in responses/events to clients.
* If the trusted hash is not provided in the config, the hash of the highest block will be used instead.

Some details worth noting:
* The `unbonding_delay` throughout was defined as an `EraId` which was inappropriate, as it represents a count of eras rather than a specific one.  This largely went unnoticed as `EraId` as defined in `casper_types::auction::types` is just an alias.  These aliases would be much more useful if they were newtypes.  In the meantime, this PR replaces usages of `EraId` with `u64` for `unbonding_delay`.
* The JSON-RPC response to `state_get_auction_info` now follows the same format as other response data, and includes the API version along with the auction info, rather than just the bare auction info (@zie1ony, @momipsl please take note)

This code cannot be tested against current delta nodes as the format of blocks has changed since then, and handling that requires updates to the linear chain sync.  Testing it by trying to upgrade against itself with only a change to the chainspec version and activation point shows that we need to resolve the issue whereby a restarted network panics with `cannot start era with total weight 0`.

### DIFF
--- a/execution_engine/src/core/engine_state/genesis.rs
+++ b/execution_engine/src/core/engine_state/genesis.rs
@@ -13,7 +13,7 @@ use serde::{Deserialize, Serialize};
 use casper_types::{
     account::AccountHash,
     auction::{
-        Bid, Bids, DelegationRate, EraId, SeigniorageRecipient, SeigniorageRecipients,
+        Bid, Bids, DelegationRate, SeigniorageRecipient, SeigniorageRecipients,
         SeigniorageRecipientsSnapshot, UnbondingPurses, ValidatorWeights, ARG_DELEGATION_RATE,
         ARG_DELEGATOR, ARG_ERA_END_TIMESTAMP_MILLIS, ARG_PUBLIC_KEY, ARG_REWARD_FACTORS,
         ARG_VALIDATOR, ARG_VALIDATOR_PUBLIC_KEY, AUCTION_DELAY_KEY, BIDS_KEY,
@@ -311,7 +311,7 @@ pub struct ExecConfig {
     auction_delay: u64,
     locked_funds_period_millis: u64,
     round_seigniorage_rate: Ratio<u64>,
-    unbonding_delay: EraId,
+    unbonding_delay: u64,
     genesis_timestamp_millis: u64,
 }
 
@@ -325,7 +325,7 @@ impl ExecConfig {
         auction_delay: u64,
         locked_funds_period_millis: u64,
         round_seigniorage_rate: Ratio<u64>,
-        unbonding_delay: EraId,
+        unbonding_delay: u64,
         genesis_timestamp_millis: u64,
     ) -> ExecConfig {
         ExecConfig {
@@ -379,7 +379,7 @@ impl ExecConfig {
         self.round_seigniorage_rate
     }
 
-    pub fn unbonding_delay(&self) -> EraId {
+    pub fn unbonding_delay(&self) -> u64 {
         self.unbonding_delay
     }
 
--- a/execution_engine/src/core/engine_state/upgrade.rs
+++ b/execution_engine/src/core/engine_state/upgrade.rs
@@ -85,9 +85,9 @@ pub struct UpgradeConfig {
     activation_point: Option<ActivationPoint>,
     new_validator_slots: Option<u32>,
     new_auction_delay: Option<u64>,
-    new_locked_funds_period: Option<EraId>,
+    new_locked_funds_period_millis: Option<u64>,
     new_round_seigniorage_rate: Option<Ratio<u64>>,
-    new_unbonding_delay: Option<EraId>,
+    new_unbonding_delay: Option<u64>,
     global_state_update: BTreeMap<Key, StoredValue>,
 }
 
@@ -102,9 +102,9 @@ impl UpgradeConfig {
         activation_point: Option<ActivationPoint>,
         new_validator_slots: Option<u32>,
         new_auction_delay: Option<u64>,
-        new_locked_funds_period: Option<EraId>,
+        new_locked_funds_period_millis: Option<EraId>,
         new_round_seigniorage_rate: Option<Ratio<u64>>,
-        new_unbonding_delay: Option<EraId>,
+        new_unbonding_delay: Option<u64>,
         global_state_update: BTreeMap<Key, StoredValue>,
     ) -> Self {
         UpgradeConfig {
@@ -116,7 +116,7 @@ impl UpgradeConfig {
             activation_point,
             new_validator_slots,
             new_auction_delay,
-            new_locked_funds_period,
+            new_locked_funds_period_millis,
             new_round_seigniorage_rate,
             new_unbonding_delay,
             global_state_update,
@@ -155,15 +155,15 @@ impl UpgradeConfig {
         self.new_auction_delay
     }
 
-    pub fn new_locked_funds_period(&self) -> Option<EraId> {
-        self.new_locked_funds_period
+    pub fn new_locked_funds_period_millis(&self) -> Option<u64> {
+        self.new_locked_funds_period_millis
     }
 
     pub fn new_round_seigniorage_rate(&self) -> Option<Ratio<u64>> {
         self.new_round_seigniorage_rate
     }
 
-    pub fn new_unbonding_delay(&self) -> Option<EraId> {
+    pub fn new_unbonding_delay(&self) -> Option<u64> {
         self.new_unbonding_delay
     }
 
--- a/execution_engine/src/core/engine_state/upgrade.rs
+++ b/execution_engine/src/core/engine_state/upgrade.rs
@@ -4,7 +4,6 @@ use num_rational::Ratio;
 use thiserror::Error;
 
 use casper_types::{
-    auction::EraId,
     bytesrepr,
     system_contract_type::{AUCTION, MINT, PROOF_OF_STAKE, STANDARD_PAYMENT},
     ContractHash, Key, ProtocolVersion,
@@ -102,7 +101,7 @@ impl UpgradeConfig {
         activation_point: Option<ActivationPoint>,
         new_validator_slots: Option<u32>,
         new_auction_delay: Option<u64>,
-        new_locked_funds_period_millis: Option<EraId>,
+        new_locked_funds_period_millis: Option<u64>,
         new_round_seigniorage_rate: Option<Ratio<u64>>,
         new_unbonding_delay: Option<u64>,
         global_state_update: BTreeMap<Key, StoredValue>,
--- a/node/src/components/block_executor.rs
+++ b/node/src/components/block_executor.rs
@@ -82,7 +82,7 @@ type BlockHeight = u64;
 /// The Block executor component.
 #[derive(DataSize, Debug, Default)]
 pub(crate) struct BlockExecutor {
-    genesis_state_root_hash: Digest,
+    genesis_state_root_hash: Option<Digest>,
     /// A mapping from proto block to executed block's ID and post-state hash, to allow
     /// identification of a parent block's details once a finalized block has been executed.
     ///
@@ -98,7 +98,7 @@ pub(crate) struct BlockExecutor {
 }
 
 impl BlockExecutor {
-    pub(crate) fn new(genesis_state_root_hash: Digest, registry: Registry) -> Self {
+    pub(crate) fn new(genesis_state_root_hash: Option<Digest>, registry: Registry) -> Self {
         let metrics = BlockExecutorMetrics::new(registry).unwrap();
         BlockExecutor {
             genesis_state_root_hash,
@@ -436,7 +436,7 @@ impl BlockExecutor {
 
     fn pre_state_hash(&mut self, finalized_block: &FinalizedBlock) -> Option<Digest> {
         if finalized_block.is_genesis_child() {
-            Some(self.genesis_state_root_hash)
+            self.genesis_state_root_hash
         } else {
             // Try to get the parent's post-state-hash from the `parent_map`.
             // We're subtracting 1 from the height as we want to get _parent's_ post-state hash.
--- a/node/src/components/block_executor.rs
+++ b/node/src/components/block_executor.rs
@@ -11,6 +11,7 @@ use std::{
 use datasize::DataSize;
 use itertools::Itertools;
 use prometheus::Registry;
+use semver::Version;
 use smallvec::SmallVec;
 use tracing::{debug, error, trace};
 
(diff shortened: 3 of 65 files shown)


goral09:
This [commit](https://github.com/CasperLabs/casper-node/pull/918/commits/ea465e14c431267ec7ac58836be859e7953550f2) won't work unless https://github.com/CasperLabs/casper-node/pull/864 is merged which in turn awaits this task to be completed https://casperlabs.atlassian.net/browse/EE-1173

goral09:
Thanks for the clean commit history :)

mpapierski (approved):

goral09:
> The JSON-RPC server and event-stream server now use the current protocol version as the API version provided in responses/events to clients.

I think those are two different things: API version and protocol version.

node/src/components/deploy_acceptor.rs
```diff
@@ -57,36 +55,21 @@ impl<REv> ReactorEventT for REv where
...
-
 /// The `DeployAcceptor` is the component which handles all new `Deploy`s immediately after they're
 /// received by this node, regardless of whether they were provided by a peer or a client.
 ///
 /// It validates a new `Deploy` as far as possible, stores it if valid, then announces the newly-
 /// accepted `Deploy`.
 #[derive(Debug)]
 pub struct DeployAcceptor {
-    cached_deploy_configs: HashMap<Version, DeployAcceptorChainspec>,
+    chainspec: Arc<Chainspec>,
```
goral09, line 65: I see what you're trying to do here but the point of having separate structs was not only to save memory but to pass on only what the component needs to know, not everything.

node/src/components/event_stream_server/http_server.rs
```diff
@@ -68,7 +70,7 @@ pub(super) async fn run(
                     if let Some(subscriber) = maybe_new_subscriber {
                         // First send the client the `ApiVersion` event.  We don't care if this
                         // errors - the client may have disconnected already.
-                        let _ = subscriber.initial_events_sender.send(SSE_INITIAL_EVENT.clone());
+                        let _ = subscriber.initial_events_sender.send(ServerSentEvent::initial_event(api_version.clone()));
```
afck, line 73: Line too long

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -102,9 +102,9 @@ impl UpgradeConfig {
         activation_point: Option<ActivationPoint>,
         new_validator_slots: Option<u32>,
         new_auction_delay: Option<u64>,
-        new_locked_funds_period: Option<EraId>,
+        new_locked_funds_period_millis: Option<EraId>,
```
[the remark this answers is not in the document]
Fraser999: Well spotted :)

Fraser999:
> I think those are two different things: API version and protocol version.

Ed and I had a discussion on this, and concluded that it's not worthwhile trying to separate the two concepts, at least for now.

goral09:
> > I think those are two different things: API version and protocol version.
> 
> Ed and I had a discussion on this, and concluded that it's not worthwhile trying to separate the two concepts, at least for now.

What I am mostly concerned about is when we end up with client <-> node versions that should be compatible but aren't.

node/src/components/deploy_acceptor.rs
Fraser999: Fair point.  Would you be ok if I just add `chain_name` and `deploy_config` as members of `DeployAcceptor` rather than creating a whole new struct for them?

Fraser999:
> What I am mostly concerned about is when we end up with client <-> node versions that should be compatible but aren't.

Yup, but I think it's better for now to run the risk of being overly severe (saying 2 client API versions are incompatible when they really could be compatible), rather than being too lax (saying 2 client API versions are compatible when they actually aren't).

Until we get solid testing in place for client API compatibility I don't think we should risk separating the client API from the protocol version.

node/src/components/block_executor.rs
```diff
@@ -98,10 +101,19 @@ pub(crate) struct BlockExecutor {
...
-    pub(crate) fn new(genesis_state_root_hash: Digest, registry: Registry) -> Self {
+    pub(crate) fn new(
+        genesis_state_root_hash: Option<Digest>,
+        protocol_version: Version,
+        registry: Registry,
+    ) -> Self {
         let metrics = BlockExecutorMetrics::new(registry).unwrap();
         BlockExecutor {
             genesis_state_root_hash,
+            protocol_version: ProtocolVersion::from_parts(
```
afck, line 112: …and maybe `impl From<Version> for ProtocolVersion`.

node/src/types/status_feed.rs
```diff
@@ -35,9 +45,39 @@ static GET_STATUS_RESULT: Lazy<GetStatusResult> = Lazy::new(|| {
...
-    GetStatusResult::from(status_feed)
+    GetStatusResult::new(status_feed, DOCS_EXAMPLE_PROTOCOL_VERSION.clone())
 });
 
+/// Summary information from the chainspec.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ChainspecInfo {
+    /// Name of the network.
+    name: String,
+    /// If `Some` then genesis or upgrade process returned a valid post state hash.
```
afck, line 56: What's the plan for emergency restarts: Will this also be `Some` then?

And for upgrades: I imagined we'd look into the latest block, and use its post-state hash, instead of the one found in the chainspec — but that works only for upgrades that don't migrate the state format; is the plan to always (automatically?) put the root hash into the chainspec somehow?

node/src/components/chainspec_loader.rs
```diff
@@ -231,42 +216,371 @@ impl ChainspecLoader {
...
+            }
+        };
+        let highest_block_era_id = highest_block.header().era_id().0;
+
+        let cached_protocol_version = maybe_cached_protocol_version
+            .expect("cached protocol version must exist if we have a stored block");
+        let current_chainspec_activation_point =
+            self.chainspec.protocol_config.activation_point.era_id.0;
+
+        if highest_block_era_id + 1 == current_chainspec_activation_point {
```
afck, line 350: You could keep them as `EraId` type:
```suggestion
        if highest_block_era_id.successor() == current_chainspec_activation_point {
```

node/src/components/chainspec_loader.rs
```diff
@@ -231,42 +216,371 @@ impl ChainspecLoader {
...
+                info!(
+                    new_point=%next_upgrade.activation_point,
+                    %current_point,
+                    "changing upgrade activation point"
+                );
+            }
+        }
+        self.next_upgrade = Some(next_upgrade);
+        Effects::new()
+    }
```
afck, line 576: I read the whole PR, so I should know it at this point, but: How does the new activation point get communicated to all the components? (Probably most of them don't need to know?)

afck (commented):
Looks great! Just a few nit-picks and questions.

node/src/components/chainspec_loader.rs
goral09: This event is only handled by components that need to decide when to upgrade (consensus). All the components that use the chainspec will get initialized with a correct one upon startup.

node/src/types/status_feed.rs
```diff
@@ -35,9 +45,39 @@ static GET_STATUS_RESULT: Lazy<GetStatusResult> = Lazy::new(|| {
...
+
+impl ChainspecInfo {
+    pub(crate) fn new(
+        chainspec_network_name: String,
+        state_root_hash: Option<Digest>,
+        next_upgrade: Option<NextUpgrade>,
+    ) -> Self {
+        ChainspecInfo {
+            name: chainspec_network_name,
+            state_root_hash,
```
goral09, line 75: This comes from the `ChainspecLoader`, if there was an upgrade will that be a hash of the state root after applying the GlobalState changes or before them?

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -13,7 +13,7 @@ use serde::{Deserialize, Serialize};
 use casper_types::{
     account::AccountHash,
     auction::{
-        Bid, Bids, DelegationRate, EraId, SeigniorageRecipient, SeigniorageRecipients,
```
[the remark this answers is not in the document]
goral09: https://casperlabs.atlassian.net/browse/HWY-258

node/src/components/block_proposer.rs
```diff
@@ -83,39 +89,29 @@ impl BlockProposer {
...
-            chainspec,
+        .event(move |sets| Event::Loaded {
             sets,
             next_finalized_block,
         });
 
         let block_proposer = BlockProposer {
             state: BlockProposerState::Initializing {
                 pending: Vec::new(),
+                state_key: deploy_sets::create_storage_key(chainspec),
```
goral09, line 113: Is that the same as `state_key` in line 98?

goral09 (approved):
Looks good :) I don't have any blockers but I am still a bit scared of approving it. I will give myself time to test it tomorrow with various scenarios and, if that works, I will approve it.

henrytill (approved):
:+1: from an EE perspective.

EdHastingsCasperLabs (approved):

node/src/components/rpc_server/rpcs.rs
```diff
@@ -231,7 +250,13 @@ pub(super) trait RpcWithOptionalParamsExt: RpcWithOptionalParams {
             .and(filters::json_rpc())
             .and(filters::method(Self::METHOD))
             .and_then(move |response_builder: Builder| {
-                Self::handle_request(effect_builder, response_builder, None).map_err(reject::custom)
+                Self::handle_request(
+                    effect_builder,
+                    response_builder,
+                    None,
+                    api_version_cloned.clone(),
```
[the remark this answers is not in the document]
Fraser999: Unfortunately yes - the closure needs to implement `Fn` and if we don't clone, it only implements `FnOnce`.

node/src/components/block_executor.rs
Fraser999: That's a problem since I believe `Version` isn't no-std compatible, and `ProtocolVersion` is in the casper-types crate which needs to support no-std.

node/src/types/status_feed.rs
Fraser999: > What's the plan for emergency restarts: Will this also be Some then?

Not the way the code is currently arranged in chainspec loader.

> And for upgrades: I imagined we'd look into the latest block, and use its post-state hash, instead of the one found in the chainspec — but that works only for upgrades that don't migrate the state format; is the plan to always (automatically?) put the root hash into the chainspec somehow?

There's no state hash in the chainspec, nor any plans to add one afaik.  This struct may be misnamed, since it does imply that the chainspec has one.  The root hash here is only held by the chainspec loader component in memory, and is the state hash returned by EE after performing the `commit_genesis()` or `upgrade()` work.

I think you're right in that it should never be `None` - we should be using the one from the highest block if we're not running `commit_genesis()` or `upgrade()`.  I'll update to handle that.

node/src/components/chainspec_loader.rs
Fraser999: Nice - thanks!

node/src/types/status_feed.rs
Fraser999: That's after the upgrade.

node/src/components/block_proposer.rs
Fraser999: Yes - I'll clone it above to make this more obvious.

node/src/components/chainspec_loader.rs
```diff
@@ -340,14 +343,14 @@ impl ChainspecLoader {
...
+        let highest_block_era_id = highest_block.header().era_id();
 
         let cached_protocol_version = maybe_cached_protocol_version
             .expect("cached protocol version must exist if we have a stored block");
         let current_chainspec_activation_point =
-            self.chainspec.protocol_config.activation_point.era_id.0;
+            self.chainspec.protocol_config.activation_point.era_id;
 
-        if highest_block_era_id + 1 == current_chainspec_activation_point {
+        if highest_block_era_id.successor() == current_chainspec_activation_point {
```
EdHastingsCasperLabs, line 353: fancy

Fraser999:
bors merge

node/src/components/block_executor.rs
afck: The impl could be `#[cfg(feature = "std")]`, could it?

node/src/components/chainspec_loader.rs
```diff
@@ -258,7 +258,7 @@ impl ChainspecLoader {
             chainspec,
             root_dir,
             reactor_exit,
-            state_root_hash: None,
+            starting_state_root_hash: Digest::default(),
```
goral09, line 261: Why would we ever reuse cached/stored state? Shouldn't we always use the one provided on the node startup?

Also, the `Digest::default()` is supposed to be a placeholder that will be replaced ones we handle the `Initialize` event?

Why not make this constructor `async` and do that in the `initializer.rs` constructor, even before we have a reactor.

Fraser999: > Why would we ever reuse cached/stored state? Shouldn't we always use the one provided on the node startup?

This would be for restarting where we didn't stop for an upgrade.

> Also, the Digest::default() is supposed to be a placeholder that will be replaced ones we handle the Initialize event?

Yes.

> Why not make this constructor async and do that in the initializer.rs constructor, even before we have a reactor.

I don't think there's a need.  We don't have any other components' constructors async.  We have a design pattern of keeping the components methods synchronous and I'd rather keep to that.

goral09: > I don't think there's a need. We don't have any other components' constructors async. We have a design pattern of keeping the components methods synchronous and I'd rather keep to that.

Sure, but the reactor's constructor is already async and I think it's worth making the component's constructor async to simplify the setup. Those `Initialize` events introduce a lot of (unnecessary) complexity. Just FYI, I will be using async constructors if they will simplify the code :)

Fraser999:
> the reactor's constructor is already async

It isn't - maybe you're thinking of the `Runner`?  We _really_ don't want to be constructing components in the Runner's constructor if that's the plan?

> Just FYI, I will be using async constructors if they will simplify the code :)

Please, please don't do that!  We agreed the design months ago, and while I acknowledge it can be tedious to work with, it's still a safer option than dealing with async code throughout our components.  There are a few "cheats" technically open to us with our pattern, but we have managed to avoid most or all of these so far.

PR #919, opened 2021-02-15
EE-1171: Automatically undelegate delegators when a validator fully unbonds

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1171

This PR introduces an automatic undelegation of delegator funds when a validator fully unbonds their stakes.


henrytill (approved):
:+1: LGTM :+1:

PR #920, opened 2021-02-15
NO-TICKET: Speed up sync/join integration test.

goral09:
This PR decreases the batch sizes of Wasm deploys and native transfers used in the `sync_test.sh`.

The main goal of the `sync_test.sh` is correctness, not performance, so the number of deploys in the network does not matter as long as that number is positive and we have a different type of deploys (native vs Wasm) both before and after the trusted hash.


momipsl (approved):
Looks good.

goral09:
bors merge

PR #924, opened 2021-02-16
NDRS-859: use next semver version rather than max when looking for upgrades

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-859

This PR changes the behaviour of the chainspec loader to look for the immediate next version of casper-node as per semver ordering as opposed to the maximum version.

This brings it in line with the launcher's behaviour, and allows for multiple future versions to be installed concurrently, letting the launcher walk up through the versions sequentially.


EdHastingsCasperLabs (approved):

Fraser999:
bors merge

PR #925, opened 2021-02-16
NDRS-889: Add `Fetcher` metrics.

goral09:
Adds three metrics do the fetcher:
* found_in_storage
* found_on_peer
* timeouts

Metrics are prefixed with the "type" of the fetcher:
* linear_chain
* deploy
* block_by_height.

### DIFF
--- a/node/src/components/fetcher/metrics.rs
+++ b/node/src/components/fetcher/metrics.rs
@@ -1,0 +1,57 @@
+use prometheus::{IntCounter, Registry};
+
+#[derive(Debug)]
+pub struct FetcherMetrics {
+    /// Number of fetch requests that found an item in the storage.
+    pub(super) found_in_storage: IntCounter,
+    /// Number of fetch requests that fetcher an item from peer.
+    pub(super) found_on_peer: IntCounter,
+    /// Number of fetch requests that timed out.
+    pub(super) timeouts: IntCounter,
+    /// Reference to the registry for unregistering.
+    registry: Registry,
+}
+
+impl FetcherMetrics {
+    pub fn new(name: &str, registry: &Registry) -> Result<Self, prometheus::Error> {
+        let found_in_storage = IntCounter::new(
+            format!("{}_found_in_storage", name),
+            format!(
+                "number of fetch requests that found {} in the storage.",
+                name
+            ),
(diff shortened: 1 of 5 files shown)


node/src/components/fetcher/metrics.rs
```diff
@@ -0,0 +1,57 @@
+use prometheus::{IntCounter, Registry};
+
+#[derive(Debug)]
+pub struct FetcherMetrics {
+    /// Number of fetch requests that found an item in the storage.
+    pub(super) found_in_storage: IntCounter,
+    /// Number of fetch requests that fetcher an item from peer.
```
Fraser999, line 7:
```suggestion
    /// Number of fetch requests that fetched an item from peer.
```

node/src/components/fetcher/metrics.rs
```diff
@@ -0,0 +1,57 @@
...
+            registry: registry.clone(),
+        })
+    }
+}
+
+impl Drop for FetcherMetrics {
+    fn drop(&mut self) {
+        self.registry
+            .unregister(Box::new(self.found_in_storage.clone()))
+            .expect("did not expect deregistering found_in_storage to fail");
```
Fraser999, line 49: I know a lot of the other components are doing it this way, but could we use `unwrap_or_else` with a warning printed rather than a panic?  Similar to how e.g. linear chain handles this.

Fraser999 (commented):
I'm not seeing where the metrics are actually updated?

node/src/components/fetcher/metrics.rs
goral09: Done in [`9960350` (#925)](https://github.com/CasperLabs/casper-node/pull/925/commits/99603501fb81193ac6820dc9eac965ab73cb98b8).

node/src/components/fetcher/metrics.rs
goral09: Done in [`9960350` (#925)](https://github.com/CasperLabs/casper-node/pull/925/commits/99603501fb81193ac6820dc9eac965ab73cb98b8).

Fraser999 (approved):

goral09:
bors merge

PR #926, opened 2021-02-16
NDRS-898: Add `ProtocolVersion` to the block header.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-898

NOTE: we still populate the header with the hardcoded version `1.0.0`. Once #918 is merged, I will create a follow-up PR that uses `BlockExecutor`'s protocol version when creating new blocks.

### DIFF
--- a/types/src/protocol_version.rs
+++ b/types/src/protocol_version.rs
@@ -1,6 +1,8 @@
 use alloc::vec::Vec;
 use core::fmt;
 
+use datasize::DataSize;
+use schemars::{gen::SchemaGenerator, schema::Schema, JsonSchema};
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -10,7 +12,18 @@ use crate::{
 
 /// A newtype wrapping a [`SemVer`] which represents a Casper Platform protocol version.
 #[derive(
-    Copy, Clone, Debug, Default, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize,
+    Copy,
+    Clone,
+    DataSize,
+    Debug,
+    Default,
+    Hash,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Serialize,
+    Deserialize,
 )]
 pub struct ProtocolVersion(SemVer);
 
@@ -146,6 +159,19 @@ impl FromBytes for ProtocolVersion {
 impl fmt::Display for ProtocolVersion {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         self.0.fmt(f)
+    }
+}
+
+impl JsonSchema for ProtocolVersion {
+    fn schema_name() -> String {
+        String::from("ProtocolVersion")
+    }
+
+    fn json_schema(gen: &mut SchemaGenerator) -> Schema {
+        let schema = gen.subschema_for::<String>();
+        let mut schema_object = schema.into_object();
+        schema_object.metadata().description = Some("Protocol version of the network".to_string());
+        schema_object.into()
     }
 }
 
--- a/types/src/protocol_version.rs
+++ b/types/src/protocol_version.rs
@@ -2,6 +2,7 @@ use alloc::vec::Vec;
 use core::fmt;
 
 use datasize::DataSize;
+#[cfg(feature = "std")]
 use schemars::{gen::SchemaGenerator, schema::Schema, JsonSchema};
 use serde::{Deserialize, Serialize};
 
@@ -162,6 +163,7 @@ impl fmt::Display for ProtocolVersion {
     }
 }
 
+#[cfg(feature = "std")]
 impl JsonSchema for ProtocolVersion {
     fn schema_name() -> String {
         String::from("ProtocolVersion")
--- a/types/src/protocol_version.rs
+++ b/types/src/protocol_version.rs
@@ -1,31 +1,21 @@
 use alloc::vec::Vec;
 use core::fmt;
+use std::{convert::TryFrom, string::String};
 
 use datasize::DataSize;
+
 #[cfg(feature = "std")]
-use schemars::{gen::SchemaGenerator, schema::Schema, JsonSchema};
-use serde::{Deserialize, Serialize};
+use schemars::JsonSchema;
+use serde::{de::Error as SerdeError, Deserialize, Deserializer, Serialize, Serializer};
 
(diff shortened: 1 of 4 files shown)


afck (approved):

types/src/protocol_version.rs
```diff
@@ -149,6 +163,20 @@ impl fmt::Display for ProtocolVersion {
     }
 }
 
+#[cfg(feature = "std")]
+impl JsonSchema for ProtocolVersion {
```
Fraser999, line 167: I think this is incorrect.  This is specifying the JSON format as a string, but because we just derive `Serialize` here and for `SemVer` it won't be just a string, it'll be a JSON object.

I think we should follow the example of e.g. `Timestamp` where we have [custom impls for `Serialize` and `Deserialize`](https://github.com/CasperLabs/casper-node/blob/master/node/src/types/timestamp.rs#L158-L178) and then we can just derive the `JsonSchema` with [an extra attribute](https://github.com/CasperLabs/casper-node/blob/master/node/src/types/timestamp.rs#L37) to tell it we use string for JSON.  That should be applied to the `SemVer` type, and here we should probably use [`#[serde(flatten)]` ](https://serde.rs/attr-flatten.html) like this:
```
pub struct ProtocolVersion(#[serde(flatten)] SemVer);
```

I _think_ that will give the best results for JSON-encoding these, but we should check how this looks before merging this PR, and we should also eyeball the generated JSON-schema.

Fraser999 (changes requested):
When this merges, we need to notify the frontend guys and Make.

types/src/protocol_version.rs
goral09: A, ok. Thanks for the pointers.

goral09: Did it a bit differently but here's the result:
```json
"header": {
    "accumulated_seed": "4cda5b116a1b66f12b77d106c2120a211c4542fdcdcb49d8560eb7fc43550398",
    "body_hash": "6a34d9ae04ed621e0ef8e460cb24b4dae9ec116ce82cb542ab5a51b6fb9953d3",
    "era_end": null,
    "era_id": 1,
    "height": 11,
    "parent_hash": "a0206acac97fb834ce9ef2f7205cb8e5edad294b83250e920846e712a98f4425",
    "protocol_version": "1.0.0",
    "random_bit": false,
    "state_root_hash": "eafd38f29dce7e17bafee6c08d1e165d8859fc15ccbe07a9f2d762f3a1966ebb",
    "timestamp": "2021-02-16T14:04:00.896Z"
  }
```

goral09: and schema:
```json
"protocol_version": {
            "$ref": "#/components/schemas/ProtocolVersion"
          }
```

types/src/protocol_version.rs
```diff
@@ -143,6 +146,36 @@ impl FromBytes for ProtocolVersion {
     }
 }
 
+impl Serialize for ProtocolVersion {
+    fn serialize<S: Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let str = format!("{}.{}.{}", self.0.major, self.0.minor, self.0.patch);
+        String::serialize(&str, serializer)
```
Fraser999, line 152: I think we shouldn't force the stringified version for non-human-readable encoders:

```suggestion
        if serializer.is_human_readable() {
            self.0.to_string().serialize(serializer)
        } else {
            self.0.serialize(serializer)
        }
```

types/src/protocol_version.rs
```diff
@@ -143,6 +146,36 @@ impl FromBytes for ProtocolVersion {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for ProtocolVersion {
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let value_as_string = String::deserialize(deserializer)?;
+        let parse_res: Result<SemVer, ParseSemVerError> =
+            SemVer::try_from(value_as_string.as_str());
+        parse_res.map_err(SerdeError::custom).map(ProtocolVersion)
```
Fraser999, line 161: Same for deserialize:

```suggestion
        let semver = if deserializer.is_human_readable() {
            let value_as_string = String::deserialize(deserializer)?;
            SemVer::try_from(value_as_string.as_str()).map_err(SerdeError::custom)?
        } else {
            SemVer::deserialize(deserializer)?
        };
        Ok(ProtocolVersion(semver))
```

types/src/protocol_version.rs
```diff
@@ -414,4 +447,12 @@ mod tests {
         let other = ProtocolVersion::from_parts(1, 99, 99);
         assert!(current.is_compatible_with(&other));
     }
+
+    #[test]
+    fn should_serialize_to_json_properly() {
```
Fraser999, line 452: I think either here or as a separate test, we should do the full round trip and deserialize from JSON back to the original.

It'd also be good to do a round trip test using `bincode` to check the non-human-readable case too.

goral09: Do we do that for all types?

goral09: Done in [`dbe8a60` (#926)](https://github.com/CasperLabs/casper-node/pull/926/commits/dbe8a601912d8ffb4ffb7eec8c9e7d191e92ba6e).

types/src/protocol_version.rs
goral09: Done in [`dbe8a60` (#926)](https://github.com/CasperLabs/casper-node/pull/926/commits/dbe8a601912d8ffb4ffb7eec8c9e7d191e92ba6e).

types/src/protocol_version.rs
goral09: Done in [`dbe8a60` (#926)](https://github.com/CasperLabs/casper-node/pull/926/commits/dbe8a601912d8ffb4ffb7eec8c9e7d191e92ba6e).

goral09:
@momipsl @zie1ony Heads-up: we're adding a `protocol_version` field to the `BlockHeader`.

types/src/protocol_version.rs
goral09: I'm assuming this is done and resolving the comment.

Fraser999 (approved):

EdHastingsCasperLabs (approved):

goral09:
bors merge

PR #927, opened 2021-02-16
NDRS-879: Log active/inactive stake.

afck:
This adds a once-per-minute INFO-level log statement listing all faulty and inactive validators, and their total weight as a percentage. Example output:

`validator participation; participation=Participation { instance_id: 248c7d082e44ad352e10dcbdac6a442f2cb675d247d9732c276aefce1dddf5a6, faulty_stake_percent: 1, inactive_stake_percent: 0, validators: [(ValidatorIndex(1), PublicKey::Ed25519(2575b4c621544fa3dda5c06bd26eed6304ea10bc57aa4bef4af9d7483ae95115), Equivocated)] }; instance_id=248c..f5a6
`

https://casperlabs.atlassian.net/browse/NDRS-879

### DIFF
--- a/node/src/components/consensus/protocols/highway/participation.rs
+++ b/node/src/components/consensus/protocols/highway/participation.rs
@@ -12,6 +12,7 @@ use crate::{
         traits::Context,
     },
     types::Timestamp,
+    utils::div_round,
 };
 
 /// A validator's participation status: whether they are faulty or inactive.
@@ -88,8 +89,8 @@ impl<C: Context> Participation<C> {
             .collect();
         Participation {
             instance_id: *highway.instance_id(),
-            inactive_stake_percent: ((inactive_w * 100 + total_w / 2) / total_w) as u8,
-            faulty_stake_percent: ((faulty_w * 100 + total_w / 2) / total_w) as u8,
+            inactive_stake_percent: div_round(inactive_w * 100, total_w) as u8,
+            faulty_stake_percent: div_round(faulty_w * 100, total_w) as u8,
             validators,
         }
     }
--- a/node/src/components/consensus/protocols/highway/participation.rs
+++ b/node/src/components/consensus/protocols/highway/participation.rs
@@ -1,0 +1,96 @@
+use std::cmp::Reverse;
+
+use itertools::Itertools;
+
+use crate::{
+    components::consensus::{
+        highway_core::{
+            highway::Highway,
+            state::{Fault, State},
+            validators::ValidatorIndex,
+        },
+        traits::Context,
+    },
+    types::Timestamp,
+};
+
+/// A validator's participation status: whether they are faulty or inactive.
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
+enum Status {
+    LastSeenSecondsAgo(u64),
+    Inactive,
+    EquivocatedInOtherEra,
+    Equivocated,
+}
+
+impl Status {
+    /// Returns a `Status` for a validator unless they are honest and online.
+    fn for_index<C: Context>(
+        idx: ValidatorIndex,
+        state: &State<C>,
+        now: Timestamp,
+    ) -> Option<Status> {
+        if let Some(fault) = state.maybe_fault(idx) {
+            return Some(match fault {
+                Fault::Banned | Fault::Indirect => Status::EquivocatedInOtherEra,
+                Fault::Direct(_) => Status::Equivocated,
+            });
+        }
+        if state.panorama()[idx].is_none() {
+            return Some(Status::Inactive);
+        }
+        if state.last_seen(idx) + state.params().max_round_length() < now {
+            let seconds = (now - state.last_seen(idx)).millis() / 1000;
+            return Some(Status::LastSeenSecondsAgo(seconds));
+        }
+        None
+    }
+}
+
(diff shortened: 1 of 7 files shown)


afck:
Replaces #900

node/src/components/consensus/protocols/highway/participation.rs
```diff
@@ -0,0 +1,96 @@
...
+                        inactive_w += u128::from(state.weight(idx).0)
+                    }
+                }
+                Some((idx, v_id.clone(), status))
+            })
+            .sorted_by_key(|(idx, _, status)| (Reverse(*status), *idx))
+            .collect();
+        Participation {
+            instance_id: *highway.instance_id(),
+            inactive_stake_percent: ((inactive_w * 100 + total_w / 2) / total_w) as u8,
```
goral09, line 91: Why do we need additional `+ total_w / 2 / total_w ? For rounding errors?

node/src/components/consensus/protocols/highway/participation.rs
```diff
@@ -0,0 +1,96 @@
...
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
+enum Status {
+    LastSeenSecondsAgo(u64),
+    Inactive,
+    EquivocatedInOtherEra,
+    Equivocated,
+}
+
+impl Status {
+    /// Returns a `Status` for a validator unless they are honest and online.
```
goral09, line 27: I think the usage of _unless_ here is wrong, right? It should be "as long as".

node/src/components/consensus/protocols/highway/participation.rs
```diff
@@ -0,0 +1,96 @@
...
+            });
+        }
+        if state.panorama()[idx].is_none() {
+            return Some(Status::Inactive);
+        }
+        if state.last_seen(idx) + state.params().max_round_length() < now {
+            let seconds = (now - state.last_seen(idx)).millis() / 1000;
+            return Some(Status::LastSeenSecondsAgo(seconds));
+        }
+        None
```
goral09, line 46: So we return `None` if validator is honest but we haven't seen new unit in the last `max_round_length` seconds window?

node/src/components/consensus/protocols/highway/participation.rs
```diff
@@ -0,0 +1,96 @@
...
+        let now = Timestamp::now();
+        let state = highway.state();
+        let mut inactive_w = 0;
+        let mut faulty_w = 0;
+        let total_w = u128::from(state.total_weight().0);
+        let validators = highway
+            .validators()
+            .enumerate_ids()
+            .filter_map(|(idx, v_id)| {
+                let status = Status::for_index(idx, state, now)?;
```
goral09, line 76: I think that the decision to use `None` for validators that are honest but haven't produced anything "lately" may be confusing.

node/src/components/consensus/protocols/highway/participation.rs
afck: To round to the nearest integer, instead of always rounding down. I.e. if it's 33.6%, I'd display 34 rather than 33.

node/src/components/consensus/protocols/highway/participation.rs
afck: No, this only returns `Some` for validators that are _not_ honest and online. We don't list the honest and online ones here, so it's easier to see if something's wrong (and who's to blame).

node/src/components/consensus/protocols/highway/participation.rs
goral09: And there's no lib for this? or helper method in std? We keep doing this over and over again manually.

node/src/components/consensus/protocols/highway/participation.rs
goral09: Synonyms for _unless_ are:

> and, barely, besides, but, except, further, however, just, merely, moreover, nevertheless, notwithstanding, notwithstanding, only, provided, save, still, that, tho, yet.

so this comment means that we will return `Status` if they're not honest and online but it's the other way around, isn't it?

node/src/components/consensus/protocols/highway/participation.rs
afck: Yes. It's stricter than the "offline" criterion for ping purposes, but I think it makes sense to log someone once the timestamps indicate they've missed a round.

node/src/components/consensus/protocols/highway/participation.rs
afck: What do you suggest?
I wanted to make a list where each entry reports a validator that isn't operating optimally.

node/src/components/consensus/protocols/highway/participation.rs
afck: No, it's exactly that:
* If they're honest and online, we return `None`.
* If they're _not_ honest and online, we return some `Status`.

node/src/components/consensus/protocols/highway/participation.rs
afck: I can't find anything in the standard library. And I'd find it overkill to add another dependency just for that. How do you feel about a `utils::div_round` helper method instead?

goral09 (approved):

PR #929, opened 2021-02-16
NO-TICKET: fix issue in Key::from_formatted_str

Fraser999:
This PR fixes a bug whereby `Key::from_formatted_str()` doesn't handle parsing a `Key::EraInfo` variant.  It also adds check for that variant to existing tests and adds a couple of serialization round trip tests, one using `bincode` and the other using `serde_json`.

### DIFF
--- a/execution_engine/src/shared/stored_value.rs
+++ b/execution_engine/src/shared/stored_value.rs
@@ -1,13 +1,7 @@
-use std::{
-    convert::TryFrom,
-    fmt::{self, Debug, Formatter},
-};
+use std::{convert::TryFrom, fmt::Debug};
 
-use serde::{
-    de::{self, SeqAccess, Visitor},
-    ser::{Error, SerializeSeq},
-    Deserialize, Deserializer, Serialize, Serializer,
-};
+use serde::{de, ser, Deserialize, Deserializer, Serialize, Serializer};
+use serde_bytes::ByteBuf;
 
 use casper_types::{
     auction::EraInfo,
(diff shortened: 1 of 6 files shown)


mpapierski (approved):

xcthulhu (approved):

execution_engine/src/shared/stored_value.rs
```diff
@@ -309,83 +303,21 @@ impl FromBytes for StoredValue {
...
-        let mut seq = serializer.serialize_seq(Some(number_of_bytes))?;
-
-        // First push the number of bytes the StoredValue takes
-        seq.serialize_element(&number_of_bytes)?;
-        for byte in bytes {
-            seq.serialize_element(&byte)?;
-        }
-        seq.end()
+            .map_err(|error| ser::Error::custom(format!("{:?}", error)))?;
+        ByteBuf::from(bytes).serialize(serializer)
```
xcthulhu, line 312: `serde_bytes` sure does make this simpler

execution_engine/src/shared/stored_value.rs
```diff
@@ -309,83 +303,21 @@ impl FromBytes for StoredValue {
...
-                         Unexpected additional bytes: {:?}",
-                        stored_value, additional_bytes
-                    )));
-                }
-                Ok(stored_value)
-            }
-        }
-        deserializer.deserialize_seq(StoredValueDeserializer)
+    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        let bytes = ByteBuf::deserialize(deserializer)?.into_vec();
```
xcthulhu, line 318: :+1:

execution_engine/src/storage/trie/mod.rs
```diff
@@ -143,23 +136,29 @@ impl<'de> Deserialize<'de> for PointerBlock {
...
-                while let Some(indexed_pointer) = seq.next_element()? {
-                    let (idx, ptr): (u8, Pointer) = indexed_pointer;
-                    ret[idx as usize] = Some(ptr);
+                while let Some((index, pointer_block_value)) =
+                    access.next_entry::<usize, Pointer>()?
+                {
+                    let element = pointer_block.0.get_mut(index).ok_or_else(|| {
+                        de::Error::custom(format!("invalid index {} in pointer block value", index))
+                    })?;
+                    *element = Some(pointer_block_value);
```
xcthulhu, line 155:
```suggestion
                    pointer_block[index as usize] = Some(pointer_block_value);
```

Pointer block radix is 256 so a `u8` is never out of bounds.

execution_engine/src/storage/trie/tests.rs
```diff
@@ -93,5 +93,12 @@ mod proptests {
              let deserialized_trie: Trie<Key, StoredValue> = serde_json::from_str(&json_str).unwrap();
              assert_eq!(trie, deserialized_trie)
         }
+
+        #[test]
+        fn bincode_roundtrip_trie(trie in trie_arb()) {
+           let bincode_bytes = bincode::serialize(&trie)?;
+           let deserialized_trie = bincode::deserialize(&bincode_bytes)?;
+           assert_eq!(trie, deserialized_trie)
+        }
```
xcthulhu, line 102: Here's another test for the `PointerBlock` we should include:

```rust
        #[test]
        fn bincode_roundtrip_trie_pointer_block(pointer_block in trie_pointer_block_arb()) {
             let bincode_bytes = bincode::serialize(&pointer_block)?;
             let deserialized_pointer_block: PointerBlock = bincode::deserialize(&bincode_bytes)?;
             assert_eq!(pointer_block, deserialized_pointer_block)
        }
```

execution_engine/src/storage/trie/mod.rs
xcthulhu: [Here][1] is an older example where we've done this.

[1]: https://github.com/xcthulhu/casper-node/blob/ae40fa08138d2a53286b3f8cfc2c09d28fc0fd2b/execution_engine/src/storage/trie/mod.rs#L123

Fraser999: The problem here is that `index` is serialized as a `usize`, not a `u8`.

We could change the serialization to use `u8` rather than `usize` and this direct access would be appropriate, but I was worried that if we change the radix to something other than 256, we'd run the risk of making this unsafe.  A malformed, serialized `PointerBlock` could cause a panic - a simple attack.

If we're convinced that we'll never change the radix from 256, then I'm fine with this suggestion (I'd change the serialization to use `u8` for the index)

EdHastingsCasperLabs (approved):

execution_engine/src/storage/trie/mod.rs
xcthulhu: Makes sense.  We should probably consider changing the old code then, but that seems out of scope.

Fraser999:
bors merge

PR #932, opened 2021-02-17
NO-TICK: set lfb hash back to original

TomVasile:
Previous change caused the synchronization function to always be true. Reverting back to how it originally was.


goral09 (approved):
Totally missed that change.

goral09:
bors merge

PR #936, opened 2021-02-17
NDRS-910: Introducing failing test for joiner with network of multiple nodes

xcthulhu:


goral09:
Why does this fail @xcthulhu ? We have a very similar test using nctl which does pass.

xcthulhu:
Here's the logs I'm seeing:

```
Feb 17 11:05:38.253 ERROR crank{; node_id=NodeId::Tls(4bd7..cd43)}: [casper_node::effect effect.rs:198] responder(()) dropped without being responded to --- this is always a bug and will likely cause another component to be stuck!
Feb 17 11:05:38.261 ERROR [casper_node::effect effect.rs:406] request for () channel closed, this is a serious bug --- a component will likely be stuck from now on ; err=oneshot canceled; queue_kind=Network
```

I suspect the issue is in the `MultiStageReactor`.  CC @marc-casperlabs

goral09:
If we have already established that this happens when the test transitions from a `joiner` reactor to `validator`, without executing all of the effects, maybe we should add a test-only `reactor::drain()` method that would just process all of the effects without enqueuing their resulting events?

goral09 (commented):
The test above that is also called `test_joiner` – can you add a comment what's the difference between them?

xcthulhu:
> The test above that is also called `test_joiner` – can you add a comment what's the difference between them?

Fixed in https://github.com/CasperLabs/casper-node/pull/936/commits/a435a66c8c74ea099af2cb06b1f3dfca893ae097

goral09 (approved):
And it's still failing b/c of the problems with joiner -> validator reactors transition?

xcthulhu:
> And it's still failing b/c of the problems with joiner -> validator reactors transition?

I think it's because we don't retry connecting...

xcthulhu:
bors merge

xcthulhu:
bors mergge

xcthulhu:
bors merge

xcthulhu:
bors retry

xcthulhu:
bors merge

PR #940, opened 2021-02-17
NDRS-501: Improve Client Error Handling

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-501

CHANGELOG:

- Defined errors for the client subcommand allowing them to gracefully exit when the node sends an error message related to the result of a client sub-command

@momipsl These changes seem to break the way `nctl` interacts with the client.

Sample Output:
Screenshot from 2021-02-17 14-43-49


goral09:
This looks great but I don't think those are all problems we had with a client. Try using a `String` for payment amount or a hex-encoded where it's not expected etc.

darthsiroftardis:
Closing due to mis-implementation

goral09:
> Closing due to mis-implementation

Why mis-implementation? This looks good, just isn't complete yet.

PR #942, opened 2021-02-18
NDRS-895: provide hard reset chainspec option

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-895

This PR adds functionality which allows an upgrade chainspec to specify that the node should ignore all blocks created during and after a given era ID.

This will hopefully never be needed, but until our upgrades can handle starting mid-era, we can use this as a fallback mechanism.


goral09 (approved):

Fraser999:
bors merge

PR #944, opened 2021-02-18
NDRS-877: handle termination signals gracefully

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-877

This PR introduces handling for termination signals which will allow the node to exit gracefully.  This should help avoid database or file corruption as all destructors should run properly before the app calls `std::process::exit()`.

I tested this locally, and the node does exit with a code of 130, 131 or 143 when sent an `INT`, `QUIT` or `TERM` signal respectively, which are the default exit codes Rust uses (128 + the signal value).


dwerner (approved):

Fraser999:
bors merge

PR #945, opened 2021-02-18
EE-1172: WIP delegators vesting schedule

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1172

- NOTE: This draft PR has temporarily modified genesis process to inject genesis delegators to support test for this feature. Genesis delegators will part of EE-1138
- This PR adds vesting schedule to genesis delegators which has similar behavior to vesting schedule of genesis validators

### DIFF
--- a/execution_engine/src/core/engine_state/genesis.rs
+++ b/execution_engine/src/core/engine_state/genesis.rs
@@ -586,6 +586,14 @@ pub enum GenesisError {
     ExecutionError(execution::Error),
     MintError(mint::Error),
     CLValue(String),
+    OrphanedDelegator {
+        validator_public_key: PublicKey,
+        delegator_public_key: PublicKey,
+    },
+    DuplicatedDelegatorEntry {
+        validator_public_key: PublicKey,
+        delegator_public_key: PublicKey,
+    },
 }
 
 pub(crate) struct GenesisInstaller<S>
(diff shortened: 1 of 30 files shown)


execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -120,13 +120,15 @@ impl GenesisResult {
...
-#[derive(DataSize, Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(DataSize, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct GenesisAccount {
     /// Assumed to be a system account if `public_key` is not specified.
     public_key: Option<PublicKey>,
     account_hash: AccountHash,
     balance: Motes,
     bonded_amount: Motes,
+    // TODO: EE-1138 implements genesis delegators
+    pub delegators: Vec<(PublicKey, U512)>,
```
henrytill, line 131: Let's try to avoid `pub`ing this field.

types/src/system/auction/bid/mod.rs
```diff
@@ -175,7 +175,21 @@ impl Bid {
...
+
+        if !vesting_schedule.initialize(staked_amount) {
+            return false;
+        }
+
+        for delegator in self.delegators_mut().values_mut() {
+            let staked_amount = *delegator.staked_amount();
+            if let Some(vesting_schedule) = delegator.vesting_schedule_mut() {
+                if !vesting_schedule.initialize(staked_amount) {
+                    return false;
```
henrytill, line 187: Seems like we would want to continue here as well.

henrytill (commented):
Looking good so far, but I think we need to take a closer look at the loop in `process`.  

As written the, the code would not properly support the case where delegators might have different `initial_release_timestamp_millis` values in their vesting schedules - namely if one delegator had already had its vesting schedule initialized in a prior call, we would exit the loop early.  

While in practice they will likely all have the same value of 90 days, I don't see any harm in writing this code in a way that supports them having different values.

mpapierski:
Had to force push bunch of times so due to that I'm closing in favor of a new fresh PR with clean history.

PR #947, opened 2021-02-18
NDRS-880: Add `nctl-upgrade-protocol` command.

goral09:
This PR adds a new nctl command `nctl-upgrade-protocol`. The command takes two arguments `version` and `era`. Version is expected to be in format `X_Y_Z`.

The command will create `version` directories in all `node-X` directories and copy required files (`casper-node` binary, chainspec and config files). Chainspec will only be updated with the new protocol version and `activation_point.era`. No other changes to the protocol are made.

### Notes

We recognize that this code is not exactly up to `nctl` standards but the command is proving very useful when testing the node upgrades so we want to give it to people early while committing to improving the code as soon as possible.


linclelinkpart5 (approved):
Looks good to me, this looks very helpful!

goral09:
bors merge

PR #948, opened 2021-02-18
EE-1174: Cap delegation rate to 100.

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1174

This PR caps delegation rate to 100.

Denominator is also changed to 100 so this way commission rate would be a simple percentage instead of a fraction of trillionths.


EdHastingsCasperLabs (approved):

EdHastingsCasperLabs:
basically good to go, after you fixup failing tests

PR #951, opened 2021-02-18
OP-1836: NCTL - random validator node shutdown and sync

TomVasile:
Implements itst01 from https://casperlabs.atlassian.net/wiki/spaces/OPS/pages/1133707277/Casper+Network+Infra+Test+Scenarios.

Also refactor nightly ci script to run multiple tests.

Test Run: https://drone-auto.casperlabs.io/TomVasile/casper-node/49/1/2


utils/nctl/sh/scenarios/itst01.sh
```diff
@@ -0,0 +1,143 @@
...
+    log_step "awaiting full synchronization of node=${NODE_ID}…"
+    while [ "$(do_read_lfb_hash "$NODE_ID")" != "$(do_read_lfb_hash "$COMPARE_NODE_ID")" ]; do
+        if [ "$WAIT_TIME_SEC" = "$SYNC_TIMEOUT_SEC" ]; then
+            log "ERROR: Failed to synchronize in ${SYNC_TIMEOUT_SEC} seconds"
+            exit 1
+        fi
+        WAIT_TIME_SEC=$((WAIT_TIME_SEC + 1))
+        sleep 1.0
+    done
+    # Wait 2 more era and then test chainheight.
```
goral09, line 97:
```suggestion
    # Wait 2 more era and then test chain height.
```

utils/nctl/sh/scenarios/itst01.sh
```diff
@@ -0,0 +1,143 @@
...
+    # This way we can verify that the node is up-to-date with the protocol state
+    # after transitioning to an active validator and the chain progressed.
+    do_await_era_change
+    log_step "verifying full synchronization of node=${NODE_ID}…"
+    while [ "$(get_chain_height "$NODE_ID")" != "$(get_chain_height "$COMPARE_NODE_ID")" ]; do
```
goral09, line 102: Why chain height and not LFB? It's more precise.

TomVasile: I noticed in my test that the restarted node would sync to the LFB and then the chain would progress but the restarted node would take a bit to catch up to the chain height. 

Thats why in the first pass I check LFB and the second pass checks the height

goral09: Yeah, there seems to be an issue with the networking component losing connections on transition. You should still be using LFB for verification.

TomVasile: Updated 👍

utils/nctl/sh/scenarios/itst01.sh
```diff
@@ -0,0 +1,138 @@
...
+    log "------------------------------------------------------------"
+
+    log "Node to be stopped: $NODE_ID"
+
+    # 0. Wait for network start up
+    do_await_genesis_era_to_complete
+    # 1. Allow chain to progress
+    do_await_era_change
+    # 2. Stop random node
+    do_stop_node "$NODE_ID"
```
goral09, line 30: You're passing in the `NODE-ID` here to stop but this node is implicitly also started in `do_start_node` (i.e. we don't pass in the node id) – let's make it consistent across the whole test.

utils/nctl/sh/scenarios/itst01.sh
```diff
@@ -0,0 +1,138 @@
...
+    do_await_era_change
+    log_step "verifying full synchronization of node=${NODE_ID}…"
+    while [ "$(do_read_lfb_hash "$NODE_ID")" != "$(do_read_lfb_hash "$COMPARE_NODE_ID")" ]; do
+        if [ "$WAIT_TIME_SEC" = "$SYNC_TIMEOUT_SEC" ]; then
+            log "ERROR: Failed to keep up with the protocol state"
+            exit 1
+        fi
+        WAIT_TIME_SEC=$((WAIT_TIME_SEC + 1))
+        sleep 1.0
+    done
```
goral09, line 104: It would be nice if we had a function to compare lfb on all the running nodes at once – so that we're definitely sure that no node is falling back.

goral09: cc @momipsl @linclelinkpart5

utils/nctl/sh/scenarios/itst01.sh
TomVasile: fixed

goral09 (approved):

PR #954, opened 2021-02-19
NO-TICKET: add public key and round length to status endpoint

Fraser999:
This PR enhances the status endpoints by providing the public signing key of the node and the next round length if the node is a validator.  Example output:

```json
{
  "api_version": "1.0.0",
  "chainspec_name": "casper-net-1",
  "starting_state_root_hash": "fd7b..0314",
  "peers": [
    {
      "node_id": "NodeId::Tls(470b..d032)",
      "address": "127.0.0.1:34554"
    },
    {
      "node_id": "NodeId::Tls(606d..b959)",
      "address": "127.0.0.1:34555"
    },
    {
      "node_id": "NodeId::Tls(a0ef..4e84)",
      "address": "127.0.0.1:34557"
    },
    {
      "node_id": "NodeId::Tls(e85b..0726)",
      "address": "127.0.0.1:34556"
    }
  ],
  "last_added_block_info": {
    "hash": "4453a99848956ad0b81883c78e6434a58997e0532314b4125746a0ee7e7accb6",
    "timestamp": "2021-02-19T00:28:12.672Z",
    "era_id": 3,
    "height": 39,
    "state_root_hash": "ba922bb17997a4481f3f3850560c7e209b3fee889d2b66926650480c8cfaa29f",
    "creator": "015ac2b49f86688057fc2426bf16067bfe428fbb33d1007f1c1583e7f59233cedd"
  },
  "our_public_signing_key": "011925e95db3df8da31ee8dc406471cc29cd413a6ee0dc08aef3eb5732a7b63f07",
  "round_length": "4s 96ms",
  "next_upgrade": null,
  "build_version": "0.9.0-7f5425bf"
}
```

### DIFF
--- a/node/src/components/consensus/consensus_protocol.rs
+++ b/node/src/components/consensus/consensus_protocol.rs
@@ -205,5 +205,6 @@ pub(crate) trait ConsensusProtocol<I, C: Context>: Send {
     /// TODO: Remove this once the Joiner no longer has a consensus component.
     fn recreate_timers(&self) -> Vec<ProtocolOutcome<I, C>>;
 
+    // TODO: Make this lees Highway-specific.
     fn next_round_length(&self) -> Option<TimeDiff>;
 }
--- a/node/src/components/consensus/consensus_protocol.rs
+++ b/node/src/components/consensus/consensus_protocol.rs
@@ -6,7 +6,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     components::consensus::{traits::Context, ActionId, TimerId},
-    types::Timestamp,
+    types::{TimeDiff, Timestamp},
     NodeRng,
 };
 
@@ -204,4 +204,6 @@ pub(crate) trait ConsensusProtocol<I, C: Context>: Send {
     /// Returns the protocol outcomes for all the required timers.
     /// TODO: Remove this once the Joiner no longer has a consensus component.
     fn recreate_timers(&self) -> Vec<ProtocolOutcome<I, C>>;
+
+    fn next_round_length(&self) -> Option<TimeDiff>;
 }
--- a/node/src/components/consensus/protocols/highway.rs
+++ b/node/src/components/consensus/protocols/highway.rs
@@ -357,22 +357,19 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
     /// Returns the median round exponent of all the validators that haven't been observed to be
     /// malicious, as seen by the current panorama.
     /// Returns `None` if there are no correct validators in the panorama.
-    pub(crate) fn median_round_exp(&self) -> Option<u8> {
+    fn median_round_exp(&self) -> Option<u8> {
         self.highway.state().median_round_exp()
     }
 
     /// Returns an instance of `RoundSuccessMeter` for the new era: resetting the counters where
     /// appropriate.
-    pub(crate) fn next_era_round_succ_meter(
-        &self,
-        era_start_timestamp: Timestamp,
-    ) -> RoundSuccessMeter<C> {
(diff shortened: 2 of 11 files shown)


goral09 (dismissed):
We cannot be publicly announcing our validator key. That's an attack vector. Random bits and leader sequence are deterministic and public. If an attacker learns the sequence of leaders and knows which nodes are theirs they can target the attack to take out the consecutive leaders to disrupt and stall the network.

goral09:
I don't know how useful announcing round exponent is. It keeps changing.

EdHastingsCasperLabs (approved):

EdHastingsCasperLabs:
> I don't know how useful announcing round exponent is. It keeps changing.
That's why we want to know what it currently is.

EdHastingsCasperLabs:
>We cannot be publicly announcing our validator key. That's an attack vector. Random bits and leader sequence are deterministic and public. If an attacker learns the sequence of leaders and knows which nodes are theirs they can target the attack to take out the consecutive leaders to disrupt and stall the network.

Lets discuss this aspect in standup tomorrow.

node/src/components/consensus/consensus_protocol.rs
```diff
@@ -204,4 +204,6 @@ pub(crate) trait ConsensusProtocol<I, C: Context>: Send {
     /// Returns the protocol outcomes for all the required timers.
     /// TODO: Remove this once the Joiner no longer has a consensus component.
     fn recreate_timers(&self) -> Vec<ProtocolOutcome<I, C>>;
+
+    fn next_round_length(&self) -> Option<TimeDiff>;
```
afck, line 208: Let's add a TODO and create a story in [FEAT-274](https://casperlabs.atlassian.net/browse/FEAT-274) to find a way to report Highway-specific details without making this trait Highway-specific.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -357,22 +357,19 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
     /// Returns the median round exponent of all the validators that haven't been observed to be
     /// malicious, as seen by the current panorama.
     /// Returns `None` if there are no correct validators in the panorama.
-    pub(crate) fn median_round_exp(&self) -> Option<u8> {
+    fn median_round_exp(&self) -> Option<u8> {
```
afck, line 360: Even by just reporting _our_ round length, we are giving away the information that this node is a validator. And in addition, and attacker could compare the reported round lengths with the round exponents specified in the different validators' units, to eventually find out _which_ validator that node is.

Instead, we could always report the median round length? Then every node in the network would return the same information.

Alternatively, we could return a whole list of the different validators' most recent round lengths, or a histogram. That would also be returned by every node in the network.

xcthulhu:
This status endpoint is clearly valuable for node operators. Couldn't we restrict it so we can only query it from `127.0.0.1`?

goral09:
> This status endpoint is clearly valuable for node operators. Couldn't we restrict it so we can only query it from `127.0.0.1`?

Node operators already know what their keys are and we log the round exponent.

Fraser999:
bors merge

node/src/components/consensus/consensus_protocol.rs
Fraser999: Done in [b23a825](https://github.com/CasperLabs/casper-node/pull/954/commits/b23a825ab7e68d4c32ee7e6ae0c449cdefac674d).

Fraser999:
bors merge

Fraser999:
bors merge

node/src/components/consensus/consensus_protocol.rs
piotr-dziubecki: https://casperlabs.atlassian.net/browse/HWY-260

PR #956, opened 2021-02-19
NDRS-912: Disable slashing.

afck:
For the first 90 days we only evict equivocating validators instead of slashing them.

https://casperlabs.atlassian.net/browse/NDRS-912

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -909,8 +909,15 @@ where
                 // the block or seen as equivocating via the consensus protocol gets slashed.
                 let era_end = terminal_block_data.map(|tbd| EraReport {
                     rewards: tbd.rewards,
-                    equivocators: era.accusations(),
-                    inactive_validators: tbd.inactive_validators,
+                    // TODO: In the first 90 days we don't slash, and we just report all
+                    // equivocators as "inactive" instead. Change this back 90 days after launch,
+                    // and put era.accusations() into equivocators instead of inactive_validators.
+                    equivocators: vec![],
+                    inactive_validators: tbd
+                        .inactive_validators
+                        .into_iter()
+                        .chain(era.accusations())
+                        .collect(),
                 });
                 let finalized_block = FinalizedBlock::new(
                     value.into(),


node/src/components/consensus/era_supervisor.rs
```diff
@@ -909,8 +909,15 @@ where
...
-                    inactive_validators: tbd.inactive_validators,
+                    // TODO: In the first 90 days we don't slash, and we just report all
+                    // equivocators as "inactive" instead. Change this back 90 days after launch,
+                    // and put era.accusations() into equivocators instead of inactive_validators.
+                    equivocators: vec![],
+                    inactive_validators: tbd
+                        .inactive_validators
+                        .into_iter()
+                        .chain(era.accusations())
+                        .collect(),
```
goral09, line 920: Shouldn't we already encode this 90-day lockup in here?

afck: Yesterday we discussed that we wouldn't; and we'd just remove this with an update.

goral09: I don't think it's feasible – since upgrades are era-based and we're supposed to disable slashing for 90 days, we can't possible announce an upgrade that will take effect exactly 90 days genesis.

afck: Good point. Let's discuss in standup today.
I had assumed it wouldn't have to be _exactly_ 90 days.

EdHastingsCasperLabs (approved):

node/src/components/consensus/era_supervisor.rs
EdHastingsCasperLabs: Andreas is correct; the upgrade to re-enable slashing could go in at any point after the 90 days; it doesn't have to be razor sharp timing.

goral09: But we might miss the 90-days delay by … days or maybe even weeks actually. If we need to give people a couple weeks notice about the upgrade, pick the era at which we will do the upgrade, between the distributing new binary version and actually hitting that upgrade point we might speed up or slow down enough to miss that deadline by days.

afck: I created [NDRS-917](https://casperlabs.atlassian.net/browse/NDRS-917) for re-enabling it.

PR #958, opened 2021-02-19
NDRS-819: DB integrity check 

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-819

CHANGELOG:

- Added a check in the intializtion of a reactor to check if the `missing_trie_keys` is empty. If it is not, the node will panic
- Added a `pub` function to allow the reactor during initialization to get the current highest block from storage if it exists
- Added a `pub` function within the contract runtime to allow for querying of the trie store


node/src/components/contract_runtime.rs
```diff
@@ -593,4 +594,16 @@ impl ContractRuntime {
             &ee_config,
         )
     }
+
+    /// Retrieve missing trie-keys for a DB check at upgrade.
```
goral09, line 598:
```suggestion
    /// Retrieves missing trie-keys for a DB check at upgrade.
```

node/src/reactor/initializer.rs
```diff
@@ -164,6 +165,15 @@ impl Reactor {
...
             ContractRuntime::new(storage_config, &config.value().contract_runtime, registry)?;
 
+        if let Some(block) = storage.get_highest_block_header_for_db_check() {
+            let highest_block_header = block.take_header();
+            let blake_hash = Blake2bHash::new(highest_block_header.state_root_hash().as_ref());
+            let missing_trie_key = contract_runtime.retrieve_missing_keys(blake_hash);
+            if !missing_trie_key.is_empty() {
+                panic!("Fatal error! Trie-Key store is not empty. Wipe the DB to ensure operations")
+            }
+        }
```
goral09, line 175: I'm wondering whether we should also validate the integrity of a DB even before having any blocks? If yes, then this can probably be part of the `CommitGenesis` process. Thoughts @xcthulhu ?

node/src/components/contract_runtime.rs
```diff
@@ -41,6 +41,7 @@ use crate::{
     utils::WithDir,
     NodeRng, StorageConfig,
 };
+use casper_execution_engine::shared::newtypes::Blake2bHash;
```
goral09, line 44: This import should be up there with the rest of its family :)

goral09:
I would like to see this PR tested in various contexts:
* node starts, gets restarted and then tries synchronizing
* node upgrades
* node joins an upgraded network (including Global State changes)

Just to be sure that going through various states of the Global State and the block storage does not break this.

node/src/reactor/initializer.rs
xcthulhu: > I'm wondering whether we should also validate the integrity of a DB even before having any blocks?

I think we need to do the following:

  1.  Check all of the blocks
  2. Check all of the block bodies
  3. Check all of the deploys/hashes
  4. Check all of the `state_root_hash`s for those blocks

I suspect that (4) will require we change `retrieve_missing_keys` to do a bulk check.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -241,7 +241,7 @@ where
     E: From<S::Error> + From<bytesrepr::Error>,
 {
     let mut missing_descendants = Vec::new();
-    let mut trie_keys_to_visit_queue = vec![trie_key];
+    let mut trie_keys_to_visit_queue = trie_key;
```
xcthulhu, line 244: Because we can have tries with common children, this should be be a `HashSet`, not a `Vec`.

We could also consider renaming `trie_key` and `trie_keys_to_visit_queue` to `trie_keys_to_visit`, since it's a bit confusing they way they are currently named.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -271,19 +270,22 @@ where
...
                     }
                 }
             }
             // If we hit an extension block, add its pointer to the queue
             Some(Trie::Extension { pointer, .. }) => {
-                trie_keys_to_visit_queue.push(pointer.into_hash())
+                trie_keys_to_visit.push(pointer.into_hash())
             }
         }
+        // TODO Use BTreeSet when pop_first() is added to stable from nightly.
```
goral09, line 284: Are we ok with leaving this TODO in here?

darthsiroftardis: Yes yes yes, because ideally we want to use a `HashSet` or `BTreeSet` and we are using `Vec`currently. The `pop_first()` stuff isn't stable yet, but it would be better if we used either of those two instead of `Vec`

execution_engine/src/storage/global_state/in_memory.rs
```diff
@@ -284,7 +284,7 @@ impl StateProvider for InMemoryGlobalState {
     fn missing_trie_keys(
         &self,
         correlation_id: CorrelationId,
-        trie_key: Blake2bHash,
+        trie_key: Vec<Blake2bHash>,
```
goral09, line 287:
```suggestion
        trie_keys: Vec<Blake2bHash>,
```

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -243,7 +243,7 @@ impl StateProvider for LmdbGlobalState {
     fn missing_trie_keys(
         &self,
         correlation_id: CorrelationId,
-        trie_key: Blake2bHash,
+        trie_key: Vec<Blake2bHash>,
```
goral09, line 246:
```suggestion
        trie_keys: Vec<Blake2bHash>,
```

execution_engine/src/storage/global_state/mod.rs
```diff
@@ -128,7 +128,7 @@ pub trait StateProvider {
     fn missing_trie_keys(
         &self,
         correlation_id: CorrelationId,
-        trie_key: Blake2bHash,
+        trie_key: Vec<Blake2bHash>,
```
goral09, line 131:
```suggestion
        trie_keys: Vec<Blake2bHash>,
```

darthsiroftardis: Applied suggestion

execution_engine/src/storage/global_state/lmdb.rs
darthsiroftardis: Applied suggestion

execution_engine/src/storage/global_state/in_memory.rs
darthsiroftardis: Applied suggestion

darthsiroftardis:
@goral09 I tested this yesterday by running a 5 node network with `nctl` and performing an upgrade from era 1 to era 2 and the node continued to function nominally

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1688,7 +1688,20 @@ where
...
+    pub fn trie_integrity_check(
+        &self,
+        correlation_id: CorrelationId,
+        trie_keys: Vec<Blake2bHash>,
+    ) -> Result<Vec<Blake2bHash>, Error>
+    where
+        Error: From<S::Error>,
+    {
+        self.state
+            .missing_trie_keys(correlation_id, trie_keys)
```
goral09, line 1704: Can't we merge those two functions into one that accepts the vector of hashes?

node/src/components/storage.rs
```diff
@@ -307,6 +308,44 @@ impl Storage {
...
+
+        info!("Checking block_metadata_db");
+        let signature_txn = env.begin_ro_txn()?;
+        let mut cursor = signature_txn.open_ro_cursor(block_metadata_db)?;
+
+        for (raw_key, raw_val) in cursor.iter() {
+            let signature: BlockSignatures = lmdb_ext::deserialize(raw_val)?;
+            // Signature verification could be very slow process
+            // It iterates over every signature and verifies them.
+            match signature.verify() {
```
goral09, line 335: I don't think we should be doing this. Signatures are verified before being stored in the storage. On the other hand, if it's being done only once at a startup maybe it's not a problem.

execution_engine/src/core/engine_state/mod.rs
darthsiroftardis: We could, but I thought it made sense to keep those functions as distinct as possible. I thought it made more sense to split off to a once off function that is named as such and used for that once specific reason.

node/src/components/storage.rs
```diff
@@ -307,6 +308,44 @@ impl Storage {
...
+            let body: BlockBody = lmdb_ext::deserialize(raw_val)?;
+            assert_eq!(
+                raw_key,
+                body.hash().as_ref(),
+                "found corrupt block body in database"
+            );
+        }
+        info!("block body db check complete");
+        drop(cursor);
+        drop(body_txn);
```
goral09, line 325: Extract to a method please.

node/src/components/storage.rs
```diff
@@ -307,6 +308,44 @@ impl Storage {
...
+            }
+            assert_eq!(
+                raw_key,
+                signature.block_hash.as_ref(),
+                "Corruption in block_metadata_db"
+            );
+        }
+        info!("Check for block_metadata_db complete");
+        drop(cursor);
+        drop(signature_txn);
```
goral09, line 347: Extract to a method please.

node/src/components/storage.rs
```diff
@@ -712,6 +751,37 @@ impl Storage {
             .transpose()
     }
 
+    pub fn get_highest_block_header_for_db_check(&self) -> Option<Block> {
```
goral09, line 754: We can drop the `_for_db_check` suffix.

goral09: Also, since this a public method please add a docstring.

node/src/components/storage.rs
darthsiroftardis: Yeah it happens once during startup, but since we are checking  the integrity of the data, we can't rely on data being stored being in the same condition when we do startup. Its a slow process, but at this time I think its worth it.

node/src/components/storage.rs
```diff
@@ -712,6 +751,37 @@ impl Storage {
...
+    pub fn get_state_root_hashes_for_trie_check(&self) -> Option<Vec<Blake2bHash>> {
+        let mut blake_hashes: Vec<Blake2bHash> = Vec::new();
+        let header_txn = match self.env.begin_ro_txn() {
+            Ok(txn) => txn,
+            Err(_) => return None,
+        };
+        let mut cursor = match header_txn.open_ro_cursor(self.block_header_db) {
+            Ok(cursor) => cursor,
+            Err(_) => return None,
+        };
```
goral09, line 772: You can replace the `match` statements here with `.ok()?`. `ok()` turns `Result` into `Option` type, ignoring the error.

node/src/components/storage.rs
```diff
@@ -712,6 +751,37 @@ impl Storage {
...
+        for (_, raw_val) in cursor.iter() {
+            let header: BlockHeader = lmdb_ext::deserialize(raw_val).ok()?;
+            let blake_hash = Blake2bHash::from(*header.state_root_hash());
+            blake_hashes.push(blake_hash);
+        }
+
+        blake_hashes.sort();
+        blake_hashes.dedup();
```
goral09, line 780: Do we need to `dedup()`? Wouldn't this mean we have duplicate headers with the same hashes in the storage? That should be a fatal error in my opinion.

darthsiroftardis: @xcthulhu You would explain this better than myself

node/src/components/storage.rs
```diff
@@ -712,6 +751,37 @@ impl Storage {
...
+        if let Some(height) = self.block_height_index.keys().last() {
+            self.get_block_by_height(&mut txn, *height).ok()?
+        } else {
+            None
+        }
+    }
+
+    pub fn get_state_root_hashes_for_trie_check(&self) -> Option<Vec<Blake2bHash>> {
+        let mut blake_hashes: Vec<Blake2bHash> = Vec::new();
+        let header_txn = match self.env.begin_ro_txn() {
```
goral09, line 765: This is just a `txn`, there's nothing specific to `header` about it.

node/src/components/storage.rs
```diff
@@ -712,6 +751,37 @@ impl Storage {
...
+    pub fn get_highest_block_header_for_db_check(&self) -> Option<Block> {
+        let mut txn = self.env.begin_ro_txn().ok()?;
+        if let Some(height) = self.block_height_index.keys().last() {
+            self.get_block_by_height(&mut txn, *height).ok()?
+        } else {
+            None
+        }
+    }
+
+    pub fn get_state_root_hashes_for_trie_check(&self) -> Option<Vec<Blake2bHash>> {
```
goral09, line 763: Please, add documentation/comment to this method.

node/src/components/storage.rs
darthsiroftardis: @marc-casperlabs Any thoughts on how to extract this

node/src/reactor/initializer.rs
```diff
@@ -164,6 +164,17 @@ impl Reactor {
...
+        if let Some(state_roots) = storage.get_state_root_hashes_for_trie_check() {
+            let missing_trie_keys = contract_runtime.trie_store_check(state_roots.clone());
+            if !missing_trie_keys.is_empty() {
+                panic!(
+                    "Fatal error! Trie-Key store is not empty.\n {:?}\n \
+                    Wipe the DB to ensure operations.\n Present state_roots: {:?}",
+                    missing_trie_keys, state_roots
+                )
+            }
+        }
```
goral09, line 176: Shouldn't this be a method on the storage itself? Something like `storage.validate()`?

goral09: Ah, it needs `contract_runtime` to access the GlobalState :thinking: :+1:

node/src/types/block.rs
```diff
@@ -997,6 +997,19 @@ impl BlockSignatures {
...
+    /// Verify the signatures contained within.
+    pub(crate) fn verify(&self) -> crypto::Result<()> {
+        let mut bytes = self.block_hash.inner().to_vec();
+        bytes.extend_from_slice(&self.era_id.0.to_le_bytes());
+        for (key, signature) in self.proofs.iter() {
+            match crypto::verify(bytes.clone(), signature, key) {
+                Ok(_) => continue,
+                Err(error) => return Err(error),
+            }
+        }
```
goral09, line 1010: Instead of recreating the signature verification code (which risks doing it incorrectly) we should recreate the `FinalitySignature` and call `verify()` on it:
```rust
let signature = FinalitySignature {
                block_hash: self.block_hash,
                era_id: self.era_id,
                signature: signature.clone(),
                public_key: public_key.clone(),
            };
            match signature.verify() {
                Ok(_) => continue,
                Err(error) => return Err(error),
            }
```

node/src/components/contract_runtime.rs
```diff
@@ -602,4 +602,28 @@ impl ContractRuntime {
...
+    pub fn retrieve_missing_keys(&self, trie_key: Blake2bHash) -> Vec<Blake2bHash> {
+        let correlation_id = CorrelationId::new();
+        match self
+            .engine_state
+            .missing_trie_keys(correlation_id, trie_key)
+        {
+            Ok(keys) => keys,
+            Err(error) => panic!("Error in retrieving keys for DB check: {:?}", error),
+        }
+    }
```
goral09, line 616: This doesn't seem to be used anywhere.

execution_engine/src/core/engine_state/mod.rs
goral09: Depends on how many places the single-hash version is called. If it's only one, and we don't break call sides, I would merge it. Not going to push, just a preference.

node/src/components/storage.rs
goral09: Yes, I agree. It's a good decision. We need to make sure that our data is correct :+1:

node/src/components/storage.rs
```diff
@@ -307,6 +308,22 @@ impl Storage {
...
+        });
+
+        // Check the integrity of the block metadata database.
+        check_block_metadata_db(&env, &block_metadata_db).unwrap_or_else(|error| {
+            panic!(
+                "Error in the integrity check of the block_metadata database: {:?}",
+                error
+            )
+        });
+
```
goral09, line 326: Can you extract the indexing above to a method as well? Thank you.

node/src/components/storage.rs
```diff
@@ -307,6 +308,22 @@ impl Storage {
...
+        // Check the integrity of the block metadata database.
+        check_block_metadata_db(&env, &block_metadata_db).unwrap_or_else(|error| {
+            panic!(
+                "Error in the integrity check of the block_metadata database: {:?}",
+                error
+            )
```
goral09, line 324: I am not sure we should be the ones deciding to panic. Can't you simply call `?` on the method and pass on the error to the caller of the constructor?

node/src/components/storage.rs
```diff
@@ -989,3 +1029,52 @@ impl Storage {
...
+        let body: BlockBody = lmdb_ext::deserialize(raw_val)?;
+        assert_eq!(
+            raw_key,
+            body.hash().as_ref(),
+            "found corrupt block body in database"
+        );
+    }
+    info!("block body db check complete");
+    drop(cursor);
+    drop(body_txn);
```
goral09, line 1049: Do we have to call `drop` explicitly? Shouldn't Rust take care of it and call `Drop::drop` on the `cursor` and `body_txn` as we exit the method?

node/src/components/storage.rs
darthsiroftardis: Extracted

node/src/components/storage.rs
darthsiroftardis: Added docstring

node/src/components/storage.rs
darthsiroftardis: Removed match statements

node/src/components/storage.rs
darthsiroftardis: Renamed to `txn`

node/src/types/block.rs
darthsiroftardis: Applied suggestion

node/src/components/storage.rs
darthsiroftardis: Aaah yes we could

node/src/components/storage.rs
darthsiroftardis: Extracted

node/src/components/storage.rs
darthsiroftardis: Do you mean the block header re-indexing, if so, then I'd rather leave that where it is and now disturb Marc's rewrite of the storage component.

node/src/components/storage.rs
darthsiroftardis: Changed to use `?`

node/src/components/storage.rs
darthsiroftardis: Exiting method w/o explicit `drop`

node/src/components/storage.rs
```diff
@@ -1056,8 +1044,8 @@ fn check_block_metadata_db(
     block_metadata_db: &Database,
 ) -> Result<(), LmdbExtError> {
     info!("Checking block_metadata_db");
-    let signature_txn = env.begin_ro_txn()?;
-    let mut cursor = signature_txn.open_ro_cursor(*block_metadata_db)?;
+    let tnx = env.begin_ro_txn()?;
```
goral09, line 1047: I think you meant `txn`?

darthsiroftardis: D A R N

node/src/components/storage.rs
goral09: I would like to see this resolved before merging. /cc @xcthulhu

node/src/components/storage.rs
goral09: I am still seeing `drop`s called explicitly here for some reason. I thought you had removed them.

darthsiroftardis: It says `outdated` at the top, perhaps from an earlier commit??

node/src/components/storage.rs
goral09: Resolved offline.

goral09 (approved):

node/src/components/storage.rs
darthsiroftardis: I've renamed it

node/src/types/block.rs
```diff
@@ -997,6 +997,21 @@ impl BlockSignatures {
...
+    pub(crate) fn verify(&self) -> crypto::Result<()> {
+        for (public_key, signature) in self.proofs.iter() {
+            let signature = FinalitySignature {
+                block_hash: self.block_hash,
+                era_id: self.era_id,
+                signature: *signature,
+                public_key: *public_key,
+            };
+            signature.verify()?;
+            continue;
```
xcthulhu, line 1011: This continue is not necessary.

node/src/components/storage.rs
```diff
@@ -989,3 +1019,48 @@ impl Storage {
...
+fn check_block_metadata_db(
+    env: &Environment,
+    block_metadata_db: &Database,
+) -> Result<(), LmdbExtError> {
+    info!("Checking block_metadata_db");
+    let txn = env.begin_ro_txn()?;
+    let mut cursor = txn.open_ro_cursor(*block_metadata_db)?;
+
+    for (raw_key, raw_val) in cursor.iter() {
+        let signature: BlockSignatures = lmdb_ext::deserialize(raw_val)?;
```
xcthulhu, line 1051:
```suggestion
        let signatures: BlockSignatures = lmdb_ext::deserialize(raw_val)?;
```

node/src/components/storage.rs
```diff
@@ -989,3 +1019,48 @@ impl Storage {
...
+        // Signature verification could be very slow process
+        // It iterates over every signature and verifies them.
+        match signature.verify() {
+            Ok(_) => continue,
+            Err(error) => error!("Error: {} in signature verification. This could indicate a corruption error in the database", error)
```
xcthulhu, line 1056:
```suggestion
            Err(error) => panic!("Error: {} in signature verification. Corruption in block_metadata_db.", error)
```

node/src/components/storage.rs
```diff
@@ -989,3 +1019,48 @@ impl Storage {
...
+    info!("Checking block_metadata_db");
+    let txn = env.begin_ro_txn()?;
+    let mut cursor = txn.open_ro_cursor(*block_metadata_db)?;
+
+    for (raw_key, raw_val) in cursor.iter() {
+        let signature: BlockSignatures = lmdb_ext::deserialize(raw_val)?;
+        // Signature verification could be very slow process
+        // It iterates over every signature and verifies them.
+        match signature.verify() {
+            Ok(_) => continue,
```
xcthulhu, line 1055: We can't `continue` here, since then we won't go to the `assert_eq!` below.

darthsiroftardis: Moving the `assert_eq!` into the match statement

node/src/types/block.rs
darthsiroftardis: Removed `continue`

node/src/components/storage.rs
darthsiroftardis: Applied suggestion

node/src/components/storage.rs
darthsiroftardis: Applied suggestion

node/src/components/storage.rs
xcthulhu: This is done elsewhere instead of just calling [lmdb::Transaction::abort][1] directly.

[1]: https://docs.rs/lmdb-rkv/0.14.0/lmdb/trait.Transaction.html#method.abort

node/src/components/storage.rs
darthsiroftardis: Removed `continue` and moved `assert_eq` into the match statement

node/src/components/storage.rs
goral09: `drop` wouldn't abort the txn though so they're different things.

xcthulhu (approved):

PR #959, opened 2021-02-19
OP-1837: NCTL - validator node shutdown >35% weight and resume 

TomVasile:
Implements itst02 from https://casperlabs.atlassian.net/wiki/spaces/OPS/pages/1133707277/Casper+Network+Infra+Test+Scenarios.

Test Run: https://drone-auto.casperlabs.io/TomVasile/casper-node/51/1/2


goral09:
I don't have access to the drone run link.

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,165 @@
...
+}
+
+function log_step() {
+    local COMMENT=${1}
+    log "------------------------------------------------------------"
+    log "STEP $STEP: $COMMENT"
+    STEP=$((STEP + 1))
+}
+
+function compare() {
```
goral09, line 66: Can you add a comment to this function? `log_stop` is kind of self-described but this one has an implicit assumption that the second argument is the expected LFB.

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,165 @@
...
+    # 0. Wait for network start up
+    do_await_genesis_era_to_complete
+    # 1. Allow chain to progress
+    do_await_era_change
+    # 2-4. Stop three nodes
+    do_stop_node "5"
+    do_stop_node "4"
+    do_stop_node "3"
+    # 5. Ensure chain stalled
+    sleep_and_compare "60"
```
goral09, line 34: Just FYI, this will work _iff_ the round length is lower than 60 seconds.

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,165 @@
...
+    log_step "awaiting full synchronization of node=${NODE_ID}…"
+    while [ "$(do_read_lfb_hash "$NODE_ID")" != "$(do_read_lfb_hash "1")" ]; do
+        if [ "$WAIT_TIME_SEC" = "$SYNC_TIMEOUT_SEC" ]; then
+            log "ERROR: Failed to synchronize node-${NODE_ID} in ${SYNC_TIMEOUT_SEC} seconds"
+            exit 1
+        fi
+        WAIT_TIME_SEC=$((WAIT_TIME_SEC + 1))
+        sleep 1.0
+    done
+}
```
goral09, line 136: We should have the double assertion that we have in the other tests. Otherwise, we might be getting false positives and ignore a case where the node after syncing does not transition to an active participant of the protocol. I know it's slower but it's the only way we can be sure it's working correctly :grimacing:

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,165 @@
...
+function main() {
+    log "------------------------------------------------------------"
+    log "Starting Scenario: itst02"
+    log "------------------------------------------------------------"
+
+    # 0. Wait for network start up
+    do_await_genesis_era_to_complete
+    # 1. Allow chain to progress
+    do_await_era_change
+    # 2-4. Stop three nodes
```
goral09, line 29: Maybe, before stopping nodes, we should verify that they're all running correctly and agree on the LFB.

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,165 @@
...
+    fi
+}
+
+function sleep_and_compare() {
+    log_step "ensuring chain stalled"
+    local SLEEP_TIME=${1}
+    local LFB1=$(do_read_lfb_hash 1)
+    log "Sleeping ${SLEEP_TIME}s..."
+    sleep $SLEEP_TIME
+    local LFB2=$(do_read_lfb_hash 1)
```
goral09, line 83: A better test will be to compare LFBs on two different nodes (the only ones that are still operating).

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,165 @@
...
+    local LFB1=$(do_read_lfb_hash ${1})
+    local LFB2=${2}
+
+    if [ "$LFB1" = "$LFB2" ]; then
+       log "error: $LFB1 = $LFB2, chain didn't progress."
+       exit 1
+    fi
+}
+
+function sleep_and_compare() {
```
goral09, line 77: This function is very similar to `compare` but the assertion is inverted. Can we have a separate function `assert_chain_stalled` and call it from the `sleep` here?

utils/nctl/sh/scenarios/itst02.sh
TomVasile: I actually do test it twice, I just rearranged how.

`# 9-11. Check sync of restarted node
    do_await_full_synchronization "5"
    do_await_full_synchronization "4"
    do_await_full_synchronization "3"
    # 12. Ensure era proceeds after restart
    do_await_era_change "2"
    # 13-15. Check sync of nodes again
    do_await_full_synchronization "5"
    do_await_full_synchronization "4"
    do_await_full_synchronization "3"`

goral09: Ah, I see. I missed that :) We should write this test function that asserts that LFB is the same across all nodes.

utils/nctl/sh/scenarios/itst02.sh
TomVasile: you are better at names 😄 I'll update it

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env bash
+
+source "$NCTL"/sh/utils/main.sh
+source "$NCTL"/sh/views/utils.sh
+source "$NCTL"/sh/node/svc_"$NCTL_DAEMON_TYPE".sh
+source "$NCTL"/sh/utils/infra.sh
```
goral09, line 6: `source "$NCTL"/sh/views/utils.sh` already sources `infra.sh`.

utils/nctl/sh/scenarios/itst02.sh
TomVasile: should be addressed here: https://github.com/CasperLabs/casper-node/pull/959/files#diff-8bd21e6750b7829a78940cd1d0405bfe8e4ba19e78acb53c085cb8cfba164d41R69-R71

utils/nctl/sh/scenarios/itst02.sh
```diff
@@ -0,0 +1,200 @@
...
+    log_step "awaiting full synchronization of node=${NODE_ID}…"
+    while [ "$(do_read_lfb_hash ${NODE_ID})" != "$(do_read_lfb_hash '1')" ]; do
+        if [ "$WAIT_TIME_SEC" = "$SYNC_TIMEOUT_SEC" ]; then
+            log "ERROR: Failed to synchronize node-${NODE_ID} in ${SYNC_TIMEOUT_SEC} seconds"
+            exit 1
+        fi
+        WAIT_TIME_SEC=$((WAIT_TIME_SEC + 1))
+        sleep 1.0
+    done
+}
```
goral09, line 171: Do we need this if we already have `check_network_sync`?

goral09 (approved):
Cool. No blockers :)

utils/nctl/sh/scenarios/itst02.sh
TomVasile: added a fuction to check across all of the nodes used in this test: https://github.com/CasperLabs/casper-node/pull/959/files#diff-8bd21e6750b7829a78940cd1d0405bfe8e4ba19e78acb53c085cb8cfba164d41R138

I call it prior to stopping the nodes (https://github.com/CasperLabs/casper-node/pull/959/files#diff-8bd21e6750b7829a78940cd1d0405bfe8e4ba19e78acb53c085cb8cfba164d41R29) and after we check that each individual node starts up (https://github.com/CasperLabs/casper-node/pull/959/files#diff-8bd21e6750b7829a78940cd1d0405bfe8e4ba19e78acb53c085cb8cfba164d41R51)

utils/nctl/sh/scenarios/itst02.sh
TomVasile: added a fuction to check across all of the nodes used in this test: https://github.com/CasperLabs/casper-node/pull/959/files#diff-8bd21e6750b7829a78940cd1d0405bfe8e4ba19e78acb53c085cb8cfba164d41R138

utils/nctl/sh/scenarios/itst02.sh
TomVasile: extended this to check that node1 and node 2 didnt progress and that their lfbs don't mismatch.

https://github.com/CasperLabs/casper-node/pull/959/files#diff-8bd21e6750b7829a78940cd1d0405bfe8e4ba19e78acb53c085cb8cfba164d41R82-R105

utils/nctl/sh/scenarios/itst02.sh
TomVasile: removed

utils/nctl/sh/scenarios/itst02.sh
TomVasile: Yea i debated removing it. I think it could go with the network sync taking its place

TomVasile: removed in: https://github.com/CasperLabs/casper-node/pull/959/commits/9b0e9d05529c9c5b7f9239cbccf8e56d89785f0b

PR #960, opened 2021-02-19
EE-1168: refactor Mint to remove use of named keys

henrytill:
Ref: https://casperlabs.atlassian.net/browse/EE-1168

This PR
- introduces a `Key::Balance` variant, under which balance values are stored.  These values are only readable and writable by host-side code.
- removes the various URefs used for indirection in the mint code.
- simplifies the structure of balance proofs now that the indirection is gone.

### DIFF
--- a/client/lib/validation.rs
+++ b/client/lib/validation.rs
@@ -187,10 +187,7 @@ pub(crate) fn validate_get_balance_response(
         .as_object()
         .ok_or(ValidateResponseError::ValidateResponseFailedToParse)?;
 
-    let (purse_proof, balance_proof): (
-        TrieMerkleProof<Key, StoredValue>,
-        TrieMerkleProof<Key, StoredValue>,
-    ) = {
+    let balance_proof: TrieMerkleProof<Key, StoredValue> = {
         let proof = object
             .get(GET_ITEM_RESULT_MERKLE_PROOF)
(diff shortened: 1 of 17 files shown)


dwerner (approved):

goral09:
> introduces a `Key::Balance` variant, under which balance values are stored. These values are only readable and writable by host-side code.

What would have to a dApp developer do to implement something similar to the Mint? Like an ERC20 token that tracks balances of all token holders?

dwerner:
This improves disk use substantially - in a 10 min run spamming with new accounts, we previously reached 1.4gb disk consumption, and with this change it's in the range of 272mb. https://casperlabs.atlassian.net/browse/NDRS-918

henrytill:
> > introduces a `Key::Balance` variant, under which balance values are stored. These values are only readable and writable by host-side code.
> 
> What would have to a dApp developer do to implement something similar to the Mint? Like an ERC20 token that tracks balances of all token holders?

We will need to reintroduce something like `Key::Local` (probably using an arbitrary `URef` as a seed to address versioning concerns).  @EdHastingsCasperLabs and I have been talking about it, but it's lower priority than some of the other optimization work ATM.

EdHastingsCasperLabs (approved):

mpapierski (approved):

PR #963, opened 2021-02-20
Update Readme

MParlikar:
After recent changes and observing community use of the node, we need to state that the node doesn't really run well by itself.


goral09:
> After recent changes and observing community use of the node, we need to state that the node doesn't really run well by itself.

I don't think it ever run fine by itself – it should detect that it's isolated and shut down.

goral09 (approved):

PR #964, opened 2021-02-20
NO TICKET increase number of buckets for contract_runtime metrics

dwerner:
Increase the number of buckets for our `contract_runtime` histograms. The old value resulted in values being skewed to 320ms exactly. I'm leaving a note that the last bucket before a similar cutoff would occur at 10.24 seconds as a result of our math here. I tested this in my environment and the metric appears to be fixed.

### DIFF
--- a/node/src/components/contract_runtime.rs
+++ b/node/src/components/contract_runtime.rs
@@ -90,10 +90,15 @@ pub struct ContractRuntimeMetrics {
 
 /// Value of upper bound of histogram.
 const EXPONENTIAL_BUCKET_START: f64 = 0.01;
+
 /// Multiplier of previous upper bound for next bound.
 const EXPONENTIAL_BUCKET_FACTOR: f64 = 2.0;
-/// Bucket count, with last going to +Inf.
-const EXPONENTIAL_BUCKET_COUNT: usize = 6;
+
+/// Bucket count, with the last bucket going to +Inf which will not be included in the results.
+/// - start = 0.01, factor = 2.0, count = 10
+/// - start * factor ^ count = 0.01 * 2.0 ^ 10 = 10.24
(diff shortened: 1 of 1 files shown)


node/src/components/contract_runtime.rs
```diff
@@ -90,10 +90,15 @@ pub struct ContractRuntimeMetrics {
...
 /// Multiplier of previous upper bound for next bound.
 const EXPONENTIAL_BUCKET_FACTOR: f64 = 2.0;
-/// Bucket count, with last going to +Inf.
-const EXPONENTIAL_BUCKET_COUNT: usize = 6;
+
+/// Bucket count, with the last bucket going to +Inf which will not be included in the results.
+/// - start = 0.01, factor = 2.0, count = 10
+/// - start * factor ^ count = 0.01 * 2.0 ^ 10 = 10.24
+/// - Values above 10.24 (f64 seconds here) will not fall in a bucket that is kept.
+const EXPONENTIAL_BUCKET_COUNT: usize = 10;
```
goral09, line 101: With `exp_bucket_count=6` the maximum bucket would have been 640ms, right? Weird that the maximum reading we had was 320ms?

goral09 (approved):

node/src/components/contract_runtime.rs
dwerner: It's mentioned that the last bucket is not used in the returned slice https://docs.rs/prometheus/0.11.0/prometheus/fn.exponential_buckets.html

goral09: Ah, thanks. Weird :)

dwerner: Yeah, I was banging my head against this trying to figure out why we were stuck at 320ms as well.

darthsiroftardis (approved):

PR #965, opened 2021-02-22
NO-TICKET: Fix `nctl` scripts.

goral09:
https://github.com/CasperLabs/casper-node/pull/957 broke `nctl` scripts – it used `staked_amount` where `bonded_amount` was [expected](https://github.com/CasperLabs/casper-node/blob/master/node/src/types/chainspec/accounts_config.rs#L24).


ghost (approved):

goral09:
bors merge

PR #971, opened 2021-02-22
NDRS-857/sample memory metrics in consensus

marc-casperlabs:
Reduces heap measuring of large collections by sampling only a subset of their contents.

### DIFF
--- a/node/src/utils/ds.rs
+++ b/node/src/utils/ds.rs
@@ -1,18 +1,96 @@
 //! Datasize helper functions.
 
-use std::mem::size_of;
+use std::{
+    collections::{HashMap, HashSet},
+    mem::size_of,
+};
+
+use datasize::DataSize;
+use rand::{
+    rngs::StdRng,
+    seq::{IteratorRandom, SliceRandom},
+    SeedableRng,
+};
+
+/// Number of items to sample when sampling a large collection.
+const SAMPLE_SIZE: usize = 100;
 
 /// Estimate memory usage of a hashmap of keys and values each with no heap allocations.
-pub fn hash_map_fixed_size<K, V>(hashmap: &std::collections::HashMap<K, V>) -> usize {
+pub fn hash_map_fixed_size<K, V>(hashmap: &HashMap<K, V>) -> usize {
     hashmap.capacity() * (size_of::<V>() + size_of::<K>() + size_of::<usize>())
 }
 
 /// Estimate memory usage of a hashset of items with no heap allocations.
-pub fn hash_set_fixed_size<T>(hashset: &std::collections::HashSet<T>) -> usize {
+pub fn hash_set_fixed_size<T>(hashset: &HashSet<T>) -> usize {
     hashset.capacity() * (size_of::<T>() + size_of::<usize>())
 }
 
 /// Estimate memory usage of a vec of items with no heap allocations.
 pub fn vec_fixed_size<T>(vec: &Vec<T>) -> usize {
     vec.capacity() * size_of::<T>()
+}
+
+/// Creates an RNG for sampling based on the length of a collection.
+fn sampling_rng(len: usize) -> StdRng {
+    // We use a fixed seed RNG here and hope the size will provide enough entropy to avoid gross
+    // misestimations. This has the added benefit that repeated measurements will consider the
+    // same nodes, reducing jitter and making this a pure function.
+
+    // Initialize a buffer suitable for the seed, which might be larger then our length bytes.
+    let mut seed = <StdRng as SeedableRng>::Seed::default();
+    let len_be = len.to_be_bytes();
+
+    // Mix in entropy from length.
+    for (b1, b2) in seed.iter_mut().zip(len_be.iter()) {
+        *b1 ^= *b2;
+    }
+
+    StdRng::from_seed(seed)
+}
+
+/// Given a length and a total of `sampled` bytes from sampling `SAMPLE_SIZE` items, return an
+/// estimate for the total heap memory consumption of the collection.
+fn scale_sample(len: usize, sampled: usize) -> usize {
+    let scaling_factor = len as f64 / SAMPLE_SIZE as f64;
+    (sampled as f64 * scaling_factor).round() as usize
+}
+
+/// Extrapolate memory usage of a Vec by from a random subset of `SAMPLE_SIZE` items.
+pub fn vec_sample<T>(vec: &Vec<T>) -> usize
+where
+    T: DataSize,
+{
+    if vec.len() < SAMPLE_SIZE {
+        vec.estimate_heap_size()
+    } else {
+        let mut rng = sampling_rng(vec.len());
+        let sampled = vec
+            .as_slice()
+            .choose_multiple(&mut rng, SAMPLE_SIZE)
+            .map(DataSize::estimate_heap_size)
+            .sum();
+        scale_sample(vec.len(), sampled)
+    }
+}
+
+/// Extrapolate memory usage of a HashMap by from a random subset of `SAMPLE_SIZE` items.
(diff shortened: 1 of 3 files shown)


node/src/utils/ds.rs
```diff
@@ -1,18 +1,96 @@
...
 /// Estimate memory usage of a hashmap of keys and values each with no heap allocations.
-pub fn hash_map_fixed_size<K, V>(hashmap: &std::collections::HashMap<K, V>) -> usize {
+pub fn hash_map_fixed_size<K, V>(hashmap: &HashMap<K, V>) -> usize {
     hashmap.capacity() * (size_of::<V>() + size_of::<K>() + size_of::<usize>())
 }
 
 /// Estimate memory usage of a hashset of items with no heap allocations.
-pub fn hash_set_fixed_size<T>(hashset: &std::collections::HashSet<T>) -> usize {
+pub fn hash_set_fixed_size<T>(hashset: &HashSet<T>) -> usize {
     hashset.capacity() * (size_of::<T>() + size_of::<usize>())
```
afck, line 25: The `size_of::<usize>()` was probably copied from `Vec` — for hash sets and maps it's more complicated, but maybe the overhead is negligible?

node/src/utils/ds.rs
```diff
@@ -0,0 +1,95 @@
...
+    vec.capacity() * size_of::<T>()
+}
+
+/// Creates an RNG for sampling based on the length of a collection.
+fn sampling_rng(len: usize) -> StdRng {
+    // We use a fixed seed RNG here and hope the size will provide enough entropy to avoid gross
+    // misestimations. This has the added benefit that repeated measurements will consider the
+    // same nodes, reducing jitter and making this a pure function.
+
+    // Initialize a buffer suitable for the seed, which might be larger then our length bytes.
```
afck, line 39:
```suggestion
    // Initialize a buffer suitable for the seed, which might be larger than our length bytes.
```

node/src/utils/ds.rs
```diff
@@ -0,0 +1,95 @@
...
+    StdRng::from_seed(seed)
+}
+
+/// Given a length and a total of `sampled` bytes from sampling `SAMPLE_SIZE` items, return an
+/// estimate for the total heap memory consumption of the collection.
+fn scale_sample(len: usize, sampled: usize) -> usize {
+    sampled * len / SAMPLE_SIZE
+}
+
+/// Extrapolate memory usage of a Vec by from a random subset of `SAMPLE_SIZE` items.
```
afck, line 57:
```suggestion
/// Extrapolate memory usage of a `Vec` by from a random subset of `SAMPLE_SIZE` items.
```

node/src/utils/ds.rs
```diff
@@ -0,0 +1,95 @@
...
+        let sampled = vec
+            .as_slice()
+            .choose_multiple(&mut rng, SAMPLE_SIZE)
+            .map(DataSize::estimate_heap_size)
+            .sum();
+        scale_sample(vec.len(), sampled)
+    }
+}
+
+/// Extrapolate memory usage of a HashMap by from a random subset of `SAMPLE_SIZE` items.
```
afck, line 75:
```suggestion
/// Extrapolate memory usage of a `HashMap` by from a random subset of `SAMPLE_SIZE` items.
```

node/src/utils/ds.rs
```diff
@@ -0,0 +1,95 @@
...
+    T: DataSize,
+{
+    if vec.len() < SAMPLE_SIZE {
+        vec.estimate_heap_size()
+    } else {
+        let mut rng = sampling_rng(vec.len());
+        let sampled = vec
+            .as_slice()
+            .choose_multiple(&mut rng, SAMPLE_SIZE)
+            .map(DataSize::estimate_heap_size)
```
afck, line 69: As discussed, let's make sure this estimates the `T` and not `&T`. (Also below.)

afck (commented):
As discussed I reviewed only https://github.com/CasperLabs/casper-node/pull/971/commits/ba270b3a0e0470e3aa3cac10b6c8e6e643644b91 and https://github.com/CasperLabs/casper-node/pull/971/commits/68561678c30448638429f0c2bc9a5b7a5d27a5aa. Looks good to me so far, but let's make sure we're not estimating `&T` instead of `T`. (See my comment.)

node/src/utils/ds.rs
marc-casperlabs: It has been copied from the estimate from `datasize` itself, unfortunately I did not look in details into the accuracy of these metrics, as that would involve very intricate knowledge of how those data structures are implemented.

Overall this is why `datasize` says "estimate" on the label.

node/src/utils/ds.rs
marc-casperlabs: Can do, I will add a unit test for this.

afck (approved):
Approving, assuming https://github.com/CasperLabs/casper-node/pull/969 gets approved and the `T` vs. `&T` tested.

node/src/utils/ds.rs
```diff
@@ -63,13 +63,15 @@ where
...
         vec.estimate_heap_size()
     } else {
+        let base_size = vec.capacity() * mem::size_of::<T>();
+
         let mut rng = sampling_rng(vec.len());
         let sampled = vec
             .as_slice()
             .choose_multiple(&mut rng, SAMPLE_SIZE)
-            .map(DataSize::estimate_heap_size)
+            .map(|v| v.estimate_heap_size())
```
afck, line 72: I thought that was exactly equivalent to `.map(DataSize::estimate_heap_size)`?? Don't you have to `(*v).estimate_heap_size()` or something?

node/src/utils/ds.rs
```diff
@@ -63,13 +63,15 @@ where
     if vec.len() < SAMPLE_SIZE {
         vec.estimate_heap_size()
     } else {
+        let base_size = vec.capacity() * mem::size_of::<T>();
```
afck, line 66: Right, good catch!

afck (approved):

node/src/utils/ds.rs
```diff
@@ -19,3 +29,134 @@ pub fn hash_set_fixed_size<T>(hashset: &HashSet<T>) -> usize {
...
+    fn hashmap_sample_above_sample_size() {
+        let num_items = SAMPLE_SIZE * 5;
+
+        // We make all items equal in size, so that we know the outcome of a random sampling.
+        let data: HashMap<usize, Vec<u32>> = (0..num_items)
+            .map(|idx| (idx, vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10]))
+            .collect();
+
+        assert_eq!(hashmap_sample(&data), data.estimate_heap_size());
+    }
```
goral09, line 161: Isn't the point of this "sampling" functions to provide a _good enough_ estimate where it's expensive to calculate the exact value? That's not the case for primitive types though, right?

afck: The exact calculation would iterate over all entries in the hash map, and sum up their sizes, whereas the estimate only takes the average of a sample, and multiplies it by the number of entries.

Even for a hash map of vectors, that should be faster than the full iteration.

goral09: I get that. My point is that this particular test is not very useful in a sense that it doesn't test what's been added here. It doesn't verify how many iterations we did so `hashmap_sample` could be doing exactly what you described.

afck: Yes, I think that test is just to verify that the new method is correct, not that it's faster. But I'll let @marc-casperlabs comment on that.

goral09: And it's trivial for the primitive types since its size is constant.

afck: But in this case it's not trivial because it's a map of _vectors_ of `u32`s, not a map of `u32`s.

PR #972, opened 2021-02-22
NDRS-880: Add an integration test for a node syncing with upgraded network.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-880
https://casperlabs.atlassian.net/browse/NDRS-849

### DIFF
--- a/utils/nctl/sh/scenarios/sync_upgrade_test.sh
+++ b/utils/nctl/sh/scenarios/sync_upgrade_test.sh
@@ -1,0 +1,220 @@
+#!/usr/bin/env bash
+
+source "$NCTL"/sh/utils/main.sh
+source "$NCTL"/sh/views/utils.sh
+source "$NCTL"/sh/assets/upgrade.sh
+source "$NCTL"/sh/node/svc_"$NCTL_DAEMON_TYPE".sh
+
+# Exit if any of the commands fail.
+set -e
+
+#######################################
+# Runs an integration tests that tries to sync a new node
+# with an upgraded network.
+#
+# Arguments:
+#   `node=XXX` ID of a new node. Default=6
+#   `timeout=XXX` timeout (in seconds) when syncing. Default=300 seconds.
+#   `version=X_Y_Z` new protocol version to upgrade to. Default=2_0_0.
+#   `era=X` at which the upgrade should take place. Default=3.
+#######################################
+function main() {
+    log "------------------------------------------------------------"
+    log "Syncing node begins"
+    log "------------------------------------------------------------"
+
+    do_await_genesis_era_to_complete
+
+    # 1. Send batch of Wasm deploys
+    do_send_wasm_deploys
+    # 2. Send batch of native transfers
+    do_send_transfers
+    # 3. Wait until they're all included in the chain.
+    do_await_deploy_inclusion
+    # 4. Upgrade the network
+    do_upgrade_network
+    # 5. Wait for the network to upgrade.
+    do_await_network_upgrade
+    # 6. Take a note of the last finalized block hash
+    do_read_lfb_hash
+    # 7. Send batch of Wasm deploys
+    do_send_wasm_deploys
+    # 8. Send batch of native transfers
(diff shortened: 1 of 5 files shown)


linclelinkpart5 (approved):
Looks good to me!

utils/nctl/sh/scenarios/sync_upgrade_test.sh
```diff
@@ -0,0 +1,220 @@
...
+}
+
+function do_upgrade_network() {
+    log_step "scheduling the network upgrade to version ${PROTOCOL_VERSION} at era ${ACTIVATE_ERA}"
+    for NODE_ID in $(seq 1 "$(get_count_of_nodes)"); do
+        _upgrade_node "$PROTOCOL_VERSION" "$ACTIVATE_ERA" "$NODE_ID"
+    done
+}
+
+function do_await_network_upgrade() {
```
TomVasile, line 79: maybe we should add a timeout or an error check here so we don't potentially wait forever in case of a bug

utils/nctl/sh/scenarios/sync_upgrade_test.sh
```diff
@@ -0,0 +1,220 @@
...
+}
+
+function do_await_deploy_inclusion() {
+    # Should be enough to await for one era.
+    log_step "awaiting one era…"
+    await_n_eras 1
+}
+
+function do_read_lfb_hash() {
+    local NODE_ID=${1}
```
TomVasile, line 121: argument is not used in line 39: https://github.com/CasperLabs/casper-node/pull/972/files?diff=split&w=1#diff-ec589c4feae2193851f12fd81a6ed2dc3f4be5de75d630910865dc4adf449bbeR39

utils/nctl/sh/scenarios/sync_upgrade_test.sh
```diff
@@ -0,0 +1,220 @@
...
+    local TRANSFERS_COUNT=5
+    local NODE_ID="random"
+
+    # Enumerate set of users.
+    for USER_ID in $(seq 1 "$(get_count_of_users)"); do
+        dispatch_native "$AMOUNT" "$USER_ID" "$TRANSFERS_COUNT" "$NODE_ID"
+    done
+}
+
+function do_await_deploy_inclusion() {
```
TomVasile, line 114: function name is a bit misleading since its not actually checking the deploys were included

utils/nctl/sh/scenarios/sync_upgrade_test.sh
goral09: Oh yeah, definitely a good idea :) especially that I hit this problem in my tests :sweat_smile:

utils/nctl/sh/scenarios/sync_upgrade_test.sh
goral09: I know but tracking the inclusion status of randomly generated deploys is going to be a pain. It's supposed to be tested with other WG-based tests. Here I just want to "pretend" they are included, which they should with nctl parameters. It's also not critical that they're all included. It's enough if some of them are. I could call `await_n_eras 1` but that wouldn't convey why I'm doing it. Hence this method.

utils/nctl/sh/scenarios/sync_upgrade_test.sh
goral09: That's intended. This method is used two-fold: one, it assigns the current LFB to the global `LFB_HASH` variable. Two, it returns it to the caller – which is used in `do_await_full_synchronization`.

utils/nctl/sh/scenarios/sync_upgrade_test.sh
goral09: Done in [`36f81a6` (#972)](https://github.com/CasperLabs/casper-node/pull/972/commits/36f81a604ef7657252e0912ae46bc115a5681649)

TomVasile (approved):

PR #978, opened 2021-02-23
NDRS-913: Push whole block to the event stream.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-913

https://github.com/CasperLabs/casper-node/pull/892 moved `deploy_hashes`, `transfer_hashes` and `proposer` from the `BlockHeader` structure to the new `BlockBody`. Up until now, we've been pushing only `BlockHeader` to the event stream, which means that since #892 got merged those three were missing from the event stream events. This PR changes what gets pushed to the event stream – we're now pushing the whole `Block`.

The main differences are in the structure of the event. Instead of `block_header` field containing `BlockHeader` structure, there's now a `block` field with `Block` struct. Structure of the `Block` can be found [here](https://github.com/CasperLabs/casper-node/blob/master/node/src/types/block.rs#L1023).

### DIFF
--- a/node/src/components/event_stream_server.rs
+++ b/node/src/components/event_stream_server.rs
@@ -91,12 +91,9 @@ where
         event: Self::Event,
     ) -> Effects<Self::Event> {
         match event {
-            Event::BlockAdded {
+            Event::BlockAdded { block_hash, block } => self.broadcast(SseData::BlockAdded {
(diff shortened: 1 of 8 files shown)


Fraser999 (approved):

goral09:
bors merge

PR #979, opened 2021-02-23
HWY-257: Clear the protocol state of unbonded eras.

afck:
All bonded eras are kept in memory because we still need to handle messages for them and detect equivocations.
However, since each era can refer to evidence from previous eras for cross-era slashing, even eras older than that are kept in memory, so that that evidence can be validated. This PR clears the protocol state of all unbonded eras and retains only evidence, to free up the memory. The dropped information is not needed anymore, since evidence cannot have any dependencies.
This should reduce consensus memory usage by almost 50%.

https://casperlabs.atlassian.net/browse/HWY-257

### DIFF
--- a/node/src/components/consensus/highway_core/state.rs
+++ b/node/src/components/consensus/highway_core/state.rs
@@ -890,6 +890,22 @@ impl<C: Context> State<C> {
         result
     }
 
+    /// Drops all state other than evidence.
+    pub(crate) fn retain_evidence_only(&mut self) {
+        self.units.clear();
+        self.blocks.clear();
+        self.panorama = self
+            .panorama
+            .iter()
+            .map(|obs| match obs {
+                Observation::Faulty => Observation::Faulty,
+                Observation::None | Observation::Correct(_) => Observation::None,
+            })
+            .collect();
+        self.endorsements.clear();
+        self.incomplete_endorsements.clear();
+    }
+
     /// Validates whether a unit with the given panorama and `endorsed` set satsifies the
     /// Limited Naïveté Criterion (LNC).
     /// Returns index of the first equivocator that was cited naively in violation of the LNC, or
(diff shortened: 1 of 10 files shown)


node/src/components/consensus/protocols/highway.rs
```diff
@@ -696,6 +701,15 @@ where
         self.highway.deactivate_validator()
     }
 
+    fn set_evidence_only(&mut self) {
+        // TODO: We could also drop the finality detector and round success meter here. Maybe make
+        // HighwayProtocol an enum with an EvidenceOnly variant?
```
goral09, line 706: We probably could, yeah.

afck: I created [NDRS-936](https://casperlabs.atlassian.net/browse/NDRS-936) for it.

node/src/components/consensus/highway_core/state.rs
```diff
@@ -890,6 +890,22 @@ impl<C: Context> State<C> {
...
+        self.units.clear();
+        self.blocks.clear();
+        self.panorama = self
+            .panorama
+            .iter()
+            .map(|obs| match obs {
+                Observation::Faulty => Observation::Faulty,
+                Observation::None | Observation::Correct(_) => Observation::None,
+            })
+            .collect();
```
xcthulhu, line 904: We could just mutate this in place, right?

xcthulhu (approved):

goral09 (approved):

node/src/components/consensus/highway_core/state.rs
afck: Done in https://github.com/CasperLabs/casper-node/pull/979/commits/ed679ab721e1af26566b8ee7e9b6125d267e4c8f.

PR #987, opened 2021-02-23
EE-1180: add Key::Bid

henrytill:
This PR adds a new `Bid` variant to the `Key` type, under which we will store values of type `Bid`.

This precedes upcoming changes to how bids are persisted in the auction module.


mpapierski (approved):

types/src/gens.rs
```diff
@@ -66,6 +66,7 @@ pub fn key_arb() -> impl Strategy<Value = Key> {
         u8_slice_32().prop_map(Key::Hash),
         uref_arb().prop_map(Key::URef),
         uref_arb().prop_map(|uref| Key::Balance(uref.addr())),
+        account_hash_arb().prop_map(Key::Bid),
```
xcthulhu, line 69: :heavy_check_mark:

xcthulhu (approved):
Looks like a minor conflict in `key.rs`, but this LGTM

EdHastingsCasperLabs (approved):

PR #989, opened 2021-02-23
EE-1181: add public key fields to auction::{Bid,Delegator}

henrytill:
This PR adds public key fields to `auction::{Bid, Delegator}`.  

In a future PR, bids will be stored under the account hash of the validator public key, so this field ensures that the public key remains associated with the bid.  It is added to `Delegator` for consistency.  We also rename `delegatee` to `validator_public_key`.


mpapierski (approved):

EdHastingsCasperLabs (approved):

henrytill:
Curses

henrytill:
bors retry

PR #990, opened 2021-02-24
EE-1182: simplify AccountHash::from_public_key

henrytill:
This PR removes the `blake2b_hash_fn` parameter from `Account::from_public_key` and hardcodes its former use site to `account::blake2b`.  This parameter was originally necessary to support system contracts, but now that we have removed system contract support, it is no longer necessary.

This precedes upcoming changes to how bids are persisted in the auction module.

### DIFF
--- a/execution_engine/src/core/runtime/auction_internal.rs
+++ b/execution_engine/src/core/runtime/auction_internal.rs
@@ -1,12 +1,11 @@
 use casper_types::{
-    account,
     account::AccountHash,
     bytesrepr::{FromBytes, ToBytes},
     system::auction::{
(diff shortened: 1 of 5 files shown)


mpapierski (commented):
There's still blake2b function available under wasm. Are we fine that now by default AccountHash::from_public_key used under a wasm contract would use blake2b compiled into wasm?

henrytill:
> There's still blake2b function available under wasm. Are we fine that now by default AccountHash::from_public_key used under a wasm contract would use blake2b crate compiled into wasm?

You are right, I hadn't considered this.  It's probably best that we leave as-is then.

PR #993, opened 2021-02-24
NDRS-940: Stop sync after very recent block.

goral09:
We used to mark syncing as done if none of the peers gave us any new block (successor of the latest block we downloaded). This proves to be problematic when we have many peers and they don't respond to us in time, before the next block is created. This way we end up being stuck in a syncing loop.

This PR considers newly downloaded block to be the tip of a chain if it was created less than 1 minute ago.

### DIFF
--- a/node/src/components/linear_chain_sync.rs
+++ b/node/src/components/linear_chain_sync.rs
@@ -283,7 +283,8 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
 
     /// Returns whether `block` can be considered tip of the chain.
     fn is_chain_end(&self, block: &Block) -> bool {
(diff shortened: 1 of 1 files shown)


afck:
I'd even make it a minute (or rather: one minimum round length). Five seconds doesn't sound realistic: It might take that long just to execute it, so by the time the peers send it it would already be older than that.

goral09:
I was thinking about using rounds here but this is highway-specific for now and `LinearChainSync` is generic. At least for now.

afck:
So should we just make it a minute for now?

goral09:
> So should we just make it a minute for now?

Already pushed.

afck (approved):

goral09:
bors merge

PR #994, opened 2021-02-24
NDRS-940: Do not return any effects after sync done.

goral09:
@fizyk20 found a bug in my code where it would replace the `self.state` with `curr_state` from before `mark_done()`.

We return `Effects::new()` (no effects) to stop the syncing process and let the reactor progress from joiner to validator.

### DIFF
--- a/node/src/components/linear_chain_sync.rs
+++ b/node/src/components/linear_chain_sync.rs
@@ -266,6 +266,7 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
                 );
                 if self.is_chain_end(&block) {
                     self.mark_done();
+                    return Effects::new()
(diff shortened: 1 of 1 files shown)


afck:
But wouldn't `State::Done` be more appropriate than leaving it as `State::None`, which was meant only as a temporary value, if I remember correctly? Not sure if it can cause any problems, though.

goral09:
> But wouldn't `State::Done` be more appropriate than leaving it as `State::None`, which was meant only as a temporary value, if I remember correctly? Not sure if it can cause any problems, though.

`mark_done()` sets `self.state` to `State::Done`.

afck (approved):

goral09:
bors merge

PR #997, opened 2021-02-24
NDRS-939: Add a delay before shutdown due to too many equivocations.

afck:
If the total weight of equivocators exceeds the configured fault tolerance threshold, consensus cannot continue and the node shuts down. This PR adds a delay for the shutdown so we can still broadcast evidence of the equivocations.

https://casperlabs.atlassian.net/browse/NDRS-939

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -742,6 +742,7 @@ where
                     "attempted to create a new era with a non-switch block header: {}",
                     block
                 )
+                .ignore()
             }
         };
         let newly_slashed = era_end.equivocators.clone();
@@ -1047,6 +1048,7 @@ where
                 inactive era = {}",
                 self.era_supervisor.current_era
             )
+            .ignore()
         } else {
             Default::default()
         }
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -1025,12 +1025,12 @@ where
                 .collect(),
             ProtocolOutcome::WeAreFaulty => Default::default(),
             ProtocolOutcome::DoppelgangerDetected => Default::default(),
-            // TODO: Replace the panic with the appropriate shutdown effect once that's possible.
-            ProtocolOutcome::FttExceeded => self
-                .effect_builder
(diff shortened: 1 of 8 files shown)


node/src/components/consensus/era_supervisor.rs
```diff
@@ -1019,6 +1024,12 @@ where
                 .collect(),
             ProtocolOutcome::WeAreFaulty => Default::default(),
             ProtocolOutcome::DoppelgangerDetected => Default::default(),
+            // TODO: Replace the panic with the appropriate shutdown effect once that's possible.
+            ProtocolOutcome::FttExceeded => self
+                .effect_builder
+                .set_timeout(Duration::from_millis(FTT_EXCEEDED_SHUTDOWN_DELAY_MILLIS))
+                .then(|_| async move { panic!("shutting down: too many faulty validators") })
+                .ignore(),
```
goral09, line 1032: Still, no solution to do it cleanly, huh?

goral09 (approved):

marc-casperlabs (changes requested):
Let's try to get this implemented without a direct panic first. I'll make a PR to this branch.

afck:
Thanks! I added your commits, @marc-casperlabs.

node/src/components/consensus/era_supervisor.rs
marc-casperlabs: Has been fixed now :)

node/src/effect.rs
```diff
@@ -1524,6 +1524,6 @@ impl<REv> EffectBuilder<REv> {
 #[macro_export]
 macro_rules! fatal {
     ($effect_builder:expr, $($arg:tt)*) => {
-        $effect_builder.fatal(file!(), line!(), format_args!($($arg)*).to_string()).ignore()
+        $effect_builder.fatal(file!(), line!(), format_args!($($arg)*).to_string())
```
goral09, line 1527: would we ever want to create an event that follows the call to `fatal!`? I thought that it's supposed to represent a stopping condition for the whole app.

afck: No, but if it contains the `ignore` it can't be chained with other futures.

marc-casperlabs: That's correct. `.ignore()` turns it from a generic future into an `Effect`. Not having `ignore` called right away makes composition a lot easier.

PR #1005, opened 2021-02-24
EE-1172: Delegators vesting schedule

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1138
Ref: https://casperlabs.atlassian.net/browse/EE-1172

This PR adds vesting schedule to genesis delegators which has similar behavior to vesting schedule of genesis validators. This PR also modifies genesis process which will create entries for each defined delegator in `accounts.toml`.

### DIFF
--- a/execution_engine/src/core/engine_state/genesis.rs
+++ b/execution_engine/src/core/engine_state/genesis.rs
@@ -586,6 +586,14 @@ pub enum GenesisError {
     ExecutionError(execution::Error),
     MintError(mint::Error),
     CLValue(String),
+    OrphanedDelegator {
+        validator_public_key: PublicKey,
+        delegator_public_key: PublicKey,
+    },
+    DuplicatedDelegatorEntry {
+        validator_public_key: PublicKey,
+        delegator_public_key: PublicKey,
+    },
 }
 
 pub(crate) struct GenesisInstaller<S>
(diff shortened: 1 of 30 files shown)


execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -620,38 +775,80 @@ where
...
-                        genesis_account.bonded_amount().value(),
-                    ))
-                } else {
-                    None
-                }
-            })
-            .collect();
+        let genesis_validators: Vec<_> = self.exec_config.get_bonded_validators().collect();
+
+        let genesis_delegators: Vec<_> = self.exec_config.get_bonded_delegators().collect();
```
henrytill, line 780: There's probably an opportunity to optimize to a single iteration here, but given that code is being run once at genesis, this seems acceptable for now.

types/src/system/auction/mod.rs
```diff
@@ -364,7 +366,9 @@ pub trait Auction:
         // Process bids
         let mut bids_modified = false;
         for (validator_public_key, bid) in bids.iter_mut() {
-            bids_modified = bid.process(era_end_timestamp_millis);
+            if bid.process(era_end_timestamp_millis) {
+                bids_modified = true;
+            }
```
henrytill, line 371: :laughing: I see you @mpapierski :laughing:

henrytill (approved):
Looking good.

PR #1010, opened 2021-02-25
NDRS-927: support flexible handshake for delta and upgraded delta nodes

Fraser999:

This PR re-introduces the relaxation of the network handshake message.  However to ensure it is compatible with 0.7.6 nodes currently running (which require the handshake to provide the chainspec hash), the handshake message has been updated to hold the chainspec hash and the network name.

The change is required in order to support future potential upgrades to the delta network.  For example, a new user would need to run this version of the node to initially join an upgraded delta network, and so the node must be able to communicate with nodes at a newer protocol version.  This would not be possible if we used the chainspec hash to handshake as the protocol version difference would be reflected in the hash, and the nodes would refuse to remain connected.

Nodes will now provide both pieces of data.  Old 0.7.6 nodes will be able to parse this message, silently ignoring the new `network_name` field.  New nodes will deserialize old nodes' handshakes such that the `network_name` is defaulted to an empty string.

On receipt of a handshake message, the nodes now favour checking the `network_name` and fall back to checking the `genesis_config_hash` only if the `network_name` is empty (indicating the handshake was sent from an old 0.7.6 node).

I tested via nctl that a network running this code connects correctly, and also that a single node running this code can remain connected to a network running the 0.7.6 nodes.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -158,8 +158,11 @@ where
     /// The interval between each fresh round of gossiping the node's public listening address.
     gossip_interval: Duration,
     /// The hash of the chainspec.  We only remain connected to peers with the same
-    /// `genesis_config_hash` as us.
+    /// `genesis_config_hash` or `network_name` as us.
     genesis_config_hash: Digest,
+    /// Name of the network we participate in.  We only remain connected to peers with the same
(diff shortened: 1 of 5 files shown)


Fraser999:
Closing to target the `release-0.8.0` branch instead.

PR #1011, opened 2021-02-25
NDRS-927: support flexible handshake for delta and upgraded delta nodes

Fraser999:

This PR re-introduces the relaxation of the network handshake message.  However to ensure it is compatible with 0.7.6 nodes currently running (which require the handshake to provide the chainspec hash), the handshake message has been updated to hold the chainspec hash and the network name.

The change is required in order to support future potential upgrades to the delta network.  For example, a new user would need to run this version of the node to initially join an upgraded delta network, and so the node must be able to communicate with nodes at a newer protocol version.  This would not be possible if we used the chainspec hash to handshake as the protocol version difference would be reflected in the hash, and the nodes would refuse to remain connected.

Nodes will now provide both pieces of data.  Old 0.7.6 nodes will be able to parse this message, silently ignoring the new `network_name` field.  New nodes will deserialize old nodes' handshakes such that the `network_name` is defaulted to an empty string.

On receipt of a handshake message, the nodes now favour checking the `network_name` and fall back to checking the `genesis_config_hash` only if the `network_name` is empty (indicating the handshake was sent from an old 0.7.6 node).

I tested via nctl that a network running this code connects correctly, and also that a single node running this code can remain connected to a network running the 0.7.6 nodes.

### DIFF
--- a/node/src/components/small_network/message.rs
+++ b/node/src/components/small_network/message.rs
@@ -2,16 +2,25 @@ use std::fmt::{self, Debug, Display, Formatter};
 
 use serde::{Deserialize, Serialize};
 
+use crate::crypto::hash::Digest;
+
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub enum Message<P> {
-    Handshake { network_name: String },
+    Handshake {
+        genesis_config_hash: Digest,
+        #[serde(default)]
+        network_name: String,
+    },
     Payload(P),
 }
 
 impl<P: Display> Display for Message<P> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
-            Message::Handshake { network_name } => write!(f, "handshake: {}", network_name),
+            Message::Handshake {
+                genesis_config_hash,
+                network_name,
+            } => write!(f, "handshake: {} -- {}", genesis_config_hash, network_name),
             Message::Payload(payload) => write!(f, "payload: {}", payload),
         }
     }
--- a/node/src/reactor/joiner.rs
+++ b/node/src/reactor/joiner.rs
@@ -408,12 +408,14 @@ impl reactor::Reactor for Reactor {
             chainspec_loader.chainspec(),
             false,
         )?;
+        let genesis_config_hash = chainspec_loader.chainspec().hash();
         let network_name = chainspec_loader.chainspec().network_config.name.clone();
         let (small_network, small_network_effects) = SmallNetwork::new(
             event_queue,
             config.network.clone(),
             registry,
             small_network_identity,
+            genesis_config_hash,
             network_name,
             false,
         )?;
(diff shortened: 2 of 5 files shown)


node/src/components/small_network/message.rs
```diff
@@ -2,16 +2,25 @@ use std::fmt::{self, Debug, Display, Formatter};
 
 use serde::{Deserialize, Serialize};
 
+use crate::crypto::hash::Digest;
+
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub enum Message<P> {
-    Handshake { network_name: String },
+    Handshake {
+        genesis_config_hash: Digest,
```
goral09, line 10: Should it be more generally – `chainspec_config_hash`? I guess that after an upgrade it's going to be a hash of the upgrade config, not the old Genesis one.

goral09: Ah, but the nodes running software with upgrade will never send us the config hash but the network name.

node/src/reactor/joiner.rs
```diff
@@ -408,12 +408,14 @@ impl reactor::Reactor for Reactor {
             chainspec_loader.chainspec(),
             false,
         )?;
+        let genesis_config_hash = chainspec_loader.chainspec().hash();
```
goral09, line 411: Not necessarily a Genesis thought. It could be an upgraded chainspec.

node/src/reactor/validator.rs
```diff
@@ -406,12 +406,14 @@ impl reactor::Reactor for Reactor {
             chainspec_loader.chainspec(),
             true,
         )?;
+        let genesis_config_hash = chainspec_loader.chainspec().hash();
```
goral09, line 409: Same here – not necessarily a Genesis.

node/src/components/small_network/message.rs
```diff
@@ -2,16 +2,25 @@ use std::fmt::{self, Debug, Display, Formatter};
...
+        #[serde(default)]
+        network_name: String,
```
goral09, line 12: In the description of this PR you say that the old delta node (running version `0.7.6`) will ignore the new field name `genesis_config_hash`. How does this happen? Programmatically I mean. What type of handshake would a `0.8.0` node send to `0.7.6`? It will include both _a_ config hash and a network name. How do `0.7.6` nodes know to ignore the network name?

goral09 (approved):
Nifty.

goral09:
Should we have a ticket to remove that? Or this is not going into `master`?

node/src/components/small_network/message.rs
Fraser999: Bincode's `deserialize()` function [sets the config option to allow trailing bytes](https://github.com/bincode-org/bincode/blob/trunk/src/lib.rs#L183) which means the 0.7.6 nodes just ignore the unused bytes comprising the network name.

It's a little fragile in that we need to ensure the fields we want ignored are at the end of the struct.  But this is temporary code for delta only I believe.

Fraser999:
bors merge

node/src/components/small_network/message.rs
goral09: Ha, nice hack! and it's not even in _our code_!

PR #1012, opened 2021-02-25
EE-1185: add TrackingCopy::get_keys

henrytill:
This PR adds `get_keys` method to TrackingCopy which reads through the cache to `StateReader::keys_with_prefix`.

The caching behavior is marginally different than how we handle key-value pairs: cached writes must be unioned with reads rather than replace them.

### DIFF
--- a/execution_engine/src/core/tracking_copy/mod.rs
+++ b/execution_engine/src/core/tracking_copy/mod.rs
@@ -5,7 +5,7 @@ pub(self) mod meter;
 mod tests;
 
 use std::{
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{BTreeSet, HashMap, HashSet, VecDeque},
     convert::{From, TryInto},
     iter,
 };
@@ -13,7 +13,7 @@ use std::{
 use linked_hash_map::LinkedHashMap;
 use thiserror::Error;
 
-use casper_types::{bytesrepr, CLType, CLValue, CLValueError, Key, U512};
+use casper_types::{bytesrepr, CLType, CLValue, CLValueError, Key, KeyTag, Tagged, U512};
 
 pub use self::ext::TrackingCopyExt;
 use self::meter::{heap_meter::HeapSize, Meter};
@@ -105,6 +105,8 @@ pub struct TrackingCopyCache<M> {
     current_cache_size: usize,
     reads_cached: LinkedHashMap<Key, StoredValue>,
     muts_cached: HashMap<Key, StoredValue>,
+    key_tag_reads_cached: LinkedHashMap<KeyTag, BTreeSet<Key>>,
+    key_tag_muts_cached: HashMap<KeyTag, BTreeSet<Key>>,
     meter: M,
 }
 
@@ -119,6 +121,8 @@ impl<M: Meter<Key, StoredValue>> TrackingCopyCache<M> {
             current_cache_size: 0,
             reads_cached: LinkedHashMap::new(),
             muts_cached: HashMap::new(),
+            key_tag_reads_cached: LinkedHashMap::new(),
+            key_tag_muts_cached: HashMap::new(),
             meter,
         }
     }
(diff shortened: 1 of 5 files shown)


types/src/key.rs
```diff
@@ -350,37 +377,30 @@ impl From<ContractPackageHash> for Key {
 impl ToBytes for Key {
     fn to_bytes(&self) -> Result<Vec<u8>, Error> {
         let mut result = bytesrepr::unchecked_allocate_buffer(self);
+        result.push(self.tag());
```
xcthulhu, line 380: From CI:

```
test storage::trie::tests::proptests::bincode_roundtrip_key ... ok
```

:heavy_check_mark:

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -231,6 +266,29 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
...
+            None => {
+                let key_tag = key_tag.to_owned();
+                let keys = self
+                    .reader
+                    .keys_with_prefix(correlation_id, &[key_tag as u8])?;
+                ret.extend(keys);
+                self.cache.insert_key_tag_read(key_tag, ret.to_owned())
+            }
+        }
+        if let Some(keys) = self.cache.get_key_tag_muts_cached(&key_tag) {
```
xcthulhu, line 286: I am a bit confused here should `key_tag_reads_cached` and `key_tag_muts_cached` be disjoint?  Or is the idea that `key_tag_muts_cached` masks `key_tag_reads_cached` and they can have duplicates?

xcthulhu: I see now; that masking happens elsewhere.  This is for key enumeration; if you mutated a key these sets will contain duplicates.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -231,6 +266,29 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
...
+            Some(keys) => ret.extend(keys),
+            None => {
+                let key_tag = key_tag.to_owned();
+                let keys = self
+                    .reader
+                    .keys_with_prefix(correlation_id, &[key_tag as u8])?;
+                ret.extend(keys);
+                self.cache.insert_key_tag_read(key_tag, ret.to_owned())
+            }
+        }
```
xcthulhu, line 285: You could do something like:

```rust
        // Do a cached read of the keys under a tag
        let key_tag_reads = match self.cache.get_key_tag_reads_cached(&key_tag) {
            Some(keys) => keys,
            None => {
                let key_tag = key_tag.to_owned();
                let keys = self
                    .reader
                    .keys_with_prefix(correlation_id, &[key_tag as u8])?;
                self.cache.insert_key_tag_read(key_tag, keys.to_owned())
                keys
            }
        }
        ret.extend(key_tag_reads);
```

xcthulhu:
LGTM

xcthulhu (approved):

execution_engine/src/core/tracking_copy/mod.rs
henrytill: yah

execution_engine/src/core/tracking_copy/mod.rs
henrytill: Good one, will do.

henrytill: Actually some impediments:
- currently `StateReader::keys_with_prefix` returns a `Vec` of keys.  I could change this to a `BTreeSet`, but...
- there's an ownership problem: `TrackingCopyCache::get_key_tags_reads_cached` returns a reference to the set but `StateReader::keys_with_prefix` returns the set itself.  The latter is unavoidable, but I don't want to have to clone if we change the former.

I may still change `StateReader::keys_with_prefix` to return a `BTreeSet` for consistency, but will likely do in a follow-up PR.

xcthulhu: Sounds good.

Fraser999 (approved):

EdHastingsCasperLabs (approved):

PR #1016, opened 2021-02-25
NDRS-942: Delegation rate for validators in accounts.toml

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/NDRS-942

This PR changes the structure of accounts.toml to separate `bonded_amount` so `delegation_rate` could be a part of it. If `[accounts.validator]` field in TOML is not specified for given array element, then such account would be normal user without stake. Otherwise it will be considered an account with a stake.

Thanks to this separation of stake related attributes both `bonded_amount` and `delegation_rate` can be properly validated.


execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -796,13 +883,28 @@ where
         let validators = {
             let mut validators = Bids::new();
 
-            for genesis_validator in genesis_validators {
-                let public_key = genesis_validator.public_key();
-                let staked_amount = genesis_validator.staked_amount();
+            for genesis_account in genesis_validators {
```
EdHastingsCasperLabs, line 886: Maybe `genesis_validators` should be renamed `genesis_accounts` ?

EdHastingsCasperLabs (approved):

PR #1018, opened 2021-02-26
EE-1186: persist bids directly in state trie

henrytill:
This PR removes the `Bids` map stored under the auction's `BIDS_KEY` named key and instead persists bids directly in the state trie under `Key::Bid` keys.  It also adds `RpcRequest::GetBids` (and related) in service of `get-auction-info`.

### DIFF
--- a/types/src/execution_result.rs
+++ b/types/src/execution_result.rs
@@ -9,6 +9,7 @@
 
 use alloc::{
     format,
+    prelude::v1::Box,
     string::{String, ToString},
     vec,
     vec::Vec,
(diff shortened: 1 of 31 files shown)


EdHastingsCasperLabs (approved):

types/src/system/auction/detail.rs
```diff
@@ -270,21 +292,27 @@ pub fn reinvest_delegator_rewards(
...
-pub fn reinvest_validator_reward(
-    bids: &mut Bids,
+pub fn reinvest_validator_reward<P>(
+    provider: &mut P,
     seigniorage_allocations: &mut Vec<SeigniorageAllocation>,
     validator_public_key: PublicKey,
     amount: U512,
-) -> Result<Option<URef>, Error> {
-    let bid = match bids.get_mut(&validator_public_key) {
+) -> Result<Option<URef>, Error>
```
mpapierski, line 306: No need for Option anymore

mpapierski (approved):

types/src/execution_result.rs
```diff
@@ -9,6 +9,7 @@
 
 use alloc::{
     format,
+    prelude::v1::Box,
```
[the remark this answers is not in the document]
henrytill: Fixed

types/src/system/auction/detail.rs
henrytill: Fixed

PR #1021, opened 2021-02-26
NO-TICKET: fix issue in gossip table

Fraser999:
This PR fixes an issue in the gossip table whereby gossip items could avoid being moved to the `finished` or `paused` state, hence leaving the item in the `current` state indefinitely, blocking any further attempts to handle gossip for that item.

We have a termination condition of knowing about `x` peers which hold the data in question, `x` being configurable and set to 15 currently.  We could reach this termination condition in a few different ways:
1. by receiving a response from a peer we gossiped the item to
2. by timing out a response we never got from a peer we gossiped the item to
3. by receiving the item as a gossip request from a peer

The issue is that we currently only move an item from `current` to `finished` or `paused` when handling situation 1.  Situation 2 was assumed to be unneeded as that should trigger a fresh attempt to gossip to a different peer, and _that_ response would eventually be received or a different peer chosen, and so on.

However, if in the meantime we received enough gossip requests from peers to reach the termination condition (situation 3), then when the timeout was triggered, we would recognise that we should no longer gossip this item, and no further gossip request was sent.  This broke the assumption that we would eventually handle a response from a peer.

To fix this, I have updated the logic to move an item to the `finished` state from `current` (or `paused`) when we reach the termination condition in all three of these scenarios.

### DIFF
--- a/node/src/components/gossiper/gossip_table.rs
+++ b/node/src/components/gossiper/gossip_table.rs
@@ -1,7 +1,7 @@
 #[cfg(not(test))]
 use std::time::Instant;
 use std::{
-    collections::{hash_map::Entry, HashMap, HashSet},
+    collections::{HashMap, HashSet},
     fmt::Display,
     hash::Hash,
     time::Duration,
@@ -10,7 +10,7 @@ use std::{
 use datasize::DataSize;
 #[cfg(test)]
 use fake_instant::FakeClock as Instant;
-use tracing::warn;
+use tracing::{error, warn};
 
 use super::Config;
 #[cfg(test)]
@@ -209,25 +209,25 @@ impl<T: Copy + Eq + Hash + Display> GossipTable<T> {
             return GossipAction::Noop;
         }
 
-        if let Some(state) = self.paused.get_mut(data_id) {
-            let _ = state.holders.insert(holder);
-            return GossipAction::Noop;
+        let update = |state: &mut State| {
+            let _ = state.holders.insert(holder.clone());
+        };
+
+        if let Some(action) = self.update_current(data_id, update) {
+            return action;
         }
 
-        match self.current.entry(*data_id) {
-            Entry::Occupied(mut entry) => {
-                let is_new = false;
-                let state = entry.get_mut();
-                let _ = state.holders.insert(holder);
-                state.action(self.infection_target, self.holders_limit, is_new)
-            }
-            Entry::Vacant(entry) => {
-                let is_new = true;
-                let state = entry.insert(State::default());
-                let _ = state.holders.insert(holder);
-                state.action(self.infection_target, self.holders_limit, is_new)
-            }
+        if let Some(action) = self.update_paused(data_id, update) {
+            return action;
         }
+
+        // This isn't in finished, current, or paused - add a new entry to current.
+        let mut state = State::default();
+        update(&mut state);
+        let is_new = true;
+        let action = state.action(self.infection_target, self.holders_limit, is_new);
+        let _ = self.current.insert(*data_id, state);
+        action
     }
 
     /// We received or generated potentially new data with given ID.  If received from a peer,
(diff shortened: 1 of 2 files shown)


EdHastingsCasperLabs (approved):

node/src/components/gossiper/gossip_table.rs
```diff
@@ -209,25 +209,25 @@ impl<T: Copy + Eq + Hash + Display> GossipTable<T> {
             return GossipAction::Noop;
         }
 
-        if let Some(state) = self.paused.get_mut(data_id) {
-            let _ = state.holders.insert(holder);
-            return GossipAction::Noop;
+        let update = |state: &mut State| {
+            let _ = state.holders.insert(holder.clone());
+        };
```
goral09, line 214: Since `state` will come from the methods that we pass the `update` to, couldn't we also pass in the `holder` and get rid of this lambda?

Fraser999: It's more like we're doing different updates depending upon where the call is being made, so we need to pass the closure to the `self.update_current()` or `self.update_paused()` since we're not always making the same changes to the state found in `current` or `paused`.

Fraser999:
bors merge

node/src/components/gossiper/gossip_table.rs
goral09: NVM, I read the rest of the code.

node/src/components/gossiper/gossip_table.rs
```diff
@@ -209,25 +209,25 @@ impl<T: Copy + Eq + Hash + Display> GossipTable<T> {
...
+            return action;
         }
+
+        // This isn't in finished, current, or paused - add a new entry to current.
+        let mut state = State::default();
+        update(&mut state);
+        let is_new = true;
+        let action = state.action(self.infection_target, self.holders_limit, is_new);
+        let _ = self.current.insert(*data_id, state);
+        action
```
goral09, line 230: Looks like this could be extracted to a method `insert_new_item(data_id) -> Action`

node/src/components/gossiper/gossip_table.rs
```diff
@@ -250,37 +250,34 @@ impl<T: Copy + Eq + Hash + Display> GossipTable<T> {
...
+        if let Some(action) = self.update_paused(data_id, update) {
+            return convert_action(action);
         }
+
+        // This isn't in finished, current, or paused - add a new entry to current.
+        let mut state = State::default();
+        update(&mut state);
+        let is_new = true;
+        let action = state.action(self.infection_target, self.holders_limit, is_new);
+        let _ = self.current.insert(*data_id, state);
```
goral09, line 279: We could reuse `insert_new_data` function here.

node/src/components/gossiper/gossip_table.rs
Fraser999: I was in two minds about it myself, but we'd also need to pass in the `update` closure or similar since we don't do the same thing when calling this inside `new_partial_data` vs `new_complete_data` (`new_complete_data` also sets `state.held_by_us = true;`)

I do like DRY, but in this case I think it'd actually make the code _less_ readable.  If you feel strongly though, I'll make the change - I'm not too fussed either way.

node/src/components/gossiper/gossip_table.rs
```diff
@@ -704,6 +716,54 @@ mod tests {
...
+
+        let limit = EXPECTED_DEFAULT_HOLDERS_LIMIT - 1;
+        for node_id in node_ids.iter().take(limit) {
+            let _ = gossip_table.new_complete_data(&data_id, Some(node_id.clone()));
+            assert!(!gossip_table.finished.contains(&data_id));
+        }
+
+        let _ =
+            gossip_table.check_timeout(&data_id, node_ids[EXPECTED_DEFAULT_HOLDERS_LIMIT].clone());
+        assert!(gossip_table.finished.contains(&data_id));
```
goral09, line 764: Can you add comments what you're doing and what you expect?

node/src/components/gossiper/gossip_table.rs
```diff
@@ -704,6 +716,54 @@ mod tests {
...
+    }
```
goral09, line 765: I don't understand why this unit test is supposed to pass. It says that it expects entries to be moved to `finished` b/c of the timeout but I don't see the timeout being set here anywhere. I don't see that we're triggering the timeout to happen etc.

goral09 (commented):
I'm glad this works but I find the approach hard to follow. The non-linearity of `update_XXX` methods makes it difficult to understand. There are many early returns (even without effects/values being returned) and state mutation inside lambda closures. I wouldn't be able to debug it myself. 

Personally, I prefer an approach that would read like this:
```
new_complete_data(data_id, holder) {
  if is_finished(data_id) {
     return NoOp
  }
  
  if is_new(data_id) {
     action = add_new_item(data_id, holder)
     return action
  }
  
  // otherwise update   
  update(data_id, holder)
}
```

node/src/components/gossiper/gossip_table.rs
Fraser999: Yeah, the gossip table is controlled entirely by the gossiper component.  The gossiper deals with setting timeouts and asking the network to send messages, etc.

The gossiper would call `gossip_table.check_timeout` when its timeout triggered, so that's why there's no actual timeouts in the gossip table tests.

node/src/components/gossiper/gossip_table.rs
goral09: if we pass in `holder: Option<NodeId>`, this should be enough to figure out if the data comes from a client and is held by us.

PR #1027, opened 2021-02-26
NDRS-899: Set ulimit

marc-casperlabs:
To address issues with networks with > 400 connections, set the open files limit (`ulimit -n`) at the startup.

If this fails, settle for the closest possible maximum. This code will never panic or abort, but warn if the limit cannot be increased.

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -54,6 +54,7 @@ use signal_hook::consts::signal::{SIGINT, SIGQUIT, SIGTERM};
 use tokio::time::{Duration, Instant};
 use tracing::{debug, debug_span, error, info, trace, warn};
 use tracing_futures::Instrument;
+use utils::rlimit::{Limit, OpenFiles, ResourceLimit};
 
 use crate::{
     effect::{Effect, EffectBuilder, Effects},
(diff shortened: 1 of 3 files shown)


EdHastingsCasperLabs (approved):

node/src/reactor.rs
```diff
@@ -81,6 +82,9 @@ static MEM_DUMP_THRESHOLD_MB: Lazy<Option<u64>> = Lazy::new(|| {
         .ok()
 });
 
+/// The desired limit for open files.
+const TARGET_OPEN_FILES_LIMIT: Limit = 64_000;
```
goral09, line 86: We should probably make it configurable via an env variable.

node/src/utils/rlimit.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    #[test]
+    fn get_and_reset_ulimit() {
+        // Retrieve limit and set the exact same limit again.
+        let limit = ResourceLimit::<OpenFiles>::get().expect("could not retrieve initial limit");
+        println!("{:?}", limit);
+
+        // Note: We could change it to something different, but we do not want to risk influencing
+        // other tests, so this is safest.
+        limit.set().expect("could not re-set limit");
```
goral09, line 143: We could always change it back to the original limit? Cargo tests run sequentially, right?

goral09 (commented):
Really educational PR. I will re-read and try to learn from it :)

PR #1033, opened 2021-03-01
NDRS-928/929: Introducing fetchers for block headers

xcthulhu:


node/src/components/storage.rs
```diff
@@ -615,6 +620,15 @@ impl Storage {
                     Some(signatures) => signatures,
                     None => BlockSignatures::new(block_hash, block.header().era_id()),
                 };
+                for (public_key, signature) in signatures.proofs.iter() {
```
Fraser999, line 623: Isn't there a PR to do this on session start?  Seems like overkill to be validating signatures every time we read a block.

If we really need it here, don't we also need it on the other block and deploy getters?

node/src/reactor/validator.rs
```diff
@@ -761,6 +841,10 @@ impl reactor::Reactor for Reactor {
                             warn!("received get request for gossiped-address from {}", sender);
                             return Effects::new();
                         }
+                        Tag::BlockHeaderByHash => todo!("Handle GET BlockHeaderByHash response"),
+                        Tag::BlockHeaderAndFinalitySignaturesByHeight => {
+                            todo!("Handle GET BlockHeaderByHeight response")
+                        }
```
Fraser999, line 847: We need to avoid `todo!`s here.  A malicious peer could trigger this easily.

node/src/types/block.rs
```diff
@@ -848,6 +848,18 @@ impl FromBytes for BlockHeader {
...
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct BlockHeaderAndFinalitySignatures {
+    pub block_header: BlockHeader,
+    pub finality_signatures: BTreeMap<PublicKey, Signature>,
+}
+
+impl Display for BlockHeaderAndFinalitySignatures {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
```
Fraser999, line 859: Outside of EE, our `Display` impls are generally not as verbose as `Debug` impls.  All of the components should have `Display` impls, you just need to decide whether you need to iterate the signatures to display that info, or whether it's reasonable to just print e.g. the length of the map as per `impl Display for BlockSignatures`.

Fraser999 (changes requested):
I think there should be unit tests added for the new `pub` storage methods.

That and the removal of the `todo!`s are the only blockers.

node/src/reactor/validator.rs
```diff
@@ -736,6 +736,86 @@ impl reactor::Reactor for Reactor {
...
+                                        Err(error) => {
+                                            error!("failed to create get-response: {}", error);
+                                            return Effects::new();
+                                        }
+                                    };
+                                }
+                                Ok(None) => {
+                                    debug!("failed to get {} for {}", block_hash, sender);
+                                    return Effects::new();
+                                }
```
goral09, line 768: Not in this PR but I think we should be signaling to the caller that we don't have the item, instead of letting the peer time out. I think it will have a performance improvement on the network.

node/src/components/storage.rs
```diff
@@ -695,6 +709,50 @@ impl Storage {
...
+            .map_or_else(BTreeMap::new, |block_signatures| block_signatures.proofs);
+        for (public_key, signature) in finality_signatures.iter() {
+            FinalitySignature::verify_bare_finality_signature(
+                &block_hash,
+                &block_header.era_id(),
+                signature,
+                public_key,
+            )
+            .map_err(Error::CryptographicSignatureVerificationError)?
+        }
```
goral09, line 737: I agree with @Fraser999 here – we already validate every finality signature on startup and we verify them before storing.

node/src/types/block.rs
goral09: Especially the part about the length of entries. This can very easily lead to blowing up the logs into gargantuan size.

node/src/components/storage.rs
xcthulhu: Let's keep this check in even if it is redundant to other checks.

node/src/reactor/validator.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1033/commits/c8a1df3e5747eb7b2ffda10af05500651ab8a206

node/src/types/block.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1033/commits/c3c5dd7449cfde784ed20a5345eecd5817aa2d1c

node/src/components/storage.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1033/commits/6b316c6752128c2fd6df8bf7bf585d680874f477

node/src/components/storage.rs
xcthulhu: Okay, having discussed this we'll get rid of this check.

Fixed in https://github.com/CasperLabs/casper-node/pull/1033/commits/6b316c6752128c2fd6df8bf7bf585d680874f477

goral09 (approved):

xcthulhu:
bors merge

Fraser999 (approved):

xcthulhu:
bors merge

xcthulhu:
bors retry

xcthulhu:
bors merge

PR #1034, opened 2021-03-01
NDRS-931: clean shutdown on fatal

marc-casperlabs:
This change introduces a special announcement called `ControlAnnouncement`, which can be used to propagate requests for runtime changes upwards.

This eliminates the need to `panic!` and makes it possible to shut down the reactor cleanly on a `fatal` error, instead of panicking.


node/src/components/network/tests_bulk_gossip.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unreachable_code)] // TODO: Figure out why this warning triggers.
```
afck, line 1: Maybe a forgotten `#[cfg(test)]` on this module?

node/src/effect/announcements.rs
```diff
@@ -25,6 +25,40 @@ use crate::{
     utils::Source,
 };
 
+/// Control announcements are special announcements handled directly by the runtime/runner.
+///
+/// Reactors are never passed control announcements back in and every reactor event must be able
+/// to be constructed from a `ControlAnnouncement` to be run.
+///
+/// Control announcements also skip the message queued and are given priority over any other effect,
```
afck, line 33:
```suggestion
/// Control announcements also skip the message queue and are given priority over any other effect,
```

node/src/reactor.rs
```diff
@@ -551,7 +577,22 @@ where
 
         // Dispatch the event, then execute the resulting effect.
         let start = self.clock.start();
-        let effects = self.reactor.dispatch_event(effect_builder, rng, event);
+
+        let (effects, keep_going) = if let Some(ctrl_ann) = event.as_control() {
+            // We've received a control event, which will _not_ be handled by the reactor.
+            match ctrl_ann {
+                ControlAnnouncement::FatalError { file, line, msg } => {
```
afck, line 584: I wonder whether it should be called `ShutdownAnnouncement` or `AbortAnnouncement` instead, since it always makes us exit?

(And whether it should be a wrapper type instead, _containing_ an event, so that `dispatch_event` doesn't need an unhandled `match` arm? Not sure how ugly that would be.)

node/src/reactor.rs
```diff
@@ -679,13 +721,15 @@ where
...
+                        break ReactorExit::ProcessShouldExit(ExitCode::Abort);
                     }
-                    self.crank(rng).await;
                 }
-                SIGINT => return ReactorExit::ProcessShouldExit(ExitCode::SigInt),
-                SIGQUIT => return ReactorExit::ProcessShouldExit(ExitCode::SigQuit),
-                SIGTERM => return ReactorExit::ProcessShouldExit(ExitCode::SigTerm),
+                SIGINT => break ReactorExit::ProcessShouldExit(ExitCode::SigInt),
+                SIGQUIT => break ReactorExit::ProcessShouldExit(ExitCode::SigQuit),
+                SIGTERM => break ReactorExit::ProcessShouldExit(ExitCode::SigTerm),
```
afck, line 732: Using `break` instead of `return` here doesn't change anything, does it?

node/src/effect/announcements.rs
```diff
@@ -25,6 +25,40 @@ use crate::{
...
+/// this is to ensure that a component that reports a fatal error is not given any more events to
```
Fraser999, line 34:
```suggestion
/// Control announcements also skip the message queues and are given priority over any other effect
/// to ensure that a component that reports a fatal error is not given any more events to
```

node/src/reactor.rs
```diff
@@ -212,6 +212,30 @@ pub trait Reactor: Sized {
...
+    ///
+    /// Returns a reference to a wrapped
+    /// [`ControlAnnouncement`](`crate::effect::announcements::ControlAnnouncement`) if the event
+    /// is indeed a control announcement variant.
+    fn as_control(&self) -> Option<&ControlAnnouncement>;
+}
+
+// Special implementation for the unit type, used in unit tests.
+// TODO: Refactor unit-testing and get rid of this implementation.
+impl ReactorEvent for () {
```
Fraser999, line 227: Let's feature-gate this then please if it's only for test.  Same for the `From` impl below.

node/src/reactor/joiner.rs
```diff
@@ -195,6 +196,10 @@ pub enum Event {
     StateStoreRequest(#[serde(skip_serializing)] StateStoreRequest),
 
     // Announcements
+    /// A control announcement.
```
afck, line 199:
```suggestion

    /// A control announcement.
```

afck (approved):

node/src/reactor/initializer.rs
```diff
@@ -233,6 +249,7 @@ impl reactor::Reactor for Reactor {
             Event::StateStoreRequest(request) => {
                 self.dispatch_event(effect_builder, rng, Event::Storage(request.into()))
             }
+            Event::ControlAnnouncement(_) => unreachable!("unhandled control announcement"),
```
Fraser999, line 252: Probably worth including the actual announcement in the panic.  Same comment applies to joiner and validator reactors.

This is probably a case for a `log_or_panic`, but since we haven't agreed an approach there, I guess we can live with panicking code here.  It seems low risk... but then again, don't they always!

Fraser999 (commented):
No real blockers from me.

node/src/reactor.rs
marc-casperlabs: No, I just wanted to be consistent and I think `break` is a little bit cleaner.

node/src/reactor.rs
marc-casperlabs: This is actually in preparation for having more than just the `FatalError` as a control announcement in the future - regular exits or some other data shall also be part of this.

Since events targeting the runner/runtime need to be handled in a special manner, grouping them inside a single enum ensure this extra logic only has to be implemented once.

node/src/components/network/tests_bulk_gossip.rs
marc-casperlabs: Unfortunately no :(

```
#[cfg(test)]
mod tests_bulk_gossip;
```

in `network.rs`.

node/src/reactor/initializer.rs
marc-casperlabs: The only reason I am not including this right now is because I want to be consistent with what the reactor macro does. Let me check.

marc-casperlabs: Okay, the macro also does it. I'll update the code :)

marc-casperlabs: > This is probably a case for a log_or_panic, but since we haven't agreed an approach there, I guess we can live with panicking code here. It seems low risk... but then again, don't they always!

Well, I have to point out that currently we are only handling `fatal` as a control announcement, so I am going out on a limb here in saying this in this case its probably for the best. :)

Fraser999 (approved):

node/src/effect/announcements.rs
```diff
@@ -27,12 +27,12 @@ use crate::{
...
-/// Reactors are never passed control announcements back in and every reactor event must be able
-/// to be constructed from a `ControlAnnouncement` to be run.
+/// Reactors are never passed control announcements back in and every reactor event must be able to
+/// be constructed from a `ControlAnnouncement` to be run.
 ///
-/// Control announcements also skip the message queue and are given priority over any other effect,
-/// this is to ensure that a component that reports a fatal error is not given any more events to
-/// handle afterwards.
+/// Control announcements also use a priority queue to ensure that a component that reports a fatal
+/// error is given as few follow up events as possible. However, there currently is no guarantee
```
afck, line 34:
```suggestion
/// error is given as few follow-up events as possible. However, there currently is no guarantee
```

node_macros/src/gen.rs
```diff
@@ -234,7 +234,7 @@ pub(crate) fn generate_reactor_impl(def: &ReactorDefinition) -> TokenStream {
                 dispatches.push(quote!(
                     #event_ident::#request_variant_ident(request) => {
                         // Request is discarded.
-                        panic!("received request that was expressively marked panic: {:?}", request)
+                        panic!("received event that was explicitly route to a panic: {:?}", request)
```
afck, line 237:
```suggestion
                        panic!("received event that was explicitly routed to a panic: {:?}", request)
```

afck (approved):

goral09:
@marc-casperlabs can we merge this?

goral09:
@marc-casperlabs Can we merge this?

PR #1036, opened 2021-03-01
NO-TICKET: Participation log message: separate inactive from faulty

afck:
Log inactive and faulty validators in separate lists.


goral09 (approved):
Thanks

PR #1039, opened 2021-03-01
NDRS-927: work around blocks not holding protocol version

Fraser999:

This PR is a workaround for the lack of protocol version in block headers.  Currently, when we restart after an upgrade from the delta network where the upgrade was a hard-reset, the joining node will start with a cached/stored version > 1.0.0.  When the second binary is started, this causes it to attempt to call the EE `commit_upgrade` passing the same version for the current and new protocols.

To get round this, we simply ignore the cached/stored version when calling `commit_upgrade`.  Instead we use a hard-coded `1.0.0` for the current version and the one provided in the chainspec as the new version.  This will need extended if we end up running 0.8.0 nodes on delta in order to be able to upgrade from those.  Hopefully this will not be needed, and it certainly won't be needed in master where we can avoid caching/storing the protocol version in the chainspec loader as it will be available in the block headers.

### DIFF
--- a/node/src/components/chainspec_loader.rs
+++ b/node/src/components/chainspec_loader.rs
@@ -373,8 +373,25 @@ impl ChainspecLoader {
             if highest_block.header().is_switch_block() {
                 // This is a valid run immediately after upgrading the node version.
                 trace!("valid run immediately after upgrade");
+
+                // TODO - remove this once we have protocol versions in block headers.
+                //
+                // It is required if we restart after the network has had a hard reset upgrade from
+                // delta currently running at v1.0.0.  In that case, the cached version could well
+                // be the same as that in the current chainspec.  However, we need to pass the
+                // previous version when calling `ContractRuntime::upgrade`.
+                //
+                // If we end up running the release-0.8.0, we'll need to update this workaround to
+                // allow for upgrading from 0.8.0.
+                let v1 = Version::new(1, 0, 0);
+                let previous_protocol_version = if cached_protocol_version > v1 {
+                    v1
+                } else {
+                    cached_protocol_version
+                };
+
                 let upgrade_config =
-                    self.new_upgrade_config(&highest_block, cached_protocol_version);
+                    self.new_upgrade_config(&highest_block, previous_protocol_version);
                 effects.extend(
                     effect_builder
                         .upgrade_contract_runtime(upgrade_config)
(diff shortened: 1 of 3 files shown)


node/src/components/chainspec_loader.rs
```diff
@@ -373,8 +373,25 @@ impl ChainspecLoader {
...
+
+                // TODO - remove this once we have protocol versions in block headers.
+                //
+                // It is required if we restart after the network has had a hard reset upgrade from
+                // delta currently running at v1.0.0.  In that case, the cached version could well
+                // be the same as that in the current chainspec.  However, we need to pass the
+                // previous version when calling `ContractRuntime::upgrade`.
+                //
+                // If we end up running the release-0.8.0, we'll need to update this workaround to
+                // allow for upgrading from 0.8.0.
```
goral09, line 385: Have you created the JIRA ticket to remove it?

goral09:
I imagine that you hit this when working on some edge-case scenario testing. Can you write that scenario down so that we can add an IT test for it?

node/src/components/chainspec_loader.rs
Fraser999: Well, no.  I don't think this will get removed.  What I meant was that if we end up running delta using `release-0.8.0` and set up a new release branch for _its_ next version, then we'll need to extend this `if` block to cater for that version upgrade too.

This won't need removed from master/dev since it should never end up ported there.

Fraser999:
> I imagine that you hit this when testing the edge-case scenario. Can you write that scenario down so that we can add an IT test for it?

Integration testing this won't be trivial since it involves using two different version of `casper-node` (and it should also really involve two versions of `casper-node-launcher` too, but that's not as critical, so I'll leave it out of the description).

To set up the scenario:

* start a testnet using the launcher and `casper-node` at `release-0.7.6`
* manually kill the testnet after era 2 has started
* install casper-node at `release-0.8.0`.  The chainspec must have `[protocol.version] = '2.0.0'`, `[protocol.hard_reset] = true` and `[protocol.activation_point] = 2`
* for any nodes which already ran, remove their `1_0_0` folders and `casper-node-launcher-state.toml`
* start these nodes and wait for them to continue generating blocks

To see the problem:

* set up a branch new node.  Provide it with `casper-node` at `release-0.7.7` for protocol version 1.0.0, and `release-0.8.0` for protocol version 2.0.0
* start the node and let it catch up with the existing nodes
* kill the node
* restart the node

After restarting, it will panic as we're passing EE the same version for current and new in the `commit_upgrade` call.

sacherjj (approved):

Fraser999:
bors merge

PR #1040, opened 2021-03-01
OP-1879: nctl pass chainspec

TomVasile:
PR allows `nctl-assets-setup` to optionally take a chainspec_path as an argument. Continues to use old default path if not passed.

ie. ` nctl-assets-setup chainspec_path="/tmp/chainspec.toml.in"`


goral09 (approved):
Code looks correct. I can't test it right now as I'm on another branch testing but if you had run it with different chainspec and without one (to use the default) then it's good to go :+1:

TomVasile:
> Code looks correct. I can't test it right now as I'm on another branch testing but if you had run it with different chainspec and without one (to use the default) then it's good to go 👍

yep, thats how I tested it:

```
# Using Old Default
$ nctl-assets-setup
$ grep -ir 'era_duration' utils/nctl/assets/net-1/chainspec/chainspec.toml
era_duration = "41seconds"

# Using Arg - era_duration +1
$ nctl-assets-setup chainspec_path="/tmp/chainspec.toml.in"
$ grep -ir 'era_duration' utils/nctl/assets/net-1/chainspec/chainspec.toml
era_duration = "42seconds"
```

goral09:
Looks good. Thanks.

PR #1046, opened 2021-03-01
NDRS-962: Add more stop conditions for syncing.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-949

This PR adds two new stop conditions for when a node can consider being sync'd with the chain:
1. When a newly downloaded block has been created less than `max_round_length` ago. Works in situations when the network is running with the highest possible round exponent and otherwise there's a long list of known peers to ask for a new block.
2. When a newly downloaded block is a switch block of the currently active era. Works in situations when the finalization takes more than the longest possible round and condition **1** wouldn't kick in.

### DIFF
--- a/utils/nctl/sh/scenarios/sync_test.sh
+++ b/utils/nctl/sh/scenarios/sync_test.sh
@@ -56,7 +56,7 @@ function log_step() {
 
 function do_await_genesis_era_to_complete() {
     log_step "awaiting genesis era to complete"
-    while [ "$(get_chain_era)" -lt 1 ]; do
+    while [ "$(get_chain_era)" != "1" ]; do
         sleep 1.0
     done
 }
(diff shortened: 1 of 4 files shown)


utils/nctl/sh/scenarios/sync_test.sh
```diff
@@ -56,7 +56,7 @@ function log_step() {
 
 function do_await_genesis_era_to_complete() {
     log_step "awaiting genesis era to complete"
-    while [ "$(get_chain_era)" -lt 1 ]; do
+    while [ "$(get_chain_era)" != "1" ]; do
```
afck, line 59: Shouldn't that be `0`?

afck (approved):

utils/nctl/sh/scenarios/sync_test.sh
goral09: Genesis is 0 and we want to transition into the first era after. Recently, code of the client was changed and it now returns `n/a` when it cannot give an answer and the numeric comparison didn't work.

goral09:
bors merge

PR #1047, opened 2021-03-01
NO-TICKET: avoid panicking on handling get responses

Fraser999:
This PR removes a couple of `todo!`s which could allow a malformed or more likely malicious message to cause the node to panic.

### DIFF
--- a/node/src/reactor/validator.rs
+++ b/node/src/reactor/validator.rs
@@ -724,8 +724,14 @@ impl reactor::Reactor for Reactor {
                                 responder: None,
                             })
                         }
-                        Tag::Block => todo!("Handle GET block response"),
-                        Tag::BlockByHeight => todo!("Handle GET BlockByHeight response"),
+                        Tag::Block => {
+                            warn!("received get block response from {}", sender);
+                            return Effects::new();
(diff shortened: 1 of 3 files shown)


node/src/reactor/validator.rs
```diff
@@ -724,8 +724,14 @@ impl reactor::Reactor for Reactor {
...
                             })
                         }
-                        Tag::Block => todo!("Handle GET block response"),
-                        Tag::BlockByHeight => todo!("Handle GET BlockByHeight response"),
+                        Tag::Block => {
+                            warn!("received get block response from {}", sender);
+                            return Effects::new();
+                        }
+                        Tag::BlockByHeight => {
+                            warn!("received get block by height response from {}", sender);
```
goral09, line 732: we should use `warn!(?sender, <msg>)` approach.

goral09 (approved):
No blockers.

node/src/reactor/validator.rs
Fraser999: Sorry - I forgot!  Updated now :)

Fraser999:
bors merge

PR #1048, opened 2021-03-02
[Delta] NDRS-962: Stop syncing early.

goral09:
This PR adds two new conditions for when to stop syncing and transition to an active node:
1. When the most recently downloaded block was created less than the longest possible round ago.
2. When the last downloaded block was a switch block of a currently active era.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -34,7 +34,7 @@ steps:
     commands:
       - cargo install cargo-audit
       - cargo generate-lockfile
-      - cargo audit --ignore RUSTSEC-2020-0123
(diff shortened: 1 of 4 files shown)


afck (approved):
Looks good to me!

In the case where the network was stalled, it can happen that the latest block is older than one maximum round length and the current era is past its minimum duration — but then we'd still finish syncing once we realize that no peer has a later block, right?

goral09:
> Looks good to me!
> 
> In the case where the network was stalled, it can happen that the latest block is older than one maximum round length and the current era is past its minimum duration — but then we'd still finish syncing once we realize that no peer has a later block, right?

Yes, or we hit a _next_ era before that.

goral09:
bors merge

PR #1051, opened 2021-03-02
NDRS-947: RFC6979 Compliant Signatures

xcthulhu:
This PR upgrades the `k256` library to use deterministic signatures conforming to [RFC6979][1]

[1]: https://tools.ietf.org/html/rfc6979


afck (commented):
Looks great! I'm just confused about the arrays instead of key/signature types.

types/src/crypto/asymmetric_key.rs
```diff
@@ -314,8 +313,8 @@ pub enum PublicKey {
     #[data_size(skip)] // Manually verified to have no data on the heap.
     Ed25519(ed25519_dalek::PublicKey),
     /// secp256k1 public key.
-    #[data_size(skip)] // Manually verified to have no data on the heap.
-    Secp256k1(k256::PublicKey),
+    #[data_size(skip)]
```
goral09, line 316: We probably don't need this `skip` anymore.

types/src/crypto/asymmetric_key.rs
```diff
@@ -314,8 +313,8 @@ pub enum PublicKey {
...
+    Secp256k1([u8; SECP256K1_COMPRESSED_PUBLIC_KEY_LENGTH]),
```
goral09, line 317: There's no full-size variant. Is that intentional?

types/src/crypto/asymmetric_key.rs
```diff
@@ -384,27 +376,59 @@ impl AsymmetricType for PublicKey {
...
-                Self::SECP256K1_LENGTH,
-                bytes.as_ref().len()
-            ))
-        })?;
-        public_key.compress();
-        Ok(PublicKey::Secp256k1(public_key))
+        match bytes.as_ref().len() {
+            Self::SECP256K1_LENGTH => {
+                let bytes =
+                    <[u8; Self::SECP256K1_LENGTH]>::try_from(bytes.as_ref()).map_err(|err| {
```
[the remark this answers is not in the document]
goral09: The `match` version will be more readable :+1:

node/src/crypto/asymmetric_key.rs
```diff
@@ -37,8 +36,16 @@ pub fn sign<T: AsRef<[u8]>>(
             Signature::Ed25519(signature.to_bytes())
         }
         (SecretKey::Secp256k1(secret_key), PublicKey::Secp256k1(_public_key)) => {
-            let signer = ecdsa::Signer::new(secret_key).expect("should create secp256k1 signer");
-            Signature::Secp256k1(signer.sign_with_rng(rng, message.as_ref()))
+            let secret_key =
+                secp256k1::SecretKey::parse(secret_key).expect("should create secp256k1 secret");
```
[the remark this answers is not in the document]
afck: (As discussed, it's because it's not `Copy`.)

afck (approved):
No blockers from my side.

node/src/crypto/asymmetric_key.rs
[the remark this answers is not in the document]
goral09: But it's `Clone`, right? I think we should not be parsing the secret for every `sign` operation – it makes it looks like it can pass on the try number 1321 and fail on 1322.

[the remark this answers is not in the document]
xcthulhu: While I agree that is the clean way to do this, `secp256k1::SecretKey` and friends don't implement `Copy` and `AsRef`

node/src/crypto/asymmetric_key.rs
```diff
@@ -282,7 +300,7 @@ mod tests {
         assert_eq!(Some(Ordering::Equal), low.partial_cmp(&low_copy));
     }
 
-    mod ed25519 {
+    mod test_ed25519 {
```
Fraser999, line 303: This is already inside a module named `test` - let's avoid adding the unnecessary prefix here.

node/src/crypto/asymmetric_key.rs
```diff
@@ -493,12 +511,12 @@ MCowBQYDK2VwAyEAGb9ECWmEzf6FQbrBZ9w7lshQhqowtrbLDFw4rXAxZuE=
...
             let public_key = PublicKey::from(&ed25519_secret_key);
             let data = b"data";
-            let signature = sign(data, &ed25519_secret_key, &public_key, &mut rng);
+            let signature = sign(data, &ed25519_secret_key, &public_key);
             bytesrepr::test_serialization_roundtrip(&signature);
         }
     }
 
-    mod secp256k1 {
+    mod test_secp256k1 {
```
Fraser999, line 519: Should remove `test_` prefix.

types/Cargo.toml
```diff
@@ -21,7 +21,7 @@ ed25519-dalek = { version = "1.0.0", default-features = false, features = ["rand
 failure = { version = "0.1.8", default-features = false, features = ["failure_derive"] }
 hex = { version = "0.4.2", default-features = false }
 hex_fmt = "0.3.0"
-k256 = { version = "0.4.2", default-features = false, features = ["ecdsa", "zeroize"] }
+libsecp256k1 = "0.3.5"
```
Fraser999, line 24: This has the feature `std` enabled by default - I think we'll need to change that so that we have `default_features = false` here and enable `std` for this when our `std` feature is enabled.

types/src/crypto/asymmetric_key.rs
```diff
@@ -48,9 +47,9 @@ const ED25519: &str = "Ed25519";
 pub const SECP256K1_TAG: u8 = 2;
 const SECP256K1: &str = "Secp256k1";
 
-const SECP256K1_SECRET_KEY_LENGTH: usize = 32;
-const SECP256K1_COMPRESSED_PUBLIC_KEY_LENGTH: usize = 33;
-const SECP256K1_SIGNATURE_LENGTH: usize = 64;
+const SECP256K1_SECRET_KEY_LENGTH: usize = secp256k1::util::SECRET_KEY_SIZE;
+const SECP256K1_COMPRESSED_PUBLIC_KEY_LENGTH: usize = secp256k1::util::COMPRESSED_PUBLIC_KEY_SIZE;
+const SECP256K1_SIGNATURE_LENGTH: usize = secp256k1::util::SIGNATURE_SIZE;
```
Fraser999, line 52: We should import these secp256k1 consts.

types/src/crypto/asymmetric_key.rs
```diff
@@ -251,10 +251,9 @@ impl ToBytes for SecretKey {
                 let ed25519_bytes = public_key.as_bytes();
                 buffer.extend_from_slice(ed25519_bytes);
             }
-            SecretKey::Secp256k1(public_key) => {
+            SecretKey::Secp256k1(secret_key) => {
```
Fraser999, line 254: Nice catch

types/src/crypto/asymmetric_key.rs
Fraser999: Yes - we didn't want to deal with a mixture of uncompressed and compressed keys.  Our constructors only allow us to hold compressed keys.

types/src/crypto/asymmetric_key.rs
```diff
@@ -384,27 +376,59 @@ impl AsymmetricType for PublicKey {
...
+            unsupported_number_of_bytes => Err(Error::AsymmetricKey(format!(
+                "failed to construct secp256k1 public key. \
+                 Unsupported number of bytes: {}",
+                unsupported_number_of_bytes
+            ))),
+        }
     }
 }
 
+// TODO: Change this to TryFrom, since all zero Secp256k1 secrets are forbidden.
```
Fraser999, line 420: I don't think we should leave this as a todo - I believe it should be handled as part of this PR.

types/src/crypto/asymmetric_key.rs
```diff
@@ -586,8 +609,8 @@ pub enum Signature {
     // it is convenient and cheap to use `signature.as_ref()`.
     Ed25519([u8; ed25519_dalek::SIGNATURE_LENGTH]),
     /// secp256k1 signature.
-    #[data_size(skip)] // Manually verified to have no data on the heap.
-    Secp256k1(ecdsa::Signature),
+    #[data_size(skip)]
```
Fraser999, line 612: Probably don't have to skip any more.

Fraser999 (changes requested):
Please add a fairly explicit description to the PR which should justify this change.

A couple of blocking comments from me: the no-std issue and the use of `openssl::sha::Sha256`.  I'd be very keen to see the `unwrap()` removed from `impl From<&SecretKey> for PublicKey` too, but I won't block the PR for that as long as it gets ticketed if left.

Otherwise just nitpicks.

types/src/crypto/asymmetric_key.rs
xcthulhu: Yes, we must have a canonical representation for public keys.

node/src/crypto/asymmetric_key.rs
xcthulhu: The `secp256k1` module was colliding with the library of the same name, so I changed it to `test_secp256k1`.  I changed this one too for symmetry.

node/src/crypto/asymmetric_key.rs
xcthulhu: Sadly, we can't because this collides with a library with the same name.

types/src/crypto/asymmetric_key.rs
xcthulhu: Fixed in 899be2413bd24f04ad763b7eafba11fa4b9a33a6

EdHastingsCasperLabs (approved):

EdHastingsCasperLabs:
Assuming you apply Fraser's blocking items, LGTM.

node/src/crypto/asymmetric_key.rs
xcthulhu: This was colliding with the crate with the same name, so I changed it to have this prefix.  I also changed `ed25519` for symmetry.

types/Cargo.toml
xcthulhu: Fixed in 8c3d61879143a7ef88bb839bdcf8a6ebdaaba02d

types/src/crypto/asymmetric_key.rs
xcthulhu: Fixed in 899be2413bd24f04ad763b7eafba11fa4b9a33a6

types/src/crypto/asymmetric_key.rs
xcthulhu: Fixed in aa7803f07b9aaa33d7c0d78c2dd138a4f5f6879e

node/src/crypto/asymmetric_key.rs
```diff
@@ -37,8 +36,16 @@ pub fn sign<T: AsRef<[u8]>>(
...
+            let secret_key =
+                secp256k1::SecretKey::parse(secret_key).expect("should create secp256k1 secret");
+            let message_hash = {
+                let mut sha256 = Sha256::new();
+                sha256.update(message.as_ref());
+                sha256.finish()
+            };
+            let (signature, _) =
+                secp256k1::sign(&secp256k1::Message::parse(&message_hash), &secret_key);
+            Signature::Secp256k1(signature.serialize())
```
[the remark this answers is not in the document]
xcthulhu: As mentioned elsewhere, we'd have to manually implement `Copy` and it makes it hard to implement `AsRef`

types/src/crypto/asymmetric_key.rs
goral09: > Yes - we didn't want to deal with a mixture of uncompressed and compressed keys. Our constructors only allow us to hold compressed keys.

Probably worth mentioning in the docs for client integration.

types/src/crypto/asymmetric_key.rs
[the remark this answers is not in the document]
afck: Maybe it doesn't! Maybe the new array pattern matching doesn't support this.

node/src/crypto/asymmetric_key.rs
```diff
@@ -4,13 +4,13 @@
 #![allow(clippy::field_reassign_with_default)]
 
 use ed25519_dalek::ExpandedSecretKey;
-use k256::{ecdsa, ecdsa::signature::Verifier};
-use signature::{RandomizedSigner, Signature as _Signature};
+use openssl::sha::Sha256;
```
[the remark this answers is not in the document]
xcthulhu: Fixed in c70107700c6d1132b373846c7f3ded94aa3e5efa

types/src/crypto/asymmetric_key.rs
xcthulhu: As discussed, we should handle this in a follow up PR.

Fraser999 (approved):

PR #1053, opened 2021-03-02
NO-TICKET: Use the era's earliest block to get activation era validators.

afck:
This changes the logic for retrieving the validator set from the global state as discussed [here](https://github.com/CasperLabs/casper-node/pull/1023#discussion_r585640105).


goral09 (approved):
Thank you.

fizyk20 (approved):

PR #1054, opened 2021-03-02
NDRS-946: Make sure Chainspec::hash is the same for everyone

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/NDRS-946

Changes the way AccountsConfig is deserialized so accounts and delegators are always sorted. Adds a test case that ensures that even with different accounts.toml both Chainspec objects are equal.

### DIFF
--- a/resources/test/valid/0_9_0_unordered/chainspec.toml
+++ b/resources/test/valid/0_9_0_unordered/chainspec.toml
@@ -1,0 +1,142 @@
+[protocol]
+version = '0.9.0'
+hard_reset = false
+
+[protocol.activation_point]
+era_id = 0
+
+[network]
+name = 'test-chain'
+timestamp = '2020-09-18T18:45:00Z'
+
+[core]
+era_duration = '3minutes'
+minimum_era_height = 9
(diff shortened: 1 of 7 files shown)


resources/test/valid/0_9_0_unordered/chainspec.toml
```diff
@@ -0,0 +1,142 @@
...
+read_base_round_reward = 10_000
+
+[system_costs.handle_payment_costs]
+get_payment_purse = 10_000
+set_refund_purse = 10_000
+get_refund_purse = 10_000
+finalize_payment = 10_000
+
+[system_costs.standard_payment_costs]
+pay = 10_000
```
afck, line 142: You could also reorder some of the sections in this file, and in each section the entries.
(That already shouldn't make a difference, of course, but why not make the test check that, too.)

afck (approved):
Looks good to me! :+1:

Fraser999 (approved):

EdHastingsCasperLabs (approved):

PR #1062, opened 2021-03-04
NO-TICKET: change activation point to represent era ID or genesis timestamp

Fraser999:
This PR unifies the genesis timestamp and upgrade activation point into a single enum type.

During a discussion with @EdHastingsCasperLabs, it was noted that genesis timestamps for all chainspecs after the genesis one become meaningless and could be omitted.  Conceptually, the genesis timestamp can be considered the activation point for the first version, hence it seems natural to absorb it into the activation point type.

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -529,12 +529,13 @@ where
             let start_height;
             let era_start_time;
 
-            if let Some(start_time) = genesis_start_time {
+            if era_id.is_genesis() {
                 newly_slashed = vec![];
                 // The validator set was read from the global state: there's no key block for era 0.
                 validators = activation_era_validators.clone();
                 start_height = 0;
-                era_start_time = start_time;
+                era_start_time =
+                    genesis_start_time.expect("must have genesis start time if era ID is 0");
             } else {
                 // If this is not era 0, there must be a key block for it.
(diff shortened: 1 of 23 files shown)


EdHastingsCasperLabs (approved):

node/src/components/consensus/era_supervisor.rs
```diff
@@ -529,12 +529,12 @@ where
             let start_height;
             let era_start_time;
 
-            if era_id.is_genesis() {
+            if let Some(start_time) = genesis_start_time {
```
afck, line 532: That still needs to check whether `era_id.is_genesis()`, otherwise it uses this start time in _all_ iterations.

afck (changes requested):

node/src/components/consensus/era_supervisor.rs
Fraser999: Ah - good catch!

I guess I would expect a test to catch this case.  I even ran nctl with this code and it got past era 0.

afck: Yes, I would have expected that, too: Do we have a test where a node joins in an era other than 1, but without an upgrade?

goral09: We do but it runs only in nightly build so it would break there.

Fraser999:
bors merge

PR #1063, opened 2021-03-04
NO-TICKET: fix gossip metrics

Fraser999:
This PR fixes the gossiper metrics, some of which are fairly useless at the moment.

* `items_received` now correctly represents the total of all gossip items received, whether passed to the component by a client, by a different component (e.g. the small network in the case of starting a new round of gossiping its own address) or by a peer gossiping it.
* `items_gossiped_onwards` has been replaced by `times_gossiped` and now represents the total number of outbound gossip messages the node has sent.

It also standardises the unregistering of metrics from the Prometheus registry to remove the risky calls to `expect` in the `Drop` impls of many of the metrics structs.  If there are objections to the introduction of a macro to achieve this, I would be happy to change it back to normal function calls.

Tagging @sacherjj & @TomVasile for info.

### DIFF
--- a/node/src/utils.rs
+++ b/node/src/utils.rs
@@ -301,3 +301,18 @@ where
 {
     (numerator + denominator / T::from(2)) / denominator
 }
+
+/// Used to unregister a metric from the Prometheus registry.
+#[macro_export]
+macro_rules! unregister_metric {
+    ($registry:expr, $metric:expr) => {
+        $registry
+            .unregister(Box::new($metric.clone()))
+            .unwrap_or_else(|_| {
+                tracing::error!(
+                    "unregistering {} failed: was not registered",
+                    stringify!($metric)
+                )
+            });
+    };
+}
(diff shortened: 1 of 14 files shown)


EdHastingsCasperLabs (approved):

node/src/utils.rs
```diff
@@ -301,3 +301,18 @@ where
...
+        $registry
+            .unregister(Box::new($metric.clone()))
+            .unwrap_or_else(|_| {
+                tracing::error!(
+                    "unregistering {} failed: was not registered",
+                    stringify!($metric)
+                )
+            });
+    };
+}
```
goral09, line 318: Why does it have to be a macro and not a plain function?

goral09 (approved):

node/src/utils.rs
Fraser999: It's just to allow us to keep printing the name of the specific metric.

Fraser999:
bors merge

PR #1065, opened 2021-03-04
typedoc upgrade to 0.19.1

hoffmannjan:
upgraded typedoc to not-latetst version but dependency version (highlight.js) is no longer giving alert about EOL.
@mpapierski please check that this version generates proper documentation :)


mpapierski (approved):
LGTM 👍 

tested with 

```
npm run prepublish-docs
```
generated proper docs

PR #1066, opened 2021-03-04
EEE-1188: Restore delegator slashing

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1188

- Sets all delegator stakes to 0 for slashed validator
- Adds a test that ensures delegators of a validator have their stake set to 0 after slashing


henrytill (approved):
:+1:

PR #1067, opened 2021-03-04
NDRS-963: chainspec loader now gets the protocol version from the highest block

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-963

This PR fixes an issue whereby a restarting node on a network which has upgraded would call the contract runtime's `commit_upgrade()` with an incorrect `current_version`.  Rather than storing the protocol version in storage and using that on restart, we now use the protocol version from the highest block now that it is available to us.

The PR also temporarily disables the trie store check as it is too slow to support as is.  It will be re-enabled in some form once a decision on how to address this issue permanently has been made.

### DIFF
--- a/node/src/components/chainspec_loader.rs
+++ b/node/src/components/chainspec_loader.rs
@@ -20,7 +20,6 @@ use std::{
 
 use datasize::DataSize;
 use derive_more::From;
-use futures::join;
 use schemars::JsonSchema;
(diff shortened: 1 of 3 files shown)


EdHastingsCasperLabs (approved):

Fraser999:
bors merge

PR #1068, opened 2021-03-04
NDRS-966: Fix retrieval of booking block hashes for eras initialized after restart.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-966

Fixes retrieval of booking block hashes for eras initialized after the restart.

### DIFF
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -118,7 +118,7 @@ pub enum Event<I> {
         /// The header of the switch block
         block: Box<Block>,
         /// Ok(block_hash) if the booking block was found, Err(height) if not
-        booking_block_hash: Result<BlockHash, u64>,
+        booking_block_hash: Result<BlockHash, EraId>,
     },
     /// Event raised upon initialization, when a number of eras have to be instantiated at once.
     InitializeEras {
@@ -323,10 +323,10 @@ where
             } => {
                 let booking_block_hash = match booking_block_hash {
                     Ok(hash) => hash,
-                    Err(height) => {
+                    Err(era_id) => {
                         error!(
-                            "could not find the booking block at height {} for era {}",
-                            height,
+                            "could not find the booking block in era {}, for era {}",
+                            era_id,
                             block.header().era_id().successor()
                         );
                         return fatal!(
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -53,6 +53,9 @@ pub(crate) use era_supervisor::{EraId, EraSupervisor};
 pub(crate) use protocols::highway::HighwayProtocol;
 use traits::NodeIdT;
 
+#[cfg(test)]
+pub(crate) use era_supervisor::oldest_bonded_era;
+
 #[derive(DataSize, Clone, Serialize, Deserialize)]
 pub enum ConsensusMessage {
     /// A protocol message, to be handled by the instance in the specified era.
@@ -125,7 +128,7 @@ pub enum Event<I> {
         /// validator set is read from the global state, not from a key block.
         validators: BTreeMap<PublicKey, U512>,
         timestamp: Timestamp,
-        genesis_start_time: Timestamp,
+        genesis_start_time: Option<Timestamp>,
     },
     /// An event instructing us to shutdown if the latest era received no votes.
     Shutdown,
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -40,11 +40,13 @@ use crate::{
         },
         EffectBuilder, Effects,
     },
+    fatal,
     protocol::Message,
     types::{ActivationPoint, Block, BlockHash, BlockHeader, ProtoBlock, Timestamp},
     NodeRng,
 };
 
+use crate::effect::EffectExt;
 pub use config::Config;
 pub(crate) use consensus_protocol::{BlockContext, EraReport};
 pub(crate) use era_supervisor::{EraId, EraSupervisor};
@@ -315,14 +317,21 @@ where
                 block,
                 booking_block_hash,
             } => {
-                let booking_block_hash = booking_block_hash.unwrap_or_else(|height| {
-                    error!(
-                        "could not find the booking block at height {} for era {}",
-                        height,
-                        block.header().era_id().successor()
-                    );
-                    panic!("couldn't get the booking block hash");
-                });
+                let booking_block_hash = match booking_block_hash {
+                    Ok(hash) => hash,
+                    Err(height) => {
+                        error!(
+                            "could not find the booking block at height {} for era {}",
+                            height,
+                            block.header().era_id().successor()
+                        );
+                        return fatal!(
+                            handling_es.effect_builder,
+                            "couldn't get the booking block hash"
+                        )
+                        .ignore();
+                    }
+                };
                 handling_es.handle_create_new_era(*block, booking_block_hash)
             }
             Event::InitializeEras {
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -347,7 +347,8 @@ where
                     key_blocks,
                     booking_blocks,
                     validators,
-                    timestamp);
+                    timestamp,
+                );
 
                 // TODO: remove that when possible
                 // This is needed because we want to make sure that we only try to handle linear
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -120,6 +120,7 @@ pub enum Event<I> {
     /// Event raised upon initialization, when a number of eras have to be instantiated at once.
     InitializeEras {
         key_blocks: HashMap<EraId, BlockHeader>,
+        booking_blocks: HashMap<EraId, BlockHash>,
         /// This is empty except if the activation era still needs to be instantiated: Its
         /// validator set is read from the global state, not from a key block.
         validators: BTreeMap<PublicKey, U512>,
@@ -336,12 +337,14 @@ where
             }
             Event::InitializeEras {
                 key_blocks,
+                booking_blocks,
                 validators,
                 timestamp,
                 genesis_start_time,
             } => {
                 let mut effects = handling_es.handle_initialize_eras(
                     key_blocks,
+                    booking_blocks,
                     validators,
                     timestamp,
                     genesis_start_time,
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -128,7 +128,6 @@ pub enum Event<I> {
         /// validator set is read from the global state, not from a key block.
         validators: BTreeMap<PublicKey, U512>,
         timestamp: Timestamp,
-        genesis_start_time: Option<Timestamp>,
     },
     /// An event instructing us to shutdown if the latest era received no votes.
     Shutdown,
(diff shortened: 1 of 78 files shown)


node/src/effect.rs
```diff
@@ -1555,6 +1555,25 @@ impl<REv> EffectBuilder<REv> {
...
+    /// It is expected that the caller of this function knows in which era the booking blocks live.
+    pub(crate) async fn collect_booking_blocks<I: IntoIterator<Item = EraId>>(
+        self,
+        booking_blocks_era_ids: I,
+    ) -> Option<HashMap<EraId, BlockHeader>>
+    where
+        REv: From<StorageRequest>,
+    {
+        futures::future::join_all(booking_blocks_era_ids.into_iter().map(|era_id| {
+            self.get_key_block_for_era_id_from_storage(era_id)
```
afck, line 1570: That gets the switch block in `era_id - 1` — is that intended?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -176,6 +179,8 @@ where
             protocol_config.protocol_version.patch as u32,
         );
         let activation_era_id = protocol_config.last_activation_point;
```
afck, line 181: You can remove that now — you duplicated it two lines below.
```suggestion
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -204,15 +209,25 @@ where
...
+            let booking_blocks = collect_booking_block_hashes(
+                effect_builder,
+                era_ids.clone(),
+                auction_delay,
+                last_activation_point,
+            )
+            .await;
+
             if current_era > activation_era_id + bonded_eras * 2 {
                 // All eras can be initialized using the key blocks only.
```
afck, line 228:
```suggestion
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -591,6 +622,64 @@ where
...
+/// In those cases, returns `None`.
+fn valid_booking_block(
+    era_id: &EraId,
+    auction_delay: u64,
+    last_activation_point: EraId,
+) -> Option<EraId> {
+    let after_booking_era_id = era_id
+        .saturating_sub(auction_delay)
+        .max(last_activation_point);
+    if after_booking_era_id.is_genesis() {
```
afck, line 639: Isn't that redundant? If it's genesis, it will also be `== last_activation_point`.

goral09: I think so, yes. I just wasn't sure (too tired).

node/src/effect.rs
goral09: No. It will get the switch block in `era_id`.

afck: But the "key block for era" is the switch block in the previous era… maybe that method should have a different name then.

goral09: Wait, it will get the switch block in era `era_id - 1` :tired_face:

node/src/components/consensus/era_supervisor.rs
goral09: Removed in [`275d960` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/275d960613df12b7732d16cf8fa6cc96f5b1270c)

node/src/components/consensus/era_supervisor.rs
goral09: Deleted in [`45e2470` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/45e247049e02d72cbd9c9044cf0e7535f192b9fe)

node/src/components/consensus/era_supervisor.rs
goral09: Removed in [`45e2470` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/45e247049e02d72cbd9c9044cf0e7535f192b9fe)

node/src/effect.rs
goral09: Actually, I think I wanted to read switch blocks (not key blocks) and just used a wrong method.

goral09: Fixed in [`ac8d609` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/ac8d609c18a52d4da666b399b6ec44360deb98d6) and then fixed again in [`42efc26` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/42efc26740f2029c1fb7a3268f07abfaa312226e)

node/src/components/consensus.rs
```diff
@@ -116,11 +118,12 @@ pub enum Event<I> {
         /// The header of the switch block
         block: Box<Block>,
         /// Ok(block_hash) if the booking block was found, Err(height) if not
```
afck, line 120:
```suggestion
        /// `Ok(block_hash)` if the booking block was found, `Err(era_id)` if not
```

node/src/components/linear_chain_sync.rs
```diff
@@ -302,12 +302,22 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
                     "Block execution result doesn't match received block."
                 );
                 if self.is_recent_block(&block) {
-                    info!("downloaded recent block. finished synchronization");
+                    info!(
+                        hash=?block.hash(),
+                        height=?block.header().height(),
+                        era=?block.header().era_id(),
```
afck, line 308:
```suggestion
                        era=block.header().era_id().0,
```
(That's what we do elsewhere, so that you read the word "era" only once.)

node/src/components/linear_chain_sync.rs
```diff
@@ -302,12 +302,22 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
...
+                    );
                     self.mark_done();
                     return Effects::new();
                 }
                 if self.is_currently_active_era(&maybe_switch_block) {
-                    info!("downloaded block from a new era. finished synchronization");
+                    info!(
+                        hash=?block.hash(),
+                        height=?block.header().height(),
+                        era=?block.header().era_id(),
```
afck, line 318:
```suggestion
                        era=block.header().era_id().0,
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -522,6 +524,11 @@ where
             let validators;
             let start_height;
             let era_start_time;
+            let seed;
+
+            let booking_block_hash = booking_blocks
+                .get(&era_id)
+                .expect("should have booking block");
```
afck, line 531: (This looks surprising… but we can `expect` that because the map _does_ have entries for all eras, and they're just `default()` for the ones without booking blocks, right?)

node/src/components/consensus/era_supervisor.rs
```diff
@@ -530,11 +537,13 @@ where
                 start_height = 0;
                 era_start_time =
                     genesis_start_time.expect("must have genesis start time if era ID is 0");
+                seed = 0;
             } else {
-                // If this is not era 0, there must be a key block for it.
                 let key_block = key_blocks.get(&era_id).expect("missing key block");
+                // If this is not era 0, there must be a key block for it.
```
afck, line 543:
```suggestion
                // If this is not era 0, there must be a key block for it.
                let key_block = key_blocks.get(&era_id).expect("missing key block");
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -595,6 +599,102 @@ where
     }
 }
 
+/// Returns an era ID in which the booking block for `era_id` lives in, if we can use it.
```
afck, line 602:
```suggestion
/// Returns an era ID in which the booking block for `era_id` lives, if we can use it.
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -595,6 +599,102 @@ where
...
+    era_id: EraId,
+    auction_delay: u64,
+    last_activation_point: EraId,
+) -> Option<EraId> {
+    // If this is the first era after the upgrade/restart/Genesis,
+    // booking block is not defined.
+    if era_id == last_activation_point {
+        return None;
+    }
+
```
afck, line 623: That seems redundant. It's already checked below.
```suggestion
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -595,6 +599,102 @@ where
...
+    let booking_era_id = era_id
+        .saturating_sub(auction_delay)
+        .saturating_sub(1)
+        .max(last_activation_point);
```
afck, line 627:
```suggestion
    let booking_era_id = era_id
        .saturating_sub(auction_delay)
        .saturating_sub(1);
```
No need to `max`, if we return `None` in those cases anyway.

If we `max` here, the `if` condition below can never be satisfied.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -595,6 +599,102 @@ where
...
+
+    // If we would have gone below the last activation point (the first `AUCTION_DELAY ` eras after
+    // an upgrade), we return `None` as there are no booking blocks there that we can use – we
+    // can't use anything from before an upgrade.
+    // NOTE that it's OK if `booking_era_id` == `last_activation_point`.
+    if booking_era_id < last_activation_point {
+        return None;
+    }
+
+    Some(booking_era_id)
```
afck, line 637: Rust supports this now — it's a matter of taste, though. :grin: 
```suggestion
    (booking_era_id >= last_activation_point).then(booking_era_id)
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -595,6 +599,102 @@ where
...
+            .await
+        {
+            Some(block) => *block.hash(),
+            None => {
+                error!(
+                    ?era_id,
+                    ?booking_block_era_id,
+                    "booking block for era must exist"
+                );
+                panic!("booking block not found in storage");
```
afck, line 664: Since we have an `effect_builder`, I guess we could use `fatal!` here, too.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -829,32 +932,35 @@ where
         effects
     }
 
+    ///  Creates new era.
```
afck, line 935:
```suggestion
    /// Creates a new era.
```

afck (commented):
Looks good; the only potential blocker is the comment on `era_supervisor.rs:627`.

node/src/components/consensus/era_supervisor.rs
goral09: Doesn't type-check. I tried.

node/src/components/consensus/era_supervisor.rs
goral09: Yes. At least, that's the plan.

node/src/components/consensus/era_supervisor.rs
goral09: OK :)

goral09: Replace in [`0fb82b1` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/0fb82b173e1fea440596ef4c214de1c4100a5d7a)

node/src/components/consensus/era_supervisor.rs
goral09: Done in [`0fb82b1` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/0fb82b173e1fea440596ef4c214de1c4100a5d7a)

node/src/components/consensus/era_supervisor.rs
goral09: Done in [`0fb82b1` (#1068)](https://github.com/CasperLabs/casper-node/pull/1068/commits/0fb82b173e1fea440596ef4c214de1c4100a5d7a)

afck (approved):

afck:
Sorry for the merge conflict! :grimacing:

goral09:
bors merge

goral09:
bors merge

PR #1070, opened 2021-03-04
NDRS-971: Restrict cleanup to unclean shutdowns

marc-casperlabs:
Adds a `Pidfile` that gets created when using the cli. The pidfile will be removed only if `Pidfile` is dropped safely, thus indicating that any previous reactor has also been cleanly dropped.

If we drop it in using logic of always continuing to operate, only omitting integrity checks on a `PidfileOutcome::Clean` we should have strictly better behavior than before, without risking to exit due to bugs in `Pidfile`, provided we do not give false positives on `PidfileOutcome::Clean`

Not as clear cut as I'd like, due to the awkward placement of the integrity check, but otherwise very reasonable. It also makes it so that we can actually cleanly unwind, from what I understand we _also_ crashed previously due to `process::exit`. E.g. when exiting due to `SIGTERM`, `process::exit` is called from within the scope where the validator lives, preventing its destructors to run.


node/src/utils.rs
```diff
@@ -315,4 +320,157 @@ macro_rules! unregister_metric {
...
+    /// The file will be locked for the lifetime of `Pidfile`.
+    pidfile: File,
+    /// The pidfile location.
+    path: PathBuf,
+    /// Previous pidfile contents.
+    previous: Option<u32>,
+}
+
+impl Pidfile {
+    /// Acquire a `Pidfile` and give an actionale outcome.
```
xcthulhu, line 392:
```suggestion
    /// Acquire a `Pidfile` and give an actionable outcome.
```

node/src/utils.rs
```diff
@@ -315,4 +320,157 @@ macro_rules! unregister_metric {
...
+            .truncate(false)
+            .create(true)
+            .read(true)
+            .write(true)
+            .open(path.as_ref())
+            .map_err(PidfileError::CouldNotOpen)?;
+
+        // Now try to acquire an exclusive lock. This will fail if another process or another
+        // instance of `Pidfile` is holding a lock onto the same pidfile.
+        pidfile.lock_exclusive().map_err(PidfileError::LockFailed);
```
xcthulhu, line 427:
```suggestion
        pidfile.lock_exclusive().map_err(PidfileError::LockFailed)?;
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    num::ParseIntError,
+    path::{Path, PathBuf},
+    process,
+};
+
+use fs2::FileExt;
+use thiserror::Error;
+use tracing::warn;
+
+/// A PID file.
```
Fraser999, line 17:
```suggestion
/// A PID (process ID) file.
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+///
+/// Records the current process' PID, removes it on exit. Can be used to determine whether or not
+/// an application was shut down cleanly.
+///
+/// The pidfile is held open with an exclusive but advisory lock.
+#[derive(Debug)]
+pub struct Pidfile {
```
Fraser999, line 24:
```suggestion
pub struct PidFile {
```

I'd like to see the module renamed too please (to `pid_file`).

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    pidfile: File,
+    /// The pidfile location.
+    path: PathBuf,
+    /// Previous pidfile contents.
+    previous: Option<u32>,
+}
+
+/// An error acquiring a pidfile.
+#[derive(Debug, Error)]
+pub enum PidfileError {
```
Fraser999, line 37:
```suggestion
pub enum PidFileError {
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// The pidfile could not be opened at all.
+    #[error("could not pidfile: {0}")]
```
Fraser999, line 39:
```suggestion
    #[error("could not open pid file: {0}")]
```

I also think all instances of "pidfile" should be changed to "pid file"

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// Error reading pidfile contents.
+    #[error("reading existing pidfile failed: {0}")]
+    ReadFailed(#[source] io::Error),
+    /// Error writing pidfile contents.
+    #[error("updating pidfile failed: {0}")]
+    WriteFailed(#[source] io::Error),
+    /// The pidfile was corrupted, its contents could not be read.
+    #[error("corrupt pidfile")]
+    Corrupted(ParseIntError),
+    /// We encountere a previous PID, but it was the same as ours.
```
Fraser999, line 53:
```suggestion
    /// We encountered a previous PID, but it was the same as ours.
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    ///
+    /// This should never happen, as the exclusive lock should prevent it.
```
Fraser999, line 55: I know it's unlikely, but we could have crashed on a previous run and have started a new process which happens to have the same PID.

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    #[error("read back our own pid from exlusively locked pidfile")]
+    DuplicatedPid,
+}
+
+/// Pidfile outcome.
+///
+/// High-level description of the outcome of opening and locking the PIDfile.
+#[must_use]
+#[derive(Debug)]
+pub enum PidfileOutcome {
```
Fraser999, line 65:
```suggestion
pub enum PidFileOutcome {
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    Crashed(Pidfile),
+    /// Clean start, pidfile lock acquired.
+    ///
+    /// **Recommendation**: Continue with intialization, but **store the `Pidfile`**.
+    Clean(Pidfile),
+    /// There was an error managing the pidfile, not sure if we have crashed or not.
+    ///
+    /// **Recommendation**: Exit, as it will not be possible to determine a crash at the next
+    /// start.
+    PidfileError(PidfileError),
```
Fraser999, line 83:
```suggestion
    PidFileError(PidFileError),
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// Error writing pidfile contents.
+    #[error("updating pidfile failed: {0}")]
+    WriteFailed(#[source] io::Error),
+    /// The pidfile was corrupted, its contents could not be read.
+    #[error("corrupt pidfile")]
+    Corrupted(ParseIntError),
+    /// We encountere a previous PID, but it was the same as ours.
+    ///
+    /// This should never happen, as the exclusive lock should prevent it.
+    #[error("read back our own pid from exlusively locked pidfile")]
```
Fraser999, line 56:
```suggestion
    #[error("read back our own pid from exclusively locked pidfile")]
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+#[must_use]
+#[derive(Debug)]
+pub enum PidfileOutcome {
+    /// Another instance of the node is likely running, or an attempt was made to reuse a pidfile.
+    ///
+    /// **Recommendation**: Exit to avoid resource conflicts.
+    AnotherNodeRunning(PidfileError),
+    /// The node crashed previously and could potentially have been corrupted.
+    ///
+    /// **Recommendation**: Run an integrity check, then potentially continue with intialization.
```
Fraser999, line 72:
```suggestion
    /// **Recommendation**: Run an integrity check, then potentially continue with initialization.
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    ///                     **Store the `Pidfile`**.
+    Crashed(Pidfile),
+    /// Clean start, pidfile lock acquired.
+    ///
+    /// **Recommendation**: Continue with intialization, but **store the `Pidfile`**.
```
Fraser999, line 77:
```suggestion
    /// **Recommendation**: Continue with initialization, but **store the `Pidfile`**.
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+            }
+            Err(err @ PidfileError::LockFailed(_)) => PidfileOutcome::AnotherNodeRunning(err),
+            Err(err) => PidfileOutcome::PidfileError(err),
+        }
+    }
+
+    /// Creates a new pidfile.
+    ///
+    /// The error-behavior of this function is important and can be used to distinguish between
+    /// different conditions according to the table below. If the `Pidfile` is instantiated before
```
Fraser999, line 108: Which table?

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+            .try_lock_exclusive()
+            .map_err(PidfileError::LockFailed)?;
+
+        // At this point, we're the exclusive users of the file and can read its contents.
+        let mut raw_contents = String::new();
+        pidfile
+            .read_to_string(&mut raw_contents)
+            .map_err(PidfileError::ReadFailed)?;
+
+        // Note: We cannot distinguish an empty file from a non-existing file, unfortunately.
```
Fraser999, line 132: Wouldn't `Path::is_file()` used at the start of this function achieve that?

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+    #[test]
+    fn pidfile_creates_file_and_cleans_it_up() {
+        let tmp_dir = TempDir::new().expect("could not create tmp_dir");
+        let pidfile_path = tmp_dir.path().join("create_and_cleanup.pid");
+
+        let outcome = Pidfile::acquire(&pidfile_path);
+
+        match outcome {
+            PidfileOutcome::Clean(pidfile) => {
+                // Check the pidfile exists, the verify it gets removed after dropping the pidfile.
```
Fraser999, line 200:
```suggestion
                // Check the pidfile exists, then verify it gets removed after dropping the pidfile.
```

node/src/utils/pidfile.rs
```diff
@@ -0,0 +1,255 @@
...
+        let tmp_dir = TempDir::new().expect("could not create tmp_dir");
+        let pidfile_path = tmp_dir.path().join("create_and_cleanup.pid");
+
+        let outcome = Pidfile::acquire(&pidfile_path);
+
+        match outcome {
+            PidfileOutcome::Clean(_pidfile) => {
+                match Pidfile::acquire(&pidfile_path) {
+                    PidfileOutcome::AnotherNodeRunning(_) => {
+                        // All good, lets ensure the loaded pid matches our current pid.
```
Fraser999, line 244:
```suggestion
                        // All good, let's ensure the loaded pid matches our current pid.
```

Also, looks like we're not actually doing the stated check.

Fraser999 (commented):
Shouldn't the checks done in storage initialization also be disabled if the node exited cleanly, not just the trie store checks?

node/src/app/main.rs
```diff
@@ -41,6 +41,7 @@ fn panic_hook(info: &PanicInfo) {
     process::abort()
 }
 
+#[quit::main]
```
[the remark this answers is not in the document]
marc-casperlabs: I've settled for just calling `process::exit()` as the last thing in main, turning the latter also into a pure wrapper.

node/src/utils/pidfile.rs
marc-casperlabs: Can do. Thanks, rust-analyzer!

node/src/utils/pidfile.rs
marc-casperlabs: In my defense, four `i`s is at least two above legal limit.

node/src/utils/pidfile.rs
marc-casperlabs: We have to open it to lock it, while creating it if missing. It's certainly to possible to do all this, but not in a race-free manner. The only alternative I can see is to create two files, a pidfile and a lockfile, but that's a whole different can of worms.

node/src/utils/pidfile.rs
marc-casperlabs: My bad, will be fixed.

node/src/utils/pidfile.rs
marc-casperlabs: Fixed.

node/src/app/cli.rs
```diff
@@ -158,7 +192,7 @@ impl Cli {
                 let registry = Registry::new();
 
                 let mut initializer_runner = Runner::<initializer::Reactor>::with_metrics(
-                    validator_config,
+                    (crashed, validator_config),
```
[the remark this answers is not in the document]
marc-casperlabs: We do not want to make this "outer" crash check official in any way - components are supposed to have a pidfile inside them.

This is just a workaround for the current structure of this particular check, which is not implemented inside the component.

node/src/utils/pidfile.rs
marc-casperlabs: I'll remove the check.

node/src/app/cli.rs
```diff
@@ -148,6 +150,40 @@ impl Cli {
                 let validator_config = Self::init(&config, config_ext)?;
                 info!(version = %casper_node::VERSION_STRING.as_str(), "node starting up");
 
+                let pidfile_outcome = {
+                    // Determine storage directory to store pidfile in.
+                    let storage_config = validator_config.map_ref(|cfg| cfg.storage.clone());
+                    let root = storage_config.with_dir(storage_config.value().path.clone());
+
+                    // Create directory if it does not exit, similar to how the storage component
```
Fraser999, line 158:
```suggestion
                    // Create directory if it does not exist, similar to how the storage component
```

node/src/utils/pid_file.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// **Recommendation**: Exit to avoid resource conflicts.
+    AnotherNodeRunning(PidFileError),
+    /// The node crashed previously and could potentially have been corrupted.
+    ///
+    /// **Recommendation**: Run an integrity check, then potentially continue with initialization.
+    ///                     **Store the `PidFile`**.
+    Crashed(PidFile),
+    /// Clean start, pidfile lock acquired.
+    ///
+    /// **Recommendation**: Continue with intialization, but **store the `PidFile`**.
```
Fraser999, line 72:
```suggestion
    /// **Recommendation**: Continue with initialization, but **store the `PidFile`**.
```

node/src/utils/pidfile.rs
Fraser999: Yes, fair point.  The scope for a race is small, but not zero.  I guess we could check the creation timestamp after opening, but all this seems like overkill.  I'm happy to leave as is.

Fraser999 (approved):
LGTM now, barring the couple of remaining typos.

marc-casperlabs:
Typos have been fixed.

bors r+

PR #1074, opened 2021-03-05
NO-TICKET: Faster trie integrity checks

xcthulhu:


execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -241,7 +246,11 @@ where
     E: From<S::Error> + From<bytesrepr::Error>,
 {
     let mut missing_descendants = Vec::new();
+    let mut visited = HashSet::new();
     while let Some(trie_key) = trie_keys_to_visit.pop() {
+        if !visited.insert(trie_key.clone()) {
```
afck, line 251: No need to clone, says Clippy.

afck (approved):

goral09:
We can also only read the last `UNBONDING_DELAY * 2 + 1` switch blocks and verify the integrity of their state root hashes.

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1074/commits/99742d03286e7e825ef94537ef644bd016db9ec6

xcthulhu:
> We can also only read the last UNBONDING_DELAY * 2 + 1 switch blocks and verify the integrity of their state root hashes.

Might break things when people read under a particular state root in the RPC...

PR #1077, opened 2021-03-05
NDRS-974: Get rid of failure

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/NDRS-974

Removes failure and uses optional thiserror for std case, and for no_std (i.e. wasm32 target) is hidden behind cfg_attr. This complication happens because thiserror does not support no_std upstream but a PR is pending https://github.com/dtolnay/thiserror/pull/64

### DIFF
--- a/types/src/system/mod.rs
+++ b/types/src/system/mod.rs
@@ -4,8 +4,6 @@ pub mod handle_payment;
 pub mod mint;
 pub mod standard_payment;
 
-// use thiserror::Error;
-
 pub use error::Error;
 pub use system_contract_type::{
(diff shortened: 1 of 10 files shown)


types/src/system/mod.rs
```diff
@@ -1,31 +1,34 @@
 //! System modules, formerly known as "system contracts"
-
 pub mod auction;
 pub mod handle_payment;
 pub mod mint;
 pub mod standard_payment;
 
+// use thiserror::Error;
+
```
Fraser999, line 8:
```suggestion
```

Fraser999 (approved):
One nitpick, but LGTM

PR #1078, opened 2021-03-05
OP-1883: itst14 nctl scenario

TomVasile:
Test:
```
GIVEN a set of Genesis validators
WHEN validator restarts within single era
THEN it can produce new consensus messages without equivocating
```

Changes:
- Implements itst14 from https://casperlabs.atlassian.net/wiki/spaces/OPS/pages/1133707277/Casper+Network+Infra+Test+Scenarios.
- Updates tomls used in other test to accommodate changes in dev.
- Moves `check_faulty`, `get_node_public_key_hex`, `get_node_public_key_hex_extended` to common utils. 
- Refactors `check_current_era` in case of null(s) or n/a(s)
- Updates `nightly_test.sh` script to accepts custom `accounts.toml`

Notes:
For the itst14 chainspec.toml I set `era_duration = '300seconds'`  to accommodate test.


goral09 (changes requested):
This does not guarantee us that the restarted node definitely did not equivocate – it would pass if it had synced up correctly and is receiving other validators' messages (to finalize a block network doesn't need that node's input). It does not verify that it itself is able to produce messages and not equivocate. Hardening would require verifying that post-restart, the restarted node, produces blocks. We can grep logs for `FinalizeBlock` with the creator key equal to the public key of the restarted node.

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,121 @@
...
+        exit 1
+    fi
+}
+
+function assert_node_proposed() {
+    local NODE_ID=${1}
+    local NODE_PATH=$(get_path_to_node $NODE_ID)
+    local PUBLIC_KEY_HEX=$(get_node_public_key_hex $NODE_ID)
+    log_step "Waiting for a node-$NODE_ID to produce a block..."
+    local OUTPUT=$(tail -f "$NODE_PATH/logs/stdout.log" | grep -m 1 "proposer: PublicKey::Ed25519($PUBLIC_KEY_HEX)")
```
goral09, line 75: Do we have a general time out on the IT tests so that we fail if this doesn't return? How is that signaled to the CI? Is it easy to notice?

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,121 @@
...
+    local NUM_T=${1:-1}
+    local NUM_RUNNING_NODES=$(nctl-status | grep -i 'running' | wc -l)
+    log_step "sending $NUM_T native transfers to $NUM_RUNNING_NODES running nodes..."
+    for ((i=0; i<=$NUM_RUNNING_NODES; i++)); do
+        log "node-$i: $NUM_T transfers sent."
+        local TRANSFER_OUT=$(nctl-transfer-native node=$i ammount=2500000000 transfers="$NUM_T" user=1)
+    done
+}
+
+function check_node_equivocated() {
```
goral09, line 90: I think a more suitable name would be `assert_no_equivocation` – since our expected outcome is to not equivocate.

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,121 @@
...
+    # 3. Restart Node
+    do_read_lfb_hash 1
+    do_start_node "5" "$LFB_HASH"
+    # 4. Verify all nodes are in sync
+    check_network_sync
+    # 5. Verify network chain height sync
+    log "Sleeping 10s"
+    sleep 10
+    check_network_sync
+    # 6. Send deploy to stopped node
```
goral09, line 41:
```suggestion
    # 6. Send deploy to all nodes
```

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,121 @@
...
+    send_transfers '100'
```
goral09, line 42: We shouldn't even need to send any transfers. Nodes will produce empty blocks. If we can simplify the test to its essential parts, I would prefer that.

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,121 @@
...
+    # 7. Verify network finalized deploy
```
goral09, line 43:
```suggestion
    # 7. Verify node proposed a block
```

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,121 @@
...
+function main() {
+    log "------------------------------------------------------------"
+    log "Starting Scenario: itst14"
+    log "------------------------------------------------------------"
+
+    # 0a. Verify all nodes are in sync
+    check_network_sync
+    log "Sleeping 10s"
+    sleep 10
+    check_network_sync
```
goral09, line 24: I would replace these with `nctl-await-n-blocks 5`, (as long as 5 is less than the minimum era height from the chainspec) to verify that network is capable of proposing blocks.

goral09 (approved):
Looks good, some clean-up and suggestions to simplify the test. It would be nice if we could get them in.

utils/nctl/sh/scenarios/itst14.sh
TomVasile: the drone run would eventually time out, (right now its set to 3hrs to accommodate more tests that come in)

utils/nctl/sh/scenarios/itst14.sh
TomVasile: didnt even know that one existed, thanks!

utils/nctl/sh/scenarios/itst14.sh
goral09: Can we wrap it with a timeout? I think the `min_era_duration` should do. After all, we expect the node to propose a block rather shortly.

TomVasile: yea we can do it, I'll have to adjust the function a bit. The `tail -f` will hold it open indefinitely.

utils/nctl/sh/scenarios/itst14.sh
TomVasile: fixed

utils/nctl/sh/scenarios/itst14.sh
TomVasile: removed this function

utils/nctl/sh/scenarios/itst14.sh
TomVasile: removed this function

utils/nctl/sh/scenarios/itst14.sh
TomVasile: fixed

utils/nctl/sh/scenarios/itst14.sh
TomVasile: added a new function to common for this. I see it being reused again.

darthsiroftardis (approved):

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,114 @@
...
+    fi
+}
+
+function assert_node_proposed() {
+    local NODE_ID=${1}
+    local NODE_PATH=$(get_path_to_node $NODE_ID)
+    local PUBLIC_KEY_HEX=$(get_node_public_key_hex $NODE_ID)
+    local TIMEOUT=${2:-300}
+    log_step "Waiting for a node-$NODE_ID to produce a block..."
+    local OUTPUT=$(timeout $TIMEOUT tail -n 1 -f "$NODE_PATH/logs/stdout.log" | grep -o -m 1 "proposer: PublicKey::Ed25519($PUBLIC_KEY_HEX)")
```
goral09, line 73: uuu nice, there's a built-in `timeout` method in bash?

utils/nctl/sh/scenarios/itst14.sh
```diff
@@ -0,0 +1,114 @@
...
+function assert_no_equivocation() {
+    local NODE_ID=${1}
+    log_step "Checking to see if node-$NODE_ID equivocated..."
+    if ( check_faulty "$NODE_ID" ); then
+        log "Error: Node-$NODE_ID equivocated!"
+        exit 1
+    else
+        log "Node-$NODE_ID did not equivocate! yay!"
+    fi
+}
```
goral09, line 92: If `node-5` had shut down, what would be the result of this function? I think we should be also checking if other nodes did not detect an equivocation. It may happen that `node-5` detected it's a doppelganger, shut down and we won't have information about equivocation in the logs.

TomVasile: @goral09 checking for 'faulty' is the way Andreas had said a node would equivocate in a previous test. Is there a better way to check?

goral09 (approved):

PR #1084, opened 2021-03-08
NDRS-1000/Restrict private key permissions

marc-casperlabs:
This change ensures that private keys generated by the casper-client are only read/writeable by the owner. Since keys are commonly generated using root, they are protected against regular user access this way.

**Note**: The documentation needs to be updated to reflect this (keys need to be either generated as `casper` or `chown`ed).

### DIFF
--- a/node/src/crypto/asymmetric_key_ext.rs
+++ b/node/src/crypto/asymmetric_key_ext.rs
@@ -103,7 +103,7 @@ impl AsymmetricKeyExt for SecretKey {
     }
 
     fn to_file<P: AsRef<Path>>(&self, file: P) -> Result<(), Error> {
-        utils::write_file(file, self.to_pem()?).map_err(Error::SecretKeySave)
+        utils::write_private_file(file, self.to_pem()?).map_err(Error::SecretKeySave)
     }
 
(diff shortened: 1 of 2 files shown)


rob-casper:
https://docs.casperlabs.io/en/latest/node-operator/setup.html would need to be updated with 

````
sudo casper-client keygen /etc/casper/validator_keys
sudo chown -R casper:casper /etc/casper/validator_keys
```

Should we print an explicit warning after keygen that keys must have perms accessible to the user that casper-node is run as?

rob-casper:
This change will also require third party docs to be updated. eg

https://github.com/make-software/how-to-casperlabs/blob/master/docs/ubuntu/setup-validator-from-scratch.md

rob-casper:
@marc-casperlabs - docs update PR https://github.com/CasperLabs/docs/pull/140

Fraser999 (approved):
LGTM, but it'd be nice to get confirmation from a macOS user that this actually works as advertised.

momipsl:
@marc-casperlabs naughty boy you did not await feedback from my testing prior to merging.

marc-casperlabs:
> naughty boy you did not await feedback from my testing prior to merging.

busted :(

PR #1085, opened 2021-03-08
NO-TICKET: Simplify linear_chain.

afck:
This fixes rustfmt and removes some unnecessary `Event` variants.

(Found in preparation of [NDRS-892](https://casperlabs.atlassian.net/browse/NDRS-892).)

### DIFF
--- a/node/src/components/linear_chain.rs
+++ b/node/src/components/linear_chain.rs
@@ -22,7 +22,7 @@ use crate::{
             ConsensusRequest, ContractRuntimeRequest, LinearChainRequest, NetworkRequest,
             StorageRequest,
         },
-        EffectBuilder, EffectExt, EffectOptionExt, EffectResultExt, Effects, Responder,
+        EffectBuilder, EffectExt, EffectOptionExt, EffectResultExt, Effects,
     },
     protocol::Message,
     types::{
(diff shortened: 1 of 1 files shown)


node/src/components/linear_chain.rs
```diff
@@ -270,10 +243,6 @@ impl<I> LinearChain<I> {
                 // TODO: disconnect from the sender.
                 continue;
             }
-            if known_signatures.proofs.contains_key(&fs.public_key) {
-                // Don't send finality signatures we already know of.
-                continue;
-            }
```
goral09, line 276: Why did you remove this?

node/src/components/linear_chain.rs
```diff
@@ -53,12 +53,6 @@ pub enum Event<I> {
         /// The deploys' execution results.
         execution_results: HashMap<DeployHash, ExecutionResult>,
     },
-    /// A continuation for `GetBlock` scenario.
-    GetBlockResult(BlockHash, Option<Box<Block>>, I),
-    /// A continuation for `BlockAtHeight` scenario.
-    GetBlockByHeightResult(u64, Option<Box<Block>>, I),
-    /// A continuation for `BlockAtHeightLocal` scenario.
-    GetBlockByHeightResultLocal(u64, Option<Box<Block>>, Responder<Option<Block>>),
```
goral09, line 61: We never construct these events? So other components read directly from the storage? :grimacing:

afck: No, the _requests_ are being made. These `*Result` events were just used internally, sent from `LinearChain` to itself, instead of doing everything in an `async` block.

node/src/components/linear_chain.rs
afck: It's duplicated: It's already in line 236.

goral09 (approved):

PR #1092, opened 2021-03-09
EE-1131: Get rid of RootNotFound

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1131

This PR removes specialized "RootNotFound" struct-error which was moved into `engine_state::error::Error` as a separate variant.

Since RootNotFound is not a special case error anymore and now changed methods shares the same error type with other call sites there is still room for improvement in the error handling.


EdHastingsCasperLabs (approved):

henrytill (approved):
Excellent

PR #1095, opened 2021-03-09
NO-TICKET/reconnect at all costs

marc-casperlabs:
This changes the logic and removes all instance of exiting the node, favoring infinite reconnection attempts to known nodes instead.  There is a hardcoded grace time of currently two seconds for this process, this could be made into a configuration variable instead.

This tries to handle most edge cases reasonably, but errs slightly on the side of not turning a node into a turbo reconnecting spammer. Some rare edge cases arise because of this, i.e. another isolated node connecting to use during the wait for reconnection.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -637,7 +637,7 @@ where
         }
         if let Some(outgoing) = self.outgoing.remove(&peer_id) {
             trace!(our_id=%self.our_id, %peer_id, "removing peer from the outgoing connections");
-            let peer_ip = format!("{}", outgoing.peer_address.ip());
+            let peer_ip = outgoing.peer_address.to_string();
             if add_to_blocklist && !self.known_addresses.contains(&peer_ip) {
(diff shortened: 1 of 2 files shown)


node/src/components/small_network.rs
```diff
@@ -106,6 +105,9 @@ const MAX_ASYMMETRIC_CONNECTION_SEEN: u16 = 3;
 static BLOCKLIST_RETAIN_DURATION: Lazy<TimeDiff> =
     Lazy::new(|| Duration::from_secs(60 * 10).into());
 
+/// Minimum amount of time that has to pass before attempting to reconnect after isolation.
+const ISOLATION_RECONNECT_DELAY: Duration = Duration::from_secs(2);
```
EdHastingsCasperLabs, line 109: config this

goral09 (approved):

marc-casperlabs:
This has been deprecated in favor of #1099

PR #1097, opened 2021-03-09
EE-1195: Auction persist SeigniorageRecipients directly in trie

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/EE-1195

CHANGELOG

- Created an `EraValidators(EraId)` variant in the `Key` enum and related enums.
- Created an `EraValidators(SeigniorageRecipients)` variant in the `StoredValue` enum and related enums.
- Added `read_era_validators` and `write_era_validators methods` to the `auction::StorageProvider` trait.
- Reimplemented `get_seigniorage_recipients_snapshot` and `set_seigniorage_recipients_snapshot` in terms of the above methods
- Added `get_seigniorage_recipients_snapshot` to the test builder.
- Changed certain assertions in EE auction tests to account for the re-factoring

PS: @henrytill thanks for the ticket description, it helped with the PR description


execution_engine/src/core/engine_state/mod.rs
```diff
@@ -179,7 +179,10 @@ where
         let handle_payment_hash = genesis_installer.create_handle_payment()?;
 
         // Create auction
-        let auction_hash = genesis_installer.create_auction()?;
+        let auction_hash = match genesis_installer.create_auction() {
```
darthsiroftardis, line 182: This should be removed, I am aware of this mistake

darthsiroftardis: I mean, I plan to revert the change I made here

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -260,6 +260,13 @@ where
                 self.named_keys.remove(name);
                 Ok(())
             }
+            Key::EraValidators(_) => {
+                //let _era_validators_info: EraValidators =
+                // self.read_gs_typed(&era_validators_info_addr)?;
```
mpapierski, line 265:
```suggestion
```

execution_engine/src/core/runtime/auction_internal.rs
```diff
@@ -99,6 +99,41 @@ where
...
     }
+
+    fn read_era_validators(
+        &mut self,
+        era_id: EraId,
+    ) -> Result<Option<SeigniorageRecipients>, Error> {
+        match self.context.read_gs(&Key::EraValidators(era_id)) {
+            Ok(Some(StoredValue::EraValidators(recipients))) => Ok(Some(recipients)),
+            Ok(Some(_)) => {
+                println!("Word");
```
mpapierski, line 110:
```suggestion
```

mpapierski (commented):
Looks good so far. Now that seigniorage recipients are written under `Key::EraValidators(era_id)` you should be able to optimize this https://github.com/darthsiroftardis/casper-node/blob/117e71b3c81c017f9b83c816069297d90b4b066a/types/src/system/auction/mod.rs#L80-L89 and call into `provider.read_era_validators(era_index)` instead of getting all the Key::EraValidators entries just to get one specific value from it.

types/src/system/auction/detail.rs
```diff
@@ -143,7 +143,23 @@ pub fn get_seigniorage_recipients_snapshot<P>(
...
+            _ => return Err(Error::InvalidKeyVariant),
+        };
+        let recipient = match provider.read_era_validators(era_id)? {
+            Some(recipient) => recipient,
+            None => return Err(Error::ValidatorNotFound),
+        };
+        ret.insert(era_id, recipient);
+    }
+
+    Ok(ret)
```
mpapierski, line 162: This function returns all seigniorage recipients, and `set_seigniorage_recipients_snapshot` writes all passed recipients. 
This might have some side effects here:
https://github.com/darthsiroftardis/casper-node/blob/117e71b3c81c017f9b83c816069297d90b4b066a/types/src/system/auction/mod.rs#L425-L444. First it gets all the seigniorage recipients, then trims it to only have N entries, and then it writes snapshot back. 

I'm thinking it will be more efficient if we'd read only a fixed amount of seigniorage recipients (directly through read_era_validators, not iterating through keys) limited by the snapshot size?

execution_engine/src/core/runtime/auction_internal.rs
darthsiroftardis: Removed (stupid debugging statments :) )

execution_engine/src/core/runtime_context/mod.rs
darthsiroftardis: Removed comments

types/src/system/auction/detail.rs
darthsiroftardis: @henrytill Thoughts??

henrytill: They key piece of @mpapierski's critique is correct: we need to limit the number of keys we read so that we aren't returning _all_ the eras, just the "window" that was there before.  The window's size is essentially the auction delay IIRC.

henrytill (changes requested):
Looking good so far, but an important consideration around the size of the map returned by `get_seigniorage_recipients_snapshot` (also affects the `TestBuilder` method).

types/src/system/auction/detail.rs
henrytill: > Edit: I think I take most of it back. We might not need to get the seigniorage snapshot amnd set the seigniorage snapshot there at all. Perhaps just write "recipients" map under Key::EraValidators(delayed_era)?

@mpapierski, not sure I fully grasp your thinking here, can you elaborate?

henrytill:
Also note that because #1089 was merged to `dev`, you'll need to use `PaddedEraId` in the `ToBytes` and `FromBytes` impls for `Key`.

types/src/system/auction/detail.rs
mpapierski: https://github.com/darthsiroftardis/casper-node/blob/117e71b3c81c017f9b83c816069297d90b4b066a/types/src/system/auction/mod.rs#L425-L444 this code could probable become just:

```rust
  let delayed_era = era_id + auction_delay; // as before
  // Update seigniorage recipients for current era
  // ... 
  let reicipients = { todo!("calculate recipients as before") };
  self.write_era_validators(delayed_era, recipients)?;
```

So my thinking is before we had to use "SeigniorageRecipientsSnapshot" container which mapped EraId into SeigniorageRecipients. Then read it from global state from single key, then extend it with new seigniorage recipients, and then trim it to contain last N eras only and write it all again under the same key. With this change we could just write new `recipients` into global state under `Key::EraValidators(delayed_era)` without reading past data. Wouldn't that work?

henrytill: > They key piece of @mpapierski's critique is correct: we need to limit the number of keys we read so that we aren't returning all the eras, just the "window" that was there before. The window's size is essentially the auction delay IIRC.

Wanted to clarify that we are talking about the _last_ N eras here with this window, so the highest `EraId`s.

henrytill: > https://github.com/darthsiroftardis/casper-node/blob/117e71b3c81c017f9b83c816069297d90b4b066a/types/src/system/auction/mod.rs#L425-L444 this code could probable become just:
> 
> ```rust
>   let delayed_era = era_id + auction_delay; // as before
>   // Update seigniorage recipients for current era
>   // ... 
>   let reicipients = { todo!("calculate recipients as before") };
>   self.write_era_validators(delayed_era, recipients)?;
> ```
> 
> So my thinking is before we had to use "SeigniorageRecipientsSnapshot" container which mapped EraId into SeigniorageRecipients. Then read it from global state from single key, then extend it with new seigniorage recipients, and then trim it to contain last N eras only and write it all again under the same key. With this change we could just write new `recipients` into global state under `Key::EraValidators(delayed_era)` without reading past data. Wouldn't that work?

Ah yes, absolutely!  I had missed the fact that @darthsiroftardis wasn't doing that because it didn't show up in the diff, so thanks for pointing this out.  

@darthsiroftardis we don't need to read the whole map and write it again in the code that @mpapierski links to.  Simply write the new recipients at `delayed_era` as he suggests.

mpapierski (approved):
Looks good 👍 

One more nitpick which just occurred to me: `Auction::read_seigniorage_recipients` is calling `get_seigniorage_recipients_snapshot` to get a collection just to get one seigniorage recipients from it. This method could just call `provider.read_era_validators(era_id)`

execution_engine_testing/test_support/src/internal/wasm_test_builder.rs
```diff
@@ -896,6 +896,36 @@ where
...
+
+        let reader = tracking_copy.reader();
+
+        let era_ids = reader
+            .keys_with_prefix(correlation_id, &[KeyTag::EraValidators as u8])
+            .unwrap_or_default();
+
+        let mut ret = BTreeMap::new();
+
+        for era_id in era_ids.into_iter() {
```
henrytill, line 917: We want to apply the same windowing here that you do below in `get_seigniorage_recipients_snapshot`: we don't want to return the validator set for every era, just the current era and every era through the auction delay.

henrytill (changes requested):
Getting close!  Looks great, but I think we have one more change to make.

execution_engine_testing/test_support/src/internal/wasm_test_builder.rs
darthsiroftardis: Ahhh, actually @mpapierski and I dicussed that and decided that for the test_builder it can remain as it

henrytill: Okay fair.

henrytill (approved):
Great work!

PR #1098, opened 2021-03-09
NDRS-925 contract runtime optimization

dwerner:
This PR combines NDRS-925 part 1+2 - it combines the block_executor with contract_runtime and reduces the number of messages sent back to the reactor to handle execute+commit for each deploy within the block being executed.


EdHastingsCasperLabs (approved):

xcthulhu (approved):

node/src/reactor/initializer.rs
```diff
@@ -47,13 +44,14 @@ pub enum Event {
     #[from]
     Chainspec(chainspec_loader::Event),
 
-    /// Storage event.
+    /// Storage event.:65
```
goral09, line 47:
```suggestion
    /// Storage event.
```

node/src/effect/announcements.rs
```diff
@@ -204,27 +204,52 @@ where
...
 
-/// A BlockExecutor announcement.
+/// A ContractRuntime announcement.
 #[derive(Debug)]
-pub enum BlockExecutorAnnouncement {
+pub enum ContractRuntimeAnnouncement {
     /// A new block from the linear chain was produced.
-    LinearChainBlock {
-        /// The block.
+    LinearChainBlock(Box<LinearChainBlock>),
```
goral09, line 211: Maybe it's a matter of naming but _contract runtime_ producing _linear chain blocks_ doesn't sound good. Maybe we should rename `ContractRuntime` now. Different PR though.

node/src/components/contract_runtime.rs
```diff
@@ -32,42 +36,129 @@ use casper_execution_engine::{
...
+        FinalizedBlock, NodeId,
+    },
     utils::WithDir,
     NodeRng, StorageConfig,
 };
 
+/// Contract runtime component event.
+#[derive(Debug, From)]
+pub enum Event {
+    /// A request made of the contract runtime component.
```
goral09, line 62:
```suggestion
    /// A request made for the contract runtime component.
```
?

node/src/components/contract_runtime.rs
```diff
@@ -202,362 +298,341 @@ where
...
+                    ContractRuntimeRequest::ExecuteBlock(finalized_block) => {
+                        debug!(?finalized_block, "execute block");
+                        effect_builder
+                            .get_block_at_height_local(finalized_block.height())
+                            .event(move |maybe_block| {
+                                maybe_block.map(Box::new).map_or_else(
+                                    || Event::BlockIsNew(Box::new(finalized_block)),
+                                    Event::BlockAlreadyExists,
+                                )
+                            })
```
goral09, line 534: If you put it inside the `async move` block then we should get away without the event?

node/src/components/contract_runtime.rs
```diff
@@ -202,362 +298,341 @@ where
...
-                    responder.respond(result).await
-                }
-                .ignore()
+            Event::BlockAlreadyExists(block) => effect_builder
+                .announce_block_already_executed(*block)
+                .ignore(),
+            // If we haven't executed the block before in the past (for example during
+            // joining), do it now.
+            Event::BlockIsNew(finalized_block) => {
+                self.get_deploys(effect_builder, *finalized_block)
```
goral09, line 579: We are in the `async fn` so we should be able to `async move { self.get_deploys.await }` and not have to send event to ourselves, right?

node/src/components/contract_runtime.rs
```diff
@@ -202,362 +298,341 @@ where
...
-                    .expect("should run");
-                    trace!(?result, "step response");
-                    responder.respond(result).await
+            Event::Result(contract_runtime_result) => match *contract_runtime_result {
+                ContractRuntimeResult::GetDeploysResult {
+                    finalized_block,
+                    deploys,
+                } => {
+                    trace!(total = %deploys.len(), ?deploys, "fetched deploys");
+                    self.handle_get_deploys_result(effect_builder, finalized_block, deploys)
```
goral09, line 587: Same here.

node/src/components/contract_runtime.rs
```diff
@@ -202,362 +298,341 @@ where
...
-                            .put_trie_and_find_missing_descendant_trie_keys(correlation_id, &*trie);
-                        metrics.put_trie.observe(start.elapsed().as_secs_f64());
-                        result
-                    })
-                    .await
-                    .expect("should run");
-                    trace!(?result, "put_trie response");
-                    responder.respond(result).await
+
+                ContractRuntimeResult::RunStepResult { mut state, result } => {
```
goral09, line 611: Since we're not trying to access `self.state`, we should be able to inline the content of this where the `RunStepResult` event is sent.

goral09:
I haven't finished reviewing but I think we should start with removing as many `Event`s sent to the component itself as possible and move it to `async { }` blocks inside `handle_event` method.

xcthulhu:
> I haven't finished reviewing but I think we should start with removing as many Events sent to the component itself as possible and move it to async { } blocks inside handle_event method.

I suspect @dwerner agrees, but we should probably merge sooner rather than later. It has a lot of conflicts.  Maybe we can make a follow up PR?

dwerner:
> > I haven't finished reviewing but I think we should start with removing as many Events sent to the component itself as possible and move it to async { } blocks inside handle_event method.
> 
> I suspect @dwerner agrees, but we should probably merge sooner rather than later. It has a lot of conflicts. Maybe we can make a follow up PR?

Yeah I'm happy to do a follow-up, but merge conflicts are an issue with this PR as other changes accumulate in dev. I would vote to merge.

goral09:
> > > I haven't finished reviewing but I think we should start with removing as many Events sent to the component itself as possible and move it to async { } blocks inside handle_event method.
> > 
> > 
> > I suspect @dwerner agrees, but we should probably merge sooner rather than later. It has a lot of conflicts. Maybe we can make a follow up PR?
> 
> Yeah I'm happy to do a follow-up, but merge conflicts are an issue with this PR as other changes accumulate in dev. I would vote to merge.

You should get approval from someone from the EE team. Henry and/or Michał.

dwerner:
Tested this with NCTL at the latest merge from dev. As per our discussion in sprint planning today, I'm going to merge this to dev.

PR #1099, opened 2021-03-09
NO-TICKET/reconnect at all costs big (bigger changeset)

marc-casperlabs:
This builds upon #1095 and is a larger changeset, attempting to be even more robust and performant. It changes substantially more code, so more testing on a large scale would be required for this one to merge.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -211,7 +211,7 @@ where
         for address in &cfg.known_addresses {
             match utils::resolve_address(address) {
                 Ok(known_address) => {
-                    if known_addresses.insert(known_address) {
+                    if !known_addresses.insert(known_address) {
                         warn!(%address, resolved=%known_address, "ignoring duplicated known address");
                     };
                 }
@@ -357,7 +357,7 @@ where
         for &address in &self.known_addresses {
             self.pending.insert(address, now);
 
-            // Add an effect to connect to the known address it.
+            // Add an effect to connect to the known address.
             effects.extend(
                 connect_outgoing(
                     address,
@@ -434,7 +434,7 @@ where
     /// Sweep and timeout pending connections.
     ///
     /// This is a reliability measure that sweeps pending connections, since leftover entries will
-    /// block any renewed connections attempts to a specific address.
+    /// block any renewed connection attempts to a specific address.
     ///
     /// In 100% bug free code, this would not be necessary, so this is in here as a stop-gap measure
     /// to avoid locking up a node with unremoved pending connections.
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -868,14 +868,11 @@ where
 
     /// Returns whether or not this node has been disconnected from all known nodes.
     fn is_not_connected_to_any_known_address(&self) -> bool {
-        if self
-            .pending
-            .intersection(&self.known_addresses)
-            .next()
-            .is_some()
-        {
-            // We have at least one pending connection to a known node, exit early.
-            return false;
+        for addr in self.pending.keys() {
+            if self.known_addresses.contains(addr) {
+                // We have at least one pending connection to a known node, exit early.
+                return false;
+            }
         }
 
         for (_node_id, outgoing) in &self.outgoing {
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -180,7 +180,7 @@ where
     net_metrics: NetworkingMetrics,
 
     /// Known addresses for this node.
-    known_addresses: Vec<SocketAddr>,
+    known_addresses: HashSet<SocketAddr>,
 }
 
 impl<REv, P> SmallNetwork<REv, P>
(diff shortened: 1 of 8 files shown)


dwerner (approved):
lgtm

node/src/components/small_network.rs
```diff
@@ -360,6 +347,41 @@ where
...
+    ///
+    /// Will schedule another reconnection if no DNS addresses could be resolved.
+    fn connect_to_known_addresses(&mut self) -> Effects<Event<P>> {
+        let mut effects = Effects::new();
+
+        let now = Instant::now();
+        for &address in &self.known_addresses {
+            self.pending.insert(address, now);
+
+            // Add an effect to connect to the known address it.
```
Fraser999, line 360:
```suggestion
            // Add an effect to connect to the known address.
```

node/src/components/small_network.rs
```diff
@@ -409,6 +431,45 @@ where
         }
     }
 
+    /// Sweep and timeout pending connections.
+    ///
+    /// This is a reliability measure that sweeps pending connections, since leftover entries will
+    /// block any renewed connections attempts to a specific address.
```
Fraser999, line 437:
```suggestion
    /// block any renewed connection attempts to a specific address.
```

node/src/components/small_network/message.rs
```diff
@@ -1,17 +1,32 @@
...
 use serde::{Deserialize, Serialize};
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub enum Message<P> {
-    Handshake { network_name: String },
+    Handshake {
+        /// Network we are connected to.
+        network_name: String,
+        /// The public address of the node connecting.
+        public_address: SocketAddr,
```
goral09, line 14: I don't understand why we need this information sent to us – we know from what IP and port combination this message comes from so why do we need to send it as well?

node/src/components/small_network.rs
```diff
@@ -590,24 +647,29 @@ where
...
         // outgoing connection, hence no need to call `self.remove()`.
-        if let Some(err) = error {
+        if let Some(ref err) = error {
             warn!(our_id=%self.our_id, %peer_address, %err, "outgoing connection failed");
         } else {
             warn!(our_id=%self.our_id, %peer_address, "outgoing connection closed");
         }
 
-        Effects::new()
+        effects
```
goral09, line 672: If you rearrange the code in this method and log at the beginning, you won't need to create the `mut effects` and just regurn from the `if`'s.

node/src/components/small_network.rs
```diff
@@ -710,8 +775,12 @@ where
                     self.update_peers_metric();
                     return remove;
                 }
+
+                // This speeds up the connection process, but masks potential bugs in the gossiper.
+                let effects = self.connect_to_peer_if_required(public_address);
                 self.update_peers_metric();
-                Effects::new()
+
+                effects
```
goral09, line 783: If you call `update_peers_metric()` first, then you can return with `self.connect_to_peer`.

node/src/components/small_network.rs
```diff
@@ -811,12 +875,23 @@ where
...
+            if self.pending.contains_key(&known_address) {
+                return false;
+            }
+
+            if self
+                .outgoing
+                .values()
+                .any(|outgoing_connection| outgoing_connection.peer_address == known_address)
+            {
+                return false;
```
goral09, line 890: Why is the asymetric connection considered as "connected"?

node/src/components/small_network/message.rs
marc-casperlabs: A an outgoing TCP connection is made from a random port. Let's assume we have two nodes `A` with IP `2.2.2.2` and `B` with IP `7.7.7.7`, both listening on a fictional node-port `9999`.

Listening addresses are thus `2.2.2.2:9999` and `7.7.7.7:9999`.

If `A` connects to be, the OS of A will allocate a random port for the outgoing connection, e.g. `40318`. So the connection will be

```
[A] 2.2.2.2:40318   ----->  7.7.7.7:9999 [B]
```

Thus the peer address of the incoming connection will show up as `2.2.2.2:40318`. However, Node B needs to connect back to `2.2.2.2:9999` instead, so we need to send the address.

When `B` connects back, it will also be from a random port, resulting in something like

```
[A] 2.2.2.2:40318   ----->  7.7.7.7:9999 [B]
[A] 2.2.2.2:9999   <-----   7.7.7.7:50223 [B]
```

node/src/components/small_network.rs
marc-casperlabs: Good catch! Although it would change the output order, this seems reasonable.

At the current point though I am loathe to make this kind of change for fear of accidentally introducing a regression. Can we maybe postpone this?

node/src/components/small_network.rs
marc-casperlabs: Unfortunately, `connect_to_peer_if_required` manipulates `self.pending`, which is read by `update_peers_metric` via `self.peers`, so we'd be off by one by this change.

node/src/components/small_network.rs
marc-casperlabs: This might be a naming problem; this function used to be called `is_isolated`. Our conditions are that if we lost the outgoing connection to all known peers we are what we formerly called "isolated" and attempt to remedy that.

Incoming connections are ignored to avoid issues where a stray incoming connection prevents us from triggering reconnection.

node/src/components/small_network.rs
goral09: Output order of what? Logs?

node/src/components/small_network.rs
marc-casperlabs: I stand corrected, you're right!

node/src/components/small_network.rs
goral09: Thanks

goral09 (approved):

Fraser999 (approved):

PR #1102, opened 2021-03-10
NDRS-892: Remove HandleLinearBlock request.

afck:
We want to move the era supervisor out of the joiner reactor. One of the joiner's components depending on the era supervisor is the linear chain, because it calls `handle_linear_chain_block` and `is_bonded_validator`. This PR removes the former: While we are joining, we don't need the era supervisor to create new finality signatures or to initialize new eras. ~It also removes the call from the block executor: We used to make the request there if an executed block was already stored, to initialize a new era, but initial eras are now initialized all at once when transitioning from the joiner to the validator reactor.~

https://casperlabs.atlassian.net/browse/NDRS-892

### DIFF
--- a/node/src/components/block_executor.rs
+++ b/node/src/components/block_executor.rs
@@ -33,12 +33,12 @@ use crate::{
     },
     crypto::hash::Digest,
     effect::{
-        announcements::BlockExecutorAnnouncement,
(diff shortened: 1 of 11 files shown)


afck:
Closing in favor of #1114.

PR #1104, opened 2021-03-10
NO-TICKET: Fix Clippy warnings, formatting and remove a redundant argument.

afck:
Formatting in `linear_chain_sync.rs` was broken — if it can't fix it, `rustfmt` won't touch it —, there were Clippy warnings in that file, and `BlockAdded` had a redundant `block_hash` argument: Since the `Block` contains its hash and doesn't need to recompute it, this is not needed.
(Found while working on [NDRS-892](https://casperlabs.atlassian.net/browse/NDRS-892).)


goral09 (approved):

goral09:
bors merge

PR #1109, opened 2021-03-10
NO-TICKET: Limit not in scope on osx

khjcph:
PR fixes building casper-node on target osx by not setting constant TARGET_OPEN_FILES_LIMIT


Fraser999 (approved):
Thanks for the PR - looks good to me.

mpapierski (approved):
Thanks for the PR @khjcph!

PR #1113, opened 2021-03-11
NO-TICKET: remove failed bootstrap address from pending set

Fraser999:
This PR effectively reverts [0dd0fdd](https://github.com/CasperLabs/casper-node/commit/0dd0fdd9ea87e88213821732857fba0dd2be3fe8) which created an issue whereby failed bootstrap connections would live forever in the small_network's `pending` set, meaning these bootstrap peers could never be connected to in that session.


EdHastingsCasperLabs (approved):

Fraser999:
bors merge

thamps-casp:
bors merge

goral09:
@Fraser999 we should probably also target this at `0.9.3`, correct?

PR #1115, opened 2021-03-11
NDRS-1010: Don't create a new era when we should shutdown for an upgrade.

afck:
https://casperlabs.atlassian.net/browse/NDRS-1010


afck:
It passed the `utils/nctl/sh/scenarios/sync_upgrade_test.sh` integration test.

goral09 (approved):

PR #1118, opened 2021-03-11
NDRS-1007/OS supplied test port

marc-casperlabs:
This hopefully reduces the number of collisions of random ports **in tests** by using OS allocated ports instead of truly random ones. This relies on the technically unspecified, but widely implemented (especially on Linux) property that requested "unused" ports are randomly selected.

A few random tests showed a floor of at least a few hundred port numbers before hitting a duplicate, so I am hopeful that this will  not hit many duplicates.

**This should not affect any production code**, all changes are contained to files that are gated behind `#[cfg(test)]`.


goral09 (approved):

goral09:
bors merge

PR #1122, opened 2021-03-12
HWY-267: Fix arithmetic overflow.

afck:
When determining whether enough nodes are online, we can't use
```
    (total_weight + target_ftt) / 2
```
because `total_weight` could already be close to `u64::MAX`.

https://casperlabs.atlassian.net/browse/HWY-267

### DIFF
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -226,7 +226,7 @@ impl<C: Context> ActiveValidator<C> {
     /// Returns whether enough validators are online to finalize values with the target fault
     /// tolerance threshold, always counting this validator as online.
     fn enough_validators_online(&self, state: &State<C>, now: Timestamp) -> bool {
-        let target_quorum = (state.total_weight() + self.target_ftt) / 2;
+        let target_quorum = state.total_weight() / 2 + self.target_ftt / 2;
         let online_weight: Weight = state
             .weights()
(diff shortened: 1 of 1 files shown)


goral09 (approved):

node/src/components/consensus/highway_core/active_validator.rs
```diff
@@ -226,7 +226,7 @@ impl<C: Context> ActiveValidator<C> {
     /// Returns whether enough validators are online to finalize values with the target fault
     /// tolerance threshold, always counting this validator as online.
     fn enough_validators_online(&self, state: &State<C>, now: Timestamp) -> bool {
-        let target_quorum = (state.total_weight() + self.target_ftt) / 2;
+        let target_quorum = state.total_weight() / 2 + self.target_ftt / 2;
```
Fraser999, line 229: I'd have maybe added a comment to try to stop someone in the future doing the obvious refactor back to what it was before?

Also, I guess there's no easy way to add a regression test for this case is there?

afck: Sorry, I'll add a comment later!

To test it, maybe we could just change one of our tests to have `total_weight` close to `u64::MAX`…

Maybe I'll make another PR for those.

afck: https://github.com/CasperLabs/casper-node/pull/1126

PR #1125, opened 2021-03-12
NO-TICKET: use actual exit code returned by validator reactor

Fraser999:
This PR fixes an issue where the exit code returned by the validator reactor is ignored.

It also borrows code from #1070 which ensures all destructors are run before exiting `main`.

### DIFF
--- a/node/src/app/cli.rs
+++ b/node/src/app/cli.rs
@@ -7,7 +7,6 @@ pub mod arglang;
 use std::{
     env, fs,
     path::{Path, PathBuf},
-    process,
     str::FromStr,
 };
 
@@ -139,7 +138,7 @@ impl FromStr for ConfigExt {
 
 impl Cli {
     /// Executes selected CLI command.
-    pub async fn run(self) -> anyhow::Result<()> {
+    pub async fn run(self) -> anyhow::Result<i32> {
         match self {
             Cli::Validator { config, config_ext } => {
                 // Setup UNIX signal hooks.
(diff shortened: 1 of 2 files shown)


node/src/app/cli.rs
```diff
@@ -173,8 +172,7 @@ impl Cli {
                 // initializer2_runner.run(&mut rng).await;
 
                 match initializer_runner.run(&mut rng).await {
-                    ReactorExit::ProcessShouldExit(ExitCode::Success) => return Ok(()),
-                    ReactorExit::ProcessShouldExit(exit_code) => process::exit(exit_code as i32),
```
goral09, line 177: What was `process::exit` doing then? If we don't need it anymore.

node/src/app/cli.rs
```diff
@@ -173,8 +172,7 @@ impl Cli {
...
+                    ReactorExit::ProcessShouldExit(exit_code) => return Ok(exit_code as i32),
```
goral09, line 175: Are we going to log the exit code anywhere? If not, then maybe add it to the `cli.rs`.

goral09 (approved):
Please log the `ExitCode` somewhere in the node.

node/src/app/cli.rs
Fraser999: It exits the process immediately without running destructors.  This PR uses Marc's approach of passing the exit code out to `main` so we can drop everything and only then call `process::exit`.

goral09:
Once this is merged, let's port it to 0.9.2 and 0.9.3 release branches.

node/src/app/cli.rs
Fraser999: Done in [ad76d63](https://github.com/CasperLabs/casper-node/pull/1125/commits/ad76d638d8701e54d764214ec0bcba7463f8134f).

Fraser999:
bors merge

afck (approved):

Fraser999:
bors merge-

Fraser999:
bors merge

Fraser999:
bors merge-

Fraser999:
bors merge

goral09:
Can you create PR against `release-0.9.3`?

Fraser999:
bors merge

PR #1127, opened 2021-03-12
EE-1134: Store system contracts at fixed addresses

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1134

This PR changes addresses of system contracts so they are stored at fixed contract hashes.

- Removes mint/handlepayment/auction/standardpayment contract hashes from ProtocolData as there is no need to store them since they are known in advance.
- Contract API "get_system_contract" is removed as the contract hashes are known in advance.
- A bit of refactoring to benefit from exhaustive pattern matching which should make adding new reserved system contracts in future much easier

The convention is to use a value from 0..255 for each system contract left padded with 0s. Address is also conveniently a valid number once you interpret it as a big endian bytes (i.e. U256 or smaller).

Currently occupied reserved contract hashes:

| System contract | Address |
| --- | --- |
| Unused[1] | 0x0 |
| Mint | 0x00(...)0001 |
| HandlePayment | 0x00(...)0002 |
| StandardPayment | 0x00(...)0003 |
| Auction | 0x00(...)0004 |

[1]: Unused but still reserved to disambiguate it from a "default address" / "system addr"


henrytill (commented):
Looks good.  Definitely glad to see the system contract cache finally get deleted.  We should wait for @EdHastingsCasperLabs to weigh in before deciding how to proceed.

Counter-proposal for addresses:

| System contract | Address |
|-----------------|---------|
| Unused[1]   |	0x0   |
| Mint	     | 0x00(...)0001 |
| Auction	     | 0x00(...)0002 |
| HandlePayment    | 0x00(...)0003 |
| StandardPayment | 0x00(...)0004 |

mpapierski:
Closed in favor of minimal version in #1150

PR #1130, opened 2021-03-12
HWY-266: Shut down era supervisor in case it hasn't received any messages for the last era in a while

afck:
Right before an upgrade it can happen that some nodes lag behind and, since the others restart, fail to find peers who complete their protocol state so they can see the switch block finalized. This change will make them restart after a while, so they can download the finalized executed blocks and then trigger the upgrade as well.

https://casperlabs.atlassian.net/browse/HWY-266

### DIFF
--- a/node/src/components/consensus/config.rs
+++ b/node/src/components/consensus/config.rs
@@ -24,6 +24,10 @@ pub struct Config {
     pub unit_hashes_folder: PathBuf,
     /// The duration for which incoming vertices with missing dependencies are kept in a queue.
     pub pending_vertex_timeout: TimeDiff,
+    /// If the current era's protocol state has not progressed for this long, shut down.
+    pub standstill_timeout: TimeDiff,
+    /// Log inactive or faulty validators periodically, with this interval.
+    pub log_participation_interval: TimeDiff,
     /// The maximum number of blocks by which execution is allowed to lag behind finalization.
     /// If it is more than that, consensus will pause, and resume once the executor has caught up.
     pub max_execution_delay: u64,
@@ -35,6 +39,8 @@ impl Default for Config {
             secret_key_path: External::Missing,
             unit_hashes_folder: Default::default(),
             pending_vertex_timeout: "10sec".parse().unwrap(),
+            standstill_timeout: "1min".parse().unwrap(),
+            log_participation_interval: "10sec".parse().unwrap(),
             max_execution_delay: 3,
         }
     }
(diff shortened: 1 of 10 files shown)


node/src/components/consensus/protocols/highway.rs
```diff
@@ -41,7 +41,11 @@ use self::{round_success_meter::RoundSuccessMeter, synchronizer::Synchronizer};
 
 /// Never allow more than this many units in a piece of evidence for conflicting endorsements,
 /// even if eras are longer than this.
-const MAX_ENDORSEMENT_EVIDENCE_LIMIT: u64 = 10000;
+const MAX_ENDORSEMENT_EVIDENCE_LIMIT: u64 = 10_000;
+/// The number of milliseconds after which an alert is raised if no progress was made.
+const STANDSTILL_ALERT_MILLIS: u64 = 5 * 60_000;
```
goral09: So it will shut down the node if it hasn't changed an era in 5 minutes?

afck: Yes… but it will only stop changing the era itself once it goes into pause mode, which is three maximum-length eras after it notices that not enough weight is online anymore.

node/src/components/consensus/config.rs
```diff
@@ -24,6 +24,10 @@ pub struct Config {
     pub unit_hashes_folder: PathBuf,
     /// The duration for which incoming vertices with missing dependencies are kept in a queue.
     pub pending_vertex_timeout: TimeDiff,
+    /// If the current era's protocol state has not progressed for this long, shut down.
+    pub standstill_timeout: TimeDiff,
+    /// Log inactive or faulty validators periodically, with this interval.
+    pub log_participation_interval: TimeDiff,
```
goral09, line 30: Hopefully, my PR merges before this so you should update this and move to highway-specific config.

afck: I know, I already thought about it. But I selflessly, heroically, approved it anyway so you could merge first. :innocent:

node/src/components/consensus/protocols/highway/tests.rs
```diff
@@ -174,44 +178,65 @@ fn send_a_wire_unit_with_too_small_a_round_exp() {
...
+        _ => panic!("Unexpected outcomes: {:?}", outcomes),
+    }
+
+    // If after another timeout, the state has not changed, an alert is raised.
+    now += standstill_timeout;
+    let outcomes = highway_protocol.handle_timer(now, TIMER_ID_STANDSTILL_ALERT);
+    match &*outcomes {
+        [ProtocolOutcome::StandstillAlert] => (),
+        _ => panic!("Unexpected outcomes: {:?}", outcomes),
+    }
```
goral09, line 239: Would
```rust
matches!(&*outcomes, [ProtocolOutcome::StandstillAlert])
```
pass?

goral09 (approved):

goral09:
The ultimate test will be building the package and trying to run it with the `nctl`. Can you do that before merging?

afck:
I ran it with nctl and the node stopped as expected.

PR #1135, opened 2021-03-15
NDRS-349/fix tracing issues

marc-casperlabs:
This PR does a little cleanup and fixes a few issues with `async` tracing instrumentation. It also adds an additional span around the runner initialization.

https://tracing.rs/tracing/struct.span#method.enter has some background information on why these changes were necessary.

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -52,7 +52,7 @@ use quanta::{Clock, IntoNanoseconds};
 use serde::Serialize;
 use signal_hook::consts::signal::{SIGINT, SIGQUIT, SIGTERM};
 use tokio::time::{Duration, Instant};
-use tracing::{debug, debug_span, error, info, trace, warn};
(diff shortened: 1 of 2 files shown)


Fraser999 (approved):

goral09:
What issues does this PR solve specifically? Apart from the cleanup.

marc-casperlabs:
> What issues does this PR solve specifically? Apart from the cleanup.

The core issues is that tracing spans were entered across an `await` point, which causes the traces to be randomly wrong --- the current span is stored on a thread local, while the task might jump threads in between.

I've written a blog post about that, look for the "async users beware" section in https://compilersaysno.com/posts/tracing-for-the-impatient/.

The worst offender in our code was here: https://github.com/CasperLabs/casper-node/pull/1135/files#diff-f37f8af7d3f5387726dc2460b5d53b8a4092e62a97a62c6491c7265b5f496801L500-L501

PR #1137, opened 2021-03-15
NDRS-811/cleanup memory usage dump in logs

marc-casperlabs:
This changes the behavior of memory metrics collection in the validator slightly. First, it removes the special casing for the `event_count == 0`, as `event_count % ANYTHING` should already cover this. Second, it simply subtracts the delay from the initial timestamp, if possible, to achieve the same effect.

A second change is that `last_metrics` is updated _after_ metrics collection; this skews the collection interval a bit but ensures that we are guaranteed to get `event_metrics_min_delay` time of uninterrupted (by memory metrics collection) event handling.

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -506,14 +506,14 @@ where
 
         // Update metrics like memory usage and event queue sizes.
         if self.event_count % self.event_metrics_threshold == 0 {
-            let now = Instant::now();
-
             // We update metrics on the first very event as well to get a good baseline.
-            if now.duration_since(self.last_metrics) >= self.event_metrics_min_delay
-                || self.event_count == 0
-            {
+            if self.last_metrics.elapsed() >= self.event_metrics_min_delay {
                 self.reactor.update_metrics(event_queue);
-                self.last_metrics = now;
+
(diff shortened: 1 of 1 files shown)


node/src/reactor.rs
```diff
@@ -757,13 +757,15 @@ impl Runner<InitializerReactor> {
...
+        let event_metrics_min_delay = Duration::from_secs(30);
+        let now = Instant::now();
         Ok(Runner {
             scheduler,
             reactor,
             event_count: 0,
             metrics: RunnerMetrics::new(&registry)?,
-            last_metrics: Instant::now(),
-            event_metrics_min_delay: Duration::from_secs(30),
+            last_metrics: now.checked_sub(event_metrics_min_delay).unwrap_or(now),
```
Fraser999, line 767: A comment explaining why we're doing this subtraction would help.  Although this avoids the special case of `|| self.event_count == 0`, the special case at least made it completely obvious what was happening :)

Fraser999 (approved):
Looks good, apart from the one non-blocker comment.

node/src/reactor.rs
marc-casperlabs: :+1: I'll add it!

PR #1139, opened 2021-03-15
HWY-268: Enable integer_arithmetic Clippy warnings to prevent overflows.

afck:
This enables the warnings for all arithmetic operations in `highway_core` that could potentially overflow. It also replaces them with safe (e.g. saturating) operations in some places.

https://casperlabs.atlassian.net/browse/HWY-268


goral09 (approved):

fizyk20 (approved):
Looking good :+1:

PR #1140, opened 2021-03-15
NDRS-821: upgrade tokio

Fraser999:
Ref: [NDRS-821](https://casperlabs.atlassian.net/browse/NDRS-821), [NDRS-977](https://casperlabs.atlassian.net/browse/NDRS-977), [NDRS-984](https://casperlabs.atlassian.net/browse/NDRS-984), [NDRS-985](https://casperlabs.atlassian.net/browse/NDRS-985), [NDRS-986](https://casperlabs.atlassian.net/browse/NDRS-986), [NDRS-995](https://casperlabs.atlassian.net/browse/NDRS-995) and [NDRS-998](https://casperlabs.atlassian.net/browse/NDRS-998).

This PR upgrades tokio and those other dependencies which also require upgrading as a result.

A significant cause for change is that the `NodeId` enum has again been made `Copy` since upgrading libp2p.  A further notable change is that small_network now uses Bincode rather than MessagePack for framing messages under the hood.


goral09:
Is it possible to separate changes to the `small_network` and handshake process to a separate PR? It shouldn't be related to the upgrade task at all.

Fraser999:
> Is it possible to separate changes to the small_network and handshake process to a separate PR? It shouldn't be related to the upgrade task at all.

I'm think with the exception of the addition of `Copy` to `NodeId`, everything else is really all tied to upgrading tokio - as in it either won't compile or it won't run without it.

node/src/components/network.rs
```diff
@@ -397,17 +398,17 @@ impl<REv: ReactorEventT<P>, P: PayloadT> Network<REv, P> {
             );
         }
 
-        for &peer_id in &peer_ids {
-            self.send_message(peer_id.clone(), payload.clone());
+        for &&peer_id in &peer_ids {
```
mpapierski, line 401: Nit: odd usage of && - if I understand correctly `peer_ids` could be `.copied()` before `.filter()` and then we can have `for peer_id in peer_ids`  (or `for &peer_id in &peer_ids`)

mpapierski (approved):
Looks good for me dependencies wise (with one non-blocking nit on syntax). Much deeper change than my original attempt some time ago - Best to have another green check from someone more knowledgeable about the node core.

node/src/components/network/event.rs
```diff
@@ -20,7 +20,7 @@ use crate::{
 use core::mem;
 
 const _NETWORK_EVENT_SIZE: usize = mem::size_of::<Event<Message>>();
-const_assert!(_NETWORK_EVENT_SIZE < 89);
+const_assert!(_NETWORK_EVENT_SIZE < 178);
```
goral09, line 23: Why this change?

node/src/components/event_stream_server.rs
```diff
@@ -58,13 +61,41 @@ pub(crate) struct EventStreamServer {
...
         let (sse_data_sender, sse_data_receiver) = mpsc::unbounded_channel();
-        let builder = utils::start_listening(&config.address)?;
+
+        // Event stream channels and filter.
+        let (broadcaster, new_subscriber_info_receiver, sse_filter) =
+            sse_server::create_channels_and_filter(config.broadcast_channel_size);
+
+        let (shutdown_sender, shutdown_receiver) = oneshot::channel::<()>();
+
+        let (address, server_with_shutdown) = warp::serve(sse_filter)
```
goral09, line 81: In line 64 we also have an `address` member. Isn't this shadowing the previous one? Does it mean we don't need one in line 64?

goral09 (commented):
Code looks correct to me but I don't feel competent enough to review the networking bits. I will trust @marc-casperlabs to review this part more closely.

node/src/components/network/event.rs
Fraser999: I'm assuming it's down to a libp2p change.  I didn't really want to track it down since we're not focused on libp2p at the moment and I think the original value of `89` was probably chosen based on what the `Event` size was at the time rather than an actual calculated requirement of event size.

node/src/components/event_stream_server.rs
Fraser999: Done in [e635dcc](https://github.com/CasperLabs/casper-node/pull/1140/commits/e635dcc00edd2194dc21bf7c51bff70af392ade8).

dwerner (approved):
lgtm

Fraser999:
bors merge

PR #1141, opened 2021-03-15
NDRS-1013 validate block size in BlockValidator

dwerner:
This PR adds validation for the size of a block (as defined by the sum of `ToBytes::serialized_length()` of it's deploys) to the `BlockValidator`. Also included is a stop-gap fix for the `DeployAcceptor` to ensure no deploy can be larger than the allowed `max_block_size` in the chainspec.

Ref: https://casperlabs.atlassian.net/browse/NDRS-1013


afck (approved):
Looks good to me! :+1:

PR #1143, opened 2021-03-16
NO-TICKET: use correct protocol version in linear chain method

Fraser999:
This PR fixes a long standing TODO which was awaiting the protocol version being added to the block header, but was overlooked when that was made available.


afck:
I think that's already fixed in https://github.com/CasperLabs/casper-node/pull/1119, which is about to be merged soon hopefully?

Fraser999:
> I think that's already fixed in #1119, which is about to be merged soon hopefully?

Oh - OK.  Thanks.  I'll close this PR in favour of #1119

PR #1144, opened 2021-03-16
NDRS-1028: Handle known linear chain blocks in validator reactor.

goral09:


node/src/components/linear_chain.rs
```diff
@@ -47,12 +48,17 @@ pub enum Event<I> {
...
         /// The block.
         block: Box<Block>,
         /// The deploys' execution results.
         execution_results: HashMap<DeployHash, ExecutionResult>,
     },
+    /// Linear chain block we already know but we may refinalize it when syncing protocol state.
+    KnownLinearChainBlock {
+        /// The block.
+        block: Box<Block>,
+    },
```
afck, line 61: Maybe a tuple struct variant would be more concise?
```suggestion
    KnownLinearChainBlock(Box<Block>),
```

node/src/components/linear_chain.rs
```diff
@@ -407,9 +427,20 @@ where
                     era_id,
                     ..
                 } = *fs;
-                if let Err(err) = fs.verify() {
-                    warn!(%block_hash, %public_key, %err, "received invalid finality signature");
-                    return Effects::new();
+                if let Some(last_block) = self.latest_block.as_ref() {
```
afck, line 430:
```suggestion
                if let Some(latest_block) = self.latest_block.as_ref() {
```
(Or `self.last_block`?)

node/src/components/linear_chain.rs
```diff
@@ -407,9 +427,20 @@ where
...
+                    let last_block_era = last_block.header().era_id();
+                    let lowest_acceptable_era_id =
+                        last_block_era + self.auction_delay - self.unbonding_delay;
```
afck, line 433: Actually, we should make a distinction here: If it's a switch block it has already forgotten its own era's validators, unbonded some old validators, and determined new ones. In that case, we should add 1 to `last_block_era`.
```suggestion
                    let current_era = last_block.header().era_id() +
                        if last_block_header.is_switch_block() { 1 } else { 0 };
                    let lowest_acceptable_era_id =
                        current_era + self.auction_delay - self.unbonding_delay;
```

afck (commented):
Looks good to me; not sure whether this will conflict with the new `BlockAlreadyExecuted` announcement in `dev`, if we want to port it there?

goral09:
> Looks good to me; not sure whether this will conflict with the new `BlockAlreadyExecuted` announcement in `dev`, if we want to port it there?

I don't know. Lets cross that bridge when we get there.

node/src/components/linear_chain.rs
goral09: Thanks. Fixed in [`31bccd8` (#1144)](https://github.com/CasperLabs/casper-node/pull/1144/commits/31bccd8eacaf40767367af900ea6e25019463f0f).

node/src/components/linear_chain.rs
goral09: Renamed in [`31bccd8` (#1144)](https://github.com/CasperLabs/casper-node/pull/1144/commits/31bccd8eacaf40767367af900ea6e25019463f0f)

node/src/components/linear_chain.rs
goral09: Done in [`31bccd8` (#1144)](https://github.com/CasperLabs/casper-node/pull/1144/commits/31bccd8eacaf40767367af900ea6e25019463f0f)

afck (approved):

node/src/components/linear_chain.rs
```diff
@@ -407,9 +424,28 @@ where
...
+                    // unbonded some old validators, and determined new ones. In that case, we
+                    // should add 1 to last_block_era.
+                    let current_era = latest_block.header().era_id()
+                        + if latest_block.header().is_switch_block() {
+                            1
+                        } else {
+                            0
+                        };
+                    let lowest_acceptable_era_id =
+                        current_era + self.auction_delay - self.unbonding_delay;
```
afck, line 438: `subtract with overflow`
Makes sense. :smile: 
```suggestion
                        (current_era + self.auction_delay).saturating_sub(self.unbonding_delay);
```

goral09: Thanks :) 

I ran `cargo clippy` locally but it didn't return those warnings.

afck: They're not enabled yet: The overflow actually _happened_, in the tests.

node/src/components/linear_chain.rs
```diff
@@ -58,7 +52,7 @@ pub enum Event<I> {
     KnownLinearChainBlock(Box<Block>),
     /// Finality signature received.
     /// Not necessarily _new_ finality signature.
-    FinalitySignatureReceived(Box<FinalitySignature>),
+    FinalitySignatureReceived(Box<FinalitySignature>, bool),
```
afck, line 55: Now that I think of it, I'd actually make this two events: `FinalitySignatureReceived` and `FinalitySignatureCreated`. But we can do that later, in another PR.

afck: …or `FinalitySignatureReceived(Signature)` with the enum you added below?

node/src/components/linear_chain.rs
```diff
@@ -174,12 +168,37 @@ impl SignatureCache {
...
+    }
+}
+
 #[derive(DataSize, Debug)]
 pub(crate) struct LinearChain<I> {
     /// The most recently added block.
     latest_block: Option<Block>,
     /// Finality signatures to be inserted in a block once it is available.
-    pending_finality_signatures: HashMap<PublicKey, HashMap<BlockHash, FinalitySignature>>,
+    pending_finality_signatures: HashMap<PublicKey, HashMap<BlockHash, Signature>>,
```
afck, line 201: Can our own ever end up there? We only sign a block if we already have that block… or can it happen that it's somehow not in storage yet when we handle our own signature?

node/src/components/linear_chain.rs
```diff
@@ -412,12 +454,12 @@ where
                 effects.extend(
                     effect_builder
                         .handle_linear_chain_block(*block.clone())
-                        .map_some(move |fs| Event::FinalitySignatureReceived(Box::new(fs))),
+                        .map_some(move |fs| Event::FinalitySignatureReceived(Box::new(fs), true)),
```
afck, line 457: Isn't this where it should be `false`? This is a signature we just created, isn't it, so it's _not_ external/gossiped?

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -156,6 +156,7 @@ where
     vertices_to_be_added: PendingVertices<I, C>,
     /// The duration for which incoming vertices with missing dependencies are kept in a queue.
     pending_vertex_timeout: TimeDiff,
+    pub(crate) current_era: bool,
```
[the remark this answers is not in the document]
goral09: It's only temporary. It will never go into `dev` or `release` branches.

node/src/components/linear_chain.rs
goral09: ups.

dwerner: Maybe make the tuple variant a struct variant? Idk what the bool means at a glance.

node/src/components/linear_chain.rs
goral09: I tried but this is actually a lot of work.

goral09:
Closing as this change has been merged as part of a separate PR.

PR #1145, opened 2021-03-16
NDRS-1029/Restore block proposer state correctly

marc-casperlabs:
There is a leftover type mismatch from storing and loading state that was not caught due to missing type annotations. Unfortunately, the serialization format, in the configuration we use, was perfectly happy to restore a `T` as an `Option<T>` in some cases, masking these bug until actual data was filled in.

This patch adds type annotations and fixes the restoration.


Fraser999 (approved):

goral09 (approved):

Fraser999:
bors merge

PR #1151, opened 2021-03-16
Backport NDRS-1013 validate block size to 0.9.5 branch

dwerner:
backport of https://github.com/CasperLabs/casper-node/pull/1141 to release-0.9.4 branch

### DIFF
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -30,6 +30,7 @@ use crate::{
     types::{BlockLike, Chainspec, Deploy, DeployHash, Timestamp},
     NodeRng,
 };
+use casper_types::bytesrepr::ToBytes;
 use keyed_counter::KeyedCounter;
 
 use super::fetcher::FetchResult;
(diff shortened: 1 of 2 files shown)


afck (approved):
I glanced over it: Approving, assuming it's identical to #1141 — or were there any merge conflicts that changed anything?

dwerner:
> I glanced over it: Approving, assuming it's identical to #1141 — or were there any merge conflicts that changed anything?

No, just cherrypicked both commits over and it seems good.

node/src/components/block_validator.rs
```diff
@@ -59,6 +63,9 @@ pub enum Event<T, I> {
 /// Tracks whether or not there are deploys still missing and who is interested in the final result.
 #[derive(DataSize, Debug)]
 pub(crate) struct BlockValidationState<T, I> {
+    /// Total running size of the deploys in the block, which is used to validate against the
+    /// chainspec's deploy.max_block_size.
+    deploy_size_running_total: usize,
```
goral09, line 68: Just for my own understanding, can't `usize` end up being a different "type" on different machines/architectures?

node/src/components/block_validator.rs
```diff
@@ -203,26 +211,54 @@ where
...
+                    }
                 }
 
                 // Now we remove all states that have finished and notify the requestors.
                 self.validation_states.retain(|key, state| {
+                    if invalid.contains(key) {
+                        state.responders.drain(..).for_each(|responder| {
+                            effects.extend(responder.respond((false, key.clone())).ignore());
+                        });
+                        return false;
```
goral09, line 252: What's with this `return` here? Does it early exit only the current iteration of `validation_state.retain`? Usually, return is exiting from the bigger context of a method call but it's not what we want to do here or we will miss other `invalid` blocks. Instead of using `return`, can we use `if { … } else { … }` for clarity? The current version really makes me uneasy.

goral09 (approved):

node/src/components/block_validator.rs
goral09: Just for my own sanity, I've confirmed empirically that it works correctly here :+1:

PR #1152, opened 2021-03-16
NDRS-1028: Optimize handling of finality signatures in the linear chain component.

goral09:
This PR makes a couple of changes that optimize how finality signatures are being handled by the node:
* we gossip finality signatures created only by our node
* we validate the cryptographic signature at the end (as it's the most expensive test)
* validator reactor will inform `LinearChain` when a already-exected block is finalized, so that `LinearChain` can update its view about the tip of the chain and handle `IsBonded` queries properly.

### DIFF
--- a/node/src/components/linear_chain.rs
+++ b/node/src/components/linear_chain.rs
@@ -477,7 +477,7 @@ where
                             0
                         };
                     let lowest_acceptable_era_id =
-                        current_era + self.auction_delay - self.unbonding_delay;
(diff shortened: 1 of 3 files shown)


goral09:
@afck you have already reviewed all but last commit.

afck (approved):
No blockers from my side. (But Clippy is always right, of course.)

goral09:
bors merge

goral09:
bors cancel

PR #1155, opened 2021-03-16
EE-{1197,1198}: add initial trie delete

henrytill:
This PR introduces a `delete` operation to the trie, along with tests validating its behavior.  Future work will surface this through the `TrackingCopy` up to the Contract API.

### DIFF
--- a/execution_engine/src/storage/trie_store/operations/mod.rs
+++ b/execution_engine/src/storage/trie_store/operations/mod.rs
@@ -18,7 +18,7 @@ use crate::{
         transaction_source::{Readable, Writable},
         trie::{
             merkle_proof::{TrieMerkleProof, TrieMerkleProofStep},
-            Parents, Pointer, Trie, RADIX, USIZE_EXCEEDS_U8,
+            Parents, Pointer, PointerBlock, Trie, RADIX, USIZE_EXCEEDS_U8,
         },
         trie_store::TrieStore,
     },
@@ -231,6 +231,7 @@ where
 /// Given a root hash, find any try keys that are descendant from it that are:
 /// 1. referenced but not present in the database
 /// 2. referenced and present but whose values' hashes do not equal their keys (ie, corrupted)
+// TODO: We only need to check one trie key at a time
 pub fn missing_trie_keys<K, V, T, S, E>(
     _correlation_id: CorrelationId,
     txn: &T,
@@ -292,6 +293,90 @@ where
         }
     }
     Ok(missing_descendants)
+}
+
+#[cfg(test)]
+pub fn check_integrity<K, V, T, S, E>(
+    _correlation_id: CorrelationId,
+    txn: &T,
+    store: &S,
+    trie_keys_to_visit: Vec<Blake2bHash>,
+) -> Result<(), E>
+where
+    K: ToBytes + FromBytes + Eq + std::fmt::Debug,
+    V: ToBytes + FromBytes + std::fmt::Debug,
+    T: Readable<Handle = S::Handle>,
+    S: TrieStore<K, V>,
+    S::Error: From<T::Error>,
+    E: From<S::Error> + From<bytesrepr::Error>,
+{
+    for state_root in &trie_keys_to_visit {
+        match store.get(txn, &state_root)? {
+            Some(Trie::Node { .. }) => {}
+            _ => panic!("Should have a pointer block node as state root"),
+        }
+    }
+    let mut trie_keys_to_visit: Vec<(Vec<u8>, Blake2bHash)> = trie_keys_to_visit
+        .into_iter()
+        .map(|blake2b_hash| (Vec::new(), blake2b_hash))
+        .collect();
+    let mut visited = HashSet::new();
+    while let Some((mut path, trie_key)) = trie_keys_to_visit.pop() {
+        if !visited.insert(trie_key) {
+            continue;
+        }
+        let maybe_retrieved_trie: Option<Trie<K, V>> = store.get(txn, &trie_key)?;
+        if let Some(trie_value) = &maybe_retrieved_trie {
+            let hash_of_trie_value = {
+                let node_bytes = trie_value.to_bytes()?;
+                Blake2bHash::new(&node_bytes)
+            };
+            if trie_key != hash_of_trie_value {
+                panic!(
+                    "Trie key {:?} has corrupted value {:?} (hash of value is {:?}); \
+                     adding to list of missing nodes",
+                    trie_key, trie_value, hash_of_trie_value
+                );
+            }
+        }
+        match maybe_retrieved_trie {
+            // If we can't find the trie_key; it is missing and we'll return it
+            None => {
+                panic!("Missing trie key: {:?}", trie_key)
+            }
+            // If we could retrieve the node and it is a leaf, the search can move on
+            Some(Trie::Leaf { key, .. }) => {
+                let key_bytes = key.to_bytes()?;
+                if !key_bytes.starts_with(&path) {
+                    panic!(
+                        "Trie key {:?} belongs to a leaf with a corrupted affix. Key bytes: {:?}, Path: {:?}.",
+                        trie_key, key_bytes, path
+                    );
+                }
+            }
+            // If we hit a pointer block, queue up all of the nodes it points to
+            Some(Trie::Node { pointer_block }) => {
+                for (byte, pointer) in pointer_block.to_indexed_pointers() {
+                    let mut new_path = path.clone();
+                    new_path.push(byte);
+                    match pointer {
+                        Pointer::LeafPointer(descendant_leaf_trie_key) => {
+                            trie_keys_to_visit.push((new_path, descendant_leaf_trie_key))
+                        }
+                        Pointer::NodePointer(descendant_node_trie_key) => {
+                            trie_keys_to_visit.push((new_path, descendant_node_trie_key))
+                        }
+                    }
+                }
+            }
+            // If we hit an extension block, add its pointer to the queue
+            Some(Trie::Extension { pointer, affix }) => {
+                path.extend_from_slice(affix.as_slice());
+                trie_keys_to_visit.push((path, pointer.into_hash()))
+            }
+        }
+    }
+    Ok(())
 }
 
 struct TrieScan<K, V> {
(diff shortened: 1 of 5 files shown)


execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+                        let trie_key = Blake2bHash::new(&trie_node.to_bytes()?);
+                        new_elements.push((trie_key, trie_node));
+                        break;
+                    }
+                    // The sibling is a leaf and the grandparent is a node.  Reseat the single leaf
+                    // sibling into the grandparent.
+                    (Pointer::LeafPointer(..), Some((idx, Trie::Node { mut pointer_block }))) => {
+                        pointer_block[idx as usize] = Some(sibling_pointer);
+                        let trie_node = Trie::Node { pointer_block };
+                        let trie_key = Blake2bHash::new(&trie_node.to_bytes()?);
```
afck, line 575: This is repeated a lot. Maybe we should add a `trie_node.compute_hash()` method?

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+                        // Elsewhere we maintain the invariant that all trie keys have corresponding
+                        // trie values.
+                        let sibling_trie = store
+                            .get(txn, &sibling_trie_key)?
+                            .expect("should have sibling");
+                        match sibling_trie {
+                            Trie::Leaf { .. } => {
+                                panic!("Node pointer should not point to leaf")
+                            }
+                            // The single sibling is a node, and the grandparent does not exist.
```
afck, line 610: It doesn't? Didn't we just push it back onto `parents`?

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+                                let trie_key = Blake2bHash::new(&new_extension.to_bytes()?);
+                                new_elements.push((trie_key, new_extension))
+                            }
+                            // The single sibling is a node, and there exists a grandparent.
+                            // Therefore the parent is not the root.  We output an extension to
+                            // replace the parent, with a single byte corresponding to the sibling
+                            // index.  In the next loop iteration, we will handle the case where
+                            // this extension might need to be combined with a grandparent
+                            // extension.
+                            Trie::Node { .. } => {
```
afck, line 630: Isn't this identical to the previous match arm? (Was the distinction about the grandparents removed at some point?)

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+                    trie_key,
+                    Trie::Extension {
+                        affix: child_affix,
+                        pointer,
+                    },
+                )),
+                Trie::Extension { affix, .. },
+            ) => {
+                let mut new_affix: Vec<u8> = Vec::<u8>::from(affix);
+                new_affix.extend(Vec::<u8>::from(child_affix.to_owned()));
```
afck, line 685:
```suggestion
                new_affix.extend(child_affix.to_owned());
```
(Not sure if even the `to_owned()` is required?)

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+                Some((
+                    trie_key,
+                    Trie::Extension {
+                        affix: child_affix,
+                        pointer,
+                    },
+                )),
+                Trie::Extension { affix, .. },
+            ) => {
+                let mut new_affix: Vec<u8> = Vec::<u8>::from(affix);
```
afck, line 684: No need to specify the type `Vec<u8>` twice.
```suggestion
                let mut new_affix: Vec<u8> = affix.into();
```
or
```suggestion
                let mut new_affix = Vec::<u8>::from(affix);
```

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+                        pointer: pointer.to_owned(),
+                    };
+                    Blake2bHash::new(&new_extension.to_bytes()?)
+                }
+            }
+            // The parent is an extension and the new element is a pointer block.
+            // The next element we add will be an extension to the pointer block we are going to
+            // add.
+            (Some((trie_key, Trie::Node { .. })), Trie::Extension { affix, mut pointer }) => {
+                pointer = Pointer::NodePointer(*trie_key);
```
afck, line 699: It would be less confusing to just not bind that pointer, since we're not actually mutating the parent.
```suggestion
            (Some((trie_key, Trie::Node { .. })), Trie::Extension { affix, .. }) => {
                let pointer = Pointer::NodePointer(*trie_key);
```

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+                let trie_extension = Trie::Extension { affix, pointer };
+                let trie_key = Blake2bHash::new(&trie_extension.to_bytes()?);
+                new_elements.push((trie_key, trie_extension))
+            }
+        }
+    }
+    let mut root_hash: Blake2bHash = root.to_owned();
```
afck, line 706: `new_elements` is never empty, is it? It would be clearer to not make this mutable then:
```suggestion
    let root_hash = new_elements.last().expect("new_elements can't be empty").0;
```

afck (commented):
Haven't reviewed the test code yet, but looks good so far.
Anyway, someone more familiar with the trie store should approve this.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -294,6 +295,90 @@ where
...
+    V: ToBytes + FromBytes + std::fmt::Debug,
+    T: Readable<Handle = S::Handle>,
+    S: TrieStore<K, V>,
+    S::Error: From<T::Error>,
+    E: From<S::Error> + From<bytesrepr::Error>,
+{
+    for state_root in &trie_keys_to_visit {
+        match store.get(txn, &state_root)? {
+            Some(Trie::Node { .. }) => {}
+            _ => panic!("Should have a pointer block node as state root"),
```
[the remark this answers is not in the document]
henrytill: The state root `Trie` value is special in the sense that it is always a `Trie::Node`, so, unless I'm misundertanding something, we need to do that check here.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -294,6 +295,90 @@ where
...
+        let maybe_retrieved_trie: Option<Trie<K, V>> = store.get(txn, &trie_key)?;
+        if let Some(trie_value) = &maybe_retrieved_trie {
+            let hash_of_trie_value = {
+                let node_bytes = trie_value.to_bytes()?;
+                Blake2bHash::new(&node_bytes)
+            };
+            if trie_key != hash_of_trie_value {
+                panic!(
+                    "Trie key {:?} has corrupted value {:?} (hash of value is {:?}); \
+                     adding to list of missing nodes",
```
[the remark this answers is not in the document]
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1155/commits/f9300c1d4cfca2adeddc897962242db28bbabd78

execution_engine/src/storage/trie_store/operations/mod.rs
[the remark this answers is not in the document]
afck: I mean we call `store.get` twice for all initial entries in `trie_keys_to_visit`.
And aren't the initial entries exactly the ones with an empty path? (The roots?)

If we did this check below, inside the `while` loop, we wouldn't need this `get` call here: Below, we'd check that if `path.is_empty()` then it's a `Trie::Node`, otherwise panic.

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: @henrytill and I agree we should do this; we will want to update all of the trie operations though.  Making a follow up ticket...

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: Yes, this comment predated that push, we will fix it.

xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1155/commits/e0718bb1afd75d046aa458c8647f39714531e3b1

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: Yes, they are identical.   Fixed in https://github.com/CasperLabs/casper-node/pull/1155/commits/e0718bb1afd75d046aa458c8647f39714531e3b1

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1155/commits/435884b23d9a6625539af2ff209c8fa84cd65567

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1155/commits/af8b66e01a5f65fe55260b1f7ed80742615528a8

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1155/commits/a2723a86125857bad79405a86dc66e52ca680154

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: Fixed https://github.com/CasperLabs/casper-node/pull/1155/commits/c2119df799123bac55303c1d81d5263fa26f9b9f

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -394,6 +479,238 @@ where
...
+    let root_trie = match store.get(txn, root)? {
+        None => return Ok(DeleteResult::RootNotFound),
+        Some(root_trie) => root_trie,
+    };
+
+    let key_bytes = key_to_delete.to_bytes()?;
+    let TrieScan { tip, mut parents } =
+        scan::<_, _, _, _, E>(correlation_id, txn, store, &key_bytes, &root_trie)?;
+
+    // Check that tip is a leaf
```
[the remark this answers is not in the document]
xcthulhu: This function will only address deleting a single `(K,V)` pair, which corresponds to a leaf.

afck (approved):

EdHastingsCasperLabs (approved):

PR #348, opened 2020-10-09
NDRS-504: replace rmp_serde with bincode

Fraser999:
We recently discovered that the rmp_serde encoder and decoder currently reports its type as human readable.  Since recent changes to our code specialise (de)serialisation for human-readable formats (generally by encoding byte strings to hex, hence doubling their size), we need to change to an encoding format which is not classed as human-readable.

Note, this PR is based off https://github.com/CasperLabs/casper-node/pull/347, so only the final commit should be reviewed here.

PR #1156, opened 2021-03-17
NO-TICKET: use bincode in small_network

Fraser999:
This PR swaps `rmp_serde` for `bincode` inside the small_network component.

This is in line with a similar replacement made in #348, and should have been changed at that stage for the same reasons.  However it was overlooked due to being a transitive dependency.

There is also a change to `ConsensusMessage` whereby it now uses [serde_bytes](https://github.com/serde-rs/bytes) to (de)serialize the `Vec<u8>` member variable, which should be significantly more efficient.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4620,27 +4620,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "rmp"
-version = "0.8.9"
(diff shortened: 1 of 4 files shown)


Fraser999:
Not required

PR #1158, opened 2021-03-17
NCTL :: shellcheck linter changes

ghost:
Ahead of mainnet I have ran the shellcheck linter over nctl bash code.  This PR encapsulates the resultant changes.


linclelinkpart5 (approved):
Looks good, turns out quoting in Bash is confusing!

PR #1165, opened 2021-03-18
NDRS-1023/add greeting before handshake

marc-casperlabs:
This adds a greeting to the networking protocol, which is 12 bytes in length and contains the SemVer protocol version.

While `V1_0_0` does nothing with this information, getting this merged into the final release ensures we can change the underlying networking protocol in an easier manner on upgrades.

**This change is not backwards compatible and will break in any existing network**

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -56,7 +56,10 @@ use std::{
 };
 
 use anyhow::Context;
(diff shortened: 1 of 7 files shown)


marc-casperlabs:
We are not going to merge this into mainnet and thus I am closing this PR. This will live on the feature branch until subsumed by a different PR.

PR #1172, opened 2021-03-19
HWY-271: Log synchronizer queue size.

goral09:
https://casperlabs.atlassian.net/browse/HWY-271

Adds a new timer `TIMER_ID_SYNCHRONIZER_QUEUE` that is recreated every 5 seconds for every _unfinalized_ era. Upon handling the timer, lengths of synchronizer queues will be logged at the `debug` level. This is to aid in debugging the protocol state synchronization problems. Example:
```
{"timestamp":"Mar 19 00:00:18.685","level":"DEBUG","fields":{"message":"synchronizer queue lengths","era_id":"2c9fb42bf82a564db5ffde1f13dcce7c36ff9f66bf70b633f06ac083b2c022fa","vertices_to_be_added_later":0,"vertex_deps":0,"vertices_to_be_added":0},"target":"casper_node::components::consensus::protocols::highway::synchronizer"}
```

### DIFF
--- a/node/src/components/consensus/protocols/highway.rs
+++ b/node/src/components/consensus/protocols/highway.rs
@@ -51,6 +51,8 @@ const TIMER_ID_VERTEX_WITH_FUTURE_TIMESTAMP: TimerId = TimerId(1);
 const TIMER_ID_PURGE_VERTICES: TimerId = TimerId(2);
 /// The timer for logging inactive validators.
 const TIMER_ID_LOG_PARTICIPATION: TimerId = TimerId(3);
+/// The timer for logging synchronizer queue size.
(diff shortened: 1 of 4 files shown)


afck (approved):
Still looks good to me, although [as discussed](https://github.com/CasperLabs/casper-node/pull/1159#discussion_r597154194) it would be less wasteful to only request the state in the most recent era.

goral09:
> Still looks good to me, although [as discussed](https://github.com/CasperLabs/casper-node/pull/1159#discussion_r597154194) it would be less wasteful to only request the state in the most recent era.

I agree, this PR does not include this change though. I will create a separate one only with that.

goral09:
bors merge

PR #1173, opened 2021-03-19
HWY-272: Log oldest seen panorama when syncing.

goral09:
https://casperlabs.atlassian.net/browse/HWY-272

Adds logging of the oldest (lowest sequence number) unit seen per validator while syncing. Shows the progress of synchronization per validator.

### DIFF
--- a/node/src/components/consensus/protocols/highway/synchronizer.rs
+++ b/node/src/components/consensus/protocols/highway/synchronizer.rs
@@ -162,6 +162,8 @@ where
     pending_vertex_timeout: TimeDiff,
     /// Instance ID of an era for which this synchronizer is constructed for.
     era_id: C::InstanceId,
+    /// Keeps track of the lowest/oldest seen unit per validator when syncing.
+    /// Used only for logging.
     oldest_seen_panorama: ValidatorMap<Option<u64>>,
     /// Boolean flag indicating whether we're synchronizing current era.
     pub(crate) current_era: bool,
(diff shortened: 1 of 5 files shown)


fizyk20 (approved):

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -211,6 +222,7 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
             vertices_to_be_added = self.vertices_to_be_added_len(),
             "synchronizer queue lengths"
         );
+        debug!(oldest_panorama=%self.oldest_seen_panorama, "oldest seen unit per validator");
```
afck, line 225: We could probably stop logging this once all entries have sequence number 0.

afck (approved):

node/src/components/consensus/protocols/highway/synchronizer.rs
goral09: I was also looking into how it would update if we started purging the vertices from the synchronizer but I am not sure if it's worth it – we would have to iterate through all entries in the `self.vertices_to_be_added` collection.

goral09: Done in [`0a9a01f` (#1173)](https://github.com/CasperLabs/casper-node/pull/1173/commits/0a9a01fb6570e0dbd21236720e8faa2c30a529c8).

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -159,18 +162,26 @@ where
     pending_vertex_timeout: TimeDiff,
     /// Instance ID of an era for which this synchronizer is constructed for.
     era_id: C::InstanceId,
+    oldest_seen_panorama: ValidatorMap<Option<u64>>,
```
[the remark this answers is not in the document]
goral09: Added in [`0a9a01f` (#1173)](https://github.com/CasperLabs/casper-node/pull/1173/commits/0a9a01fb6570e0dbd21236720e8faa2c30a529c8)

goral09:
bors merge

afck (approved):

PR #1175, opened 2021-03-19
HWY-273: Add cache for prevalidated vertices.

goral09:
https://casperlabs.atlassian.net/browse/HWY-273

Adds a cache for pending prevalidated vertices (vertices that are currently being synchronized) and for every new vertex received, checks if it's already in that cache. If not, don't request it again.

Also, checks the node's protocol state if the incoming vertex is already known before trying to validate it.

### DIFF
--- a/node/src/components/consensus/protocols/highway.rs
+++ b/node/src/components/consensus/protocols/highway.rs
@@ -429,20 +429,15 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
     fn pre_validate_vertex(
         &mut self,
         v: Vertex<C>,
-    ) -> Result<PreValidationStatus<C>, (Vertex<C>, VertexError)> {
+    ) -> Result<PreValidatedVertex<C>, (Vertex<C>, VertexError)> {
         let id = v.id();
         if let Some(prev_pvv) = self.pvv_cache.get(&id) {
-            return Ok(PreValidationStatus::Known(prev_pvv.clone()));
+            return Ok(prev_pvv.clone());
         }
         let pvv = self.highway.pre_validate_vertex(v)?;
         self.pvv_cache.insert(id, pvv.clone());
-        Ok(PreValidationStatus::New(pvv))
+        Ok(pvv)
     }
-}
-
-enum PreValidationStatus<C: Context> {
-    New(PreValidatedVertex<C>),
-    Known(PreValidatedVertex<C>),
 }
 
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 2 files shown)


afck (approved):

goral09:
@afck can I ask you for another review? I have added a piece of logic that I think we had missed originally.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -415,6 +423,26 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
             .last_finalized()
             .map_or(false, is_switch)
     }
+
+    /// Prevalidates the vertex but checks the cache for previously validated vertices.
+    /// Avoids multiplate validation of the same vertex.
```
afck, line 428:
```suggestion
    /// Avoids multiple validations of the same vertex.
```
(Or "repeated validation", or "redundant validation".)

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -395,6 +411,27 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
...
+    /// Adds `sender` as alternative source for downloading vertex.
+    /// Returns protocol outcomes scheduled as a result of this action.
+    pub(crate) fn alternative_source(
+        &mut self,
+        pvv: PreValidatedVertex<C>,
+        sender: I,
+        now: Timestamp,
+    ) -> ProtocolOutcomes<I, C> {
+        match self
+            .vertices_to_be_added
```
afck, line 423: Aren't there several other collections, like `vertex_deps`? We can't expect the vertex to be in `vertices_to_be_added`, in general.

afck (changes requested):
I don't understand the added commit: Isn't all of that logic (adding alternative source, etc.) already taken care of by the existing synchronizer implementation?

goral09:
> I don't understand the added commit: Isn't all of that logic (adding alternative source, etc.) already taken care of by the existing synchronizer implementation?

You're right. I thought I can save some iterations and checks with this but it looks like it's not that bug of a difference in the end. I will revert that commit.

node/src/components/consensus/protocols/highway.rs
goral09: Reverted in [`81c8391` (#1175)](https://github.com/CasperLabs/casper-node/pull/1175/commits/81c8391fc53aaf2512a9398393fefaba1277d608).

node/src/components/consensus/protocols/highway/synchronizer.rs
goral09: Reverted in [`81c8391` (#1175)](https://github.com/CasperLabs/casper-node/pull/1175/commits/81c8391fc53aaf2512a9398393fefaba1277d608).

node/src/components/consensus/protocols/highway.rs
afck: No, this one I had overlooked in my first review.

afck (approved):

node/src/components/consensus/protocols/highway.rs
goral09: Fixed [`c23aa66` (#1175)](https://github.com/CasperLabs/casper-node/pull/1175/commits/c23aa668be091222c846512f13de3ec2db2ddcf2)

goral09:
Just FYI @afck , I will close this PR and opened a new one against `dev`. As discussed earlier today about what branch to target and what will be released next.

PR #1190, opened 2021-03-20
NDRS-1037/monitor potential bans

marc-casperlabs:
This change makes it possible to monitor potential blocklisted peers (which are not blocked during the current grace period) and determine their IP address as well from the logs. This is done through two changes:

1. The existing `warn!` log message that logged when a peer would have been blocked has been updated to include the peer's node id (sometimes called peer ID).
2. The existing log for established connections has been upgrade from a `debug!` log level to an `info!` log level.

The code changes are minimal, containing only a slight alteration to the logging calls. The most dramatic effect is a great increase in log messages, as every successful connection will now result in an INFO level log message.

## Node operator recommendations

To find peers that would have been blocked (either due to misconfiguration, bugs or genuine malicious behavior), a node operator should monitor the logs for the WARN level message that says `peer deemed potentially byzantine`. Upon finding such a message, they should search the log backwards to find a INFO level message that maps a peer's node ID to an actual address: "established incoming connection".

Please note that the address listed there is not the public address of the peer, but rather the outgoing one; for the listening address the port will be different. This may be an issue in cases where a single IP hosts multiple nodes, but we believe this better than the alternative of using the node supplied address, which is would be prone to forgery by the sender. In most cases, the IP address alone should be sufficient to find the node in question.


xcthulhu (approved):

node/src/reactor/validator.rs
```diff
@@ -998,9 +998,9 @@ impl reactor::Reactor for Reactor {
                             });
                         self.dispatch_event(effect_builder, rng, reactor_event)
                     }
-                    ConsensusAnnouncement::DisconnectFromPeer(_peer) => {
+                    ConsensusAnnouncement::DisconnectFromPeer(peer) => {
                         // TODO: handle the announcement and actually disconnect
-                        warn!("Disconnecting from a given peer not yet implemented.");
+                        warn!(%peer, "peer deemed potentially byzantine");
```
goral09, line 1003: byzantine is a bad choice of word – consensus will disconnect from a peer also for other types of offences (invalid deploys for example).

goral09 (approved):

xcthulhu:
bors merge

goral09:
> bors merge

Don't do that for other people's PRs. If you can't wait for this to get merged, ask the author first.

node/src/reactor/validator.rs
marc-casperlabs: Alright, has been changed.

PR #1191, opened 2021-03-21
NO-TICKET: Add more logging to track block proposals.

afck:
This adds INFO-level logs for receiving and validating proposed blocks, and whenever we send a witness unit in a round where we didn't receive a block.

### DIFF
--- a/node/src/components/consensus/protocols/highway.rs
+++ b/node/src/components/consensus/protocols/highway.rs
@@ -333,6 +333,7 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
         let vertex = vv.inner();
         if let (Some(value), Some(timestamp)) = (vertex.value(), vertex.timestamp()) {
             if value.needs_validation() {
+                self.log_proposal(vertex, "requesting proposal validation");
                 let cv = value.clone();
                 self.pending_values
                     .entry(value.hash())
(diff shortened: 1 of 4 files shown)


node/src/components/consensus/protocols/highway.rs
```diff
@@ -333,6 +333,7 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
         let vertex = vv.inner();
         if let (Some(value), Some(timestamp)) = (vertex.value(), vertex.timestamp()) {
             if value.needs_validation() {
+                self.log_proposal(vertex, "requesting proposal validation");
```
goral09, line 336: With vertices having 100 deploys and 2500 transfers this can be too much. Can we log just the hash instead? We can always look it up later on Clarity or via node's RPC.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -381,6 +384,7 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
             error!(vertex = ?vv.inner(), "unexpected vertex in evidence-only mode");
             return vec![];
         }
+        self.log_proposal(vv.inner(), "adding valid proposal to the protocol state");
```
goral09, line 387: Same here – just a hash would be better.

goral09 (approved):

afck:
(Yes, we're not logging all the deploy hashes.)

PR #1194, opened 2021-03-21
NDRS-1034

linclelinkpart5:
Add nctl functions to verify transfers and deploys


utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -125,6 +125,68 @@ function do_await_n_blocks() {
...
+}
+
+function verify_deploy() {
+    local NODE_ID=${1}
+    local TRANSFER_HASH=${2}
+    local WALKBACK=${3}
+    local BLOCK_HASH=${4}
+
+    _verify_deploy_transfer "$NODE_ID" "$TARGET_HASH" "deploy" "$WALKBACK" "$PARENT"
+}
```
goral09, line 188: Can you add comments to these public methods?

linclelinkpart5: Done!

utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -125,6 +125,72 @@ function do_await_n_blocks() {
...
+    fi
+}
+
+# Walks the chain on a given node, looking for a specific transfer hash.
+# Prints out the block hash that contains the target hash, if found.
+function verify_transfer() {
+    local NODE_ID=${1}
+    local TRANSFER_HASH=${2}
+    local WALKBACK=${3}
+    local BLOCK_HASH=${4}
```
goral09, line 178: The method says that it prints out the block hash that contains the transfer we're looking for, what's _this_ `BLOCK_HASH` argument then?

Second question, can we also return the hash of the block containing the transfer? It might be useful to have downstream.

utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -125,6 +125,72 @@ function do_await_n_blocks() {
...
+        exit 1
+    fi
+}
+
+# Walks the chain on a given node, looking for a specific transfer hash.
+# Prints out the block hash that contains the target hash, if found.
+function verify_transfer() {
+    local NODE_ID=${1}
+    local TRANSFER_HASH=${2}
+    local WALKBACK=${3}
```
goral09, line 177: What's this? Please add to the method comment.

utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -125,6 +125,72 @@ function do_await_n_blocks() {
...
+        fi
+    else
+        log "Error: Block with $TYPE hash not found within walkback range!"
+        exit 1
+    fi
+}
+
+# Walks the chain on a given node, looking for a specific transfer hash.
+# Prints out the block hash that contains the target hash, if found.
+function verify_transfer() {
```
goral09, line 174: Can we have a version of this method that doesn't expect the block hash to begin the search for? It should be the current LFB of the node.

goral09, line 174: I don't think the name of the method really conveys what it's doing – it doesn't verify anything, it just looks it up. So maybe `find_transfer` will be better (and `find_deploy`).

PR #1198, opened 2021-03-22
NDRS-1031: Port back NDRS-1028 to `dev` branch.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-1031

NDRS-1028: Optimize handling of finality signatures in the linear chain component. r=goral09 a=goral09

This PR makes a couple of changes that optimize how finality signatures are being handled by the node:
* We gossip finality signatures created only by our node
* We validate the cryptographic signature at the end (as it's the most expensive test)
* validator reactor will inform `LinearChain` when an already-exected block is finalized so that `LinearChain` can update its view about the tip of the chain and handle `IsBonded` queries properly.

### DIFF
--- a/node/src/reactor/joiner.rs
+++ b/node/src/reactor/joiner.rs
@@ -509,8 +509,10 @@ impl reactor::Reactor for Reactor {
 
         let linear_chain = linear_chain::LinearChain::new(
             &registry,
-            chainspec_loader.initial_state_root_hash(),
             protocol_version.clone(),
+            chainspec_loader.initial_state_root_hash(),
+            chainspec_loader.chainspec().core_config.auction_delay,
+            chainspec_loader.chainspec().core_config.unbonding_delay,
             chainspec_loader
                 .chainspec()
                 .protocol_config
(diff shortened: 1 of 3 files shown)


afck (approved):

node/src/reactor/joiner.rs
```diff
@@ -509,8 +509,10 @@ impl reactor::Reactor for Reactor {
 
         let linear_chain = linear_chain::LinearChain::new(
             &registry,
-            chainspec_loader.initial_state_root_hash(),
             protocol_version.clone(),
+            chainspec_loader.initial_state_root_hash(),
+            chainspec_loader.chainspec().core_config.auction_delay,
+            chainspec_loader.chainspec().core_config.unbonding_delay,
```
afck, line 515: Might be safer to pass in a chainspec reference: Then it's impossible to accidentally switch auction and unbonding delay.

goral09: I did that initially but then I thought that passing in the whole thing might be problematic when we want to test it/mock it.

PR #1201, opened 2021-03-22
HWY-273: Add cache for prevalidated vertices.

goral09:
https://casperlabs.atlassian.net/browse/HWY-273

Adds a cache for pending prevalidated vertices (vertices that are currently being synchronized) and for every new vertex received, checks if it's already in that cache. If not, don't request it again.

Also, checks the node's protocol state if the incoming vertex is already known before trying to validate it.

### DIFF
--- a/node/src/components/consensus/protocols/highway.rs
+++ b/node/src/components/consensus/protocols/highway.rs
@@ -28,7 +28,10 @@ use crate::{
         highway_core::{
             active_validator::Effect as AvEffect,
             finality_detector::{FinalityDetector, FttExceeded},
-            highway::{Dependency, GetDepOutcome, Highway, Params, ValidVertex, Vertex},
+            highway::{
+                Dependency, GetDepOutcome, Highway, Params, PreValidatedVertex, ValidVertex,
+                Vertex, VertexError,
+            },
             state,
             state::{Observation, Panorama},
             validators::{ValidatorIndex, Validators},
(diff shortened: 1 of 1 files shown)


node/src/components/consensus/protocols/highway.rs
```diff
@@ -677,7 +704,8 @@ where
             }
             TIMER_ID_PURGE_VERTICES => {
                 self.synchronizer.purge_vertices(now);
-                let next_time = now + self.synchronizer.pending_vertex_timeout();
+                self.pvv_cache.clear();
+                let next_time = Timestamp::now() + self.synchronizer.pending_vertex_timeout();
```
afck, line 708: I think I removed all calls to `now()` to make the unit tests deterministic. (Not sure whether this one matters, but it would still be nice to keep the wall clock out of this module.)
```suggestion
                let next_time = now + self.synchronizer.pending_vertex_timeout();
```

afck (approved):
No blockers.

node/src/components/consensus/protocols/highway.rs
goral09: Done in [`96422dc` (#1201)](https://github.com/CasperLabs/casper-node/pull/1201/commits/96422dc1c1e4ab54cd2167d68a799365ab6928ec)

goral09:
bors  merge

PR #1207, opened 2021-03-23
NDRS-1043: change trigger for upgrade checks to timed interval

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1043.

This PR changes the trigger for the chainspec loader to check for the next installed version of the chainspec.  The current trigger is the announcement of a new block; with this PR it is on a repeating minute timer.

With this being on a regular timed interval now, an upgrade will be detected even if block production has stalled.

I don't believe we need the interval to be a config setting, but if reviewers disagree, that can be changed.

I have tested this works via the `nctl-upgrade-protocol` command.

### DIFF
--- a/node/src/components/chainspec_loader.rs
+++ b/node/src/components/chainspec_loader.rs
@@ -16,6 +16,7 @@ use std::{
     path::{Path, PathBuf},
     str::FromStr,
     sync::Arc,
+    time::Duration,
 };
 
 use datasize::DataSize;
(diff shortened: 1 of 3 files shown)


goral09:
I don't believe this is required for correctness – if blocks are not produced then there will be no upgrade anyway.

goral09 (approved):

Fraser999:
bors merge

PR #1208, opened 2021-03-23
NDRS-1051: Use highest block as an init state for components when no syncing is performed.

goral09:
All components use `LinearChainSync`'s last known block as an initial state. In the case when there's was syncing performed they failed to initialize properly. We need to return the highest block when there's no synchronization.


fizyk20 (approved):

goral09:
bors merge

goral09:
bors merge

PR #1210, opened 2021-03-24
NDRS-{1050,1051}: Port fixes from release-0.9.5 branch to dev.

goral09:


node/src/components/linear_chain_sync.rs
```diff
@@ -91,7 +92,7 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
         chainspec: &Chainspec,
         storage: &Storage,
         init_hash: Option<BlockHash>,
```
afck, line 93: (Not necessarily in this PR.)
Should we rename that to `trusted_block_hash`, to avoid confusion with the initial state root hash?

node/src/components/linear_chain_sync.rs
```diff
@@ -614,6 +633,24 @@ where
                                 ),
                             );
                         }
+                        if block.protocol_version() != self.protocol_version {
+                            warn!(
+                                %peer,
+                                protocol_version = ?self.protocol_version,
+                                block_version = ?block.protocol_version(),
```
afck, line 640: `%` would probably look nicer than `?`? Not sure what they look like for protocol versions.

afck (approved):

node/src/components/linear_chain_sync.rs
goral09: Changed in [`339b851` (#1210)](https://github.com/CasperLabs/casper-node/pull/1210/commits/339b85175e29df8f895c3bdb0bbddf666ab0325e).

afck (approved):

goral09:
bors merge

PR #1213, opened 2021-03-24
EE-1076: Recursive query limit

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1076

This PR add a query depth limit.
    
This commit will ensure that arbitrarily long chains of named keys either a contract/account that references an account/contract, or a chain of urefs that references other urefs - will stop after exceeding fixed depth limit.


goral09:
Maybe there should be a cycle-detection logic here as well?

mpapierski:
> 
> 
> Maybe there should be a cycle-detection logic here as well?

There is https://github.com/CasperLabs/casperlabs-node/blob/cf7920a03b5b7b85bc2856c08568aa48bc0a06c5/execution_engine/src/core/engine_state/query.rs#L13

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -402,6 +419,7 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
     pub fn query(
         &self,
         correlation_id: CorrelationId,
+        config: &EngineConfig,
```
goral09, line 422: Personally, I wouldn't pass in the whole `EngineConfig` – only the bits that are required by the method.

darthsiroftardis (approved):
LGTM!

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -410,9 +428,14 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
         let mut proofs = Vec::new();
 
         loop {
+            if query.depth > config.query_depth_limit {
```
Fraser999, line 431: I think this lets the query exceed the limit by 1.  For example, it the limit is 1, this will only cause us to return a result on the third iteration of the loop, i.e. after two iterations have completed.

Maybe that's the intent, but it seems a bit non-intuitive to me.  If we want to keep this behaviour, I think we should document it more thoroughly - particularly in more user-facing places like the config file.

node/src/components/contract_runtime/config.rs
```diff
@@ -21,6 +22,10 @@ pub struct Config {
     ///
     /// Defaults to 512.
     max_readers: Option<u32>,
+    /// The limit of depth of recursive global state queries.
+    ///
+    /// Defaults to 5.
+    query_depth_limit: Option<u64>,
```
Fraser999, line 28: I'd be inclined to name it `max_query_depth` for consistency with the other limiting config values.

execution_engine/src/core/tracking_copy/mod.rs
mpapierski: Fair point. I fixed the looping so it will hard stop exactly at `query.depth >= config.query_depth_limit` and also renamed the config name

Fraser999 (approved):

resources/local/config.toml
```diff
@@ -283,3 +283,8 @@ verify_accounts = true
 #
 # The size should be a multiple of the OS page size.
 #max_global_state_size = 32_212_254_720
+
+# Optional depth limit to use for global state queries.
+#
+# If unset, defaults to 5.
+#max_query_depth = 5
```
henrytill, line 290: Missing trailing newline

resources/production/config-example.toml
```diff
@@ -287,3 +287,8 @@ verify_accounts = true
 #
 # The size should be a multiple of the OS page size.
 #max_global_state_size = 805306368000
+
+# Optional depth limit to use for global state queries.
+#
+# If unset, defaults to 5.
+#max_query_depth = 5
```
henrytill, line 294: Missing trailing newline

henrytill (approved):
:+1: Go ahead and merge when ready

PR #1219, opened 2021-03-24
NDRS-1053: Add a parent to CandidateBlock.

afck:
With deploy replay protection the validity of a candidate block depends not only on its contents and its timestamp, but also on its ancestors. So if there are two different vertices with identical consensus values, i.e. candidate blocks, one of them could be valid and one invalid. Whether a node would consider them valid would depend on which one was received first, because that would determine the set of ancestors we check against.

Adding the parent hash makes candidate block ancestors explicit so that consensus can rely on the validity result again: If a candidate block is deemed valid, we know that all vertices that carry it are valid.

https://casperlabs.atlassian.net/browse/NDRS-1053

### DIFF
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -31,6 +31,7 @@ use casper_types::{PublicKey, U512};
 
 use crate::{
     components::Component,
+    crypto::hash::Digest,
     effect::{
         announcements::ConsensusAnnouncement,
         requests::{
@@ -96,6 +97,7 @@ pub enum Event<I> {
         era_id: EraId,
         proto_block: ProtoBlock,
         block_context: BlockContext,
+        parent: Option<Digest>,
     },
     #[from]
     ConsensusRequest(ConsensusRequest),
@@ -189,10 +191,11 @@ impl<I: Debug> Display for Event<I> {
                 era_id,
                 proto_block,
                 block_context,
+                parent,
             } => write!(
                 f,
-                "New proto-block for era {:?}: {:?}, {:?}",
-                era_id, proto_block, block_context
+                "New proto-block for era {:?}: {:?}, {:?}, parent: {:?}",
+                era_id, proto_block, block_context, parent
             ),
             Event::ConsensusRequest(request) => write!(
                 f,
@@ -303,7 +306,8 @@ where
                 era_id,
                 proto_block,
                 block_context,
-            } => handling_es.handle_new_proto_block(era_id, proto_block, block_context),
+                parent,
+            } => handling_es.handle_new_proto_block(era_id, proto_block, block_context, parent),
             Event::ConsensusRequest(ConsensusRequest::HandleLinearBlock(block, responder)) => {
                 handling_es.handle_linear_chain_block(*block, responder)
             }
--- a/node/src/components/consensus.rs
+++ b/node/src/components/consensus.rs
@@ -106,7 +106,6 @@ pub enum Event<I> {
         era_id: EraId,
         sender: I,
         proto_block: ProtoBlock,
-        timestamp: Timestamp,
         parent: Option<Digest>,
         valid: bool,
     },
@@ -207,14 +206,12 @@ impl<I: Debug> Display for Event<I> {
                 era_id,
                 sender,
                 proto_block,
-                timestamp,
                 parent,
                 valid,
             } => write!(
                 f,
-                "Candidate block received from {:?} at {} for {} with parent {:?} is {}: {:?}",
+                "Candidate block received from {:?} for {} with parent {:?} is {}: {:?}",
                 sender,
-                timestamp,
                 era_id,
                 parent,
                 if *valid { "valid" } else { "invalid" },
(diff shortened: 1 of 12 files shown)


node/src/components/consensus/candidate_block.rs
```diff
@@ -53,6 +57,11 @@ impl CandidateBlock {
     pub(crate) fn accusations(&self) -> &Vec<PublicKey> {
         &self.accusations
     }
+
+    /// Returns the parent candidate block, or `Non` if this is the first in this era.
```
goral09, line 61:
```suggestion
    /// Returns the parent candidate block, or `None` if this is the first in this era.
```

node/src/components/consensus/consensus_protocol.rs
```diff
@@ -93,6 +93,7 @@ pub(crate) enum ProtocolOutcome<I, C: Context> {
     CreateNewBlock {
         block_context: BlockContext,
         past_values: Vec<C::ConsensusValue>,
+        parent_value: Option<C::ConsensusValue>,
```
goral09, line 96: Why not make it part of the `BlockContext`?

afck: I'm trying, but this turns out to be a major change: We'd have to make `BlockContext` (which is created inside `highway_core`) generic in `C: Context`, which causes other issues…
Since I'd like to rethink that solution before porting it to `dev` anyway, maybe it's not worth doing this in this PR?

node/src/components/consensus.rs
```diff
@@ -105,6 +107,7 @@ pub enum Event<I> {
         sender: I,
         proto_block: ProtoBlock,
         timestamp: Timestamp,
```
goral09, line 109: isn't the timestamp now part of the `ProtoBlock`?

afck: Thanks! Removed.

node/src/components/consensus.rs
```diff
@@ -203,16 +206,16 @@ impl<I: Debug> Display for Event<I> {
                 era_id,
                 sender,
                 proto_block,
-                timestamp,
+                parent,
                 valid,
             } => write!(
                 f,
-                "Proto-block received from {:?} at {} for {} is {}: {:?}",
+                "Candidate block received from {:?} for {} with parent {:?} is {}: {:?}",
```
xcthulhu, line 213:
```suggestion
                "Proto-block received from {:?} for {} with parent {:?} is {}: {:?}",
```

node/src/components/consensus.rs
```diff
@@ -104,7 +106,7 @@ pub enum Event<I> {
         era_id: EraId,
         sender: I,
         proto_block: ProtoBlock,
-        timestamp: Timestamp,
+        parent: Option<Digest>,
```
xcthulhu, line 109: Note: move parent into protoblock

node/src/components/consensus/candidate_block.rs
```diff
@@ -85,4 +78,12 @@ impl ConsensusValueT for CandidateBlock {
     fn needs_validation(&self) -> bool {
         !self.proto_block.wasm_deploys().is_empty() || !self.proto_block.transfers().is_empty()
     }
+
+    fn timestamp(&self) -> Timestamp {
+        self.proto_block.timestamp()
```
xcthulhu, line 83: Note: timestamp has moved into `ProtoBlock`, that is why it is missing from all of the arguments.

TBD: move parent into `ProtoBlock` as well.

node/src/components/consensus/protocols/highway.rs
```diff
@@ -311,10 +315,23 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
         if let (Some(value), Some(timestamp), Some(swunit)) =
             (vertex.value(), vertex.timestamp(), vertex.unit())
```
xcthulhu, line 316: Note: this vertex is a proposal

node/src/components/consensus/protocols/highway.rs
```diff
@@ -311,10 +315,23 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
...
         {
+            let panorama = &swunit.wire_unit().panorama;
+            let fork_choice = self.highway.state().fork_choice(panorama);
+            let parent_value =
+                fork_choice.map(|hash| self.highway.state().block(hash).value.hash());
+            if timestamp != value.timestamp() || parent_value.as_ref() != value.parent() {
```
xcthulhu, line 322: Note: here we enforce the unit wrapping the proposed candidate block must have the same timestamp as the candidate block.

xcthulhu (approved):

xcthulhu:
After review, it looks like we want timestamps to *NOT* be in protoblocks/candidate blocks.  They are already in units, and putting them in another structure introduces an invariant we have to enforce.

Similarly the parent is also redundant (it can be determined through the `fork_choice` method).

These simplifications do not need to be done here, however.

node/src/components/consensus/candidate_block.rs
afck: …or move timestamp out again, as discussed. :sweat_smile:

goral09 (approved):

PR #1224, opened 2021-03-25
OP-1940: add back removed true

TomVasile:
Adding back a pipe to true that was removed in a recent pr.


TomVasile:
for reference to the issue: https://github.com/Supervisor/supervisor/issues/1223

goral09 (approved):

TomVasile:
reference: https://github.com/CasperLabs/casper-node/pull/899/files

TomVasile:
reference: removed here, https://github.com/CasperLabs/casper-node/pull/1220

PR #1226, opened 2021-03-25
NDRS-1030: Disconnect from faulty peers

marc-casperlabs:
This PR creates a new announcement for blocklisting, which is currently picked up by `small_network`. If enabled via configuration (currently disabled by default), the peer will be blocked. Regardless of configuration the event is logged.

As a consequence, it is possible to punish peers not only from consensus. As a nice side effect, the `ConsensusAnnouncement` is no longer generic over `I`.

### DIFF
--- a/node/src/reactor/validator.rs
+++ b/node/src/reactor/validator.rs
@@ -170,7 +170,7 @@ pub enum Event {
     DeployAcceptorAnnouncement(#[serde(skip_serializing)] DeployAcceptorAnnouncement<NodeId>),
     /// Consensus announcement.
     #[from]
-    ConsensusAnnouncement(#[serde(skip_serializing)] ConsensusAnnouncement<NodeId>),
+    ConsensusAnnouncement(#[serde(skip_serializing)] ConsensusAnnouncement),
     /// ContractRuntime announcement.
     #[from]
     ContractRuntimeAnnouncement(#[serde(skip_serializing)] ContractRuntimeAnnouncement),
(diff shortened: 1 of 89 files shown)


node/src/reactor/validator.rs
```diff
@@ -185,6 +186,9 @@ pub enum Event {
     /// Chainspec loader announcement.
     #[from]
     ChainspecLoaderAnnouncement(#[serde(skip_serializing)] ChainspecLoaderAnnouncement),
+    /// Blocklist announcement.
+    #[from]
+    BlocklistAnnouncement(BlocklistAnnouncement<NodeId>),
```
afck, line 191: Most variants `#[serde(skip_serializing)]` — not sure which ones should and shouldn't?

afck (approved):
Looks good to me! :+1: 
It's not actually a blacklist/"blocklist" (yet?), though — it just disconnects but would allow the peer to reconnect immediately, would it?

node/src/reactor/validator.rs
marc-casperlabs: `skip_serializing` is used for values where there is no `Serialize` implementation. Some might not have been updated.

marc-casperlabs:
> It's not actually a blacklist/"blocklist" (yet?), though — it just disconnects but would allow the peer to reconnect immediately, would it?

There's a parameter that decides blocking on `fn remove()`, the last one.

xcthulhu (approved):

Fraser999 (approved):
LGTM, but shouldn't we wire it up for the joiner reactor too?  And if so, should we try and persist the blocklist from joiner to validator instances of small_network?

PR #1227, opened 2021-03-25
NDRS-{1052,1053}: Prevent deploy replays. (Ports #1216, #1218, #1219, #1223 to dev.)

afck:
This is a port of https://github.com/CasperLabs/casper-node/pull/1216, https://github.com/CasperLabs/casper-node/pull/1218, https://github.com/CasperLabs/casper-node/pull/1219, https://github.com/CasperLabs/casper-node/pull/1223 to the `dev` branch.

* Instead of persisting the `BlockProposerDeploySets` to LMDB and loading in on startup, we recreate that state from the collection of finalized deploys. This fixes a bug where that set would be incorrectly initialized for joining/restarting/upgrading nodes.
* Blocks are considered invalid if they contain the same deploy twice, or if they contain a deploy that was already present in any of their ancestors.

https://casperlabs.atlassian.net/browse/NDRS-1052
https://casperlabs.atlassian.net/browse/NDRS-1053

### DIFF
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -10,7 +10,7 @@
 mod keyed_counter;
 
 use std::{
-    collections::{hash_map::Entry, HashMap, HashSet, VecDeque},
+    collections::{hash_map::Entry, BTreeMap, HashMap, HashSet, VecDeque},
     convert::Infallible,
     fmt::Debug,
     sync::Arc,
@@ -18,6 +18,7 @@ use std::{
 
 use datasize::DataSize;
 use derive_more::{Display, From};
+use itertools::Itertools;
 use smallvec::{smallvec, SmallVec};
 use tracing::info;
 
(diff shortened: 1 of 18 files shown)


node/src/components/block_validator.rs
```diff
@@ -149,11 +149,20 @@ where
...
+                let deploy_count = block_deploys.len();
+                // Collect the deploys in a set; this also deduplicates them.
+                let block_deploys: HashSet<_> = block_deploys
                     .iter()
                     .map(|deploy_hash| **deploy_hash)
-                    .collect::<HashSet<_>>();
+                    .collect();
+                if block_deploys.len() != deploy_count {
+                    info!(
+                        deploys = ?block.deploys(), ?sender,
```
goral09, line 161: really, we're logging 2600 deploy hashes? or even whole bodies?

goral09 (approved):

node/src/components/block_validator.rs
afck: If we receive a block with an attempted replay attack? Why not?

afck: On the other hand: The deploy hashes won't actually tell us much; we should rather log the block proposer maybe?

goral09: My question is _why_?

goral09: I would say that we should log the block proposer and replayed hashes. This shouldn't happen often so we can afford finding duplicates.

afck: Turns out we only have a `BlockLike` there, not a `Block`, so no proposer etc.
Maybe we should move this check into consensus after all?

afck: As discussed, I changed the log to only list the duplicated deploys.

PR #1228, opened 2021-03-25
NDRS-967: SecretKey constructor should return Result

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-967

CHANGELOG:

- Changed the `SecretKey::ed25519` and `secp256k1` constructors to return `Result` 
- Added two `From` impls to reduce the amount of refactoring required.
- Refactored code as necessary


types/src/crypto/asymmetric_key.rs
```diff
@@ -133,15 +133,19 @@ impl SecretKey {
...
+            Ok(key) => Ok(SecretKey::Ed25519(key)),
+            Err(error) => Err(Error::AsymmetricKey(error.to_string())),
+        }
     }
 
     /// Constructs a new secp256k1 variant from a byte array.
-    pub fn secp256k1(bytes: [u8; Self::SECP256K1_LENGTH]) -> Self {
-        SecretKey::Secp256k1(bytes)
+    pub fn secp256k1(bytes: [u8; Self::SECP256K1_LENGTH]) -> Result<Self, Error> {
+        Ok(SecretKey::Secp256k1(bytes))
```
xcthulhu, line 148: This has to return an error if bytes is all zeros (that's an invalid private key for Secp256k1)

types/src/crypto/asymmetric_key.rs
```diff
@@ -417,6 +421,19 @@ impl From<SecretKey> for PublicKey {
...
+impl From<&Result<SecretKey, Error>> for PublicKey {
+    fn from(secret_key: &Result<SecretKey, Error>) -> Self {
+        let key = secret_key.as_ref().unwrap();
+        PublicKey::from(key)
+    }
+}
+
+impl From<Result<SecretKey, Error>> for PublicKey {
+    fn from(key: Result<SecretKey, Error>) -> Self {
+        PublicKey::from(key.unwrap())
```
xcthulhu, line 433: These need to be `TryFrom`s, and in general we should never have `From<Result<_, _>>`

xcthulhu (changes requested):
Sadly, we can't move the `unwrap()` callsites around.  Bit of a pain, but these are going to eventually surface in the node.  We have to emit `fatal!` effects as appropriate so we can properly shutdown when we hit these.

types/src/crypto/asymmetric_key.rs
xcthulhu: I think if we call the k256 private key constructor here it'll throw that error for us (we'll need to have a `From` impl for the `Error` it thows however).

types/src/crypto/asymmetric_key.rs
```diff
@@ -133,15 +133,19 @@ impl SecretKey {
...
-        // safe to unwrap as `SecretKey::from_bytes` can only fail if the provided slice is the
-        // wrong length.
-        SecretKey::Ed25519(ed25519_dalek::SecretKey::from_bytes(&bytes).unwrap())
+    pub fn ed25519(bytes: [u8; Self::ED25519_LENGTH]) -> Result<Self, Error> {
+        // while it is safe to unwrap as `SecretKey::from_bytes` can only fail if the provided slice
+        // is the wrong length.
+        // We should be disciplined about unwrapping in constructors.
+        match ed25519_dalek::SecretKey::from_bytes(&bytes) {
+            Ok(key) => Ok(SecretKey::Ed25519(key)),
+            Err(error) => Err(Error::AsymmetricKey(error.to_string())),
```
xcthulhu, line 142: Can we make a `From` impl for these `ed25519_dalek` errors?  We should just have variants that wrap them if possible rather than have these strings.

types/src/crypto/asymmetric_key.rs
```diff
@@ -209,8 +213,8 @@ impl Clone for SecretKey {
     fn clone(&self) -> Self {
         match self {
             SecretKey::System => SecretKey::System,
-            SecretKey::Ed25519(sk) => Self::ed25519(*sk.as_bytes()),
-            SecretKey::Secp256k1(sk) => Self::secp256k1(*sk),
+            SecretKey::Ed25519(sk) => Self::ed25519(*sk.as_bytes()).unwrap(),
```
Fraser999, line 216: If we're being strict about not unwrapping in constructors, then I think we should be consistently zealous here.  We should remove the `Clone` impl and provide something like `try_clone(&self) -> Result<Self, Error>` I guess?

types/src/crypto/asymmetric_key.rs
```diff
@@ -283,13 +287,13 @@ impl FromBytes for SecretKey {
             ED25519_TAG => {
                 let (raw_bytes, remainder): ([u8; Self::ED25519_LENGTH], _) =
                     FromBytes::from_bytes(remainder)?;
-                let secret_key = Self::ed25519(raw_bytes);
+                let secret_key = Self::ed25519(raw_bytes).unwrap();
```
Fraser999, line 290: This method already returns a `Result` - why are we unwrapping here and just below?

types/src/crypto/asymmetric_key.rs
Fraser999: Agreed, but we should also remove `impl From<&SecretKey> for PublicKey` and `impl From<SecretKey> for PublicKey` since they're effectively constructors which have panicking code.

types/src/crypto/asymmetric_key.rs
Fraser999: I think our unusual approach of just making the errors a string in this case is justified.  The [`ed25519::Error` type](https://docs.rs/signature/1.2.2/signature/struct.Error.html) and [`k256::Error` type](https://docs.rs/elliptic-curve/0.9.6/src/elliptic_curve/error.rs.html#10) are deliberately opaque, and the resulting log messages extremely unhelpful.

I don't think we should just do `Err(Error::AsymmetricKey(error.to_string()))` though, as that will just produce the unhelpful text "signature error".  Instead I think we should have something like "failed to construct Ed25519 secret key from bytes" which leaks no details about the secret key bytes, but is specific enough to be meaningful for debugging purposes.

types/src/crypto/asymmetric_key.rs
Fraser999: I agree about using the `k256::SecretKey::from_bytes()` constructor to validate these bytes, but for the same reason as my comment on the Ed25519 constructor, I'd avoid providing a `From` impl for that error type.

Fraser999 (changes requested):

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -894,23 +896,23 @@ where
         let genesis_delegators: Vec<_> = self.exec_config.get_bonded_delegators().collect();
 
         // Make sure all delegators have corresponding genesis validator entries
-        for (&validator_public_key, &delegator_public_key, _balance, delegated_amount) in
-            &genesis_delegators
+        for (validator_public_key, delegator_public_key, _balance, delegated_amount) in
+            genesis_delegators.clone().into_iter()
```
xcthulhu, line 900:
```suggestion
            genesis_delegators.iter()
```

You are cloning below anyway, no need to clone here.

node/src/crypto/asymmetric_key.rs
```diff
@@ -64,25 +61,10 @@ pub fn verify<T: AsRef<[u8]>>(
...
-                message.as_ref(),
-                &ed25519_dalek::Signature::from_bytes(signature).map_err(|_| {
-                    Error::AsymmetricKey(format!(
-                        "failed to construct Ed25519 signature from {:?}",
-                        &signature[..]
-                    ))
-                })?,
-            )
+            .verify_strict(message.as_ref(), signature)
             .map_err(|_| Error::AsymmetricKey(String::from("failed to verify Ed25519 signature"))),
```
xcthulhu, line 65: Can we add an error variant here instead of using this string?

node/src/crypto/error.rs
```diff
@@ -1,6 +1,7 @@
 use std::result;
 
 use base64::DecodeError;
+use ed25519_dalek::ed25519::Error as SignatureError;
```
xcthulhu, line 4:
```suggestion
use signature::Error as SignatureError;
```

I _think_ that's what we want here.

node/src/crypto/error.rs
```diff
@@ -53,6 +54,10 @@ pub enum Error {
     /// Error trying to manipulate the system key.
     #[error("invalid operation on system key: {0}")]
     System(String),
+
+    /// Error related to the underlying signature crate
+    #[error("Error in signature")]
+    Signature(SignatureError),
```
xcthulhu, line 60: @Fraser999 With this error message @darthsiroftardis isn't leaking any secrets in the Display impl.  Do we want to have a custom Debug impl too?

xcthulhu: We have this other error variant that might be problematic:

```rust
    /// Pem format error.
    #[error("pem error: {0}")]
    FromPem(String),
```

https://github.com/CasperLabs/casper-node/pull/1228/files#diff-c23caa49a8f80268f2deff3f2e467bdf5d855e1784c2927151f040c0018caeedR40

@Fraser999 I propose getting rid of these error variants that have arbitrary strings as payloads (in a separate story), as they are security hazards.

node/src/crypto/asymmetric_key_ext.rs
```diff
@@ -93,13 +93,13 @@ impl AsymmetricKeyExt for SecretKey {
...
         getrandom::getrandom(&mut bytes[..]).expect("RNG failure!");
-        Ok(SecretKey::ed25519(bytes))
+        Ok(SecretKey::ed25519_from_bytes(bytes).unwrap())
     }
 
     fn generate_secp256k1() -> Result<Self, Error> {
         let mut bytes = [0u8; Self::SECP256K1_LENGTH];
         getrandom::getrandom(&mut bytes[..]).expect("RNG failure!");
-        Ok(SecretKey::secp256k1(bytes))
+        Ok(SecretKey::secp256k1_from_bytes(bytes).unwrap())
```
xcthulhu, line 102: Why the `unwrap`?  We're returning a `Result<Self, Error>`...

node/src/crypto/asymmetric_key_ext.rs
```diff
@@ -93,13 +93,13 @@ impl AsymmetricKeyExt for SecretKey {
     fn generate_ed25519() -> Result<Self, Error> {
         let mut bytes = [0u8; Self::ED25519_LENGTH];
         getrandom::getrandom(&mut bytes[..]).expect("RNG failure!");
```
xcthulhu, line 95: Maybe we can extend the `Error` variant here to include these `rand::Error`s and our crypto errors?

node/src/crypto/asymmetric_key_ext.rs
```diff
@@ -139,7 +139,7 @@ impl AsymmetricKeyExt for SecretKey {
                 der = Der::new(&mut encoded);
                 der.sequence(|der| {
                     der.integer(&[1])?;
-                    der.octet_string(secret_key)?;
+                    der.octet_string(secret_key.to_bytes().as_slice())?;
```
xcthulhu, line 142: `to_bytes` is definitely not what we want here :(

We need to do `.into::<Vec<u8>>()`

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: Right, but then it would be moved here then. I should ideally remove the following clone then

types/src/crypto/asymmetric_key.rs
```diff
@@ -209,8 +184,12 @@ impl Clone for SecretKey {
     fn clone(&self) -> Self {
         match self {
             SecretKey::System => SecretKey::System,
-            SecretKey::Ed25519(sk) => Self::ed25519(*sk.as_bytes()),
-            SecretKey::Secp256k1(sk) => Self::secp256k1(*sk),
+            SecretKey::Ed25519(sk) => Self::ed25519_from_bytes(*sk.as_bytes()).unwrap(),
+            SecretKey::Secp256k1(sk) => {
+                let raw_bytes: [u8; SECP256K1_SECRET_KEY_LENGTH] =
+                    sk.to_bytes().as_slice().try_into().unwrap();
+                Self::secp256k1_from_bytes(&raw_bytes).unwrap()
```
xcthulhu, line 191: `k256::ecdsa::SigningKey` doesn't implement `Clone` according to [the docs][1].

We should consider following suite. 

[1]: https://docs.rs/k256/0.7.2/k256/ecdsa/struct.SigningKey.html

xcthulhu: (in a separate ticket)

node/src/crypto/asymmetric_key.rs
darthsiroftardis: I figure that work would be part of the follow up PR

types/src/crypto/asymmetric_key.rs
```diff
@@ -168,39 +148,34 @@ impl AsymmetricType for SecretKey {
...
+
+impl From<SecretKey> for Vec<u8> {
+    fn from(secret_key: SecretKey) -> Vec<u8> {
+        match secret_key {
+            SecretKey::System => Vec::new(),
+            SecretKey::Ed25519(key) => key.as_ref().into(),
+            SecretKey::Secp256k1(key) => key.to_bytes().as_slice().into(),
+        }
+    }
+}
```
xcthulhu, line 171: We shouldn't have this variant, only `impl From<&SecretKey> for Vec<u8> {`

xcthulhu: Well... maybe we can't get away with that...

types/src/crypto/asymmetric_key.rs
```diff
@@ -433,7 +377,7 @@ impl Debug for PublicKey {
             formatter,
             "PublicKey::{}({})",
             self.variant_name(),
-            HexFmt(self)
+            HexFmt(Into::<Vec<u8>>::into(self.clone()))
```
xcthulhu, line 380:
```suggestion
            HexFmt(Into::<Vec<u8>>::into(self))
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -444,7 +388,7 @@ impl Display for PublicKey {
             formatter,
             "PubKey::{}({:10})",
             self.variant_name(),
-            HexFmt(self)
+            HexFmt(Into::<Vec<u8>>::into(self.clone()))
```
xcthulhu, line 391:
```suggestion
            HexFmt(Into::<Vec<u8>>::into(self))
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -460,7 +404,7 @@ impl Ord for PublicKey {
         let self_tag = self.tag();
         let other_tag = other.tag();
         if self_tag == other_tag {
-            self.as_ref().cmp(other.as_ref())
+            Into::<Vec<u8>>::into(self.clone()).cmp(&Into::<Vec<u8>>::into(other.clone()))
```
xcthulhu, line 407:
```suggestion
            Into::<Vec<u8>>::into(self).cmp(&Into::<Vec<u8>>::into(other))
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -473,7 +417,7 @@ impl Ord for PublicKey {
 impl Hash for PublicKey {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.tag().hash(state);
-        self.as_ref().hash(state);
+        Into::<Vec<u8>>::into(self.clone()).hash(state);
```
xcthulhu, line 420:
```suggestion
        Into::<Vec<u8>>::into(self).hash(state);
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -831,41 +786,43 @@ mod detail {
...
 
-    impl<'a> From<&'a SecretKey> for AsymmetricTypeAsBytes<'a> {
-        fn from(secret_key: &'a SecretKey) -> Self {
+    impl From<&SecretKey> for AsymmetricTypeAsBytes {
+        fn from(secret_key: &SecretKey) -> Self {
             match secret_key {
                 SecretKey::System => AsymmetricTypeAsBytes::System,
-                SecretKey::Ed25519(ed25519) => AsymmetricTypeAsBytes::Ed25519(ed25519.as_ref()),
-                SecretKey::Secp256k1(secp256k1) => AsymmetricTypeAsBytes::Secp256k1(secp256k1),
+                key @ SecretKey::Ed25519(_) => AsymmetricTypeAsBytes::Ed25519(key.clone().into()),
```
xcthulhu, line 799:
```suggestion
                key @ SecretKey::Ed25519(_) => AsymmetricTypeAsBytes::Ed25519(key.into()),
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -831,41 +786,43 @@ mod detail {
...
+                key @ SecretKey::Secp256k1(_) => {
+                    AsymmetricTypeAsBytes::Secp256k1(key.clone().into())
```
xcthulhu, line 801:
```suggestion
                    AsymmetricTypeAsBytes::Secp256k1(key.into())
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -831,41 +786,43 @@ mod detail {
...
-    impl<'a> From<&'a PublicKey> for AsymmetricTypeAsBytes<'a> {
-        fn from(public_key: &'a PublicKey) -> Self {
+    impl From<&PublicKey> for AsymmetricTypeAsBytes {
+        fn from(public_key: &PublicKey) -> Self {
             match public_key {
                 PublicKey::System => AsymmetricTypeAsBytes::System,
-                PublicKey::Ed25519(ed25519) => AsymmetricTypeAsBytes::Ed25519(ed25519.as_ref()),
-                PublicKey::Secp256k1(secp256k1) => {
-                    AsymmetricTypeAsBytes::Secp256k1(secp256k1.as_ref())
+                key @ PublicKey::Ed25519(_) => AsymmetricTypeAsBytes::Ed25519(key.clone().into()),
```
xcthulhu, line 811:
```suggestion
                key @ PublicKey::Ed25519(_) => AsymmetricTypeAsBytes::Ed25519(key.into()),
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -831,41 +786,43 @@ mod detail {
...
+                key @ PublicKey::Secp256k1(_) => {
+                    AsymmetricTypeAsBytes::Secp256k1(key.clone().into())
```
xcthulhu, line 813:
```suggestion
                    AsymmetricTypeAsBytes::Secp256k1(key.into())
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -831,41 +786,43 @@ mod detail {
...
-    impl<'a> From<&'a Signature> for AsymmetricTypeAsBytes<'a> {
-        fn from(signature: &'a Signature) -> Self {
+    impl From<&Signature> for AsymmetricTypeAsBytes {
+        fn from(signature: &Signature) -> Self {
             match signature {
                 Signature::System => AsymmetricTypeAsBytes::System,
-                Signature::Ed25519(ed25519) => AsymmetricTypeAsBytes::Ed25519(ed25519.as_ref()),
-                Signature::Secp256k1(secp256k1) => {
-                    AsymmetricTypeAsBytes::Secp256k1(secp256k1.as_ref())
+                key @ Signature::Ed25519(_) => AsymmetricTypeAsBytes::Ed25519(key.clone().into()),
```
xcthulhu, line 823:
```suggestion
                key @ Signature::Ed25519(_) => AsymmetricTypeAsBytes::Ed25519(key.into()),
```

types/src/crypto/asymmetric_key.rs
```diff
@@ -831,41 +786,43 @@ mod detail {
...
+                key @ Signature::Secp256k1(_) => {
+                    AsymmetricTypeAsBytes::Secp256k1(key.clone().into())
```
xcthulhu, line 825:
```suggestion
                    AsymmetricTypeAsBytes::Secp256k1(key.into())
```

types/src/crypto/asymmetric_key.rs
xcthulhu: I agree that `SecretKey` should not have `Clone`, but it will add more scope to an already complex PR.

I am nervous about `try_clone`.  We really should never be cloning a private key.

node/src/crypto/asymmetric_key.rs
xcthulhu: I agree.  We shouldn't have error variants with strings since they are security hazards, so we'll clean this up when we do remove those.

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: @xcthulhu I got it working with `iter`

node/src/crypto/error.rs
darthsiroftardis: Suggestion applied

node/src/crypto/asymmetric_key_ext.rs
darthsiroftardis: `unwrap` removed

node/src/crypto/asymmetric_key_ext.rs
darthsiroftardis: Extended the `Error` variant with a new one called `RNGFailure`

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Suggestion applied

types/src/crypto/asymmetric_key.rs
darthsiroftardis: I've changed the original `impl` to `From<&Secret>` but I have kept a thin wrapper for `From<Secret>` as it matches previous code patterns. If it must be removed, I can do so

node/src/crypto/error.rs
xcthulhu: Chatting with @Fraser999, we are thinking that the `Signature` variant here should be broken up into Ed25519 and Secp256k1 variants and we should get rid of the `From` impl.  This way we can better debug.  We'll earmark this for follow up work.

types/src/crypto/asymmetric_key.rs
xcthulhu: Yeah, let's get rid of it I think.

darthsiroftardis: Oh I can't remove it cause its a trait bound on `AsymetricType`

xcthulhu (approved):

node/src/types/chainspec/accounts_config/delegator_config.rs
```diff
@@ -9,12 +9,12 @@ use casper_types::{
     PublicKey,
 };
 #[cfg(test)]
-use casper_types::{SecretKey, U512};
+use casper_types::{AsymmetricType, SecretKey, U512};
 
 #[cfg(test)]
 use crate::testing::TestRng;
 
-#[derive(PartialEq, Ord, PartialOrd, Eq, Serialize, Deserialize, DataSize, Debug, Copy, Clone)]
```
goral09, line 17: Why is it not `Copy` anymore?

node/src/types/block.rs
```diff
@@ -450,26 +458,30 @@ impl FinalizedBlock {
...
+                    PublicKey::from(&SecretKey::ed25519_from_bytes(rng.gen::<[u8; 32]>()).unwrap())
                 })
                 .take(inactive_count)
                 .collect(),
             })
         } else {
             None
         };
-        let secret_key: SecretKey = SecretKey::ed25519(rng.gen());
+        let secret_key: SecretKey = SecretKey::ed25519_from_bytes(rng.gen::<[u8; 32]>()).unwrap();
```
goral09, line 484: What's the point of making the constructor return `Result` if we're not taking advantage of it? Is there going to be a follow-up task to do this?

xcthulhu: This is just test code so we can just unwrap here I think...

node/src/types/chainspec/accounts_config/delegator_config.rs
xcthulhu: `k256::ecdsa::Signature` doesn't support `Copy` so we can't.

darthsiroftardis:
> The majority of the changes in this PR is about adding `clone()` method call b/c some types are not `Copy` anymore. What's the reason behind it? What type changed that forced this ripple effect?

So the definition of the `Secp256k1` variant changed, it now wraps around the underlying `k256` type which doesn't implement `Copy` . Hence the ripple effect of removing `Copy`.

node/src/types/chainspec/accounts_config/account_config.rs
```diff
@@ -10,14 +10,14 @@ use casper_types::{
...
 #[cfg(test)]
-use casper_types::{SecretKey, U512};
+use casper_types::{AsymmetricType, SecretKey, U512};
 
 #[cfg(test)]
 use crate::testing::TestRng;
 
 use super::ValidatorConfig;
 
-#[derive(PartialEq, Ord, PartialOrd, Eq, Serialize, Deserialize, DataSize, Debug, Copy, Clone)]
```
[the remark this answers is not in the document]
darthsiroftardis: No, we now wrap around `k256::ecdsa::VerifyingKey` for the Secp variant for the public key, it doesn't implement copy, so we had to get rid of that trait

node/src/types/block.rs
goral09: It was a random place where I put the comment. I meant the whole PR .

node/src/types/chainspec/accounts_config/account_config.rs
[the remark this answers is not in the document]
goral09: This should have been done in a separate PR though. I tried reviewing "constructor should return `Result`" PR and not "change the type used for the secret key" PR.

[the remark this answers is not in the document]
darthsiroftardis: It also changes how the constructor ought to behave, previously we were creating "valid" keys from `[0;32]` slices, which I believe is incorrect , @xcthulhu you can correct me on this. At worst I think the PR is improperly titled

Fraser999 (approved):

PR #1231, opened 2021-03-27
NO-TICKET: improve Key error reporting and tests

Fraser999:
This PR adjusts the way in which `Key` enums are parsed from their formatted string formats.  The objective is to give more meaningful feedback to users when a parsing error occurs.

Previously we'd try to parse to each variant, and once all attempts had failed we'd report `InvalidPrefix`.  This was misleading in the case where the user provided a valid prefix, but the Key still failed parsing.  For example, we saw multiple cases where the user mistakenly supplied a public key instead of an account hash.  This would result in a string like `account-hash-<66 hex chars>` yielding an `InvalidPrefix` error, when in fact the prefix was correct, but the body was incorrect.

Now parsing will be attempted based on the supplied prefix, and will only fall through to the `UnknownPrefix` error if the prefix doesn't match one of the valid prefixes.  The example mentioned above now yields an error stating `account-key from string error: address portion is wrong length: could not convert slice to array`.

This PR also improves the unit test coverage of `Key` since some tests were not covering newer `Key` variants.


types/src/key.rs
```diff
@@ -239,39 +234,73 @@ impl Key {
...
+                .map_err(|error| FromStrError::DeployInfo(error.to_string()))?;
+            return Ok(Key::DeployInfo(DeployHash::new(hash_array)));
         }
+
+        if input.starts_with(TRANSFER_ADDR_FORMATTED_STRING_PREFIX) {
+            let transfer_addr = TransferAddr::from_formatted_str(input)?;
+            return Ok(Key::Transfer(transfer_addr));
+        }
+
+        if input.starts_with(UREF_FORMATTED_STRING_PREFIX) {
```
mpapierski, line 263: Not a blocker but just something that occurred to me that there is duplicated logic - first, this checks prefix with `starts_with` and then `URef::from_formatted_str` does `strip_prefix` inside doing the same work twice. Alternatively, wouldn't this work for `T::from_formatted_str` if we already have T's error type connected through From traits?

```rust
match URef::from_formatted_str(input) {
  Ok(uref) => return Key::URef(uref),
  Err(uref::FromStrError::InvalidPrefix) => {}
  Err(e) => return Err(e.into()),
}
```

Fraser999: Yup, makes sense.  Updated in [28599eb](https://github.com/CasperLabs/casper-node/pull/1231/commits/28599ebdd84231ce637399d24bbafb75e900a2a3).

mpapierski (approved):

darthsiroftardis (approved):

Fraser999:
bors merge

PR #1237, opened 2021-03-29
NDRS-1049: Drain reactor queue switch

marc-casperlabs:
This prevents fatal errors from being skipped by clearing the `Control` queue and looking for datal events before successfully exiting.

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -702,6 +702,24 @@ where
             match TERMINATION_REQUESTED.load(Ordering::SeqCst) as i32 {
                 0 => {
                     if let Some(reactor_exit) = self.reactor.maybe_exit() {
+                        // TODO: Workaround, until we actually use control announcements for
+                        // exiting: Go over the entire remaining event queue and look for a control
+                        // announcement. This approach is hacky, and should be replaced with
+                        // `ControlAnnouncement` handling instead.
+
+                        for event in self.scheduler.drain_queue(QueueKind::Control).await {
+                            if let Some(ctrl_ann) = event.as_control() {
+                                match ctrl_ann {
+                                    ControlAnnouncement::FatalError { file, line, msg } => {
+                                        warn!(%file, line=*line, %msg, "exiting due to fatal error scheduled before reactor completion");
+                                        return ReactorExit::ProcessShouldExit(ExitCode::Abort);
+                                    }
+                                }
+                            } else {
+                                warn!(%event, "found non-control announcement while draining queue")
+                            }
+                        }
+
                         break reactor_exit;
                     }
                     if !self.crank(rng).await {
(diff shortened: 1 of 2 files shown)


node/src/utils/round_robin.rs
```diff
@@ -58,6 +61,14 @@ impl<I> QueueState<I> {
         }
     }
 
+    /// Remove all events from a queue.
+    async fn drain(&self) -> Vec<I> {
```
goral09, line 65: Could this return an iterator instead?

node/src/reactor.rs
```diff
@@ -702,6 +702,24 @@ where
...
+                        for event in self.scheduler.drain_queue(QueueKind::Control).await {
+                            if let Some(ctrl_ann) = event.as_control() {
+                                match ctrl_ann {
+                                    ControlAnnouncement::FatalError { file, line, msg } => {
+                                        warn!(%file, line=*line, %msg, "exiting due to fatal error scheduled before reactor completion");
+                                        return ReactorExit::ProcessShouldExit(ExitCode::Abort);
+                                    }
+                                }
+                            } else {
+                                warn!(%event, "found non-control announcement while draining queue")
```
goral09, line 719: Won't we end up with many many warns?

goral09 (approved):

goral09:
@marc-casperlabs Are you planning on merging this?

marc-casperlabs:
This is temporarily going back into draft mode to address a potential issue with the queue's concurrency safety.

Fraser999 (approved):

node/src/utils/round_robin.rs
marc-casperlabs: An iterator would block the queue permanently while it exists.

```
1. LOCK QUEUE
2. DRAIN ITEMS (<--- iterator would replace this step and would need to hold the lock)
3. UNLOCK QUEUE
```

is what is happening here.

If we returned an iterator, we'd essentially hold the lock as long as it is being iterated, and we would need a guard type to do so. I'm not sure this is worth the extra effort?

node/src/reactor.rs
marc-casperlabs: Yes, you're right! I'll downgrade this to a `debug!` message.

node/src/utils/round_robin.rs
goral09: Maybe not, I was just worried that "materializing" all the events into a `Vec` may be problematic.

marc-casperlabs: They are just being moved. The "draining" happens twice over the lifetime of a node, as in only when we change from initializer to joiner, and from joiner to validator. I think it is fairly unproblematic, and we are aiming to replace this in the mid-term.

PR #1240, opened 2021-03-29
NO-TICKET: Add a unit test for ping out of bound exception.

goral09:


node/src/components/consensus/highway_core/highway.rs
```diff
@@ -904,4 +905,23 @@ pub(crate) mod tests {
...
+        let highway = Highway {
+            instance_id: TEST_INSTANCE_ID,
+            validators: test_validators(),
+            state,
+            active_validator: None,
+        };
+
+        // Ping by validator that is not bonded, with an index that is outside of boundaries of the
+        // state.
+        let ping: Vertex<TestContext> = Vertex::Ping(Ping::new(DAN, now, &DAN_SEC));
```
afck, line 923: Maybe just so this test remains relevant even if we change the default weights:
```rust
        assert!(DAN_SEC.0 >= WEIGHTS.len());
```

afck (approved):

node/src/components/consensus/highway_core/highway.rs
goral09: I was just adding this right now :)

goral09: Added in [`37c07c1` (#1240)](https://github.com/CasperLabs/casper-node/pull/1240/commits/37c07c15d45fb29a9f0d251252722e59b0f1411f)

PR #1244, opened 2021-03-29
HWY-278: Add an instance ID to Pings.

afck:
While not as important as for other vertex types, it's still better to distinguish pings by instance ID, so that they are only handled in the era and the network they were meant for.

https://casperlabs.atlassian.net/browse/HWY-278


goral09 (commented):
Are we going to validate this additional information on incoming pings?

afck:
Good catch! :man_facepalming: 
Fixed in https://github.com/CasperLabs/casper-node/pull/1244/commits/a0c2c94cf74f113b1775850934d392c3344dd038.

goral09 (approved):

PR #1249, opened 2021-03-30
NDSR-1048: Fix fetcher metrics.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-1048

This PR fixes a bug where `Fetcher` would incorrectly count the number of requests that time out. Previously, we would consider every request made as timed out as `TimeoutPeer` events are not canceled if the original fetch request had completed successfully. 

### Note: ###
I am not happy with the fact that I had to put `metrics` method into the `ItemFetcher` trait but b/c of how the abstractions are written and the fact that the whole logic/information about whether the request really timed-out or not is hidden inside the `signal` method (that is implemented on the `ItemFetcher` trait), there was no easy way to extract this information without a major rewrite. I've spent a lot of time on this already so I went for a correct solution right now.

### DIFF
--- a/node/src/components/fetcher.rs
+++ b/node/src/components/fetcher.rs
@@ -32,6 +32,8 @@ pub use config::Config;
 pub use event::{Event, FetchResult};
 use metrics::FetcherMetrics;
 
+use self::event::ItemResponders;
+
 /// A helper trait constraining `Fetcher` compatible reactor events.
 pub trait ReactorEventT<T>:
     From<Event<T>>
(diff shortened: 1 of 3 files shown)


xcthulhu (changes requested):
This isn't compatible with NDRS-1001, so we have to rethink how we are going to do this...

goral09:
> This isn't compatible with NDRS-1001, so we have to rethink how we are going to do this...

What is NDRS-1001? I can't see PR for it. Also, this isn't changing the functionality, it's just fixing a bug.

goral09:
> This isn't compatible with NDRS-1001, so we have to rethink how we are going to do this...

Can you explain why _this_ PR makes the `Fetcher` incompatible with NDRS-1001?

xcthulhu:
I've incorporated these changes into  #1080, closing in favour of that PR.

PR #1256, opened 2021-03-31
NO-TICKET: Limit message size in chainspec

marc-casperlabs:
This ensures that nodes are in agreement about what the maximum message size is through the chainspec.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -158,6 +158,8 @@ where
     /// Name of the network we participate in. We only remain connected to peers with the same
     /// network name as us.
     network_name: String,
+    /// The maximum message size for a network message, as supplied from the chainspec.
+    maximum_net_message_size: u32,
     /// Channel signaling a shutdown of the small network.
     // Note: This channel is closed when `SmallNetwork` is dropped, signalling the receivers that
     // they should cease operation.
@@ -196,6 +198,7 @@ where
         registry: &Registry,
         small_network_identity: SmallNetworkIdentity,
         network_name: String,
+        maximum_net_message_size: u32,
         notify: bool,
     ) -> Result<(SmallNetwork<REv, P>, Effects<Event<P>>)> {
         let mut known_addresses = HashSet::new();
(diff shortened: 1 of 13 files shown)


goral09:
I don't think this should be a chainspec config after all. I think the networking should allow arbitrarily large messages and if there's a business/domain limit then that should be applied in a proper component rather than the networking layer.

marc-casperlabs:
> I don't think this should be a chainspec config after all.

This is somewhat debateable, but the maximum size of anything sent across the network depends on values from the chainspec, so in theory we should be able to calculate it from the chainspec anyway. Right now, we just shortcut the otherwise somewhat complex estimate by putting in an explicit value.

It would be perfectly fine to have this value configurable by node operators, except that it would not gain us anything while at the same potentially breaking things. If nodes differ in their opinion on the maximum acceptable message size for network messages, we will run into issues.

To the best of my knowledge, the chainspec contains "values all nodes need to agree on" as a rule of thumb, so using that definition, I think it is a good place to keep it.

> I think the networking should allow arbitrarily large messages 

We can never allow arbitrarily large messages because we cannot afford to spend arbitrarily large amounts of memory before starting to validate contents. If large sizes (the current limit is an initial one, bound to come down soon) are required for contents, a chunking scheme will have to be implemented (which is in the works, just not for this release).

> and if there's a business/domain limit then that should be applied in a proper component rather than the networking layer.

This is purely a layer in the defense against a resource exhaustion; the acceptable size for a deploy or other things is still not determined by the networking layer. There just has to be _some_ bound for the network to function, otherwise nodes are trivially crashable.

node/src/components/small_network.rs
```diff
@@ -550,6 +556,7 @@ where
         effect_builder: EffectBuilder<REv>,
         peer_id: NodeId,
         transport: Transport,
+        maximum_net_message_size: u32,
```
goral09, line 559: Does it have to be an argument? Can't you take that from `self`?

goral09 (approved):

node/src/components/small_network.rs
marc-casperlabs: Good catch!

Fraser999 (approved):

PR #1258, opened 2021-03-31
NO-TICKET: limit deploy size to 1MiB

Fraser999:
This PR limits the deploy size, both when accepting one in the node, and when creating one in the client.

The node was previously rejecting deploys greater than the permitted maximum block size.  It now rejects deploys with a serialized size of > 1MiB.  The serialized size is that provided by the `ToBytes` trait for the `Deploy` type.

The client also now checks that any deploy created is within this limit and avoids sending such an invalid deploy to the node.


goral09 (commented):
AFAIU, we wanted to only update the client so that we can release it and it would still be compatible with the current node (even before we get to upgrading the network). I would prefer doing it in separate PRs and make the deploy size configurable via chainspec. We don't need to rush the node changes, we have time before the next release (upgrade).

client/lib/ffi.rs
```diff
@@ -35,23 +35,24 @@ pub enum casper_error_t {
...
-    CASPER_IO_ERROR = -15,
-    CASPER_TO_BYTES_ERROR = -16,
-    CASPER_CRYPTO_ERROR = -17,
-    CASPER_INVALID_CL_VALUE = -18,
-    CASPER_INVALID_ARGUMENT = -19,
-    CASPER_INVALID_RESPONSE = -20,
-    CASPER_FFI_SETUP_NOT_CALLED = -21,
-    CASPER_FFI_PTR_NULL_BUT_REQUIRED = -22,
-    CASPER_CONFLICTING_ARGUMENTS = -23,
+    CASPER_DEPLOY_SIZE_TOO_LARGE = -7,
```
goral09, line 38: Wouldn't it be better to make this backwards compatible and add the new error type at the end, rather than changing all of the exit codes that come after this?

node/src/types/deploy.rs
```diff
@@ -46,6 +46,8 @@ use crate::{
     utils::DisplayIter,
 };
 
+const MAX_SERIALIZED_SIZE: u64 = 1_024 * 1_024;
```
goral09, line 49: This should be part of the chainspec, rather than hardcoded. With this setup it's possible to run from the same chainspec with different binary that has this limit set differently. IMO this type of network-wide configuration should be introduced via chainspec, rather than software version.

dwerner (approved):

client/lib/ffi.rs
EdHastingsCasperLabs: agreed

node/src/types/deploy.rs
EdHastingsCasperLabs: agreed

EdHastingsCasperLabs (approved):
Assuming Mateusz's points are resolved.

client/lib/deploy.rs
```diff
@@ -341,6 +354,40 @@ mod tests {
...
     }
 
+    #[test]
+    fn should_fail_to_create_large_deploy() {
+        let deploy_params = deploy_params();
+        let payment_params =
+            PaymentStrParams::with_package_hash(PKG_HASH, VERSION, ENTRYPOINT, args_simple(), "");
+        // Create a string arg of 1048576 letter 'a's to ensure the deploy is greater than 1048576
+        // bytes.
+        let large_args_simple = format!("name_01:string='{:a<1048576}'", "");
```
goral09, line 364: nice!

goral09 (approved):

Fraser999:
bors merge

PR #1263, opened 2021-03-31
NO-TICKET: Include chainspec version in handshake

marc-casperlabs:
This introduces a protocol version field into the handshake, which is ignored by the V1.0.0 node. At the same time, if absent, V1.0.0 is filled in. As a result, every version of the release node software has a correct protocol version field.

~~**NOTE** This PR is due a rebase & force push once https://github.com/CasperLabs/casper-node/pull/1256 is merged to master.~~

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -32,6 +32,7 @@
 //! No explicit reconnect is attempted. Instead, if the peer is still online, the normal gossiping
 //! process will cause both peers to connect again.
 
+mod chain_info;
 mod config;
 mod error;
 mod event;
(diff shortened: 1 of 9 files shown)


dwerner (approved):
looks good to me

Fraser999 (approved):

node/src/components/small_network.rs
```diff
@@ -760,13 +755,16 @@ where
             Message::Handshake {
                 network_name,
                 public_address,
+                protocol_version,
             } => {
-                if network_name != self.network_name {
+                if network_name != self.chain_info.network_name {
```
goral09, line 760: Hm, so we are including the protocol version in the handshake but not using it? Is that intentional?

marc-casperlabs: Yes, it is right now. It is for future use.

PR #1266, opened 2021-04-01
NDRS-1061: Remove ProtoBlock from FinalizedBlock and rename wasm_deploys.

afck:
This is a refactoring, and a preparation for simplifying the block types, specifically for replacing `ProtoBlock` and `CandidateBlock` with a single type.

This PR removes `ProtoBlock` from `FinalizedBlock`, because the new type will contain fields (`accusations`) that would be redundant as a part of `FinalizedBlock`. It also renames `ProtoBlock::wasm_deploys` to `deploy_hashes`, and `transfers` to `transfer_hashes`, so that it is in line with `BlockBody`.

https://casperlabs.atlassian.net/browse/NDRS-1061

### DIFF
--- a/node/src/components/block_proposer.rs
+++ b/node/src/components/block_proposer.rs
@@ -414,14 +414,14 @@ impl BlockProposerReady {
         let max_block_size_bytes = deploy_config.max_block_size as usize;
         let block_gas_limit = Gas::from(deploy_config.block_gas_limit);
 
-        let mut transfers = Vec::new();
-        let mut wasm_deploys = Vec::new();
+        let mut transfer_hashes = Vec::new();
+        let mut deploy_hashes = Vec::new();
         let mut block_gas_running_total = Gas::zero();
         let mut block_size_running_total = 0usize;
 
         for (hash, deploy_type) in self.sets.pending.iter() {
-            let at_max_transfers = transfers.len() == max_transfers;
-            let at_max_deploys = wasm_deploys.len() == max_deploys
+            let at_max_transfers = transfer_hashes.len() == max_transfers;
+            let at_max_deploys = deploy_hashes.len() == max_deploys
                 || (deploy_type.is_wasm()
                     && block_size_running_total + DEPLOY_APPROX_MIN_SIZE >= max_block_size_bytes);
 
@@ -442,7 +442,7 @@ impl BlockProposerReady {
 
             // always include wasm-less transfers if we are under the max for them
             if deploy_type.is_transfer() && !at_max_transfers {
-                transfers.push(*hash);
+                transfer_hashes.push(*hash);
             } else if deploy_type.is_wasm() && !at_max_deploys {
                 if block_size_running_total + deploy_type.size() > max_block_size_bytes {
                     continue;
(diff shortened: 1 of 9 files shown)


node/src/types/block.rs
```diff
@@ -254,41 +254,35 @@ impl ProtoBlock {
...
     /// The list of deploy hashes included in the block.
-    pub(crate) fn transfers(&self) -> &Vec<DeployHash> {
-        &self.transfers
+    pub(crate) fn transfer_hashes(&self) -> &Vec<DeployHash> {
+        &self.transfer_hashes
     }
 
-    pub(crate) fn deploys_iter(&self) -> impl Iterator<Item = &DeployHash> {
-        self.wasm_deploys().iter().chain(self.transfers().iter())
+    pub(crate) fn deploys_transfers_iter(&self) -> impl Iterator<Item = &DeployHash> {
```
dwerner, line 266: Naming nitpick: I think this should be named `deploys_and_transfers_iter` if that's what it does, as the plurality suggested by `deploys_` and `transfers_` seems less clear as it is currently.

node/src/types/block.rs
```diff
@@ -405,6 +400,11 @@ impl FinalizedBlock {
         self.proposer
     }
 
+    /// Returns an iterator over all deploy and transfer hashes.
+    pub(crate) fn deploys_transfers_iter(&self) -> impl Iterator<Item = &DeployHash> {
```
dwerner, line 404: See my above nitpick on naming.

dwerner (approved):
Looks good to me.

node/src/types/block.rs
afck: Good point; done in https://github.com/CasperLabs/casper-node/pull/1266/commits/929416ca55d109a3547044d9f84532927872aff7.

node/src/components/block_proposer.rs
```diff
@@ -413,14 +414,14 @@ impl BlockProposerReady {
         let max_block_size_bytes = deploy_config.max_block_size as usize;
         let block_gas_limit = Gas::from(deploy_config.block_gas_limit);
 
-        let mut transfers = Vec::new();
-        let mut wasm_deploys = Vec::new();
+        let mut transfer_hashes = Vec::new();
+        let mut deploy_hashes = Vec::new();
```
goral09, line 418: I mean, they're both deploys really.

afck: They are hashes, not full deploys/transfers.
Or do you mean that a "transfer" is also a "deploy"?
I made this change to be in line with the `BlockBody`, where we also use these identifiers. Unfortunately we're a bit inconsistent: In some places transfers are considered to be deploys, in other places "deploy" refers to only the Wasm ones.

goral09 (approved):

node/src/components/block_proposer.rs
goral09: > Or do you mean that a "transfer" is also a "deploy"?

Yes, this.

afck: I'd also like to address this inconsistency, but that's going to be a bigger PR.
And it needs discussion first: Is a "transfer" a "deploy" or isn't it? And what do we call a non-transfer deploy then, or what do we use as an umbrella term?
But I want to avoid just introducing a third kind of terminology and make everything even less consistent.

PR #1268, opened 2021-04-01
NDRS-833: `LinearChain` cleanup part 3.

goral09:
Cleanup of the `LinearChain` component (in preparation for unit tests) continued. The main thing that changed in this PR is the addition of `LinearChainComponent` struct that wraps `LinearChain` and `LinearChainMetrics`. All component-level logic is now moved to the `LinearChainComponent` struct, rather than `LinearChain`. The latter does not deal with `Effect`, `REv`, `EffectBuilder`, `I` (peer id type), or any other code related to the reactor or networking. The functionality should be mostly unchanged. I tried working iteratively so reviewing commit-by-commit should be simpler than trying to do it all at once (although not necessarily take less time). I've found a couple of inconsistencies and potential sources of bugs around the handling of pending/cached finality signatures. Those should be now resolved.

### DIFF
--- a/node/src/components/linear_chain.rs
+++ b/node/src/components/linear_chain.rs
@@ -23,7 +23,7 @@ use crate::{
         EffectBuilder, EffectExt, EffectResultExt, Effects,
     },
     protocol::Message,
-    types::{BlockByHeight, FinalitySignature, Timestamp},
+    types::{BlockByHeight, BlockSignatures, FinalitySignature, Timestamp},
     NodeRng,
 };
 use casper_types::{EraId, ProtocolVersion};
@@ -124,12 +124,35 @@ where
                 block,
                 execution_results,
             } => {
-                let (signatures, mut effects) =
-                    self.linear_chain_state.new_block(effect_builder, &*block);
+                let signatures = self.linear_chain_state.new_block(&*block);
+                let mut effects = Effects::new();
+                if !signatures.is_empty() {
+                    let mut block_signatures =
+                        BlockSignatures::new(*block.hash(), block.header().era_id());
+                    for sig in signatures.iter() {
+                        block_signatures.insert_proof(sig.public_key(), sig.signature());
+                    }
+                    effects.extend(
+                        effect_builder
+                            .put_signatures_to_storage(block_signatures)
+                            .ignore(),
+                    );
+                    for signature in signatures {
+                        if signature.is_local() {
+                            let message =
+                                Message::FinalitySignature(Box::new(signature.to_inner().clone()));
+                            effects.extend(effect_builder.broadcast_message(message).ignore());
+                        }
+                        effects.extend(
+                            effect_builder
+                                .announce_finality_signature(signature.take())
+                                .ignore(),
+                        );
+                    }
+                };
                 let block_to_store = block.clone();
                 effects.extend(
                     async move {
-                        effect_builder.put_signatures_to_storage(signatures).await;
                         let block_hash = *block_to_store.hash();
                         effect_builder.put_block_to_storage(block_to_store).await;
                         effect_builder
--- a/node/src/components/linear_chain.rs
+++ b/node/src/components/linear_chain.rs
@@ -187,7 +187,7 @@ where
                     // correct – block was created in a specific era so the IDs have to match.
                     if known_signatures.era_id != fs.era_id {
                         warn!(public_key=%fs.public_key,
-                            expected=%known_signatures.era_id, 
+                            expected=%known_signatures.era_id,
                             got=%fs.era_id,
                             "finality signature with invalid era id.");
                         self.linear_chain_state.remove_from_pending_fs(&*fs);
@@ -199,7 +199,8 @@ where
                         return Effects::new();
                     }
                     // Populate cache so that next incoming signatures don't trigger read from the
-                    // storage. If `known_signatures` are already from cache then this will be a noop.
+                    // storage. If `known_signatures` are already from cache then this will be a
+                    // noop.
                     self.linear_chain_state
                         .cache_signatures(*known_signatures.clone());
                 }
--- a/node/src/components/linear_chain.rs
+++ b/node/src/components/linear_chain.rs
@@ -227,36 +227,29 @@ where
                         },
                     )
             }
-            Event::IsBonded(Some(mut signatures), fs, true) => {
-                // Known block and signature from a bonded validator.
-                // Check if we had already seen this signature before.
-                if signatures.has_proof(&fs.public_key, &fs.signature) {
-                    self.linear_chain_state.remove_from_pending_fs(&*fs);
-                    Effects::new()
-                } else {
-                    signatures.insert_proof(fs.public_key, fs.signature);
-                    // Cache the results in case we receive the same finality signature before we
-                    // manage to store it in the database.
-                    self.linear_chain_state
-                        .cache_signatures(*signatures.clone());
-                    debug!(hash=%signatures.block_hash, "storing finality signatures");
-                    let mut effects = effect_builder
-                        .announce_finality_signature(fs.clone())
-                        .ignore();
-                    // If new, gossip and store.
-                    if let Some(signature) = self.linear_chain_state.remove_from_pending_fs(&*fs) {
-                        if signature.is_local() {
-                            let message = Message::FinalitySignature(fs.clone());
-                            effects.extend(effect_builder.broadcast_message(message).ignore());
-                        }
-                    };
-                    effects.extend(
-                        effect_builder
-                            .put_signatures_to_storage(*signatures)
-                            .ignore(),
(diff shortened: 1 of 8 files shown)


node/src/components/linear_chain.rs
```diff
@@ -134,39 +180,58 @@ where
...
                             "finality signature with invalid era id.");
+                        self.linear_chain_state.remove_from_pending_fs(&*fs);
                         // TODO: Disconnect from the sender.
-                        self.remove_from_pending_fs(&*fs);
                         return Effects::new();
                     }
-                    // Populate cache so that next finality signatures don't have to read from the
-                    // storage. If signature is already from cache then this will be a noop.
-                    self.cache_signatures(*signatures.clone());
+                    if known_signatures.has_proof(&fs.public_key, &fs.signature) {
```
afck, line 220: I thought we decided that `public_key` is enough? Because it doesn't matter if the same validator created another, different, but also valid signature.

node/src/components/linear_chain.rs
```diff
@@ -185,49 +250,43 @@ where
...
-                    effects
-                }
+            Event::IsBonded(Some(mut known_signatures), fs, true) => {
+                // New finality signature from a bonded validator.
+                known_signatures.insert_proof(fs.public_key, fs.signature);
+                // Cache the results in case we receive the same finality signature before we
+                // manage to store it in the database.
+                self.linear_chain_state
+                    .cache_signatures(*known_signatures.clone());
+                debug!(hash=%known_signatures.block_hash, "storing finality signatures");
```
afck, line 260:
```suggestion
                debug!(hash = %known_signatures.block_hash, "storing finality signatures");
```

node/src/components/linear_chain.rs
```diff
@@ -185,49 +250,43 @@ where
...
+                    // flow must have already removed it. If it's a signature
+                    // created by this node, gossip it.
+                    if signature.is_local() {
+                        let message = Message::FinalitySignature(fs.clone());
+                        effects.extend(effect_builder.broadcast_message(message).ignore());
+                    }
+                };
+                effects.extend(
+                    effect_builder
+                        .put_signatures_to_storage(*known_signatures)
```
afck, line 277: I wonder whether it can happen that the put events get handled in the wrong order.
I don't think it can, currently, but it's also not something components should rely on, in general.

E.g. if you got a signature A from Alice, put it in the cache, and made the put event for `[A]`. Then you got B from Bob, and made the put event for `[A, B]`. Then the latter even is handled, and then the former. The state would now contain only `[A]`.

(Not something for this PR, but maybe to revisit later?)

node/src/components/linear_chain/pending_signatures.rs
```diff
@@ -68,10 +62,10 @@ impl PendingSignatures {
                 %block_hash, %public_key,
                 "received too many finality signatures for unknown blocks"
             );
-            return;
+            return false;
         }
         // Add the pending signature.
-        let _ = sigs.insert(block_hash, signature);
+        sigs.insert(block_hash, signature).is_some()
```
afck, line 68:
```suggestion
        sigs.insert(block_hash, signature).is_none()
```
Would be good to have a unit test for this. (Or for something higher level, that also tests this.)

node/src/components/linear_chain/signature_cache.rs
```diff
@@ -40,28 +40,16 @@ impl SignatureCache {
...
         let FinalitySignature {
             block_hash,
             public_key,
+            signature,
             ..
         } = fs;
         self.signatures
             .get(block_hash)
-            .map_or(false, |bs| bs.has_proof(public_key))
+            .map_or(false, |bs| bs.has_proof(public_key, signature))
```
afck, line 53: I don't think the signature itself should matter.

node/src/types/block.rs
```diff
@@ -1000,8 +1000,11 @@ impl BlockSignatures {
         self.proofs.insert(public_key, signature)
     }
 
-    pub(crate) fn has_proof(&self, public_key: &PublicKey) -> bool {
-        self.proofs.contains_key(public_key)
+    pub(crate) fn has_proof(&self, public_key: &PublicKey, signature: &Signature) -> bool {
+        self.proofs
+            .get(public_key)
+            .iter()
+            .any(|sig| *sig == signature)
```
afck, line 1007: (Shouldn't be necessary; see other comments.)

node/src/components/linear_chain/state.rs
```diff
@@ -191,8 +142,16 @@ impl<I> LinearChain<I> {
...
     }
 
-    // Caches the signature.
-    pub(super) fn cache_signatures(&mut self, signatures: BlockSignatures) {
+    /// Caches the signature.
+    pub(super) fn cache_signatures(&mut self, mut signatures: BlockSignatures) {
+        // Merge already-known signatures and the new ones.
+        self.get_signatures(&signatures.block_hash)
+            .iter()
+            .for_each(|bs| {
```
afck, line 150: This would be less confusing:
```suggestion
        if let Some(bs) = self.get_signatures(&signatures.block_hash) {
```

node/src/components/linear_chain/pending_signatures.rs
goral09: Actually, I think this should always be true anyway. This `add` method should return `true` if the value was added and it doesn't matter whether there was entry under `block_hash` previously or not.

node/src/components/linear_chain.rs
```diff
@@ -134,39 +180,58 @@ where
...
-                    if signatures.era_id != fs.era_id {
+                if let Some(known_signatures) = &maybe_signatures {
+                    // If the newly-received finality signature does not match the era of previously
+                    // validated signatures reject it as they can't both be
+                    // correct – block was created in a specific era so the IDs have to match.
+                    if known_signatures.era_id != fs.era_id {
                         warn!(public_key=%fs.public_key,
-                            expected=%signatures.era_id, got=%fs.era_id,
+                            expected=%known_signatures.era_id,
+                            got=%fs.era_id,
```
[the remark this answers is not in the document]
goral09: Done in [`5ffb8ab` (#1268)](https://github.com/CasperLabs/casper-node/pull/1268/commits/5ffb8ab4bee2a6e7972a9678ef4a3b7ccfc57dd3)

node/src/components/linear_chain.rs
goral09: Done in [`5ffb8ab` (#1268)](https://github.com/CasperLabs/casper-node/pull/1268/commits/5ffb8ab4bee2a6e7972a9678ef4a3b7ccfc57dd3)

node/src/types/block.rs
goral09: Done in [`f5f6038` (#1268)](https://github.com/CasperLabs/casper-node/pull/1268/commits/f5f603830d33577186c4c7072cbb578714d8e773)

node/src/components/linear_chain/signature_cache.rs
goral09: Done in [`f5f6038` (#1268)](https://github.com/CasperLabs/casper-node/pull/1268/commits/f5f603830d33577186c4c7072cbb578714d8e773)

node/src/components/linear_chain.rs
goral09: Done in [`f5f6038` (#1268)](https://github.com/CasperLabs/casper-node/pull/1268/commits/f5f603830d33577186c4c7072cbb578714d8e773)

node/src/components/linear_chain/pending_signatures.rs
goral09: Fixed in [`50f753a` (#1268)](https://github.com/CasperLabs/casper-node/pull/1268/commits/50f753a8773becc78c914139bb5035ba830f7631)

node/src/components/linear_chain.rs
goral09: Storage also combined the signatures together – it adds the new data to the old one. I made sure that's the case when that piece was written.

afck (approved):

node/src/components/linear_chain.rs
```diff
@@ -20,14 +23,46 @@ use crate::{
...
-pub(crate) use state::LinearChain;
+use state::LinearChain;
 
-impl<I, REv> Component<REv> for LinearChain<I>
+#[derive(DataSize, Debug)]
+pub(crate) struct LinearChainComponent<I> {
+    linear_chain_state: LinearChain,
+    #[data_size(skip)]
+    metrics: LinearChainMetrics,
+    _marker: PhantomData<I>,
```
xcthulhu, line 39: Let's get rid of this `PhatomData` marker field

afck: That wouldn't compile, would it?

PR #1271, opened 2021-04-04
NDRS-1113: fix reconnecting if isolated

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1113.

This PR fixes a bug where a node which is a member of the known addresses can become isolated but not retry to connect to other known addresses.

It also improves the logging of an outgoing connection failure in two ways:
1. A lost outgoing connection where the peer ID is known is now only logged once
2. The log messages for peer ID known vs unknown are now even more different to help disambiguate them.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -582,7 +582,7 @@ where
                 local_address=?transport.get_ref().local_addr(),
                 "connected outgoing to ourself - closing connection",
             );
-            return Effects::new();
+            return self.reconnect_if_not_connected_to_any_known_addresses(effect_builder);
         }
 
         // The stream is only used to receive a single handshake message and then dropped.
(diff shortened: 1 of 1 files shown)


goral09:
I honestly thought that it's the intended behavior. I've raised this a couple of weeks back and it was decided to leave as-is.

node/src/components/small_network.rs
```diff
@@ -582,7 +582,7 @@ where
                 local_address=?transport.get_ref().local_addr(),
                 "connected outgoing to ourself - closing connection",
             );
-            return Effects::new();
+            return self.reconnect_if_not_connected_to_any_known_addresses(effect_builder);
```
goral09, line 585: Why does the node even connect to itself in the first place? Is that b/c it may be in the `known_addresses`? Wouldn't it be simpler if the node filtered out its own address from that list?

node/src/components/small_network.rs
```diff
@@ -638,29 +638,39 @@ where
...
+            warn!(
+                our_id=%self.our_id,
+                %peer_address,
+                "outgoing connection to known address closed"
+            );
         }
-
-        effects
+        // Since we are not calling `self.remove()`, call the reconnection check explicitly.
+        self.reconnect_if_not_connected_to_any_known_addresses(effect_builder)
```
goral09, line 673: Would it be possible to rewrite this whole function to get rid of early return from above? By the time I am reading _this_ comment I already lost the context and information form the earlier `if`s and `else`s and have no idea why we're not calling `self.remove()`, why we should have or what would it have meant.

node/src/components/small_network.rs
Fraser999: Not sure if it'd be simpler or not, but this is how the small_network has worked since inception I believe.  Perhaps that behaviour is due to change with the upcoming overhaul by @marc-casperlabs?

Either way, I'd be OK with changing the logic - just not in this PR :)

node/src/components/small_network.rs
Fraser999: I actually just changed it to have an early return, since that makes it simpler imo.  I'd rather not change it back.  I tend to prefer early returns as they generally make the code simpler to understand and require less nested levels.

The diff is pretty hard to read, but looking at [the actual function itself](https://github.com/CasperLabs/casper-node/blob/da770ea81b8ac121bb9db9fb50789d7cb552d4e7/node/src/components/small_network.rs#L632-L674) should hopefully answer your concerns.  It's a fairly small, succinct method really.

EdHastingsCasperLabs (approved):

Fraser999:
bors merge

PR #1272, opened 2021-04-05
NDRS-1076: on hard reset of storage, remove all data associated with a removed block

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1076.

In a hard reset, blocks created in or after the specified era are deleted from storage, but associated data is left behind.  Associated data includes block bodies, block signatures and crucially deploy execution results associated with the deleted blocks.

This PR resolves that issue by removing the above mentioned associated data on a hard reset.

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1266,6 +1266,7 @@ fn initialize_deploy_metadata_db(
             }
         }
 
+        // If the deploy's execution results are now empty, we just remove them entirely.
         if deploy_metadata.execution_results.is_empty() {
             cursor.del(WriteFlags::empty())?;
         } else if removed {
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -43,7 +43,7 @@ mod tests;
 #[cfg(test)]
 use std::{collections::BTreeSet, convert::TryFrom};
 use std::{
-    collections::{btree_map::Entry, BTreeMap},
+    collections::{btree_map::Entry, BTreeMap, HashSet},
     fmt::{self, Display, Formatter},
     fs, io, mem,
     path::PathBuf,
@@ -289,6 +289,7 @@ impl Storage {
         let mut block_txn = env.begin_rw_txn()?;
         let mut cursor = block_txn.open_rw_cursor(block_header_db)?;
 
+        let mut deleted_block_hashes = HashSet::new();
         // Note: `iter_start` has an undocumented panic if called on an empty database. We rely on
         //       the iterator being at the start when created.
         for (raw_key, raw_val) in cursor.iter() {
@@ -298,6 +299,7 @@ impl Storage {
                 // versions - they were most likely created before the upgrade and should be
                 // reverted.
                 if block.era_id() >= invalid_era && block.protocol_version() < protocol_version {
+                    let _ = deleted_block_hashes.insert(block.hash());
                     cursor.del(WriteFlags::empty())?;
                     continue;
                 }
@@ -318,11 +320,11 @@ impl Storage {
         drop(cursor);
         block_txn.commit()?;
 
-        // Check the integrity of the block body database.
-        check_block_body_db(&env, &block_body_db)?;
+        let deleted_block_hashes_raw = deleted_block_hashes.iter().map(BlockHash::as_ref).collect();
 
-        // Check the integrity of the block metadata database.
-        check_block_metadata_db(&env, &block_metadata_db)?;
+        initialize_block_body_db(&env, &block_body_db, &deleted_block_hashes_raw)?;
+        initialize_block_metadata_db(&env, &block_metadata_db, &deleted_block_hashes_raw)?;
+        initialize_deploy_metadata_db(&env, &deploy_metadata_db, &deleted_block_hashes)?;
 
         Ok(Storage {
             root,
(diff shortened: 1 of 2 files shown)


node/src/components/storage.rs
```diff
@@ -318,11 +320,11 @@ impl Storage {
...
 
-        // Check the integrity of the block body database.
-        check_block_body_db(&env, &block_body_db)?;
+        let deleted_block_hashes_raw = deleted_block_hashes.iter().map(BlockHash::as_ref).collect();
 
-        // Check the integrity of the block metadata database.
-        check_block_metadata_db(&env, &block_metadata_db)?;
+        initialize_block_body_db(&env, &block_body_db, &deleted_block_hashes_raw)?;
+        initialize_block_metadata_db(&env, &block_metadata_db, &deleted_block_hashes_raw)?;
+        initialize_deploy_metadata_db(&env, &deploy_metadata_db, &deleted_block_hashes)?;
```
goral09, line 327: I would separate the removal of old entries from the integrity check.

node/src/components/storage.rs
```diff
@@ -1212,6 +1234,49 @@ fn check_block_metadata_db(
...
+    info!("block metadata database initialized");
+    Ok(())
+}
+
+/// Purges stale entries from the deploy metadata database.
+fn initialize_deploy_metadata_db(
+    env: &Environment,
+    deploy_metadata_db: &Database,
+    deleted_block_hashes: &HashSet<BlockHash>,
+) -> Result<(), LmdbExtError> {
```
goral09, line 1250: We should probably also check the integrity of the deploy data. Can you ticket that?

node/src/components/storage.rs
```diff
@@ -1212,6 +1234,49 @@ fn check_block_metadata_db(
...
+
+    for (raw_key, raw_val) in cursor.iter() {
+        let mut deploy_metadata: DeployMetadata = lmdb_ext::deserialize(raw_val)?;
+
+        let mut removed = false;
+        for block_hash in deleted_block_hashes {
+            if deploy_metadata
+                .execution_results
+                .remove(block_hash)
+                .is_some()
```
goral09, line 1263: Wouldn't it be better, in terms of performance, to check it the other way around? Instead of checking every block hash we removed for existence in the `deploy_metadata.execution_results` (that test should be `true` only for one block hash in that collection, so we're making a lot of wasteful iterations), we check if `delete_block_hashes` exists in `delete_block_hashes`.

node/src/components/storage.rs
```diff
@@ -1212,6 +1234,49 @@ fn check_block_metadata_db(
...
+            {
+                removed = true;
+            }
+        }
+
+        if deploy_metadata.execution_results.is_empty() {
+            cursor.del(WriteFlags::empty())?;
+        } else if removed {
+            let buffer = lmdb_ext::serialize(&deploy_metadata)?;
+            cursor.put(&raw_key, &buffer, WriteFlags::empty())?;
```
goral09, line 1273: Can you add a comment that we remove the whole entry under the cursor, from the DB, if there are no more blocks it is contained in?

darthsiroftardis (approved):
LGTM!

node/src/components/storage/tests.rs
```diff
@@ -929,30 +982,90 @@ fn should_hard_reset() {
...
+        for (index, deploy) in deploys.iter().enumerate() {
+            let (_deploy, metadata) =
+                get_deploy_and_metadata(&mut harness, &mut storage, *deploy.id()).unwrap();
+            let should_have_exec_results = index < blocks_per_era * reset_era;
+            assert_eq!(
+                should_have_exec_results,
+                !metadata.execution_results.is_empty()
+            );
+        }
+    };
```
goral09, line 1063: All of the 3 checks above follow the same pattern, they could be extracted to a method.

node/src/components/storage/tests.rs
```diff
@@ -929,30 +982,90 @@ fn should_hard_reset() {
...
+
+    // Test with a hard reset to era 2, deleting blocks 6 and 7.
```
goral09, line 1065:
```suggestion
    // Test with a hard reset to era 2, deleting blocks (and associated data) 6 and 7.
```

goral09 (approved):
No blockers. Thanks for the tests.

node/src/components/storage.rs
Fraser999: I'd rather avoid that since we'd have to iterate each DB twice then.

goral09: Oh right, makes sense :+1:

node/src/components/storage.rs
Fraser999: I think we'd need to discuss that one first, just to agree the scope.  Do we just check the signature(s), or do we try and validate more than that?

goral09: Sure, we can discuss the scope during the sprint planning but the blank ticket would serve as a reminder.

node/src/components/storage.rs
Fraser999: I think you're right in that we could assume `execution_results` will only ever have 0 or 1 entry, but I'd rather leave this as is for now just to be safe.  I'm hoping we soon get to fix my initial mistake in making the `execution_resuilts` a `HashMap`, at which time, this code will automatically become optimal.

goral09: It won't become optimal anyway. Here, for every element (deploy) under the cursor (`m` elements) we iterate over all delete blocks again (`n` elements). So we're making `m x n` iterations. We need to change the second check to
```rust
for b in deploy_metadata.execution_results {
  if deleted_block_hashes.contains(b) {
    deploy_metadata.execution_results.remove(b);
  }
}
```
then it makes far fewer iterations and when we change the structure of `execution_results` to only ever contain single block, then that code becomes optimal.

node/src/components/storage.rs
Fraser999: Fair point - I'll make a ticket.

node/src/components/storage.rs
Fraser999: So, leaving the future changes out of the discussion, I think for now we need to keep `m x n` iterations to be safe.

goral09: Not really, my variant would still be vastly better b/c it would iterate `m` times (for every deploy) and then only as many times as there are elements in the `deploy_metadata.execution_results` (which should be max 1 anyway).

node/src/components/storage/tests.rs
Fraser999: I don't see a way to make that more succinct than what's here already.  Each check is just 3 lines, and the only thing common to all 3 is the variable set to `index < blocks_per_era * reset_era`.

goral09: They are all basically the same code with the exception for the `get_*` method the second element in `assert_eq!` call.

goral09: If you don't think this is worth changing then I'm fine with that. It's not a blocker.

Fraser999:
bors merge

PR #1276, opened 2021-04-07
NDRS-1121: Deduplicate block validity criteria.

afck:
This adds the `AppendableBlock` type, which checks block validity when adding deploys and transfers, and uses that type both for proposing and validating blocks, to make sure the same criteria are applied in both cases.

https://casperlabs.atlassian.net/browse/NDRS-1121

### DIFF
--- a/node/src/components/block_proposer.rs
+++ b/node/src/components/block_proposer.rs
@@ -19,7 +19,7 @@ use std::{
 use datasize::DataSize;
 use itertools::Itertools;
 use prometheus::{self, Registry};
-use tracing::{debug, error, info, trace, warn};
+use tracing::{debug, error, info, trace};
 
 use crate::{
     components::Component,
@@ -27,14 +27,16 @@ use crate::{
         requests::{BlockProposerRequest, ProtoBlockRequest, StateStoreRequest, StorageRequest},
         EffectBuilder, EffectExt, Effects,
     },
-    types::{chainspec::DeployConfig, Chainspec, DeployHash, DeployHeader, ProtoBlock, Timestamp},
+    types::{
+        appendable_block::{AddError, AppendableBlock},
+        chainspec::DeployConfig,
+        Chainspec, DeployHash, DeployHeader, ProtoBlock, Timestamp,
+    },
     NodeRng,
 };
-use casper_execution_engine::shared::gas::Gas;
 pub(crate) use deploy_sets::BlockProposerDeploySets;
 pub(crate) use event::{DeployType, Event};
 use metrics::BlockProposerMetrics;
-use num_traits::Zero;
 
 /// Block proposer component.
 #[derive(DataSize, Debug)]
@@ -385,20 +387,11 @@ impl BlockProposerReady {
     }
 
     /// Checks if a deploy is valid (for inclusion into the next block).
-    fn is_deploy_valid(
-        &self,
-        header: &DeployHeader,
-        block_timestamp: Timestamp,
-        deploy_config: &DeployConfig,
-        past_deploys: &HashSet<DeployHash>,
-    ) -> bool {
-        let all_deps_resolved = || {
-            header
-                .dependencies()
-                .iter()
-                .all(|dep| past_deploys.contains(dep) || self.contains_finalized(dep))
-        };
-        header.is_valid(deploy_config, block_timestamp) && all_deps_resolved()
+    fn deps_resolved(&self, header: &DeployHeader, past_deploys: &HashSet<DeployHash>) -> bool {
+        header
+            .dependencies()
+            .iter()
+            .all(|dep| past_deploys.contains(dep) || self.contains_finalized(dep))
     }
 
     /// Returns a list of candidates for inclusion into a block.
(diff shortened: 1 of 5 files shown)


node/src/types/appendable_block.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub(crate) fn add(
+        &mut self,
+        hash: DeployHash,
+        deploy_type: &DeployType,
+    ) -> Result<(), AddError> {
+        if self.deploy_and_transfer_set.contains(&hash) {
+            return Err(AddError::Duplicate);
+        }
+        if !deploy_type
+            .header()
```
goral09, line 74: It's a weird thing to see – a header on deploy _type_. IMO a _type_ should be either `Wasm` or `NativeTransfer`. Maybe we need a better name?

node/src/types/appendable_block.rs
```diff
@@ -0,0 +1,121 @@
...
+            .is_valid(&self.deploy_config, self.timestamp)
+        {
+            return Err(AddError::InvalidDeploy);
+        }
+        if deploy_type.is_transfer() {
+            if self.transfer_hashes.len() == self.deploy_config.block_max_transfer_count as usize {
```
goral09, line 80: Can you extract
```rust
self.transfer_hashes.len() == self.deploy_config.block_max_transfer_count as usize
```
to a method?

node/src/types/appendable_block.rs
```diff
@@ -0,0 +1,121 @@
...
+                return Err(AddError::TransferCount);
+            }
+            self.transfer_hashes.push(hash);
+        } else {
+            if self.deploy_hashes.len() == self.deploy_config.block_max_deploy_count as usize {
```
goral09, line 85: Same as for the transfer check above.

node/src/types/appendable_block.rs
```diff
@@ -0,0 +1,121 @@
...
+                return Err(AddError::DeployCount);
+            }
+            // Only deploys count towards the size and gas limits.
+            if self.total_size
+                > (self.deploy_config.max_block_size as usize).saturating_sub(deploy_type.size())
```
goral09, line 90: I don't understand why you prefer it this way :sweat_smile: To me it's just much simpler to see the deploy's size added to the `self.total_size`.

node/src/components/block_proposer/tests.rs
```diff
@@ -354,16 +356,16 @@ fn should_keep_track_of_unhandled_deploys() {
...
-            deploy2.header(),
-            test_time,
-            &proposer.deploy_config,
-            &past_deploys
-        ),
-        "deploy2 should -not- be valid"
+        deploy2
+            .header()
+            .is_valid(&proposer.deploy_config, test_time),
+        "deploy2 should be valid"
```
goral09, line 368: Why did this assertion change? It also does not expect the `past_deploys` anymore, is that test equivalent now?

node/src/components/block_validator.rs
```diff
@@ -68,14 +67,14 @@ pub enum Event<T, I> {
...
-        deploy_size: usize,
+        deploy_type: Box<DeployType>,
     },
 
     /// A request to find a specific deploy, potentially from a peer, failed.
     #[display(fmt = "deploy {} missing", _0)]
     DeployMissing(DeployHash),
 
-    /// Deploy was invalid. Failed the chainspec test.
+    /// Deploy was invalid. Unable to convert to a deploy type.
```
goral09, line 77: Is `DeployInvalid` now meaning _only_ inability to convert to a deploy type? If so, then we should change the name of this enum variant.

node/src/components/block_validator.rs
```diff
@@ -403,20 +378,14 @@ where
...
 {
     let validate_deploy = move |result: FetchResult<Deploy, I>| match result {
-        FetchResult::FromStorage(deploy) | FetchResult::FromPeer(deploy, _) => {
-            if deploy
-                .header()
-                .is_valid(&chainspec.deploy_config, block_timestamp)
-            {
-                let deploy_size = deploy.serialized_length();
+        FetchResult::FromStorage(deploy) | FetchResult::FromPeer(deploy, _) => deploy
+            .deploy_type()
```
goral09, line 382: Why would that even fail?

node/src/types/appendable_block.rs
afck: I agree, but this wasn't added in this PR, so I'd like to leave it for now.

node/src/components/block_proposer.rs
```diff
@@ -409,73 +402,68 @@ impl BlockProposerReady {
...
-            let at_max_deploys = deploy_hashes.len() == max_deploys
-                || (deploy_type.is_wasm()
-                    && block_size_running_total + DEPLOY_APPROX_MIN_SIZE >= max_block_size_bytes);
-
-            if at_max_deploys && at_max_transfers {
-                break;
+        let mut appendable_block = AppendableBlock::new(deploy_config, block_timestamp);
+
+        // We prioritize transfers over deploys, so we try to include them first.
+        for (hash, deploy_type) in &self.sets.pending {
```
goral09, line 408: I think it's a good idea to separate Wasm from native transfers in the pending sets.

node/src/types/appendable_block.rs
afck: Right… trying to cover all the overflow/underflow and `usize::MAX` cases. But maybe I can use a `checked_add`. :+1:

node/src/components/block_proposer/tests.rs
afck: It is, and the "not" must have been a typo, because the assertion was that it _was_ valid.
Anyway, we now check that the header `is_valid` and above that `proposer.contains_finalized`, which I believe was the original intention. Maybe @dwerner can confirm?

goral09 (commented):
Looks good to me so far.

node/src/components/block_validator.rs
afck: Yes, although failing to convert still means that it's invalid. Maybe `DeployConversionFailed` or `CouldNotCreateDeployType`?

node/src/components/block_validator.rs
afck: These are the relevant lines from `Deploy::deploy_type`:
```rust
                let value = payment_item
                    .args()
                    .get(ARG_AMOUNT)
                    .ok_or(Error::InvalidPayment)?;
                let value = value
                    .clone()
                    .into_t::<U512>()
                    .map_err(|_| Error::InvalidPayment)?;
```

node/src/components/block_proposer.rs
afck: I agree, but that may also be a separate task, for after the next release.

node/src/components/block_proposer/tests.rs
dwerner: The original intent of this test was to ensure that the block proposer would correctly handle the case where:

- We received a finalized deploy hash (through `.finalized_deploys`)
- We have NOT yet received the DeployType

This scenario can happen since messages from the reactor can arrive out of order.

From reading the above test case, it would appear the original intent is preserved, minus my typo.

node/src/types/appendable_block.rs
afck: Done in https://github.com/CasperLabs/casper-node/pull/1276/commits/0994e702c9eba2795f6cd3d669c9398b925b213c.

node/src/types/appendable_block.rs
afck: Doen in https://github.com/CasperLabs/casper-node/pull/1276/commits/0994e702c9eba2795f6cd3d669c9398b925b213c.

node/src/types/appendable_block.rs
afck: Done in https://github.com/CasperLabs/casper-node/pull/1276/commits/0994e702c9eba2795f6cd3d669c9398b925b213c.

goral09 (approved):

xcthulhu (approved):

PR #1277, opened 2021-04-07
NDRS-1117: Backport handshake

marc-casperlabs:
This backports two 1.1.0 changes to `dev`, with the necessary alterations: https://github.com/CasperLabs/casper-node/pull/1256 (making the maximum message size depend on the chainspec instead of being hardcoded) https://github.com/CasperLabs/casper-node/pull/1263 (include the chainspec version in the handshake)

It also introduces the first set of rountrip tests that are ensuring we can serialize/deserialize network messages from 1.0.0 and onwards.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -32,6 +32,7 @@
 //! No explicit reconnect is attempted. Instead, if the peer is still online, the normal gossiping
 //! process will cause both peers to connect again.
 
+mod chain_info;
 mod config;
 mod error;
 mod event;
@@ -99,6 +100,7 @@ use crate::{
     types::{NodeId, TimeDiff, Timestamp},
     utils, NodeRng,
 };
+use chain_info::ChainInfo;
 pub use config::Config;
 pub use error::Error;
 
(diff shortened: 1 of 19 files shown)


node/src/components/small_network.rs
```diff
@@ -502,11 +507,9 @@ where
 
                 info!(our_id=%self.our_id, %peer_id, %peer_address, "established incoming connection");
                 // The sink is only used to send a single handshake message, then dropped.
-                let (mut sink, stream) = framed::<P>(transport).split();
-                let handshake = Message::Handshake {
-                    network_name: self.network_name.clone(),
-                    public_address: self.public_address,
-                };
+                let (mut sink, stream) =
+                    framed::<P>(transport, self.chain_info.maximum_net_message_size).split();
```
dwerner, line 511: IMO it's worth mentioning this in the PR description - that we're allowing configuration of the max framed size for messages.

dwerner (approved):
lgtm, but `maximum_net_message_size` should be mentioned in the description (admittedly it's a backport)

node/src/components/small_network.rs
marc-casperlabs: Hmm, technically it's not configuration, but a chain spec value? It's mentioned in the linked PR https://github.com/CasperLabs/casper-node/pull/1256

marc-casperlabs: I'll add it though.

PR #1281, opened 2021-04-08
NO-TICKET: Log the event type if it is too large.

afck:
I was wondering about the "large event size" warnings in the tests, so I added more logging. Turns out the event itself is test-only, but it's still better to log the event type:

```
WARN  [c:reactor :448] large event size, consider reducing it or boxing; event_size=328; type_name="casper_node::components::gossiper::tests::Event"
```


marc-casperlabs (approved):
Sounds like a good idea.

Sadly, we cannot log the biggest variant on enums, but this should help somewhat.

PR #1283, opened 2021-04-08
NDRS-1120: Add a delay before proposing dependencies.

afck:
This adds a configurable delay between the time we receive a deploy (from a client or another full node) and the time we propose it in a block. The intention is to give deploys time to spread across the network via gossip, so that when we include them in a new block, the other nodes won't all have to ask us for the deploy.

https://casperlabs.atlassian.net/browse/NDRS-1120


fizyk20 (approved):

goral09 (approved):

afck:
```
    Address already in use (os error 98)', node/src/components/small_network
```
Retrying…

PR #1286, opened 2021-04-08
EE-1223: Remove get refund purse

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1223

Removes `get_refund_purse` entry point altogether as we were returning an `URef` there without access rights which could pose a security risk.

### DIFF
--- a/execution_engine/src/core/engine_state/genesis.rs
+++ b/execution_engine/src/core/engine_state/genesis.rs
@@ -31,7 +31,7 @@ use casper_types::{
         },
         handle_payment::{
             self, ARG_ACCOUNT, METHOD_FINALIZE_PAYMENT, METHOD_GET_PAYMENT_PURSE,
-            METHOD_GET_REFUND_PURSE, METHOD_SET_REFUND_PURSE,
+            METHOD_SET_REFUND_PURSE,
         },
         mint::{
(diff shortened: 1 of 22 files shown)


darthsiroftardis (approved):
LGTM! I would wait for another greencheck

goral09:
What security risks? A purse without access rights can't be used for taking the money out of it. Even the comment says:
```
// We purposely choose to remove the access rights so that we do not
// accidentally give rights for a purse to some contract that is not
// supposed to have it.
```

henrytill (approved):
Merge away!

mpapierski:
> 
> 
> What security risks? A purse without access rights can't be used for taking the money out of it. Even the comment says:
> 
> ```
> // We purposely choose to remove the access rights so that we do not
> // accidentally give rights for a purse to some contract that is not
> // supposed to have it.
> ```

Correct - if there are no access rights, then there's no reason to expose this URef at all to contract developer

goral09:
> > What security risks? A purse without access rights can't be used for taking the money out of it. Even the comment says:
> > ```
> > // We purposely choose to remove the access rights so that we do not
> > // accidentally give rights for a purse to some contract that is not
> > // supposed to have it.
> > ```
> 
> Correct - if there are no access rights, then there's no reason to expose this URef at all to contract developer

Thanks. It can still be used in some in-smart-contract manipulation like equality check. It can be transferred to other smart contracts (also without access rights). It's just not useful when trying to read/write/add to it but still pretty useful to me.

PR #1290, opened 2021-04-09
NDRS-1061, CEP 40: Simplify block types and remove redundant fields.

afck:
This deduplicates the timestamp, height and parent information in the consensus proposal messages and replaces `ProtoBlock` and `CandidateBlock` with a single `BlockPayload` type.

That also means the `mem_proto_block_validator` metric gets renamed to `mem_block_payload_validator`.

[CEP 40](https://github.com/CasperLabs/ceps/blob/master/text/0040-block-proposal-message-cleanup.md)
https://casperlabs.atlassian.net/browse/NDRS-1061


node/src/types/block.rs
```diff
@@ -180,87 +187,41 @@ impl From<TryFromSliceError> for Error {
...
 /// The piece of information that will become the content of a future block (isn't finalized or
 /// executed yet)
 ///
 /// From the view of the consensus protocol this is the "consensus value": The protocol deals with
-/// finalizing an order of `ProtoBlock`s. Only after consensus has been reached, the block's
+/// finalizing an order of `BlockPayload`s. Only after consensus has been reached, the block's
 /// deploys actually get executed, and the executed block gets signed.
 ///
 /// The word "proto" does _not_ refer to "protocol" or "protobuf"! It is just a prefix to highlight
 /// that this comes before a block in the linear, executed, finalized blockchain is produced.
```
Fraser999, line 198:
```suggestion
```

Fraser999 (approved):
Only one nitpick from me, but some of the changes to the consensus functionality should be checked by someone more familiar with the details.  Otherwise LGTM.

node/src/types/block.rs
afck: Good catch! Fixed in https://github.com/casper-network/casper-node/pull/1290/commits/74aafadcaf6484ad8fcd2dcb7065349869471683.

node/src/components/consensus/consensus_protocol.rs
```diff
@@ -11,25 +11,42 @@ use crate::{
...
 /// Information about the context in which a new block is created.
 #[derive(Clone, DataSize, Eq, PartialEq, Debug, Ord, PartialOrd, Hash)]
-pub struct BlockContext {
+pub(crate) struct BlockContext<C>
+where
+    C: Context,
+{
     timestamp: Timestamp,
-    height: u64,
+    ancestor_values: Vec<C::ConsensusValue>,
```
goral09, line 19: I'm reviewing the PR commit-by-commit, it would be nice to have the explanation (in the description of the commit) why we're adding the ancestor values.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1059,38 +1062,38 @@ where
...
+                    .iter()
+                    .flat_map(|ancestor| ancestor.deploys_and_transfers_iter())
+                    .any(|deploy| block_deploys_set.contains(deploy))
+                {
+                    return self.resolve_validity(ResolveValidity {
+                        era_id,
+                        sender,
+                        proposed_block,
+                        valid: false,
+                    });
```
goral09, line 1096: Can we extract this to a method (or at least add a comment) saying this is a replay attack validation?

node/src/components/consensus/era_supervisor/era.rs
```diff
@@ -85,78 +80,62 @@ impl<I> Era<I> {
...
+    pub(crate) fn resolve_evidence(
+        &mut self,
+        pub_key: &PublicKey,
+    ) -> Vec<ProposedBlock<ClContext>> {
+        for pc in self.validation_states.values_mut() {
             pc.missing_evidence.retain(|pk| pk != pub_key);
         }
         self.consensus.mark_faulty(pub_key);
-        self.remove_complete_candidates()
+        let (complete, validation_states): (HashMap<_, _>, HashMap<_, _>) = self
```
goral09, line 111:
```suggestion
        let (complete, incomplete): (HashMap<_, _>, HashMap<_, _>) = self
```
Can we change the names to be somewhat clearer what it is?

goral09 (approved):
No blockers.

node/src/components/consensus/consensus_protocol.rs
afck: That would mean force-pushing, though. :grimacing: 

Concretely, I added `ancestor_values` to the block context because they make sense there: They are part of the context information that you need, both to propose a new block payload, and to validate an incoming block payload. It also allows us to pass fewer arguments in some places, where we had both context (or timestamp) and ancestor values.

goral09: Oh, definitely not force-push. Just a general comment for the future.

node/src/components/consensus/era_supervisor/era.rs
afck: Done in https://github.com/casper-network/casper-node/pull/1290/commits/794ed59f41d5773bdfe4357c8ae592e4f9670aaa.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/1290/commits/794ed59f41d5773bdfe4357c8ae592e4f9670aaa.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1329,3 +1319,17 @@ where
...
+    /// If this block contains a deploy that's also present in an ancestor, this returns the deploy
+    /// hash, otherwise `None`.
+    fn contains_replay(&self) -> Option<&DeployHash> {
+        let block_deploys_set: BTreeSet<DeployHash> =
+            self.value().deploys_and_transfers_iter().cloned().collect();
+        self.context()
+            .ancestor_values()
+            .iter()
+            .flat_map(|ancestor| ancestor.deploys_and_transfers_iter())
+            .find(|deploy| block_deploys_set.contains(deploy))
```
goral09, line 1333: Only one thing I wonder – is this guaranteed to be the same `DeployHash` on every node that executes this method? Or is the order non-deterministic? I was imagining validators debugging this and _maybe_ comparing their logs.

afck: The order of iteration should be deterministic. In the block payload it's even `Vec`s, and the ancestors are sorted in reverse chronological order. So every validator will see the same message.

fizyk20 (approved):

sacherjj:
bors merge

sacherjj:
bors merge

sacherjj:
bors merge

sacherjj:
bors try

TomVasile:
bors try

sacherjj:
bors ping

PR #1292, opened 2021-04-09
NDRS-1116: display full cause chain when logging errors

marc-casperlabs:
This PR introduces the `display_error` function, which causes errors to be properly **lazily** formatted when logging errors.

Currently, in our codebase, we often log errors like this:

```rust
match result {
    Error(err) {
        warn!(%err, "something happened');
    }
    // ...
}
```

This has the drawback of swallowing the errors causes, which are often more interesting than the potentially wrapping error itself.

This can now be addressed by using the `display_error` function:

```rust
match result {
    Error(ref err) {
        warn!(err=display_error(err), "something happened');
    }
    // ...
}
```

Which will cause the error to be output as "error: cause1: cause 2" instead.

Ideally this would be handled using `tracing`'s built-in functionality, but the implementation is lagging behind/buggy; the `display_error` functions documentation has details on how this should be replaced once `tracing` upstream has addressed the issue.

The whole codebase will need to be ultimately updated, this PR addresses `small_network` as a first proving grounds for the change.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -86,7 +86,7 @@ use tracing::{debug, error, info, trace, warn};
 
 use self::error::Result;
 pub(crate) use self::{
-    error::cl_error, event::Event, gossiped_address::GossipedAddress, message::Message,
+    error::display_error, event::Event, gossiped_address::GossipedAddress, message::Message,
 };
 use crate::{
     components::{
@@ -214,7 +214,7 @@ where
                     };
                 }
                 Err(ref err) => {
-                    warn!(%address, err=cl_error(err), "failed to resolve known address");
+                    warn!(%address, err=display_error(err), "failed to resolve known address");
                 }
             }
         }
@@ -501,7 +501,7 @@ where
                         our_id=%self.our_id,
                         %peer_address,
                         local_address=?transport.get_ref().local_addr(),
-                        err=cl_error(err),
+                        err=display_error(err),
                         "incoming connection dropped",
                     );
                     return Effects::new();
@@ -549,7 +549,7 @@ where
                 effects
             }
             Err(ref err) => {
-                warn!(our_id=%self.our_id, %peer_address, err=cl_error(err), "TLS handshake failed");
+                warn!(our_id=%self.our_id, %peer_address, err=display_error(err), "TLS handshake failed");
                 Effects::new()
             }
         }
@@ -568,7 +568,7 @@ where
             Err(ref err) => {
                 // The peer address disappeared, likely because the connection was closed while
                 // we are setting up.
-                warn!(%peer_id, err=cl_error(err), "peer connection terminated while setting up outgoing connection, dropping");
+                warn!(%peer_id, err=display_error(err), "peer connection terminated while setting up outgoing connection, dropping");
 
                 // We still need to clean up any trace of the connection.
                 return self.remove(effect_builder, &peer_id, false);
@@ -648,7 +648,7 @@ where
                     our_id=%self.our_id,
                     %peer_id,
                     %peer_address,
-                    err=cl_error(err),
+                    err=display_error(err),
                     "outgoing connection failed"
                 );
             } else {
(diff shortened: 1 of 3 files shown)


Fraser999 (approved):
I'm fine to approve this PR, but I would absolutely limit `e!`'s usage to small_network for now until we see how things shake out with tracing.  I think displaying the full cause stack for every logged error could well be a waste of dev time and effort, and actually result in a more painful debugging experience compared to just using the `Display` impl of the given error.

I'm also not keen on any single-character names, but unless others object, given the limited usage of this macro, I can live with it :)

marc-casperlabs:
> I'm also not keen on any single-character names, but unless others object, given the limited usage of this macro, I can live with it :)

Your sacrifice is appreciated. :)

bors r+

goral09:
I don't like the single-character macro either. It looks really unnatural.

marc-casperlabs:
> I don't like the single-character macro either. It looks really unnatural.

It was chosen because as `tracing` catches up,

```rust
warn!(err = e!(err), "...");
```

should eventually become

```rust
warn!(err = !err, "...");

// possibly:
warn!(!err, "...")
```

goral09:
Why not `err!` or `error!`?

node/src/components/small_network.rs
```diff
@@ -90,6 +90,7 @@ use crate::{
     components::{
         network::ENABLE_LIBP2P_NET_ENV_VAR, networking_metrics::NetworkingMetrics, Component,
     },
+    e,
```
dwerner, line 93: Maybe make it a local macro if it's only going to be used in small_network, as per @Fraser999 's suggestion, which I agree with.

marc-casperlabs: This usually gets us in trouble because doctests and tests need it available externally, IIRC.

xcthulhu (changes requested):
We should not be making single character functions or macros.

marc-casperlabs:
> Why not err! or error!?

This would clash too much with the tracing `error!` macro, in my opinion.

xcthulhu:
@marc-casperlabs Did you see some slowdown that made this necessary?

xcthulhu:
bors cancel

marc-casperlabs:
> Did you see some slowdown that made this necessary?

It's to retain the flexibility to possibly upgrade to `tracing`'s own handling of errors, instead of using the `Display<T>` wrapper, once https://github.com/tokio-rs/tracing/issues/1347 is fixed, without having to rewrite the code that uses it.

I "staged" this in this repo and verified in a simple throwaway app, to ensure that it can be upgraded.

Fraser999:
Can I suggest we rename to `cl_error`, and try limiting its scope to small_network?  If we need it for doctests or integration tests, we can expose it further.

node/src/components/small_network.rs
marc-casperlabs: So it can't be entirely local (no `#[macro_export]`) since I do not want the `impl` in the root `small_net` module.

marc-casperlabs:
> Can I suggest we rename to cl_error, and try limiting its scope to small_network? If we need it for doctests or integration tests, we can expose it further.

It's now been moved and limited as much as possible.

@xcthulhu @goral09  It's also been renamed as `cl_error!`

goral09 (approved):

xcthulhu:
Couldn't we just wait until tokio-rs/tracing#1347 lands before doing this?

marc-casperlabs:
> Couldn't we just wait until tokio-rs/tracing#1347 lands before doing this?

Unfortunately as far as I know we have no idea when that might be :(

Fraser999:
> Couldn't we just wait until tokio-rs/tracing#1347 lands before doing this?

I'd be keen to see this merged ASAP - we need more info on network log messages right now really.

xcthulhu (approved):
LGTM

xcthulhu:
> I'd be keen to see this merged ASAP - we need more info on network log messages right now really.

Good point.  Approved.

PR #1293, opened 2021-04-09
EE-1225: Payment purse invariants

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1225

This PR contains a regression test that tries to set a refund purse into a payment purse to violate `finalize_payment` postconditions. Fix consists of a check inside finalize_payment (rather than in set_refund_purse) to ensure we will not refund into the payment purse.


types/src/system/handle_payment/mod.rs
```diff
@@ -139,6 +151,10 @@ mod internal {
             return Ok(());
         }
 
+        // if let Some(refund_purse) = refund_purse {
+        //     assert_ne!(refund_purse.addr(), payment_purse.a);
+        // }
+
```
henrytill, line 157: Did you mean to commit this?

henrytill (changes requested):
Holding until further notice

dwerner (approved):

execution_engine/CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Changed
 * Update pinned version of Rust to `nightly-2021-06-17`
-
+* Fixed a case where user could potentially supply a refund purse as a payment purse.
```
```suggestion

### Fixed
* Fix a case where user could potentially supply a refund purse as a payment purse.

```

mpapierski:
I'll re-run tests, and update the description with the error that this issue could cause

smart_contracts/contracts/test/ee-1225-regression/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[[bin]]
+name = "ee_1225_regression"
+path = "src/main.rs"
+bench = false
+doctest = false
+test = false
+
+[features]
+std = ["casper-contract/std", "casper-types/std"]
```
Fraser999:
```suggestion
```

Fraser999 (approved):

Fraser999 (approved):

mpapierski:
After removing the fix the regression test executes fine which means the code silently accepts refund purse as a payment purse which could lead to unexpected results once we'd change the refund percentage.

PR #1294, opened 2021-04-09
NDRS-1086: change SSE broadcast channel size to be a factor of event stream buffer size

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1086.

This PR removes `broadcast_channel_size` from the SSE config, instead making the value a factor (120%) of the `event_stream_buffer_length`.

We always want the broadcast channel size to be greater than the event stream buffer length so that a new client can retrieve the entire set of buffered events if desired.

@sacherjj @momipsl note that this PR changes the config.toml.

### DIFF
--- a/node/src/components/event_stream_server.rs
+++ b/node/src/components/event_stream_server.rs
@@ -46,6 +46,16 @@ pub use config::Config;
 pub(crate) use event::Event;
 pub use sse_server::SseData;
 
+/// This is used to define the number of events to buffer in the tokio broadcast channel to help
+/// slower clients to try to avoid missing events (See
+/// https://docs.rs/tokio/1.4.0/tokio/sync/broadcast/index.html#lagging for further details).  The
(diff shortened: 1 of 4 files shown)


node/src/components/event_stream_server.rs
```diff
@@ -77,8 +87,10 @@ impl EventStreamServer {
         let (sse_data_sender, sse_data_receiver) = mpsc::unbounded_channel();
 
         // Event stream channels and filter.
+        let broadcast_channel_size = config.event_stream_buffer_length / 100
```
goral09, line 90: Is that value in bytes? If so, can we add the suffix `_bytes` to it?

goral09 (approved):

node/src/components/event_stream_server.rs
Fraser999: No, they both measure event counts.

dwerner (approved):

Fraser999:
bors merge

Fraser999:
bors merge

PR #1295, opened 2021-04-10
NDRS-1133 - Unifying fast-sync and linear-chain-sync

xcthulhu:


goral09:
Like I said in the comment under the other PR – I hope the "normal" `LinearChain` will not be replaced with the fast-sync version. We still need the original implementation for archival nodes.

dwerner (approved):
lgtm

PR #1299, opened 2021-04-12
NDRS-1070: Signing Should Not Panic

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-1070

CHANGELOG:

- Changed the return of the `sign` function to return `Result<Signature>` instead of simply `Signature`
- Refactored code to account for the above change


node/src/components/consensus/cl_context.rs
```diff
@@ -43,7 +43,7 @@ impl ValidatorSecret for Keypair {
     type Signature = Signature;
 
     fn sign(&self, hash: &Digest) -> Signature {
-        crypto::sign(hash, self.secret_key.as_ref(), &self.public_key)
+        crypto::sign(hash, self.secret_key.as_ref(), &self.public_key).unwrap()
```
xcthulhu, line 46: Sadly, we can't panic here either.

These errors need to be surfaced all the way up to the `handle` method on our components, and we have to emit `fatal!(efffect_builder,...)` effects.

xcthulhu (changes requested):
Sadly, we have more work to handle these errors properly.

Maybe this could be 1 of 2, if we come back and work to surface these errors as `fatal!` effects?

xcthulhu:
Since this isn't critical, let's wait until part 2 is completed until merging.

PR #1300, opened 2021-04-12
EE-1229: Validator slots clarification

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1229

Validate validator slots & validators.
    
This commit ensures that the `validator_slot` config value at genesis should
be lower than or equal to a number of genesis validators.
    
In case of validator slots value is smaller than genesis validators the
the result would be 0 non-genesis validators and N genesis validators
regardless of the configured limit.

CC @henrytill


darthsiroftardis (approved):
LGTM!

types/src/system/auction/mod.rs
```diff
@@ -399,6 +399,11 @@ pub trait Auction:
 
             non_founder_weights.sort_by(|(_, lhs), (_, rhs)| rhs.cmp(lhs));
 
+            // This assumes that amount of founding validators does not exceed configured validator
+            // slots. For a case where there is exactly N validators and the limit is N, only
```
afck, line 403:
```suggestion
            // slots. For a case where there are exactly N validators and the limit is N, only
```

types/src/system/auction/mod.rs
```diff
@@ -399,6 +399,11 @@ pub trait Auction:
...
+            // founding validators will be the in the winning set. It is advised to set
+            // `validator_slots` larger than amount of founding validators in accounts.toml to
+            // accomodate non-genesis validators.
```
afck, line 406: That part doesn't apply anymore, does it? We now make sure it's never larger.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -2959,3 +2959,51 @@ fn should_validate_genesis_delegators_bond_amount() {
 
     builder.run_genesis(&run_genesis_request);
 }
+
+fn check_validator_slots_for_accounts(accounts: usize) {
+    let accounts = {
+        let range = 1..=accounts;
+        let (capacity, _) = range.size_hint();
```
afck, line 2966: Isn't `capacity == accounts` now? Would it be easier to just use `accounts`?
Alternatively, don't make a mutable vector at all:
```rust
let accounts: Vec<_> = (1..=accounts).map(|i| {
    let count = U256::from(i);
    let mut secret_key_bytes = [0; 32];
    …
}).collect();
```

afck (approved):
Looks good; just some nit-picks.

types/src/system/auction/mod.rs
mpapierski: Yes - now validator slots should be configured to be greater or equal to the number of genesis validator accounts. This comment is for the reader to bring the config validation process.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
mpapierski: Oh, right, that `size_hint` was a bit of an overkill 🤦 I just used `accounts` value. Thanks!

henrytill (changes requested):
Holding until further notice

Fraser999 (approved):

PR #1303, opened 2021-04-12
Master into release-1.1.0

sacherjj:


client/lib/rpc.rs
```diff
@@ -201,7 +201,7 @@ impl RpcCall {
         let session = ExecutableDeployItem::Transfer {
             args: transfer_args,
         };
-        let deploy = Deploy::with_payment_and_session(deploy_params, payment, session);
+        let deploy = Deploy::with_payment_and_session(deploy_params, payment, session).unwrap();
```
xcthulhu, line 204: We should not be unwrapping here...

goral09:
@xcthulhu @sacherjj This is a "cleanup" PR, there should be no commits made directly to it, it's only moving changes from one branch to another. This is not how git flow works.

sacherjj:
You are correct @goral09 I reverted back and we should fix these other ways.

xcthulhu (approved):

PR #1304, opened 2021-04-13
EE-1220: Add Read check in Mint::transfer

darthsiroftardis:
CHANGELOG:

- Added a checked in `Mint::transfer` that checks the `Read` access for the source of the transfer and returns an `InvalidAccessRights` error should the check fail


mpapierski (approved):
LGTM

PR #1305, opened 2021-04-13
HWY-275: Do not schedule same vertex for synchronization more than once.

goral09:


goral09:
For some reason `dev` was broken and clippy started failing for me on totally unrelated changes. Fixed in [`c7b0a0d` (#1305)](https://github.com/CasperLabs/casper-node/pull/1305/commits/c7b0a0d8957f8674f079df839a81e3c7f857b73e)

execution_engine_testing/tests/src/test/get_balance.rs
```diff
@@ -132,7 +132,7 @@ fn get_balance_using_public_key_should_work() {
 
     let alice_main_purse = alice_account.main_purse();
 
-    let alice_balance_result = builder.get_public_key_balance_result(*ALICE_KEY);
```
afck, line 135: Was this broken on `dev`?

afck (commented):
This seems to prevent later senders (i.e. alternative sources for a vertex) from being added to `vertices_to_be_added`.

In general: Currently, if Alice and Bob send us a vertex with a dependency, we'd request the dependency from both, so if Alice doesn't answer, we receive it from Bob.

With this change, if Alice sent it first, would we ignore Bob? Would be be stuck now if Alice doesn't answer, because we'd never request it from Bob?

execution_engine_testing/tests/src/test/get_balance.rs
goral09: Apparently, I had to merge with `dev` and resolve Clippy issues on my branch.

goral09:
> This seems to prevent later senders (i.e. alternative sources for a vertex) from being added to `vertices_to_be_added`.
> 
> In general: Currently, if Alice and Bob send us a vertex with a dependency, we'd request the dependency from both, so if Alice doesn't answer, we receive it from Bob.
> 
> With this change, if Alice sent it first, would we ignore Bob? Would be be stuck now if Alice doesn't answer, because we'd never request it from Bob?

Ah, you're right! I don't know what I was thinking.

PR #1311, opened 2021-04-14
OP-1838: itst06 scenario

TomVasile:
Scenario mimics a node that fails as transfers are coming in mid-stream. Verifies that transfers that are sent get included in blocks.

https://casperlabs.atlassian.net/browse/OP-1838

### DIFF
--- a/utils/nctl/sh/scenarios/common/itst.sh
+++ b/utils/nctl/sh/scenarios/common/itst.sh
@@ -156,6 +156,9 @@ function get_switch_block_equivocators() {
     fi
 }
 
+# Function is used to walk back the blocks to check if a transfer
+# is included under transfer_hashes. If the transfer is not found
+# within the walkback, it will error out.
 function verify_transfer_inclusion() {
     local NODE_ID=${1}
     # Number of blocks to walkback before erroring out
--- a/utils/nctl/sh/scenarios/common/itst.sh
+++ b/utils/nctl/sh/scenarios/common/itst.sh
@@ -156,6 +156,40 @@ function get_switch_block_equivocators() {
     fi
 }
 
+function verify_transfer_inclusion() {
+    local NODE_ID=${1}
+    # Number of blocks to walkback before erroring out
+    local WALKBACK=${2}
+    local TRANSFER=${3}
+    local BLOCK_HASH=${4}
+    local JSON_OUT
+    local PARENT
+    local BLOCK_HEADER
+    local BLOCK_TRANSFER_HASHES
+
+    if [ -z "$BLOCK_HASH" ]; then
+        JSON_OUT=$($(get_path_to_client) get-block --node-address $(get_node_address_rpc "$NODE_ID"))
+    else
+        JSON_OUT=$($(get_path_to_client) get-block --node-address $(get_node_address_rpc "$NODE_ID") -b "$BLOCK_HASH")
+    fi
+
+    if [ "$WALKBACK" -gt 0 ]; then
+        BLOCK_HEADER=$(echo "$JSON_OUT" | jq '.result.block.header')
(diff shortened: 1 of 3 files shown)


sacherjj (approved):

utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -156,6 +156,43 @@ function get_switch_block_equivocators() {
     fi
 }
 
+# Function is used to walk back the blocks to check if a transfer
+# is included under transfer_hashes. If the transfer is not found
+# within the walkback, it will error out.
+function verify_transfer_inclusion() {
```
goral09, line 162: I think we should do similar test for deploys (not transfers).

utils/nctl/sh/scenarios/itst06.sh
```diff
@@ -0,0 +1,112 @@
...
+        node="$NODE_ID") > "$DEPLOY_LOG" 2>&1 &
+    sleep 1
+}
+
+# Loops lines the hashes in the temp file. Check if all transfers we recieved a hash
+# back from the client are included in a block.
+function check_transfer_inclusion() {
+    local NODE_ID=${1}
+    local WALKBACK=${2}
+    local TRANSFER_HASHES=$(cat "$DEPLOY_LOG" | awk -F'::' '{print $4}' | sed 's/^[ \t]*//' | sed '/^[[:space:]]*$/d')
```
goral09, line 78: :dizzy_face: Good for you :upside_down_face:

goral09 (commented):
The code looks correct but I have some thoughts about the test scenario in general and I think this test will fail non-deterministically on occasion. 

We're sending deploys to node-5, saving every deploy hash, after 1 second we stop the node. The background process keeps trying to send the deploys now shutdown node. After it's done, we restart node-5 and verify that all of the deploy hashes we persisted can be found in the linear chain. In order for the deploys to be included node-5 had to either propose the deploy itself or gossip it to other nodes. If we send a deploy, persist its hash but node-5 gets shut down before it manages to propose a block with that deploy or gossip it, then we won't find it in the chain. 


Also, my second question, what are we trying to test here exactly? The gossiper? Or whether node persists deploys across restarts?

utils/nctl/sh/scenarios/common/itst.sh
TomVasile: yep, thats up next!
https://casperlabs.atlassian.net/browse/OP-1839

TomVasile:
> The code looks correct but I have some thoughts about the test scenario in general and I think this test will fail non-deterministically on occasion.
> 
> We're sending deploys to node-5, saving every deploy hash, after 1 second we stop the node. The background process keeps trying to send the deploys now shutdown node. After it's done, we restart node-5 and verify that all of the deploy hashes we persisted can be found in the linear chain. In order for the deploys to be included node-5 had to either propose the deploy itself or gossip it to other nodes. If we send a deploy, persist its hash but node-5 gets shut down before it manages to propose a block with that deploy or gossip it, then we won't find it in the chain and test will fail.
> 

It was operating under an assumption that node-5 would still retain the transfers in a buffer and then either gossip/propose them when he came back up. Is this not the case?

> Also, my second question, what are we trying to test here exactly? The gossiper? Or whether node persists deploys across restarts?

The latter, my focus was to test what happens to the deploys (transfers) we received a hash back for when a node dies mid-stream and whether they persist across the restart or are lost forever.

goral09:
I think the test does as much as it can – it interacts with the node as with a black box. We just need to understand its limitations.

Also, like I said, I think it may fail occassionally.

goral09 (approved):
As discussed – this test has its limitations but also covers some new ground. It will be useful to run it in nightly tests.

TomVasile:
As discussed in the meeting, looped the test a bunch of times with no failures yet. Going to merge with the note of the possibility of flakiness. If we see issues from the test in nightly runs i'll remove it.

PR #1312, opened 2021-04-14
HWY-282: Do not request dependency that is already in the synchronizer queue.

goral09:
https://casperlabs.atlassian.net/browse/HWY-282

Prior to the addition of [new logic](https://github.com/goral09/casper-node-1/blob/hwy-282/node/src/components/consensus/protocols/highway/synchronizer.rs#L373) we would try downloading a dependency (`c1`) that is already in the synchronizer's state. Now, instead, we will follow the vertices we already have and what those depend on and request the "lowest" one.

### DIFF
--- a/node/src/components/consensus/protocols/highway/synchronizer.rs
+++ b/node/src/components/consensus/protocols/highway/synchronizer.rs
@@ -151,13 +151,13 @@ where
     C: Context,
 {
     /// Incoming vertices we can't add yet because they are still missing a dependency.
-    vertex_deps: BTreeMap<Dependency<C>, PendingVertices<I, C>>,
+    vertices_awaiting_deps: BTreeMap<Dependency<C>, PendingVertices<I, C>>,
     /// The vertices that are scheduled to be processed at a later time.  The keys of this
     /// `BTreeMap` are timestamps when the corresponding vector of vertices will be added.
     vertices_to_be_added_later: BTreeMap<Timestamp, PendingVertices<I, C>>,
     /// Vertices that might be ready to add to the protocol state: We are not currently waiting for
     /// a requested dependency.
-    vertices_to_be_added: PendingVertices<I, C>,
+    vertices_no_deps: PendingVertices<I, C>,
     /// The duration for which incoming vertices with missing dependencies are kept in a queue.
     pending_vertex_timeout: TimeDiff,
     /// The duration between two consecutive requests of the latest state.
@@ -180,9 +180,9 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
         instance_id: C::InstanceId,
     ) -> Self {
         Synchronizer {
-            vertex_deps: BTreeMap::new(),
+            vertices_awaiting_deps: BTreeMap::new(),
             vertices_to_be_added_later: BTreeMap::new(),
-            vertices_to_be_added: Default::default(),
+            vertices_no_deps: Default::default(),
             pending_vertex_timeout,
             request_latest_state_timeout,
             oldest_seen_panorama: iter::repeat(None).take(validator_len).collect(),
@@ -194,9 +194,9 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
     /// Removes expired pending vertices from the queues, and schedules the next purge.
     pub(crate) fn purge_vertices(&mut self, now: Timestamp) {
         let oldest = now.saturating_sub(self.pending_vertex_timeout);
-        self.vertices_to_be_added.remove_expired(oldest);
+        self.vertices_no_deps.remove_expired(oldest);
         Self::remove_expired(&mut self.vertices_to_be_added_later, oldest);
-        Self::remove_expired(&mut self.vertex_deps, oldest);
+        Self::remove_expired(&mut self.vertices_awaiting_deps, oldest);
     }
 
     // Returns number of elements in the `verties_to_be_added_later` queue.
@@ -210,12 +210,15 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
 
     // Returns number of elements in `vertex_deps` queue.
     fn vertex_deps_len(&self) -> u64 {
-        self.vertex_deps.iter().map(|(_, pv)| pv.len()).sum()
+        self.vertices_awaiting_deps
+            .iter()
+            .map(|(_, pv)| pv.len())
+            .sum()
     }
 
     // Returns number of elements in `vertices_to_be_added` queue.
     fn vertices_to_be_added_len(&self) -> u64 {
-        self.vertices_to_be_added.len()
+        self.vertices_no_deps.len()
     }
 
     pub(crate) fn log_len(&self) {
@@ -301,7 +304,7 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
     /// `vertices_to_be_added` queue.
     pub(crate) fn remove_satisfied_deps(&mut self, highway: &Highway<C>) -> ProtocolOutcomes<I, C> {
         let satisfied_deps = self
-            .vertex_deps
+            .vertices_awaiting_deps
             .keys()
             .filter(|dep| highway.has_dependency(dep))
             .cloned()
@@ -309,7 +312,7 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
         // Safe to unwrap: We know the keys exist. TODO: Replace with BTreeMap::retain once stable.
         let pvs = satisfied_deps
             .into_iter()
-            .flat_map(|dep| self.vertex_deps.remove(&dep).unwrap())
+            .flat_map(|dep| self.vertices_awaiting_deps.remove(&dep).unwrap())
             .collect_vec();
         self.schedule_add_vertices(pvs)
     }
@@ -325,7 +328,7 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
         // Get the next vertex to be added; skip the ones that are already in the protocol state,
         // and the ones that are still missing dependencies.
         loop {
-            let pv = match self.vertices_to_be_added.pop() {
+            let pv = match self.vertices_no_deps.pop() {
                 None => return (None, outcomes),
                 Some(pv) if highway.has_vertex(pv.vertex()) => continue,
                 Some(pv) => pv,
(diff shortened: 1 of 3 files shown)


node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -210,12 +215,15 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
 
     // Returns number of elements in `vertex_deps` queue.
     fn vertex_deps_len(&self) -> u64 {
```
afck, line 217:
```suggestion
    fn vertices_awaiting_deps_len(&self) -> u64 {
```

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -210,12 +215,15 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
...
-        self.vertex_deps.iter().map(|(_, pv)| pv.len()).sum()
+        self.vertices_awaiting_deps
+            .iter()
+            .map(|(_, pv)| pv.len())
+            .sum()
     }
 
     // Returns number of elements in `vertices_to_be_added` queue.
     fn vertices_to_be_added_len(&self) -> u64 {
```
afck, line 225:
```suggestion
    fn vertices_no_deps_len(&self) -> u64 {
```

node/src/components/consensus/protocols/highway/synchronizer/tests.rs
```diff
@@ -111,3 +108,128 @@ fn purge_vertices() {
...
+        let vv = highway
+            .validate_vertex(pvv(*unit))
+            .unwrap_or_else(|_| panic!("{:?} unit is valid", unit));
+        highway.add_valid_vertex(vv, now);
+        let _ = sync.remove_satisfied_deps(&highway);
+    }
+    assert!(sync.is_empty());
+}
+
+#[cfg(test)]
```
afck, line 209: Isn't that annotation on the whole `tests` module already? No need to repeat it on individual items.

afck (commented):
This has an unintended side effect that we should probably address. In the example from the test:

Say Carol sent us c2 and then c1, so now we requested c0 from her.
Now Bob sends us b0.
What happens if Carol goes offline now?

With the existing code, Bob would eventually (and redundantly) send us c1, and then we'd request c0 from Bob. :heavy_check_mark: 

With the new code, we wouldn't request c1 from Bob, and never receive an answer from Carol, so we'd be stuck. :x: 
Even if Bob now sent us a new vertex b1, depending on b0, we wouldn't request anything from him.

node/src/components/consensus/protocols/highway/synchronizer/tests.rs
goral09: Removed in [`d3e8a82` (#1312)](https://github.com/CasperLabs/casper-node/pull/1312/commits/d3e8a82bae9fc19cf5a7839a640b0a219720e6c5).

node/src/components/consensus/protocols/highway/synchronizer.rs
goral09: Changed in [`d3e8a82` (#1312)](https://github.com/CasperLabs/casper-node/pull/1312/commits/d3e8a82bae9fc19cf5a7839a640b0a219720e6c5)

node/src/components/consensus/protocols/highway/synchronizer.rs
goral09: Changed in [`d3e8a82` (#1312)](https://github.com/CasperLabs/casper-node/pull/1312/commits/d3e8a82bae9fc19cf5a7839a640b0a219720e6c5)

goral09:
> This has an unintended side effect that we should probably address. In the example from the test:
> […]

You're correct. Thanks for pointing this out. I will move that PR to a Draft mode for now and continue working.

node/src/components/consensus/protocols/highway/synchronizer/tests.rs
```diff
@@ -183,12 +180,21 @@ fn do_not_download_synchronized_dependencies() {
...
     // "Download" the last dependency.
     let _ = sync.schedule_add_vertex(peer0, pvv(c0), now);
     // Now, the whole chain can be added to the protocol state.
-    for unit in &[c0, c1, b0, c2] {
+    let units: std::collections::BTreeSet<Dependency<TestContext>> = vec![c0, c1, b0, c2]
+        .into_iter()
+        .map(Dependency::Unit)
+        .collect();
+    for _ in 0..units.len() {
         let (maybe_pv, outcomes) = sync.pop_vertex_to_add(&highway);
```
afck, line 210: The test would be slightly stronger if you'd instead iterate over the vertices to add:
```rust
while let (Some(pv), outcomes) = sync.pop_vertex_to_add(&highway) {
    // …
    assert!(units.remove(&pv_dep));
    // …
}
```
That way we don't just check that we get at least `units.len()` vertices and they are all contained in `units`, but actually that none of them repeat and they are exactly identical to `units`.

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -339,20 +339,24 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
...
+                    // any of the dependencies currently so it should be retried soon.
                     self.add_missing_dependency(dep.clone(), pv);
                     continue;
                 }
                 // We are still missing a dependency. Store the vertex in the map and request
                 // the dependency from the sender.
+                // Find the first dependency that `pv` needs that we haven't synchronized yet
+                // and request it from the sender of `pv`. Since it relies on it, it should have
+                // it as well.
+                let transitive_dependency = self.find_transitive_dependency(dep.clone());
```
afck, line 355: We should move this line to the top (above 342):
```rust
let transitive_dependency = self.find_transitive_dependency(dep.clone());
if self.vertices_no_deps.contains_dependency(&transitive_dependency) {
    self.add_missing_dependency(dep.clone(), pv);
    continue;
}
```

Because if the _transitive_ dependency is already in `vertices_no_deps`, there's also no reason to request it.

E.g.: Bob sends us `c2`, while `c1` is in `vertices_awaiting_deps`. But we already received even `c0` from Alice, and it's in `vertices_no_deps`. Then we also don't need to request `c0` from Bob.

node/src/components/consensus/protocols/highway/synchronizer/tests.rs
```diff
@@ -67,7 +67,10 @@ fn purge_vertices() {
     assert!(maybe_pv.is_none());
     assert!(matches!(
         *outcomes,
-        [ProtocolOutcome::CreatedTargetedMessage(_, NodeId(0))]
+        [
+            ProtocolOutcome::CreatedTargetedMessage(_, NodeId(0)),
+            ProtocolOutcome::CreatedTargetedMessage(_, NodeId(0))
```
afck, line 72: It would be good to check what we're requesting here. Maybe re-create the message and serialize it, so we can make the full comparison?

node/src/components/consensus/protocols/highway/synchronizer.rs
goral09: Good point.

node/src/components/consensus/protocols/highway/synchronizer/tests.rs
goral09: After changing the code as you asked in https://github.com/CasperLabs/casper-node/pull/1312#discussion_r614081731, there's only one dependency now being requested.

goral09: Done in [`60627c6` (#1312)](https://github.com/CasperLabs/casper-node/pull/1312/commits/60627c68a9551e909c18b7ad4107de17aea904a9).

node/src/components/consensus/protocols/highway/synchronizer/tests.rs
```diff
@@ -216,7 +213,7 @@ fn do_not_download_synchronized_dependencies() {
             "unexpected dependency request {:?}",
             outcomes
         );
-        let pv_dep = maybe_pv.unwrap().vertex().id();
+        let pv_dep = pv.vertex().id();
         assert!(units.contains(&pv_dep), "unexpected dependency");
```
afck, line 217: Why not remove it, so we also check that there are no duplicates?
```suggestion
        assert!(units.remove(&pv_dep), "unexpected dependency");
```

node/src/components/consensus/protocols/highway/synchronizer.rs
```diff
@@ -349,13 +356,18 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
...
                 // We are still missing a dependency. Store the vertex in the map and request
                 // the dependency from the sender.
-                // Find the first dependency that `pv` needs that we haven't synchronized yet
-                // and request it from the sender of `pv`. Since it relies on it, it should have
-                // it as well.
-                let transitive_dependency = self.find_transitive_dependency(dep.clone());
                 let sender = pv.sender().clone();
                 let ser_msg = HighwayMessage::RequestDependency(transitive_dependency).serialize();
-                outcomes.push(ProtocolOutcome::CreatedTargetedMessage(ser_msg, sender));
+                if !outcomes.iter().any(|o| match o {
```
afck, line 361: I'd prefer to either make `outcomes` a set (and then turn it into a `Vec` only at the end of the method) or, even better:
Add a set `requested_deps: HashSet<(I, Dependency<C>)>`, so we don't even serialize the same message twice.

Although neither of those will prevent us from requesting the same vertex twice from the same peer in two separate `pop_vertex_to_add` calls.

afck (approved):
No blockers; just wondering how it can be simplified.

node/src/components/consensus/protocols/highway/synchronizer.rs
goral09: > Although neither of those will prevent us from requesting the same vertex twice from the same peer in two separate `pop_vertex_to_add` calls.

Deduplicating in-flight requests was not the goal of this PR.

> I'd prefer to either make outcomes a set (and then turn it into a `Vec` only at the end of the method) or, even better:
Add a set `requested_deps: HashSet<(I, Dependency<C>)>`, so we don't even serialize the same message twice.

Sure, I was too lazy to do it :)

goral09: Done in [`12bb6c2` (#1312)](https://github.com/CasperLabs/casper-node/pull/1312/commits/12bb6c28701da5269fcc8d5cec73b2509afa6f36).

node/src/components/consensus/protocols/highway/synchronizer/tests.rs
goral09: Done in [`12bb6c2` (#1312)](https://github.com/CasperLabs/casper-node/pull/1312/commits/12bb6c28701da5269fcc8d5cec73b2509afa6f36)

afck (approved):
Thanks! :+1:

fizyk20 (approved):

goral09:
bors merge

PR #1314, opened 2021-04-15
NDRS-1130: Remove BlockLike; verify deploy type in block validator.

afck:
This replaces the `BlockLike` trait with an enum that is used only within the block validator, and adds checks that the `transfer_hashes` contain only transfers, and the `deploy_hashes` contain none.

**Renames the metric `mem_block_payload_validator` to `mem_block_validator`!**

https://casperlabs.atlassian.net/browse/NDRS-1130

### DIFF
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -65,12 +65,24 @@ impl DeployOrTransferHash {
     }
 }
 
-#[derive(DataSize, Debug, Display, From, Clone, Hash, Eq, PartialEq)]
+#[derive(DataSize, Debug, Display, Clone, Hash, Eq, PartialEq)]
 pub enum ValidatingBlock {
     #[display(fmt = "{}", _0.display())]
     Block(Box<Block>),
     #[display(fmt = "{}", _0.display())]
     ProposedBlock(Box<ProposedBlock<ClContext>>),
+}
+
+impl From<Block> for ValidatingBlock {
+    fn from(block: Block) -> ValidatingBlock {
+        ValidatingBlock::Block(Box::new(block))
+    }
+}
+
+impl From<ProposedBlock<ClContext>> for ValidatingBlock {
+    fn from(proposed_block: ProposedBlock<ClContext>) -> ValidatingBlock {
+        ValidatingBlock::ProposedBlock(Box::new(proposed_block))
+    }
 }
 
 impl ValidatingBlock {
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -163,6 +163,13 @@ where
     fn source(&mut self) -> Option<I> {
         self.sources.pop_front()
     }
+
+    fn respond<REv>(&mut self, value: bool) -> Effects<REv> {
+        self.responders
+            .drain(..)
+            .flat_map(|responder| responder.respond(value).ignore())
+            .collect()
+    }
 }
 
 #[derive(DataSize, Debug)]
@@ -314,16 +321,12 @@ where
                 // Now we remove all states that have finished and notify the requestors.
                 self.validation_states.retain(|key, state| {
                     if invalid.contains(key) {
-                        state.responders.drain(..).for_each(|responder| {
-                            effects.extend(responder.respond(false).ignore());
-                        });
+                        effects.extend(state.respond(false));
                         return false;
                     }
                     if state.missing_deploys.is_empty() {
                         // This one is done and valid.
-                        state.responders.drain(..).for_each(|responder| {
-                            effects.extend(responder.respond(true).ignore());
-                        });
+                        effects.extend(state.respond(true));
                         return false;
                     }
                     true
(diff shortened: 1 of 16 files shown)


node/src/components/block_validator.rs
```diff
@@ -186,33 +241,26 @@ where
...
-                    // If there are no deploys, return early.
-                    return responder.respond((true, block)).ignore();
+                    return responder.respond(false).ignore();
                 }
 
-                // TODO: Clean this up to use `or_insert_with_key` once
-                // https://github.com/rust-lang/rust/issues/71024 is stabilized.
                 match self.validation_states.entry(block) {
                     Entry::Occupied(mut entry) => {
                         // The entry already exists.
```
xcthulhu, line 259: Note: this may be unreachable, but this is the correct behavior even in the case that we do reach it.

node/src/components/block_validator/tests.rs
```diff
@@ -0,0 +1,282 @@
...
+    )
+}
+
+/// Validates a block using a `BlockValidator` component, and returns the result.
+async fn validate_block(
+    rng: &mut TestRng,
+    timestamp: Timestamp,
+    deploys: Vec<Deploy>,
+    transfers: Vec<Deploy>,
+) -> bool {
```
xcthulhu, line 154: Note: this will not model the case in which a block in invalid because we could not fetch its deploys.

New JIRA ticket: https://casperlabs.atlassian.net/browse/NDRS-1153

node/src/components/block_validator/tests.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    // Otherwise the effects must be requests to fetch the block's deploys.
+    let fetch_results: Vec<_> = effects.into_iter().map(tokio::spawn).collect();
+
+    // We make our mock reactor answer with the expected deploys and transfers:
+    for deploy in deploys.into_iter().chain(transfers) {
+        reactor.expect_fetch_deploy(deploy).await;
+    }
+
```
xcthulhu, line 189: Note: Reactor should be empty now

node/src/components/block_validator/tests.rs
```diff
@@ -0,0 +1,282 @@
...
+    let mut block_validator = BlockValidator::<NodeId>::new(chainspec);
+
+    // Pass the block to the component. This future will eventually resolve to the result, i.e.
+    // whether the block is valid or not.
+    let validation_result =
+        tokio::spawn(effect_builder.validate_proposed_block("Bob", proposed_block.clone()));
+    let event = reactor.expect_block_validator_event().await;
+    let effects = block_validator.handle_event(effect_builder, rng, event);
+
+    // If validity could already be determined, the effect will be the validation response.
```
xcthulhu, line 173: This is the empty block case or the case where the block had duplicate deploy hashes.  There may be other cases.

node/src/effect.rs
```diff
@@ -1183,48 +1184,38 @@ impl<REv> EffectBuilder<REv> {
...
-    /// within the block.
-    pub(crate) async fn validate_block<I>(
-        self,
-        sender: I,
-        block: Block,
-        block_timestamp: Timestamp,
-    ) -> (bool, Block)
+    /// Checks whether the deploys included in the block exist on the network and the block is
+    /// valid.
+    pub(crate) async fn validate_block<I>(self, sender: I, block: Block) -> bool
```
xcthulhu, line 1189: Chatting with @afck , let's have this be generic in `Into<ValidatingBlock>`.

node/src/reactor/joiner.rs
```diff
@@ -168,18 +168,12 @@ pub enum Event {
...
 
     /// Block proposer request.
     #[from]
     BlockProposerRequest(#[serde(skip_serializing)] BlockProposerRequest),
 
-    /// Block payload validator request.
-    #[from]
-    BlockPayloadValidatorRequest(
-        #[serde(skip_serializing)] BlockValidationRequest<BlockPayload, NodeId>,
-    ),
```
xcthulhu, line 181: Good catch that this is no longer used!

node/src/reactor/validator/memory_metrics.rs
```diff
@@ -86,9 +86,9 @@ impl MemoryMetrics {
         )?;
         let mem_block_proposer =
             IntGauge::new("mem_block_proposer", "block_proposer memory usage in bytes")?;
-        let mem_block_payload_validator = IntGauge::new(
-            "mem_block_payload_validator",
-            "block_payload_validator memory usage in bytes",
+        let mem_block_validator = IntGauge::new(
+            "mem_block_validator",
```
xcthulhu, line 90: CC @sacherjj @TomVasile looks like the metric `mem_block_payload_validator` is being changed to `mem_block_validator` here... not sure we are surfacing this in our dashboard.

xcthulhu (approved):
Note that we might want to make `EffectBuilder::validate_block` generic for `Into<ValidatingBlock>`.

node/src/components/block_validator/tests.rs
afck: As discussed, I wouldn't explicitly check that here, to keep the test as simple as possible.

node/src/effect.rs
afck: Done in https://github.com/CasperLabs/casper-node/pull/1314/commits/a1f34f21f5632b03aeb0da8b85588ba348409c47.

fizyk20 (approved):

PR #1316, opened 2021-04-15
NDRS-1137: small net instrumentation

marc-casperlabs:
Adds some observability to the small_net networking component.

# Message classification

Network messages are classified into the following categories:

* `protocol`: Handshakes
* `consensus`
* `deploy_gossip`
* `address_gossip`
* `deploy_transfer`: Any traffic of deploys being requested directly (via joining or when downloading blocks)
* `block_transfer`: Any traffic of blocks being requested directly (mostly joining)
* `other`: Other traffic. Should be very rare.

# Metrics

Additional metrics are added, namely `net_out_count_*` and `net_out_bytes_`. These contain the summed up single message counts and total number of bytes transferred across the entire node for each category (e.g. `net_out_bytes_block_transfer` is the total number of bytes transferred to satisfy block requests).

# Tracing

Two tracing targets have been introduced, `net_out` and `net_in`. If set to the `TRACE` level, they mark and log each outgoing packet. `msg_id` or trace IDs are unique per message sent on a connection, even if the payload is identical.

Example:

```
{
  "timestamp": "Apr 19 14:49:15.239",
  "level": "TRACE",
  "fields": {
    "message": "sending",
    "msg_id": "19e5aee6f6880df3",
    "msg_size": 25,
    "msg_kind": "protocol"
  },
  "target": "net_out"
}
```

The log message above was recorded on node 1. On node 2, we can find the resulting `net_in` message:

```
{
  "timestamp": "Apr 19 14:49:15.239",
  "level": "TRACE",
  "fields": {
    "message": "received",
    "msg_id": "19e5aee6f6880df3",
    "msg_size": 25,
    "msg_kind": "protocol"
  },
  "target": "net_in"
}
```

The timestamp difference between both is 0, which means it got transferred in less than half a millisecond (not surprising, since we used loopback networking and are under no load).

To run with these traces enabled, `RUST_LOG` or equivalent can be used:

```
RUST_LOG=info,net_out=trace,net_in=trace
```


goral09:
Why can't we have `msg_kind` and `msg_size` on the receiving node as well?

node/src/protocol.rs
```diff
@@ -43,6 +46,29 @@ pub enum Message {
...
+                    Tag::Deploy => MessageKind::DeployTransfer,
+                    Tag::Block => MessageKind::BlockTransfer,
+                    // This is a weird message, which we should not encounter here?
+                    Tag::GossipedAddress => MessageKind::Other,
+                    Tag::BlockByHeight => MessageKind::BlockTransfer,
+                    Tag::BlockHeaderByHash => MessageKind::BlockTransfer,
+                    Tag::BlockHeaderAndFinalitySignaturesByHeight => MessageKind::BlockTransfer,
+                }
+            }
+            Message::FinalitySignature(_) => MessageKind::BlockTransfer,
```
goral09, line 67: We should disambiguate between `Message::FinalitySignature` and `Tag::BlockByHeight`, `Tag::BlockHeaderByHash` and `Tag::BlockHeaderAndFinalitySignaturesByHeight` – the latter are issues on joining node while the latter (`FinalitySignature(_)`) is created during normal node operation.

node/src/components/small_network.rs
```diff
@@ -1,9 +1,8 @@
 //! Fully connected overlay network
 //!
-//! The *small network* is an overlay network where each node participating is connected to every
-//! other node on the network. The *small* portion of the name stems from the fact that this
-//! approach is not scalable, as it requires at least $O(n)$ network connections and broadcast will
-//! result in $O(n^2)$ messages.
+//! The *small network* is an overlay network where each node participating is attempting to main a
```
goral09, line 3:
```suggestion
//! The *small network* is an overlay network where each node participating is attempting to maintain a
```

node/src/components/small_network/counting_format.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    #[inline]
+    fn deserialize(self: Pin<&mut Self>, src: &BytesMut) -> Result<Message<P>, Self::Error> {
+        let this = self.project();
+        let projection: Pin<&mut F> = this.inner;
+
+        // We do not include additional meta info here, since we do not want the deserialization
+        // time to be added to our measurements.
+        trace!(target: "net_in",
+            msg_id=%TraceId(&src),
```
goral09, line 93: Since the `msg_id` is derived only from the serialized message, it might be duplicated across nodes (if we send the same data to multiple nodes) but we might also have the same `msg_id` for _incoming_ messages (if they are the same message)?

goral09 (commented):
Nice, really clean :clap:

marc-casperlabs:
*cough* Unfortunately, my GPG key expired an hour ago or so, so I will have to force push this. Sorry!

node/src/protocol.rs
marc-casperlabs: I wanted to be fairly course and not explode the number of categories.

Should I move `FinalitySignature` into the `Consensus` class maybe?

goral09: If we don't want to add more categories then `Consensus` is :+1:

node/src/components/small_network/counting_format.rs
marc-casperlabs: That's true. I guess we could add a second ID depending on what we want to track.

goral09:
> Why can't we have `msg_kind` and `msg_size` on the receiving node as well?
> 
> **EDIT**: I can see the comment now:
> 
> ```
> // We do not include additional meta info here, since we do not want the deserialization
> // time to be added to our measurements.
> ```
> 
> but you call `F::deserialize` a couple of lines below. Couldn't we assign the `received` timestamp to a variable, deserialize and then log? This way we can have more detailed measurements on the receiving nodes as well without any degradation to the performance.

Ping! Can it be done @marc-casperlabs ?

EdHastingsCasperLabs (approved):

node/src/components/small_network/counting_format.rs
goral09: I have no idea how to fix that from inside the (de)serializer.

marc-casperlabs:
> Ping! Can it be done @marc-casperlabs ?

I am working on a follow-up.

node/src/components/small_network/counting_format.rs
```diff
@@ -87,26 +128,179 @@ where
...
+    #[inline]
+    fn in_flag(self) -> u8 {
+        !(self.out_flag())
+    }
+
+    #[inline]
+    fn out_flag(self) -> u8 {
+        match self {
+            Role::Dialer => 0xaa,
+            Role::Listener => !0xaa,
```
goral09, line 307: Why this complicated? Why not simply `0u8` and `1u8`?

node/src/components/small_network/counting_format.rs
```diff
@@ -38,15 +53,35 @@ pub(super) struct CountingFormat<F> {
     /// The actual serializer performing the work.
     #[pin]
     inner: F,
```
goral09, line 57: Can you explain why `inner` has to be pinned?

goral09 (approved):
This got a bit more complicated, I hope I didn't miss anything. Looks good :)

node/src/components/small_network/counting_format.rs
marc-casperlabs: Because I want to call things like [`Deserializer::deserialize`](https://docs.rs/tokio-serde/0.8.0/tokio_serde/trait.Deserializer.html#tymethod.deserialize), which take a `self: Pin<&mut Self>, ` :)

node/src/components/small_network/counting_format.rs
marc-casperlabs: Having them be the inverse of each other is a useful property for the code below, so it'd be `0` and `256` instead. I thought it'd be neat to have the `1010_1010` and `0101_0101` bit patterns here.

goral09: Oh, I missed that completely. Can you use the `0b` format then and/or add a comment?

marc-casperlabs: I'll do both!

node/src/components/small_network/counting_format.rs
goral09: Ok, I thought it's the other way around – b/c you used `#[pin]` the self is now `Pin<…>`.

marc-casperlabs:
> This got a bit more complicated, I hope I didn't miss anything. Looks good :)

Yeah, sorry, I completely underestimated the effort to track messages that are identical, but sent to different nodes.

On the plus side, we no longer hash the serialized representation, so it got a little faster :)

node/src/components/small_network/counting_format.rs
marc-casperlabs: It's fixed with the latest changes, although slightly differently. Each connection issues IDs not based on content, but using a sequence number on each connection instead (`ConnectionId::create_trace_id`).

PR #1317, opened 2021-04-15
EE-1226: Fix bid docstrings

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1226

Improves rust docs around bid/delegate entry points and related comments.


types/src/system/auction/mod.rs
```diff
@@ -52,10 +52,9 @@ pub type UnbondingPurses = BTreeMap<AccountHash, Vec<UnbondingPurse>>;
 pub trait Auction:
     StorageProvider + SystemProvider + RuntimeProvider + MintProvider + AccountProvider + Sized
 {
-    /// Returns era_validators.
+    /// Returns era validators for each configured trusted eras.
     ///
-    /// Publicly accessible, but intended for periodic use by the Handle Payment contract to update
-    /// its own internal data structures recording current and past winners.
+    /// Publicly accessible, but intended for periodic use by the consensus through an API call.
```
EdHastingsCasperLabs, line 57: I would delete this line altogether.

types/src/system/auction/mod.rs
```diff
@@ -52,10 +52,9 @@ pub type UnbondingPurses = BTreeMap<AccountHash, Vec<UnbondingPurse>>;
 pub trait Auction:
     StorageProvider + SystemProvider + RuntimeProvider + MintProvider + AccountProvider + Sized
 {
-    /// Returns era_validators.
+    /// Returns era validators for each configured trusted eras.
```
EdHastingsCasperLabs, line 55: /// Returns active validators and auction winners for a number of future eras determined by the configured auction_delay.

dwerner (approved):

types/src/system/auction/mod.rs
```diff
@@ -90,9 +88,18 @@ pub trait Auction:
...
-    /// calls `bond` in the Mint contract to create (or top off) a bid purse. It also adjusts the
-    /// delegation rate.
+    /// This entry point adds or modifies an entry in the `Key::Bid` section of the global state and
+    /// creates (or tops off) a bid purse. Post genesis, any new call on this entry point causes a
+    /// non-founding validator in the system to exist.
+    ///
+    /// The logic works for both founding and non-founding validators, making it possible to adjust
+    /// their delegation rate and increase their stakes.
+    ///
+    /// A validator with its bid inactive due to slashing can activate his bid again by increasing
```
Fraser999:
```suggestion
    /// A validator with its bid inactive due to slashing can activate its bid again by increasing
```

types/src/system/auction/mod.rs
```diff
@@ -90,9 +88,18 @@ pub trait Auction:
...
+    /// its stake.
+    ///
+    /// Validators cannot create a bid with 0 amount, and the delegation rate can't exceed
+    /// [`DELEGATION_RATE_DENOMINATOR`].
```
Fraser999, line 100: The doc comment should explain what is returned by the method.

types/src/system/auction/mod.rs
```diff
@@ -152,14 +159,17 @@ pub trait Auction:
         Ok(updated_amount)
     }
 
-    /// For a non-founder validator, implements essentially the same logic as add_bid, but reducing
-    /// the number of tokens and calling unbond in lieu of bond.
+    /// For a non-founder validator, this entry point implements essentially the same logic as
```
Fraser999: I don't think we should refer to `add_bid` here since it leaves too many open questions that I think the doc comment should answer.

For example,
* what happens if `amount` is more than they have in their bid account?
* what happens if they pass `amount` set to `0`? (It's not the same as for `add_bid` by the looks of it - we don't error here)
* if the bid account is left with `0` is the validator's entry in the `Key::Bid` section of the global state removed or left with a value of `0`?

types/src/system/auction/mod.rs
```diff
@@ -152,14 +159,17 @@ pub trait Auction:
...
+    /// `add_bid` but reducing the number of tokens staked in the bid.
+    ///
+    /// For a founding validator, this function first checks whether they are released and fails if
+    /// they are not.
     ///
-    /// For a founding validator, this function first checks whether they are released, and fails
-    /// if they are not.
+    /// All the delegators that bid their tokens on a particular validator are automatically
```
Fraser999: This is only true if the bid account is reduced to `0` though, yes?

types/src/system/auction/mod.rs
```diff
@@ -215,11 +225,12 @@ pub trait Auction:
         Ok(updated_stake)
     }
 
-    /// Adds a new delegator to delegators, or tops off a current one. If the target validator is
-    /// not in founders, the function call returns an error and does nothing.
+    /// Adds a new delegator to delegators or tops its current stake. If the target validator is
```
Fraser999:
```suggestion
    /// Adds a new delegator to delegators or increases its current stake. If the target validator is
```

types/src/system/auction/mod.rs
```diff
@@ -286,12 +297,12 @@ pub trait Auction:
...
-    /// the entry in delegators and calls unbond in the Mint contract to create a new unbonding
-    /// purse.
+    /// Removes specified amount of motes (or the value from the collection altogether, if the
+    /// remaining amount is 0) from the entry in delegators map for given validator and creates a
+    /// new unbonding request to the queue.
     ///
-    /// The arguments are the delegator’s key, the validator key and quantity of motes and
-    /// returns a tuple of the unbonding purse along with the remaining bid amount.
+    /// The arguments are the delegator's key, the validator's key, and the remaining stake and
+    /// return the remaining bid amount.
```
Fraser999: I'm a little confused by "return the remaining bid amount".

I think it'd be good to have this as a separate sentence, but that aside, can we try and be fairly consistent in our description of the return types for all the methods being documented in this PR please?

execution_engine/CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Changed
 * Update pinned version of Rust to `nightly-2021-06-17`
-
+* Changed comments to clarify behavior of the bidding functionality
```
```suggestion
* Improve doc comments to clarify behavior of the bidding functionality.

```
(there should be 3 blank lines before the start of a new version's section)

Fraser999 (approved):
LGTM

PR #1319, opened 2021-04-15
NDRS-1120: Add a delay before proposing dependencies. (1.0.2 backport)

afck:
This adds a configurable delay between the time we receive a deploy (from a client or another full node) and the time we propose it in a block. The intention is to give deploys time to spread across the network via gossip, so that when we include them in a new block, the other nodes won't all have to ask us for the deploy.

Backport of https://github.com/CasperLabs/casper-node/pull/1283.

### DIFF
--- a/node/src/types/timestamp.rs
+++ b/node/src/types/timestamp.rs
@@ -247,11 +247,6 @@ impl TimeDiff {
     pub const fn from_seconds(seconds: u32) -> Self {
         TimeDiff(seconds as u64 * 1_000)
     }
-
-    /// Returns the product, or `TimeDiff(u64::MAX)` if it would overflow.
-    pub fn saturating_mul(self, rhs: u64) -> Self {
-        TimeDiff(self.0.saturating_mul(rhs))
-    }
 }
 
 impl Mul<u64> for TimeDiff {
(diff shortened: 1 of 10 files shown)


goral09 (approved):

node/src/types/timestamp.rs
```diff
@@ -247,6 +247,11 @@ impl TimeDiff {
     pub const fn from_seconds(seconds: u32) -> Self {
         TimeDiff(seconds as u64 * 1_000)
     }
+
+    /// Returns the product, or `TimeDiff(u64::MAX)` if it would overflow.
+    pub fn saturating_mul(self, rhs: u64) -> Self {
+        TimeDiff(self.0.saturating_mul(rhs))
+    }
```
xcthulhu, line 254:
```suggestion
```

This is not used.

resources/production/config-example.toml
```diff
@@ -41,6 +41,7 @@ pending_vertex_timeout = '30min'
 max_execution_delay = 3
 
 
+
```
xcthulhu, line 44:
```suggestion
```

Spurious line.

xcthulhu (commented):
I've reviewed this, but we must test this.  I have assigned this to @TomVasile

afck:
Closing in favor of https://github.com/CasperLabs/casper-node/pull/1323.

PR #1321, opened 2021-04-16
EE-1245: revert EE-1205

henrytill:
Ref: 
https://github.com/CasperLabs/casper-node/pull/1262


dwerner (approved):

types/src/execution_result.rs
```diff
@@ -571,16 +562,7 @@ impl ToBytes for Transform {
...
-            Transform::WriteWithdraw(value) => value.serialized_length() + U8_SERIALIZED_LENGTH,
-            Transform::WriteEraValidators(value) => {
-                value.serialized_length() + U8_SERIALIZED_LENGTH
-            }
-            Transform::Identity
-            | Transform::WriteContractWasm
-            | Transform::WriteContract
-            | Transform::WriteContractPackage
-            | Transform::Delete => U8_SERIALIZED_LENGTH,
+            _ => U8_SERIALIZED_LENGTH,
```
Fraser999, line 565: This change looks wrong.  We should also probably avoid using a catchall case here and explicitly list every variant.

types/src/execution_result.rs
```diff
@@ -652,7 +633,7 @@ impl FromBytes for Transform {
 impl Distribution<Transform> for Standard {
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Transform {
         // TODO - include WriteDeployInfo and WriteTransfer as options
-        match rng.gen_range(0..14) {
+        match rng.gen_range(0..13) {
```
Fraser999, line 636: I guess not for this PR, but it's worrying that we're not covering all variants here.  We use this in the bytesrepr roundtrip test, so only 13 of the 19 variants are being tested.

Fraser999 (changes requested):
Just a simple bug in the `Transform` code blocking this.

types/src/execution_result.rs
dwerner: This is supposed to be a revert PR only, so we should put that in a separate PR.

henrytill: Agreed, I would like to do this in a separate PR.  These are purely reverting things to where they were.

mpapierski (approved):

EdHastingsCasperLabs (approved):

types/src/execution_result.rs
EdHastingsCasperLabs: double plus agreed

Fraser999: Sure - I'll approve on that assumption

Fraser999 (approved):

PR #1325, opened 2021-04-16
NDRS-1142: Use only block headers, not bodies, in the consensus component.

afck:
Fast sync won't download all block bodies, so in preparation, the consensus component needs to not rely on them.

https://casperlabs.atlassian.net/browse/NDRS-1142

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -50,8 +50,8 @@ use crate::{
     },
     fatal,
     types::{
-        ActivationPoint, Block, BlockHash, BlockHeader, DeployHash, FinalitySignature,
-        FinalizedBlock, TimeDiff, Timestamp,
+        ActivationPoint, BlockHash, BlockHeader, DeployHash, FinalitySignature, FinalizedBlock,
+        TimeDiff, Timestamp,
     },
     utils::WithDir,
     NodeRng,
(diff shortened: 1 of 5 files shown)


goral09 (approved):
Code looks correct and I haven't found any residual usages of `Block` in the consensus component.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -711,19 +711,19 @@ where
         })
     }
 
-    pub(super) fn handle_block_added(&mut self, block: Block) -> Effects<Event<I>> {
+    pub(super) fn handle_block_added(
+        &mut self,
+        block_hash: BlockHash,
```
xcthulhu, line 716: `BlockHash` isn't necessary here; I am more leery of juggling an invariant than I am of micro-optimizing.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -711,19 +711,19 @@ where
...
+        self.era_supervisor.executed_block(&block_header);
         let mut effects = if self.era_supervisor.is_validator_in(&our_pk, era_id) {
-            let block_hash = block.hash();
             self.effect_builder
                 .announce_created_finality_signature(FinalitySignature::new(
-                    *block_hash,
-                    era_id,
-                    &our_sk,
-                    our_pk,
+                    block_hash, era_id, &our_sk, our_pk,
```
xcthulhu, line 726:
```suggestion
                    block_header.hash(), era_id, &our_sk, our_pk,
```

afck: Done in https://github.com/CasperLabs/casper-node/pull/1325/commits/e56eeeefffda3e6d09a9a0d7184240c0e872cb92.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/CasperLabs/casper-node/pull/1325/commits/e56eeeefffda3e6d09a9a0d7184240c0e872cb92.

xcthulhu (approved):

PR #1326, opened 2021-04-16
EE-1221: Increase create purse / mint entry point cost

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1221

To discourage creating many empty purses to deplete storage resources, this PR is increasing create_purse/mint cost to 2.5cspr.


darthsiroftardis (approved):

execution_engine/src/shared/host_function_costs.rs
```diff
@@ -23,7 +23,7 @@ const DEFAULT_ADD_COST: u32 = 5_800;
 const DEFAULT_CALL_CONTRACT_COST: u32 = 4_500;
 const DEFAULT_CALL_CONTRACT_ARGS_SIZE_WEIGHT: u32 = 420;
 
-const DEFAULT_CREATE_PURSE_COST: u32 = 170_000;
+const DEFAULT_CREATE_PURSE_COST: u32 = 1_700_000;
```
EdHastingsCasperLabs, line 26: shouldn't this also be 2_500_000_000; ?

Fraser999 (approved):

execution_engine/CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Changed
 * Support building and testing using stable Rust.
+* Changed price of `create_purse` to 2.5CSPR to discourage people from creating purses in the payment code.
```
goral09, line 21: The "in the payment code" is confusing b/c it increases the price across the whole execution, including session code. Is that correct @mpapierski ?

mpapierski: Correct, but it was increased specifically to disallow creating those in the payment code.

goral09: But the comment makes it sound as if the cost is increased only in the payment code. I don't think the last part needs to be included in the changelog.

PR #1332, opened 2021-04-16
Bump version 1.1.1 and known address prune

sacherjj:
Bumped version to 1.1.1
Set activation Era to 360. (~ 2021-04-30 15:30:00Z)
Pruned 5 inactive nodes from known_addresses
add_bid and delegate to 2_500_000_000 cost for defense.


TomVasile:
our bumps look good, deferring to ed/fraser on the dependency bumps

TomVasile (approved):

PR #1336, opened 2021-04-17
Do not truncate state root hash.

mcdee:
A call to the status endpoint at `:8888/status` returns the starting state root hash as truncated, e.g.

```
...
"starting_state_root_hash": "8e22..b679",
...
```

This seems unnecessary, and removes information from the API result.  This PR removes the truncation and returns the full starting state root hash.

### DIFF
--- a/node/src/types/status_feed.rs
+++ b/node/src/types/status_feed.rs
@@ -157,7 +157,7 @@ pub struct GetStatusResult {
     /// The chainspec name.
     pub chainspec_name: String,
     /// The state root hash used at the start of the current session.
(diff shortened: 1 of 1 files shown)


sacherjj:
bors try

sacherjj:
@mcdee This would need to target dev. We do not merge directly to master. I will let the core devs comment on the reason for thumbprint.

sacherjj:
You would also need to rebase this on dev.

Fraser999:
Closed in favour of #1517.

PR #1338, opened 2021-04-19
NDRS-1160: Merge Fast Sync Prototype

xcthulhu:


node/src/components/contract_runtime.rs
```diff
@@ -483,13 +484,8 @@ where
...
                     } => {
                         trace!(?trie_key, "read_trie request");
-                        let engine_state = Arc::clone(&self.engine_state);
-                        let metrics = Arc::clone(&self.metrics);
+                        let result = self.read_trie(trie_key);
                         async move {
-                            let correlation_id = CorrelationId::new();
-                            let start = Instant::now();
-                            let result = engine_state.read_trie(correlation_id, trie_key);
-                            metrics.read_trie.observe(start.elapsed().as_secs_f64());
```
xcthulhu, line 492: @dwerner Got rid of `read_trie` metric; it is unused in production.

node/src/components/fetcher/event.rs
```diff
@@ -67,6 +69,22 @@ pub enum Event<T: Item> {
     TimeoutPeer { id: T::Id, peer: NodeId },
 }
 
+impl<T: Item> Event<T> {
+    pub fn from_get_response_serialized_item(peer: NodeId, serialized_item: &[u8]) -> Option<Self> {
```
xcthulhu, line 73: @fizyk20 Wasn't sure how to handle the server side of the fetcher using a parameter, but I managed to abstract the client side.

node/src/components/linear_chain_sync.rs
```diff
@@ -67,6 +73,10 @@ pub use traits::ReactorEventT;
 
 #[derive(DataSize, Debug)]
 pub(crate) struct LinearChainSync<I> {
+    init_hash: Option<BlockHash>,
+    chainspec: Chainspec,
+    started: bool,
```
xcthulhu, line 78: Since fast syncing is going to change wire formats, the rest of this code is going to break.  For now, the prototype is part of this...

node/src/components/linear_chain_sync.rs
```diff
@@ -778,18 +799,43 @@ where
...
+                                init_hash,
+                                chainspec,
+                            )
+                            .await
+                            {
+                                Ok(()) => Some(Event::Start(peer_id)),
+                                Err(error) => {
+                                    effect_builder
+                                        .fatal(file!(), line!(), format!("{:?}", error))
+                                        .await;
```
xcthulhu, line 819: @marc-casperlabs Wasn't sure how to use the macro here...

node/src/components/linear_chain_sync.rs
```diff
@@ -778,18 +799,43 @@ where
...
-                            .immediately()
-                            .event(move |_| Event::Start(cloned_peer_id)),
-                    );
+                self.peers.push(peer_id.clone());
+                match self.init_hash {
+                    Some(init_hash) if !self.started => {
+                        self.started = true;
+                        let chainspec = self.chainspec.clone();
+                        (async move {
+                            match operations::run_fast_sync_task(
```
xcthulhu, line 808: Here is the entry point for the fast sync task.

I intend to remove the other events for this component.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+const TIMEOUT_DURATION: Duration = Duration::from_millis(100);
+
+/// Get the currently connected peers from the networking component and shuffle them
+async fn get_and_shuffle_network_peers<REv, I>(effect_builder: EffectBuilder<REv>) -> Vec<I>
+where
+    REv: From<NetworkInfoRequest<I>>,
+    I: Send + 'static,
+{
+    let mut vector_of_peers: Vec<I> = effect_builder
+        .network_peers::<I>()
```
xcthulhu, line 35: @marc-casperlabs We might want to expose a different effect for this...

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+async fn fetch_retry_forever<T, REv, I>(
+    effect_builder: EffectBuilder<REv>,
+    id: T::Id,
+) -> Result<Box<T>, FetcherError<T, I>>
+where
+    T: Item + 'static,
+    REv: From<FetcherRequest<I, T>> + From<NetworkInfoRequest<I>>,
+    I: Eq + Debug + Clone + Send + 'static,
+{
+    loop {
```
xcthulhu, line 56: @Fraser999 As discussed, this loops forever

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+            .await
+        {
+            Ok(FetchedData::FromStorage { item }) => return Ok(Some(item)),
+            Ok(FetchedData::FromPeer { item, .. }) => {
+                // TODO: Validate item, ban peer if necessary
+                // Compute the total weight of the validators
+                let _total_weight: U512 = trusted_validator_weights
+                    .iter()
+                    .map(|(_, weight)| *weight)
+                    .sum();
```
xcthulhu, line 117: @afck @fizyk20 @momipsl I started to implement finality signature checking here; should I be doing `1 - ftt` to ensure >66%

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+    // The era supervisor may slash up to 3*delay + 1 eras back
+    let historical_eras_needed = delay.saturating_mul(3).saturating_add(1);
+
+    match maybe_last_emergency_restart_era_id {
+        Some(last_emergency_restart_era_id)
+            if last_emergency_restart_era_id
+                > most_recent_block_header
+                    .era_id()
+                    .saturating_sub(historical_eras_needed) =>
+        {
```
xcthulhu, line 259: @Fraser999 @fizyk20 I'm not sure if the logic below does the right thing in this corner case

node/src/components/storage.rs
```diff
@@ -653,7 +653,8 @@ impl Storage {
                 let finality_signatures =
                     match self.get_finality_signatures(&mut txn, &block_hash)? {
                         Some(signatures) => signatures,
-                        None => BlockSignatures::new(block_hash, block.header().era_id()),
```
xcthulhu, line 656: @darthsiroftardis @marc-casperlabs If there are no block signatures, this should not return success...

node/src/components/storage.rs
```diff
@@ -704,7 +706,8 @@ impl Storage {
                 let hash = block.hash();
                 let finality_signatures = match self.get_finality_signatures(&mut txn, hash)? {
                     Some(signatures) => signatures,
-                    None => BlockSignatures::new(*hash, block.header().era_id()),
```
xcthulhu, line 707: @darthsiroftardis @marc-casperlabs Ditto here.  This should be merged upstream...

node/src/testing/multi_stage_test_reactor/test_chain.rs
```diff
@@ -328,17 +328,25 @@ async fn test_joiner() {
         "There should be just one bonded validator in the network"
     );
 
-    // Get the first switch block hash
-    let first_switch_block_hash = get_switch_block_hash(1, &mut chain.network, &mut rng).await;
+    // Get the second switch block hash
+    // As part of the fast sync process, we will need to retrieve the first switch block
+    let switch_block_hash = get_switch_block_hash(2, &mut chain.network, &mut rng).await;
```
xcthulhu, line 333: @afck This needs to start with the second era now.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+            *current_walk_back_header.parent_hash(),
+        )
+        .await?;
+    };
+
+    // Get the most recent header which has the same version as ours
+    // We keep fetching by height until none of our peers have a block at that height
+    let mut most_recent_block_header = *trusted_header;
+    loop {
+        // If we encounter a block header of a version which is newer than ours we return an error
```
afck, line 203: Will this result in a shutdown with exit code 0, as in [CEP 37](https://github.com/casper-network/ceps/pull/37)?

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+                }
+            }
+            // If we could not fetch, we can stop if the most recent has our protocol version
+            None if most_recent_block_header.protocol_version()
+                == chainspec.protocol_config.version =>
+            {
+                break
+            }
+            // Otherwise keep trying to fetch until we get a block with our version
+            _ => {
```
afck, line 235:
```suggestion
            None => {
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+    // Synchronize the trie store for the most recent block header.
+
+    // Corner case: if an emergency restart happened recently, it is necessary to synchronize the
+    // state for the block right after the emergency restart.  This is needed for the EraSupervisor.
+
+    // The era supervisor needs validator information from previous eras it may potentially slash.
+    // The number of previous eras is determined by a *delay* in which consensus participants become
+    // bonded validators or unbond.
+    let delay = era_supervisor::bonded_eras(&(&chainspec).into());
+    // The era supervisor may slash up to 3*delay + 1 eras back
```
afck, line 250: It's more complicated; it's not strictly true that it can slash/evict someone for an equivocation further back than just a single delay.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+                    effect_builder,
+                    *most_recent_block_header.parent_hash(),
+                )
+                .await?;
+                if previous_block_header.era_id() == last_emergency_restart_era_id
+                    && previous_block_header.is_switch_block()
+                {
+                    break;
+                }
+                most_recent_block_header = *previous_block_header;
```
afck, line 272: Should we really overwrite `most_recent_block_header`?
I think we need the full global state for both `previous_block_header` and `most_recent_block_header`?

node/src/testing/multi_stage_test_reactor/test_chain.rs
```diff
@@ -349,7 +357,7 @@ async fn test_joiner() {
         "There should be two validators in the network (one bonded and one read only)"
     );
 
-    let era_num = 3;
+    let era_num = 6;
```
afck, line 360: Why not 4?

node/src/components/consensus.rs
```diff
@@ -5,7 +5,7 @@
 mod cl_context;
 mod config;
 mod consensus_protocol;
-mod era_supervisor;
+pub(crate) mod era_supervisor;
```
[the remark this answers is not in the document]
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1338/commits/10ee32ede625af1f4cbfaa3f7296991d4fe8602c

node/src/reactor/joiner.rs
```diff
@@ -872,24 +939,39 @@ impl reactor::Reactor for Reactor {
...
+                Event::BlockHeaderByHeightFetcher(request.into()),
+            ),
         }
     }
 
     fn maybe_exit(&self) -> Option<ReactorExit> {
         if self.linear_chain_sync.stopped_for_upgrade() {
-            Some(ReactorExit::ProcessShouldExit(ExitCode::Success))
+            Some(ReactorExit::ProcessShouldExit(
+                crate::types::ExitCode::Success,
```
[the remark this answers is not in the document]
xcthulhu: Fixed in https://github.com/CasperLabs/casper-node/pull/1338/commits/2fcb270e6161665e76eb13a76a5efe2e1f07fcae

node/src/components/contract_runtime.rs
xcthulhu: Ah, you are right... moving forward these metrics are a maintenance burden, especially if they are not used.

node/src/components/linear_chain_sync.rs
xcthulhu: > So the plan is still to get to a version that supports both slow and fast sync?

Doesn't look like we can support slow sync if we change the wire format to the fetchers.  New nodes would send messages with the current payload and old nodes would not be able to parse them.

> And this is a temporary workaround to keep the linear_chain_sync component functional by just letting fast sync do the work?

We will gradually be transitioning from the slow sync process.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+    REv: From<NetworkInfoRequest<I>>,
+    I: Send + 'static,
+{
+    let mut vector_of_peers: Vec<I> = effect_builder
+        .network_peers::<I>()
+        .await
+        .into_iter()
+        .map(|(peer, _)| peer)
+        .collect();
+    let mut rng = rand::thread_rng();
```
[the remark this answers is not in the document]
xcthulhu: The current tests for this code are non-deterministic, since they are somewhat high level.

It is a bit complicated to thread it in, but if it's a show stopper I can ticket carrying it through the code.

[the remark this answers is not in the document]
afck: Can't we just call `NodeRng::new()` here?

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+    effect::{
+        requests::{ContractRuntimeRequest, FetcherRequest, NetworkInfoRequest},
+        EffectBuilder,
+    },
+    types::{BlockHash, BlockHeader, BlockHeaderWithMetadata, Chainspec, Item},
+};
+
+const TIMEOUT_DURATION: Duration = Duration::from_millis(100);
+
+/// Get the currently connected peers from the networking component and shuffle them
```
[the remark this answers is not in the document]
xcthulhu: Fixed in 1116a36e6735fbdef277ce6a984850760c203f2d

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+                    )
+                }
+                Err(error) => return Err(error),
+            }
+        }
+        tokio::time::sleep(TIMEOUT_DURATION).await
+    }
+}
+
+// fn verify_finality_signatures
```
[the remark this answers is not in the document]
xcthulhu: Fixed in 4b1aed17e0f39d7e12d96a86006808c2c49aef06

node/src/components/linear_chain_sync/operations.rs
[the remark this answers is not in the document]
afck: Thanks! It was on other functions, too, though.

node/src/types/item.rs
```diff
@@ -43,6 +43,8 @@ pub enum Tag {
     BlockHeaderByHash,
     /// A block header and its finality signatures requested by its height in the linear chain.
     BlockHeaderAndFinalitySignaturesByHeight,
+    /// A trie from the global trie merkle tree in the execution engine.
```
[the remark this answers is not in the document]
xcthulhu: Fixed in af7ad996f5f2a78e4a1f3465db37be993c683247

node/src/types.rs
```diff
@@ -5,7 +5,7 @@ mod block;
 pub mod chainspec;
 mod deploy;
 mod exit_code;
-mod item;
+pub(crate) mod item;
```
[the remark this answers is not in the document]
xcthulhu: Fixed in bd4e5e0899b06d80b2f0c99f32f811cfafd46468

node/src/reactor/joiner.rs
```diff
@@ -588,79 +632,85 @@ impl reactor::Reactor for Reactor {
...
-                        }
-                    };
-                    self.dispatch_event(effect_builder, rng, Event::BlockByHeightFetcher(event))
+                    match fetcher::Event::<BlockWithMetadata>::from_get_response_serialized_item(
+                        sender,
+                        &serialized_item,
+                    ) {
+                        Some(fetcher_event) => {
+                            self.dispatch_event(effect_builder, rng, Event::BlockByHeightFetcher(fetcher_event))
+                        } ,
```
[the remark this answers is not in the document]
xcthulhu: Fixed in 32b3dfbb8ab00773f96f63b6b5fb34071fdad129

node/src/effect/requests.rs
```diff
@@ -270,6 +270,14 @@ pub enum StorageRequest {
         /// local storage.
         responder: Responder<Option<BlockHeader>>,
     },
+    /// Retrieve block header with metadata by height.
+    GetBlockHeaderAndMetadataByHeight {
+        /// Hash of block to get header of.
```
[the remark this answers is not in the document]
xcthulhu: Fixed in fa00c5fa3282ce1f6297f7b29cf386259e3e251a

node/src/effect/requests.rs
```diff
@@ -270,6 +270,14 @@ pub enum StorageRequest {
...
+        block_height: u64,
+        /// Responder to call with the result.  Returns `None` is the block header doesn't exist in
```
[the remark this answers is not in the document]
xcthulhu: Fixed in b5700b92d15ce3566b6a1d7d152452e9e72a4735

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Yes!  The error here is ultimately handled as a `fatal` effect in `linear_chain_sync.rs`

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,286 @@
...
+                    .saturating_sub(historical_eras_needed) =>
+        {
+            // Walk backwards until we have the first block after the emergency upgrade
+            loop {
+                let previous_block_header = fetch_retry_forever::<BlockHeader, REv, I>(
+                    effect_builder,
+                    *most_recent_block_header.parent_hash(),
+                )
+                .await?;
+                if previous_block_header.era_id() == last_emergency_restart_era_id
```
xcthulhu, line 265: As per discussion with @afck , this should be `last_emergency_restart_era_id - 1`

node/src/components/linear_chain_sync/operations.rs
xcthulhu: If we are near an emergency upgrade, we want to treat the block right after the upgrade as the `most_recent_block_header`.  After syncing that we will execute the deploys for the blocks after this point.

node/src/components/linear_chain_sync/operations.rs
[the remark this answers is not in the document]
xcthulhu: Fixed in c379588f655a74483cdc8dac7f5449380d5f038d

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -0,0 +1,288 @@
...
+where
+    REv: From<ContractRuntimeRequest>
+        + From<FetcherRequest<I, Trie<Key, StoredValue>>>
+        + From<NetworkInfoRequest<I>>,
+    I: Eq + Clone + Debug + Send + 'static,
+{
+    let trie =
+        fetch_retry_forever::<Trie<Key, StoredValue>, REv, I>(effect_builder, trie_key).await?;
+    let outstanding_tries = effect_builder
+        .put_trie_and_find_missing_descendant_trie_keys(trie)
```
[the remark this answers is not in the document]
xcthulhu: We don't mark incomplete ones, no.  We should mark incomplete block headers in storage upon syncing a trie, and mark them complete afterwards.  I'll make a follow-up story for that.

node/src/components/linear_chain_sync/operations.rs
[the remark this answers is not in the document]
xcthulhu: Hmm... I've made further changes in 05779ea0b5811e3ee4159ac0f91718506ea3a468

node/src/components/linear_chain_sync/operations.rs
xcthulhu: I've updated the comment in 20d5d940d979fdb2c6f86644523c817c56191994

afck (approved):

node/src/components/storage.rs
xcthulhu: This needs further review.  This logic is used by the RPC, where this is not the correct behavior. However, a _fetcher_ should refuse to return a payload which does not finality signatures.

xcthulhu:
This is already merged into `feat-fast-sync`, closing,

PR #1339, opened 2021-04-19
NDRS-1173: revert to messagepack for framed transport

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1173

This PR reverts small_network to use messagepack 0.14.4 for framing network messages, in order to remain compatible with the current running mainnet nodes.

The new `FramedTransport` type is compatible with the [`impl futures_core::Stream for tokio_serde::SymmetricallyFramed`](https://github.com/carllerche/tokio-serde/blob/v0.6.1/src/lib.rs#L255-L274) and [`impl futures_sink::Sink for tokio_serde::SymmetricallyFramed`](https://github.com/carllerche/tokio-serde/blob/v0.6.1/src/lib.rs#L276-L305), and this has been tested via running an upgrade test using nctl, and seeing that a new node is able to join an upgraded network (meaning that nodes running `release-1.0.0` code are able to communicate with nodes running the code in this PR).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -805,6 +805,7 @@ dependencies = [
  "base64",
  "bincode",
  "blake2",
+ "bytes 1.0.1",
  "casper-execution-engine",
  "casper-node-macros",
  "casper-types",
@@ -818,7 +819,9 @@ dependencies = [
  "fake_instant",
  "fs2",
  "futures",
+ "futures-core",
  "futures-io",
+ "futures-sink",
  "getrandom 0.2.2",
  "hex",
  "hex-buffer-serde",
@@ -846,6 +849,7 @@ dependencies = [
  "openssl",
  "parking_lot",
  "pem",
+ "pin-project 1.0.6",
  "pnet",
  "prometheus",
  "proptest",
@@ -874,7 +878,6 @@ dependencies = [
  "thiserror",
  "tokio 1.4.0",
  "tokio-openssl",
- "tokio-serde",
  "tokio-stream",
  "tokio-util 0.6.5",
  "toml",
@@ -1663,18 +1666,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "educe"
-version = "0.4.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "54ed56329d95e524ef98177ad672881bdfe7f22f254eb6ae80deb6fdd2ab20c4"
-dependencies = [
- "enum-ordinalize",
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "ee-1071-regression"
 version = "0.1.0"
 dependencies = [
@@ -1899,19 +1890,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "enum-ordinalize"
-version = "3.1.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2d52ff39419d3e16961ecfb9e32f5042bdaacf9a4cc553d2d688057117bae49b"
-dependencies = [
- "num-bigint 0.3.2",
- "num-traits",
- "proc-macro2",
- "quote",
- "syn",
-]
-
-[[package]]
 name = "env_logger"
 version = "0.8.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2118,9 +2096,9 @@ dependencies = [
 
 [[package]]
 name = "futures-core"
-version = "0.3.13"
+version = "0.3.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15496a72fabf0e62bdc3df11a59a3787429221dd0710ba8ef163d6f7a9112c94"
+checksum = "098cd1c6dda6ca01650f1a37a794245eb73181d0d4d4e955e2f3c37db7af1815"
 
 [[package]]
 name = "futures-executor"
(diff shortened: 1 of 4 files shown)


marc-casperlabs:
The change seems to be a little bigger than required, at least from my point of view.

# Problem statement

Rephrasing the problem, the situation we are in (as far as I can tell is) that mainnet (`release-1.0.0`) currently uses message pack with the [following instantiation](https://github.com/CasperLabs/casper-node/blob/de259adc5c86635bf80f6a8004d39bd0fde15d8e/node/src/components/small_network.rs#L1297-L1302):

```
/// A framed transport for `Message`s.
type FramedTransport<P> = SymmetricallyFramed<
    Framed<Transport, LengthDelimitedCodec>,
    Message<P>,
    SymmetricalMessagePack<Message<P>>,
>;
```

In this version, the lockfile also [gives a version](https://github.com/CasperLabs/casper-node/blob/de259adc5c86635bf80f6a8004d39bd0fde15d8e/Cargo.lock#L4646-L4655) of `rmp-serde` at `0.14.4`.

As we know, the `rmp-serde` versions >= `0.15` change the `is_human_readable` flag's default value, which breaks communication, as things like `SocketAddr` will serialize differently (by the way, this is [demonstrated in the handshake tests](https://github.com/CasperLabs/casper-node/blob/458842132823d04779ad7c9747f3ae80fcfc0821/node/src/components/small_network/message.rs#L155-L202), which test that message pack serialization works even against preserved binary snapshots of 1.0.0 handshakes. The only reason this test is currently passing after the bincode change is that it does not use `Transport`, but instead instantiates directly).

As state above, the [current `dev` version](https://github.com/CasperLabs/casper-node/blob/458842132823d04779ad7c9747f3ae80fcfc0821/node/src/components/small_network.rs#L1319-L1324) currently uses `bincode`, which is entirely incompatible:

```
/// A framed transport for `Message`s.
type FramedTransport<P> = SymmetricallyFramed<
    Framed<Transport, LengthDelimitedCodec>,
    Message<P>,
    SymmetricalBincode<Message<P>>,
>;
```

Just changing it back to `SymmetricalMessagePack` _may_ be problematic, since it there is the 0.14/0.15 `is_human_readable` issue to consider.

# Current solution

In 5fdcfb5634366e2dc786e063713d0778c3c36d86, it throws away the entire transport. This has some consequences, for one it won't merge against `dev` in a few minutes with the merge of #1316, which changes this to

```
type FramedTransport<P> = SymmetricallyFramed<
    Framed<Transport, LengthDelimitedCodec>,
    Message<P>,
    CountingFormat<SymmetricalBincode<Message<P>>>,
>;
```

The counting formatter hooks directly into the existing `SymetricallyFramed` instantiation, at the message encoding level.

This defeats the entire purpose of using the modular crate and copies over some really non-trivial manual async code from the crate itself (see `framed_transport.rs`).

So overall, I am not a fan of this approach.

# Alternative solutions

There are some variants that I like better: Change only the parts that are actually relevant, _i.e._ the serializer. The [`tokio_serde::messagepack::MessagePack`](https://github.com/carllerche/tokio-serde/blob/cbd4a0e4a5ee4b0090c493954af496b6201e4c85/src/lib.rs#L460-L492) is only 30 lines, and [there is an example of how to write one in our own codebase](https://github.com/CasperLabs/casper-node/blob/c4fb335da2a50983769d2ba0ae70e2c57306a5f6/node/src/components/small_network/counting_format.rs).

## Replace the serializer with rmp-serde 0.14

Creating our own serializer/deserializer, even just copy&pasting (remember to copy the MIT license header) it, would instantly allow us to use `rmp-serde` instead. Should be quite simple and a minimal change.

## Replace the serializer with configurable `rmp-serde`.

Even better would be to change the one thing about the `tokio_serde` `MessagePack` serializer that is actually the problem: No configurability. `rmp-serde` works (IIRC) even at version `0.15` if the correct configuration is used. The only issue of the `tokio_serde` implementation is thus that it uses `rmp-serde`'s the top level methods, which in turn do not allow configuration, and thus do not allow toggling the [`HumanReadableConfig`](https://docs.rs/rmp-serde/0.15.4/rmp_serde/config/struct.HumanReadableConfig.html).

For bonus points, we could actually do both, use 0.14 and be forward-compatible by setting the configuration option.

## Just lock the version

I noticed that the assumption that we are forced into a newer message pack version is not true! Looking at the [`Cargo.lock` file in `dev`](https://github.com/CasperLabs/casper-node/blob/458842132823d04779ad7c9747f3ae80fcfc0821/Cargo.lock#L4786-L4795), it is still set at `0.14.4`:

```
[[package]]
name = "rmp-serde"
version = "0.14.4"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "4ce7d70c926fe472aed493b902010bccc17fa9f7284145cb8772fd22fdb052d8"
dependencies = [
 "byteorder",
 "rmp",
 "serde",
]
```

The [version given in `tokio-serde`](https://github.com/carllerche/tokio-serde/blob/cbd4a0e4a5ee4b0090c493954af496b6201e4c85/Cargo.toml#L33) is still at `0.15`:

```
rmp-serde = { version = "0.15", optional = true }
```

While this _should_ not work, it is worth a try to see what happens. I remember wondering why it was not breaking when I messed around with this due to the handshake, but likely it was because I was already working against a bincode version.

Fraser999:
Closed in favour of incoming PR based off https://github.com/marc-casperlabs/casper-node/tree/mbrfraser/fix-rmp-serde-version.

PR #1343, opened 2021-04-20
NDRS-1175: Minor fixes in block proposer and era supervisor.

afck:
This removes some code left over from when the block proposer state used to be persisted. It also fixes an issue in the era supervisor where `next_block_height` could _decrease_ (only temporarily, though), and increases log levels for the messages related to the block proposer queues.

(Related to [NDRS-1175](https://casperlabs.atlassian.net/browse/NDRS-1175).)

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -1009,6 +1009,10 @@ where
                 equivocators,
                 proposer,
             }) => {
+                if era_id != self.era_supervisor.current_era {
+                    debug!(era = era_id.value(), "finalized block in old era");
+                    return Effects::new();
+                }
                 let era = self.era_supervisor.active_eras.get_mut(&era_id).unwrap();
                 era.add_accusations(&equivocators);
                 era.add_accusations(value.accusations());
@@ -1042,7 +1046,10 @@ where
                     .effect_builder
                     .announce_finalized_block(finalized_block.clone())
                     .ignore();
-                self.era_supervisor.next_block_height = finalized_block.height() + 1;
+                self.era_supervisor.next_block_height = self
+                    .era_supervisor
+                    .next_block_height
+                    .max(finalized_block.height() + 1);
                 if finalized_block.era_report().is_some() {
                     // This was the era's last block. Schedule deactivating this era.
                     let delay = Timestamp::now().saturating_diff(timestamp).into();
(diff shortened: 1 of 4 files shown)


fizyk20 (approved):
Good catch!

dwerner (approved):

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1009,6 +1009,10 @@ where
                 equivocators,
                 proposer,
             }) => {
+                if era_id != self.era_supervisor.current_era {
+                    debug!(era = era_id.value(), "finalized block in old era");
```
goral09, line 1013: Shouldn't this be a warn or err? After all, this should never happen and it would be a bug.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1042,7 +1046,10 @@ where
                     .effect_builder
                     .announce_finalized_block(finalized_block.clone())
                     .ignore();
-                self.era_supervisor.next_block_height = finalized_block.height() + 1;
+                self.era_supervisor.next_block_height = self
+                    .era_supervisor
+                    .next_block_height
+                    .max(finalized_block.height() + 1);
```
goral09, line 1052: If the `max` version was to return a different result from the old one, then it would have meant that we finalize blocks out of order or re-finalize previously finalized blocks. This also sounds like a bug to me.

node/src/components/consensus/era_supervisor.rs
afck: No, it can definitely happen: If we just joined, and initialized eras 1, 2 and 3, we'll simultaneously download their protocol states, and see blocks finalized in all of them.

It's just that we can safely ignore that, because if we already initialized era 3 that means we have all of era 2's blocks in the linear chain.

node/src/components/consensus/era_supervisor.rs
afck: It can happen if we join in the middle of an era: `next_block_height` gets initialized based on the latest linear chain block, but then we synchronize the protocol state from the beginning of that era, and will see all its blocks finalized again.

I was actually considering initializing it with 0 instead, but I'm not sure that's a good idea.

node/src/components/consensus/era_supervisor.rs
goral09: Ah, of course. Thanks.

goral09 (approved):

afck:
bors merge

PR #1344, opened 2021-04-20
NDRS-1175: Make sure that validator has access to the correct latest block

goral09:
Fixes a bug where a restarted node might end up with an incorrectly seeded block proposer.

https://casperlabs.atlassian.net/browse/NDRS-1175


afck (approved):

fizyk20 (approved):

goral09:
bors merge

PR #1345, opened 2021-04-20
NDRS-1186: Fix a bug where finality signatures received after storing a block were not gossiped or stored.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-1186

### DIFF
--- a/node/src/components/linear_chain/state.rs
+++ b/node/src/components/linear_chain/state.rs
@@ -92,9 +92,6 @@ impl LinearChain {
     // were waiting for that block.
     fn new_block(&mut self, block: &Block) -> Vec<Signature> {
         let signatures = self.collect_pending_finality_signatures(block.hash());
-        if signatures.is_empty() {
-            return vec![];
-        }
         let mut block_signatures = BlockSignatures::new(*block.hash(), block.header().era_id());
         for sig in signatures.iter() {
             block_signatures.insert_proof(sig.public_key(), sig.signature());
@@ -267,6 +264,8 @@ impl LinearChain {
         match self.get_signatures(&block_hash) {
             // Not found in the cache, look in the storage.
             None => vec![Outcome::LoadSignatures(fs)],
+            // We know about the block but we haven't seen any signatures for it yet.
+            Some(signatures) if signatures.proofs.is_empty() => vec![Outcome::LoadSignatures(fs)],
             Some(signatures) => self.handle_cached_signatures(Some(Box::new(signatures)), fs),
         }
     }
--- a/node/src/components/linear_chain/state.rs
+++ b/node/src/components/linear_chain/state.rs
@@ -637,4 +637,48 @@ mod tests {
         let outcomes = lc.handle_finality_signature(Box::new(invalid_sig), false);
         assert!(outcomes.is_empty())
     }
+
+    #[test]
+    fn new_block_then_own_sig() {
+        let _ = logging::init();
+        let mut rng = TestRng::new();
+        let protocol_version = ProtocolVersion::V1_0_0;
+        let auction_delay = 1;
+        let unbonding_delay = 2;
+        let mut lc = LinearChain::new(protocol_version, auction_delay, unbonding_delay);
+        // Set the latest known block so that we can trigger the following checks.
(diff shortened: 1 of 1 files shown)


node/src/components/linear_chain/state.rs
```diff
@@ -267,6 +264,8 @@ impl LinearChain {
         match self.get_signatures(&block_hash) {
             // Not found in the cache, look in the storage.
             None => vec![Outcome::LoadSignatures(fs)],
+            // We know about the block but we haven't seen any signatures for it yet.
+            Some(signatures) if signatures.proofs.is_empty() => vec![Outcome::LoadSignatures(fs)],
```
goral09, line 268: This is not necessarily required but it would load the signatures from storage into the in-memory cache if the cache was pruned in the meantime.

Fraser999 (approved):

node/src/components/linear_chain/state.rs
```diff
@@ -92,9 +92,6 @@ impl LinearChain {
     // were waiting for that block.
     fn new_block(&mut self, block: &Block) -> Vec<Signature> {
         let signatures = self.collect_pending_finality_signatures(block.hash());
-        if signatures.is_empty() {
-            return vec![];
-        }
```
afck, line 97: Trying to understand this fix…
So this would now put an empty `BlockSignatures` into the cache. Where does that make a difference? (In `handle_finality_signature`, at least, it seems like it doesn't.)

goral09: Yeah, the trick is that now we're caching empty `BlockSignatures` which later in `handle_is_bonded` will return `Some(_)` from `or_else(|| self.get_signatures)` method – which means that finality signature is for an already known block it's just that we don't have any signatures for it yet.

afck: It would be good to document that somewhere, maybe on the `signature_cache` field.

afck (approved):

node/src/components/linear_chain/state.rs
```diff
@@ -637,4 +636,48 @@ mod tests {
...
+        let block = Block::random_with_specifics(&mut rng, EraId::new(3), 10, false);
+        let block_hash = *block.hash();
+        let block_era = block.header().era_id();
+        let put_block_outcomes = lc.handle_new_block(Box::new(block.clone()), HashMap::new());
+        let expected_outcomes = {
+            let mut tmp = vec![];
+            tmp.push(Outcome::StoreBlock(Box::new(block)));
+            tmp.push(Outcome::StoreExecutionResults(block_hash, HashMap::new()));
+            tmp
+        };
```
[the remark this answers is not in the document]
goral09: Done [`2d855cf` (#1345)](https://github.com/CasperLabs/casper-node/pull/1345/commits/2d855cf6575f55b83a1ad924c8a78ad1dc1221ba)

node/src/components/linear_chain/state.rs
```diff
@@ -637,4 +636,48 @@ mod tests {
...
+        assert!(!outcomes.is_empty(), "{:?} should not be empty", outcomes);
+        let expected_outcomes = {
+            let mut tmp = vec![];
+            let mut block_signatures = BlockSignatures::new(block_hash, block_era);
+            block_signatures.insert_proof(valid_sig.public_key.clone(), valid_sig.signature);
+            tmp.push(Outcome::StoreBlockSignatures(block_signatures));
+            tmp.push(Outcome::Gossip(Box::new(valid_sig.clone())));
+            tmp.push(Outcome::AnnounceSignature(Box::new(valid_sig)));
+            tmp
+        };
```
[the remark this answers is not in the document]
goral09: Done [`2d855cf` (#1345)](https://github.com/CasperLabs/casper-node/pull/1345/commits/2d855cf6575f55b83a1ad924c8a78ad1dc1221ba)

node/src/components/linear_chain/state.rs
goral09: Added in [`4b3ceea` (#1345)](https://github.com/CasperLabs/casper-node/pull/1345/commits/4b3ceea288596405a5977548ce5353b797c626a3).

goral09:
bors merge

PR #1347, opened 2021-04-21
NDRS-1159: Add more consensus and block proposer log messages.

afck:
* Log who missed their slot as a leader.
* List the individual steps of the rewards calculation: Which block got which quorum and who got rewards.
* Log when the block proposer queues a request due to the `next_finalized` value.

https://casperlabs.atlassian.net/browse/NDRS-1159

### DIFF
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -209,19 +209,17 @@ impl<C: Context> ActiveValidator<C> {
                 return effects;
             } else if timestamp == r_id + self.witness_offset(r_len) {
                 let panorama = self.panorama_at(state, timestamp);
-                if panorama.has_correct() {
-                    if let Some(witness_unit) =
-                        self.new_unit(panorama, timestamp, None, state, instance_id)
+                if let Some(witness_unit) =
+                    self.new_unit(panorama, timestamp, None, state, instance_id)
+                {
+                    if self
+                        .latest_unit(state)
+                        .map_or(true, |latest_unit| latest_unit.round_id() != r_id)
                     {
-                        if self
-                            .latest_unit(state)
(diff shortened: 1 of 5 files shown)


node/src/components/consensus/highway_core/active_validator.rs
```diff
@@ -614,6 +611,40 @@ impl<C: Context> ActiveValidator<C> {
     pub(crate) fn next_round_length(&self) -> TimeDiff {
         state::round_len(self.next_round_exp)
     }
+
+    /// Logs a message if the witness unit is in a round without a proposal.
+    fn log_if_missing_proposal(
```
goral09, line 616: This will only fire if node is an active validator, is there a way to have similar functionality for read-only nodes?

afck: That's a good point.
We'd need another timer for that, then. But a read-only node doesn't even have its own round length — so would we always assume minimum round length and log every round without a proposal? With our current configuration that would probably make sense.

goral09: I was rather thinking about inspecting the panorama of a unit and checking if it's in a round that doesn't have a proposal. I _think_ it should be doable:
* if fork choice would return a different block than that round's leader block, we can check if there was a leader block in the panorama
* if fork choice returns a block from the previous round and there's no new leader block in the panorama, then it would have meant that the rounds misses the leader's block.

afck: Yes, but we don't want to have 100 lines of lots for each missed round, i.e. we don't want to log it for _every_ incoming witness message in that round.

afck:
Closing this for now. I'll remove one commit, rebase, and re-implement logging of missing proposals.

PR #1348, opened 2021-04-21
NDRS-1188: Metrics race condition cleanup

marc-casperlabs:
This fixes a race condition where new metrics were being replaced before the networking component had shut down completely, resulting in a panic. As a fix `small_network` now drops its own strong reference first and waits for any stragglers.


goral09 (approved):
Looks good although I am a bit worried that this still may fail to drop all the references in time, leading to a node crash. "Fortunately", that could only happen during the transition between different reactors so won't cause any data/state corruption or lead to an equivocation.

PR #1351, opened 2021-04-22
NDRS-1190: Make sure an era is not activated twice without a unit hash file.

afck:
* Don't activate old eras on initialization: When the era supervisor starts, the validator should only actively participate in the latest era.
* Delete the unit hash file only once the era becomes obsolete.

https://casperlabs.atlassian.net/browse/NDRS-1190

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -295,8 +295,6 @@ where
         if self.active_eras.contains_key(&era_id) {
             panic!("{} already exists", era_id);
         }
-        self.current_era = era_id;
-        self.metrics.current_era.set(era_id.value() as i64);
         let instance_id = instance_id(&self.protocol_config, era_id);
 
         info!(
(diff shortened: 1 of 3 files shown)


fizyk20 (approved):
Nice :+1:

node/src/components/consensus/era_supervisor.rs
```diff
@@ -312,13 +311,24 @@ where
...
         // ongoing based on its minimum duration, and we are one of the validators.
         let our_id = &self.public_signing_key;
-        let should_activate = if !validators.contains_key(&our_id) {
+        let should_activate = if self.current_era > era_id {
+            trace!(
+                era = era_id.value(),
+                current_era = self.current_era.value(),
+                "not voting; initializing past era"
+            );
+            false
```
goral09, line 320: I thought eras are always initialized in order – from the oldest one to the latest one – so `self.current_era` should never be bigger than `era_id`.

goral09 (approved):

node/src/components/consensus/era_supervisor.rs
afck: But on initialization we set `current_era`, before we initialize the past eras: https://github.com/CasperLabs/casper-node/blob/42ec609300d6bf934fe86fdeed5a0b64c7c967e0/node/src/reactor/validator.rs#L469

afck:
bors merge

node/src/components/consensus/era_supervisor.rs
goral09: Thanks, I did not remember that.

PR #1352, opened 2021-04-22
NDRS-1159: Add more consensus and block proposer log messages. 

afck:
* Log who missed their slot as a leader.
* List the individual steps of the rewards calculation: Which block got which quorum and who got rewards.
* Log when the block proposer queues a request due to the `next_finalized` value.

https://casperlabs.atlassian.net/browse/NDRS-1159

### DIFF
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -197,19 +197,17 @@ impl<C: Context> ActiveValidator<C> {
                 return effects;
             } else if timestamp == r_id + self.witness_offset(r_len) {
                 let panorama = self.panorama_at(state, timestamp);
-                if panorama.has_correct() {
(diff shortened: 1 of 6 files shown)


goral09 (approved):
Nice one.

node/src/components/consensus/highway_core/active_validator.rs
```diff
@@ -416,6 +414,9 @@ impl<C: Context> ActiveValidator<C> {
         state: &State<C>,
         instance_id: C::InstanceId,
     ) -> Option<SignedWireUnit<C>> {
+        if value.is_none() && !panorama.has_correct() {
+            return None; // Wait for the first proposal before creating a unit without a value.
+        }
```
goral09, line 419: Is there a reason to do this that is inherent to this PR or is it just for code readability?

fizyk20 (approved):

PR #1354, opened 2021-04-23
NO-TICKET: Fix liveness_test_some_mute.

afck:
When we introduced the `TemporarilyMute` variant, we accidentally made `PermanentlyMute` validators not mute at all! This re-enables simulation of permanently mute validators (i.e. offline/crashed nodes) in the Highway DES tests.

### DIFF
--- a/node/src/components/consensus/highway_core/highway_testing.rs
+++ b/node/src/components/consensus/highway_core/highway_testing.rs
@@ -235,8 +235,7 @@ impl HighwayValidator {
                 }
             }
             Some(DesFault::PermanentlyMute) => {
-                // For mute validators we add it to the state but not gossip, if the delivery time
-                // is in the interval in which they are muted.
+                // For mute validators we add it to the state but not gossip.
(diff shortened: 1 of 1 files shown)


node/src/components/consensus/highway_core/highway_testing.rs
```diff
@@ -234,7 +234,21 @@ impl HighwayValidator {
                     }
                 }
             }
-            None | Some(DesFault::TemporarilyMute { .. }) | Some(DesFault::PermanentlyMute) => {
+            Some(DesFault::PermanentlyMute) => {
+                // For mute validators we add it to the state but not gossip, if the delivery time
```
goral09, line 238: The "if …" part of the comment does not apply to the `PermanentlyMute` variant.

goral09 (approved):
Thanks for catching this.

afck:
bors merge

xcthulhu (approved):

PR #1357, opened 2021-04-23
NDRS-1197: Fix flaky storage tests.

goral09:
https://casperlabs.atlassian.net/browse/NDRS-1197

Fixes random block generators used in the storage tests so that they don't lead to flaky tests.

### DIFF
--- a/node/src/components/storage/tests.rs
+++ b/node/src/components/storage/tests.rs
@@ -574,8 +574,18 @@ fn different_block_at_height_is_fatal() {
     let mut storage = storage_fixture(&harness);
 
     // Create two different blocks at the same height.
-    let block_44_a = random_block_at_height(&mut harness.rng, 44);
-    let block_44_b = random_block_at_height(&mut harness.rng, 44);
+    let block_44_a = Box::new(Block::random_with_specifics(
+        &mut harness.rng,
+        EraId::new(1),
+        44,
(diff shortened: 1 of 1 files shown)


node/src/components/storage/tests.rs
```diff
@@ -559,8 +574,18 @@ fn different_block_at_height_is_fatal() {
...
+        EraId::new(1),
+        44,
+        false,
+    ));
+    let block_44_b = Box::new(Block::random_with_specifics(
+        &mut harness.rng,
+        EraId::new(1),
+        44,
+        false,
+    ));
```
afck, line 588: Should we add a similar `should_panic` test for two switch blocks in the same era (but possibly different height)?

afck (approved):

node/src/components/storage/tests.rs
goral09: In a separate PR?

goral09:
bors merge

node/src/components/storage/tests.rs
afck: (I'm not sure if it's needed, but I created [NDRS-1198](https://casperlabs.atlassian.net/browse/NDRS-1198) anyway.)

xcthulhu (approved):

PR #1358, opened 2021-04-23
NDRS-1194: OpenRPC schema REST endpoint

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-1194

CHANGELOG:

- Added an `/rpc-schema` REST endpoint which gives a OpenRPC compatible schema

### DIFF
--- a/node/src/components/rest_server/filters.rs
+++ b/node/src/components/rest_server/filters.rs
@@ -26,7 +26,7 @@ pub const STATUS_API_PATH: &str = "status";
 pub const METRICS_API_PATH: &str = "metrics";
 
 /// The OpenRPC scehma URL path.
-pub const OPEN_RPC_API_PATH: &str = "openrpc";
+pub const JSON_RPC_SCHEMA_API_PATH: &str = "rpc-schema";
 
 pub(super) fn create_status_filter<REv: ReactorEventT>(
     effect_builder: EffectBuilder<REv>,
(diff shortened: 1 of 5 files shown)


node/src/components/rest_server/filters.rs
```diff
@@ -25,6 +25,9 @@ pub const STATUS_API_PATH: &str = "status";
 /// The metrics URL path.
 pub const METRICS_API_PATH: &str = "metrics";
 
+/// The OpenRPC scehma URL path.
+pub const OPEN_RPC_API_PATH: &str = "openrpc";
```
Fraser999, line 29: I think we should be leaning towards emphasising that this returns a schema rather than focusing on the format of the returned schema.  I guess I'd go for something more like
```
pub const JSON_RPC_SCHEMA_API_PATH: &str = "rpc-schema";
```

node/src/components/rest_server/filters.rs
```diff
@@ -72,3 +75,21 @@ pub(super) fn create_metrics_filter<REv: ReactorEventT>(
         })
         .boxed()
 }
+
+pub(super) fn create_rpc_filter<REv: ReactorEventT>(
```
Fraser999, line 79:
```suggestion
pub(super) fn create_rpc_schema_filter<REv: ReactorEventT>(
```

node/src/components/rpc_server/rpcs/docs.rs
```diff
@@ -103,7 +103,7 @@ pub trait DocExample {
 
 /// The main schema for the casper node's RPC server, compliant with https://spec.open-rpc.org.
 #[derive(Clone, Serialize, Deserialize, Debug)]
-struct OpenRpcSchema {
+pub struct OpenRpcSchema {
```
Fraser999, line 106: Do we need this to be `pub` rather than `pub(crate)`?

node/src/effect/requests.rs
```diff
@@ -697,13 +697,19 @@ pub enum RestRequest<I> {
         /// Responder to call with the result.
         responder: Responder<Option<String>>,
     },
+    /// Returns OpenRPC compatible payload.
+    GetOpenRpc {
```
Fraser999, line 701:
```suggestion
    /// Returns schema of client-facing JSON-RPCs in OpenRPC format.
    GetRpcSchema {
```

Fraser999 (commented):
Just some naming changes requested.

node/src/components/rpc_server/rpcs/docs.rs
darthsiroftardis: Yes, I tried `pub(crate)` originally

node/src/components/rest_server/filters.rs
darthsiroftardis: Used suggestion

node/src/components/rest_server/filters.rs
darthsiroftardis: Applied suggestion

Fraser999 (approved):

PR #1360, opened 2021-04-23
EE-1248: Test fixtures

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1248

This PR introduces an initial implementation of externalized test fixtures for ABI. There is also an automated test case loader that executes defined tests.

These fixtures are already used in this PR to detect a failure in `Transform` serialization routine which should also make it easier to do further refactor in EE-1246

ABI test fixture contains a JSON object where keys are the names of test cases. Each value is a dictionary with "input" (array of input values) and "output" a base16 encoded binary output of serialization of the input.

Each input value can be encoded as:

- A tuple/array of two elements - type (string) and a value as defined by a specification
- A dictionary with two keys: "type" and "value" as explained above.

```json
{
  "SerializeBool": {
    "input": [
      {
        "type": "Bool",
        "value": true
      },
      {
        "type": "Bool",
        "value": false
      }
    ],
    "output": "0100"
  },
  "SerializeString": {
    "input": [
      {
        "type": "String",
        "value": "Hello, world!"
      }
    ],
    "output": "0d00000048656c6c6f2c20776f726c6421"
  }
}
```

A repository of fixtures is a directory with subdirectories in it. Each subdirectory in the root dir is dedicated to specific test cases i.e. `$ROOT/ABI/basic.json`. A fixtures subdirectory can contain multiple json files to represent different groups of test cases.


dwerner:
bors try

henrytill (commented):
:+1: on the basic JSON format.  I'm sure we'll continue to iterate on it as the need arises, but this is a great starting point.

I haven't dug into the supporting code yet, but I will take a closer look before the EOW.

dwerner (approved):
looks good to me, but let's make sure not to merge this until after 1.3 is cut

utils/validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "casper-validation"
+version = "0.1.0"
+authors = ["Michał Papierski <michal@casperlabs.io>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
```
Fraser999:
```suggestion
```

utils/validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dependencies]
```
Fraser999, line 7: Can we sort the deps alphabetically please?

utils/validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+casper-types = { version = "1.0.0", path = "../../types", features = ["std"] }
```
Fraser999:
```suggestion
casper-types = { path = "../../types", features = ["std"] }
```

utils/validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+hex = { version = "0.4.2", features = ["serde"] }
+serde = "1"
+serde_json = "1"
+thiserror = "1.0.18"
+base16 = "0.2.1"
+clap = "3.0.0-beta.2"
+derive_more = "0.99.13"
+anyhow = "1"
+casper-execution-engine = { version = "1.0.0", path = "../../execution_engine" }
```
Fraser999:
```suggestion
casper-execution-engine = { path = "../../execution_engine" }
```

execution_engine/CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Changed
 * Update pinned version of Rust to `nightly-2021-06-17`
-
+* Added serialized test fixtures for ABI and an automated verification tool.
```
Fraser999: I'm not sure we actually need a changelog entry for this.  It mostly affects `casper_types`, but there are no real user-facing changes there.

I'm not strongly opposed to updating a changelog, but if you do want to, I think it'd be better suited to `casper_types` as many of the EE shared types are being moved there.

utils/validation/src/abi.rs
```diff
@@ -0,0 +1,149 @@
+use std::collections::BTreeMap;
+
+use casper_execution_engine::shared::stored_value::StoredValue;
+use casper_types::{
+    bytesrepr::{self, ToBytes},
+    CLValue, Key, Transform, U512,
+};
+
+use serde::{Deserialize, Serialize};
```
Fraser999: This should come before the casper imports.

utils/validation/src/abi.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    pub fn input(&self) -> &[Input] {
+        &self.input
+    }
+
+    pub fn output(&self) -> &[u8] {
+        &self.output
+    }
+
+    /// This gets executed for each test case.
```
Fraser999, line 103: The doc comment should describe what the method is doing.

utils/validation/src/error.rs
```diff
@@ -0,0 +1,19 @@
+use casper_types::bytesrepr;
+use std::{io, path::PathBuf};
+use thiserror::Error;
```
Fraser999:
```suggestion
use std::{io, path::PathBuf};

use thiserror::Error;

use casper_types::bytesrepr;
```

utils/validation/src/generators.rs
```diff
@@ -0,0 +1,379 @@
...
+    error::Error,
+    Fixture, TestFixtures,
+};
+
+pub fn make_abi_test_fixtures() -> Result<TestFixtures, Error> {
+    let basic = {
+        let mut basic = BTreeMap::new();
+        basic.insert(
+            "SerializeU8".to_string(),
+            ABITestCase::from_inputs(vec![254u8.into()])?,
```
Fraser999, line 28: Maybe having the max and min values in each case would be useful?

Also, should we include a test for `()`?

utils/validation/src/generators.rs
```diff
@@ -0,0 +1,379 @@
...
+            "SerializeU32".to_string(),
+            ABITestCase::from_inputs(vec![3490072870u32.into()])?,
+        );
+        basic.insert(
+            "SerializeU64".to_string(),
+            ABITestCase::from_inputs(vec![10829133186225377555u64.into()])?,
+        );
+        basic.insert(
+            "SerializeString".to_string(),
+            ABITestCase::from_inputs(vec!["Hello, world!".to_string().into()])?,
```
Fraser999, line 48: Also an empty string?

utils/validation/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+use abi::ABIFixture;
+use error::Error;
+
+pub const ABI_TEST_FIXTURES: &str = "ABI";
+const JSON_FILE_EXT: &str = "json";
+
+#[derive(Debug)]
+pub enum Fixture {
+    /// Name of the test fixture (taken from a file name) and ABI fixture itself.
+    ABI(String, ABIFixture),
```
Fraser999: Let's use a tuple style variant for clarity please.

utils/validation/src/main.rs
```diff
@@ -0,0 +1,58 @@
+mod generators;
+
+use std::{fs::File, io::BufWriter, path::PathBuf};
+
+use casper_validation::{Fixture, ABI_TEST_FIXTURES};
+use clap::Clap;
```
Fraser999:
```suggestion
use clap::Clap;

use casper_validation::{Fixture, ABI_TEST_FIXTURES};
```

utils/validation/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Clap)]
+enum SubCommand {
+    Generate(Generate),
+}
+
+/// Generates example test fixtures from the code.
+///
+/// Do not use with day to day development - for example to fix an error in serialization code by
+/// replacing the fixture with possibly invalid code.
```
Fraser999, line 25: This probably would be worth pointing out in bold in the README.

utils/validation/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+        Ok(())
+    }
+}
+
+fn main() -> anyhow::Result<()> {
+    let opts: Opts = Opts::parse();
+    match opts.subcmd {
+        SubCommand::Generate(generate) => generate.run()?,
+    }
+    Ok(())
```
Fraser999:
```suggestion
        SubCommand::Generate(generate) => generate.run(),
    }
```

utils/validation/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        for Fixture::ABI(file_name, fixture) in fixtures {
+            let output_path = {
+                let mut output_path = self.output.clone();
+                output_path.push(ABI_TEST_FIXTURES);
+                output_path.push(file_name + ".json");
+                output_path
+            };
+
+            let file = File::create(&output_path)?;
```
Fraser999: This fails if the `ABI_TEST_FIXTURES` subdirectory doesn't exist.

Since you're using anyhow errors here, it'd be useful to add context to the error.  I got the following error message which didn't help too much in diagnosing what I was doing wrong:
```
Error: No such file or directory (os error 2)
```

utils/validation/tests/validation_test.rs
```diff
@@ -0,0 +1,75 @@
+extern crate rustc_test as test;
```
Fraser999: Let's not alias the crate unless we have to please.

Fraser999 (commented):
Could you add a README to the new validation folder please?  It should give a summary of the pieces of the puzzle and how we use them.

Some of the files in `utils/validation/tests/fixtures/ABI` aren't generated by running the tool - how were they created, and should that be something the tool generates?

When a test fails, the output is very opaque.  E.g. when I modified a single char in one of the fixtures, the output was
```
fraser@fraser-ryzen-3700x:~/Rust/casper-node/utils/validation (ee-1248-test-fixtures *+$|MERGING=)$ cargo t -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.10s
     Running unittests (/home/fraser/Rust/casper-node/target/debug/deps/casper_validation-2ffcd560dd0f1093)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests (/home/fraser/Rust/casper-node/target/debug/deps/casper_validation-3c4bd9c8c5b806ef)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/validation_test.rs (/home/fraser/Rust/casper-node/target/debug/deps/validation_test-172c646da81e8c84)
Error: Odd number of digits at line 13 column 19
error: test failed, to rerun pass '--test validation_test'
```

I think our high-level naming is somewhat off here.  I think what we're calling test fixtures are actually test resources, or test inputs.  And the term "ABI" should be replaced by "serialization" throughout IMO.

utils/validation/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[dev-dependencies]
+rustc-test = "0.3.1"
+
```
Fraser999, line 18: Adding the following helps keep the test output to a minimum:
```

[[bin]]
name = "casper-validation"
test = false
doctest = false

[lib]
test = false
doctest = false
```

utils/validation/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+        let dir_entries = match entry.path().file_name() {
+            Some(file_name) if file_name == ABI_TEST_FIXTURES => {
+                utils::recursive_read_dir(&entry.path())?
+            }
+            None | Some(_) => continue,
+        };
+
+        for dir_entry in dir_entries {
+            let fixture = load_fixture(dir_entry.path())?;
+            let filename = path
```
Fraser999: Shouldn't this be `dir_entry.path()` rather than `path`?

mpapierski: Good catch. Fixed. Instead of 'fixtures' it will now display 'basic' (from the filename of json file)

utils/validation/src/main.rs
mpapierski: Addressed

utils/validation/tests/validation_test.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+#[derive(Clap, Debug)]
+struct Opts {
+    #[clap(long = "nocapture", parse(from_flag = ::std::ops::Not::not))]
+    capture: bool,
+    #[allow(unused_attributes)]
+    #[clap(name = "FILTERS")]
+    filters: Vec<String>,
+}
```
Fraser999: I don't think we need this (yet).

The `--nocapture` flag shouldn't stop the test runner from printing out the name and result of each test case, and the `filters` member is unused.

I'd recommend just removing this for now so we can get this PR merged, and we can then iterate on it over time to increase its functionality as required.

mpapierski: Addressed in commit 7e1a3e11e19040ceeae81a40faff90f1790abfa2 where I just simply do non-zero exit code if at least one test case failed

Fraser999 (approved):

PR #1364, opened 2021-04-26
HWY-286: Make unchecked timestamp subtraction test-only.

afck:
That way we can't use the potentially underflowing `timestamp - timediff` in production anymore, and have to use `timestamp.saturating_sub(timediff)`. A panic in production would be unlikely, since the result of the subtraction would have to be before 1970, but it's better to err on the side of caution. The problem did show up in a unit test that used timestamps close to 0 (i.e. beginning of 1970).

https://casperlabs.atlassian.net/browse/HWY-286

### DIFF
--- a/node/src/types/timestamp.rs
+++ b/node/src/types/timestamp.rs
@@ -3,7 +3,7 @@
 
 use std::{
     fmt::{self, Display, Formatter},
-    ops::{Add, AddAssign, Div, Mul, Rem, Sub},
+    ops::{Add, AddAssign, Div, Mul, Rem},
     str::FromStr,
     time::{Duration, SystemTime},
 };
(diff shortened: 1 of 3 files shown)


goral09 (approved):

afck:
bors merge

node/src/types/timestamp.rs
```diff
@@ -123,7 +123,8 @@ impl AddAssign<TimeDiff> for Timestamp {
     }
 }
 
-impl Sub<TimeDiff> for Timestamp {
+#[cfg(test)]
+impl std::ops::Sub<TimeDiff> for Timestamp {
```
marc-casperlabs, line 127: Common coding style would be to `#[cfg(test)] use std::ops::Sub`, but if @Fraser999 does not catch you, you're fine :)

afck: Too late! :stuck_out_tongue:

PR #1365, opened 2021-04-26
NDRS-1193: Introduce task manager

marc-casperlabs:
This introduces a task manager to manage background tasks that need to be shutdown and awaited. This functionality replaces the introduced `Arc` based waiting.

The task manager will wait for task completion when `finalize` is called.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -634,24 +634,16 @@ where
 
         let mut effects = self.check_connection_complete(effect_builder, peer_id);
 
-        let handshake = self.chain_info.create_handshake(self.public_address);
-
         effects.extend(
-            message_sender(
-                receiver,
-                sink,
-                self.net_metrics.queued_messages.clone(),
-                handshake,
-            )
-            .event(move |result| Event::OutgoingFailed {
-                peer_id: Box::new(Some(peer_id)),
-                peer_address: Box::new(peer_address),
-                error: Box::new(result.err().map(Into::into)),
-            }),
-        );
-        effects.extend(
-            handshake_reader(self.event_queue, stream, self.our_id, peer_id, peer_address)
-                .ignore::<Event<P>>(),
+            effect_builder
+                .immediately()
+                .event(move |_| Event::OutgoingReady {
+                    receiver: receiver,
+                    sink: Box::new(sink),
+                    stream: Box::new(stream),
+                    peer_id: Box::new(peer_id),
+                    peer_address,
+                }),
         );
 
         effects
@@ -1039,6 +1031,49 @@ where
             Event::OutgoingEstablished { peer_id, transport } => {
                 self.setup_outgoing(effect_builder, *peer_id, transport)
             }
+            Event::OutgoingReady {
+                receiver,
+                sink,
+                stream,
+                peer_id,
+                peer_address,
+            } => {
+                let handshake = self.chain_info.create_handshake(self.public_address);
+
+                // Start the sender task.
+                let event_queue = self.event_queue;
+                let counter = self.net_metrics.queued_messages.clone();
+                let peer_id = *peer_id;
+
+                self.task_manager.spawn(move |shutdown| {
+                    message_sender(
+                        event_queue,
+                        receiver,
+                        *sink,
+                        shutdown,
+                        counter,
+                        handshake,
+                        peer_id,
+                        peer_address,
+                    )
+                });
+
+                // Start the handshake reader task.
+                let our_id = self.our_id;
+
+                self.task_manager.spawn(move |shutdown| {
+                    handshake_reader(
+                        event_queue,
+                        *stream,
+                        shutdown,
+                        our_id,
+                        peer_id,
+                        peer_address,
+                    )
+                });
+
+                Effects::new()
+            }
             Event::OutgoingFailed {
                 peer_id,
                 peer_address,
@@ -1228,6 +1263,7 @@ async fn setup_tls(
 async fn handshake_reader<REv, P>(
     event_queue: EventQueueHandle<REv>,
     mut stream: SplitStream<FramedTransport<P>>,
+    mut shutdown: ShutdownReceiver,
     our_id: NodeId,
     peer_id: NodeId,
     peer_address: SocketAddr,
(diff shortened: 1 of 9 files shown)


node/src/components/small_network/task_manager.rs
```diff
@@ -0,0 +1,231 @@
...
+/// Dropping a task manager will instruct all tasks to shutdown, but the `drop` will not await their
+/// termination.
+#[derive(DataSize, Debug)]
+pub struct TaskManager {
+    /// A counter for joined tasks.
+    ///
+    /// Incremented to provide unique IDs for tasks.
+    task_counter: usize,
+    /// Join handles for all spawned tasks, mapped from task ID to join handle.
+    join_handles: Arc<Mutex<HashMap<usize, JoinHandle<()>>>>,
```
xcthulhu, line 115: I think we should avoid a mutex if we can.

node/src/components/small_network/task_manager.rs
```diff
@@ -0,0 +1,231 @@
...
+//!     task_manager.shutdown_and_wait(Duration::from_secs(10))
+//! ).expect("shutdown failed");
+//!
+//! assert_eq!(tasks_leftover, 0);
+//! ```
+//!
+//! # `Drop`
+//!
+//! Dropping the `TaskManager` will cause all tasks to be sent a shutdown notice, but the `Drop`
+//! implementation will **not** wait them to finish.
```
dwerner, line 48: Let's have a unit test that proves that `TaskManager` maintains the invariants you specify here.

node/src/components/small_network/task_manager.rs
```diff
@@ -0,0 +1,231 @@
...
+    /// If the internal log has been poisoned, outputs a warning, but otherwise keeps on going.
+    pub fn spawn<F, G>(&mut self, gen: F)
+    where
+        F: FnOnce(ShutdownReceiver) -> G,
+        G: Future<Output = ()> + Send + 'static,
+    {
+        let task_id = self.task_counter;
+        self.task_counter += 1;
+
+        let join_handles = self.join_handles.clone();
```
dwerner, line 157: If we close over a strong `Arc` in a closure/future, it can keep the item alive longer than intended. This can be a source of resource leaks. Instead, let's prefer `Weak` here, with `Arc::downgrade()`, and test with `weak_join_handles.upgrade()` in the future itself.

node/src/components/small_network/task_manager.rs
xcthulhu: If we are going to use a `Mutex`, it's more idiomatic to use a `futures::Lock::Mutex`.

node/src/components/small_network.rs
```diff
@@ -1259,116 +1263,185 @@ async fn setup_tls(
...
+                    }
+                    None => {
+                        // The receive channel was closed. We're shutting down unclean?
+                        break;
+                    }
+                }
+            }
 
-    Ok(())
+            _ = shutdown.wait_for_shutdown() => {
```
dwerner, line 1439: We're relying on tokio quite a bit when `futures` based stream-primitives would do the job.

Couldn't we do something like:
```
match queue.select(shutdown.wait_for_shutdown().into_stream()) {
    Either::Left((payload, _)) => {
...
    }, 
    Either::Right((value2, _)) => break,
...
}
```

node/src/components/small_network/task_manager.rs
```diff
@@ -0,0 +1,231 @@
...
+                ShutdownOutcome::JoinFailure(err) => {
+                    warn!(task_id, %err, "could not join task");
+                    failed += 1;
+                }
+                ShutdownOutcome::TimedOut => {
+                    error!(task_id, "task joining timed out");
+                    failed += 1;
+                }
+            }
+        }
```
xcthulhu, line 227: Couldn't we just `.map` here and then `.sum`?

dwerner (changes requested):

node/src/components/small_network/task_manager.rs
goral09: I wanted to write that as well but saw that PR was still in Draft. It's not anymore, let's add tests.

node/src/components/small_network/task_manager.rs
goral09: I actually prefer this version. It's more readable IMO and makes it more obvious what we're doing.

node/src/components/small_network/task_manager.rs
marc-casperlabs: > I think we should avoid a mutex if we can.

Agreed, although the only alternative I can see if not bounding the number of "dead" join handles, as we have no chance of cleaning them up if a task finishes. That seems a worse alternative.

I'm very much open to a better suggestion :)

> `futures::Lock::Mutex`

AFAIK there's no need to use a futures-aware mutex here, at least it's not recommended by `tokio::sync::Mutex` in this case at least: https://tokio-rs.github.io/tokio/doc/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

node/src/components/small_network/task_manager.rs
marc-casperlabs: > Couldn't we just .map here and then .sum?

It's a stream, not an iterator, so out-of-the-box, I don't think there's a `.sum` impl available? The only one I am aware of is https://docs.rs/async-std/1.9.0/async_std/stream/trait.Sum.html.

If it was available, imagine the code'd look like this

```rust
        Ok(joins
            .map(|(task_id, outcome)| match outcome {
                ShutdownOutcome::Ok => {
                    trace!(task_id, "successfully joined task");
                    0
                }
                ShutdownOutcome::JoinFailure(err) => {
                    warn!(task_id, %err, "could not join task");
                    1
                }
                ShutdownOutcome::TimedOut => {
                    error!(task_id, "task joining timed out");
                    1
                }
            })
            .sum());
```

Overall, I don't think the improvement is that great, I can certainly see both. The explicit version is a little easier to maintain/debug if things change in my mind; the pattern of mapping-for-side-effects has always felt weird to me.

marc-casperlabs: Side note: This code used to run `.for_each_concurrent`, but with the change to `FuturesUnordered` as recommend by the docs, I simplified it a bit.

node/src/components/small_network/task_manager.rs
marc-casperlabs: The only alternative I can think of is having another task with a channel that manages these things.

node/src/components/small_network/task_manager.rs
marc-casperlabs: Good point. The comment specifying the invariant was added the implementation, but I guess it's fair to test it, now that we are advertising it :)

marc-casperlabs:
Closing this for now due to concerns.

PR #1368, opened 2021-04-26
RES-419: Add more logging for Highway units and finality.

afck:
This adds a config option for logging the size of every incoming and outgoing unit, and an INFO-level log message for when finality is detected.

https://casperlabs.atlassian.net/browse/RES-419


goral09 (approved):

afck:
bors merge

PR #1371, opened 2021-04-26
NDRS-1179: Store block headers on retrieval

xcthulhu:
When retrieving a block header, we need to store them in storage.  This is because the `EraSupervisor` will use them later when determining era validators.

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -434,22 +434,21 @@ impl Storage {
         Ok(match req {
             StorageRequest::PutBlock { block, responder } => {
                 let mut txn = self.env.begin_rw_txn()?;
-                if !txn.put_value(
-                    self.block_body_db,
-                    block.header().body_hash(),
-                    block.body(),
-                    true,
-                )? {
+                if !txn.put_value(self.block_body_db, &block.body().hash(), block.body(), true)? {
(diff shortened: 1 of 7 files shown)


node/src/components/storage.rs
```diff
@@ -434,22 +434,21 @@ impl Storage {
...
-                )? {
+                if !txn.put_value(self.block_body_db, &block.body().hash(), block.body(), true)? {
                     error!("Could not insert block body for block: {}", block);
                     txn.abort();
                     return Ok(responder.respond(false).ignore());
                 }
-                if !txn.put_value(self.block_header_db, block.hash(), block.header(), true)? {
+                if !txn.put_value(
+                    self.block_header_db,
+                    &block.header().hash(),
```
afck, line 444: (This is the same as `block.hash()`, is it? You just changed it because it's more natural to write it this way when storing the header?)

afck (approved):

node/src/components/storage.rs
xcthulhu: This way relies less on the invariant that `block.header.body_hash == block.body.hash()`. It may be worth writing a check here for that invariant.

dwerner (approved):

PR #1372, opened 2021-04-27
NDRS-1205: ValidatorInitConfig Should Use BlockHeaders

xcthulhu:
- Remove linear_chain_fast_sync.rs


node/src/components/linear_chain_sync.rs
```diff
@@ -503,6 +504,17 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
         }
     }
 
+    pub fn into_maybe_latest_block_header(self) -> Option<BlockHeader> {
+        match self.state {
+            State::SyncingTrustedHash { latest_block, .. } => latest_block.map(Block::take_header),
```
afck, line 509: I imagine even the `latest_block` field itself doesn't need the full block anymore?

afck (approved):

node/src/components/linear_chain_sync.rs
xcthulhu: I want to keep the changes to `linear_chain_sync.rs` minimal; the plan is that we'll sunset it in `feat-fast-sync`.

dwerner (approved):

xcthulhu:
bors merge

PR #1374, opened 2021-04-27
OP-2012: Check for equivocators in nctl scenarios

TomVasile:
Adds function to nctl common which checks logs for equivocators.
Called in tests where equivocators are not expected (some tests have checks already specific to them).

https://casperlabs.atlassian.net/browse/OP-2012

spawned from discussion on: https://casperlabs.atlassian.net/browse/NDRS-1190


utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -208,3 +208,18 @@ function assert_node_proposed() {
...
     log "$OUTPUT"
 }
+
+# Checks logs for nodes 1-10 for equivocators
+function assert_no_equivocators_logs() {
+    local LOGS
+    log_step "Looking for equivocators in logs..."
+    for i in {1..10}; do
+        # true is because grep exits 1 on no match
+        LOGS=$(cat "$NCTL"/assets/net-1/nodes/node-"$i"/logs/stdout.log 2>/dev/null | grep -w 'Equivocated') || true
```
afck, line 218: Isn't this case-sensitive? [The log message is lowercase](https://github.com/casper-network/casper-node/blob/c195ed3d5de81e68a3cefdb8caf8e97fe0fa9ea4/node/src/components/consensus/era_supervisor.rs#L1142):
```
validator equivocated
```

utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -208,3 +208,18 @@ function assert_node_proposed() {
     log "node-$NODE_ID created a block!"
     log "$OUTPUT"
 }
+
+# Checks logs for nodes 1-10 for equivocators
+function assert_no_equivocators_logs() {
+    local LOGS
+    log_step "Looking for equivocators in logs..."
+    for i in {1..10}; do
```
goral09, line 216: I think there's an `nctl` function that returns the number of nodes created. That would be more resilient.

utils/nctl/sh/scenarios/common/itst.sh
TomVasile: spoke to Andreas offline, was looking at a different message which is more periodic. Updating to his suggestion.

utils/nctl/sh/scenarios/common/itst.sh
TomVasile: added in https://github.com/casper-network/casper-node/pull/1374/commits/27419f591aebc07149358edad135ea558f78d81c

goral09 (approved):

PR #1375, opened 2021-04-28
OP-2022: Extend nctl eviction scenario

TomVasile:
Adds re-join to itst13 eviction scenario.
Syncs chainspecs with new value from local/reseorces.
Moves function to common for reuse in scenarios.

### DIFF
--- a/utils/nctl/sh/scenarios/itst13.sh
+++ b/utils/nctl/sh/scenarios/itst13.sh
@@ -7,7 +7,7 @@ set -e
 
 #######################################
 # Runs an integration test that tries to simulate
-# and verify validator ejection.
+# and verify validator ejection and rejoining.
 #
 # Arguments:
 #   `timeout=XXX` timeout (in seconds) when syncing.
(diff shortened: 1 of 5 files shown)


utils/nctl/sh/scenarios/itst13.sh
```diff
@@ -29,6 +29,14 @@ function main() {
...
     # 5. Validate eviction occured
     assert_eviction '5'
+    # 6. Re-bid & restart node 5
+    do_submit_auction_bids "5"
+    do_read_lfb_hash "1"
+    do_start_node "5" "$LFB_HASH"
+    # 7. wait auction_delay + 1
+    do_await_era_change "4"
+    # 6. Assert that restarted validator is producing blocks.
+    assert_node_proposed "5" "180"
```
goral09, line 39: Does `assert_node_proposed` walk back the linear chain until the Genesis, when looking for a block proposed by `node-5`? If yes, then this will always return true b/c `node-5` proposed blocks before being evicted.

TomVasile: it does not walk the blocks, it looks at only new log entries that are coming in

PR #1376, opened 2021-04-28
 NDRS-1212: Fast sync must store headers and finality signatures when fetching block headers by height

xcthulhu:


node/src/components/contract_runtime.rs
```diff
@@ -529,9 +529,6 @@ where
                                 engine_state.upgrade(),
                             ) {
                                 (Some(block_at_height), Some(engine_state))
-                                    // TODO: This is a band-aid. If a block is not executed, we
-                                    // should fetch old blocks backwards until we find one with a
-                                    // state root that exists and execute forwards.
```
xcthulhu, line 534: @henrytill @dwerner So while I see some failure modes, I am not sure how to handle them in the long run, so I'm removing this comment.

node/src/components/linear_chain_sync/error.rs
```diff
@@ -47,7 +96,4 @@ where
 
     #[error(transparent)]
     TrieFetcherError(#[from] FetcherError<Trie<Key, StoredValue>, I>),
-
-    #[error("Could not store block header: {block_header}")]
-    CouldNotStoreBlockHeader { block_header: Box<BlockHeader> },
```
xcthulhu, line 52: @afck This was being thrown when the storage component was reporting the block header was already present.  But this _could_ happen on main-net, especially if the user joins with the same trusted hash over and over again.  So I am removing this error.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -101,24 +107,83 @@ where
...
+        .await;
+    Ok(block_header)
+}
+
+fn validate_finality_signatures(
+    block_header: &BlockHeader,
+    trusted_validator_weights: &BTreeMap<PublicKey, U512>,
+    finality_threshold_fraction: Ratio<u64>,
+    block_signatures: &BlockSignatures,
+) -> Result<(), FinalitySignatureError> {
```
xcthulhu, line 121: @afck Please review!  I am going to be testing finality signatures from here on out... this passes my unit test with multiple nodes with random weights.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -142,10 +237,12 @@ where
...
                     height,
                     peer
-                )
+                );
+                // If the peer we requested doesn't have the item, continue with the next peer
+                continue;
             }
-            Err(error) => return Err(error),
-        }
+            Err(error) => return Err(error.into()),
```
xcthulhu, line 244: Now that I look at this again, this might still be a fetcher error...

node/src/components/linear_chain_sync/error.rs
afck: Is the header being already present the only thing that can go wrong in `put_block_header_to_storage`? Otherwise, instead of a `bool`, this should probably return a `Result`?

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -91,34 +96,96 @@ where
...
+    let total_weight: U512 = trusted_validator_weights
+        .iter()
+        .map(|(_, weight)| *weight)
+        .sum();
+
+    let lower_bound = finality_threshold_fraction.add(1).div(2);
+    // Verify: signature_weight / total_weight >= lower_bound
+    // Equivalent to the following
+    if signature_weight * U512::from(*lower_bound.denom())
+        < total_weight * U512::from(*lower_bound.numer())
```
afck, line 164: We are basically assuming here that
* numerator and denominator in the FTT have at most 63 bits, and
* the total weight has at most 448 bits,

so this doesn't overflow. All of that is, of course, true in mainnet (and probably always, in practice), but it would be good to make it explicit somewhere.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -128,12 +195,42 @@ where
         {
             Ok(FetchedData::FromStorage { item }) => return Ok(Some(item)),
             Ok(FetchedData::FromPeer { item, .. }) => {
-                // TODO: Validate item, ban peer if necessary
                 // Compute the total weight of the validators
```
afck, line 198:
```suggestion
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -128,12 +195,42 @@ where
...
+                } = *item.clone();
+
+                match validate_finality_signatures(
+                    &block_header,
+                    &trusted_validator_weights,
+                    finality_threshold_fraction,
+                    &block_signatures,
+                ) {
+                    Ok(()) => {}
+                    Err(error) => {
```
afck, line 211: Maybe `if let Err(error) = validate_…` would be more concise than `match`, since we don't do anything in the `Ok(())` arm.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -128,12 +195,42 @@ where
...
+                        warn!(
+                            "Error validating finality signatures from peer. \
+                             peer: {peer:?}, \
+                             error: {error}",
+                            peer = peer.clone(),
+                            error = error
+                        );
```
afck, line 218:
```suggestion
                        warn!(
                            ?peer, %error,
                            "Error validating finality signatures from peer.",
                        );
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -128,12 +195,42 @@ where
...
+
+                // Store the block header
+                effect_builder
+                    .put_block_header_to_storage(Box::new(block_header.clone()))
+                    .await;
+
+                // Store the finality signatures
+                effect_builder
+                    .put_signatures_to_storage(block_signatures.clone())
+                    .await;
```
afck, line 232: These are two transactions, right? So it could in theory happen that the header gets stored but the signatures don't.
We just need to keep that in mind so we never assume we necessarily have signatures for stored headers.
But I guess we don't assume that anyway? (E.g. when going backwards from the trusted hash, we don't need signatures at all?)

afck (dismissed):
Looks good; no blockers! :+1:

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -91,34 +96,96 @@ where
...
+    let lower_bound = finality_threshold_fraction.add(1).div(2);
+    // Verify: signature_weight / total_weight >= lower_bound
+    // Equivalent to the following
+    if signature_weight * U512::from(*lower_bound.denom())
+        < total_weight * U512::from(*lower_bound.numer())
     {
-        Err(LinearChainSyncError::CouldNotStoreBlockHeader { block_header })
-    } else {
-        Ok(block_header)
+        return Err(FinalitySignatureError::InsufficientWeightForFinality {
```
goral09, line 166: This would return an error if this particular `BlockSignatures`, received from _a single peer_ does not have more than the required `lower_bound` signatures (by weight) but I don't think it is correct to expect that every peer we query for a block at height will already have enough signatures to pass this test. For example, if the block has just been finalized and/or signatures haven't been gossiped then that peer won't have enough of them. Maybe it doesn't matter in practice as we expect that nodes finalized and exchange signatures quickly enough.

afck: That's true:
* Either we need to say we always expect a full set of signatures when requesting by height. Then we should reply "we don't have it" if someone requests a block that we don't yet have enough signatures for.
* Or we send the block anyway, but then we need to start collecting signatures here, on the receiver end, and put the block in some cache.

I guess the first option is easier?

node/src/components/linear_chain_sync/operations.rs
xcthulhu: We could also just promote everything to bigints here too... not 100% on how to do that with U512s.

node/src/components/linear_chain_sync/error.rs
xcthulhu: It _should_, but storage currently swallows errors. It's worth revisiting that architecture, but it's beyond the scope of this PR.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: > This will return an error if this particular BlockSignatures instance, received from a single peer does not have more than the required lower_bound signatures (by weight) but I don't think it is correct to expect that every peer we query for a block at height will already have enough signatures to pass this test.

This will keep trying until it finds a peer with sufficient weight.

In the future, if a peer does not have sufficient weight it will report the block/block-header as missing, but that is beyond the scope of this PR.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Fixed in 01b0c705bac9f18e3f39f1adba231340f5c3e0b5

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Fixed in 24650baf9ab7725c17151ea550dc074206c98a18

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Fixed in 98c5063df841b3bc09ac729c64b0770ca8ce67e4

node/src/components/linear_chain_sync/operations.rs
afck: I believe we make some assumptions like this elsewhere, too, and in practice, 63 bits will always be enough. We should just enforce it on initialization somehow.

node/src/components/contract_runtime.rs
```diff
@@ -519,43 +513,9 @@ where
...
-                        .event(std::convert::identity)
-
-                        // .event(move |maybe_block| {
-                        //     maybe_block.map(Box::new).map_or_else(
-                        //         || Event::BlockIsNew(Box::new(finalized_block)),
-                        //         Event::BlockAlreadyExists,
-                        //     )
-                        // })
+                        info!(?finalized_block, "execute finalized block");
+                        operations::get_deploys_and_finalize_block(effect_builder, finalized_block)
```
afck, line 517: Should it be called `get_deploys_and_execute_block`?

node/src/components/contract_runtime/operations.rs
```diff
@@ -99,3 +107,44 @@ pub(super) async fn execute(
...
+    for maybe_deploy in effect_builder.get_deploys_from_storage(deploy_hashes).await {
+        if let Some(deploy) = maybe_deploy {
+            deploys.push_back(deploy)
+        } else {
+            fatal!(
+                effect_builder,
+                "deploy for block in era={} and height={} is expected to exist in the storage",
+                era_id,
+                height
+            )
```
afck, line 139: ~Can we really expect that we have all those deploys in storage?~
~Consensus will currently synchronize the protocol state of the past six eras, and will send those "finalized" events for all their blocks.~

Ah, the reason is that we synchronize the global state for the _oldest_ block we downloaded, and then run all the deploys, so we _will_ have them, right? (Just another invariant we need to make sure we never break.)

node/src/effect.rs
```diff
@@ -1676,6 +1663,6 @@ impl<REv> EffectBuilder<REv> {
 #[macro_export]
 macro_rules! fatal {
     ($effect_builder:expr, $($arg:tt)*) => {
-        $effect_builder.fatal(file!(), line!(), format_args!($($arg)*).to_string())
+        $effect_builder.fatal(file!(), line!(), format!($($arg)*).to_string())
```
afck, line 1666: Is the `.to_string()` redundant now?

`format!(…)` is equivalent to `format_args!(…).to_string()`, is it?

afck (approved):
Looks great! :+1: 
No blockers, just some remaining open questions.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: > We just need to keep that in mind so we never assume we necessarily have signatures for stored headers.

Yes.

> But I guess we don't assume that anyway? (E.g. when going backwards from the trusted hash, we don't need signatures at all?)

That's right.  It's safe to fetch a block header by hash, even if the node does not have the finality signatures for it yet.

xcthulhu:
bors merge

PR #1377, opened 2021-04-28
NDRS-1210: Deploy acceptor fix

darthsiroftardis:
REF: https://casperlabs.atlassian.net/browse/NDRS-1210

CHANGELOG:

- Changed the deploy acceptor to respond to the RPC call once a storage request is made. Previously it would respond before making storage calls.


node/src/components/rpc_server/rpcs/docs.rs
```diff
@@ -101,7 +101,7 @@ pub trait DocExample {
     fn doc_example() -> &'static Self;
 }
 
-/// The main schema for the casper node's RPC server, compliant with https://spec.open-rpc.org.
+/// The main schema for the casper node's RPC server, compliant with <https://spec.open-rpc.org.>
```
Fraser999, line 104: Thanks for squashing the warning, but we probably don't want the trailing period included in the link.  Also, we usually handle links like this:
```suggestion
/// The main schema for the casper node's RPC server, compliant with
/// [the OpenRPC Specification](https://spec.open-rpc.org).
```

Fraser999 (approved):
One nitpick, but otherwise LGTM.

darthsiroftardis:
bors merge

PR #1380, opened 2021-04-29
NO-TICKET: Change activation point of an upgrade.

goral09:
Recently, the event that triggers the check for an upgrade has been changed to fire every 60 seconds rather than on every block. This means that in tests where block creation is fast and eras take less than 1 minute it may happen that the network already goes past the activation point even before it detects scheduled upgrade. In that case, it would upgrade on the next switch block which causes all kinds of problems.

This PR moves the activation point by 2 eras which should give the test enough time to detect the upgrade on time.


Fraser999 (approved):

goral09:
bors merge

PR #1381, opened 2021-04-29
HWY-288: Avoid redundant requests caused by the Highway synchronizer.

afck:
Once all Highway dependencies are resolved, the synchronizer forgets about a proposed block and gives it to `HighwayProtocol` for inclusion into the protocol state. However, nonempty blocks are not immediately added: They are kept in the `pending_values` map while the block validator is downloading the deploys and checking the block limits. During that period, the synchronizer will request the same block again, if any dependent unit is added.

This PR should fix both the redundant requests for the block itself as well as the redundant requests to the block validator.

https://casperlabs.atlassian.net/browse/HWY-288


fizyk20 (approved):

afck:
Good catch, @goral09! Thanks for fixing this! :+1:

xcthulhu (approved):

fizyk20 (approved):

goral09:
bors merge

PR #1383, opened 2021-05-03
NDRS-1227: Use Genesis Validators When Fast Sync Joins Around Genesis

xcthulhu:
When joining right after genesis, fast sync will now use the genesis validators for finality signature verification.

### DIFF
--- a/node/src/components/linear_chain_sync/operations.rs
+++ b/node/src/components/linear_chain_sync/operations.rs
@@ -113,6 +113,66 @@ where
     Ok(block_header)
 }
 
+/// Get trusted block headers; falls back to genesis if none are available
+async fn get_trusted_validator_weights<REv, I>(
+    effect_builder: EffectBuilder<REv>,
+    chainspec: &Chainspec,
+    trusted_header: &BlockHeader,
+) -> Result<BTreeMap<PublicKey, U512>, LinearChainSyncError<I>>
+where
+    REv: From<FetcherRequest<I, BlockHeader>> + From<NetworkInfoRequest<I>> + From<StorageRequest>,
+    I: Eq + Debug + Clone + Send + 'static,
+{
+    // If we are right after genesis, use the genesis validators.
+    if trusted_header.era_id() == EraId::new(0) {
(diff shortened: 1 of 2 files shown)


node/src/components/linear_chain_sync/operations.rs
```diff
@@ -113,6 +113,66 @@ where
...
+        {
+            return Ok(trusted_validator_weights.clone());
+        }
+        current_header_to_walk_back_from = fetch_and_store_block_header(
+            effect_builder,
+            *current_header_to_walk_back_from.parent_hash(),
+        )
+        .await?;
+    }
+}
```
afck, line 173: Doesn't this duplicate [`Effect::get_era_validators`](https://github.com/casper-network/casper-node/blob/5dda5052bd1265de9164d7d3be45d24750686697/node/src/effect.rs#L1557), except that that retrieves the validators from the global state? But at that point, we should have the global state available so it should work, shouldn't it?

xcthulhu: Almost, but not exactly.  `Effect::get_era_validators` can does storage and contract-runtime requests.  We don't have anything in either storage or the state trie at the point this code is called.  The functions are similar, however: they both fallback to the chainspec in the special case where we are asking for the validators for era 0.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -113,6 +113,66 @@ where
...
+        let genesis_validator_weights: BTreeMap<PublicKey, U512> = chainspec
+            .network_config
+            .accounts_config
+            .accounts()
+            .iter()
+            .filter_map(|account_config| {
+                if account_config.is_genesis_validator() {
+                    Some((
+                        account_config.public_key(),
+                        account_config.bonded_amount().value(),
```
afck, line 137: Does that include delegators? (Not sure how that works at genesis.)

xcthulhu: I _think_ so?  Paging @Fraser999 ... maybe we should have a function that shares this logic in both the genesis code and here as well.

PR #1384, opened 2021-05-03
NDRS-1161: Fast Sync Download Linear Chain Blocks And Deploys

xcthulhu:


node/src/components/linear_chain_sync/operations.rs
```diff
@@ -263,16 +269,120 @@ where
     return Ok(outstanding_tries);
 }
 
+struct BlockWithMetadataFetchedFromRemote<I> {
+    block_with_metadata: Box<BlockWithMetadata>,
+    sender: I,
+}
+
+/// Fetches a block from the network by height.
+async fn fetch_and_store_block_by_height_from_remote<REv, I>(
```
afck, line 278: Maybe:
```suggestion
async fn fetch_and_store_block_by_height<REv, I>(
```
Because it's not necessarily from remote, if we already have it in storage?

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -399,5 +512,34 @@ where
         outstanding_trie_keys.extend(missing_descendant_trie_keys);
     }
 
+    // Process blocks to until we are current.
```
afck, line 515:
```suggestion
    // Process blocks until we are current.
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -263,16 +269,120 @@ where
...
+        + From<StorageRequest>,
+    I: Eq + Debug + Clone + Send + 'static,
+{
+    for peer in get_and_shuffle_network_peers(effect_builder).await {
+        match effect_builder
+            .fetch::<BlockWithMetadata, I>(height, peer.clone())
+            .await
+        {
+            Ok(FetchedData::FromStorage { item }) => {
+                // TODO: We do not know this peer has the block, don't use a fetcher
```
afck, line 296: Not sure I understand this comment. You mean we should basically set `sender: None` (i.e. make it an option)?

afck (approved):

node/src/components/linear_chain_sync/operations.rs
xcthulhu: I'm not sure.  What happens if I try to validate this with the `block_validator` and I don't have the right sender?

afck: Ah, you mean then it will conclude that nobody has the deploys? :confused:

xcthulhu: I made a follow up ticket to remedy this.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Alternatively, we could just break out the code for fetching deploys, since fast sync does not need to actually validate blocks, just fetch deploys.

xcthulhu:
bors merge

PR #1385, opened 2021-05-03
EE-1257: Remove transfer gas field

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1257

This commit removes the `gas` field from the public interface (as defined by Serialize/Deserialize traits). The API internally is restricted to make sure the gas field is unused.

For backward compatibility, ToBytes/FromBytes still serializes and deserializes the value, although it's assumed to be "zero."

The correct way to measure the cost of a native transfer is to get the deploys execution result.


darthsiroftardis (approved):
LGTM!

xcthulhu:
Could we get rid of the field from the internal data structures?  I think we would have to serialize a sentinel value in the `ToBytes` impl, but other than that we could just get rid of the field entirely.

Would that work?

types/src/transfer.rs
```diff
@@ -168,7 +199,7 @@ impl FromBytes for Transfer {
         let (source, rem) = URef::from_bytes(rem)?;
         let (target, rem) = URef::from_bytes(rem)?;
         let (amount, rem) = U512::from_bytes(rem)?;
-        let (gas, rem) = U512::from_bytes(rem)?;
+        let (_reserved_u512, rem) = U512::from_bytes(rem)?;
```
Fraser999, line 202: This should be commented to explain why it's required.  Probably `_legacy_gas` might be a better name, as "reserved" implies reserved for future use to me.

Similarly for the two places in the `ToBytes` impl.

Fraser999 (approved):
Just a nitpick, but optimistic approval assuming it's handled.

Fraser999:
This will break backwards compatibility for any consumers of our JSON representation of the `Transfer` struct, so we'll postpone this work until the 2.0.0 release.

PR #1391, opened 2021-05-04
OP-2032: Removing need for wrapper script in CI

TomVasile:
Removes wrapper script.
Moves put to using drone plugin.
No plugin existed for the del/get so step(s) were reworked so that the wrapper is not necessary.


sacherjj (approved):

TomVasile:
bors try

TomVasile:
bors try

PR #1394, opened 2021-05-05
EE-1246: Exhaustive pattern match

mpapierski:
Ref: https://casperlabs.atlassian.net/browse/EE-1246

This PR refactors constants for enum tags with exhaustive wildcard pattern matching and avoids wildcards (`_ => { something; }`) around ToBytes/FromBytes methods.

This is done by grouping constants under enum and using FromPrimitive/ToPrimitive to make ToBytes/FromBytes always do exhaustive pattern matching. Proptests' arbitrary generators are updated where possible as well as there are more fixtures generated that rely on #1360


cspramit:
@mpapierski why did we reference PR #1627 to this ticket? when 1627 was in 1.3 and this is 1.4 - is the work done?

mpapierski:
> 
> 
> @mpapierski why did we reference PR #1627 to this ticket? when 1627 was in 1.3 and this is 1.4 - is the work done?

At the time #1627 was in development I pointed out that we should already use exhaustive pattern matching just like in this PR

mpapierski:
Closing in favor of #2627

PR #1398, opened 2021-05-06
NDRS-1241: Only respond to all when fetcher is infallible

xcthulhu:
For some fetchers (namely those for deploys, block headers by hash, and tries) it's safe to respond to all when we've fetched an item with the right ID.  Others may respond with incorrect block bodies, signatures, or deploys. For those we will ultimately need to ban that peer.

### DIFF
--- a/node/src/components/fetcher.rs
+++ b/node/src/components/fetcher.rs
@@ -71,6 +71,13 @@ pub enum FetchedOrNotFound<T, Id> {
 }
 
 pub trait ItemFetcher<T: Item + 'static> {
+    /// Indicator on whether it is safe to respond to all of our responders. For example, [Deploy]s
+    /// and [BlockHeader]s are safe because their [Item::id] is all that is needed for
+    /// authentication. But other structures have _finality signatures_ or have substructures that
+    /// require validation. These are not infallible, and only the responders corresponding to the
+    /// node queried may be responded to.
+    const SAFE_TO_RESPOND_TO_ALL: bool;
+
     fn responders(&mut self) -> &mut HashMap<T::Id, HashMap<NodeId, Vec<FetchResponder<T>>>>;
 
     fn metrics(&mut self) -> &FetcherMetrics;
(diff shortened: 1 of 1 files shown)


node/src/components/fetcher.rs
```diff
@@ -263,6 +292,8 @@ impl ItemFetcher<Deploy> for Fetcher<Deploy> {
 }
 
 impl ItemFetcher<Block> for Fetcher<Block> {
+    const SAFE_TO_RESPOND_TO_ALL: bool = false;
```
afck, line 295: Why not? Isn't this by hash?

node/src/components/fetcher.rs
```diff
@@ -154,53 +161,73 @@ pub trait ItemFetcher<T: Item + 'static> {
         }
     }
 
+    /// Sends result to all responders
+    fn respond_to_all(&mut self, id: T::Id, result: FetchResult<T, NodeId>) -> Effects<Event<T>> {
```
afck, line 165: So this is never meant to be called with an `Err` result? Would be good to document that.

afck (approved):

node/src/components/fetcher.rs
xcthulhu: So, heres the block data structure:

```rust
pub struct Block {
    hash: BlockHash,
    header: BlockHeader,
    body: BlockBody,
}
```

We have a couple rules here:

   - `hash == header.hash()`
   - `body.hash() == header.body_hash`
   
If these are violated then the peer is faulty...

afck: Then shouldn't it also be `false` for deploys? E.g. their `ttl` could exceed the maximum as specified in the chainspec.

node/src/components/fetcher.rs
xcthulhu: Fixed in 9fedf1eef373bce79649cb8849b461b95515f386

node/src/components/fetcher.rs
afck: Sorry for the confusion! It's correct, of course: It only needs to be `false` for item types where a matching hash is insufficient to respond to everyone.
If someone sends us a block with the correct hash, it _is_ the block we asked for, even if it's invalid.

PR #1400, opened 2021-05-06
NDRS-1226: BlockWithMetadata fetcher no longer uses storage

xcthulhu:
Whenever fast-sync fetches a `BlockWithMetadata`, it needs to validate it in order to retrieve its deploys.  Validation requires providing the peer from which the block was fetched.  In the case that the block was retrieved from storage, no such peer is available.  To avoid this, we prevent the `BlockWithMetadata` fetcher from retrieving from storage.

### DIFF
--- a/node/src/components/fetcher.rs
+++ b/node/src/components/fetcher.rs
@@ -515,7 +515,7 @@ where
                     self.metrics.found_in_storage.inc();
                     self.got_from_storage(item, peer)
                 }
-                _ => self.failed_to_get_from_storage(effect_builder, id, peer),
+                None => self.failed_to_get_from_storage(effect_builder, id, peer),
             },
             Event::GotRemotely { item, source } => {
                 match source {


node/src/components/fetcher.rs
```diff
@@ -450,7 +471,7 @@ where
                     self.metrics.found_in_storage.inc();
                     self.got_from_storage(item, peer)
                 }
-                None => self.failed_to_get_from_storage(effect_builder, id, peer),
+                _ => self.failed_to_get_from_storage(effect_builder, id, peer),
```
afck, line 474: Why not `None`?

afck (commented):
If we don't check storage and don't reply to all responders — are we still using much of the fetcher logic at all? Would it be cleaner if fast sync just had its own message type and sent requests and handled responses over the network for blocks with metadata itself?

afck:
Actually: Now that we are dropping the assumption that with every block we also have its deploys in storage, we also can't assume that about our peers anymore. So if we received a block from a peer, that still doesn't mean that peer also has the deploys?

node/src/components/fetcher.rs
xcthulhu: Good catch!  Fixed in a2dd9b2bed8f877e50e5ab7d6aeb49483165f3e3

xcthulhu:
> Actually: Now that we are dropping the assumption that with every block we also have its deploys in storage, we also can't assume that about our peers anymore. So if we received a block from a peer, that still doesn't mean that peer also has the deploys?

We can expect ourselves to not necessarily have the deploys when we are in the joiner reactor.  The joiner reactor does not currently allow other nodes to fetch from it.  When we are in the validator reactor we should always have deploys, however.  But either way, a peer should not report having a block if they do not have its deploys (or finality signatures for that matter).  I'll ensure this in follow up work.

afck (approved):

afck (approved):

xcthulhu:
I did this because to use the block validator it is necessary to hand it a peer that provided the block.

However, I have decided to simply use the deploy fetcher to fetch the deploys necessary for fast sync. I don't need really need to validate a block.

PR #1402, opened 2021-05-07
HWY-292, HWY-293, NDRS-1237: Reduce redundant data and requests in Highway. (ports to dev)

afck:
Port https://github.com/casper-network/casper-node/pull/1389 and https://github.com/casper-network/casper-node/pull/1397 to the `dev` branch:
* Avoid requesting the same vertex redundantly from lots of peers.
* Add more logging to the synchronizer.
* Use `Arc` to avoid cloning consensus values (i.e. lists of deploy hashes) in memory.
* Change the default `pending_vertex_timeout` to 30 minutes.
* Only request our peers' latest states if we suspect we might be stuck.

https://casperlabs.atlassian.net/browse/HWY-292
https://casperlabs.atlassian.net/browse/HWY-293
https://casperlabs.atlassian.net/browse/NDRS-1237

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -993,12 +993,6 @@ where
                 .immediately()
                 .event(move |()| Event::Action { era_id, action_id }),
             ProtocolOutcome::CreateNewBlock(block_context) => {
-                let past_deploys = block_context
-                    .ancestor_values()
-                    .iter()
-                    .flat_map(|block_payload| block_payload.deploys_and_transfers_iter())
-                    .cloned()
-                    .collect();
                 let accusations = self
                     .era_supervisor
                     .iter_past(era_id, self.era_supervisor.bonded_eras())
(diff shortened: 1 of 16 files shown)


afck:
I cherry-picked the merge commit for https://github.com/casper-network/casper-node/pull/1397; otherwise I'd have to manually resolve all the merge conflicts that were already resolved in https://github.com/casper-network/casper-node/pull/1397/commits/f012c0aff07b49336c6fb1c45df01d7553f1e7e5.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1053,7 +1046,7 @@ where
                         .collect(),
                 });
                 let finalized_block = FinalizedBlock::new(
-                    value,
+                    Arc::try_unwrap(value).unwrap_or_else(|arc| (*arc).clone()),
```
goral09, line 1049: Can you extract this to a method like discussed? Unless it's used only in this single place.

goral09 (approved):

node/src/components/consensus/era_supervisor.rs
afck: It's only used here, but it would still be nicer.

Anyway, I was considering making a follow-up PR with cleanups anyway, and limiting this PR to just the backports.

goral09: That's fine.

fizyk20 (approved):

afck:
bors merge

PR #1406, opened 2021-05-07
NDRS-1244: add client support for multisig transfers

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1244

This PR introduces a separate subcommand to the Rust client `make-transfer` which is a mixture of the existing two subcommands `make-deploy` and `transfer`.  It allows for the creation of a Deploy for a transfer as a JSON-encoded file.  Using the existing `sign-deploy` and `send-deploy` subcommands on such a file allows us to use the Rust client to create multisig transfers.

### DIFF
--- a/client/lib/deploy.rs
+++ b/client/lib/deploy.rs
@@ -10,11 +10,11 @@ use casper_node::{
     rpcs::{account::PutDeploy, chain::GetBlockResult, info::GetDeploy, RpcWithParams},
     types::{Deploy, DeployHash, TimeDiff, Timestamp},
 };
-use casper_types::{ProtocolVersion, SecretKey};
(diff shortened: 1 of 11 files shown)


darthsiroftardis:
Could we change the command to something like `make-signed-transfer` or `multisig-transfer`. This would help differentiate the two commands better on a superficial level.

darthsiroftardis (approved):
Code looks good, don't want to block the merge on something like the name

Fraser999:
bors merge

Fraser999:
> Could we change the command to something like make-signed-transfer or multisig-transfer. This would help differentiate the two commands better on a superficial level.

As discussed offline, we'll leave as is for now.  If we come up with better names for these, we can ticket for future work.

PR #1409, opened 2021-05-08
NDRS-1217: handle chainspec in casper-updater

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1217

This PR updates the `casper-updater` tool to handle changes required in the production chainspec and associated files.

The existing `--bump` arg which accepts values of `major`, `minor` or `patch` is used to also bump the appropriate part of the protocol version in the `resources/production/chainspec.toml` file.  If we need more fine-grained control over setting the protocol version, a new arg can be added to the tool for that.  But for now, it seems like our build version will be kept in lockstep with the protocol version.

The tool now has an `--activation-point` arg, which takes an integer used to set the activation point in the `resources/production/chainspec.toml` file.  There is no check that the value is being replaced with a higher one, since under an emergency restart, the value could validly be lower or equal to the existing one.

If `--bump` is specified, then `--activation-point` must also be specified, and vice-versa.

The `--dry-run` command has been adapted to ensure these new affected files are displayed too, showing what the protocol version and activation point are currently, what they'll be updated to, and which files are affected by that.

If the tool is run with no args, it expects the user to input each change explicitly, allowing for fine-grained control over the version update for each crate.  This functionality has also been extended to include inputs for the protocol version and activation point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,9 +44,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.14.1"
+version = "0.15.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a55f82cfe485775d02112886f4169bde0c5894d75e79ead7eafe7e40a25e45f7"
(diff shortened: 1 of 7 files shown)


Fraser999:
bors merge

PR #1413, opened 2021-05-10
NDRS-1247: start recurring connection housekeeping task

Fraser999:
Ref: https://casperlabs.atlassian.net/browse/NDRS-1247

This PR fixes an oversight in the recent cleanup of small_network, whereby a recurring housekeeping task was never actually started on initialisation of the component.

I have run the integration test `itst01.sh` successfully three times after making this change.


marc-casperlabs (approved):
:+1:

PR #1419, opened 2021-05-10
NDRS-1202: Connection negotiation tasks

marc-casperlabs:
Note: The size of this PR is in part due to some things being moved from one file to another, or fields being renamed.

This PR extracts tasks (asynchronous functions, most run as effects) from the small networking component into a submodule. In doing so it unifies what used to be broken up steps for establishing a new connection into linear `async` functions.

As a result there is a noticeable complexity reduction in the component itself. Handshake handling benefits immensely from this, as the code can now be properly shared and no longer leaks into the handling of other events.

Tasks become testable as a result.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -464,6 +464,7 @@ where
     /// Sets up an established outgoing connection.
     ///
     /// Initiates sending of the handshake as soon as the connection is established.
+    #[allow(clippy::redundant_clone)]
     fn handle_outgoing_connection(
         &mut self,
         effect_builder: EffectBuilder<REv>,
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -33,7 +33,7 @@ mod message;
 mod message_pack_format;
 mod outgoing;
 mod symmetry;
-mod tasks;
+pub(crate) mod tasks;
 #[cfg(test)]
 mod tests;
 
@@ -72,10 +72,11 @@ use tracing::{debug, error, info, trace, warn, Instrument, Span};
 use self::{
     counting_format::{ConnectionId, CountingFormat, Role},
     error::Result,
+    event::IncomingConnection,
     message_pack_format::MessagePackFormat,
     outgoing::{DialOutcome, DialRequest, OutgoingConfig, OutgoingManager},
     symmetry::ConnectionSymmetry,
-    tasks::{IncomingConnection, NetworkContext},
+    tasks::NetworkContext,
 };
 pub(crate) use self::{
     error::display_error,
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -44,10 +44,7 @@ use std::{
     mem,
     net::{SocketAddr, TcpListener},
     result,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc, Weak,
-    },
+    sync::{Arc, Weak},
     time::{Duration, Instant},
 };
 
@@ -173,8 +170,6 @@ where
     /// order that they can be gracefully terminated.
     #[data_size(skip)]
     shutdown_receiver: watch::Receiver<()>,
-    /// Flag to indicate the server has stopped running.
-    is_stopped: Arc<AtomicBool>,
     /// Join handle for the server thread.
     #[data_size(skip)]
     server_join_handle: Option<JoinHandle<()>>,
@@ -278,7 +273,6 @@ where
             shutdown_sender: Some(server_shutdown_sender),
             shutdown_receiver,
             server_join_handle: Some(server_join_handle),
-            is_stopped: Arc::new(AtomicBool::new(false)),
             net_metrics: Arc::new(net_metrics),
         };
 
@@ -598,7 +592,6 @@ where
                         addr,
                         self.certificate.clone(),
                         self.secret_key.clone(),
-                        self.is_stopped.clone(),
                     )
                     .instrument(span)
                     .result(
@@ -721,10 +714,6 @@ where
         async move {
             // Close the shutdown socket, causing the server to exit.
             drop(self.shutdown_sender.take());
-
-            // Set the flag to true, ensuring any ongoing attempts to establish outgoing TLS
-            // connections return errors.
-            self.is_stopped.store(true, Ordering::SeqCst);
 
             // Wait for the server to exit cleanly.
             if let Some(join_handle) = self.server_join_handle.take() {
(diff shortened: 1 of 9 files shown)


node/src/components/small_network/tasks.rs
```diff
@@ -0,0 +1,478 @@
...
+            Ok(public_addr)
+        }
+    } else {
+        // Received a non-handshake, this is an error.
+        Err(ConnectionError::DidNotSendHandshake)
+    }
+}
+
+/// Core accept loop for the networking server.
+pub(super) async fn server<P, REv>(
```
Fraser999, line 330: A verb would be more appropriate - `serve` or `run_server` maybe?

node/src/components/small_network/tasks.rs
```diff
@@ -0,0 +1,478 @@
...
+        .ssl()
+        .peer_certificate()
+        .ok_or(ConnectionError::NoPeerCertificate)?;
+
+    let peer_id = NodeId::from(
+        tls::validate_cert(peer_cert)
+            .map_err(ConnectionError::PeerCertificateInvalid)?
+            .public_key_fingerprint(),
+    );
+
```
Fraser999, line 91: We used to check for the server having stopped at this point, and return an error rather than continue.  I seem to recall that this addressed an issue when transitioning from the joiner reactor to the validator.  Maybe I'm missing something, but it doesn't look like we're handling that case now in this PR?

node/src/components/small_network/tasks.rs
marc-casperlabs: This was intentional, the noun-named tasks are long-running until terminated, while the verb functions/tasks are self-terminating. `server`, `message_sender` and `message_receiver` are in the former category. Maybe it's worth keeping this little scheme?

node/src/components/small_network.rs
```diff
@@ -360,187 +339,247 @@ where
...
+        result: io::Result<()>,
+        peer_id: Box<NodeId>,
+        peer_addr: SocketAddr,
+        span: Span,
+    ) -> Effects<Event<P>> {
+        span.in_scope(|| {
+            // Log the outcome.
+            match result {
+                Ok(()) => {
+                    info!("regular connection close")
```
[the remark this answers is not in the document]
marc-casperlabs: That was intentional, the "regular" refers to the act of closing, not the connection.

So maybe it should be

* regular connection closing
* connection closed regularly

?

node/src/components/small_network/tasks.rs
marc-casperlabs: Correct me if I am wrong, but I believe you are referring to this? https://github.com/casper-network/casper-node/blob/2569ebd51923d0353542c93c5d47480246170cbb/node/src/components/small_network.rs#L1350-L1359

This was based on the following ticket "back then" (about 7 months ago):

> Problem: the joiner reactor connects to peers using small_network, and when the switch to the validator reactor occurs a new network is started and a new certificate is issued. As the the NodeId is based on the cert, this results in the same ip address having two different NodeId’s and two different ports being bound. Eventually the entry from the joiner reactor will go away when it is closed.
>
> Ideally, a clean finalization of the joiner reactor related state would clear any remaining incoming / outgoing connections before the validator reactor starts.

There are multiple reasons for removing this functionality, although not all aligned on a common goal:

1. The workaround was flawed: The code listed above turned a larger race condition into a smaller one. Core issue that an outgoing connection that is being established while the server is shut down will be kept open, since the handle that could cause a closing is stuck in the old event queue that is no longer being polled (the connection has a stream and a sink; the stream is closed as soon as possible, while the sink is shipped around with the `IncomingConnection` event). Adding the boolean flag would terminate it in many cases, but still leaves the door open for a connection to reach the event queue only to not be processed.

2. The original ticket is a non-issue. We should not be relying on the 1:1 identity of these IDs in the first place, I hope this has not been silently made a requirement, since we explicitly removed the persistence of IDs before that: https://github.com/casper-network/casper-node/commit/df7638a12d27170ee4b4d5e83ac9a94be5d5e7df#diff-4acfd86427b0a531d080292cd244708eaf7856fcd084459fce07ea3c6d2749ef

3. We silently re-introduced (2.) later on: https://github.com/casper-network/casper-node/commit/33dc7e5b4247e47021eb7fd6ab05c9bd293144cb#diff-5a2e839b41d4bdf6e8b228edbf0ef6b37bef11d7023cc469b06ee996301d8a76, making the original issue moot.

4. There is no longer potentially lingering state, as the connection process, which used to visit `SmallNetwork` three times during the process now only reports back with entirely finished connection. Otherwise, only the IP is known to the component. In a running component, these outgoing connections would also be cleaned up during housekeeping.

5. There was a proposed solution for cleaning up *everything* on shutdown as suggested in the ticket, but it was rejected: https://github.com/casper-network/casper-node/pull/1365

6. A situation where this is a problem will look like this: Node on machine A will have two outgoing connections, O->B and P->B, to the node on machine B. O->B will be from a dead reactor and a shutdown server tasks, so it will quickly be one-sided. There will be no messages outgoing from O->B as well, as there is no reactor, no sender for this (the handle is stuck in the unpolled event queue). P->B is from B's perspective an entirely different node, and any prior outgoing connection will have been terminated.
   O->B will just linger as dead weight, although we could clean it up, we currently don't purge incoming asymmetric connections, only outgoing ones.

If we are to prevent this in the future, a better way would be to simply nullify event queue handles once we've shutdown a reactor, and empty the queue entirely. Adding back the previous solution is a workaround at best though.

Fraser999: OK - thanks for the detailed history and suggested future plan.  It does seem like this should be a non-issue, so I'll approve the PR as is.  I think we should be prepared to drop in a short-term workaround if required after further testing though.  We've been wrong about non-issues being non-issues before :)  Hopefully that won't be required though.

node/src/components/small_network/tasks.rs
Fraser999: It was a nitpick, so since there's a rationale for it, I'm fine to leave it as is.

node/src/components/small_network.rs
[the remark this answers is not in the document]
Fraser999: `regular connection closing` gets my vote

Fraser999 (approved):

node/src/components/small_network/tasks.rs
marc-casperlabs: Alright, permission to tell me "I told you so" officially granted!

PR #1428, opened 2021-05-10
NDRS-1231: Authenticate validator key in handshake

marc-casperlabs:
This authenticates a peer based on their validator key. Note that #1419 needs to be merged (and this PR rebased) first, for this reason it is still in draft mode.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -405,6 +405,7 @@ where
                 peer_addr,
                 public_addr,
                 peer_id,
+                peer_consensus_public_key: _,
                 stream,
             } => {
                 info!("new incoming connection established");
@@ -548,6 +549,7 @@ where
             OutgoingConnection::Established {
                 peer_addr,
                 peer_id,
+                peer_consensus_public_key: _,
                 sink,
             } => {
                 info!("new outgoing connection established");
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -242,7 +242,7 @@ where
                 let root = cfg.dir();
                 cfg.value()
                     .load_keys(root)
-                    .map(|(secret_key, public_key)| ConsensusKeyPair::new(secret_key, public_key))
+                    .map(|(secret_key, public_key)| ConsensusKeyPair::new(Arc::try_unwrap(secret_key).expect("did not expect additional references on freshly created secret key Arc"), public_key))
             })
             .transpose()
             .map_err(Error::LoadConsensusKeys)?;
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -237,15 +237,19 @@ where
         }
 
         // If given consensus key configuration, load it for handshake signing.
-        let consensus_keys = consensus_cfg
-            .map(|cfg| {
-                let root = cfg.dir();
-                cfg.value()
-                    .load_keys(root)
-                    .map(|(secret_key, public_key)| ConsensusKeyPair::new(Arc::try_unwrap(secret_key).expect("did not expect additional references on freshly created secret key Arc"), public_key))
-            })
-            .transpose()
-            .map_err(Error::LoadConsensusKeys)?;
+        let consensus_keys =
+            consensus_cfg
+                .map(|cfg| {
+                    let root = cfg.dir();
+                    cfg.value().load_keys(root)
+                })
+                .transpose()
+                .map_err(Error::LoadConsensusKeys)?
+                .map(|(secret_key, public_key)| {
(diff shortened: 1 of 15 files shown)


node/src/components/small_network/message.rs
```diff
@@ -44,6 +44,27 @@ impl<P: Payload> Message<P> {
     }
 }
 
+/// A pair of secret keys used by consensus.
+pub(super) struct ConsensusKeyPair {
+    secret_key: SecretKey,
```
goral09, line 49: IIRC, the intent was to never clone a `SecretKey` but use an `Arc`.

goral09:
Can I ask, why do we need this?

node/src/components/small_network/message.rs
marc-casperlabs: It is not cloned, as the outer structure is providing the `Arc` here. The `struct` itself does not implement `Clone` as well :)

goral09: But you do pass `SecretKey` to the constructor, not `Arc`. So it has to be moved into this struct – which means it's either forgotten on the caller end or it's cloned, isn't that right?

marc-casperlabs: That depends on the compiler. It is moved, but whether or not that copies it from stack to heap here is not for us to specify.

I did notice that there have been some changes to how this is handled in `dev`, let's re-examine it after I have finished rebasing.

marc-casperlabs:
The PR has been rebased and is now ready for review.

node/src/components/small_network/message.rs
marc-casperlabs: Done. So the question remains if we want to be strict and move all responsibility of creating a heap-only copy into the `Loadable` impl of `Arc<SecretKey>`.

If that's the case, then it would be correct for `ConsensusKeyPair` to hold an `Arc<SecretKey>`.

On the other hand, I think the preferable alternative would be to alter `SecretKey` itself to always use the heap and security features instead, as that would make it impossible to construct stack-allocated keys at all. Otherwise we are relying on convention instead of types ("always use an `Arc` around `SecretKey`").

afck: So we actually _do_ end up with multiple copies in memory here, because we call `load_keys` twice, right?
I don't think that's an issue, but if it is, `load_keys` itself could memoize it the first time, and then only clone the `Arc`.

afck (approved):

node/src/components/small_network/message.rs
marc-casperlabs: That's fair, but I imagine the logic would be somewhat complicated, as we would have to ensure not to load the same key twice. Is that worth the effort right now? I feel it'd be better spent in a follow-up item, implementing a secure conclave for the key?

afck: I agree, I think it's fine as it is. :+1:

node/src/components/small_network.rs
```diff
@@ -232,6 +236,17 @@ where
             public_addr.set_port(local_addr.port());
         }
 
+        // If given consensus key configuration, load it for handshake signing.
+        let consensus_keys = consensus_cfg
+            .map(|cfg| {
+                let root = cfg.dir();
+                cfg.value()
+                    .load_keys(root)
+                    .map(|(secret_key, public_key)| ConsensusKeyPair::new(Arc::try_unwrap(secret_key).expect("did not expect additional references on freshly created secret key Arc"), public_key))
```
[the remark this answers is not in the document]
marc-casperlabs: There's no clean way to tame this monstrosity other than splitting it up, I think, so that's what I'll do :)

node/src/components/small_network.rs
```diff
@@ -237,15 +237,19 @@ where
...
+                .map(|cfg| {
+                    let root = cfg.dir();
+                    cfg.value().load_keys(root)
+                })
+                .transpose()
+                .map_err(Error::LoadConsensusKeys)?
+                .map(|(secret_key, public_key)| {
+                    ConsensusKeyPair::new(Arc::try_unwrap(secret_key).expect(
+                    "did not expect additional references on freshly created secret key Arc",
+                ), public_key)
```
afck, line 251:
```suggestion
                        "did not expect additional references on freshly created secret key Arc",
                    ), public_key)
```
(…or is this how rustfmt did it?)

afck (approved):

node/src/components/small_network.rs
marc-casperlabs: I'll double check :)

marc-casperlabs: It seems `rustfmt` has been waving the white flag facing this one. I manually improved the formatting.

afck (approved):

PR #1447, opened 2021-05-14
Persist event stream event index

### LINKED ISSUE #1417: Avoid restarting event stream event ID from 0 on each restart
Currently, the event stream indexes every event it produces.  The numbering starts at 0 and increases by 1 for every subsequent event.  However, if the node stops and restarts (which it would do e.g. during an upgrade) then the numbering starts from 0 again.

We want to persist the indexing and on the restart, continue from where we left off.

Fraser999:
This PR provides the SSE server with a persistent cache for the last event index.  On dropping the server, the value is written to a file and read on starting.

The same folder used by the storage component is used for holding this file.  This has the advantage of not requiring any change to the config file type, and also restricts the number of locations the app will write to while running.

As well as the provided unit tests, I verified this works as expected by running an nctl upgrade test and asserting that the event indices were persisted across the upgrade by all nodes.

Closes #1417

### DIFF
--- a/node/src/components/event_stream_server.rs
+++ b/node/src/components/event_stream_server.rs
@@ -22,10 +22,11 @@
 
 mod config;
 mod event;
+mod event_indexer;
 mod http_server;
 mod sse_server;
 
-use std::{convert::Infallible, fmt::Debug};
+use std::{convert::Infallible, fmt::Debug, path::PathBuf};
 
 use datasize::DataSize;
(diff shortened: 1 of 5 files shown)


marc-casperlabs (approved):
Looks good to me for now :)

EdHastingsCasperLabs (approved):

Fraser999:
bors merge

PR #1464, opened 2021-05-20
EE-1217: add initial support for `get_call_stack`

henrytill:


execution_engine_testing/tests/src/test/regression/mod.rs
```diff
@@ -24,7 +25,6 @@ mod ee_550;
 mod ee_572;
 mod ee_584;
 mod ee_597;
-mod ee_598;
```
henrytill, line 27: Note that we've disabled the EE-598 regression for the moment, as it uses proxy contract in a way that no longer works with the current checks.  

A fix will follow...

dwerner:
This is passing the upgrade test:

https://gist.github.com/dwerner/eade3ce34f71788c8122b0b66c88cba2

EdHastingsCasperLabs (approved):

goral09:
Hey Henry, can you include the description of that new function? Or link a GH/ZenHub issue that describes it?

henrytill:
Converting back to a draft after realizing it would be more useful to include contract hash rather than contract wasm hash in the `Contract` variant.

henrytill:
closed in favor of #1627

PR #1467, opened 2021-05-21
EE-1242: Implement Dictionary

### LINKED ISSUE #1420: Dictionary
This task is about implementing support for improved local keys, which intends to resolve past problems that lead to their removal. Contract developers will use this to store data in a separate keyspace prefixed by the URef address provisioned to a user by the system.

CEP for reference: casper-network/ceps#39

mpapierski:
This PR implements support for improved local keys, which intends to resolve past problems that lead to their removal. Contract developers will use this to store data in a separate keyspace prefixed by the URef address provisioned to a user by the system.

CEP for reference: casper-network/ceps#39

Closes #1420


types/src/key.rs
```diff
@@ -501,14 +550,18 @@ impl FromBytes for Key {
...
+            }
             _ => Err(Error::Formatting),
         }
     }
 }
 
 impl Distribution<Key> for Standard {
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Key {
-        match rng.gen_range(0..9) {
+        match rng.gen_range(0..=9) {
```
dwerner, line 564: Why don't we just bump to 10?

smart_contracts/contracts_as/test/local-state/assembly/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../../../../../../.nvm/versions/node/v10.16.3/lib/node_modules/assemblyscript/std/assembly.json",
```
dwerner, line 2: This looks brittle...

smart_contracts/contracts_as/test/local-state/assembly/index.ts
```diff
@@ -0,0 +1,112 @@
...
+import {CLValue, CLType, CLTypeTag} from "../../../../contract_as/assembly/clvalue";
+import { Key } from "../../../../contract_as/assembly/key";
+import { AccessRights, URef } from "../../../../contract_as/assembly/uref";
+import { Pair } from "../../../../contract_as/assembly/pair";
+import { createLocal, readLocal, writeLocal } from "../../../../contract_as/assembly/local";
+import { fromBytesString, toBytesString } from "../../../../contract_as/assembly/bytesrepr";
+import { revert } from "../../../../contract_as/assembly/externals";
+
+const LOCAL_KEY_NAME = "local";
+const WRITE_LOCAL_KEY: Array<u8> = [66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66, 66];
```
dwerner, line 16: Do we have a ts formatter in use here?

henrytill (commented):
First pass, overall looks good.

A couple big points to discuss:
1. What name are we going to use instead of "local"?
2. Does it make sense to have a convenience function like `create_local` or should we just have people create and use urefs themselves?

cc @EdHastingsCasperLabs

henrytill:
Recording the decisions of today's offline discussion...

> 1. What name are we going to use instead of "local"?

Decided to call it "dictionary":
* Key name should be `Dictionary`.  
* API Functions: `new_dictionary`, `dictionary_put`, `dictionary_get` (these last two are new suggestions)

> 2. Does it make sense to have a convenience function like `create_local` or should we just have people create and use urefs themselves?
 
Yes, but we should call it `new_dictionary` and it should take a string/str name argument from which we will automatically generate a named key entry.  The other functions should continue to take a `URef` seed argument.

execution_engine/src/shared/host_function_costs.rs
```diff
@@ -186,9 +188,9 @@ where
 #[derive(Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Debug, DataSize)]
 pub struct HostFunctionCosts {
     pub read_value: HostFunction<[Cost; 3]>,
-    pub read_value_local: HostFunction<[Cost; 3]>,
+    pub read_local: HostFunction<[Cost; 5]>,
```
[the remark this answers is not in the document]
mpapierski: No, unfortunately - needs #1487, or we can ship v1 with hardcoded defaults without modifying `ProtocolData`

execution_engine/src/core/runtime/externals.rs
```diff
@@ -992,10 +996,10 @@ where
                     _,
                 ) = Args::parse(args)?;
                 self.charge_host_function_call(
-                    &host_function_costs.dictionary_get,
+                    &host_function_costs.read_value_local,
                     [
-                        uref_ptr,
-                        uref_size,
+                        // uref_ptr,
+                        // uref_size,
```
dwerner, line 1002: Are these a `todo` perhaps or are they intentionally left commented-out as a note? (If intentional, perhaps we could add a note as to why they are not included in the call to `charge_host_function_call()`).

execution_engine/src/core/runtime/externals.rs
```diff
@@ -1014,10 +1018,10 @@ where
             FunctionIndex::DictionaryPutFuncIndex => {
                 let (uref_ptr, uref_size, key_bytes_ptr, key_bytes_size, value_ptr, value_ptr_size): (_, u32, _, u32, _, u32) = Args::parse(args)?;
                 self.charge_host_function_call(
-                    &host_function_costs.dictionary_put,
+                    &host_function_costs.write_local,
                     [
-                        uref_ptr,
-                        uref_size,
+                        // uref_ptr,
+                        // uref_size,
```
dwerner, line 1024: Same question as above.

execution_engine/src/core/runtime/externals.rs
```diff
@@ -992,10 +996,10 @@ where
                     _,
                 ) = Args::parse(args)?;
                 self.charge_host_function_call(
-                    &host_function_costs.dictionary_get,
+                    &host_function_costs.read_value_local,
```
dwerner, line 999: I see there's a commit `Revert backwards incompatible changes` where you reverted your approach of `s/local/dictionary/g` for several field and type names in `HostFunctionCosts` among other places. Is this due to `serde`'s `Serialize` and `Deserialize` traits? If so, could we leverage `serde`'s `#[serde(alias = "...")]` attribute here?

e.g.
```
#[derive(Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Debug, DataSize)]
pub struct HostFunctionCosts {
...
    #[serde(alias = "read_local")]
    pub dictionary_get: HostFunction<[Cost; 4]>,
...
}
```
This would preserve the ability to deserialize from the old name, while moving the changing the field name to reflect the `dictionary` name.

execution_engine/src/core/runtime/externals.rs
```diff
@@ -1014,10 +1018,10 @@ where
             FunctionIndex::DictionaryPutFuncIndex => {
                 let (uref_ptr, uref_size, key_bytes_ptr, key_bytes_size, value_ptr, value_ptr_size): (_, u32, _, u32, _, u32) = Args::parse(args)?;
                 self.charge_host_function_call(
-                    &host_function_costs.dictionary_put,
+                    &host_function_costs.write_local,
```
dwerner, line 1021: See comment above about possibly using `#[serde(alias = "...")]` here.

dwerner (changes requested):
This is looking pretty good! I made a comment about a reverted change to `s/local/dictionary/g`. Hopefully we are able to pass the regression test with something like `#[serde(alias = "...")]` on struct fields.

types/src/key.rs
```diff
@@ -319,12 +341,33 @@ impl Key {
...
+
+    /// Creates a new [`Key::Local`] variant based on a `uref` and a `key` bytes.
+    pub fn dictionary(uref: URef, key: &[u8]) -> Key {
+        // NOTE: Expect below is safe because the length passed is supported.
+        let mut hasher = VarBlake2b::new(BLAKE2B_DIGEST_LENGTH).expect("should create hasher");
+        hasher.update(uref.addr().as_ref());
+        hasher.update(key);
+        // NOTE: Assumed safe as size of `HashAddr` equals to the output provided by hasher.
+        let mut local_addr = HashAddr::default();
+        hasher.finalize_variable(|hash| local_addr.clone_from_slice(hash));
```
henrytill, line 368: I would prefer to use the `Blake2bHash` and its `new` method here, even if we are "unwrapping" from the `Blake2bHash` newtype itself when we write the dictionary key.  That newtype was designed to abstract over this operation, and prevents us from duplicating (and/or getting wrong) the hashing logic.

smart_contracts/contracts/test/dictionary-call/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+doctest = false
+test = false
+
+[features]
+std = ["casper-contract/std", "casper-types/std"]
+
+[dependencies]
+casper-contract = { path = "../../../contract" }
+casper-types = { path = "../../../../types" }
+dictionary = { path = "../dictionary" }
```
henrytill, line 20: Add trailing newline

Also:
https://stackoverflow.com/questions/44704968/visual-studio-code-insert-new-line-at-the-end-of-files

:wink:

henrytill (changes requested):
Coupla small things.  Also, could you respond to @dwerner's review from last week?

types/src/key.rs
mpapierski: I used inclusive range as it felt more natural to use `0..=9` when there are exactly 9 match branches

smart_contracts/contracts_as/test/local-state/assembly/tsconfig.json
mpapierski: This file is copy-pasted from other AS contracts. If there's a better way to use tsconfig.json files I'd be happy to do a separate pass at addressing it in #1524

execution_engine/src/core/runtime/externals.rs
mpapierski: Fair point, changed

execution_engine/src/core/runtime/externals.rs
mpapierski: Removed - when we'll change host function costs to increase arity compiler will complain here

execution_engine/src/core/runtime_context/local_key.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Wraps a [`CLValue`] and extends it with a seed [`URef`] and original key bytes.
+#[derive(Clone)]
+pub struct LocalKeyValue {
+    /// Actual [`CLValue`] written to global state.
+    cl_value: CLValue,
+    /// Original key bytes.
+    key_bytes: Bytes,
+    /// [`URef`] seed bytes.
+    seed_address: Bytes,
```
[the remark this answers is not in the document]
mpapierski: Addressed

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -448,6 +451,18 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
 
             let value = stored_value.value().to_owned();
 
+            // Following code does a patching on the `StoredValue` that unwraps an inner
+            // `LocalKeyValue` for local keys only.
+            let value = match local_key::monkey_patch(query.current_key, value) {
```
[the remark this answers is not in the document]
mpapierski: Addressed with `handle_stored_value`

types/src/key.rs
mpapierski: Unfortunately `Blake2bHash` is in ./execution_engine, but this code is in ./types. I could use the `account::blake2b` function, or do we want to move the `Blake2bHash` struct into the `types` crate as part of this PR?

smart_contracts/contracts/test/dictionary-call/Cargo.toml
mpapierski: Fixed (and reconfigured my IDE)

smart_contracts/contracts_as/test/local-state/assembly/tsconfig.json
dwerner: When in rome I guess, thanks for answering. I suppose this is dependent on a certain dev environment setup for ts as well, one using nvm.

dwerner (approved):

types/src/key.rs
henrytill: Tragic, let's keep as-is for now

smart_contracts/contracts/test/dictionary-call/src/bin/main.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use dictionary::{DEFAULT_DICTIONARY_NAME, DEFAULT_DICTIONARY_VALUE};
+use dictionary_call::{
+    Operation, ARG_CONTRACT_HASH, ARG_FORGED_UREF, ARG_OPERATION, ARG_SHARE_UREF_ENTRYPOINT,
+    NEW_DICTIONARY_NAME, NEW_DICTIONARY_VALUE,
+};
+
+/// Calls local state contract by hash as passed by `ARG_CONTRACT_HASH` argument and returns a
+/// single value.
+fn call_local_state_contract<T: CLTyped + FromBytes>(entrypoint: &str) -> T {
```
henrytill, line 23:
```suggestion
fn call_dictionary_test_contract<T: CLTyped + FromBytes>(entrypoint: &str) -> T {
```

henrytill (commented):
Almost there!

execution_engine/src/core/runtime/externals.rs
```diff
@@ -968,6 +972,61 @@ where
...
                 Ok(Some(RuntimeValue::I32(0)))
             }
+            FunctionIndex::NewDictionaryFuncIndex => {
+                let (output_size_ptr,): (u32,) = Args::parse(args)?;
+
+                self.charge_host_function_call(
+                    &DEFAULT_HOST_FUNCTION_NEW_DICTIONARY,
+                    [output_size_ptr],
+                )?;
+                let ret = self.create_local(output_size_ptr)?;
```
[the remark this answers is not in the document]
henrytill: Addressed

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3345,6 +3346,84 @@ where
         self.gas(cost)?;
         Ok(())
     }
+
+    fn create_local(&mut self, output_size_ptr: u32) -> Result<Result<(), ApiError>, Error> {
```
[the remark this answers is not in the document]
henrytill: Addressed

smart_contracts/contracts/test/dictionary-call/src/bin/main.rs
henrytill: Addressed

execution_engine/src/core/runtime/externals.rs
```diff
@@ -968,6 +972,61 @@ where
                 self.record_era_info(era_id, era_info)?;
                 Ok(Some(RuntimeValue::I32(0)))
             }
+            FunctionIndex::NewDictionaryFuncIndex => {
+                let (output_size_ptr,): (u32,) = Args::parse(args)?;
```
EdHastingsCasperLabs, line 976: I appreciate args description comments in these methods, such as:
// args(0) = _description of what this arg is supposed to be_

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3345,6 +3346,84 @@ where
...
+        uref_ptr: u32,
+        uref_size: u32,
+        key_bytes_ptr: u32,
+        key_bytes_size: u32,
+        output_size_ptr: u32,
+    ) -> Result<Result<(), ApiError>, Trap> {
+        if !self.can_write_to_host_buffer() {
+            // Exit early if the host buffer is already occupied
+            return Ok(Err(ApiError::HostBufferFull));
+        }
```
EdHastingsCasperLabs, line 3387: why not use the same `if let Err(err) = self.check_host_buffer() {            return Ok(Err(err));        }` pattern as in above method?

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3345,6 +3346,84 @@ where
...
+
+        let value_bytes = value_size.to_le_bytes(); // Wasm is little-endian
+        if let Err(error) = self.memory.set(output_size_ptr, &value_bytes) {
+            return Err(Error::Interpreter(error.into()).into());
+        }
+
+        Ok(Ok(()))
+    }
+
+    /// Writes `value` under a key derived from `key` in the dictionary partition of GlobalState
```
EdHastingsCasperLabs, line 3410: If I didn't already know what you meant, this comment would be useless to me. Consider expounding on it a bit to be less inscrutable?

execution_engine/src/core/runtime_context/dictionary.rs
```diff
@@ -0,0 +1,106 @@
+use casper_types::{
+    bytesrepr::{self, Bytes, FromBytes, ToBytes},
+    CLType, CLTyped, CLValue, CLValueError, Key,
+};
+
+use crate::shared::stored_value::StoredValue;
+
+/// Wraps a [`CLValue`] and extends it with a seed [`casper_types::URef`] and original key bytes.
```
EdHastingsCasperLabs, line 8: Consider explaining _why_ we want to retain the uref_addr bytes and key bytes for posterity.

execution_engine/src/core/runtime/mod.rs
henrytill: Addressed

execution_engine/src/core/runtime/mod.rs
henrytill: Addressed

execution_engine/src/core/runtime_context/dictionary.rs
henrytill: Addressed

execution_engine/src/core/runtime/externals.rs
henrytill: Addressed

henrytill:
bors try

darthsiroftardis (commented):
LGTM!

darthsiroftardis (approved):

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -1028,4 +1053,52 @@ where
...
+        cl_value: CLValue,
+    ) -> Result<(), Error> {
+        self.validate_writeable(&uref.into())?;
+        self.validate_uref(&uref)?;
+
+        self.validate_cl_value(&cl_value)?;
+
+        let wrapped_cl_value =
+            DictionaryValue::new(cl_value, key_bytes.to_vec(), uref.addr().to_vec());
+        let wrapped_cl_value = CLValue::from_t(wrapped_cl_value).map_err(Error::from)?;
```
EdHastingsCasperLabs, line 1098: nitpick: maybe something like

let wrapped_cl_value = { 
        let dictionary_value = DictionaryValue::new(cl_value, key_bytes.to_vec(), uref.addr().to_vec());
        CLValue::from_t(dictionary_value).map_err(Error::from)?
 }; 

or 

let dictionary_value = DictionaryValue::new(cl_value, key_bytes.to_vec(), uref.addr().to_vec());
let wrapped_cl_value = CLValue::from_t(dictionary_value).map_err(Error::from)?;

smart_contracts/contract/src/contract_api/storage.rs
```diff
@@ -335,3 +335,72 @@ pub fn disable_contract_version(
 
     api_error::result_from(result)
 }
+
+/// Creates new [`URef`] that points to a context-local partition of global state.
+/// Optionally you can pass a [`str`] to automatically put this dictionary under named keys.
```
EdHastingsCasperLabs, line 340: This comment is incorrect; "context-local partition" is a misleading / opaque explanation, and the &str argument isn't optional.

smart_contracts/contract/src/contract_api/storage.rs
```diff
@@ -335,3 +335,72 @@ pub fn disable_contract_version(
...
+pub fn new_dictionary(key_name: &str) -> Result<URef, ApiError> {
```
EdHastingsCasperLabs, line 341: i think `dictionary_name` is a less confusing name for this variable...

smart_contracts/contract/src/contract_api/storage.rs
```diff
@@ -335,3 +335,72 @@ pub fn disable_contract_version(
...
+    let uref: URef = bytesrepr::deserialize(value_bytes).unwrap_or_revert();
+    runtime::put_key(key_name, Key::from(uref));
+    Ok(uref)
+}
+
+/// Reads data from dictionary pointed at by [`URef`].
+pub fn dictionary_get<K: ToBytes, V: CLTyped + FromBytes>(
+    uref: URef,
+    key: K,
+) -> Result<Option<V>, bytesrepr::Error> {
```
EdHastingsCasperLabs, line 362: I think the comment is unhelpful. Maybe something like...
/// Retrieve value stored under `key` in the dictionary accessed by `dictionary_uref`.

smart_contracts/contract/src/contract_api/storage.rs
```diff
@@ -335,3 +335,72 @@ pub fn disable_contract_version(
...
+            Err(e) => runtime::revert(e),
+        }
+    };
+
+    let value_bytes = runtime::read_host_buffer(value_size).unwrap_or_revert();
+    Ok(Some(bytesrepr::deserialize(value_bytes)?))
+}
+
+/// Writes `value` under `key` in the context-local partition of global state.
+pub fn dictionary_put<K: ToBytes, V: CLTyped + ToBytes>(uref: URef, key: K, value: V) {
```
EdHastingsCasperLabs, line 389: Similar desire to see improved clarity in the comment and variable naming.

EdHastingsCasperLabs (approved):

henrytill (approved):

PR #1472, opened 2021-05-22
Move finality signatures to separate event stream

### LINKED ISSUE #1452: Move finality signatures to a separate event stream

Fraser999:
This PR splits the SSE event stream into two discrete streams.  One stream emits only `FinalitySignature`s, the other emits all remaining event types.

The URL for the former is now `<IP:Port>/events/sigs` and the latter `<IP:Port>/events/main`.  The previous URL will now return a 404 with a body of `invalid path: expected 'events/main' or 'events/sigs'`.

The events continue to be indexed globally rather than per stream.  So for example, the main stream could yield event IDs of 1, 2, 4, 6.  In this case events 3 and 5 are `FinalitySignature`s filtered out of the main stream.  This allows consumers of both streams to interleave the events if desired, to recreate the original ordering.

Tagging @sacherjj @TomVasile @momipsl to draw attention to the changed URL path.

Closes #1452.

### DIFF
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -2,6 +2,10 @@
 
 use datasize::DataSize;
 use futures::{Stream, StreamExt};
+use http::status::StatusCode;
+use hyper::Body;
+#[cfg(test)]
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use tokio::sync::{
     broadcast::{self, error::RecvError},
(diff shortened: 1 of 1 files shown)


sacherjj:
@Fraser999 is this landing with 1.3 or 1.4?

darthsiroftardis (approved):
LGTM!

Fraser999:
@sacherjj I don't think there's a rush for this?  I'd suggest 1.4 personally.

Fraser999:
bors merge

Issue #1479, opened 2021-05-24
Create test plan for bandwidth limiting


marc-casperlabs:
We need a plan that describes how to exercise the new bandwidth limiting feature, taking into account the specs of our testing nodes.

The core issue here is to find the minimum amount of data and nodes that need to be active to actually overload the network, as this will likely stretch our testing capabilities. An alternative is to artificially shrink nodes to get the same effect.

PR #1494, opened 2021-05-26
Improve bandwidth limiting testing and edge cases

marc-casperlabs:
Contains a few minor fixes that are most prominent in testing:

* Disables bandwidth limiting when there is no information about validators at all. This can, for example, be triggered by setting absurdly low bandwidth limits and then starting a new network from genesis, stalling immediately.
* Respect existing known era IDs that are available. This is, due to the fix above, only a minor improvement, causing us to not having to wait until the next switch block.
* Log changes in known validator sets. This makes it possible to inspect traffic shaping behavior at all.

Changes in preparation to #1479.


fizyk20 (approved):
Looks OK to me :+1:

Issue #1498, opened 2021-05-26
Stalled node continues to run


Fraser999:
The stalled node 2 in #1496 ran for 22 minutes in a stalled state (stuck at the switch block immediately before the upgrade).  I was expecting that after being stalled for 10 minutes that consensus would trigger a shutdown.

(Related: #1501)

afck:
In principle the standstill alert works. I set
```toml
maximum_round_exponent = 12
```
in the chainspec, so the maximum round length would be 4 seconds. A node counts as offline after being silent for [three maximum-length rounds](https://github.com/casper-network/casper-node/blob/a2d6511d17aa7df75a633542d207c280a775248e/node/src/components/consensus/highway_core/state.rs#L55), i.e. after 12 seconds in this test.

I also set
```toml
standstill_timeout = "1min"
```
in the config of node 1.

I then started the network, stopped nodes 2 and 3, and restarted node 1:
```
$ nctl-start
$ nctl-stop node=2
$ nctl-stop node=3

$ nctl-view-chain-lfb
2021-05-31T11:59:01.402070 [INFO] [51782] NCTL :: last finalized block hash @ node-1 = f2556c54efa4216882157b8359e911ca137c21efbfa87f2f8c8c9cedd59ed1cd
2021-05-31T11:59:01.497436 [INFO] [51782] NCTL :: last finalized block hash @ node-2 = 'N/A'
2021-05-31T11:59:01.554904 [INFO] [51782] NCTL :: last finalized block hash @ node-3 = 'N/A'
2021-05-31T11:59:01.644949 [INFO] [51782] NCTL :: last finalized block hash @ node-4 = f2556c54efa4216882157b8359e911ca137c21efbfa87f2f8c8c9cedd59ed1cd
2021-05-31T11:59:01.800020 [INFO] [51782] NCTL :: last finalized block hash @ node-5 = f2556c54efa4216882157b8359e911ca137c21efbfa87f2f8c8c9cedd59ed1cd
2021-05-31T11:59:01.884051 [INFO] [51782] NCTL :: last finalized block hash @ node-6 = 'N/A'
2021-05-31T11:59:02.019905 [INFO] [51782] NCTL :: last finalized block hash @ node-7 = 'N/A'
2021-05-31T11:59:02.075080 [INFO] [51782] NCTL :: last finalized block hash @ node-8 = 'N/A'
[...]

PR #1500, opened 2021-05-27
Log more progress/standstill alert information.

afck:
This makes it more visible in the logs when a node periodically checks whether its protocol state is stalled.

#1498


goral09 (approved):

fizyk20 (approved):

afck:
bors merge

PR #1503, opened 2021-05-27
Drain joiner reactor before joining

marc-casperlabs:
Drains the queue, to eliminate stale connections, whose handshake has completed, but which currently live on the queue.


Fraser999 (approved):

node/src/app/cli.rs
```diff
@@ -228,7 +228,15 @@ impl Cli {
                     ReactorExit::ProcessShouldContinue => info!("finished joining"),
                 }
 
-                let config = joiner_runner.into_inner().into_validator_config().await?;
+                let (joiner_reactor, joiner_queue) = joiner_runner.into_inners();
+                let config = joiner_reactor.into_validator_config().await?;
+
+                // At this point, the joiner is shut down, so we clear the queue to ensure any
+                // connections whose handshake completed but have not been registered are dropped.
```
goral09, line 235: Isn't dropping the event queue achieving the same effect?

marc-casperlabs: No, we only keep handles to it unfortunately.

Issue #1470, opened 2021-05-21
EE: call_contract does not enforce Group access restriction or args type checking


zie1ony:
call_versioned_contract has the correct functionality, but call_contract does not

this is a serious userland security violation

change call_contract to enforce Group access and args type checking, exactly as call_versioned_contract

EdHastingsCasperLabs:
something like this...

Index: execution_engine/src/core/runtime/mod.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/execution_engine/src/core/runtime/mod.rs b/execution_engine/src/core/runtime/mod.rs
--- a/execution_engine/src/core/runtime/mod.rs	(revision da04b479901cc51af7df32210b66dae58b65fcb8)
+++ b/execution_engine/src/core/runtime/mod.rs	(date 1621619648159)
@@ -1800,16 +1800,17 @@
             .cloned()
             .ok_or_else(|| Error::NoSuchMethod(entry_point_name.to_owned()))?;
 
-        let context_key = self.get_context_key_for_contract_call(contract_hash, &entry_point)?;
+        let contract_package_hash = contract.contract_package_hash().into();
 
-        self.execute_contract(
-            key,
-            context_key,
-            contract,
-            args,
-            entry_point,
-            self.context.protocol_version(),
-        )
+        let contract_package = match self.context.read_gs(&key)? {
+            Some(StoredValue::ContractPackage(contract_package)) => contract_package,
+            Some(_) => {
+                return Err(Error::InvalidContractPackage(contract_package_hash));
+            }
+            None => return Err(Error::KeyNotFound(key)),
+        };
+
+        self.call_contract_checked(contract_package, contract, entry_point, args)
     }
 
     /// Calls `version` of the contract living at `key`, invoking `method` with
@@ -1827,10 +1828,7 @@
         let contract_package = match self.context.read_gs(&key)? {
             Some(StoredValue::ContractPackage(contract_package)) => contract_package,
             Some(_) => {
-                return Err(Error::FunctionNotFound(format!(
-                    "Value at {:?} is not a versioned contract",
-                    contract_package_hash
-                )));
+                return Err(Error::InvalidContractPackage(contract_package_hash));
             }
             None => return Err(Error::KeyNotFound(key)),
         };
@@ -1868,8 +1866,22 @@
             .cloned()
[...]

PR #1504, opened 2021-05-27
GH-1470: Group restriction fix

mpapierski:
Ref #1470 

This PR is an improved version of #1485 (reverted in #1495) with an emphasis on upgrade testing. It introduces the concept of upgrade regression testing by persisting a global state from a recent release into the version control and reusing it while testing. This way EE can iterate faster to detect post-upgrade regressions.

The fix itself consists of changes to how groups are verified in `call_contract` API call and additionally makes improvements arguments types/length verification. For example, an entry point defined with `CLType::Option` parameters makes this argument not checked.


execution_engine/src/core/runtime/mod.rs
```diff
@@ -1852,36 +1846,123 @@ where
...
-            .map(|a| a.cl_type())
-            .cloned()
-            .zip(args.to_values().into_iter().map(|v| v.cl_type()).cloned())
+        // This will skip arguments check for system contracts only. This code should be removed on
+        // next major version bump. Argument checks for system contract is still done during
+        // execution of a system contract.
+        if !self
+            .protocol_data()
+            .system_contracts()
+            .contains(&contract_hash)
```
mpapierski, line 1885: @henrytill this line skips entry point arguments check for system contracts only. We can't upgrade entry points at the next minor/patch version bump without changing the upgrade procedure, so I figured the safest thing to do is just to not check the arguments for these until the next major version bump. This PR is ready for the next major bump (which will install fixed entrypoints) and has regression tests in place to verify it.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -1852,36 +1846,118 @@ where
...
+            // Unused args are the ones that are passed, but are not present as required nor
+            // optionals.
+            let unused_args = &(&passed_args - &required_args) - &optional_args;
+
+            if !unused_args.is_empty() {
+                return Err(Error::UnusedArgumentsFound {
+                    required: required_args.len(),
+                    unused: unused_args.len(),
+                });
+            }
```
EdHastingsCasperLabs, line 1925: I don't think it is safe for us to error in this case; some 3rd parties may tail their passed args with extra marker or readback data that is not needed for the called logic but is used for some off chain post process triggered by inclusion of the deploy in a block.

EdHastingsCasperLabs (approved):
Generally LGTM, but see note on unused params

execution_engine/src/core/runtime/mod.rs
mpapierski: Fair point. I removed this error condition, and added an extra test to ensure contracts can accept arbitrary arguments (not listed in entry point as extra markers; for example)

henrytill:
I would like to rebase this branch against #1627 so that it is aligned with that work.

henrytill (changes requested):
I would like to rebase this branch against #1627 so that it is aligned with that work.

henrytill (commented):
Otherwise looks good.

PR #1508, opened 2021-05-28
Skip storage integrity checks if the node didn't previously crash

### LINKED ISSUE #1507: Optimise, background or skip initialize_block_metadata_db
During a recent network upgrade, the nodes took a significant time to re-establish block production.  A major factor seems to be a call to [`storage::initialize_block_metadata_db()`](https://github.com/casper-network/casper-node/blob/dev/node/src/components/storage.rs#L1319-L1357) as evidenced by the logs below:

```json
{"timestamp":"May 28 13:01:05.432","level":"INFO","fields":{"message":"initializing block metadata database"},"target":"casper_node::components::storage"}
{"timestamp":"May 28 13:05:58.609","level":"INFO","fields":{"message":"block metadata database initialized"},"target":"casper_node::components::storage"}
```

We should probably only execute this check if the node has just restarted after a crash.

Fraser999:
The behaviour where we usually skip integrity checks of global state has been extended to the storage component in this PR.  The integrity checks are now only performed if the node was detected to have crashed on the previous run.

Closes #1507.

### DIFF
--- a/node/src/components/fetcher/tests.rs
+++ b/node/src/components/fetcher/tests.rs
@@ -76,6 +76,7 @@ reactor!(Reactor {
             &WithDir::new(cfg.temp_dir.path(), cfg.storage_config),
             chainspec_loader.hard_reset_to_start_of_era(),
             ProtocolVersion::from_parts(1, 0, 0),
+            false,
         );
         deploy_acceptor = infallible DeployAcceptor(cfg.deploy_acceptor_config, &*chainspec_loader.chainspec());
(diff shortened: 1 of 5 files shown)


marc-casperlabs (approved):
Ideally we wanted each component to manage its own crash-state and this is sort of pulling us further from that goal, but given the current code state, this should certainly do the trick :)

tl;dr LGTM

goral09 (approved):

Fraser999:
bors merge

PR #1513, opened 2021-05-29
Shut down sse event streams gracefully

### LINKED ISSUE #1512: Properly terminate event stream on shutdown
The SSE server should be updated so that it terminates the stream properly rather than returning an error.

This should also have the effect of avoiding needless logging errors produced by the `warp` crate.

Fraser999:
This PR causes SSE event streams to be shut down gracefully rather than returning an error to the client.

For example, if `curl` is used to follow an event stream as a node stops for an upgrade, the stream exits with an error, reported to the user as `curl: (18) transfer closed with outstanding read data remaining` and `curl` exits with code `18`.  With the update in this PR, `curl` now exits with `0` and reports no error.

This also has the effect of avoiding `warp` logging errors on shutdown of the SSE server when client(s) are attached.

Closes #1512.

### DIFF
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -1,7 +1,7 @@
 //! Types and functions used by the http server to manage the event-stream.
 
 use datasize::DataSize;
-use futures::{Stream, StreamExt};
+use futures::{future, Stream, StreamExt};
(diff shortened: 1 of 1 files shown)


marc-casperlabs (approved):
Looks good to me!

Fraser999:
bors merge

PR #1518, opened 2021-06-01
Reference counted network messages

marc-casperlabs:
This removes the in-memory copying of outgoing network messages, saving a substantial amount of memory in the case of broadcasts and, to some extent, gossips.

Splits the formerly symmetrical networking transport into an outgoing one that shares references, while the incoming stream is still value based.


goral09 (approved):
The code looks good, I assume you've run it and tested.

marc-casperlabs:
> The code looks good, I assume you've run it and tested.

Yes, tests and one `nctl` run.

Fraser999 (approved):
LGTM!

Issue #1487, opened 2021-05-24
Easy ProtocolData upgradability


mpapierski:
Follow up a ticket for #1309 

Code in #1309 includes a new field in the middle of `ProtocolData`'s nested structures. Unfortunately, as for now upgrade code does first read `ProtocolData` with the current version here:

https://github.com/casper-network/casper-node/blob/b1e9e78c72127e844e78f3bcc4718a68a911572b/execution_engine/src/core/engine_state/mod.rs#L235-L245

This will be problematic if, say, 1.1 upgrades to 1.2, adding or removing a field. 

This task is about introducing a migration code where instead we serialize and deserialize all ProtocolData fields with a BTreeMap-like format with keys as enums, and do post-processing to handle missing fields in a similar fashion to what protocol buffers does.

piotr-dziubecki:
@Fraser999 -> @darthsiroftardis - is this still relevant?

PR #1531, opened 2021-06-02
GH-1487: Protocol data upgrade

mpapierski:
Ref #1487 

This PR handles migration of the old "legacy" ProtocolData format during upgrade into a new extensible format. It will read the legacy format for protocol versions <=1.2.0, and it will read the legacy format, and write the new format. New networks starting with genesis will always serialize new format regardless of the genesis protocol version.

- Serializes into a `BTreeMap<u64, Vec<u8>>`-compatible format with a dedicated API to give more type safety - encourages the use of enums with a From/ToPrimitive
- Ability to skip unknown fields (i.e., if a given field is removed/deprecated)
- Adding new fields enforces using a default instance of deserialized objects for an implicit migration
- Example test for versioned structs with migration steps https://github.com/mpapierski/casper-node/blob/26d84ad73ea7eb75a4fa6d3c71447dfc7393c159/types/src/bytesrepr.rs#L1321-L1449


execution_engine/src/legacy/protocol_data.rs
```diff
@@ -0,0 +1,65 @@
...
+};
+
+use crate::storage::protocol_data::ProtocolData;
+
+use self::{system_config::LegacySystemConfig, wasm_config::LegacyWasmConfig};
+
+use super::types::LegacyHashAddr;
+
+/// Protocol versions lower than or equal to this value should be deserialized with a legacy format.
+pub const LEGACY_PROTOCOL_DATA_VERSION: ProtocolVersion = ProtocolVersion::from_parts(1, 2, 0);
```
xcthulhu, line 20: This should probably be switched via a compile-time feature.  CC @fizyk20

mpapierski:
Closed in favor of #1814

Issue #1424, opened 2021-05-10
Use different namespaces for different networks.


piotr-dziubecki:
Currently a node uses a single directory to hold its LMDB database files and various "unit" files created by consensus.  To avoid problems if a node is run against e.g. the testnet and later the mainnet, we need to create a subdirectory in the storage folder which is named as per the given network, i.e. it will use the [`chainspec.network.name`](https://github.com/casper-network/casper-node/blob/85a7c37793d2bee48931f89e3b6dc216a700935d/resources/production/chainspec.toml#L22) value for the subdir name.

There are a few aspects to this task:

* need to pass the network name in via [`Storage::new()`](https://github.com/casper-network/casper-node/blob/85a7c37793d2bee48931f89e3b6dc216a700935d/node/src/components/storage.rs#L250-L255).  This will be done similarly to the `protocol_version`: it's also specified in the chainspec, and the initializer reactor (which constructs the `Storage` component) [accesses the chainspec by calling the `ChainspecLoader` component](https://github.com/casper-network/casper-node/blob/85a7c37793d2bee48931f89e3b6dc216a700935d/node/src/reactor/initializer.rs#L197).
* the [`ContractRuntime` component's constructor](https://github.com/casper-network/casper-node/blob/85a7c37793d2bee48931f89e3b6dc216a700935d/node/src/components/contract_runtime.rs#L655-L662) should be changed.  Currently it takes a copy of the Storage component's config (`storage_config: WithDir<StorageConfig>`), but only needs that to deduce the path in which to create its own LMDB files.  We should replace that variable with a simple `PathBuf` or `&Path` which will provided via a new getter on the `Storage` component.  The initializer reactor has access to the newly-constructed `Storage` just before it constructs the `ContractRuntime`.
[...]

PR #1532, opened 2021-06-02
Storage creates a subdir named after network.

nilz3ro:
# About These Changes

* changed `Storage` constructor to accept a network name.
* changed `Storage` constructor to move specific lmdb files to a subdirectory named after the network.
* changed the initializer reactor to pass the `Chainspec`'s network name to the `Storage` constructor.
* changed the `ContractRuntime`'s constructor to accept a path to the storage dir directly, instead of using a `WithDir<T>` 
* updated tests that construct `Storage`
* updated tests that construct `ContractRuntime`
* added unit tests for the helper functions that run in the `Storage` constructor.
* closes casper-network/casper-node#1424


node/src/components/storage.rs
```diff
@@ -1153,6 +1173,29 @@ fn insert_to_deploy_index(
...
+fn move_storage_files_to_network_subdir(
+    root: &PathBuf,
+    subdir: &PathBuf,
+    file_names: &[&str],
+) -> Result<(), Error> {
+    file_names
+        .iter()
+        .map(|file_name| (root.join(file_name), subdir.join(file_name)))
+        .map(|(src, dest)| {
+            fs::rename(src.clone(), dest.clone()).map_err(|_| Error::UnableToMoveFile(src, dest))
```
nilz3ro, line 1192: Note: I actually do want to preserve the original fs error here, I could use some advice on what the cleanest way to get that into the Thiserror enum.

node/src/components/storage.rs
```diff
@@ -252,16 +263,32 @@ impl Storage {
...
-                .map_err(|err| Error::CreateDatabaseDirectory(root.clone(), err))?;
+        let mut root = cfg.with_dir(config.path.clone());
+        let network_subdir = root.join(network_name);
+
+        if !network_subdir.exists() {
+            fs::create_dir_all(&network_subdir)
+                .map_err(|err| Error::CreateDatabaseDirectory(network_subdir.clone(), err))?;
+        }
+
+        let lmdb_files = [
```
Fraser999, line 279: We normally define things like this as `const` variables towards the top of the file (after the imports).

That would also mean it wouldn't need passed in as an arg to `should_move_storage_files_to_network_subdir()` or `move_storage_files_to_network_subdir()`.

node/src/components/storage.rs
```diff
@@ -1153,6 +1180,35 @@ fn insert_to_deploy_index(
     Ok(())
 }
 
+fn should_move_storage_files_to_network_subdir(root: &PathBuf, file_names: &[&str]) -> bool {
+    file_names
```
Fraser999, line 1184: This nice and neat, but unfortunately I think it'd be good to log a warning for the case where some but not all of these files exist.  The warning should give details of which files exist and which do not.

node/src/components/storage.rs
```diff
@@ -1153,6 +1180,35 @@ fn insert_to_deploy_index(
...
+}
+
+fn move_storage_files_to_network_subdir(
+    root: &PathBuf,
+    subdir: &PathBuf,
+    file_names: &[&str],
+) -> Result<(), Error> {
+    file_names
+        .iter()
+        .map(|file_name| (root.join(file_name), subdir.join(file_name)))
```
Fraser999, line 1197: I think rather than calling `map()` twice, it might be simpler and more readable to just do e.g.

```rs
        .map(|file_name| {
            let source_path = root.join(file_name);
            let dest_path = subdir.join(file_name);
            fs::rename(&source_path, &dest_path).map_err(|original_error| Error::UnableToMoveFile {
```
Also note that (with or without the change suggested above) we can avoid cloning the `PathBuf`s by passing refs in `fs::rename()`.

node/src/components/storage/tests.rs
```diff
@@ -1144,3 +1153,112 @@ fn should_hard_reset() {
...
+    .unwrap();
+
+    let expected_path = cfg.path.join("test");
+
+    assert!(expected_path.exists());
+    assert_eq!(expected_path, storage.root_path());
+}
+
+#[test]
+fn should_not_move_nonexistent_files() {
```
Fraser999, line 1178:
```suggestion
fn should_not_try_to_move_nonexistent_files() {
```

node/src/components/storage/tests.rs
```diff
@@ -1144,3 +1153,112 @@ fn should_hard_reset() {
...
+    let src_path1 = root.join(file_names[0]);
+    let src_path2 = root.join(file_names[1]);
+    let src_path3 = root.join(file_names[2]);
+    let dest_path1 = subdir.join(file_names[0]);
+    let dest_path2 = subdir.join(file_names[1]);
+    let dest_path3 = subdir.join(file_names[2]);
+
+    // Storage will create this in the constructor,
+    // doing this manually since we're not calling the constructor in this test.
+    fs::create_dir(root.clone()).unwrap();
```
Fraser999, line 1243: You could use `fs::create_dir_all()` here.  It probably doesn't matter that it doesn't return a `Result` since you're asserting the files exist before trying to move them.

Fraser999 (commented):
Mostly looks good to me.  A couple of nitpicks, one of which will cause the tests to need refactored slightly.  Otherwise, very nice work.

node/src/components/storage.rs
nilz3ro: Ohh for some reason I thought it had to take ownership. I didn't know it could borrow

nilz3ro:
NOTE: This should not be merged until version `1.4.0`.

Fraser999 (approved):

PR #1534, opened 2021-06-03
OP-2072: rename log_step used in upgrade scenarios

TomVasile:
Changes the name of the `log_step` function used by the nctl upgrade scenarios. 

See Ticket: https://casperlabs.atlassian.net/browse/OP-2072


momipsl (approved):
Looks good.

PR #1539, opened 2021-06-07
Add SSE server tests and fix issues exposed by these

### LINKED ISSUE #1525: Add tests for SSE server

Fraser999:
This PR adds a comprehensive test suite for the SSE server.  In doing so, it revealed two issues in the server, both of which are fixed in this PR:
* The server did not handle malformed queries well.  In some cases it ignored them and in others it treated them as a bad path
* The server did not handle the `start_from` query properly in the case where the buffered events' IDs wrapped past `u32::MAX` back to `0`

The PR also introduces a test which checks the generated JSON schema for the `SseData` enum against a previously-generated, hard-coded schema (added to "resources/test").  Rather than being particularly useful as a unit test, it will instead make it somewhat simple to view differences between discrete versions of the data produced by the SSE server.

Closes #1525.

### DIFF
--- a/node/src/components/event_stream_server/tests.rs
+++ b/node/src/components/event_stream_server/tests.rs
@@ -1,0 +1,899 @@
+use std::{
+    cell::RefCell,
+    error::Error,
+    fs, io, iter, str,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use futures::{join, StreamExt};
+use http::StatusCode;
+use pretty_assertions::assert_eq;
+use rand::Rng;
+use schemars::schema_for;
+use tempfile::TempDir;
+use tokio::{
+    sync::{Barrier, Notify},
+    task::JoinHandle,
+    time,
+};
+use tracing::debug;
+
+use super::*;
+use crate::{logging, testing::TestRng};
+use sse_server::{
+    EventFilter, Id, QUERY_FIELD, SSE_API_MAIN_PATH, SSE_API_ROOT_PATH, SSE_API_SIGNATURES_PATH,
+};
+
+/// The total number of random events each `EventStreamServer` will emit by default, excluding the
+/// initial `ApiVersion` event.
+const EVENT_COUNT: u32 = 100;
+/// The event stream buffer length, set in the server's config.  Set to half of the total event
+/// count to allow for the buffer purging events in the test.
+const BUFFER_LENGTH: u32 = EVENT_COUNT / 2;
+/// The maximum amount of time to wait for a test server to complete.  If this time is exceeded, the
+/// test has probably hung, and should be deemed to have failed.
+const MAX_TEST_TIME: Duration = Duration::from_secs(1);
+/// The duration of the sleep called between each event being sent by the server.
+const DELAY_BETWEEN_EVENTS: Duration = Duration::from_millis(1);
+
+thread_local!(
+    /// A per-test identifier for the clients.
+    static CLIENT_ID: RefCell<usize> = RefCell::new(0)
+);
+
+/// A helper to allow the synchronization of a single client joining the SSE server.
+///
+/// It provides the primitives to allow the client to connect to the server just before a specific
+/// event is emitted by the server.
+#[derive(Clone)]
+struct ClientSyncBehavior {
+    /// The event ID before which the server should wait at the barrier for the client to join.
+    join_before_event: Id,
+    /// The barrier to sync the client joining the server.
+    barrier: Arc<Barrier>,
+}
+
+impl ClientSyncBehavior {
+    fn new(join_before_event: Id) -> (Self, Arc<Barrier>) {
+        let barrier = Arc::new(Barrier::new(2));
+        let behavior = ClientSyncBehavior {
+            join_before_event,
+            barrier: Arc::clone(&barrier),
+        };
+        (behavior, barrier)
+    }
+}
+
+/// A helper to allow the synchronization of all clients joining a single SSE server.
+#[derive(Clone)]
+struct ServerSyncBehavior {
+    /// Whether the server should have a delay between sending events, to allow a client to keep up
+    /// and not be disconnected for lagging.
+    has_delay_between_events: bool,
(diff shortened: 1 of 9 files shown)


nilz3ro (approved):
I think these changes look good, but I would like another person to review them since I'm not familiar with these components.

node/src/components/event_stream_server/tests.rs
```diff
@@ -0,0 +1,899 @@
...
+const BUFFER_LENGTH: u32 = EVENT_COUNT / 2;
+/// The maximum amount of time to wait for a test server to complete.  If this time is exceeded, the
+/// test has probably hung, and should be deemed to have failed.
+const MAX_TEST_TIME: Duration = Duration::from_secs(1);
+/// The duration of the sleep called between each event being sent by the server.
+const DELAY_BETWEEN_EVENTS: Duration = Duration::from_millis(1);
+
+thread_local!(
+    /// A per-test identifier for the clients.
+    static CLIENT_ID: RefCell<usize> = RefCell::new(0)
```
marc-casperlabs, line 45: Is it possibly sufficient to use a random `u64` instead and not deal with any thread local stuff?

node/src/components/event_stream_server/tests.rs
```diff
@@ -0,0 +1,899 @@
...
+    protocol_version: ProtocolVersion,
+    events: Vec<SseData>,
+    first_event_id: Id,
+    server_join_handle: Option<JoinHandle<()>>,
+    server_stopper: ServerStopper,
+}
+
+impl TestFixture {
+    /// Constructs a new `TestFixture` including `EVENT_COUNT` random events ready to be served.
+    fn new(rng: &mut TestRng) -> Self {
```
marc-casperlabs, line 175: Maybe pass in `EVENT_COUNT` here, since we might want to use different settings for different tests?

node/src/components/event_stream_server/tests.rs
```diff
@@ -0,0 +1,899 @@
...
+            first_event_id: 0,
+            server_join_handle: None,
+            server_stopper: ServerStopper::new(),
+        }
+    }
+
+    /// Creates a new `EventStreamServer` and runs it in a tokio task, returning the actual address
+    /// the server is listening on.
+    ///
+    /// Only one server can be run at a time; this panics if there is already a server task running.
```
marc-casperlabs, line 205: Are we limiting ourselves to one instance (and thus one test in parallel) due to port collision issues?

node/src/components/event_stream_server/tests.rs
```diff
@@ -0,0 +1,899 @@
...
+    //   * data:<JSON-encoded ApiVersion> (note, no ID line follows this first event)
+    // then the following three repeated for as many events as are applicable to that stream:
+    //   * data:<JSON-encoded event>
+    //   * id:<integer>
+    //   * empty line
+    // then finally, repeated keepalive lines until the server is shut down.
+    let mut received_events = Vec::new();
+    let mut line_itr = response_text.lines();
+    while let Some(data_line) = line_itr.next() {
+        let data = match data_line.strip_prefix("data:") {
```
marc-casperlabs, line 419: Maybe I've written a few too many parsers in my recent time, but how about adding a function

`fn parse_line<T: FromStr>(expected_key: &str, value: T) -> Result<T, ParseError>` (or a custom trait instead of `FromStr`)

and then writing this more readable as something roughly along the lines of 

```
let data: ApiVersion = parse_line("data", line_itr.next().ok_or(ParseError::ExpectedLine)?)?;
let data: Event = parse_line("data, line_itr.next().ok_or(ParseError::ExpectedLine)?)?;
let data: i64 = parse_line("id, line_itr.next().ok_or(ParseError::ExpectedLine)?)?;
// ...
```

marc-casperlabs (approved):
Some suggestions, but if none were addressed, I'd still be fine merging this :)

node/src/components/event_stream_server/tests.rs
Fraser999: Passing the RNG in here is painful, as in some tests this is called in parallel and the borrow checker has thoughts about that.  I'll pass a `&str` so we can keep with the current approach of having the clients named sequentially to make the log output in the tests more understandable.

node/src/components/event_stream_server/tests.rs
Fraser999: None of the tests need this flexibility, so I'd rather avoid adding an extra parameter.  We can always do it later if we add new tests which might need that.

node/src/components/event_stream_server/tests.rs
Fraser999: No, purely for simplicity in the tests.  The separate unit tests all run in parallel, but I couldn't see a use for running two servers in parallel in a single test.

node/src/components/event_stream_server/tests.rs
Fraser999: Hmm - that seems overly complex for what we need here.  While it would simplify the call site a little, it wouldn't be as simple as you show.  For example, parsing the data line is as simple as stripping the prefix `data:`.  However, we also handle the case where the expected data line is empty or is a keepalive.

I'll move the parsing wholesale to a new function as you suggest in the comment above to see if that helps a little.

Fraser999: Hmm - that seems overly complex for what we need here.  While it would simplify the call site a little, it wouldn't be as simple as you show.  For example, parsing the data line is as simple as stripping the prefix `data:`.  However, we also handle the case where the expected data line is empty or is a keepalive.

I'll move the parsing wholesale to a new function as you suggest in the comment above to see if that helps a little.

Fraser999:
bors merge

node/src/components/event_stream_server/tests.rs
marc-casperlabs: Sounds good to me. In the end, we probably want to have a proper SSE parser and not some quick inline solution I guess though.

node/src/components/event_stream_server/tests.rs
marc-casperlabs: Well, it's just tests, so I am fine with that :)

PR #1540, opened 2021-06-07
Add EffectBuilder::get_consensus_validator_info.

### LINKED ISSUE #1526: Add EffectBuilder::get_consensus_validator_info.
`EffectBuilder::get_consensus_validator_info` should return a map, keyed by the public key of each current and recent validator. The value should be a list of status change events, including the eras in which they occurred:

* Were they added to the validator set?
* Did they equivocate?
* Were they inactive?
* Were they excluded from the proposer sequence?
* Were they banned?
* Were they removed from the validator set?

This is meant to be used in a new endpoint or to add data to the `status` endpoint, to make this information more accessible.

afck:
This returns a list of changes to validators' states, i.e. whether they were added, removed, banned, etc.

Merging to a feature branch since it still requires work on the RPC side.

Closes #1526.

### DIFF
--- a/node/src/effect.rs
+++ b/node/src/effect.rs
@@ -104,7 +104,7 @@ use crate::{
     components::{
         block_validator::ValidatingBlock,
         chainspec_loader::{CurrentRunInfo, NextUpgrade},
-        consensus::{BlockContext, ClContext},
+        consensus::{BlockContext, ClContext, ValidatorChange},
         contract_runtime::EraValidatorsRequest,
         deploy_acceptor,
         fetcher::FetchResult,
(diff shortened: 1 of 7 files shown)


fizyk20 (approved):

goral09:
This functionality could be implemented outside of the node itself – all the information should be in the switch blocks.

afck:
Not quite: When _we_ saw someone equivocate is not in the switch block. But maybe it's not important.

Anyway, the idea was that this would be more convenient for the user.

goral09:
> Not quite: When _we_ saw someone equivocate is not in the switch block. But maybe it's not important.

This event is part of the event stream as well.

node/src/effect.rs
```diff
@@ -1718,6 +1718,18 @@ impl<REv> EffectBuilder<REv> {
...
+    // TODO: Expose this via RPC.
+    pub(crate) async fn _get_consensus_validator_info(
+        self,
+    ) -> BTreeMap<PublicKey, Vec<(EraId, ValidatorChange)>>
+    where
+        REv: From<ConsensusRequest>,
+    {
+        self.make_request(ConsensusRequest::ValidatorInfo, QueueKind::Regular)
+            .await
+    }
```
EdHastingsCasperLabs, line 1731: @piotr-dziubecki , a ticket needs to be created to call this method and wire the resultant data up to the json-rpc & rest status endpoints; assign to @Fraser999 to oversee and @darthsiroftardis to implement for rel 1.4

EdHastingsCasperLabs (approved):

afck:
bors merge

PR #1546, opened 2021-06-07
Flexible protocol serialization scheme

marc-casperlabs:
Hardcodes the handshake, allowing a switchover of the serialization scheme based on the protocol scheme.

There's an open question of what `V2` serialization should be, although the rest is in place.


marc-casperlabs:
This is currently on ice, as long as fast sync is still slated for 1.4.

Issue #1481, opened 2021-05-24
CEP 48: Implement sequence number-based panoramas.


afck:
This is part of [CEP 48](https://github.com/casperlabs/ceps/pull/0048).

afck:
The implementation in https://github.com/casper-network/casper-node/pull/1550 is complete, but merging was postponed because it's a complicated change to an already complicated component. Needs discussion.

PR #1550, opened 2021-06-07
CEP 48 part 1: Sequence number-based panoramas.

afck:
The main component of Highway's protocol state can be viewed as a directed acyclic graph (DAG) of _units_ (the most common kind of Highway message) representing a "happened before" relation. What actually matters to the algorithm is the partial order generated by that graph: Each unit `u` cites other units `v` that `u`'s creator knew about, giving rise to an arrow `u → v` in the DAG. However, `u → v → w` implies that the creator of `u` also knew about `w`. So we write `u > v` in that situation, but also `u > w`, even if `w` is only reachable via multiple arrows. The units form a partial order, and how the generating DAG is represented doesn't matter.

Except for performance: We want to send as little data as necessary over the wire to communicate the [(strict) lower closure](https://en.wikipedia.org/wiki/Upper_set#Upper_closure_and_lower_closure) of `u`, i.e. the set of all `v` with `v < u`.

Currently we send a [panorama](https://github.com/casper-network/casper-node/blob/fa10b20aa8504f60a12e02adde00dee83ee6657b/node/src/components/consensus/highway_core/state/panorama.rs#L102), a list with one entry for each validator, containing the hash of the latest unit `v` by that validator such that `v < u`. A hash is 32 bytes.

This change reduces the size of the panorama in the wire format by replacing the hash of `v` with its sequence number, which is only 8 bytes. So when I receive `u`, and `u → v`, I don't immediately know `v`'s hash, I only know that `v` is e.g. the 42nd unit of that validator.

Since faulty validators can create forks, "the 42nd unit" is not necessarily uniquely defined anymore. A large part of this PR is about double-checking and resolving that ambiguity.

This is part of [CEP 48](https://github.com/casperlabs/ceps/pull/0048), and closes https://github.com/casper-network/casper-node/issues/1481.

### DIFF
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -433,12 +433,14 @@ impl<C: Context> ActiveValidator<C> {
             );
             return None;
         }
+        let previous = panorama[self.vidx].correct().cloned();
         let seq_number = panorama.next_seq_num(state, self.vidx);
         let endorsed = state.seen_endorsed(&panorama);
         let panorama_hash = panorama.hash();
         let hwunit = WireUnit {
             panorama: panorama.clone(),
             panorama_hash,
+            previous,
             creator: self.vidx,
             instance_id,
             value,
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -437,8 +437,9 @@ impl<C: Context> ActiveValidator<C> {
         let seq_number = panorama.next_seq_num(state, self.vidx);
         let endorsed = state.seen_endorsed(&panorama);
         let panorama_hash = panorama.hash();
+        let seq_num_panorama = panorama.to_seq_num_panorama(&state);
         let hwunit = WireUnit {
-            panorama: panorama.clone(),
+            seq_num_panorama,
             panorama_hash,
             previous,
             creator: self.vidx,
@@ -644,7 +645,7 @@ mod tests {
     use tempfile::tempdir;
 
     use crate::components::consensus::highway_core::{
-        highway_testing::TEST_INSTANCE_ID, validators::ValidatorMap,
+        highway::ObservationSeqNum, highway_testing::TEST_INSTANCE_ID, validators::ValidatorMap,
     };
 
     use super::{
@@ -652,7 +653,7 @@ mod tests {
             finality_detector::FinalityDetector,
             state::{tests::*, State, Weight},
         },
-        Vertex, *,
+        *,
     };
 
     type Eff = Effect<TestContext>;
@@ -989,7 +990,6 @@ mod tests {
 
         // Store `a2` unit as the Alice's last unit.
         write_last_unit(&unit_file, &a2, &a2_panorama).expect("storing unit should succeed");
-
         // Alice's last unit is `a2` but `State` is empty. She must synchronize first.
         let (mut alice, alice_init_effects) = ActiveValidator::new(
             ALICE,
@@ -1037,8 +1037,8 @@ mod tests {
             "new unit should have correct seq_number"
         );
         assert_eq!(
-            proposal_wunit.wire_unit().panorama,
-            panorama!(a2.hash()),
+            proposal_wunit.wire_unit().seq_num_panorama[ALICE],
+            ObservationSeqNum::Correct(2),
             "new unit should cite the latest unit"
         );
 
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -435,8 +435,10 @@ impl<C: Context> ActiveValidator<C> {
         }
         let seq_number = panorama.next_seq_num(state, self.vidx);
         let endorsed = state.seen_endorsed(&panorama);
+        let panorama_hash = panorama.hash();
         let hwunit = WireUnit {
             panorama: panorama.clone(),
+            panorama_hash,
             creator: self.vidx,
             instance_id,
             value,
--- a/node/src/components/consensus/highway_core/active_validator.rs
+++ b/node/src/components/consensus/highway_core/active_validator.rs
@@ -74,7 +74,7 @@ where
     /// The path to the file storing the hash of our latest known unit (if any).
     unit_file: Option<PathBuf>,
     /// The last known unit created by us.
-    own_last_unit: Option<SignedWireUnit<C>>,
+    own_last_unit: Option<(SignedWireUnit<C>, Panorama<C>)>,
     /// The target fault tolerance threshold. The validator pauses (i.e. doesn't create new units)
     /// if not enough validators are online to finalize values at this FTT.
     target_ftt: Weight,
@@ -141,22 +141,20 @@ impl<C: Context> ActiveValidator<C> {
     fn can_vote(&self, state: &State<C>) -> bool {
         self.own_last_unit
             .as_ref()
-            .map_or(true, |swunit| state.has_unit(&swunit.hash()))
+            .map_or(true, |(swunit, _)| state.has_unit(&swunit.hash()))
     }
 
     /// Returns whether validator's protocol state is synchronized up until the panorama of its own
     /// last unit.
     pub(crate) fn is_own_last_unit_panorama_sync(&self, state: &State<C>) -> bool {
-        self.own_last_unit.as_ref().map_or(true, |swunit| {
-            swunit
-                .wire_unit()
-                .panorama
+        self.own_last_unit.as_ref().map_or(true, |(_, panorama)| {
+            panorama
                 .iter_correct_hashes()
                 .all(|hash| state.has_unit(hash))
         })
     }
 
-    pub(crate) fn take_own_last_unit(&mut self) -> Option<SignedWireUnit<C>> {
+    pub(crate) fn take_own_last_unit(&mut self) -> Option<(SignedWireUnit<C>, Panorama<C>)> {
         self.own_last_unit.take()
     }
 
@@ -197,16 +195,14 @@ impl<C: Context> ActiveValidator<C> {
                 return effects;
             } else if timestamp == r_id + self.witness_offset(r_len) {
                 let panorama = self.panorama_at(state, timestamp);
-                if let Some(witness_unit) =
-                    self.new_unit(panorama, timestamp, None, state, instance_id)
-                {
+                if let Some(vv) = self.new_unit(panorama, timestamp, None, state, instance_id) {
                     if self
                         .latest_unit(state)
                         .map_or(true, |latest_unit| latest_unit.round_id() != r_id)
                     {
                         info!(round_id = %r_id, "sending witness in round with no proposal");
                     }
-                    effects.push(Effect::NewVertex(ValidVertex(Vertex::Unit(witness_unit))));
+                    effects.push(Effect::NewVertex(vv));
                     return effects;
                 }
             }
@@ -256,10 +252,7 @@ impl<C: Context> ActiveValidator<C> {
         if self.should_send_confirmation(uhash, now, state) {
             let panorama = state.confirmation_panorama(self.vidx, uhash);
             if panorama.has_correct() {
-                if let Some(confirmation_unit) =
-                    self.new_unit(panorama, now, None, state, instance_id)
-                {
-                    let vv = ValidVertex(Vertex::Unit(confirmation_unit));
+                if let Some(vv) = self.new_unit(panorama, now, None, state, instance_id) {
                     effects.push(Effect::NewVertex(vv));
                 }
             }
@@ -311,7 +304,7 @@ impl<C: Context> ActiveValidator<C> {
         if maybe_parent_hash.map_or(false, |hash| state.is_terminal_block(hash)) {
             return self
                 .new_unit(panorama, timestamp, None, state, instance_id)
-                .map(|proposal_unit| Effect::NewVertex(ValidVertex(Vertex::Unit(proposal_unit))));
+                .map(Effect::NewVertex);
         }
         // Otherwise we need to request a new consensus value to propose.
         let ancestor_values = match maybe_parent_hash {
@@ -354,7 +347,7 @@ impl<C: Context> ActiveValidator<C> {
             return vec![];
         }
         self.new_unit(panorama, timestamp, Some(value), state, instance_id)
-            .map(|proposal_unit| Effect::NewVertex(ValidVertex(Vertex::Unit(proposal_unit))))
+            .map(Effect::NewVertex)
             .into_iter()
             .collect()
     }
@@ -413,7 +406,7 @@ impl<C: Context> ActiveValidator<C> {
         value: Option<C::ConsensusValue>,
         state: &State<C>,
         instance_id: C::InstanceId,
-    ) -> Option<SignedWireUnit<C>> {
+    ) -> Option<ValidVertex<C>> {
         if value.is_none() && !panorama.has_correct() {
             return None; // Wait for the first proposal before creating a unit without a value.
         }
@@ -443,7 +436,7 @@ impl<C: Context> ActiveValidator<C> {
         let seq_number = panorama.next_seq_num(state, self.vidx);
         let endorsed = state.seen_endorsed(&panorama);
         let hwunit = WireUnit {
-            panorama,
+            panorama: panorama.clone(),
             creator: self.vidx,
             instance_id,
             value,
@@ -454,13 +447,13 @@ impl<C: Context> ActiveValidator<C> {
         }
         .into_hashed();
         let swunit = SignedWireUnit::new(hwunit, &self.secret);
-        write_last_unit(&self.unit_file, swunit.clone()).unwrap_or_else(|err| {
+        write_last_unit(&self.unit_file, &swunit, &panorama).unwrap_or_else(|err| {
             panic!(
                 "should successfully write unit's hash to {:?}, got {:?}",
                 self.unit_file, err
             )
         });
-        Some(swunit)
+        Some(ValidVertex(Vertex::UnitWithPanorama(swunit, panorama)))
     }
 
     /// Returns a `ScheduleTimer` effect for the next time we need to be called.
@@ -577,7 +570,7 @@ impl<C: Context> ActiveValidator<C> {
             return false;
         }
         match vertex {
-            Vertex::Unit(swunit) => {
+            Vertex::Unit(swunit) | Vertex::UnitWithPanorama(swunit, _) => {
                 // If we already have the unit in our local state,
                 // we must have had created it ourselves earlier and it is now gossiped back to us.
                 self.is_our_unit(swunit.wire_unit()) && !state.has_unit(&swunit.hash())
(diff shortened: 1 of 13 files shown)


fizyk20 (approved):

goral09 (commented):
I'm 1/3 through reviewing your PR but I've got a question.

node/src/components/consensus/highway_core/highway/vertex.rs
```diff
@@ -28,6 +28,11 @@ where
     C: Context,
 {
     Unit(C::Hash),
+    /// The unit with its full panorama.
+    UnitWithPanorama(C::Hash),
+    /// A unit specified by sequence number and validator index, as seen by another unit, given by
+    /// hash.
+    UnitBySeqNum(u64, ValidatorIndex, C::Hash),
```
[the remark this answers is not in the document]
afck: We already have the skip list to find them easily — but only for honest validators, yes.

The problem with omitting the hash is the following situation:
Imagine `v < u` and I already have multiple forks by `v`'s creator in my protocol state, but the unit `u` doesn't happen to know about them: It cites `v` directly.
Now I send you `u`, and since you don't have it yet you request `v` from me — by sequence number!
Since your request doesn't contain `u`'s hash anymore I don't know what it refers to, and I have multiple forks by `v`'s creator, so I can't find the unit. We'd be stuck.

You could argue that I should just pick _any_ unit with the right sequence number. Maybe that would be better; but it would mean that I pick a random fork for every equivocator.

To avoid complicating this feature even further I'd prefer to leave it as is for now, and consider adding such an additional index separately in the future?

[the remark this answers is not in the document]
goral09: > You could argue that I should just pick any unit with the right sequence number. Maybe that would be better; but it would mean that I pick a random fork for every equivocator.

Exactly! And if the end, the panorama's control hash doesn't match I will fall back to requesting the dependencies by hash.

Also, if you know about `V`'s forks but you don't mark `V` as faulty in your latest unit `u`, isn't that an invalid behavior and you will also end up marked as `Faulty`?

node/src/components/consensus/highway_core/highway/vertex.rs
```diff
@@ -53,11 +68,37 @@ where
...
+    ) -> Result<S::Ok, S::Error> {
+        (swunit, panorama).serialize(serializer)
+    }
+
+    pub(super) fn deserialize<'de, D: Deserializer<'de>, C: Context>(
+        deserializer: D,
+    ) -> Result<(SignedWireUnit<C>, Panorama<C>), D::Error> {
+        let (swunit, panorama) = <(SignedWireUnit<C>, Panorama<C>)>::deserialize(deserializer)?;
+        if panorama.hash() != swunit.wire_unit().panorama_hash {
+            return Err(serde::de::Error::custom("wrong panorama hash"));
```
goral09, line 96: If this is a serialization error, will we be able to ban peers if they give us invalid data? This is domain knowledge and IMO it shouldn't live in (de)serialization logic.

node/src/components/consensus/highway_core/highway/vertex.rs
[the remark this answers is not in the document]
afck: `u` isn't necessarily my latest unit. It's not necessarily _my_ unit at all. Maybe I'm not even a validator!

I'm a peer who sent you (maybe even as a dependency of something else) `u` (which might be old, if you've just joined).
If `u` doesn't mark `v`'s creator as faulty, that's totally fine in general.
But I should be able to answer all requests for dependencies of `u` somehow.
If I give you the wrong unit with the right sequence number, that's also fine, because as you say, you'll request it by hash if it doesn't match in the end. It might just add more rounds of communication if I send you the wrong fork first and thereby force you to request the full hash-based panorama and the other fork by hash.

Anyway, maybe you're right; at least it would simplify the different kinds of dependencies. I can give it a try and make the change. Do you think it should be part of this PR?

node/src/components/consensus/highway_core/highway/vertex.rs
afck: I'm not sure: In general, what _do_ we do if we receive data that doesn't deserialize? I think we already report the peer as bad in that case?

afck: I definitely think that invariants that can be verified without using any data can be considered deserialization errors, though. This is just a hash that is expected to match the hash of some of the message's data.

goral09: That rule isn't intrinsic to the bytes/object itself though – the message with an invalid hash can still correctly deserialize to the expected structure. To me, that rule is a domain rule and (de)serialization logic shouldn't know anything about domain, that' all.

afck: OK, I'll change it then, and do the check in `pre_validate_vertex` instead.

node/src/components/consensus/highway_core/highway/vertex.rs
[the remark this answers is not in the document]
goral09: Thanks for the more detailed explanation, I missed some nuances originally.

> It might just add more rounds of communication if I send you the wrong fork first and thereby force you to request the full hash-based panorama and the other fork by hash.

The new approach, by design, adds more rounds of communication but that's only for the case where we have faulty/non-honest validators, and we're fine with that. This particular use-case we're discussing also involves equivocations.

I would prefer if we added it to this PR, so that it marks that piece of work as complete.

node/src/components/consensus/highway_core/highway/vertex.rs
goral09: That's only my opinion (only I feel pretty strong about it) and you don't have to do it if you strongly disagree.

node/src/components/consensus/highway_core/highway/vertex.rs
```diff
@@ -38,6 +43,15 @@ impl<C: Context> Dependency<C> {
     pub(crate) fn is_unit(&self) -> bool {
         matches!(self, Dependency::Unit(_))
     }
+    /// Returns `true` if both refer to the same vertex, even if they differ in whether they
```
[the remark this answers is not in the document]
afck: (Unresolving, so I don't forget. :sweat_smile: )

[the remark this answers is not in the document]
goral09: In the final version of the code it looks like the newline is there. I commented on a _commit_ originally and I don't understand why GH doesn't show _outdated_ tag here.

node/src/components/consensus/highway_core/highway/vertex.rs
afck: I only disagree weakly, so I'll do it. :smile: 

I'd _like_ to make it impossible to create an _instance_ of `Vertex` that violates the invariant, and that would require us to make it part of the deserialization logic.
However, we already instantiate by directly calling `UnitWithPanorama(unit, panorama)` in the code anyway — an enum can only enfoce its invariants if it is made private, and wrapped in some other type.

node/src/components/consensus/highway_core/highway/vertex.rs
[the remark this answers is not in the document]
afck: Ah, I see, thanks! Resolving again.
(It's not "outdated" because the comment is on the _next_ line, probably. Can't put a comment on a line that's not there.)

node/src/components/consensus/highway_core/highway/vertex.rs
goral09: If we had a separate representation of each struct for inter-node communication, then the mapping between the two could return errors. The constructor of `Vertex` is the place where _I_ would put these validation steps.

afck: > The constructor of `Vertex`

That's what I mean: An enum always has canonical constructors (its variant names), so we can't add validation there unless we wrap the whole enum in another type.

Also, if you do validation of a type in all constructors, you _should_ also do the same validation in the deserialization code, otherwise you create a way to instantiate surprising invalid objects.

goral09: If there's no way to deserialize directly into an enum, and there's a single constructor that includes the domain validity rules then there's no way to create an instance of an invalid object. In that setup you would need an intermediate DTO that has a some mapping to your domain types and that DTO could be invalid from the POV of the domain but it's OK – b/c DTO is not a domain _yet_.

afck: Sure, but as long as the enum isn't private (to its module) and hidden inside another data type, you can always create (outside its module) an invalid instance by directly creating a particular variant.

goral09: Sure, our job is to write the best code we can and even after making it defensive as much as possible, a malicious developer tries to work around the guards we placed and hurts himself – we can't do anything about it.

node/src/components/consensus/highway_core/state/tests.rs
```diff
@@ -188,30 +194,85 @@ fn add_unit() -> Result<(), AddUnitError<TestContext>> {
...
+    let maybe_err = state
+        .add_unit(unit.clone(), panorama!(N, b1, c0, N))
+        .err()
+        .map(unit_err);
+    assert_eq!(Some(UnitError::PanoramaLength(4)), maybe_err);
+    let maybe_err = state
+        .add_unit(unit.clone(), panorama!(a0, b1, c0))
+        .err()
+        .map(unit_err);
+    assert_eq!(Some(UnitError::PanoramaMismatch(ALICE)), maybe_err);
```
goral09, line 230: Looks like it's duplicated with lines 216-220.

node/src/components/consensus/highway_core/state/tests.rs
```diff
@@ -188,30 +194,85 @@ fn add_unit() -> Result<(), AddUnitError<TestContext>> {
...
+    let maybe_err = add_unit!(state, CAROL, 0, 4u8, None; N, N, N)
+        .err()
+        .map(unit_err);
+    assert_eq!(Some(UnitError::MissingBlock), maybe_err);
+
+    // The panorama length must be the number of validators.
+    let maybe_err = add_unit!(state, CAROL, 0, 4u8, 0xC; N, N, N, N)
+        .err()
+        .map(unit_err);
+    assert_eq!(Some(UnitError::PanoramaLength(4)), maybe_err);
```
goral09, line 268: Isn't that the same as in https://github.com/casper-network/casper-node/pull/1550/files#diff-5edd9079fa700dab6deb47fd36ebe675615f67a4d00a3112e0cbaeac2d1a63c4R221-R225 ?

node/src/components/consensus/highway_core/state/tests.rs
```diff
@@ -236,6 +297,84 @@ fn add_unit() -> Result<(), AddUnitError<TestContext>> {
...
+    // We need to request the full hash-based panoramas. They contain the right c0 by hash.
+    let res = new_state.compute_panorama(&b1_swunit);
+    assert_eq!(Err(Dependency::UnitWithPanorama(b1)), res);
+    let b1_panorama = state.unit(&b1).panorama.clone();
+    let c0_swunit = state.wire_unit(&c0, TEST_INSTANCE_ID).unwrap();
+    let c0_panorama = new_state.compute_panorama(&c0_swunit).unwrap();
+    new_state.add_unit(c0_swunit, c0_panorama)?;
+
+    // Finally we can add b1 and a2. Since a2 already knows that Carol is faulty, its panorama can
+    // be computed despite the multiple forks.
```
goral09, line 370: Can you add an assertion that verifies that `a2`'s panorama indeed sees Carol as faulty?

node/src/components/consensus/highway_core/state.rs
```diff
@@ -768,31 +862,76 @@ impl<C: Context> State<C> {
...
         }
+        for ((vidx, obs), seq_obs) in panorama.enumerate().zip(&wunit.seq_num_panorama) {
+            match (obs, seq_obs) {
+                (Observation::Faulty, ObservationSeqNum::Faulty)
+                | (Observation::None, ObservationSeqNum::None) => {}
+                (Observation::Correct(hash), ObservationSeqNum::Correct(seq_number))
+                    if self.unit(hash).seq_number == *seq_number => {}
+                (_, _) => return Err(UnitError::PanoramaMismatch(vidx)),
+            }
+        }
```
goral09, line 934: Can we extract this to a method on `Panorama` type? Also, it seems like you do that in `pre_validate_unit` as well.

node/src/components/consensus/highway_core/state.rs
```diff
@@ -768,31 +862,76 @@ impl<C: Context> State<C> {
...
+            (ObservationSeqNum::Correct(seq_num), Some(_)) => {
+                if seq_num.checked_add(1) != Some(wunit.seq_number) {
+                    return Err(UnitError::SequenceNumber);
+                }
+            }
+            (ObservationSeqNum::None, None) if wunit.seq_number != 0 => {
+                return Err(UnitError::SequenceNumber);
+            }
+            _ => {}
+        }
```
goral09, line 890: Can you add a comment that we validate unit's `previous` against the panorama for consistency?

node/src/components/consensus/highway_core/highway.rs
```diff
@@ -352,21 +419,36 @@ impl<C: Context> Highway<C> {
...
+            Dependency::UnitBySeqNum(seq_num, vidx, hash) => self
+                .validators
+                .id(*vidx)
+                .and_then(|_| self.state.maybe_unit(hash))
+                .and_then(|unit| unit.panorama[*vidx].correct())
+                .and_then(|hash| self.state.find_in_swimlane(hash, *seq_num))
+                .and_then(|hash| self.state.wire_unit(hash, self.instance_id))
+                .map_or(GetDepOutcome::None, |wunit| {
+                    GetDepOutcome::Vertex(Vertex::Unit(wunit))
+                }),
```
goral09, line 440: This should get simplified if we get rid of `hash` from `UnitBySeqNum`.

goral09 (approved):
Looks good, I don't think I had any blocking comments.

node/src/components/consensus/highway_core/active_validator.rs
```diff
@@ -197,16 +195,14 @@ impl<C: Context> ActiveValidator<C> {
                 return effects;
             } else if timestamp == r_id + self.witness_offset(r_len) {
                 let panorama = self.panorama_at(state, timestamp);
-                if let Some(witness_unit) =
-                    self.new_unit(panorama, timestamp, None, state, instance_id)
-                {
+                if let Some(vv) = self.new_unit(panorama, timestamp, None, state, instance_id) {
```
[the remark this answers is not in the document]
afck: Or `witness_vv` and `confirmation_vv`, since it's a `ValidVertex` now?

node/src/components/consensus/highway_core/state/panorama.rs
```diff
@@ -213,6 +213,11 @@ impl<C: Context> Panorama<C> {
         self.enumerate().filter_map(missing_dep).next()
     }
 
+    /// Returns the checksum of all hashes in this panorama.
+    pub(crate) fn hash(&self) -> C::Hash {
+        <C as Context>::hash(&bincode::serialize(self).expect("serialize Panorama"))
```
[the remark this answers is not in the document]
afck: Isn't that identical? The `ValidatorMap` is just a `Vec` wrapper.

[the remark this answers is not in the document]
goral09: Isn't the wrapper adding to the serialization format? If not, then it's all good already.

node/src/components/consensus/highway_core/state/tests.rs
afck: Good catch! :+1:

node/src/components/consensus/highway_core/state/tests.rs
afck: But that's just the `F` in line 319:
```rust
let a2 = add_unit!(state, ALICE, None; a1, b1, F)?;
```

node/src/components/consensus/highway_core/state/panorama.rs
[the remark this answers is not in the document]
afck: I'll double check, but I'm pretty sure it never does.

node/src/components/consensus/highway_core/active_validator.rs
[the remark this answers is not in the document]
goral09: Ok, sounds good.

node/src/components/consensus/highway_core/state.rs
afck: It's not exactly the same, though, because in `validate_unit` we compare sequence numbers, and in `pre_validate_unit` we can't (because we don't necessarily have the unit with the cited hash yet).

node/src/components/consensus/highway_core/highway.rs
```diff
@@ -280,7 +295,14 @@ impl<C: Context> Highway<C> {
     ) -> Result<ValidVertex<C>, (PreValidatedVertex<C>, VertexError)> {
         match self.do_validate_vertex(pvv.inner()) {
             Err(err) => Err((pvv, err)),
-            Ok(()) => Ok(ValidVertex(pvv.0)),
+            Ok(Some(panorama)) => match pvv.0 {
+                Vertex::Unit(swunit) => Ok(ValidVertex(Vertex::UnitWithPanorama(swunit, panorama))),
+                vertex => {
+                    error!(?vertex, "unexpected computed panorama for vertex");
```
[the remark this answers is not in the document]
afck: You're right; we actually don't need the separate `do_validate_vertex` at all anymore, then this problem goes away, too.

node/src/components/consensus/highway_core/highway.rs
```diff
@@ -373,23 +373,41 @@ impl<C: Context> Highway<C> {
     }
 
     /// Returns whether we have a vertex that satisfies the dependency.
-    pub(crate) fn has_dependency(&self, dependency: &Dependency<C>) -> bool {
+    ///
+    /// If it's ambiguous, i.e. the dependency is specified by sequence number and we don't have
+    /// the citing unit, `None` is returned.
```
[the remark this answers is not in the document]
afck: It actually should always return `None` or `Some(false)` for `UnitBySeqNum` now, because we can _never_ know whether that's the right unit.

afck:
Thanks, @goral09! I addressed or replied to all your comments. Please take another look.

goral09 (approved):
Thanks for the updates @afck. Looks like this https://github.com/casper-network/casper-node/pull/1550#discussion_r649991739 is the only one outstanding but that's not a blocker. Still approve :heavy_check_mark:

xcthulhu (changes requested):
Since these changes are substantial, it would be best to extend the DES tests in `highway_core` to cover this code and various corner cases to avoid a regression.

A less attractive alternative is to do gray-box testing with the `test_chain` rig.

node/src/components/consensus/highway_core/highway/vertex.rs
afck: Done in https://github.com/casper-network/casper-node/pull/1550/commits/bda70086de3433cf17a4a8341138fce929fafd12.

goral09:
> Since these changes are substantial, it would be best to extend the DES tests in `highway_core` to cover this code and various corner cases to avoid a regression.
> 
> A less attractive alternative is to do gray-box testing with the `test_chain` rig.

This PR is being merged to a feature branch, not `dev`. DES tests in `highway_core` can't be extended to test the synchronizer b/c it [synchronizer] is not part of `highway_core` in a sense that can be tested there.

PR #1555, opened 2021-06-09
Allow longer strings to be displayed in jsons. Was 100, now 150.

zie1ony:
It's needed to display full signature instead of `130 chars`.


EdHastingsCasperLabs (approved):

EdHastingsCasperLabs:
Fix the tests to allow new value; otherwise LGTM

Issue #1549, opened 2021-06-07
Use Legacy Sync's Rule For When To Stop Downloading Blocks In Fast-Sync


xcthulhu:

xcthulhu:
Closed with #1654

PR #1562, opened 2021-06-10
Sync blocks

xcthulhu:


node/src/components/linear_chain_sync/operations.rs
```diff
@@ -279,6 +307,116 @@ where
...
+                        "Error validating finality signatures from peer.",
+                    );
+                    // TODO: ban peer
+                    continue;
+                }
+
+                // Store the block
+                // effect_builder
+                //     .put_block_to_storage(Box::new(block.clone()))
+                //     .await;
```
afck, line 391: Why not?

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -421,5 +561,29 @@ where
         outstanding_trie_keys.extend(missing_descendant_trie_keys);
     }
 
-    Ok(())
+    while let Some(block_with_metadata) = fetch_and_store_block_by_height(
```
afck, line 564: Should we add another condition to exit the loop? Whenever we have a reasonably recent block, e.g. one that is in the current era?

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -421,5 +561,29 @@ where
...
+        effect_builder,
+        most_recent_block_header.height(),
```
afck, line 566: Shouldn't this be `+ 1`?

node/src/components/linear_chain_sync/operations.rs
xcthulhu: This is issue #1549, but @fizyk20 and I just discussed this and I could tackle this here...

node/src/components/linear_chain_sync/operations.rs
xcthulhu: I am calling legacy sync after this loop.  Sadly, legacy sync makes some assumptions if it finds blocks in the database.  So this will have to wait until fast sync can fly on its own.

node/src/components/storage.rs
```diff
@@ -879,6 +879,7 @@ impl Storage {
         tx: &mut Tx,
         era_id: EraId,
     ) -> Result<Option<BlockHeader>, LmdbExtError> {
+        info!(switch_block_era_id_index = ?self.switch_block_era_id_index);
```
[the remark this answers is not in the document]
xcthulhu: You are right.  Fixed in 0a104771f8c46bec4b5ca71b45036d2fa41dfe42

node/src/components/linear_chain_sync/operations.rs
afck: Ah, right, I didn't know there's an issue for it. :+1:

afck (approved):

PR #1570, opened 2021-06-11
Fix intermittent failure in SSE server test

### LINKED ISSUE #1560: Intermittent failure in SSE test
https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/515/3/3

Fraser999:
This makes the lagging client SSE server test more robust by changing it in a couple of ways:
* the test server now emits up to 100,000,000 events rather than the 100 previously
* the lagging clients now run slower by having a small sleep between reading each chunk of received data from the TCP stream
* the clients also pause for 5 seconds after connecting to the server but before starting to read data, to help ensure the server's TCP send buffer and client's TCP receive buffer are filled

Closes #1560.

### DIFF
--- a/node/src/components/event_stream_server/tests.rs
+++ b/node/src/components/event_stream_server/tests.rs
@@ -16,7 +16,7 @@ use schemars::schema_for;
 use tempfile::TempDir;
 use tokio::{
     sync::{Barrier, Notify},
-    task::JoinHandle,
+    task::{self, JoinHandle},
     time,
 };
 use tracing::debug;
@@ -30,6 +30,9 @@ use sse_server::{
 /// The total number of random events each `EventStreamServer` will emit by default, excluding the
 /// initial `ApiVersion` event.
 const EVENT_COUNT: u32 = 100;
+/// The maximum number of random events each `EventStreamServer` will emit, excluding the initial
+/// `ApiVersion` event.
+const MAX_EVENT_COUNT: u32 = 100_000_000;
 /// The event stream buffer length, set in the server's config.  Set to half of the total event
 /// count to allow for the buffer purging events in the test.
 const BUFFER_LENGTH: u32 = EVENT_COUNT / 2;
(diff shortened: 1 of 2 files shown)


node/src/components/event_stream_server/tests.rs
```diff
@@ -111,6 +126,8 @@ impl ServerSyncBehavior {
     async fn sleep_if_required(&self) {
         if self.has_delay_between_events {
             time::sleep(DELAY_BETWEEN_EVENTS).await;
+        } else {
+            task::yield_now().await;
```
marc-casperlabs, line 130: I'm not sure this is necessary, since we're calling an `async` function, thus giving the runtime a chance to intercept anyway? :)

marc-casperlabs (approved):
Looks good to me, although I can't judge right now if this fixes the behavior deterministically or we just ended up tweaking the numbers. The two barriers make me think the former though :)

darthsiroftardis (approved):
LGTM

node/src/components/event_stream_server/tests.rs
Fraser999: Yes, it shouldn't be strictly necessary.  However, the local tests seemed to benefit from this in the case of the lagging node.  My thinking is that this helps interleave the time slices more evenly between the client task and the server task.  We need the server task to actually progress (without a sleep) in order for the server to fill the buffers and kick the client off.  

However, it the server gets too big a time slice, it gets a chance to enqueue a _lot_ more events to the stream.  The only thing stopping the server doing that in its time slice is for the clients to detect the test has ended, allowing their futures to join and the call to stop the running server can happen.

So, it benefits us to give the clients control as frequently as possible.

Fraser999:
> I can't judge right now if this fixes the behavior deterministically or we just ended up tweaking the numbers.

It's probably somewhere in between.  The clients are explicitly forced to run more slowly than the server now (which wasn't the case previously) which is more than just tweaking the numbers.  Given enough time, that should _always_ cause the clients to get kicked.

I didn't want to leave open the possibility of the test running for an unacceptably long time however, hence the additional (failure) stopping condition of allowing a maximum of 100,000,000 events in the test.  Local testing showed that in success cases, the event count didn't exceed 700,000.

Ultimately the test might still fail due to some unforeseen condition I haven't considered, but I think this change makes it much less likely.

Fraser999:
bors merge

Fraser999:
bors merge

PR #1582, opened 2021-06-11
NDRS-1569: Fix peers metric

darthsiroftardis:
Cause for the bug:

- Recent changes to the `small_network` component removed the `peers` member from the component which was used to update the corresponding and the function `update_peers_metric` which did the update

Fix:

- The peers metrics now is set based on the current amount of symmetric connections


nilz3ro (approved):
Looks good to me 💯

marc-casperlabs (approved):
Nice work, @darthsiroftardis !

Issue #1558, opened 2021-06-09
Issue when rejoining an inactive validator


TomVasile:
### Overview:

TL;DR: If we restart a node after it is marked inactive but prior to ejection, then rebid it after ejection, it fails to rejoin.

Fails consistently in case 1 ( labeled in my branch ) and passes consistently in case 2 ( labeled in my branch ).
- Case 1 implements what was proposed in the issue. ie. restarting the node prior to verifying ejection
- Case 2 is basically the same as the current test, restarts the node after verified ejection

### Steps to repro:
```
cd "$NCTL/sh/scenarios"
nctl-assets-teardown
nctl-assets-setup chainspec_path="$NCTL/sh/scenarios/chainspecs/itst13.chainspec.toml.in"
nctl-start
sleep 90
bash -i itst13.sh
```

### Links:
Test description: https://app.zenhub.com/workspaces/engineering-60953fafb1945f0011a3592d/issues/casper-network/governance/14

My branch: https://github.com/TomVasile/casper-node/tree/governance-issue-14

Test in question: [itst13.sh](https://github.com/TomVasile/casper-node/blob/governance-issue-14/utils/nctl/sh/scenarios/itst13.sh)

afck:
If I set

```toml
standstill_timeout = '30s'
```

in the `config.toml` it passes for me. In general it doesn't make sense to have a timeout that's longer than an era (41 seconds in this test). Otherwise the joining node could start era 4 just when the others move on to 5 and make no progress for a long time.

afck:
@fizyk20 also saw the test fail with an unexpected proposed block:
I realized I described the feature the wrong way. You're really only excluded from the leader sequence if you were inactive in the _previous_ era. Not in general if you're about to be evicted. So: If the node was inactive in era 4 then it doesn't get to propose blocks in era 5. However, if it sends any message in era 5, it _does_ get to propose blocks in era 6, even if it'll be evicted in era 7.

So I think the test should be modified: The era length should be increased, the standstill timeout decreased, and `auction_delay` set to 1 (like in mainnet).

afck:
@fizyk20 and I probably finally found the root cause. This branch should work: https://github.com/afck/casper-node/tree/governance-issue-14 (Testing right now…)

[...]

PR #1583, opened 2021-06-11
Fix new_era() arguments mix-up.

afck:
This fixes the root cause of https://github.com/casper-network/casper-node/issues/1558.

To make the tests work reliably we should also set the standstill alert to a value lower than the minimum era duration and the `auction_delay` to 1, though. (See issue comment.)

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -886,8 +886,8 @@ where
             Timestamp::now(), // TODO: This should be passed in.
             next_era_validators_weights.clone(),
             newly_slashed,
-            era_end.inactive_validators.iter().cloned().collect(),
             slashed,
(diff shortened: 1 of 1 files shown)


xcthulhu (approved):
So... I'll approve this but I worry this function has too many arguments.

Maybe it could take a `switch_block_header` as an argument?  Then you could derive:
     - `era_id`
     - `start_time`
     -  `start_height`

Switch block headers also have an `era_end`, which can be used to derive inactive validators and newly slashed validators.  It also has the next era validator weights.  You might want a constructor that relies on those to throw an error, however, since those fields are optional.

goral09 (approved):

fizyk20 (approved):

afck:
> Maybe it could take a `switch_block_header` as an argument?

Yes, I'm planning to do something like that next week.

afck:
bors merge

PR #1592, opened 2021-06-14
Add holders of transitive dependencies in synchronizer.

### LINKED ISSUE #1584: Highway synchronizer: transitive dependencies should take note of holder
If there are units `c < b < a` and we receive `b` from Bob, the synchronizer makes a note in `vertices_awaiting_deps` that Bob has `b`, and requests `c` from Bob. Now assume Bob goes offline and Alice sends us `a`. The synchronizer will identify `c` as a transitive dependency of `a`, and request `c` from Alice. It makes a note in `vertices_awaiting_deps` that Alice has `a`, but not that Alice has `b`! Now if Alice sends us `c` and `b` has another dependency `d`, we will request `d` only from Bob, not from Alice, and be stuck!

In `find_transitive_dependency`, we need to mark the sender as a holder of all intermediate vertices.

afck:
If `c < b < a` and we already have `b` in the synchronizer queue when Alice sends us `a`, then we also know that Alice must be a holder of `c`, and request it from her.

With this PR, we will also remember later that Alice is a holder of `b`, and if `b` has further dependencies, we can request them from Alice.

Closes #1584.

### DIFF
--- a/node/src/components/consensus/highway_core/synchronizer.rs
+++ b/node/src/components/consensus/highway_core/synchronizer.rs
@@ -57,6 +57,13 @@ impl<I: NodeIdT, C: Context> PendingVertices<I, C> {
             .or_insert(time_received);
     }
 
+    /// Adds a holder to the vertex that satisfies `dep`.
+    fn add_holder(&mut self, dep: &Dependency<C>, sender: I, time_received: Timestamp) {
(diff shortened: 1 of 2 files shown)


fizyk20 (approved):

afck:
The test failed because the order of the messages is not deterministic (`HashMap`). Fixing it now…

goral09 (approved):

afck:
bors merge

afck:
bors merge

PR #1600, opened 2021-06-14
Make SSE tests not depend upon server emitting keepalives to terminate

### LINKED ISSUE #1591: Fix SSE server tests
The SSE server tests run clients which subscribe to an SSE server.  The clients finish running when they receive a keepalive from the server, indicating a period of inactivity.  The server is set up so that such a period of inactivity should only happen once all test events have been emitted.

To stop the tests running for too long, the keepalive frequency is reduced from 15 seconds to 100 milliseconds.  However, it seems that this value is too low, as indicated in a sporadic failure of `should_handle_wrapping_past_max_event_id_for_signatures` where the first client must have received a keepalive before the test had actually completed.

We should change how the clients are stopped.  Rather than relying on the keepalive, we should pass in the actual expected final event ID, and use the keepalive as a fallback.  Since the keepalive will then not normally be expected to be received, we can revert to 15 second intervals for keepalives in tests.

Fraser999:
This PR changes how the SSE server tests terminate to avoid spurious failures.

Previously they used the arrival of the first keepalive as the signal that the server had finished sending all events in the tests.  However, even with a relatively large margin of error, this was insufficient to guarantee that a server in a running test wouldn't be so slow that it would never emit a keepalive before all events had been emitted.

The stopping criterion for a test client has been changed to a specific event ID.  The client consumes events until the expected ID is received, or panics if a keepalive is received.  As the keepalive interval has been reverted from 100 milliseconds for tests to the default 15 seconds, there is no reasonable way in which a client should receive a keepalive mid-test now.

Closes #1591.

### DIFF
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -1,6 +1,6 @@
 //! Types and functions used by the http server to manage the event-stream.
 
-use std::{collections::HashMap, time::Duration};
+use std::collections::HashMap;
 
 use datasize::DataSize;
 use futures::{future, Stream, StreamExt};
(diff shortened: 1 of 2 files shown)


darthsiroftardis (approved):
LGTM!

nilz3ro (approved):
# 💯

Fraser999:
bors merge

Issue #1602, opened 2021-06-15
dependabot config


mpapierski:
We're using release branches as defaults from now on, and dependabot opens its PRs against it. We need to migrate into GH dependabot config, and have an explicit `target-branch` option.

mpapierski:
We're not using dependabot anymore

PR #1603, opened 2021-06-15
GH-1602: dependabot config

mpapierski:
This PR adds explicit config for GitHub's dependabot with config for both npm and cargo with daily intervals.


TomVasile (approved):
Looks good 👍

PR #1607, opened 2021-06-17
Mark known faulty validators when initializing eras.

### LINKED ISSUE #1516: Mark known faulty validators when initializing eras.
We have two cross-era equivocation handling mechanisms:
* If a validator was known to be faulty by the previous switch block, we ban them in the next era and don't accept any of their messages. ([Currently deactivated](https://github.com/casper-network/casper-node/blob/a2d6511d17aa7df75a633542d207c280a775248e/node/src/components/consensus/era_supervisor.rs#L1056).)
* If a validator equivocates in a recent era, we start citing them as `Faulty` in all future eras as well, as if they had equivocated there, too.

Even after re-enabling banning, this still leaves a gap between the creation of the switch block and the initialization of the next era. So:

When initializing a new era, we should immediately mark everyone as `Faulty` who equivocated in the past `bonded_eras()` eras. The implementation is probably as easy as adding this to `EraSupervisor::new_era`:

```diff
diff --git a/node/src/components/consensus/era_supervisor.rs b/node/src/components/consensus/era_supervisor.rs
index d11573894..9d5cf7540 100644
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -348,6 +348,15 @@ where
             now,
         );
 
+        for past_era_id in self.iter_past_other(era_id, self.bonded_eras()) {
+            for pub_key in self.active_eras[&past_era_id]
+                .consensus
+                .validators_with_evidence()
+            {
+                consensus.mark_faulty(pub_key);
+            }
+        }
+
         if should_activate {
             let secret = Keypair::new(self.secret_signing_key.clone(), our_id.clone());
             outcomes.extend(consensus.activate_validator(
```

The hard part is writing tests for it.

Also related is #1538: If that is done before this issue, it shouldn't be `bonded_eras()`, of course, and just the auction delay instead.

afck:
If someone equivocates in an old era N, we also mark them as faulty in era N + 1, N + 2, etc. (up to `unbonding_delay - auction_delay`). However, if the equivocation was detected _before_ era N + 1 was initialized, we failed to mark them as faulty in era N + 1 on initialization. This PR fixes that:

Now, when initializing a new era, we immediately mark everyone as faulty against whom we have evidence in a recent era. Since the synchronizer ignores faulty validators' units, unless cited by a non-faulty one, this means you can't effectively equivocate two eras in a row.

I also renamed all "slashing" references in consensus, since there are no plans to enable slashing, and that isn't strictly a consensus question anyway.

Closes #1516.

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -371,7 +371,7 @@ where
         for e_id in self.iter_past_other(era_id, self.bonded_eras()) {
             if let Some(old_era) = self.active_eras.get_mut(&e_id) {
                 for pub_key in old_era.consensus.validators_with_evidence() {
-                    let proposed_blocks = era.resolve_evidence(&pub_key);
+                    let proposed_blocks = era.resolve_evidence_and_mark_faulty(&pub_key);
                     if !proposed_blocks.is_empty() {
                         error!(
                             ?proposed_blocks,
@@ -1190,7 +1190,7 @@ where
                 {
                     let proposed_blocks =
                         if let Some(era) = self.era_supervisor.active_eras.get_mut(&e_id) {
-                            era.resolve_evidence(&pub_key)
+                            era.resolve_evidence_and_mark_faulty(&pub_key)
                         } else {
                             continue;
                         };
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -425,12 +425,6 @@ where
         self.active_eras.get(&era_id).map_or(false, has_validator)
     }
 
-    /// Returns the most recent active era.
-    #[cfg(test)]
-    pub(crate) fn current_era(&self) -> EraId {
-        self.current_era
-    }
-
     pub(crate) fn stop_for_upgrade(&self) -> bool {
         self.stop_for_upgrade
     }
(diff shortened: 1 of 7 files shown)


node/src/components/consensus/era_supervisor.rs
```diff
@@ -358,19 +358,36 @@ where
...
+            new_faulty,
+            faulty,
             validators,
         );
+
+        // Mark validators as faulty for which we have evidence in a recent era.
+        for e_id in self.iter_past_other(era_id, self.bonded_eras()) {
+            if let Some(old_era) = self.active_eras.get_mut(&e_id) {
+                for pub_key in old_era.consensus.validators_with_evidence() {
+                    let proposed_blocks = era.resolve_evidence(&pub_key);
```
xcthulhu, line 374: Resolve evidence is side-effecting, so we might want to change its name to `resolve_evidence_and_mark_faulty`

afck: Done in https://github.com/casper-network/casper-node/pull/1607/commits/2d39fd4d382e3790f88fdde206a359ef9f94a204.

xcthulhu (approved):
LGTM

goral09 (approved):

afck:
bors merge

PR #1619, opened 2021-06-18
Add regression tests for cryptographic signatures

### LINKED ISSUE #1581: Regression tests for signing function
We need regression tests for [`asymmetric_key::sign()`](https://github.com/casper-network/casper-node/blob/release-1.2/node/src/crypto/asymmetric_key.rs#L27-L51) which use hard-coded values in a test per key variant.

Fraser999:
This PR adds regression tests for signing via the Ed25519 and secp256k1 keys.  The intent is that should either of these change their implementations in a non-backwards-compatible manner, these tests will alert us.

Closes #1581.

### DIFF
--- a/node/src/crypto/asymmetric_key.rs
+++ b/node/src/crypto/asymmetric_key.rs
@@ -473,6 +473,43 @@ MCowBQYDK2VwAyEAGb9ECWmEzf6FQbrBZ9w7lshQhqowtrbLDFw4rXAxZuE=
             let signature = sign(data, &ed25519_secret_key, &public_key);
             bytesrepr::test_serialization_roundtrip(&signature);
         }
+
+        #[test]
+        fn validate_known_signature() {
+            // In the event that this test fails, we need to consider pinning the version of the
+            // `ed25519-dalek` crate to maintain backwards compatibility with existing data on the
+            // Casper network.
+
+            // Values taken from:
+            // https://github.com/dalek-cryptography/ed25519-dalek/blob/925eb9ea56192053c9eb93b9d30d1b9419eee128/TESTVECTORS#L62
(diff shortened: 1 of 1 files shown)


xcthulhu (commented):
LGTM, but we might consider some sad-path tests as well.

nilz3ro (approved):
lgtm

Fraser999:
@xcthulhu 
> LGTM, but we might consider some sad-path tests as well.

We do already have some sad path tests for signatures.  I don't think there'd be much point in hard-coded sad path tests too, unless I'm missing something?

xcthulhu:
> @xcthulhu
> 
> > LGTM, but we might consider some sad-path tests as well.
> 
> We do already have some sad path tests for signatures. I don't think there'd be much point in hard-coded sad path tests too, unless I'm missing something?

It'd possibly be redundant but until we have better instrumentation of code coverage I'd err on the side of caution myself.

Fraser999:
>>> LGTM, but we might consider some sad-path tests as well.
>>
>> We do already have some sad path tests for signatures. I don't think there'd be much point in hard-coded sad path tests too, unless I'm missing something?
>
> It'd possibly be redundant but until we have better instrumentation of code coverage I'd err on the side of caution myself.

After discussion, I'm still unconvinced.  The tests would be identical to [those already in the suite](https://github.com/casper-network/casper-node/blob/09c9d02b315e171d32b834600646e4d2e3dd80b8/node/src/crypto/asymmetric_key.rs#L754-L780), so I can't bring myself to just copy paste that exact test for this PR.

Having said that, I won't block a PR if you do choose to duplicate these tests yourself :)

darthsiroftardis (approved):
LGTM

Fraser999:
bors merge

PR #1621, opened 2021-06-18
NDRS-1617: Resolve doc warnings

darthsiroftardis:
CHANGELOG:

- Added `RUSTDOCFLAGS="-D warnings"` to the `Makefile` causing doc warnings to escalated to errors instead
- Resolve resulting errors in both `node` and `types`

### DIFF
--- a/smart_contracts/contract/src/ext_ffi.rs
+++ b/smart_contracts/contract/src/ext_ffi.rs
@@ -717,7 +717,7 @@ extern "C" {
     /// The bytes in wasm memory from offset `key_ptr` to `key_ptr + key_size`
     /// will be used together with the current context’s seed to form a dictionary.
     /// The value at that dictionary is read from the global state, serialized and
-    /// buffered in the runtime. This result can be obtained via the [`read_host_buffer`]
+    /// buffered in the runtime. This result can be obtained via the \[`read_host_buffer`]\
     /// function.
     ///
     /// # Arguments
--- a/smart_contracts/contract/src/ext_ffi.rs
+++ b/smart_contracts/contract/src/ext_ffi.rs
@@ -717,7 +717,7 @@ extern "C" {
     /// The bytes in wasm memory from offset `key_ptr` to `key_ptr + key_size`
     /// will be used together with the current context’s seed to form a dictionary.
     /// The value at that dictionary is read from the global state, serialized and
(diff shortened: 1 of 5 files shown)


smart_contracts/contract/src/ext_ffi.rs
```diff
@@ -717,7 +717,7 @@ extern "C" {
     /// The bytes in wasm memory from offset `key_ptr` to `key_ptr + key_size`
     /// will be used together with the current context’s seed to form a dictionary.
     /// The value at that dictionary is read from the global state, serialized and
-    /// buffered in the runtime. This result can be obtained via the [`read_host_buffer`]
+    /// buffered in the runtime. This result can be obtained via the \[`read_host_buffer`]\
```
Fraser999, line 720: This should be renamed from `read_host_buffer` to `casper_read_host_buffer`.

types/src/key.rs
```diff
@@ -345,7 +345,7 @@ impl Key {
         }
     }
 
-    /// Returns a reference to the inner [`DictionaryAddr`] if `self` is of type
+    /// Returns a reference to the inner `DictionaryAddr` if `self` is of type
```
Fraser999, line 348: This has actually made me look at why the doc warning was saying the type isn't public.

Here, I think we should leave the doc comment as it was, but instead in lib.rs we should be publicly exporting `KEY_DICTIONARY_LENGTH` and `DictionaryAddr`.

Also, the doc comment for `DictionaryAddr` should be changed from ``/// A newtype for [`Key`]s dictionary variant.`` to ``/// An alias for [`Key`]s dictionary variant.``

types/src/system/mint/error.rs
```diff
@@ -25,10 +25,10 @@ pub enum Error {
     /// Destination purse not found.
     #[cfg_attr(feature = "std", error("Destination not found"))]
     DestNotFound = 2,
-    /// See [`PurseError::InvalidURef`].
+    /// See `PurseError::InvalidURef`.
```
Fraser999, line 28: These are unhelpful as `PurseError` is private.  Let's move the comments currently on the `PurseError` variants to here, and have the `PurseError` doc comments refer the reader to here to avoid duplicating them.

Fraser999 (changes requested):

Fraser999 (approved):

PR #1624, opened 2021-06-21
Deploy deduplication

marc-casperlabs:
Adds in-memory deduplication for deploys by adding a cache of `Arc` handles to the storage component. For historic reasons this is still driven by code that lives in the participant routing logic, until that can be refactored. The `SharedObject` enum is a wrapper type distinguishing between shared and owned types, it is mainly used to allow for toggling the behavior at runtime.

Some tests are included to ensure backwards compatible serialization/deserialization of the now `SharedObejct` wrapped types.

Currently we only deduplicate deploys, as these can form by far the largest objects in memory due to their WASM payload.

Reviewer hint: `BlobCache` and `SharedObject` are self-contained and can be reviewed on their own.

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -36,7 +36,9 @@
 //! The storage component itself is panic free and in general reports three classes of errors:
 //! Corruption, temporary resource exhaustion and potential bugs.
 
+pub(crate) mod blob_cache;
 mod lmdb_ext;
+
 #[cfg(test)]
 mod tests;
 
@@ -64,6 +66,8 @@ use tracing::{debug, error, info};
 use casper_execution_engine::shared::newtypes::Blake2bHash;
 use casper_types::{EraId, ExecutionResult, ProtocolVersion, Transfer, Transform};
 
+use self::blob_cache::BlobCache;
+
 use super::Component;
 #[cfg(test)]
 use crate::crypto::hash::Digest;
@@ -78,7 +82,7 @@ use crate::{
         Block, BlockBody, BlockHash, BlockHeader, BlockHeaderWithMetadata, BlockSignatures, Deploy,
         DeployHash, DeployHeader, DeployMetadata, Item, TimeDiff,
     },
-    utils::{blob_cache::BlobCache, WithDir},
+    utils::WithDir,
     NodeRng,
 };
 use lmdb_ext::{LmdbExtError, TransactionExt, WriteTransactionExt};
@@ -212,7 +216,7 @@ pub struct Storage {
     /// Whether or not memory deduplication is enabled.
     enable_mem_deduplication: bool,
     /// Pool of loaded items.
-    pub deploy_cache: BlobCache<<Deploy as Item>::Id>,
+    pub(crate) deploy_cache: BlobCache<<Deploy as Item>::Id>,
 }
 
 impl<REv> Component<REv> for Storage
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -209,6 +209,10 @@ pub struct Storage {
     switch_block_era_id_index: BTreeMap<EraId, BlockHash>,
     /// A map of deploy hashes to hashes of blocks containing them.
     deploy_hash_index: BTreeMap<DeployHash, BlockHash>,
+    /// Whether or not memory deduplication is enabled.
+    enable_mem_deduplication: bool,
+    /// How many loads before memory duplication checks for dead references.
+    mem_pool_gc_interval: u16,
 }
 
 impl<REv> Component<REv> for Storage
@@ -376,6 +380,8 @@ impl Storage {
             block_height_index,
             switch_block_era_id_index,
             deploy_hash_index,
+            enable_mem_deduplication: config.enable_mem_deduplication,
+            mem_pool_gc_interval: config.mem_pool_gc_interval,
         })
     }
 
@@ -1177,6 +1183,10 @@ pub struct Config {
     ///
     /// The size should be a multiple of the OS page size.
     max_state_store_size: usize,
+    /// Whether or not memory deduplication is enabled.
+    enable_mem_deduplication: bool,
+    /// How many loads before memory duplication checks for dead references.
+    mem_pool_gc_interval: u16,
 }
 
 impl Default for Config {
@@ -1188,6 +1198,8 @@ impl Default for Config {
             max_deploy_store_size: DEFAULT_MAX_DEPLOY_STORE_SIZE,
             max_deploy_metadata_store_size: DEFAULT_MAX_DEPLOY_METADATA_STORE_SIZE,
             max_state_store_size: DEFAULT_MAX_STATE_STORE_SIZE,
+            enable_mem_deduplication: false,
+            mem_pool_gc_interval: 1024,
         }
     }
 }
@@ -1225,7 +1237,7 @@ impl Display for Event {
 // ON OR BUILD UPON THIS CODE.
 
 impl Storage {
-    // Retrieves a deploy from the deploy store to handle a legacy network request.
+    /// Retrieves a deploy from the deploy store to handle a legacy network request.
     pub fn handle_legacy_direct_deploy_request(&self, deploy_hash: DeployHash) -> Option<Deploy> {
         // NOTE: This function was formerly called `get_deploy_for_peer` and used to create an event
         // directly. This caused a dependency of the storage component on networking functionality,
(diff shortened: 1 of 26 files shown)


node/src/utils/blob_cache.rs
```diff
@@ -0,0 +1,167 @@
...
+    pub fn put(&mut self, id: I, item: Weak<Vec<u8>>) {
+        self.items.insert(id, item);
+
+        if self.put_count >= self.garbage_collect_interval {
+            self.items.retain(|_, item| item.strong_count() > 0);
+
+            self.put_count = 0;
+        }
+
+        self.put_count += 1;
```
Fraser999, line 54: Might be worth making an alias for `u16` and using that as the type for `garbage_collect_interval` and `put_count` to help avoid future refactoring allowing this to overflow?

node/src/reactor/participating.rs
```diff
@@ -682,28 +682,80 @@ impl reactor::Reactor for Reactor {
...
-                                        Err(error) => {
-                                            error!("failed to create get-response: {}", error);
-                                            return Effects::new();
-                                        }
-                                    };
+                            // Depending on the runtime configuration, fetch as a shared or owned
+                            // object.
+                            if self.storage.mem_duplication_enabled() {
+                                let outcome =
+                                    self.storage.deploy_cache.get(&deploy_hash).map(Ok).or_else(
```
Fraser999, line 689: Couldn't the caching be handled entirely inside `Storage` rather than requiring explicit cache read/writes here?

node/src/reactor/participating.rs
```diff
@@ -682,28 +682,80 @@ impl reactor::Reactor for Reactor {
...
+                                    .handle_legacy_direct_deploy_request(deploy_hash)
+                                {
+                                    // This functionality was moved out of the storage component and
+                                    // should be refactored ASAP.
+                                    Some(deploy) => {
+                                        match Message::new_get_response(&deploy) {
+                                            Ok(message) => {
+                                                if self.storage.mem_duplication_enabled() {
+                                                    todo!("update cache if enabled")
+                                                }
```
Fraser999, line 743: Was this left in by mistake?  We're in an `else` branch here where `self.storage.mem_duplication_enabled()` is `false`.

node/src/components/storage.rs
```diff
@@ -209,6 +209,10 @@ pub struct Storage {
     switch_block_era_id_index: BTreeMap<EraId, BlockHash>,
     /// A map of deploy hashes to hashes of blocks containing them.
     deploy_hash_index: BTreeMap<DeployHash, BlockHash>,
+    /// Whether or not memory deduplication is enabled.
+    enable_mem_deduplication: bool,
+    /// Pool of loaded items.
+    pub deploy_cache: BlobCache<<Deploy as Item>::Id>,
```
Fraser999, line 215: I'd prefer if we kept the cache private and added whatever `pub(crate)` helpers are required.

Fraser999 (commented):
Some questions/nitpicks, but generally looking good.

xcthulhu:
So, I am probably out of the loop but I have to ask: did we identify a memory issue here using a profiler? It'd be nice if we could get away without doing this since it adds a lot of complexity to the storage component.

goral09:
> So, I am probably out of the loop but I have to ask: did we identify a memory issue here using a profiler? It'd be nice if we could get away without doing this since it adds a lot of complexity to the storage component.

Which complexity do you mean exactly? From what I see, there are only 24 lines added to the `components/storage.rs` and none are modified.

node/src/types.rs
```diff
@@ -48,3 +52,151 @@ pub type NodeRng = ChaCha20Rng;
 /// The RNG used throughout the node for testing.
 #[cfg(test)]
 pub type NodeRng = crate::testing::TestRng;
+
+/// An in-memory object that can possibly be shared with user parts of the system.
```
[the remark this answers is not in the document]
marc-casperlabs: "other" :(

node/src/types.rs
```diff
@@ -48,3 +52,151 @@ pub type NodeRng = ChaCha20Rng;
...
+///
+/// In general, this should only be used for immutable, content-addressed objects.
+///
+/// This type exists solely to switch between `Box` and `Arc` based behavior, future updates should
+/// deprecate this in favor of using `Arc`s directly or turning `LoadedItem` into a newtype.
+#[derive(Clone, DataSize, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub enum SharedObject<T> {
```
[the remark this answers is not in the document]
marc-casperlabs: Fair enough.

node/src/types.rs
```diff
@@ -48,3 +52,151 @@ pub type NodeRng = ChaCha20Rng;
...
+impl<'de, T> Deserialize<'de> for SharedObject<T>
+where
+    T: Deserialize<'de>,
+{
+    #[inline]
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        T::deserialize(deserializer).map(SharedObject::owned)
```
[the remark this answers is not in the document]
marc-casperlabs: There's no way of storing it in the cache at this point - we do not have access to the cache here. So any sharing would just be an `Arc` with a single reference.

There is merit to using `shared` regardless, as we would be able to avoid a copy if we update the code to reuse the `Arc`. But currently this has no effect, since the contained data is straight up deserialized and thrown away.

So the gain is not as big as it could be right now and I err'd on the side of keeping the old behavior for the sake of not making any more behavioral changes than necessary.

node/src/utils.rs
```diff
@@ -1,6 +1,7 @@
 //! Various functions that are not limited to a particular module, but are too small to warrant
 //! being factored out into standalone crates.
 
+pub mod blob_cache;
```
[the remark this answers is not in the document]
marc-casperlabs: We seem to have a fundamental difference on where to stash things, I hope we never have to share a silverware drawer!

Other than that, I'm with moving it :)

node/src/utils/blob_cache.rs
marc-casperlabs: I do like that `u16` puts a reasonable limit on the garbage collection interval, though it might be too small.

That aside, anyone changing the type will get a type error due to line 48?

```
        if self.put_count >= self.garbage_collect_interval {
```

marc-casperlabs: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=912b43e614b9bdb9d16c05203ad65107

node/src/reactor/participating.rs
marc-casperlabs: :( You're right.

node/src/reactor/participating.rs
marc-casperlabs: That's a damn good catch. Moved everything inside a separate storage component method.

node/src/types.rs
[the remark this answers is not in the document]
Fraser999: Thanks - makes sense.

node/src/utils/blob_cache.rs
Fraser999: OK.  It's a nitpick, so I won't insist.  I knew there'd be a compiler error, but the knee-jerk reaction to fix the compiler error I think would be to cast the smaller type to the larger.  Using an alias for `u16` as the type for `put_count` and for `garbage_collect_interval` gives a slightly stringer hint as to the author's intent that these really should remain the same type.

Not a big deal though - I'm ok with it as is.

node/src/components/storage/blob_cache.rs
```diff
@@ -0,0 +1,167 @@
...
+};
+
+use datasize::DataSize;
+
+/// A cache of serialized items.
+///
+/// Maintains a collection of weak references and automatically purges them in configurable
+/// intervals.
+#[derive(DataSize, Debug)]
+pub(crate) struct BlobCache<I> {
```
Fraser999, line 16:
```suggestion
pub(super) struct BlobCache<I> {
```

Same change for all `pub(crate)` in this module.

Fraser999 (approved):

nilz3ro (approved):
I think this change looks very interesting, have you done any performance profiling?

marc-casperlabs:
> I think this change looks very interesting, have you done any performance profiling?

Yes, but there will be some official ones later. You can test locally by setting the feature on and off.

PR #1629, opened 2021-06-24
CEP58 / request rate limiting

marc-casperlabs:
This introduces request rate limiting as described in CEP58. There are two resource pools, an unlimited one for known current and upcoming validators, and a shared one controlled by `max_non_validating_peer_request_rate`. For example, setting `max_non_validating_peer_request_rate` to `10` limits the amount of allowed `GetRequest`s for `Deploy`s to 10 per second _total_.

A large part of this PR is just the renaming of the `bandwidth_limiter` to `limiter` and resulting changes of types, comments and documentation. It may help to look at the diff that omits this renaming commit (9851442).

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -182,9 +182,15 @@ where
     /// era, so we treat the highest era we have seen as the active era.
     highest_era_seen: EraId,
 
-    /// The active bandwidth limiter.
+    /// The outgoing bandwidth limiter.
     #[data_size(skip)]
-    limiter: Box<dyn Limiter>,
+    upstream_limiter: Box<dyn Limiter>,
+
+    /// The limiter for incoming resource usage.
+    ///
+    /// This is not incoming bandwidth but an independent resource estimate.
+    #[data_size(skip)]
+    incoming_limiter: Box<dyn Limiter>,
 }
 
 impl<REv, P> SmallNetwork<REv, P>
@@ -228,13 +234,15 @@ where
             return Err(Error::EmptyKnownHosts);
         }
 
-        let limiter: Box<dyn Limiter> = if cfg.max_non_validating_peer_bps == 0 {
+        let upstream_limiter: Box<dyn Limiter> = if cfg.max_non_validating_peer_bps == 0 {
             Box::new(limiter::Unlimited)
         } else {
             Box::new(limiter::ClassBasedLimiter::new(
                 cfg.max_non_validating_peer_bps,
             ))
         };
+
+        let incoming_limiter: Box<dyn Limiter> = Box::new(limiter::Unlimited);
 
         let outgoing_manager = OutgoingManager::new(OutgoingConfig {
             retry_attempts: RECONNECTION_ATTEMPTS,
@@ -308,7 +316,8 @@ where
             server_join_handle: Some(server_join_handle),
             net_metrics,
             highest_era_seen: EraId::new(0),
-            limiter,
+            upstream_limiter,
+            incoming_limiter,
         };
 
         let effect_builder = EffectBuilder::new(event_queue);
@@ -440,7 +449,7 @@ where
                 peer_addr,
                 public_addr,
                 peer_id,
-                peer_consensus_public_key: _,
+                peer_consensus_public_key,
                 stream,
             } => {
                 info!("new incoming connection established");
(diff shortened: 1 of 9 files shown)


node/src/components/small_network.rs
```diff
@@ -182,9 +182,15 @@ where
     /// era, so we treat the highest era we have seen as the active era.
     highest_era_seen: EraId,
 
-    /// The active bandwidth limiter.
+    /// The outgoing bandwidth limiter.
     #[data_size(skip)]
-    limiter: Box<dyn BandwidthLimiter>,
+    upstream_limiter: Box<dyn Limiter>,
```
goral09, line 187: Why `Box<dyn Limiter>` and not something like `L1: Limiter, L2: Limiter`?

node/src/protocol.rs
```diff
@@ -67,6 +67,24 @@ impl Payload for Message {
...
+            Message::Consensus(_) => 0,
+            Message::DeployGossiper(_) => 0,
+            Message::AddressGossiper(_) => 0,
+            Message::GetRequest { tag, .. } | Message::GetResponse { tag, .. } => match tag {
+                Tag::Deploy => 1,
+                Tag::Block => 0,
+                Tag::GossipedAddress => 0,
+                Tag::BlockByHeight => 0,
+                Tag::BlockHeaderByHash => 0,
+                Tag::BlockHeaderAndFinalitySignaturesByHeight => 0,
```
goral09, line 83: By setting `0` for all the `Tag::Block*` variants, we're allowing for almost unlimited joining, even for non-validating nodes. Is that intentional?

Fraser999 (commented):
Mainly nitpicks - basically LGTM.

node/src/components/small_network.rs
goral09: > My own preference would be to avoid the `Limiter` trait altogether and change the `ClassBasedLimiter` to handle the single special case of being constructed with resources_per_second == 0, implying unlimited resources.

I wanted to suggest the same :+1:.

marc-casperlabs: > I believe it's so we can choose at runtime whether to hold a ClassBasedLimiter or an Unlimited one.

That's correct. Otherwise we'd only be able to switch them at compile time.

Otherwise I disagree, we may want to try additional strategies for limiting in the future (fixed rate per peer, HTBs, etc.) and this neatly captures it. Every one of these would need a special case for `== 0` or similar.

The alternative I can see is to use an `Option`, but that seemed less elegant.

node/src/protocol.rs
marc-casperlabs: Yes, we're being very conservative with the introduction of this feature. We can tweak the numbers if we have to though.

The reason deploys are throttled as the highest priority is because they offer the largest possible size. AFAIK any other message is limited to a few K in size, while deploys can be 1M.

goral09: Block's limit is 10M.

goral09: Ah, but block's limit takes into account its constituents deploys so a block _message_ should be smaller than 10MB :+1:

resources/local/config.toml
```diff
@@ -125,6 +125,11 @@ max_addr_pending_time = '1min'
 # A value of `0` means unlimited.
 max_non_validating_peer_bps = 0
 
+# The maximum amount of requests from validating peers per second answered.
```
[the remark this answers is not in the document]
marc-casperlabs: It's likely that we will be extending this to other requests, but somewhat less likely for non-requests. I kinda split the difference here.

node/src/components/small_network.rs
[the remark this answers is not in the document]
marc-casperlabs: To me `upstream` implies bandwidth, while `outgoing` does not. But I can change it.

resources/local/config.toml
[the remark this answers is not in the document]
Fraser999: Hmm.  Either way, we'll have to try and remember to update this comment when the upcoming changes happen (unless we decide to exclude get-deploy responses, which I think would be a mistake), so it makes sense to me to have this be as accurate as possible right now.

Fraser999 (approved):
LGTM

node/src/components/small_network.rs
goral09: I'm fine with the answer, resolving the comment Fraser. If you still want to continue a discussion you can unresolve it.

PR #1631, opened 2021-06-24
Query State With Optional Block Identifier

nilz3ro:
About These Changes:
* updated the RPC component to make state root hash an optional block identifier.

### DIFF
--- a/client/lib/validation.rs
+++ b/client/lib/validation.rs
@@ -17,6 +17,7 @@ use casper_types::{bytesrepr, Key, U512};
 const GET_ITEM_RESULT_BALANCE_VALUE: &str = "balance_value";
 const GET_ITEM_RESULT_STORED_VALUE: &str = "stored_value";
 const GET_ITEM_RESULT_MERKLE_PROOF: &str = "merkle_proof";
+const GET_ITEM_RESULT_STATE_ROOT_HASH: &str = "state_root_hash";
 
 /// Error that can be returned when validating a block returned from a JSON-RPC method.
 #[derive(Error, Debug)]
@@ -116,7 +117,6 @@ pub(crate) fn validate_get_era_info_response(
 
 pub(crate) fn validate_query_response(
     response: &JsonRpc,
-    state_root_hash: &Digest,
     key: &Key,
     path: &[String],
 ) -> Result<(), ValidateResponseError> {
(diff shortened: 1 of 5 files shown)


client/lib/validation.rs
```diff
@@ -164,14 +164,23 @@ pub(crate) fn validate_query_response(
...
+        let value = object
+            .get(GET_ITEM_RESULT_STATE_ROOT_HASH)
+            .ok_or(ValidateResponseError::ValidateResponseFailedToParse)?;
+
+        let value_str = value
+            .as_str()
+            .ok_or(ValidateResponseError::ValidateResponseFailedToParse)?;
+
+        let value_bytes = hex::decode(value_str)
+            .map_err(|_| ValidateResponseError::ValidateResponseFailedToParse)?;
```
xcthulhu, line 177: Perhaps we could just extend `ValidateResponseError` so we have a variant for this?

ie,

```rust
#[error(transparent)]
HexDecodeError(#[from] hex::decode::Error)
```

or something like that?  But maybe we have some downstream formatting stuff for `ValidateResponserError` this would make more complicated.

client/lib/validation.rs
```diff
@@ -164,14 +164,23 @@ pub(crate) fn validate_query_response(
...
+
+        bytesrepr::deserialize(value_bytes)?
+    };
+
+    core::validate_query_proof(&state_root_hash.into(), &proofs, key, path, proof_value)
```
xcthulhu, line 182: This is sadly more complicated now.

You have to do a few things to check integrity now:

    1. Get the block header and check it has the state root provided.
    2. Hash the block header and check it matches what was queried.
    3. Validate the proof like this

Sadly, it kind of breaks security to skip the first two steps which is a pain :(

We have an implementation of [CEP 41][1] on a branch.  Once that lands we could replace steps 1 and 2 with another kind of Merkle proof.

[1]: https://github.com/casper-network/ceps/blob/master/text/0041-extensible-block-body.md

client/lib/validation.rs
nilz3ro: Yeah I agree, I think it would be clearer

goral09 (changes requested):
This is changing an already-existing RPC endpoint and client. We don't know how many existing users rely on this behavior. We should add a new method/endpoint instead.

nilz3ro:
Deprecated by casper-network/casper-node#1645

PR #1634, opened 2021-06-26
Add event stream emitting accepted Deploys to SSE server

### LINKED ISSUE #1616: Add new event stream endpoint for accepted deploys

Fraser999:
This PR adds a new event stream to SSE server which emits `Deploy`s in full as they are accepted by a node.

The existing two streams are unaffected by this change.  The new stream can be accessed via `<IP:Port>/events/deploys` again with the optional query string `?start_from=<N>` as per the existing streams.

The `DeployAccepted` events emitted by the new stream are different from the others in that the relevant info (in this case the `Deploy`s) aren't actually held in the SSE event, but instead are retrieved from storage on demand in order to avoid exceeding reasonable memory consumption.

Note that while the stream will be accessible to clients during the node's joining phase, it will only emit keepalives during this time.  It will start emitting newly-accepted `Deploy`s once it has transitioned into the participating state (i.e. once it has finished joining).

Closes #1616.


goral09:
> Note that while the stream will be accessible to clients during the node's joining phase, it will only emit keepalives during this time. It will start emitting newly-accepted Deploys once it has transitioned into the participating state (i.e. once it has finished joining).

After finished joining, will it emit all of the deploys that were accepted during the joining phase? Why it can't emit them while joining?

node/src/reactor/participating.rs
```diff
@@ -909,6 +911,13 @@ impl reactor::Reactor for Reactor {
                 let mut effects =
                     self.dispatch_event(effect_builder, rng, Event::DeployGossiper(event));
 
+                let event = event_stream_server::Event::DeployAccepted(*deploy.id());
```
goral09, line 914: I read the description of the PR about not allocating too much memory but I am not sure I understand why. We should accept every deploy only once and immediately emit it to the event stream. Either there is a consumer subscribed – in which case we emit the deploy to the event and drop the deploy, or not – in which case we can drop the deploy immediately. 

Why do you think sending a whole deploy in `Event::DeployAccepted` could be a problem?

goral09: I think I see it now – b/c we keep `n` last events in the ring buffer of the event server, it's possible that all of those slots become occupied with `DeployAccepted` events which, if they were wrapping whole `Deploy` instances, could start occupying a lot of memory.

Fraser999: Yup - exactly.

Fraser999:
@goral09 
> After finished joining, will it emit all of the deploys that were accepted during the joining phase? Why it can't emit them while joining?

I was thinking that it shouldn't do that rather than couldn't.  I was basically mirroring what I thought was the behaviour for blocks.  However, I see you're right.  The joining node _does_ emit everything (all the way from genesis if need be), so it should do so for accepted deploys too.

I'll update the behaviour, but it'll require a bit of a rethink since transitioning from one reactor to another is a problem here.

Nice catch!  Thanks.

marc-casperlabs (approved):
I think this is as good as it gets for now, without solving the underlying architectural wrinkle.

node/src/components/event_stream_server/deploy_getter.rs
```diff
@@ -0,0 +1,107 @@
...
+                    .await
+            } else {
+                error!("no effect builder set in deploy getter");
+                return None;
+            };
+
+        if maybe_deploys.len() != 1 {
+            panic!("should return exactly one deploy");
+        }
+        maybe_deploys.pop().unwrap()
```
darthsiroftardis, line 88: I'm confused what is the type here that is being unwrapped, is it `Result<Option<Deploy>>` ?

nilz3ro (approved):
Nice work 🎊

darthsiroftardis (approved):
LGTM!!

node/src/components/event_stream_server/deploy_getter.rs
Fraser999: It's an `Option`.  `maybe_deploys` is of type `Vec<Option<Deploy>>`, and [`pop()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.pop) returns an `Option`, so in this case we're unwrapping `Option<Option<Deploy>>`.

Fraser999:
bors merge

Fraser999:
bors merge

PR #1636, opened 2021-06-27
Add configurable upper limit of concurrent event stream clients

### LINKED ISSUE #1633: Limit number of concurrent subscribers to the event stream
We should change [the unused `[event_stream_server][qps_limit]` config option](https://github.com/casper-network/casper-node/blob/release-1.2/resources/production/config-example.toml#L183-L185) to `max_concurrent_subscribers`, and limit the number of clients subscribed to the stream at any one time to this value.

Fraser999:
This PR introduces a new config option `[event_stream_server][max_concurrent_subscribers]`, which defines the maximum number of clients allowed to subscribe to a single event stream server at any given time.

The limit is coarse-grained, relating to the total number of clients, regardless of which stream(s) are subscribed to.  It is not a per-stream limit.

Tagging @sacherjj and @momipsl to advise of the change to config file options.

Closes #1633.

### DIFF
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -314,14 +314,11 @@ fn create_422() -> Response {
     response
 }
 
-/// Creates a 204 response (No Content) to be returned if the server has too many subscribers.
-///
-/// HTTP 204 is mentioned in
-/// https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events-intro as a
-/// suitable way to tell the client to stop reconnecting.
-fn create_204() -> Response {
-    let mut response = Response::new(Body::empty());
-    *response.status_mut() = StatusCode::NO_CONTENT;
+/// Creates a 503 response (Service Unavailable) to be returned if the server has too many
+/// subscribers.
+fn create_503() -> Response {
+    let mut response = Response::new(Body::from("server has reached limit of subscribers"));
+    *response.status_mut() = StatusCode::SERVICE_UNAVAILABLE;
     response
 }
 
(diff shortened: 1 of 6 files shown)


node/src/components/event_stream_server/sse_server.rs
```diff
@@ -314,10 +314,22 @@ fn create_422() -> Response {
     response
 }
 
+/// Creates a 204 response (No Content) to be returned if the server has too many subscribers.
+///
+/// HTTP 204 is mentioned in
+/// https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events-intro as a
+/// suitable way to tell the client to stop reconnecting.
+fn create_204() -> Response {
```
marc-casperlabs, line 322: I'm not sure 204 is the right response type here, as it is very misleading. The correct HTTP error code, in my opinion, is [HTTP 503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503). If you are concerned about reconnections, maybe set the [`Retry-After` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)?

node/src/components/event_stream_server/tests.rs
```diff
@@ -419,6 +433,38 @@ async fn subscribe(
...
+    handle_response(response, final_event_id, client_id).await
+}
+
+/// Handles a response from the server.
+async fn handle_response(
+    response: Response,
+    final_event_id: Id,
+    client_id: &str,
+) -> Result<Vec<ReceivedEvent>, reqwest::Error> {
+    if response.status() == StatusCode::NO_CONTENT {
```
marc-casperlabs, line 464: See notes above :).

marc-casperlabs (commented):
It would be okay with merging this, but I would prefer we do not abuse the 204 status code here and use the more appropriate 5XX status message.

marc-casperlabs (approved):
Looks good! :)

Fraser999:
bors merge

PR #1637, opened 2021-06-28
Re-enable banning faulty validators

### LINKED ISSUE #1520: Re-enable banning equivocators.
[Banning is currently disabled as all equivocators are reported in the `inactive` field instead of `equivocators`.](https://github.com/casper-network/casper-node/blob/a2d6511d17aa7df75a633542d207c280a775248e/node/src/components/consensus/era_supervisor.rs#L1056)
We need to put them back into `equivocators` so that they are banned in all future eras, even before eviction becomes effective. Since the plan is to not re-enable slashing, the EE should then treat equivocators the same as inactive validators, and only evict them.

See #1516 for more details.

afck:
Currently equivocators are treated like inactive validators. This PR will re-enable banning, and put the equivocators back into the block's `equivocators` field, making it (almost) impossible to equivocate in multiple successive eras.

It does _not_ enable slashing: Equivocators are still only evicted, just like inactive validators.

Closes #1520.


fizyk20 (approved):
Looks good, as far as I can tell :+1:

afck:
I'll put this back into draft mode and get rid of the (dead) slashing code. It's safer that way.

goral09 (commented):
Have you gotten a green light to merge this into `dev` prior to 1.3.0 release?

fizyk20 (approved):

fizyk20 (approved):

xcthulhu (approved):

afck:
bors merge

PR #1645, opened 2021-06-28
Add "query_global_state" query RPC to node

darthsiroftardis:
CHANGELOG:

- Added a new "query_global_state" RPC endpoint that takes either a `BlockHash` or `state_root_hash` param to make query to `GlobalState`
- Added a new `GlobalStateIdentifier` enum to the rpcs to support the query
- Added a new `query-global-state` command the rust client that takes in a hash which is identified as either a `block` hash or `state` root hash along with a `key` and `path` arg (similar to the preexisting) query

PS: We could use a better distinction between the "state_get_item" query and the newly added with one this PR on the client side. I am open to suggestions for names that are more distinct

Closes: [1608](https://app.zenhub.com/workspaces/engineering-60953fafb1945f0011a3592d/issues/casper-network/casper-node/1608)


nilz3ro (approved):
LGTM 🎊

client/src/global_state_query.rs
```diff
@@ -1,71 +1,85 @@
...
+            hash_value: state_root_hash,
+        };
+    }
+    if let Some(block_hash) = block_hash::get(matches) {
+        return GlobalStateStrParams {
+            is_block_hash: true,
+            hash_value: block_hash,
+        };
     }
+    unreachable!("clap arg groups and parsing should prevent this")
```
nilz3ro: Could we change this message to include the function name `global_state_str_params`? That way if this `unreachable!` is ever reached, someone will have a nice and easy time grepping for it.

nilz3ro (commented):
I found one small nitpick

client/src/global_state_query.rs
darthsiroftardis: Nice point

client/lib/lib.rs
```diff
@@ -28,9 +28,13 @@ use jsonrpc_lite::JsonRpc;
...
 
 use casper_execution_engine::core::engine_state::ExecutableDeployItem;
-use casper_node::{rpcs::state::DictionaryIdentifier, types::Deploy};
+use casper_node::{
+    rpcs::state::DictionaryIdentifier,
+    types::{BlockHash, Deploy},
+};
 use casper_types::{Key, UIntParseError, U512};
 
+use casper_node::{crypto::hash::Digest, rpcs::state::GlobalStateIdentifier};
```
Fraser999: This is in the wrong place.

client/lib/lib.rs
```diff
@@ -376,7 +380,8 @@ pub async fn get_state_root_hash(
         .await
 }
 
-/// Retrieves a stored value from the network.
+/// Retrieves a stored value from the network. This function is deprecated, use
+/// `casper_client::query_global_state` instead.
```
Fraser999: The doc comment doesn't need updated.  Instead the function should have the following attribute applied:
```rust
#[deprecated(note = "Users should use `casper_client::query_global_state` instead.")]
```
This will automatically mark the function as deprecated in the generated rustdocs.

client/lib/lib.rs
```diff
@@ -522,6 +527,48 @@ pub async fn get_account_info(
...
+///   response. If it can be parsed as an `i64` it will be used as a JSON integer. If empty, a
+///   random `i64` will be assigned. Otherwise the provided string will be used verbatim.
+/// * `node_address` is the hostname or IP and port of the node on which the HTTP service is
+///   running, e.g. `"http://127.0.0.1:7777"`.
+/// * When `verbosity_level` is `1`, the JSON-RPC request will be printed to `stdout` with long
+///   string fields (e.g. hex-formatted raw Wasm bytes) shortened to a string indicating the char
+///   count of the field.  When `verbosity_level` is greater than `1`, the request will be printed
+///   to `stdout` with no abbreviation of long fields.  When `verbosity_level` is `0`, the request
+///   will not be printed to `stdout`.
+/// * `state_identifier` is a str value that identifies the supplied 32-byte hex encoded hash as as
```
Fraser999: There is no such parameter.  This should describe `global_state_str_params`.

node/src/components/rpc_server/rpcs/state.rs
```diff
@@ -830,3 +842,141 @@ impl RpcWithParamsExt for GetDictionaryItem {
...
 }
+
+/// Identifier for possible ways to query Global State
+#[derive(Serialize, Deserialize, Debug, JsonSchema, Clone)]
+#[serde(deny_unknown_fields)]
+pub enum GlobalStateIdentifier {
+    /// Query using a block hash.
+    Block(BlockHash),
+    /// Query using the state root hash.
+    StateRoot(Digest),
```
Fraser999:
```suggestion
    BlockHash(BlockHash),
    /// Query using the state root hash.
    StateRootHash(Digest),
```

client/src/main.rs
```diff
@@ -58,6 +57,7 @@ enum DisplayOrder {
     GetRpcs,
     AccountAddress,
     GetDictionaryItem,
+    QueryGlobalState,
```
Fraser999: I think this makes more sense to go in where the `QueryState` was removed from.  Please move `GetDictionaryItem` to immediately after `QueryGlobalState` too.  This affects the order as displayed if you run `casper-client -h`, and it'd be good to group similar subcommands in as logical an order as possible.

client/src/common.rs
```diff
@@ -246,3 +246,83 @@ pub mod public_key {
...
+
+    const ARG_NAME: &str = "key";
+    const ARG_SHORT: &str = "k";
+    const ARG_VALUE_NAME: &str = "FORMATTED STRING or PATH";
+    const ARG_HELP: &str =
+        "The base key for the query. This must be a properly formatted public key, account hash, \
+        contract address hash, URef, transfer hash or deploy-info hash. The format for each \
+        respectively is \"<HEX STRING>\", \"account-hash-<HEX STRING>\", \"hash-<HEX STRING>\", \
+        \"uref-<HEX STRING>-<THREE DIGIT INTEGER>\", \"transfer-<HEX-STRING>\" and \
+        \"deploy-<HEX-STRING>\". The public key may instead be read in from a file, in which case \
```
Fraser999: The `Key` variants has been extended somewhat since this was first written - please extend the comment to cover all relevant `Key` variants.

client/lib/lib.rs
```diff
@@ -522,6 +527,48 @@ pub async fn get_account_info(
...
+/// * `key` must be a formatted [`PublicKey`](https://docs.rs/casper-node/latest/casper-node/crypto/asymmetric_key/enum.PublicKey.html)
+///   or [`Key`](https://docs.rs/casper-types/latest/casper-types/enum.PublicKey.html). This will
+///   take one of the following forms:
+/// ```text
+/// 01c9e33693951aaac23c49bee44ad6f863eedcd38c084a3a8f11237716a3df9c2c           # PublicKey
+/// account-hash-0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20  # Key::Account
+/// hash-0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20        # Key::Hash
+/// uref-0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20-007    # Key::URef
+/// transfer-0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20    # Key::Transfer
+/// deploy-0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20      # Key::DeployInfo
```
Fraser999: The `Key` variants has been extended somewhat since this was first written - please extend the comment to cover all relevant `Key` variants.

client/lib/lib.rs
```diff
@@ -1143,6 +1190,33 @@ impl<'a> TryInto<DictionaryIdentifier> for DictionaryItemStrParams<'a> {
     }
 }
 
+/// Container for `GlobalStateIdentifier` construction options.
+#[derive(Default, Debug)]
+pub struct GlobalStateStrParams<'a> {
+    /// Identifier to mark the hash as either a Block hash or `state_root_hash`
+    /// When true, the hash provided is a Block hash.
+    pub is_block_hash: bool,
+    /// The hash value.
```
Fraser999:
```suggestion
    /// The hex-encoded hash value.
```

client/src/common.rs
```diff
@@ -246,3 +246,83 @@ pub mod public_key {
         Ok(value.to_string())
     }
 }
+
+/// Handles providing the arg for and retrieval of the key.
+pub mod key {
```
Fraser999: This module and the `path` module aren't shared, so they should live in query_global_state.rs.

Fraser999 (changes requested):
There should be a `casper_query_global_state()` function added to the FFI, and `casper_get_item` should have the `#[deprecated]` attribute applied similar to the one suggested for `get_item()`.

Also, there should be test cases for this new functionality added to integration_test.rs.

client/src/common.rs
darthsiroftardis: Moved to query_global_state

client/lib/lib.rs
darthsiroftardis: Applied suggestion

client/lib/lib.rs
darthsiroftardis: Extended to include `dictionary`

client/src/common.rs
darthsiroftardis: Extended to include `dictionary`

client/src/main.rs
darthsiroftardis: Moved to where `QueryState` previously was, `GetDictionaryItem` follows `QueryGlobalState`

node/src/components/rpc_server/rpcs/state.rs
darthsiroftardis: Applied suggestion

client/lib/lib.rs
darthsiroftardis: The relevant doc string: 
```
/// * `global_state_str_params` contains global state identifier related options for this query. See
///   [`GlobalStateStrParams`](struct.GlobalStateStrParams.html) for more details.

```

client/lib/lib.rs
darthsiroftardis: Moved

client/lib/lib.rs
Fraser999: You didn't remove the description for `state_identifier` and I just noticed that `hash` should be removed too.

client/lib/ffi.rs
```diff
@@ -551,6 +551,8 @@ pub extern "C" fn casper_get_state_root_hash(
 ///
 /// See [get_item](super::get_item) for more details.
 #[no_mangle]
+#[allow(deprecated)]
```
Fraser999: We shouldn't `allow` here - only at any callsites.

darthsiroftardis: Right, so it would be directly at the callsite not at the function declaration??

client/lib/lib.rs
darthsiroftardis: Yeah, I am also noticing that the doc string for `Key` also points to an incorrect section of docs.rs

darthsiroftardis: Will change that and correct as well

client/lib/ffi.rs
darthsiroftardis: Moved to callsite

client/lib/lib.rs
darthsiroftardis: Updated the docs for `get_item` and `query_global_state` and updated the help message for the command

Fraser999 (approved):

darthsiroftardis:
bors try

Issue #1574, opened 2021-06-11
Avoid accidental overwriting of files via client


Fraser999:
The following client functions write output files:
* [`make_deploy()`](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/lib.rs#L77-L113)
* [`sign_deploy()`](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/lib.rs#L115-L140)
* [`make_transfer()`](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/lib.rs#L212-L262)

Currently these all overwrite the file if it already exists.  We want to modify the behaviour of each to require a `force: bool` parameter, [similar to `keygen::generate_files()`](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/keygen.rs#L25-L97).

Note that [the display string for the `Error::FileAlreadyExists` variant](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/error.rs#L58) should be amended to not specify `key file`, just `file`.

PR #1647, opened 2021-06-29
[NDRS-1574] Add `--force` Boolean flag to commands that may overwrite files.

nilz3ro:
# About These Changes
* Adds a `--force` Boolean flag that can be passed to `make-deploy`, `sign-deploy` or `make-transfer`.

closes casper-network/casper-node#1574

### DIFF
--- a/client/lib/lib.rs
+++ b/client/lib/lib.rs
@@ -22,7 +22,7 @@ mod parsing;
 mod rpc;
 mod validation;
 
-use std::{convert::TryInto, fs, io::Cursor};
+use std::{convert::TryInto, fs, io::Cursor, path::PathBuf};
 
 use jsonrpc_lite::JsonRpc;
 use serde::Serialize;
@@ -93,23 +93,35 @@ pub fn make_deploy(
     deploy_params: DeployStrParams<'_>,
     session_params: SessionStrParams<'_>,
     payment_params: PaymentStrParams<'_>,
+    force: bool,
 ) -> Result<()> {
-    let output = deploy::output_or_stdout(none_if_empty(maybe_output_path)).map_err(|error| {
-        Error::IoError {
-            context: format!(
-                "unable to get file or stdout, provided '{:?}'",
-                maybe_output_path
-            ),
-            error,
-        }
+    let maybe_output_path = none_if_empty(maybe_output_path);
+    let path_exists = if let Some(p) = maybe_output_path {
+        PathBuf::from(p).exists()
+    } else {
+        false
+    };
+
+    let output = deploy::output_or_stdout(maybe_output_path).map_err(|error| Error::IoError {
+        context: format!(
+            "unable to get file or stdout, provided '{:?}'",
+            maybe_output_path
+        ),
+        error,
     })?;
 
-    Deploy::with_payment_and_session(
-        deploy_params.try_into()?,
-        payment_params.try_into()?,
-        session_params.try_into()?,
-    )?
-    .write_deploy(output)
+    match (path_exists, force) {
+        (true, false) => {
+            let pb = PathBuf::from(maybe_output_path.unwrap());
+            Err(Error::FileAlreadyExists(pb))
+        }
+        (true, true) | (false, true) | (false, false) => Deploy::with_payment_and_session(
+            deploy_params.try_into()?,
+            payment_params.try_into()?,
+            session_params.try_into()?,
+        )?
+        .write_deploy(output),
+    }
 }
 
 /// Reads a previously-saved `Deploy` from a file, cryptographically signs it, and outputs it to a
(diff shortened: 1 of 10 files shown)


client/lib/lib.rs
```diff
@@ -93,23 +93,35 @@ pub fn make_deploy(
...
+        error,
     })?;
 
-    Deploy::with_payment_and_session(
-        deploy_params.try_into()?,
-        payment_params.try_into()?,
-        session_params.try_into()?,
-    )?
-    .write_deploy(output)
+    match (path_exists, force) {
```
darthsiroftardis, line 113: Maybe change this `match` to an `if let (true, false) = (path_exists, force) { err } else { o/p }`

client/lib/lib.rs
```diff
@@ -93,23 +93,35 @@ pub fn make_deploy(
...
+        (true, false) => {
+            let pb = PathBuf::from(maybe_output_path.unwrap());
+            Err(Error::FileAlreadyExists(pb))
+        }
+        (true, true) | (false, true) | (false, false) => Deploy::with_payment_and_session(
```
darthsiroftardis, line 118: Okay nitpick, but maybe mark these as constants, `const PATH_EXISTS: bool = true` , I don't think we should have naked booleans especially in this case.

darthsiroftardis (commented):
Some nitpicks

client/lib/lib.rs
nilz3ro: I don't understand how using a constant would be better than using an immutable `let` here, because `path_exists` could be true or false depending on if the path can be found.

client/lib/lib.rs
nilz3ro: I agree, the `if let (...)` will be cleaner and slightly more readable. The tradeoff is that we lose some of the logic encapsulated in the other three cases listed in the second match arm.

client/src/deploy/creation_common.rs
```diff
@@ -939,3 +940,23 @@ pub(super) mod payment_version {
         matches.value_of(ARG_NAME).unwrap_or_default()
     }
 }
+
+pub(super) mod force {
```
Fraser999, line 944: We shouldn't need this module at all - there's a suitable one in common.rs.

[Its `arg()` function](https://github.com/casper-network/casper-node/blob/dev/client/src/common.rs#L128-L139) can be passed the `DisplayOrder::Force` which you added above, and in each case the `singular` arg should be true.

client/tests/integration_test.rs
```diff
@@ -558,7 +559,50 @@ mod make_deploy {
...
+            casper_client::make_deploy(
+                file_path.to_str().unwrap(),
+                deploy_params::test_data_valid(),
+                session_params::test_data_with_package_hash(),
+                payment_params::test_data_with_name(),
+                false
+            )
+            .map_err(ErrWrapper),
+            Err(Error::FileAlreadyExists(file_path).into())
+        );
```
Fraser999, line 588: Let's add a check that the original file is unmodified by this attempt.

client/tests/integration_test.rs
```diff
@@ -668,18 +717,98 @@ mod sign_deploy {
...
+        assert_eq!(
+            casper_client::sign_deploy_file(
+                unsigned_file_path.to_str().unwrap(),
+                "../resources/local/secret_keys/node-1.pem",
+                signed_file_path.to_str().unwrap(),
+                false
+            )
+            .map_err(ErrWrapper),
+            Err(Error::FileAlreadyExists(signed_file_path).into())
+        );
```
Fraser999, line 771: Also here, let's check the failed attempt doesn't modify the pre-existing output file.

client/tests/integration_test.rs
```diff
@@ -718,7 +848,54 @@ mod make_transfer {
...
+                AMOUNT,
+                TARGET_ACCOUNT,
+                TRANSFER_ID,
+                deploy_params::test_data_valid(),
+                payment_params::test_data_with_name(),
+                false
+            )
+            .map_err(ErrWrapper),
+            Err(Error::FileAlreadyExists(file_path).into())
+        );
```
Fraser999, line 879: And again here.

Fraser999 (changes requested):
A few points to be changed, but heading in the right direction, particularly in the addition of tests.  [The ticket](https://github.com/casper-network/casper-node/issues/1574) also mentions the following which is not actioned in this PR yet:

> Note that [the display string for the `Error::FileAlreadyExists` variant](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/error.rs#L58) should be amended to not specify `key file`, just `file`.

client/lib/lib.rs
```diff
@@ -93,23 +93,35 @@ pub fn make_deploy(
...
-            error,
-        }
+    let maybe_output_path = none_if_empty(maybe_output_path);
+    let path_exists = if let Some(p) = maybe_output_path {
+        PathBuf::from(p).exists()
+    } else {
+        false
+    };
+
+    let output = deploy::output_or_stdout(maybe_output_path).map_err(|error| Error::IoError {
```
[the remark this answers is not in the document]
nilz3ro: done

client/lib/lib.rs
```diff
@@ -80,29 +79,28 @@ pub fn put_deploy(
...
+/// * `maybe_output_path` specifies the output file, or if empty, will print it to `stdout`.
 /// * `deploy_params` contains deploy-related options for this `Deploy`. See
 ///   [`DeployStrParams`](struct.DeployStrParams.html) for more details.
 /// * `session_params` contains session-related options for this `Deploy`. See
 ///   [`SessionStrParams`](struct.SessionStrParams.html) for more details.
 /// * `payment_params` contains payment-related options for this `Deploy`. See
 ///   [`PaymentStrParams`](struct.PaymentStrParams.html) for more details.
+/// * If `force` is true, existing files will be overwritten. If `force` is false and any of the
+/// files exist, [`Error::FileAlreadyExists`](../enum.Error.html#variant.FileAlreadyExists) is
+/// returned and no files are written.
```
Fraser999, line 91: This was copied verbatim from a function where multiple output files are used.  In this case it's a single output file.  Also, the generated link is dead, since it points to `../` which is incorrect here.

```suggestion
/// * If `force` is true, and the output file exists, it will be overwritten. If `force` is false
///   and the output file exists,
///   [`Error::FileAlreadyExists`](enum.Error.html#variant.FileAlreadyExists) is returned and the
///   output file is left unchanged.
```

This change should be applied to `sign_deploy_file()` and `make_transfer()` too.

Issue #1575, opened 2021-06-11
Ensure client sign-deploy subcommand doesn't damage input file on failure


Fraser999:
Currently [the execution flow in `sign_deploy_file()`](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/lib.rs#L115-L140) is:
* parse the deploy from the input path
* create the output writer (could be a new file, or stdout)
* sign the deploy and write it to the output writer

This will be a problem in the case where the input path is the same as the output path, and signing or writing fails, since step 2 will have deleted the contents of the file.

We need to ensure that in the case where the output file pre-exists, it is not damaged if the command fails.  It seems like an appropriate way to do this is to create the output writer as a temporary file, and once successfully written, move it to the actual specified output location.

PR #1652, opened 2021-07-01
NDRS-1575 Ensure Client Does Not Damage Files On Failure.

nilz3ro:
# About These Changes
* changes `make-deploy`, `make-transfer` and `sign-deploy` to ensure that they do not damage or overwrite existing files if something goes wrong when writing the output file.

closes casper-network/casper-node#1575

### DIFF
--- a/client/lib/deploy.rs
+++ b/client/lib/deploy.rs
@@ -1,10 +1,10 @@
+use rand::{self, distributions::Alphanumeric, Rng};
+use serde::{Deserialize, Serialize};
 use std::{
     fs::{self, File},
     io::{self, BufReader, Read, Write},
-    path::PathBuf,
+    path::{Path, PathBuf},
 };
-
-use serde::{Deserialize, Serialize};
 
 use casper_execution_engine::core::engine_state::ExecutableDeployItem;
 use casper_node::{
@@ -76,7 +76,7 @@ impl From<GetBlockResult> for ListDeploysResult {
 pub(super) enum OutputKind<'a> {
     File {
         path: &'a str,
-        tmp_path: String,
+        tmp_path: PathBuf,
         overwrite_if_exists: bool,
     },
     Stdout,
@@ -84,7 +84,13 @@ pub(super) enum OutputKind<'a> {
 
 impl<'a> OutputKind<'a> {
     pub(super) fn file(path: &'a str, overwrite_if_exists: bool) -> Self {
-        let tmp_path = Path::new(path).with_extension(".tmp");
+        let collision_resistant_string = rand::thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(64)
+            .map(char::from)
+            .collect::<String>();
+        let extension = format!(".{}.tmp", &collision_resistant_string);
+        let tmp_path = Path::new(path).with_extension(extension);
         OutputKind::File {
             path,
             tmp_path,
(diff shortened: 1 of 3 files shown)


client/lib/deploy.rs
```diff
@@ -76,31 +76,37 @@ impl From<GetBlockResult> for ListDeploysResult {
 pub(super) enum OutputKind<'a> {
     File {
         path: &'a str,
+        tmp_path: String,
```
Fraser999, line 79: Since we're generating `tmp_path` using `Path` functionality (see later comment) I'd be more inclined to hold this as a `PathBuf`.

client/lib/deploy.rs
```diff
@@ -76,31 +76,37 @@ impl From<GetBlockResult> for ListDeploysResult {
...
         overwrite_if_exists: bool,
     },
     Stdout,
 }
 
 impl<'a> OutputKind<'a> {
     pub(super) fn file(path: &'a str, overwrite_if_exists: bool) -> Self {
+        let tmp_path = [path.clone(), ".tmp"].join("");
```
Fraser999, line 87: Probably more canonical to do:

```suggestion
        let tmp_path = Path::new(path).with_extension(".tmp");
```

client/lib/deploy.rs
```diff
@@ -76,31 +76,37 @@ impl From<GetBlockResult> for ListDeploysResult {
...
+    pub(super) fn get(&self) -> Result<Box<dyn Write>> {
         match self {
             OutputKind::File {
                 path,
+                tmp_path,
                 overwrite_if_exists,
+                ..
             } => {
                 let path = PathBuf::from(path);
+                let tmp_path = PathBuf::from(tmp_path);
```
Fraser999, line 104: We won't need this if `tmp_path` is changed to be a `PathBuf`, but if this were still needed, I'd suggest moving it to just before it's actually used.  I think it's better to not create variables until they're actually needed.

client/lib/deploy.rs
```diff
@@ -111,6 +117,22 @@ impl<'a> OutputKind<'a> {
...
+            OutputKind::File { path, tmp_path, .. } => match fs::rename(tmp_path.clone(), path) {
+                Ok(_) => Ok(()),
+                Err(error) => Err(Error::IoError {
+                    context: format!(
+                        "Could not move tmp file {} to destination {}",
+                        tmp_path, path
+                    ),
+                    error,
+                }),
+            },
```
Fraser999, line 132: Since `fs::rename()` returns a `Result<(), io::Error>`, we don't need to change the `Ok` type here - just mapping the `Err` type is enough.  Also, we don't need to clone `tmp_path`:

```suggestion
            OutputKind::File { path, tmp_path, .. } => {
                fs::rename(&tmp_path, path).map_err(|error| Error::IoError {
                    context: format!(
                        "Could not move tmp file {} to destination {}",
                        tmp_path.display(),
                        path
                    ),
                    error,
                })
            }
```

Fraser999 (commented):
Just some nitpicks.

client/lib/deploy.rs
nilz3ro: Good point

Fraser999 (approved):

nilz3ro:
bors merge

PR #1656, opened 2021-07-05
Apply standstill alert only for joining nodes.

### LINKED ISSUE #1501: Apply standstill alert only for joining nodes.
We addressed two scenarios with somewhat conflicting solutions:

**Scenario 1**: 34% of the validators go temporarily offline. During that time, the remaining nodes cannot finalize anything, so the units they add to the protocol state are not very useful.

**Solution 1**: If 34% haven't created a message for a while we go into "pause mode", where instead of new units we only send ping messages. As soon as we receive pings from 67% again, we resume creating units. That way the protocol state doesn't unnecessarily grow during the outage.

**Scenario 2**: A joining node is badly connected and fails syncing its protocol state.

**Solution 2**: We have a "standstill timeout" after which we once request our peers protocol states, and if after another timeout we still haven't added any new units, we shutdown: After a restart the connections may be better.

The problem is that solution 2 causes all nodes in the network to restart if the outage in scenario 1 takes too long. During the restart they lose their protocol state, so the only way to recover is an emergency restart.

Let's apply the standstill timeout only once, for a newly joining node, so that a long running node will never shutdown due to standstill.

afck:
Don't automatically restart a running node when the current era doesn't make progress; only restart if that happens shortly after joining the network.

Closes #1501.

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -89,7 +89,7 @@ pub struct EraSupervisor<I> {
     /// current one.
     active_eras: HashMap<EraId, Era<I>>,
     secret_signing_key: Arc<SecretKey>,
-    pub(super) public_signing_key: PublicKey,
+    public_signing_key: PublicKey,
     current_era: EraId,
     protocol_config: ProtocolConfig,
     config: Config,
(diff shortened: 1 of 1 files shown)


goral09 (approved):

fizyk20 (approved):

xcthulhu (approved):

afck:
bors merge

PR #1657, opened 2021-07-05
Avoid emitting duplicate events in the event stream

### LINKED ISSUE #1648: Avoid duplicate events in event stream
A recent CI test failure which has not been able to be reproduced yet exposed an issue in the event stream server whereby a client using the `?start_from=N` query string could receive a duplicated event.

This happens due a timing issue.  When the client requests historical events from the ring buffer, these are sent via an `mpsc` channel to the tokio task handling that specific client, and the task continues to receive further events generated by the server via a `broadcast` channel.  

This leaves a small time window for the following sequence:
* a new client connects, ready to receive broadcasted events
* a new event is sent by the http server via the broadcast channel
* that event is added to the end of the http server's buffer
* the http server handles the new client's request for historical data by cloning events from the buffer, up to and including the last one
* the tokio task handling that client now receives all historical events, followed by broadcasted events since it connected.  The last buffered event is duplicated as the first broadcasted event.

It seems like the most robust way to handle this is to have the tokio task handling the client's events deduplicate events after it transitions from receiving the historical events to receiving the new events.

Fraser999:
As described in #1648, there was an edge case where the event stream server could emit duplicate events when a client requested a number of historic events via the `?start_from=N` query string.

This PR fixes the issue and adds a regression test for it.

Closes #1648.

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -47,6 +47,7 @@ All notable changes to this project will be documented in this file.  The format
 * Drain the joiner reactor queue on exit, to eliminate stale connections whose handshake has completed, but which live on the queue.
 * Shut down SSE event streams gracefully.
 * Limit the maximum number of clients connected to the event stream server via the `[event_stream_server][max_concurrent_subscribers]` config option.
+* Avoid emitting duplicate events in the event stream.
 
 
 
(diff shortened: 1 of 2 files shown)


darthsiroftardis (approved):
LGTM!

nilz3ro (approved):
LGTM

Fraser999:
bors merge

Issue #1577, opened 2021-06-11
transfer_id should be a u64 throughout client


Fraser999:
In `client/lib/lib.rs`, all instances of `transfer_id` should be changed from a `&str`s to `u64`s (ensure the doc comments are appropriately updated).

Likewise, in `client/lib/ffi.rs`, they should be changed from `*const c_char`s to `u64`s.

This should also allow us to delete [`parsing::transfer_id()`](https://github.com/casper-network/casper-node/blob/release-1.2/client/lib/parsing.rs#L575-L579).

PR #1661, opened 2021-07-07
Transfer Id should be u64.

nilz3ro:
# About These Changes
* Makes `transfer_id` `u64` throughout the client lib and FFI.
* closes casper-network/casper-node#1577


client/lib/ffi.rs
```diff
@@ -379,7 +379,7 @@ pub extern "C" fn casper_transfer(
     let node_address = try_unsafe_arg!(node_address);
     let amount = try_unsafe_arg!(amount);
     let maybe_target_account = try_unsafe_arg!(maybe_target_account);
-    let transfer_id = try_unsafe_arg!(transfer_id);
+    // let transfer_id = try_unsafe_arg!(transfer_id);
```
darthsiroftardis, line 382: Leftover comment from the change

client/lib/ffi.rs
```diff
@@ -407,15 +407,15 @@ pub extern "C" fn casper_make_transfer(
...
+    transfer_id: u64,
     deploy_params: *const casper_deploy_params_t,
     payment_params: *const casper_payment_params_t,
     force: bool,
 ) -> casper_error_t {
     let maybe_output_path = try_unsafe_arg!(maybe_output_path);
     let amount = try_unsafe_arg!(amount);
     let maybe_target_account = try_unsafe_arg!(maybe_target_account);
-    let transfer_id = try_unsafe_arg!(transfer_id);
+    // let transfer_id = try_unsafe_arg!(transfer_id);
```
darthsiroftardis, line 418: Leftover comment

darthsiroftardis (approved):
LGTM, can be merged once stray comments have been removed

client/tests/integration_test.rs
```diff
@@ -833,7 +833,7 @@ mod make_transfer {
     const AMOUNT: &str = "30000000000";
     const TARGET_ACCOUNT: &str =
         "01aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
-    const TRANSFER_ID: &str = "1314";
+    const TRANSFER_ID: u64 = 1314;
```
goral09, line 836: This means all clients' (if we have any :sweat_smile:) using Rust client in their bigger ecosystem will have to update? If I have scripts that use `casper-client` Rust binary, would they still be able to use a string version?

Fraser999: Other clients using our Rust lib (either as a Rust lib or as a C lib) will need to update.  Usage of the binary is unaffected by this change - the binary currently errors out if a user passes something which doesn't parse as a `u64` for the transfer ID arg.

Fraser999 (changes requested):
This PR should also include an update to the client's CHANGELOG.md.

Otherwise LGTM.

nilz3ro:
bors merge

Fraser999 (approved):

nilz3ro:
bors merge

EdHastingsCasperLabs (changes requested):
hold up on this; we've been requested to change TransferId's to support strings...needs to be discussed prior to merging this

PR #1663, opened 2021-07-08
Fast sync: Download all blocks with not-yet-expired deploys.

### LINKED ISSUE #1596: Fast Sync Must Fetch Deploy Hashes And Save Going Back to The TTL

afck:
Download all blocks younger than the maximum deploy TTL when joining:
These deploy hashes are necessary for detecting replayed deploys in new blocks.

Closes #1596.


afck:
Currently in draft mode: This is blocked because the contract runtime currently fails if there are blocks for which we don't have the global state.

Issue #1658, opened 2021-07-06
Fast Sync Must Execute Blocks To Current


xcthulhu:
- [ ] Allow the `ContractRuntime` to make execute block execution calls with a responder (PR #1669)
- [ ] Call `ContractRuntime` from the fast-sync loop (PR #1712)

xcthulhu:
Closed with #1669

PR #1669, opened 2021-07-08
Refactor ContractRuntime

xcthulhu:
Refactors the contract runtime so it can handle calls using a responder.

Part of https://github.com/casper-network/casper-node/issues/1658

Closes https://github.com/casper-network/casper-node/issues/1520

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -56,12 +56,12 @@ pub use self::{
     execute_request::ExecuteRequest,
     execution::Error as ExecError,
     execution_result::{ExecutionResult, ExecutionResults, ForcedTransferResult},
-    genesis::{ExecConfig, GenesisAccount, GenesisResult},
+    genesis::{ExecConfig, GenesisAccount, GenesisSuccess},
     query::{GetBidsRequest, GetBidsResult, QueryRequest, QueryResult},
-    step::{RewardItem, SlashItem, StepRequest, StepResult},
+    step::{RewardItem, SlashItem, StepError, StepRequest, StepSuccess},
     system_contract_cache::SystemContractCache,
     transfer::{TransferArgs, TransferRuntimeArgsBuilder, TransferTargetMode},
-    upgrade::{UpgradeConfig, UpgradeResult},
+    upgrade::{UpgradeConfig, UpgradeSuccess},
 };
 use crate::{
     core::{
@@ -82,11 +82,7 @@ use crate::{
         transform::Transform,
         wasm_prep::Preprocessor,
     },
-    storage::{
-        global_state::{CommitResult, StateProvider},
-        protocol_data::ProtocolData,
-        trie::Trie,
-    },
+    storage::{global_state::StateProvider, protocol_data::ProtocolData, trie::Trie},
 };
 
 pub const MAX_PAYMENT_AMOUNT: u64 = 2_500_000_000;
@@ -138,7 +134,7 @@ where
         genesis_config_hash: Blake2bHash,
         protocol_version: ProtocolVersion,
         ee_config: &ExecConfig,
-    ) -> Result<GenesisResult, Error> {
+    ) -> Result<GenesisSuccess, Error> {
         // Preliminaries
         let initial_root_hash = self.state.empty_root();
         let system_config = ee_config.system_config();
@@ -202,26 +198,27 @@ where
         // Commit the transforms.
         let execution_effect = genesis_installer.finalize();
 
-        let commit_result = self
+        let post_state_hash = self
             .state
             .commit(
                 correlation_id,
                 initial_root_hash,
                 execution_effect.transforms.to_owned(),
             )
-            .map_err(Into::into)?;
+            .map_err(Into::<execution::Error>::into)?;
 
         // Return the result
-        let genesis_result = GenesisResult::from_commit_result(commit_result, execution_effect);
-
-        Ok(genesis_result)
+        Ok(GenesisSuccess {
+            post_state_hash,
+            execution_effect,
+        })
     }
 
     pub fn commit_upgrade(
         &self,
         correlation_id: CorrelationId,
         upgrade_config: UpgradeConfig,
-    ) -> Result<UpgradeResult, Error> {
+    ) -> Result<UpgradeSuccess, Error> {
         // per specification:
         // https://casperlabs.atlassian.net/wiki/spaces/EN/pages/139854367/Upgrading+System+Contracts+Specification
 
@@ -230,7 +227,7 @@ where
         let pre_state_hash = upgrade_config.pre_state_hash();
         let tracking_copy = match self.tracking_copy(pre_state_hash)? {
             Some(tracking_copy) => Rc::new(RefCell::new(tracking_copy)),
-            None => return Ok(UpgradeResult::RootNotFound),
+            None => return Err(Error::RootNotFound(pre_state_hash)),
         };
 
         // 3.1.1.1.1.2 current protocol version is required
@@ -375,20 +372,23 @@ where
             tracking_copy.borrow_mut().write(*key, value.clone());
         }
 
-        let effects = tracking_copy.borrow().effect();
+        let execution_effect = tracking_copy.borrow().effect();
 
         // commit
-        let commit_result = self
+        let post_state_hash = self
             .state
             .commit(
                 correlation_id,
                 pre_state_hash,
-                effects.transforms.to_owned(),
+                execution_effect.transforms.to_owned(),
             )
             .map_err(Into::into)?;
 
         // return result and effects
-        Ok(UpgradeResult::from_commit_result(commit_result, effects))
+        Ok(UpgradeSuccess {
+            post_state_hash,
+            execution_effect,
+        })
     }
 
     pub fn tracking_copy(
@@ -1640,7 +1640,7 @@ where
         correlation_id: CorrelationId,
         pre_state_hash: Blake2bHash,
         effects: AdditiveMap<Key, Transform>,
-    ) -> Result<CommitResult, Error>
+    ) -> Result<Blake2bHash, Error>
     where
         Error: From<S::Error>,
     {
@@ -1822,18 +1822,20 @@ where
         &self,
         correlation_id: CorrelationId,
         step_request: StepRequest,
-    ) -> Result<StepResult, Error> {
+    ) -> Result<StepSuccess, StepError> {
         let protocol_data = match self.state.get_protocol_data(step_request.protocol_version) {
             Ok(Some(protocol_data)) => protocol_data,
             Ok(None) => {
-                return Ok(StepResult::InvalidProtocolVersion);
+                return Err(StepError::InvalidProtocolVersion(
+                    step_request.protocol_version,
+                ));
             }
-            Err(error) => return Ok(StepResult::GetProtocolDataError(Error::Exec(error.into()))),
+            Err(error) => return Err(StepError::GetProtocolDataError(Error::Exec(error.into()))),
         };
 
         let tracking_copy = match self.tracking_copy(step_request.pre_state_hash) {
-            Err(error) => return Ok(StepResult::TrackingCopyError(error)),
-            Ok(None) => return Ok(StepResult::RootNotFound),
+            Err(error) => return Err(StepError::TrackingCopyError(error)),
+            Ok(None) => return Err(StepError::RootNotFound(step_request.pre_state_hash)),
             Ok(Some(tracking_copy)) => Rc::new(RefCell::new(tracking_copy)),
         };
 
@@ -1852,14 +1854,14 @@ where
         {
             Ok(contract) => contract,
             Err(error) => {
-                return Ok(StepResult::GetContractError(error.into()));
+                return Err(StepError::GetContractError(error.into()));
             }
         };
 
         let system_module = match tracking_copy.borrow_mut().get_system_module(&preprocessor) {
             Ok(module) => module,
             Err(error) => {
-                return Ok(StepResult::GetSystemModuleError(error.into()));
+                return Err(StepError::GetSystemModuleError(error.into()));
             }
         };
 
@@ -1895,21 +1897,14 @@ where
                     "failed to deserialize reward factors: {}",
                     error.to_string()
                 );
-                return Ok(StepResult::Serialization(error));
+                return Err(StepError::BytesRepr(error));
             }
         };
 
-        let reward_args = {
-            let maybe_runtime_args = RuntimeArgs::try_new(|args| {
-                args.insert(ARG_REWARD_FACTORS, reward_factors)?;
-                Ok(())
-            });
-
-            match maybe_runtime_args {
-                Ok(runtime_args) => runtime_args,
-                Err(error) => return Ok(StepResult::CLValueError(error)),
-            }
-        };
+        let reward_args = RuntimeArgs::try_new(|args| {
+            args.insert(ARG_REWARD_FACTORS, reward_factors)?;
+            Ok(())
+        })?;
 
         let distribute_rewards_call_stack = {
             let system = CallStackElement::session(PublicKey::System.to_account_hash());
@@ -1941,7 +1936,7 @@ where
         );
 
         if let Some(exec_error) = execution_result.take_error() {
-            return Ok(StepResult::DistributeError(exec_error));
+            return Err(StepError::DistributeError(exec_error));
         }
 
         let slashed_validators = match step_request.slashed_validators() {
@@ -1951,7 +1946,7 @@ where
                     "failed to deserialize validator_ids for slashing: {}",
                     error.to_string()
                 );
-                return Ok(StepResult::Serialization(error));
+                return Err(StepError::BytesRepr(error));
             }
         };
 
@@ -1993,32 +1988,25 @@ where
         );
 
         if let Some(exec_error) = execution_result.take_error() {
-            return Ok(StepResult::SlashingError(exec_error));
+            return Err(StepError::SlashingError(exec_error));
         }
 
         if step_request.run_auction {
-            let run_auction_args = {
-                let maybe_runtime_args = RuntimeArgs::try_new(|args| {
-                    args.insert(
-                        ARG_ERA_END_TIMESTAMP_MILLIS,
-                        step_request.era_end_timestamp_millis,
-                    )?;
-                    args.insert(
-                        ARG_EVICTED_VALIDATORS,
-                        step_request
-                            .evict_items
-                            .iter()
-                            .map(|item| item.validator_id.clone())
-                            .collect::<Vec<PublicKey>>(),
-                    )?;
-                    Ok(())
-                });
-
-                match maybe_runtime_args {
-                    Ok(runtime_args) => runtime_args,
-                    Err(error) => return Ok(StepResult::CLValueError(error)),
-                }
-            };
+            let run_auction_args = RuntimeArgs::try_new(|args| {
+                args.insert(
+                    ARG_ERA_END_TIMESTAMP_MILLIS,
+                    step_request.era_end_timestamp_millis,
+                )?;
+                args.insert(
+                    ARG_EVICTED_VALIDATORS,
+                    step_request
+                        .evict_items
+                        .iter()
+                        .map(|item| item.validator_id.clone())
+                        .collect::<Vec<PublicKey>>(),
+                )?;
+                Ok(())
+            })?;
 
             let run_auction_call_stack = {
                 let system = CallStackElement::session(PublicKey::System.to_account_hash());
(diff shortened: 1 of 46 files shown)


node/src/components/contract_runtime/error.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// An error that occurred when committing execution results to the trie.
+    #[error(transparent)]
+    CommitError(#[from] CommitError),
+    /// An error that occurred while creating a block.
+    #[error(transparent)]
+    BlockCreationError(#[from] BlockCreationError),
+}
+
+/// An error emitted by the execution engine on commit
+// TODO: Consider replacing CommitResult with Result<Blake2bHash, CommitError>
```
afck, line 45: That seems obsolete now?
```suggestion
```

node/src/components/contract_runtime/operations.rs
```diff
@@ -1,31 +1,145 @@
...
+        trace!(?deploy_hash, ?result, "deploy execution result");
+        // As for now a given state is expected to exist.
+        let (state_hash, execution_result) =
+            commit_execution_effects(engine_state, metrics, state_root_hash, deploy_hash, result)?;
+        execution_results.insert(deploy_hash, (deploy_header, execution_result));
+        state_root_hash = state_hash;
+    }
+
+    // If the finalized block has an era report, run the auction contract
+    let maybe_step_result_success = match finalized_block.era_report() {
```
afck, line 79:
```suggestion
    let maybe_step_success = match finalized_block.era_report() {
```

node/src/components/contract_runtime/operations.rs
```diff
@@ -48,47 +162,33 @@ pub(super) async fn commit_execution_effects(
...
-        }
-    }
+    )?;
+    Ok((new_state_root, execution_result))
 }
 
-pub(super) async fn commit(
-    engine_state: Arc<EngineState<LmdbGlobalState>>,
-    metrics: Arc<ContractRuntimeMetrics>,
+fn commit(
```
afck, line 174:
```suggestion
fn commit_transforms(
```

node/src/components/contract_runtime/operations.rs
```diff
@@ -48,47 +162,33 @@ pub(super) async fn commit_execution_effects(
...
+    engine_state: &EngineState<LmdbGlobalState>,
+    metrics: &ContractRuntimeMetrics,
     state_root_hash: Digest,
     effects: AdditiveMap<Key, Transform>,
-) -> Result<CommitResult, engine_state::Error> {
+) -> Result<Digest, engine_state::Error> {
     trace!(?state_root_hash, ?effects, "commit");
     let correlation_id = CorrelationId::new();
     let start = Instant::now();
     let result = engine_state.apply_effect(correlation_id, state_root_hash.into(), effects);
```
afck, line 183: So `apply_effect` already writes the result to the database?

node/src/components/contract_runtime/operations.rs
```diff
@@ -99,3 +199,52 @@ pub(super) async fn execute(
...
+    metrics: &ContractRuntimeMetrics,
+    protocol_version: ProtocolVersion,
+    pre_state_root_hash: Digest,
+    era_report: &EraReport<PublicKey>,
+    era_end_timestamp_millis: u64,
+    next_era_id: EraId,
+) -> Result<StepSuccess, StepError> {
+    // Extract the rewards and the inactive validators if this is a switch block
+    let EraReport {
+        // Note: Highway does not slash, do nothing with the equivocators
```
afck, line 214: Should we log an error if it's not empty? Or just evict them? (That's what https://github.com/casper-network/casper-node/pull/1637 will do anyway.)

node/src/components/contract_runtime/operations.rs
```diff
@@ -99,3 +199,52 @@ pub(super) async fn execute(
...
+        .clone()
+        .into_iter()
+        .map(EvictItem::new)
+        .collect();
+
+    let step_request = StepRequest {
+        pre_state_hash: pre_state_root_hash.into(),
+        protocol_version,
+        reward_items,
+        // Note: Highway does not slash; but another consensus protocol (e.g., BABE) could
```
afck, line 235: Whether to slash or not is not determined by the consensus protocol. (Also above.)
```suggestion
        // Note: The Casper Network does not slash, but another network could
```

node/src/components/linear_chain_sync.rs
```diff
@@ -952,3 +957,36 @@ pub(crate) fn clean_linear_chain_state(
     let key = create_state_key(&chainspec);
     storage.del_state_store(key)
 }
+
+async fn execute_finalized_block<REv>(
```
afck, line 961: This is duplicated in the era supervisor, is it? We can deduplicate this if we move it to e.g. `utils` — or some new `common` module for this type of logic?

node/src/effect.rs
```diff
@@ -1196,13 +1169,16 @@ impl<REv> EffectBuilder<REv> {
     }
 
     /// Passes a finalized proto-block to the block executor component to execute it.
```
afck, line 1171:
```suggestion
    /// Passes a finalized proto-block to the block executor component to execute it.
    /// The `deploys` must match the deploy and transfer hashes in the finalized block, in order.
```

node/src/effect.rs
```diff
@@ -1756,6 +1714,6 @@ impl<REv> EffectBuilder<REv> {
 #[macro_export]
 macro_rules! fatal {
     ($effect_builder:expr, $($arg:tt)*) => {
-        $effect_builder.fatal(file!(), line!(), format_args!($($arg)*).to_string())
+        $effect_builder.fatal(file!(), line!(), format!($($arg)*).to_string())
```
afck, line 1717: This should be redundant now?
```suggestion
        $effect_builder.fatal(file!(), line!(), format!($($arg)*))
```

node/src/reactor/participating.rs
```diff
@@ -125,6 +122,9 @@ pub enum Event {
     LinearChain(#[serde(skip_serializing)] linear_chain::Event<NodeId>),
 
     // Requests
+    /// Contract runtime event.
```
afck, line 125:
```suggestion
    /// Contract runtime request.
```

afck (commented):
Looks great to me; it simplifies the code a lot! :+1: 

Someone more familiar with the EE should review it, too, though.

node/src/components/contract_runtime.rs
```diff
@@ -726,411 +541,79 @@ impl ContractRuntime {
...
+                    .announce_step_success(era_id, step_execution_effect)
+                    .ignore(),
+            );
+        }
+
         // If the child is already finalized, start execution.
-        if let Some((finalized_block, deploys)) = self.exec_queue.remove(&next_height) {
-            effects.extend(self.handle_get_deploys_result(
+        if let Some((finalized_block, deploys)) = self.exec_queue.remove(&(block_height + 1)) {
+            effects.extend(self.execute_finalized_block(
```
dwerner, line 595: @xcthulhu you mentioned this in our review of these changes. Does this mean a breaking change? Could you outline a bit about that?

execution_engine/src/core/engine_state/step.rs
```diff
@@ -104,32 +104,47 @@ impl StepRequest {
...
     GetProtocolDataError(Error),
+    #[error("Tracking copy error: {0}")]
     TrackingCopyError(Error),
+    #[error("Get contract error: {0}")]
     GetContractError(Error),
+    #[error("Get system module error: {0}")]
     GetSystemModuleError(Error),
+    #[error("Slashing error: {0}")]
     SlashingError(Error),
+    #[error("Slashing error: {0}")]
```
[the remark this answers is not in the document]
xcthulhu: :heavy_check_mark:

execution_engine/src/core/engine_state/step.rs
```diff
@@ -104,32 +104,47 @@ impl StepRequest {
...
-        next_era_validators: BTreeMap<PublicKey, U512>,
-        execution_effect: ExecutionEffect,
-    },
+    #[error(transparent)]
+    BytesRepr(#[from] bytesrepr::Error),
+    #[error(transparent)]
+    CLValueError(#[from] CLValueError),
+    #[error(transparent)]
+    GetEraValidatorsError(#[from] GetEraValidatorsError),
+    #[error(transparent)]
```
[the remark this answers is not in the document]
xcthulhu: :heavy_check_mark:

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -200,28 +196,29 @@ where
...
-        let commit_result = self
+        let post_state_hash = self
             .state
             .commit(
                 correlation_id,
                 initial_root_hash,
-                execution_effect.transforms.to_owned(),
+                effect.transforms.to_owned(),
             )
             .map_err(Into::into)?;
```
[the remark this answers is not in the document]
xcthulhu: Gives me a compilation error: 

```
the trait `From<<S as StateProvider>::Error>` is not implemented for `engine_state::error::Error`
```

I've got the constraint above:

```rust
impl<S> EngineState<S>
where
    S: StateProvider,
    S::Error: Into<execution::Error>,
```

I've gone ahead and typed the `Into::into` to make it explicit in b6100bbca6b82629110d8398bff0437ff18db54d

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -19,59 +19,19 @@ use crate::{
...
-    KeyNotFound(Key),
-    TypeMismatch(TypeMismatch),
-    Serialization(bytesrepr::Error),
-    Success {
-        post_state_hash: Blake2bHash,
-        effect: ExecutionEffect,
-    },
+pub struct UpgradeSuccess {
+    pub post_state_hash: Blake2bHash,
+    pub effect: ExecutionEffect,
```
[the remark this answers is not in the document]
xcthulhu: Fixed in 0fb1074d1bc1affc62d1dd4733c0556ec2b34546

node/src/components/chainspec_loader.rs
```diff
@@ -518,28 +548,19 @@ impl ChainspecLoader {
 
     fn handle_commit_genesis_result(
         &mut self,
-        result: Result<GenesisResult, engine_state::Error>,
+        result: Result<GenesisSuccess, engine_state::Error>,
     ) -> Effects<Event> {
```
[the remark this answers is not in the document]
xcthulhu: Fixed in 3fbb1b0da68cf1f76a8fbfcf634f1f40aa28f40d

node/src/components/chainspec_loader.rs
```diff
@@ -550,28 +571,19 @@ impl ChainspecLoader {
 
     fn handle_upgrade_result(
         &mut self,
-        result: Result<UpgradeResult, engine_state::Error>,
+        result: Result<UpgradeSuccess, engine_state::Error>,
     ) -> Effects<Event> {
```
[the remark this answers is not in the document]
xcthulhu: Fixed in 531aacb0020a323d618d7cfbefe1f6de48c28150

dwerner:
Looking good to me.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1269,6 +1272,39 @@ where
     }
 }
 
+async fn execute_finalized_block<REv>(
```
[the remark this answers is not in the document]
xcthulhu: I agree that the fact this code has to get the deploys is a smell.

Here is my proposal (if you agree, we can make a ticket):

We change `EffectBuilder::validate_block` to be:

```rust
async fn validate_block<I>(self, sender: I, block: ProposedBlock) -> Result<ValidatedBlock, BlockValidationError>
```

Where:

```rust
struct ValidatedBlock {
   proposed_block: ProposedBlock,
   deploys: Deploys
}
```

Replay detection logic would move into the block validator and be rolled up into `BlockValidationError`.

This would improve type correctness.  Any method that should operate on a `ValidatedBlock`, rather than a merely `ProposedBlock`, will have to say-so in its type signature.

While it's true that `deploys` would still be `ValidatedBlock`, it's an implementation detail.  If we want, we could put `ValidatedBlock` in `linear_chain.rs` and make `ValidatedBlock::deploys` a private method.

WDYT?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1269,6 +1272,39 @@ where
...
+    effect_builder: EffectBuilder<REv>,
+    finalized_block: FinalizedBlock,
+) where
+    REv: From<StorageRequest> + From<ControlAnnouncement> + From<ContractRuntimeRequest>,
+{
+    // TODO: We already had all these deploys from when we validated the
+    // ProtoBlock. We should just carry them around instead of doing this.
```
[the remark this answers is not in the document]
xcthulhu: Are there?  There's like only one unfinalized block at a time, right?  Why do we hold onto them after the proposer's round?

node/src/components/contract_runtime.rs
xcthulhu: Fixed in https://github.com/casper-network/casper-node/pull/1669/commits/302374993792ed12d36a47094ae61e4cef65eae1

If we recur here, we'll never emit any effects until the queue is drained, which might mean downstream components will wait longer before they can do their thing.

node/src/components/linear_chain_sync.rs
xcthulhu: Well, `linear_chain_sync` is going away so I didn't think having this duplicated code was a big deal.

node/src/effect.rs
xcthulhu: Fixed in 5f80f2f45340c67321a7adc3cc30dba8ed510b25

node/src/effect.rs
xcthulhu: Fixed in a9f3fcfce4f7b72bd1577e401c322cdf6c057e03

execution_engine_testing/test_support/src/internal/wasm_test_builder.rs
```diff
@@ -524,16 +516,8 @@ where
...
-        if let CommitResult::Success { state_root } = commit_result {
-            self.post_state_hash = Some(state_root);
-            return self;
-        }
-        panic!(
-            "Expected commit success but received a failure instead: {:?}",
-            commit_result
-        );
+        self.post_state_hash = Some(self.commit_transforms(prestate_hash, effects));
+        self
```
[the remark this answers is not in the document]
xcthulhu: `commit_transforms` is only used here, so I inlined it, and changed the name as suggested.

Fixed in 3c82b40bcb83a74d8cb140c2493f3b1ee7ef3eca

execution_engine/src/storage/global_state/mod.rs
```diff
@@ -161,13 +139,15 @@ where
...
+                return Err(CommitError::KeyNotFound(key).into());
             }
             (ReadResult::Found(current_value), transform) => match transform.apply(current_value) {
                 Ok(updated_value) => updated_value,
-                Err(err) => return Ok(err.into()),
+                Err(err) => return Err(CommitError::TransformError(err).into()),
             },
-            _x @ (ReadResult::RootNotFound, _) => panic!(stringify!(_x._1)),
+            (ReadResult::RootNotFound, _) => {
+                return Err(CommitError::ReadRootNotFound(state_root).into())
```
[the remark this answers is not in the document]
xcthulhu: If this happens we should shut down the node.  I have attempted to have all of the top-level call sites for `commit` shut down if it fails.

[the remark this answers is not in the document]
xcthulhu: I've added the transform to this context here in a9c1aea64df9aa0d01430c1aec28f6802744e9fa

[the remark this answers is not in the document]
xcthulhu: I went and added `error!` level logging for all of these error cases in 27e99a546

types/src/execution_result.rs
```diff
@@ -111,13 +111,17 @@ static EXECUTION_RESULT: Lazy<ExecutionResult> = Lazy::new(|| {
 });
 
 /// The result of executing a single deploy.
+// Note: For backward compatibility reasons, the `execution_effect` fields are serialized under the
+// name `effect`. We could consider introducing a json representation struct here as we have done
+// elsewhere for `Block`.
```
afck, line 116: Ah, I see! Sorry for complicating this!
Maybe then we should use `effect` everywhere instead of `execution_effect`?

execution_engine/src/storage/global_state/mod.rs
```diff
@@ -152,22 +134,44 @@ where
...
         let value = match (read_result, transform) {
             (ReadResult::NotFound, Transform::Write(new_value)) => new_value,
-            (ReadResult::NotFound, _) => {
-                return Ok(CommitResult::KeyNotFound(key));
+            (ReadResult::NotFound, transform) => {
+                error!(
+                    ?state_root,
+                    ?key,
+                    ?transform,
+                    "Key not found wile attempting to apply transform"
```
afck, line 150:
```suggestion
                    "Key not found while attempting to apply transform"
```

node/src/components/consensus/era_supervisor.rs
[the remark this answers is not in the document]
afck: I like the idea of a `ValidatedBlock` type — similar to what we do with vertices in Highway. It would also require a change to the Highway code, however: Currently the `ProposedBlock` is part of the Highway unit wire format, but in the protocol state we'd want to use `ValidatedBlock`. Anyway, that's probably something for a different PR.

I'd still like to move everything that uses `Deploy` or `DeployHash` out of `consensus`, however.

node/src/components/consensus/era_supervisor.rs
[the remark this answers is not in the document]
afck: Yes, it's unlikely, but it can happen. E.g. if lots of validators are offline, multiple subsequent blocks could be unfinalized at the same time. Or if there's an equivocator there could be several sibling blocks.

We need to hold on to the block for the rest of the era because it will be a dependency of all our future units. If we send someone a unit who is missing a dependency, we need to be able to provide it.

node/src/components/contract_runtime.rs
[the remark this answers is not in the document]
xcthulhu: Fixed in 3023749

node/src/components/contract_runtime.rs
```diff
@@ -404,10 +408,40 @@ where
...
+                .ignore()
+            }
+            ContractRuntimeRequest::EnqueueBlockForExecution {
                 finalized_block,
                 deploys,
             } => {
-                info!(?finalized_block, "execute finalized block");
+                info!(?finalized_block, "enqueuing finalized block for execution");
                 if self.execution_pre_state.next_block_height == finalized_block.height() {
                     self.execute_finalized_block(
```
afck, line 446: Should we replace this call with `operations::execute_finalized_block` as well now?

node/src/components/consensus/era_supervisor.rs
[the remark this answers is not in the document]
xcthulhu: Okay, this will be moved out of consensus and into `LinearChain` in an upcoming PR I am working on.

node/src/components/consensus/era_supervisor.rs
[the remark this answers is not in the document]
xcthulhu: Okay, we'll retrieve them from storage later.  I got rid fo this comment in 82de6d9b69d18ed1a4e0467fab6081d5e9089bc2

node/src/components/contract_runtime/operations.rs
xcthulhu: Fixed in ab987022cd28168aa6451e947e27c3914366b66c

node/src/components/contract_runtime/operations.rs
xcthulhu: I went ahead and evicted them in c4757fec6839f056de203cd5f9da8a8dffa1e29f

types/src/execution_result.rs
xcthulhu: Fixed in b122856458683d0de58deb3914bda944b9afd05c

node/src/components/contract_runtime.rs
xcthulhu: This call returns effects.  Ultimately, the plan is to move this functionality to `LinearChain`

dwerner (approved):

afck (commented):
I just realized that this will now execute each block as a whole, in a single, blocking step, _even in normal operation_ (i.e. not just while joining) — does it? Execution can take a long time, so if we keep it that way, it should probably run in a separate thread?

node/src/components/contract_runtime/operations.rs
xcthulhu: Yes.

afck (approved):

afck (approved):

dwerner (approved):

xcthulhu:
bors merge

PR #1671, opened 2021-07-08
Update dependencies and rust compiler version

### LINKED ISSUE #1667: Resolve tokio vulnerability

Fraser999:
This PR updates our dependencies (triggered by a RUSTSEC warning), which in turn required an update of the pinned Rust version, and from there, many new clippy warnings needed resolved.

Note that [`RUSTSEC-2021-0073` security warning](https://rustsec.org/advisories/RUSTSEC-2021-0073) is now ignored in the Makefile since that particular one only affects a dependency of libp2p which is effectively unused and scheduled for removal in the near future.

Closes #1667.


nilz3ro (approved):
LGTM!
I read all of it. Definitely cleans up a lot. I like that we're using `From<T>` instead of `Into<T>` everywhere because it seems more conventional to me. It seems like we're removing the `&` where we're already borrowing, which is nice and clean.

darthsiroftardis (approved):

Fraser999:
bors merge

Issue #1688, opened 2021-07-14
Using StoredVersionedContractByHash uses wrong context for execution


mpapierski:
On v1.2.0 when a contract uses `put_key` API, and the contract is invoked through `StoredVersionedContractByHash` or `StoredVersionedContractByName` a context of a contract package is used, but a contract hash is necessary.

This is currently solved on dev by commit https://github.com/casper-network/casper-node/commit/aca6a6f822722cd50d877c33aeeca4fac8f02e8a and this task is about creating a regression test to ensure we will not run into this again in the future.

PR #1689, opened 2021-07-14
GH-1688: Base key regression test

mpapierski:
Ref #1688 

This commit adds a regression test, which applied on top of `release-1.2.0` fails, but passes on `dev` as the issue is already fixed by commit aca6a6f822722cd50d877c33aeeca4fac8f02e8a

This test exercises all Stored* variant of `ExecutableDeployItem` to ensure that computed `base_key` always refers to either `Account` or a `Contract` but never a `ContractPackage`.


execution_engine_testing/tests/src/test/regression/gh_1688.rs
```diff
@@ -0,0 +1,225 @@
...
+        "expected {} in {:?}",
+        NEW_KEY_NAME,
+        contract.named_keys()
+    );
+    assert!(
+        !account.named_keys().contains_key(NEW_KEY_NAME),
+        "unexpected {} in {:?}",
+        NEW_KEY_NAME,
+        contract.named_keys()
+    );
```
goral09, line 224: These lines seem to be duplicated across all tests – is that correct? If so, can we extract it?

darthsiroftardis: I agree, the difference seems to be in how the deploy is created, we could extract it out and build a certain deploy based on a flag?

darthsiroftardis (approved):
LGTM!

execution_engine_testing/tests/src/test/regression/gh_1688.rs
mpapierski: Fair point - did this in 89bf83740fc6694e1fade4ec3563fe7103a6226b

EdHastingsCasperLabs (approved):

PR #1694, opened 2021-07-14
Remove `ErrWrapper` in favor of `matches!` in tests.

### LINKED ISSUE #1576: Remove ErrWrapper from client integration tests
This type exists to allow equality comparison of `Error`s.  We can avoid that by just using `matches!`.

nilz3ro:
# About These Changes
* Removes `ErrWrapper` in favor of `matches!`.
* closes #1576


dwerner:
ErrWrapper was implemented in order to facilitate the output of unit tests, while simply `assert!(matches!(...` results in a failure but inferior logging as to what error was recieved. Is that desirable?

nilz3ro:
@dwerner I think the trade off is that we're losing a little bit of the precision that the `ErrWrapper` provided, since it allowed us to use `assert_eq`. By removing it, we now have one less type to maintain and we're able to use something built into Rust. I think that overall this makes the tests cleaner and more maintainable. I am, however, curious to know if there's a way to make error assertions in a way that's simple and direct like `assert!(matches!(...))` but that gives us clear output like `assert_eq!`.

dwerner:
> @dwerner I think the trade off is that we're losing a little bit of the precision that the `ErrWrapper` provided, since it allowed us to use `assert_eq`. By removing it, we now have one less type to maintain and we're able to use something built into Rust. I think that overall this makes the tests cleaner and more maintainable. I am, however, curious to know if there's a way to make error assertions in a way that's simple and direct like `assert!(matches!(...))` but that gives us clear output like `assert_eq!`.

I think you'd get everything you're after and the ability to `assert_eq!` if you ensured that all the underlying error types first implement Eq and PartialEq.

goral09:
It looks like the `ErrWrapper`'s `PartialEq` relies on the string representation of the underlying error. That's not a good practice though.

client/tests/integration_test.rs
```diff
@@ -119,34 +119,24 @@ impl MockServerHandle {
         }
     }
 
-    fn get_balance(&self, state_root_hash: &str, purse_uref: &str) -> Result<(), ErrWrapper> {
-        casper_client::get_balance("1", &self.url(), 0, state_root_hash, purse_uref)
-            .map(|_| ())
-            .map_err(ErrWrapper)
+    fn get_balance(&self, state_root_hash: &str, purse_uref: &str) -> Result<(), Error> {
+        casper_client::get_balance("1", &self.url(), 0, state_root_hash, purse_uref).map(|_| ())
```
darthsiroftardis, line 123: I think you can get rid of this `.map` here, cause you don't need to create an `ErrWrapper`

nilz3ro: I left those in there because the functions returned `Result<(), Error>`, I could go back and change everything to `Result<T, Error>` depending on what client function is being called if that's better.

EdHastingsCasperLabs (approved):

Fraser999 (approved):

PR #1695, opened 2021-07-14
implement lmdb manual flushing for EE

dwerner:
This PR hardcodes flags for LMDB (in the EE only) to make flushing manual, which can improve the performance of block execution by:

image


node/src/components/contract_runtime.rs
```diff
@@ -472,6 +474,7 @@ where
                             let correlation_id = CorrelationId::new();
                             let start = Instant::now();
                             let result = engine_state.commit_step(correlation_id, step_request);
```
goral09, line 476: Why not put it inside the `commit*` functions? Those are the ones that "flush" changes to the disk.

execution_engine/src/storage/transaction_source/lmdb.rs
```diff
@@ -75,9 +75,14 @@ impl LmdbEnvironment {
         map_size: usize,
         max_readers: u32,
     ) -> Result<Self, error::Error> {
+        // These options mean we must manually call sync on the environment for the EE.
+        let fast_options = EnvironmentFlags::NO_SYNC
+            | EnvironmentFlags::NO_META_SYNC
+            | EnvironmentFlags::NO_LOCK
+            | EnvironmentFlags::WRITE_MAP;
```
xcthulhu, line 82: We might consider having these flags come from config settings...

node/src/components/contract_runtime.rs
xcthulhu: This may reflect the benchmarks Dan has currently implemented.  The last time I spoke he was actually instrumenting the loop that was extracted [here][1], which is slated for the 1.4.0 release. If he was flushing only after executing the whole block, this is the place in the code as-is to do so.

But it's straightforward to use benchmarks to figure out what to do here:

  - Put the flush after executing the entire block
  - Put the flush after every `commit`/`commit_step`

goral09: You're right, we should definitely benchmark the alternative.

dwerner: Here are the results of flushing every commit vs the baseline (current functionality):
image

goral09: Thanks for this @dwerner ! That's a great analysis. Case closed :)

dwerner:
Bumping to 1.4 for more testing.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -99,6 +103,16 @@ pub struct EngineState<S> {
     state: S,
 }
 
+impl EngineState<LmdbGlobalState> {
+    pub fn flush_environment(&self) {
+        self.state
+            .environment
+            .env()
+            .sync(true)
+            .expect("should flush environment")
```
xcthulhu: Let's just propagate the error here, rather than expect.

mpapierski:
I'm fine with this change - in my opinion, looking at the numbers, the benefits outweigh potential problems, although I'll be happy if we'd have more time to experiment and understand LMDB transactions and the potential impact of manual flushing on the consistency of global state in case of crashes. Here are also my thoughts:

I have a theory that we could achieve a similar result by putting one giant LMDB transaction that will execute multiple commits - my reasoning for this is that if you have 2500 operations each in a separate transaction in a relational database, it would be faster to have one transaction that does 2500 operations (operation == commit_upgrade). I didn't have a chance yet to measure this using LMDB, although the way we currently use LMDB transactions, and how we execute commit calls suggests it we could make it more efficient (and safer comparing to manual flush). I think the same for reads that we're doing with LMDB: each read creates its own transaction, which also calls commit() (which also forces a `fsync`?).

dwerner:
As per my conversation with @EdHastingsCasperLabs, I made this configurable behavior in `config.toml` under `[contract_runtime]`'s `enable_manual_sync` bool option.

execution_engine/src/core/engine_state/execute_request.rs
```diff
@@ -5,7 +5,7 @@ use casper_types::{ProtocolVersion, PublicKey, SecretKey};
 use super::deploy_item::DeployItem;
 use crate::shared::newtypes::Blake2bHash;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
goral09: Is it necessary to `Clone` this object? It can be expensive to have multiple copies of it and we really should only need one.

node/CHANGELOG.md
```diff
@@ -47,7 +50,8 @@ All notable changes to this project will be documented in this file.  The format
 * Enable getting validators for future eras in `EffectBuilder::get_era_validators()`.
 * Improve logging around stalled consensus detection.
 * Skip storage integrity checks if the node didn't previously crash.
-* Update pinned version of Rust to `nightly-2021-06-17`.
+* Update pinned version of Rust to `nightly-2021-06-17`
```
```suggestion
* Update pinned version of Rust to `nightly-2021-06-17`.
```

goral09 (commented):
Looks good but I would prefer if we didn't have to clone the `ExecRequest`. If that's possible.

xcthulhu (approved):

execution_engine/src/core/engine_state/execute_request.rs
dwerner: It's needed in benches, but isn't used in the node this way.

goral09: Got it. Thanks.

goral09 (approved):

execution_engine_testing/tests/benches/transfer_bench.rs
```diff
@@ -247,6 +250,67 @@ pub fn transfer_to_existing_accounts(group: &mut BenchmarkGroup<WallTime>, shoul
...
+        |b| b.iter(|| transfer_to_account_multiple_native_transfers(&mut builder, &exec_requests)),
+    );
+}
+
+fn transfer_to_account_multiple_native_transfers(
+    builder: &mut LmdbWasmTestBuilder,
+    execute_requests: &[ExecuteRequest],
+) {
+    for exec_request in execute_requests.iter().cloned() {
+        let builder = builder.exec(exec_request).expect_success();
```
Fraser999: This would allow us to avoid making the `ExecuteRequest` clonable:
```suggestion
    for exec_request in execute_requests {
        let request = ExecuteRequest::new(
            exec_request.parent_state_hash,
            exec_request.block_time,
            exec_request.deploys.clone(),
            exec_request.protocol_version,
            exec_request.proposer.clone(),
        );
        let builder = builder.exec(request).expect_success();
```

node/src/components/contract_runtime/config.rs
```diff
@@ -26,6 +26,11 @@ pub struct Config {
     ///
     /// Defaults to 5.
     max_query_depth: Option<u64>,
+
+    /// Enable synchronizing to disk only after each block is written.
+    ///
+    /// Defaults to true.
+    enable_manual_sync: bool,
```
Fraser999: This needs to be changed to an `Option<bool>` if we want to leave it commented out in the config.toml files.

Fraser999 (changes requested):
Even if the `enable_manual_sync` config field is changed to `Option<bool>`, I think we should also update the following config files:
* resources/production/config-example.toml
* utils/nctl/sh/scenarios/configs/bond_its.config.toml
* utils/nctl/sh/scenarios/configs/gov96.config.toml
* utils/nctl/sh/scenarios/configs/itst13.config.toml

node/src/components/contract_runtime/config.rs
dwerner: Done

node/src/components/contract_runtime.rs
```diff
@@ -392,6 +392,10 @@ where
                     let start = Instant::now();
                     let result = engine_state
                         .put_trie_and_find_missing_descendant_trie_keys(correlation_id, &*trie);
+                    // PERF: this *could* be called only periodically.
+                    if let Err(lmdb_error) = engine_state.flush_environment() {
+                        error!(?lmdb_error, "error flushing lmdb environment");
```
Fraser999: I think we're treating an error while flushing as fatal at the other callsite.  That would be appropriate here too wouldn't it?

dwerner: Nice catch. I'll do that now.

dwerner: Done

Fraser999 (approved):
LGTM

PR #1696, opened 2021-07-15
Check signature weight when loading block or header by height.

### LINKED ISSUE #1655: Never Return A Block/BlockHeader With Insufficient Finality Signatures To a Fetcher

afck:
Closes #1655.


node/src/components/linear_chain_sync/operations.rs
```diff
@@ -193,7 +193,18 @@ fn validate_finality_signatures(
...
     // Cryptographically verify block signatures
     block_signatures.verify()?;
+    weigh_finality_signatures(
+        trusted_validator_weights,
+        finality_threshold_fraction,
+        block_signatures,
+    )
+}
 
+pub(crate) fn weigh_finality_signatures(
```
goral09, line 203: Can you add an API doc comment to this method?

node/src/reactor/validator.rs
```diff
@@ -783,36 +782,51 @@ impl reactor::Reactor for Reactor {
...
+                                match Message::new_get_response(
+                                    &fetch_or_not_found_header_with_sigs,
+                                ) {
+                                    Ok(message) => {
+                                        effect_builder.send_message(sender, message).await
+                                    }
+                                    Err(error) => error!("failed to create get-response {}", error),
                                 }
-                            };
+                            }
```
goral09, line 828: This looks very similar to the method in `linear_chain_sync::operations`, is that what the `TODO` there is about?

afck: It looks like I'll probably have to rewrite this PR anyway after `feat-fast-sync` has been rebased and @xcthulhu's latest refactorings merged. Then I'll try to deduplicate this, too.

afck:
Closing this since we're rewriting it. We'll take your comments into account!

PR #1702, opened 2021-07-16
Add changelog entry for shutdown_on_standstill option.

afck:
Add a changelog entry for https://github.com/casper-network/casper-node/pull/1681.

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -36,6 +36,7 @@ All notable changes to this project will be documented in this file.  The format
 * Improve logging around stalled consensus detection.
 * Skip storage integrity checks if the node didn't previously crash.
 * Update pinned version of Rust to `nightly-2021-06-17`
+* Don't shut down by default anymore if stalled. To enable set config option `shutdown_on_standstill = true` in `[consensus.highway]`.
 
 ### Removed
 * Remove systemd notify support, including removal of `[network][systemd_support]` config option.


goral09 (approved):
I think you could add a couple more words around what _stalled_ means but it's not a blocker, more like an opinion.

fizyk20 (approved):

afck:
Yes, I'm not sure how much detail should go into the changelog. Maybe:
```
* Don't shut down anymore if consensus stops making progress.
  This used to be the default so the node would restart and possibly be better connected.
  To re-enable it, set the config option `shutdown_on_standstill = true` in `[consensus.highway]`.
```

goral09:
Personally, I would give enough information in the changelog so that a person not involved in the project can understand (more or less) what the change was about.

afck:
Just taked to @Fraser999 and he likes the shorter one, so I'd merge it for now.
There's also a little bit more detail in the `config-example.toml`. Maybe I should expand that instead?

afck:
bors merge

goral09:
> Just taked to @Fraser999 and he likes the shorter one, so I'd merge it for now.
> There's also a little bit more detail in the `config-example.toml`. Maybe I should expand that instead?

I think the target audience of the changelog shouldn't have to read the code to understand the change. The changelog isn't for devs, it's for managers, people using the project – validators making a decision whether they want to use the new binary, users determining whether they can use the new version of the client, etc.

afck:
Sure, but "stalled", "no progres in consensus", "no protocol state change" all would require a lot more explanation to really give any additional information; much more than what fits in a changelog, I think. Do you have a suggestion for the entry?

PR #1711, opened 2021-07-19
Storage Component No Longer Panics

xcthulhu:


node/src/components/storage.rs
```diff
@@ -684,7 +775,7 @@ impl Storage {
                         .insert(*block_hash, execution_result);
                     let was_written =
                         txn.put_value(self.deploy_metadata_db, &deploy_hash, &metadata, true)?;
-                    assert!(
+                    debug_assert!(
```
afck, line 778: This will now fail completely silently in production. We should probably at least log an error? (Also below.)
But even then: Is it okay to just continue if we failed to write to storage?

node/src/components/storage.rs
```diff
@@ -1035,50 +1130,44 @@ impl Storage {
...
         let block_header: BlockHeader = match tx.get_value(self.block_header_db, &block_hash)? {
             Some(block_header) => block_header,
             None => return Ok(None),
         };
         let found_block_header_hash = block_header.hash();
         if found_block_header_hash != *block_hash {
-            return Err(LmdbExtError::BlockHeaderNotStoredUnderItsHash {
-                queried_block_hash: *block_hash,
+            return Err(Error::BlockHeaderNotStoredUnderItsHash {
+                queried_block_hash_bytes: Digest::from(*block_hash).to_vec(),
```
afck, line 1170: Why the conversion? Isn't `found_block_header_hash` also a `BlockHash`?

xcthulhu: Elsewhere, this same error is used when reading values under a raw key in the block header DB using a cursor.  This is why it has this structure.

I've switched to `block_hash.as_ref().to_vec()` in https://github.com/casper-network/casper-node/pull/1711/commits/c3982d4ac73d6e54481bfd4691f52214b059b99c

node/src/components/storage.rs
xcthulhu: It depends on the failure. In general, if we can continue to function with some loss of functionality, we should.

In this case, the execution results have already been applied to the Merkle trie in the `ContractRuntime`. The system can continue to make blocks with the rest of the chain. The loss of functionality appears to be that certain queries by the JSON-RPC may fail.

xcthulhu: I have added logging in this PR: https://github.com/casper-network/casper-node/pull/1711/commits/ac69ed9fda3315ff4bcab237adc6eca0509008d2

node/src/components/storage.rs
```diff
@@ -173,6 +174,92 @@ pub enum Error {
     /// Error when validating a block.
     #[error(transparent)]
```
Fraser999, line 175: The doc comment for this `Error` enum should be changed (I can't attach this comment to the actual relevant line: it's line 135)

node/src/components/storage.rs
```diff
@@ -735,13 +824,13 @@ impl Storage {
                         return Ok(responder.respond(None).ignore());
                     };
                 // Check that the hash of the block retrieved is correct.
-                assert_eq!(&block_hash, block.hash());
+                debug_assert_eq!(&block_hash, block.hash());
```
Fraser999: I think we should log an error here, and at the new `debug_assert!` just below too.

node/src/components/storage.rs
```diff
@@ -1452,10 +1539,6 @@ impl Storage {
     }
 
     /// Reads all known deploy hashes from the internal store.
-    ///
```
Fraser999: Shouldn't this comment block remain unchanged?  It's still valid and contains useful info.  Or did you mean to update this method to return a `Result` too?

node/src/components/storage.rs
```diff
@@ -1474,29 +1557,18 @@ impl Storage {
             .collect()
     }
 
-    /// Get the switch block for a specified era number in a read-only LMDB database transaction.
-    ///
-    /// # Panics
-    ///
-    /// Panics on any IO or db corruption error.
-    pub fn transactional_get_switch_block_by_era_id(
+    /// Get the switch block for a specified era number.
```
```suggestion
    /// Gets the switch block for a specified era ID.
```

node/src/components/storage.rs
```diff
@@ -1474,29 +1557,18 @@ impl Storage {
...
+    pub fn read_switch_block_by_era_num(
```
Fraser999: Nitpick: we generally refer to "era IDs" rather than "era numbers"

node/src/components/storage.rs
```diff
@@ -1474,29 +1557,18 @@ impl Storage {
...
-            .commit()
-            .expect("Could not commit transaction");
-        switch_block
+    ) -> Result<Option<Block>, Error> {
+        let mut read_only_lmdb_transaction = self.env().begin_ro_txn()?;
+        let maybe_switch_block = self.get_switch_block_by_era_id(
+            &mut read_only_lmdb_transaction,
+            EraId::from(switch_block_era_num),
+        )?;
+        drop(read_only_lmdb_transaction);
```
Fraser999: I think this line could just be deleted.

Fraser999 (commented):
A few nitpicks, but generally LGTM.

node/src/components/storage.rs
xcthulhu: Fixed in 897417c96

node/src/components/storage.rs
xcthulhu: Fixed in d15a3e057

node/src/components/storage.rs
xcthulhu: My bad.  I've reverted this in ce02f1da6

node/src/components/storage.rs
xcthulhu: Fixed in f8326e350

node/src/components/storage.rs
xcthulhu: Fixed in e669360a7

node/src/components/storage.rs
xcthulhu: Fixed in f8326e3

node/src/components/storage.rs
```diff
@@ -735,13 +824,21 @@ impl Storage {
...
+                           "block not stored under hash");
+                    debug_assert_eq!(&block_hash, block.hash());
+                }
                 let finality_signatures =
                     match self.get_finality_signatures(&mut txn, &block_hash)? {
                         Some(signatures) => signatures,
                         None => BlockSignatures::new(block_hash, block.header().era_id()),
                     };
-                assert!(finality_signatures.verify().is_ok());
+                if finality_signatures.verify().is_ok() {
```
```suggestion
                if finality_signatures.verify().is_err() {
```

afck (approved):

Fraser999 (approved):

PR #1712, opened 2021-07-19
Execute blocks to current

xcthulhu:


node/src/types/block.rs
```diff
@@ -378,20 +378,14 @@ impl FinalizedBlock {
...
-            .iter()
-            .copied()
-            .map(DeployOrTransferHash::Deploy)
-            .chain(
-                self.transfer_hashes
-                    .iter()
-                    .copied()
-                    .map(DeployOrTransferHash::Transfer),
-            )
+    /// Returns the deploy hashes for the finalized block.
```
afck, line 381: Maybe to clarify, since "deploy" is a bit ambiguous:

```suggestion
    /// Returns the deploy hashes for the finalized block, excluding transfers.
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1300,8 +1296,30 @@ async fn execute_finalized_block<REv>(
             return;
         }
     }
+
+    // Get the deploy hashes for the finalized block.
+    let transaction_hashes = finalized_block.transfer_hashes().to_owned();
```
afck, line 1301:
```suggestion
    // Get the transfer hashes for the finalized block.
    let transfer_hashes = finalized_block.transfer_hashes().to_owned();
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1300,8 +1296,30 @@ async fn execute_finalized_block<REv>(
...
+
+    // Get all deploys in order they appear in the finalized block.
+    let mut transactions: Vec<Deploy> = Vec::with_capacity(transaction_hashes.len());
```
afck, line 1304:
```suggestion
    // Get all transfers in order they appear in the finalized block.
    let mut transfers: Vec<Deploy> = Vec::with_capacity(transfer_hashes.len());
```

node/src/components/contract_runtime/operations.rs
```diff
@@ -54,7 +55,7 @@ pub(super) fn execute_finalized_block(
         HashMap::new();
     // Run any deploys that must be executed
     let block_time = finalized_block.timestamp().millis();
-    for deploy in deploys {
+    for deploy in deploys.into_iter().chain(transfers.into_iter()) {
```
afck, line 58:
```suggestion
    for deploy in deploys.into_iter().chain(transfers) {
```

node/src/components/contract_runtime/types.rs
```diff
@@ -92,3 +92,16 @@ pub struct BlockAndExecutionEffects {
     /// An [`ExecutionEffect`] created if an era ended.
     pub maybe_step_execution_effect: Option<ExecutionEffect>,
 }
+
+impl BlockAndExecutionEffects {
+    /// Get the block.
```
afck, line 97:
```suggestion
    /// Gets the block.
```

node/src/components/linear_chain_sync.rs
```diff
@@ -1048,26 +1047,52 @@ async fn execute_block<REv>(
...
+                )
+                .await;
+                return;
+            }
         }
-    }
+        deploys
+    };
+
+    // Get the deploy hashes for the block.
```
afck, line 1071: (Also below.)

```suggestion
    // Get the transfer hashes for the block.
```

node/src/types/block.rs
xcthulhu: I've tried to clarify the documentation in https://github.com/casper-network/casper-node/pull/1712/commits/b1f535c93aa991626a75eb3a0750c8d4d8fdc99b

node/src/components/consensus/era_supervisor.rs
xcthulhu: Fixed in 8e89b9c98236e80661953dce42d8328fd93a994f

node/src/components/consensus/era_supervisor.rs
xcthulhu: Fixed in https://github.com/casper-network/casper-node/commit/8e89b9c98236e80661953dce42d8328fd93a994f

node/src/components/contract_runtime/operations.rs
xcthulhu: Fixed in 9a1366523

node/src/components/contract_runtime/types.rs
xcthulhu: Fixed in d09df6d36

node/src/components/linear_chain_sync.rs
xcthulhu: Fixed in 132941fc2

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1272,52 +1272,66 @@ where
...
 }
 
+async fn get_deploys_or_transfers<REv>(
+    effect_builder: EffectBuilder<REv>,
+    hashes: Vec<DeployHash>,
+) -> Option<Vec<Deploy>>
+where
+    REv: From<StorageRequest>,
+{
+    let mut deploys_or_transfer: Vec<Deploy> = Vec::with_capacity(hashes.len());
```
afck, line 1282:
```suggestion
    let mut deploys_or_transfers: Vec<Deploy> = Vec::with_capacity(hashes.len());
```

afck (approved):

goral09:
We had this separation into deploys and transfers at the beginning, months ago but we purposefully started handling them both together b/c that separation lead to bugs where on every call-site developers had to remember that there are _two_ items they need to handle and that was easily missed. Why are we separating them again? I cannot see anything in the code here that would require this? Both `deploys` and `transfers` are of type `Vec<Deploy>` (same goes for `deploy_hashes` and `transfer_hashes` which are `&[DeployHash]`).

This PR includes a lot of (unnecessary and in my opinion even  dangerous) refactoring. If not for it, I think the [important bit](https://github.com/casper-network/casper-node/pull/1712/commits/bac733eebe398404f94814fe87384c292ff45522#diff-be0c11c9a6fcff5e4229eae55fb52c24d0a59100e637a36eaeed02f1321394a9R601-R621) would be a 20 line PR. Currently, because of all the noise, the important part is easy to miss during review.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1273,35 +1272,68 @@ where
     }
 }
 
+async fn get_deploys_or_transfers<REv>(
```
goral09, line 1275: ~Isn't this really `get_deploys_AND_transfers`?~ Ah, this separation into transfers and deploys again.

node/src/components/event_stream_server/deploy_getter.rs
```diff
@@ -67,7 +66,7 @@ impl DeployGetter {
         #[cfg(test)]
         return self.get_test_deploy(deploy_hash);
 
-        let deploy_hashes = smallvec![deploy_hash];
+        let deploy_hashes = vec![deploy_hash];
```
goral09, line 69: What was wrong with `smallvec!`? It's an optimization where it won't make as many allocations as `vec!` for collections of less than 2 elements.

fizyk20 (approved):
Looks good, though like Mateusz, I'm wondering about the smallvec → vec change.

node/src/components/event_stream_server/deploy_getter.rs
xcthulhu: The effect builder was just calling `to_vec()` on it anyway.

PR #1715, opened 2021-07-19
fromPublicKey function

HansWestman74:
Added missing function fromPublicKey and imported necessary blake method to external.


mpapierski:
@HansWestman74 I think you need to change the base from release-1.2.1 into dev

PR #1720, opened 2021-07-19
Deduplicate Transfer Arg Parsing in Client.

### LINKED ISSUE #1578: Avoid duplicating parsing of transfer args in client
Move the parsing of transfer args from lib.rs to `DeployExt::new_transfer()`.

Also consider removing the `DeployExt` trait, as it probably doesn't give much benefit over free functions.

nilz3ro:
# About These Changes
* closes #1578


Fraser999 (approved):

Fraser999 (approved):

client/lib/deploy.rs
```diff
@@ -251,6 +252,11 @@ impl DeployExt for Deploy {
         const TRANSFER_ARG_TARGET: &str = "target";
         const TRANSFER_ARG_ID: &str = "id";
 
+        let amount = U512::from_dec_str(amount)
+            .map_err(|err| Error::FailedToParseUint("amount", UIntParseError::FromDecStr(err)))?;
```
rafal-ch: Could it be that we can use the `const TRANSFER_ARG_AMOUNT` here instead of hardcoding the `"amount"` string?

rafal-ch (approved):
LGTM!

I left a single minor comment.

Additionally, consider checking if the code that creates `session` in `fn new_transfer(...)` is (or should be) covered by tests. None of the tests in `deploy.rs` uses the mentioned function to create a `Deploy`, they use the `with_payment_and_session` and similar constructors instead.

client/lib/deploy.rs
nilz3ro: Yes that makes sense.

PR #1728, opened 2021-07-20
Block/BlockHeader requests with sufficient finality signatures

### LINKED ISSUE #1655: Never Return A Block/BlockHeader With Insufficient Finality Signatures To a Fetcher

afck:
Closes #1655

### DIFF
--- a/node/src/components/linear_chain.rs
+++ b/node/src/components/linear_chain.rs
@@ -16,15 +16,13 @@ use casper_types::ProtocolVersion;
 
 use crate::{
     components::{
-        fetcher::FetchedOrNotFound,
         linear_chain::state::{Outcome, Outcomes},
         Component,
     },
     effect::{
         announcements::LinearChainAnnouncement,
         requests::{
-            ChainspecLoaderRequest, ContractRuntimeRequest, LinearChainRequest, NetworkRequest,
-            StorageRequest,
+            ChainspecLoaderRequest, ContractRuntimeRequest, NetworkRequest, StorageRequest,
         },
         EffectBuilder, EffectExt, EffectResultExt, Effects,
     },
@@ -64,7 +62,7 @@ impl<I> LinearChainComponent<I> {
 fn outcomes_to_effects<REv, I>(
     effect_builder: EffectBuilder<REv>,
     outcomes: Outcomes,
-) -> Effects<Event<I>>
+) -> Effects<Event>
 where
     REv: From<StorageRequest>
         + From<NetworkRequest<I, Message>>
@@ -132,7 +130,7 @@ where
         + Send,
     I: Display + Send + 'static,
 {
-    type Event = Event<I>;
+    type Event = Event;
     type ConstructionError = Infallible;
 
     fn handle_event(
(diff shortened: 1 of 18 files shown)


node/src/components/linear_chain.rs
```diff
@@ -142,36 +140,6 @@ where
         event: Self::Event,
     ) -> Effects<Self::Event> {
         match event {
-            Event::Request(LinearChainRequest::BlockRequest(block_hash, sender)) => async move {
```
goral09, line 145: I will probably learn it later in the review process but why are we removing these? It should be the `LinearChain` component that answers these queries about the _linear chain_ blocks.

node/src/components/storage.rs
```diff
@@ -895,18 +904,57 @@ impl Storage {
...
+            .get_block_and_sufficient_finality_signatures_by_height(
+                &mut txn,
+                height,
+                genesis_validator_weights,
+                finality_threshold_fraction,
+                last_emergency_restart,
+            )?;
+        drop(txn);
+        Ok(maybe_block_and_finality_signatures)
+    }
```
goral09, line 1054: These belong to the `LinearChain` component in my opinion, and not storage. Storage shouldn't have _any_ domain logic, it should only handle retrieving and storing data.

node/src/components/storage.rs
```diff
@@ -1275,6 +1331,142 @@ impl Storage {
...
+                    ?err,
+                    ?block_header,
+                    "insufficient finality signatures for block header read from storage"
+                );
+                Ok(None)
+            }
+            Err(err) => Err(err.into()),
+            Ok(()) => Ok(Some(block_signatures)),
+        }
+    }
```
goral09, line 1468: Same here – domain logic should not be in storage layer.

node/src/types/chainspec/network_config.rs
```diff
@@ -27,13 +28,16 @@ pub struct NetworkConfig {
 
 impl NetworkConfig {
     /// Returns a vector of chainspec validators' public key and their stake.
-    pub fn chainspec_validator_stakes(&self) -> Vec<(PublicKey, Motes)> {
+    pub fn chainspec_validator_stakes(&self) -> BTreeMap<PublicKey, U512> {
```
goral09, line 31: What's the reason to change the return type here?

goral09 (commented):
I haven't reviewed the correctness of the new logic yet, just the code organization and I have some comments.

node/src/types/chainspec/network_config.rs
afck: We were only using this in places where we ended up doing this conversion.

node/src/components/storage.rs
xcthulhu: While it is true that the storage component now contains domain logic, involving the linear chain has significant downsides.

1. It does not access LMDB in a transactional manner. We are losing the A in ACID. The code in this PR is transactional.

2. The JSON-RPC accesses the Storage component directly rather than using the `LinearChain`. It does not require sufficient finality signatures. Doing so is not backward compatible. While the JSON-RPC _could_ go through the `LinearChain`, that would require creating several different message types and extending various handlers.

3. There are 6 different fetchers. While two of them were routed to the linear chain, the rest were not.  All of the fetcher requests are now handled uniformly [here][1].

4. Unit testing this code has been straightforward. Unit testing a reactor component is far more involved.

[1]: https://github.com/casper-network/casper-node/pull/1728/files#diff-212250b2151062263c01ec4511602e2ada0b8a83dd02b8727589737f3d3312ebR366-R427

afck:
Closing this, since it seems to be more controversial than I thought. Let's have a discussion first.

PR #1741, opened 2021-07-21
Move to stable rust for all but casper-contract and smart contracts

### LINKED ISSUE #1706: Move majority of Rust code to stable Rust
This will pave the way for allowing better organisation of the dependencies between our crates, with an emphasis on splitting data shared between the client and node into the `casper-types` crate.

Fraser999:
The main thrust of this PR is to allow the majority of the codebase to be compiled and tested using stable Rust.

The `casper-contract` crate and all smart contracts still require to be built using nightly, since they require nightly features relating to alloc error handlers due to the `no_std` environment.

The smart contracts have almost all had their `std` feature removed as it was at odds with the unconditional `#![no_std]` in each, and was never enabled.  The ones which have retained the `std` feature all provide lib targets which are consumed by other crates requiring this feature.

There was a single benchmark test in casper-node which has been commented out in order to allow the removal of the `#![feature(test)]`.  This test will be ticketed for re-inclusion using criterion to avoid the need for nightly Rust.

The Makefile has been updated to accommodate these changes and the total number of tests being executed by `make check-rs` has remained effectively consistent (we currently run one extraneous test in casper-types when `std` is not explicitly enabled - this is rectified in this PR).  We currently omit building many of the crates' doc targets, and this too has been rectified.

The drone file has been updated to make more use of the Makefile, ensuring less disparity between what devs run and what CI runs, and reducing duplication (e.g. of `--ignore RUSTSEC-xxx`).  CI will also now build docs for all published crates and fail on any warnings.

Finally, regarding performance, the following table provides a comparison between current `dev` and the code in this PR:

| | dev | This PR |
| --- | --- | --- |
| Time to run `make clean && make check-rs` in Debug mode | 14m36.643s | 13m6.501s |
| Time to rerun `make check-rs` in Debug mode | 6m53.304s | 6m43.968s |
| Disk usage after Debug run | 18,476,280 bytes | 11,480,920 bytes |
| Time to run `make clean && make check-rs` in Release mode | 25m26.351s | 28m22.404s |
| Time to rerun `make check-rs` in Release mode | 6m3.701s | 3m37.964s |
| Disk usage after Release run | 8,517,968 bytes | 7,559,520 bytes |

Closes #1706.


execution_engine/src/shared/newtypes/blake2b256.rs
```diff
@@ -32,7 +32,7 @@ impl Blake2bHash {
     }
 
     /// Converts the underlying BLAKE2b hash digest array to a `Vec`
-    pub fn to_vec(&self) -> Vec<u8> {
```
mpapierski, line 35: Wouldn't that become `into_vec`?

node/src/crypto/hash.rs
```diff
@@ -56,12 +56,12 @@ impl Digest {
     pub const LENGTH: usize = 32;
 
     /// Returns a copy of the wrapped `u8` array.
-    pub fn to_array(&self) -> [u8; Digest::LENGTH] {
+    pub fn to_array(self) -> [u8; Digest::LENGTH] {
```
mpapierski, line 59: `into_`, no?

smart_contracts/contract/src/no_std_handlers.rs
```diff
@@ -1,25 +1,20 @@
...
 
 /// A panic handler for use in a `no_std` environment which simply aborts the process.
 #[panic_handler]
 #[no_mangle]
-pub fn panic(_info: &::core::panic::PanicInfo) -> ! {
+pub fn panic(_info: &core::panic::PanicInfo) -> ! {
     #[cfg(feature = "test-support")]
-    runtime::print(&format!("Panic: {}", _info));
-    ::core::intrinsics::abort();
+    crate::contract_api::runtime::print(&alloc::format!("Panic: {}", _info));
```
mpapierski: As we discussed offline - with an improved Makefile `build-contracts-rs` we have to remove this line as we've observed negative performance hit on transaction benchmarks by 20% on average

mpapierski (approved):
👍

nilz3ro (approved):

Fraser999:
bors merge

PR #1751, opened 2021-07-21
Wrap deploy-accepted event stream event

### LINKED ISSUE #1743: Wrap deploy accepted event in SSE in DeployAccepted
We should change the deploy accepted event emitted by the SSE server from the contents of a deploy to one wrapped in a newtype named `DeployAccepted` so this stream is consistent with the others, which will simplify things for consumers of the streams.

Fraser999:
This PR changes how an accepted deploy is represented in JSON on the event stream.  Previously, only the contents of the deploy were output.  This now wraps the contents in a `DeployAccepted` type, making it consistent with the style of other SSEs.

Closes #1743.

### DIFF
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -205,6 +205,12 @@ impl SseData {
     }
 }
 
+#[derive(Serialize)]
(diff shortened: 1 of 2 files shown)


darthsiroftardis (approved):

Fraser999:
bors merge

PR #1753, opened 2021-07-21
removed return unreachables added fromPublicKey and more

HansWestman74:
This pull request 

- removes incorrect code that causes errors when upgrading to the latest AssemblyScript compiler. Specifically, AssemblyScript does not have a never return type like Rust and TypeScript. For this reason the latest asc compiler gets confused with code like

       return <T>unreachable().
       
- adds missing function  fromPublicKey needed for the AssemblyScript implementation of the keys-manager.
- declares and makes use of casper-blake2b in externals.ts

**bytesrepr.ts**
let bytes = bytes.subarray(currentPos); --> bytes = bytes.subarray(currentPos);
return <T>unreachable(); ---> unreachable();

**index.ts**
replaced all  return <T>unreachable() with unreachable()
getNamedArgSize(name: String) ---> getNamedArgSize(name: string)
Replaced all U32 with u32 in getNamedArgSize

**key.ts**
replaced all  return <T>unreachable() ---> unreachable()
fixed ending of Key.toBytes so that the compiler is happy with the return type
Implemented   fromPublicKey(publicKey:PublicKey) : AccountHash  in AccountHash class.

**public_key.ts**
Implemented getRawBytes and getAlgorithmName in PublicKey class

**purse.ts** 
replaced all  return <T>unreachable() ---> unreachable()

**externals.ts**
added declaration of casper_blake2b function used in implemented fromPublicKey function

### DIFF
--- a/smart_contracts/contract_as/assembly/key.ts
+++ b/smart_contracts/contract_as/assembly/key.ts
@@ -4,7 +4,7 @@ import {UREF_SERIALIZED_LENGTH} from "./constants";
 import {URef} from "./uref";
 import {CLValue} from "./clvalue";
 import {Error, ErrorCode} from "./error";
-import {checkTypedArrayEqual, typedToArray,encodeUTF8} from "./utils";
+import {checkTypedArrayEqual, typedToArray, encodeUTF8} from "./utils";
 import {Ref} from "./ref";
 import {Result, Error as BytesreprError} from "./bytesrepr";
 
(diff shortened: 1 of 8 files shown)


smart_contracts/contract_as/assembly/key.ts
```diff
@@ -4,10 +4,12 @@ import {UREF_SERIALIZED_LENGTH} from "./constants";
 import {URef} from "./uref";
 import {CLValue} from "./clvalue";
 import {Error, ErrorCode} from "./error";
-import {checkTypedArrayEqual, typedToArray} from "./utils";
+import {checkTypedArrayEqual, typedToArray,encodeUTF8} from "./utils";
```
mpapierski:
```suggestion
import {checkTypedArrayEqual, typedToArray, encodeUTF8} from "./utils";
```

mpapierski (approved):
I'd suggest consistent formatting in the rest of new code but other than that 👍

Fraser999 (approved):
LGTM

HansWestman74:
bors merge

PR #1757, opened 2021-07-22
GOV-35: Do not synchronize the linear chain immediately after an upgrade.

goral09:
This PR adds a logic that detects whether the current node instance is running immediately after an upgrade and if that's the case – won't perform any linear chain synchronization.

### DIFF
--- a/node/src/components/chainspec_loader.rs
+++ b/node/src/components/chainspec_loader.rs
@@ -305,7 +305,7 @@ impl ChainspecLoader {
     }
 
     /// Returns whether the current node instance is started immediately after an upgrade –
-    /// i.e. whether the last/highest block stored is a block that trigerred the upgrade.
+    /// i.e. whether the last/highest block stored is a block that triggerred the upgrade.
     pub(crate) fn after_upgrade(&self) -> bool {
         self.after_upgrade
     }
--- a/node/src/components/chainspec_loader.rs
+++ b/node/src/components/chainspec_loader.rs
@@ -176,6 +176,7 @@ pub struct ChainspecLoader {
     initial_state_root_hash: Digest,
     next_upgrade: Option<NextUpgrade>,
     initial_block: Option<Block>,
+    after_upgrade: bool,
 }
 
 impl ChainspecLoader {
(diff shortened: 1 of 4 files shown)


node/src/components/chainspec_loader.rs
```diff
@@ -301,6 +304,12 @@ impl ChainspecLoader {
         self.reactor_exit
     }
 
+    /// Returns whether the current node instance is started immediately after an upgrade –
+    /// i.e. whether the last/highest block stored is a block that trigerred the upgrade.
```
afck:
```suggestion
    /// i.e. whether the last/highest block stored is a block that triggered the upgrade.
```

node/src/types/chainspec/network_config.rs
```diff
@@ -25,23 +21,6 @@ pub struct NetworkConfig {
     pub(crate) accounts_config: AccountsConfig,
 }
 
-impl NetworkConfig {
-    /// Returns a vector of chainspec validators' public key and their stake.
-    pub fn chainspec_validator_stakes(&self) -> Vec<(PublicKey, Motes)> {
```
afck, line 30: This will actually be used by fast sync again, so we might as well keep it to avoid more merge conflicts.

afck (approved):

node/src/types/chainspec/network_config.rs
goral09: I would say that it's not the reason to keep an unused code around but I know this will be the cause of a lot of tension so I will bring it back.

node/src/components/chainspec_loader.rs
goral09: Fixed in [`b1f152a` (#1757)](https://github.com/casper-network/casper-node/pull/1757/commits/b1f152a9a6db463faf34ae4b3804b004a8a9f017)

node/src/types/chainspec/network_config.rs
goral09: Should be brought back in [`4307d72` (#1757)](https://github.com/casper-network/casper-node/pull/1757/commits/4307d726c27c8a225fe43d21d577450912a9e229)

EdHastingsCasperAssociation (approved):

PR #1759, opened 2021-07-22
Parameterize Sync Timeout

nilz3ro:
This PR introduces `sync_timeout` as a config option that allows an operator to control the duration of time that a node will sync before shutting down.

### DIFF
--- a/node/src/components/linear_chain_sync.rs
+++ b/node/src/components/linear_chain_sync.rs
@@ -23,13 +23,14 @@
 //! execute (as we do in the first, SynchronizeTrustedHash, phase) it would have taken more time and
 //! we might miss more eras.
 
+mod config;
 mod event;
 mod metrics;
 mod peers;
 mod state;
 mod traits;
 
-use std::{convert::Infallible, fmt::Display, mem, str::FromStr};
+use std::{convert::Infallible, fmt::Display, mem};
 
 use datasize::DataSize;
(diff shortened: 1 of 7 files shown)


node/src/components/linear_chain_sync.rs
```diff
@@ -94,15 +96,14 @@ impl<I: Clone + PartialEq + 'static> LinearChainSync<I> {
         highest_block: Option<Block>,
         after_upgrade: bool,
         next_upgrade_activation_point: Option<ActivationPoint>,
+        config: Config,
```
goral09, line 99: I think we should only pass in the timeout part, not the whole object.

resources/production/config-example.toml
```diff
@@ -312,3 +312,7 @@ verify_accounts = true
 #
 # If unset, defaults to 5.
 #max_query_depth = 5
+
+[linear_chain_sync]
+# Use sync_timeout to set the amount of time that the node will try to sync before shutting down.
+sync_timeout = "1min"
```
goral09: IMO the default should be at least the 5min it used to be.

Fraser999 (approved):

resources/production/config-example.toml
nilz3ro: done

sacherjj (approved):

PR #1762, opened 2021-07-23
Fix issues with pings and doppelgänger detection.

afck:
Closes https://github.com/casper-network/Governance/issues/36.

### DIFF
--- a/node/src/components/consensus/highway_core/highway.rs
+++ b/node/src/components/consensus/highway_core/highway.rs
@@ -997,4 +997,27 @@ pub(crate) mod tests {
         // Verify that sending a Ping from a non-existing validator does not panic.
         assert!(!highway.has_vertex(&ping));
     }
+
+    #[test]
+    fn own_initial_ping_is_not_from_doppelganger() {
+        let now: Timestamp = 500.into();
+        let later = 501.into();
+
+        let state: State<TestContext> = State::new_test(WEIGHTS, 0);
+        let target_ftt = state.total_weight() / 3;
+        let mut highway = Highway {
+            instance_id: TEST_INSTANCE_ID,
+            validators: test_validators(),
+            state,
+            active_validator: None,
+        };
+
+        let _effects =
+            highway.activate_validator(ALICE.0, ALICE_SEC.clone(), now, None, target_ftt);
+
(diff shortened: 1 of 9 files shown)


node/src/components/consensus/highway_core/state.rs
```diff
@@ -568,13 +568,11 @@ impl<C: Context> State<C> {
         self.pings[creator] = self.pings[creator].max(timestamp);
     }
 
-    /// Returns `true` if the latest timestamp we have is less than one maximum round length older
-    /// than the given timestamp.
-    ///
-    /// This is to prevent ping spam: If the incoming ping is only slightly newer than a unit or
-    /// ping we have already received, we drop it without forwarding it to our peers.
+    /// Returns `true` if the latest timestamp we have is older than the given timestamp.
```
xcthulhu, line 571: I'm trying to understand.

If the ping was not created by us, it seems the old behavior makes sense.

And if the ping was from us, perhaps we could prevent ourselves from forwarding it elsewhere?

xcthulhu: Looks like we aren't forwarding, so this is the correct behavior.

xcthulhu (approved):
LGTM

node/src/components/consensus/highway_core/highway.rs
```diff
@@ -219,7 +219,7 @@ impl<C: Context> Highway<C> {
             self.instance_id,
         );
         self.active_validator = Some(av);
-        effects
+        self.add_new_own_vertices(effects, current_time)
```
goral09, line 222: That's the root of the problem? We were not adding our own ping to the state?

goral09 (approved):

node/src/components/consensus/highway_core/highway.rs
afck: Yes, exactly.

And the other potential problem was that we weren't counting pings from someone else as doppelgänger pings if they weren't _much_ younger (more than one maximum round length) than the latest one we had in the state.

goral09: Thanks 👍

goral09 (approved):

afck:
bors merge

PR #1772, opened 2021-07-23
Revert to use fixed delay before including a deploy in a proposed block

### LINKED ISSUE #1771: Revert usage of DeployGossiper finished gossiping announcement by BlockProposer
The deploy gossiper can announce that it has finished gossiping for a given deploy before it has received the body.  This causes the block proposer to fail to include the deploy in the next block, as it fails to retrieve the deploy from storage.  There is no re-attempt made by the block proposer, hence the deploy is never included in a block proposed by that node.

To fix this for now, the previous approach will be used whereby the block proposer will listen to  deploy acceptor announcements (which implies the deploy has definitely been stored locally) and will then include the deploy after a fixed delay, configurable, defaulted to 1 minute.

### LINKED ISSUE #1778: Missed Wasm transfer in Stests are pending or not executed
Scenario 3: Rate Limits with Load directed at Validators
Casper Node: e9b28da7e
Network : 5V + 5NV + 5RO
Instance: t3a.large
Dashboard:
https://grafana.casperlabs.io/goto/Z0flowink
https://grafana.casperlabs.io/goto/C4UuTwmnz
Config Override:
{'config': {'max_outgoing_byte_rate_non_validators': 6553600, 'max_incoming_message_rate_non_validators': 3000, 'enable_mem_deduplication': 'true'}, 'chainspec': {}}
Logs: http://genesis.casperlabs.io/ajith-kaboom-130/casper-node-logs/16Jul21-5V5NV-all-limits-validator-nodes-e9b28da7e/16072021_0415/log_download_list.txt
Observations:
- network stable
- load targeted to ONLY validator nodes with stest
- 30k transfers completed
- :x:  3.6k wasm transfers not fully executed. ~200 were pending/not-executed
- 5 joiner nodes successfully synced
- 1 out of 5 reloaded twice to sync to height
- :x: SSE reports shows empty wasm blocks with workload executions (edited)

Fraser999:
This PR fixes an edge-case issue where the `BlockProposer` could try to include a deploy in a proposed block before it has stored that deploy locally.  The deploy would not be included in that proposed block, and that node would never try to include that deploy in a proposed block again.

The fix is a temporary change back to the previous approach for this mechanism: the `BlockProposer` only tries to include deploys for which there has been a `DeployAcceptor` announcement (implying the deploy has definitely been stored locally).  Such deploys are only proposed after a configurable fixed period of time has passed.

For reference, this PR reverts most of the changes in #1484.

Closes #1771.
Closes #1778.

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -32,7 +32,6 @@ All notable changes to this project will be documented in this file.  The format
 * Persist event stream event index across node restarts.
 * Separate transfers from other deploys in the block proposer.
 * Enable getting validators for future eras in `EffectBuilder::get_era_validators()`.
-* Replace config option `[block_propser][deploy_delay]` (which specified a fixed delay before proposing a deploy) with a gossip-finished announcement.
 * Improve logging around stalled consensus detection.
 * Skip storage integrity checks if the node didn't previously crash.
 * Update pinned version of Rust to `nightly-2021-06-17`
--- a/node/src/components/block_proposer.rs
+++ b/node/src/components/block_proposer.rs
@@ -3,6 +3,7 @@
 //! The block proposer stores deploy hashes in memory, tracking their suitability for inclusion into
 //! a new block. Upon request, it returns a list of candidates that can be included.
 
+mod config;
 mod deploy_sets;
 mod event;
 mod metrics;
@@ -16,10 +17,10 @@ use std::{
     time::Duration,
 };
 
+pub use config::Config;
 use datasize::DataSize;
 use itertools::Itertools;
 use prometheus::{self, Registry};
-use smallvec::smallvec;
 use tracing::{debug, error, info, trace, warn};
 
 use casper_types::PublicKey;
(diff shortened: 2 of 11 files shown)


goral09:
The PR we're trying to revert had also changes to `gossip_table.rs` and `gossiper.rs` - we don't need to revert those?

Fraser999:
> The PR we're trying to revert had also changes to gossip_table.rs and gossiper.rs - we don't need to revert those?

I don't believe so.  Those changes were all additional functionality in support of the new `FinishedGossiping` announcement.

goral09 (approved):
In the PR that introduced this feature there were some changes around metrics and I'm not sure if it's OK if we don't revert those changes as well. They're not related to the problems we're seeing so it's not a blocker. LGTM.

node/CHANGELOG.md
```diff
@@ -155,4 +154,4 @@ All notable changes to this project will be documented in this file.  The format
 [1.1.1]: https://github.com/casper-network/casper-node/compare/v1.0.1...v1.1.1
 [1.1.0]: https://github.com/casper-network/casper-node/compare/v1.0.1...v1.1.1
 [1.0.1]: https://github.com/casper-network/casper-node/compare/v1.0.0...v1.0.1
-[1.0.0]: https://github.com/casper-network/casper-node/releases/tag/v1.0.0
\ No newline at end of file
+[1.0.0]: https://github.com/casper-network/casper-node/releases/tag/v1.0.0
```
afck, line 157: Did anything actually change in this line??

afck (approved):

nilz3ro (approved):
LGTM 🎊

node/CHANGELOG.md
Fraser999: Just a newline which my editor automatically applies at the end of all files on saving.

Fraser999:
bors merge

PR #1790, opened 2021-07-27
Check for emergency restart before fast syncing

### LINKED ISSUE #1770: Fast Sync Must Not Try To Join Before An Emergency Restart

xcthulhu:
Before fast syncing, check that we are not trying to join before an emergency restart.

Closes https://github.com/casper-network/casper-node/issues/1770

### DIFF
--- a/node/src/components/linear_chain_sync/operations.rs
+++ b/node/src/components/linear_chain_sync/operations.rs
@@ -438,12 +438,28 @@ where
     I: Eq + Debug + Clone + Send + 'static,
 {
     // Fetch the trusted header
-    let trusted_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
+    let trusted_block_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
+
+    let maybe_last_emergency_restart_era_id = chainspec.protocol_config.last_emergency_restart;
+    match maybe_last_emergency_restart_era_id {
+        Some(last_emergency_restart_era)
+            if last_emergency_restart_era >= trusted_block_header.era_id() =>
+        {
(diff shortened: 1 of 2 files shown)


node/src/components/linear_chain_sync/operations.rs
```diff
@@ -438,12 +438,28 @@ where
     I: Eq + Debug + Clone + Send + 'static,
 {
     // Fetch the trusted header
-    let trusted_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
+    let trusted_block_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
+
+    let maybe_last_emergency_restart_era_id = chainspec.protocol_config.last_emergency_restart;
+    match maybe_last_emergency_restart_era_id {
+        Some(last_emergency_restart_era)
+            if last_emergency_restart_era >= trusted_block_header.era_id() =>
```
afck, line 446: Once we have the immediate switch block, will this become a strict inequality? Because then joining in the era immediately following the restart should be supported by fast sync, should it?

afck (approved):

node/src/components/linear_chain_sync/operations.rs
xcthulhu: So if we provided the immediate switch block's hash as the trusted hash, wouldn't we get equality here?  Worthy of a unit test...

nilz3ro (approved):

node/src/components/linear_chain_sync/operations.rs
afck: You're right, `>=` makes sense. :+1:

PR #1792, opened 2021-07-27
Revert to use fixed delay before including a deploy in a proposed block

### LINKED ISSUE #1771: Revert usage of DeployGossiper finished gossiping announcement by BlockProposer

### LINKED ISSUE #1778: Missed Wasm transfer in Stests are pending or not executed

Fraser999:
This PR is a port of #1772 and fixes an edge-case issue where the `BlockProposer` could try to include a deploy in a proposed block before it has stored that deploy locally.  The deploy would not be included in that proposed block, and that node would never try to include that deploy in a proposed block again.

The fix is a temporary change back to the previous approach for this mechanism: the `BlockProposer` only tries to include deploys for which there has been a `DeployAcceptor` announcement (implying the deploy has definitely been stored locally).  Such deploys are only proposed after a configurable fixed period of time has passed.

For reference, this PR reverts most of the changes in #1484.

Closes #1771.
Closes #1778.

### DIFF
--- a/node/src/reactor/participating.rs
+++ b/node/src/reactor/participating.rs
@@ -444,6 +444,7 @@ impl reactor::Reactor for Reactor {
                 .map(|block_header| block_header.height() + 1)
                 .unwrap_or(0),
             chainspec_loader.chainspec().as_ref(),
+            config.block_proposer,
         )?;
 
         let initial_era = maybe_latest_block_header.as_ref().map_or_else(
@@ -914,12 +915,30 @@ impl reactor::Reactor for Reactor {
                 deploy,
                 source,
             }) => {
+                let deploy_info = match deploy.deploy_info() {
+                    Ok(deploy_info) => deploy_info,
+                    Err(error) => {
+                        error!(%error, "invalid deploy");
+                        return Effects::new();
+                    }
+                };
+
+                let event = block_proposer::Event::BufferDeploy {
+                    hash: deploy.deploy_or_transfer_hash(),
+                    deploy_info: Box::new(deploy_info),
+                };
+                let mut effects =
+                    self.dispatch_event(effect_builder, rng, Event::BlockProposer(event));
+
                 let event = gossiper::Event::ItemReceived {
                     item_id: *deploy.id(),
                     source: source.clone(),
                 };
-                let mut effects =
-                    self.dispatch_event(effect_builder, rng, Event::DeployGossiper(event));
+                effects.extend(self.dispatch_event(
+                    effect_builder,
+                    rng,
+                    Event::DeployGossiper(event),
+                ));
 
                 let event = event_stream_server::Event::DeployAccepted(*deploy.id());
                 effects.extend(self.dispatch_event(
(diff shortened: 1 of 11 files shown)


nilz3ro (approved):

node/src/reactor/participating.rs
```diff
@@ -1020,11 +1039,13 @@ impl reactor::Reactor for Reactor {
                 Effects::new()
             }
             Event::DeployGossiperAnnouncement(GossiperAnnouncement::FinishedGossiping(
-                gossiped_deploy_id,
+                _gossiped_deploy_id,
             )) => {
-                let reactor_event =
-                    Event::BlockProposer(block_proposer::Event::BufferDeploy(gossiped_deploy_id));
-                self.dispatch_event(effect_builder, rng, reactor_event)
+                // let reactor_event =
```
marc-casperlabs, line 1044: Any reason we need to keep this commented out code around?

node/src/reactor/participating/config.rs
```diff
@@ -38,4 +39,7 @@ pub struct Config {
     pub deploy_acceptor: DeployAcceptorConfig,
     /// Linear chain sync configuration.
     pub linear_chain_sync: LinearChainSyncConfig,
+    /// Block proposer configuration.
+    #[serde(default)]
```
marc-casperlabs, line 43: Is this to make this configuration optional? I can see why that is useful, but on the other hand, wouldn't that be slightly out of policy regarding the rest of our code?

marc-casperlabs (approved):
No blockers, just comments.

node/src/reactor/participating.rs
Fraser999: Not really I suppose, but we'll be re-instating the commented code very soon.  Either way, I'd like to keep this changeset identical to the one it's ported from unless you have strong objections (which it seems from subsequent comments you don't :) )

node/src/reactor/participating/config.rs
Fraser999: Yes, I noticed that too.  This was done to keep it as close to the original code in release-1.2 as possible, to help simplify the original PR's review.  Again, this should be getting removed in the very near future, so I wasn't too concerned.

Fraser999:
bors merge

PR #1793, opened 2021-07-27
Split casper-types account.rs into multiple files.

### LINKED ISSUE #1787: Split `casper-types/src/account.rs` into multiple files
We want accounts.rs split as follows in preparation for moving more types from EE shared:

* account/account_hash.rs
* account/weight.rs
* account/action_type.rs
* account/error.rs (holder for all the various error types)

We should also remove `AccountHashBytes` altogether (we're aiming to not have aliases for our many byte array types, but only newtypes proper).

nilz3ro:
# About These Changes

## Edits to `types/src/account.rs`: 
* extract new  `mod account_hash`.
* extract new `mod action_type`
* extract new `mod error`.
* extract new `mod weight`.

## Other Edits:
* remove `AccountHashBytes`


types/src/account/account_hash.rs
```diff
@@ -0,0 +1,207 @@
...
+    fmt::{Debug, Display, Formatter},
+};
+use datasize::DataSize;
+use rand::{
+    distributions::{Distribution, Standard},
+    Rng,
+};
+#[cfg(feature = "std")]
+use schemars::{gen::SchemaGenerator, schema::Schema, JsonSchema};
+use serde::{de::Error as SerdeError, Deserialize, Deserializer, Serialize, Serializer};
```
Fraser999: We aim to group our imports in separate sections with a blank line between each as follows:
* std
* third party crates
* other casper crates
* this crate

This comment applies to all files in this PR.

types/src/account/account_hash.rs
```diff
@@ -0,0 +1,207 @@
...
+
+/// The length in bytes of a [`AccountHash`].
+pub const ACCOUNT_HASH_LENGTH: usize = 32;
+/// TODO: Document this.
```
Fraser999:
```suggestion
/// The prefix applied to the hex-encoded `AccountHash` to produce a formatted string representation.
```

Fraser999 (commented):
Nitpicks only.

rafal-ch (approved):
LGTM 👍

Issue #1765, opened 2021-07-23
Docs update: execution_engine/shared


mpapierski:
Add or update existing rust docs in a `shared` rust module in EE:

- [x] shared/account
- [x] shared/logging
- [x] shared/newtypes
- [x] shared/system_config
- [x] shared/*

mpapierski:
#2110 merged

PR #1810, opened 2021-07-29
Move EE Shared Account Types Into `casper_types`.

### LINKED ISSUE #1788: Move account types from EE to casper_types
The following types should be moved:
* `Account`
* `ActionThresholds`
* `AssociatedKeys`

nilz3ro:
# About These Changes
> This PR depends on #1793.
* moved `Account`, `ActionThresholds` and `AssociatedKeys` types and implementations to `casper_types::account` module.
* closes #1788.


Fraser999 (changes requested):
The instances of `/// TODO: doc comment.` should be addressed.  Please liaise with @mpapierski on this.

rafal-ch (approved):
LGTM 👍

mpapierski:
> 
> 
> The instances of `/// TODO: doc comment.` should be addressed. Please liaise with @mpapierski on this.

@nilz3ro @Fraser999 Those items you moved out from a `shared` module into `casper_types` were ticketed for documenting as part of #1765. It should be possible for you to start document these in this PR, and as far as I'm concerned there will be no collision there (I'm documenting EE's core, and Dan started EE's storage IIRC). I'm sure you'll be able to move far, and we can sync up tomorrow and I can help you with remaining items.

mpapierski (approved):

mpapierski:
I think you should move last bits from https://github.com/nilz3ro/casper-node/blob/GH-1788/execution_engine/src/shared/account.rs and move it into types crate

execution_engine/src/shared/transform.rs
```diff
@@ -1,11 +1,12 @@
 use std::{
     any,
-    convert::TryFrom,
     default::Default,
     fmt::{self, Display, Formatter},
     ops::{Add, AddAssign},
 };
 
+use core::convert::TryFrom;
+
```
Fraser999: Was this change necessary?

types/src/account.rs
```diff
@@ -16,11 +23,264 @@ use thiserror::Error;
...
+        Account {
+            account_hash,
+            named_keys,
+            main_purse,
+            associated_keys,
+            action_thresholds,
+        }
+    }
+
+    /// An Account constructor with presets for associated_keys and action_thresholds.
```
Fraser999, line 61: I think we should describe what the presets are - not necessarily in this first paragraph, but as a further paragraph.

types/src/account.rs
```diff
@@ -16,11 +23,264 @@ use thiserror::Error;
...
+    pub fn account_hash(&self) -> AccountHash {
+        self.account_hash
+    }
+
+    /// Returns main purse.
+    pub fn main_purse(&self) -> URef {
+        self.main_purse
+    }
+
+    /// Returns an [`AccessRights::ADD`]-only version of the [`URef`].
```
Fraser999:
```suggestion
    /// Returns an [`AccessRights::ADD`]-only version of the main purse's [`URef`].
```

types/src/account.rs
```diff
@@ -16,11 +23,264 @@ use thiserror::Error;
...
+
+        // Returns true if the new weight would be greater or equal to all of
+        // the thresholds.
+        new_weight >= self.action_thresholds().deployment().value()
+            && new_weight >= self.action_thresholds().key_management().value()
+    }
+
+    /// Removes an associated key from an account.
+    ///
+    /// Verifies that removing the key will not cause the remaining weight to fall below any action
```
Fraser999, line 158: Maybe rephrase to something like "Returns an error if removing the key will cause the remaining weight ..."

I think an example of how thresholds work would be useful, particularly wrt to changing them.  Probably something in the top-level of this module.  The we could link to that from here and from the docs for `update_associated_key` and any others as relevant.

types/src/account.rs
```diff
@@ -16,11 +23,264 @@ use thiserror::Error;
...
+                return Err(RemoveKeyFailure::ThresholdViolation);
+            }
+        }
+        self.associated_keys.remove_key(&account_hash)
+    }
+
+    /// Updates an associated key.
+    ///
+    /// This method validates that the key that's being updated does not have a lower weight than
+    /// the key management threshold.
```
Fraser999: I'd rephrase to something like "Returns an error if the update would result in a violation of the key management thresholds."

Similar point for `set_action_threshold`.

types/src/account.rs
```diff
@@ -16,11 +23,264 @@ use thiserror::Error;
...
+        weight: Weight,
+    ) -> Result<(), SetThresholdFailure> {
+        // Verify if new threshold weight exceeds total weight of all associated
+        // keys.
+        self.can_set_threshold(weight)?;
+        // Set new weight for given action
+        self.action_thresholds.set_threshold(action_type, weight)
+    }
+
+    /// Verifies if user can set action threshold
```
Fraser999:
```suggestion
    /// Verifies if user can set action threshold.
```

types/src/account.rs
```diff
@@ -16,11 +23,264 @@ use thiserror::Error;
...
+    pub fn can_set_threshold(&self, new_threshold: Weight) -> Result<(), SetThresholdFailure> {
+        let total_weight = self.associated_keys.total_keys_weight();
+        if new_threshold > total_weight {
+            return Err(SetThresholdFailure::InsufficientTotalWeight);
+        }
+        Ok(())
+    }
+
+    /// Checks whether all authorization keys are associated with this account
```
Fraser999:
```suggestion
    /// Checks whether all authorization keys are associated with this account.
```

Fraser999 (commented):
Nitpicks only now.

execution_engine/src/shared/transform.rs
nilz3ro: No, I will undo this change.

types/src/account.rs
nilz3ro: I think a practical example with doctests would be great. Do you think the examples should be done as part of this PR or should we make a new issue to add some practical examples with doc tests?

Fraser999 (approved):

types/src/account.rs
nilz3ro: Note for posterity: added #1905

PR #1814, opened 2021-07-29
Remove protocol data

darthsiroftardis:
CHANGELOG:

- Removed `ProtocolData` and by extension, the protocol_data store 
- Moved `SystemConfig` and `WasmConfig` over to `EngineConfig`
- System contract hashes are now stored in Global state under a unique `Key::SystemContractRegistry` in the form of a map
- Added a `release_1_3_1` test fixture to check upgrade from a prexisting global state to a new binary that does not include `ProtocolData`
- Added a test to check whether upgrades are committed successfully

Closes: [1685](https://app.zenhub.com/workspaces/engineering-60953fafb1945f0011a3592d/issues/casper-network/casper-node/1685)


goral09:
Task/PR description should include presentation of the problem and explanation why the ticket/PR is needed, what was done and why. In the case of this particular ticket - why are we removing the `ProtocolData`?

dwerner:
> Task/PR description should include presentation of the problem and explanation why the ticket/PR is needed, what was done and why. In the case of this particular ticket - why are we removing the `ProtocolData`?

It's required for fast-sync.

goral09:
> > Task/PR description should include presentation of the problem and explanation why the ticket/PR is needed, what was done and why. In the case of this particular ticket - why are we removing the `ProtocolData`?
> 
> It's required for fast-sync.

I figured that much but it still doesn't explain why. Is it b/c fast sync may not have access to the protocol files and can only synchronize the Global State? If so, then that explanation should be included in the ticket. Reviewers are not only judging the correctness of the code but also whether it's solving a real problem.

execution_engine/src/core/engine_state/executable_deploy_item.rs
```diff
@@ -320,7 +320,12 @@ impl ExecutableDeployItem {
                 error::Error::Exec(execution::Error::NoSuchMethod(entry_point_name.to_owned()))
             })?;
 
-        if protocol_data.system_contracts().contains(&contract_hash) {
+        let mut system_hashes = tracking_copy
+            .borrow_mut()
+            .get_system_contracts(correlation_id)?
+            .into_values();
+
+        if system_hashes.any(|x| x == contract_hash) {
```
mpapierski: Better to use find than any as find would stop on first `true` condition.

```suggestion
        if system_hashes.find(|&value| value == contract_hash) {
```

execution_engine/src/core/engine_state/executable_deploy_item.rs
```diff
@@ -128,7 +128,7 @@ impl ExecutableDeployItem {
         correlation_id: CorrelationId,
         preprocessor: &Preprocessor,
         protocol_version: &ProtocolVersion,
-        protocol_data: &ProtocolData,
```
mpapierski, line 131: I think you don't need to pass standard payment hash only - passing a system contract registry would be equivalent, as inside of `get_deploy_metadata` you're retrieving all the system contracts again. It is retrieved already before `get_deploy_metadata` is called so you can save on global state access by passing whole registry.

execution_engine_testing/test_support/src/internal/exec_with_return.rs
```diff
@@ -12,16 +12,15 @@ use casper_execution_engine::{
...
 };
 use casper_types::{
     account::AccountHash, bytesrepr::FromBytes, system::CallStackElement, BlockTime, CLTyped,
     DeployHash, EntryPointType, Key, Phase, ProtocolVersion, RuntimeArgs, URef, U512,
 };
 
 use crate::internal::{utils, WasmTestBuilder, DEFAULT_WASM_CONFIG};
-
-use super::DEFAULT_SYSTEM_CONFIG;
+use casper_types::system::STANDARD_PAYMENT;
```
mpapierski: Merge with line 17

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -1240,6 +1265,33 @@ where
...
+            .tracking_copy
+            .borrow_mut()
+            .read(self.correlation_id, &Key::SystemContractRegistry)
+            .map_err(|_| GenesisError::FailedToCreateRegistry)?
+            .expect("should have stored value")
+            .as_cl_value()
+            .expect("should be able to convert to CLValue")
+            .to_owned();
+        let mut partial_registry =
+            CLValue::into_t::<BTreeMap<String, ContractHash>>(partial_cl_registry)
```
mpapierski: My general bias would be to add a

```rust
pub type SystemContractRegistry = BTreeMap<String, ContractHash>;
```

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -799,7 +797,19 @@ where
 
         let (_, mint_hash) = self.store_contract(access_key, named_keys, entry_points);
 
-        self.protocol_data = ProtocolData::partial_with_mint(mint_hash);
+        {
+            let mut partial_registry = BTreeMap::<String, ContractHash>::new();
```
mpapierski, line 811:
```suggestion
            let mut partial_registry = BTreeMap::new();
```

(or see my other comment about a type alias)

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -826,6 +836,9 @@ where
 
         let (_, handle_payment_hash) = self.store_contract(access_key, named_keys, entry_points);
 
+        self.store_system_contract(HANDLE_PAYMENT, handle_payment_hash)
+            .expect("should install handle payment");
+
```
mpapierski: To be consistent with other `store_system_contract` calls:

```suggestion
        self.store_system_contract(HANDLE_PAYMENT, handle_payment_hash)?;
```

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -158,42 +152,25 @@ where
...
-            let protocol_data = ProtocolData::new(
-                *wasm_config,
-                *system_config,
-                mint_hash,
-                handle_payment_hash,
-                standard_payment_hash,
-                auction_hash,
-            );
-
-            self.state
```
mpapierski, line 193: I have a suggestion - the code as written now would write partial `BTreeMap` with system contracts. You could keep a `BTreeMap<String, ContractHash>` in the `GenesisInstaller` (just like before there was `self.protocol_data`), and extend it. And write it once to global state at this point (as it was done before)

execution_engine/src/core/engine_state/engine_config.rs
```diff
@@ -1,22 +1,48 @@
-const DEFAULT_MAX_QUERY_DEPTH: u64 = 5;
+use crate::shared::{system_config::SystemConfig, wasm_config::WasmConfig};
+
+pub const DEFAULT_MAX_QUERY_DEPTH: u64 = 5;
+
+pub const DEFAULT_WASMLESS_TRANSFER_COST: u32 = 10_000;
```
mpapierski: Do we need this to be defined here though? I think I've originally put it in protocol_data.rs - what about moving it to where it is used which would be: `execution_engine/src/shared/system_config.rs`

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1265,7 +1235,7 @@ where
                 correlation_id,
                 &preprocessor,
                 &protocol_version,
-                &protocol_data,
+                *standard_payment_hash,
```
mpapierski: See related comment in get_deploy_metadata

```suggestion
                &system_contract_registry,
```

execution_engine_testing/test_support/src/internal/wasm_test_builder.rs
```diff
@@ -367,22 +361,41 @@ where
...
+        let registry = match self.query(
+            Some(post_state_hash),
+            Key::SystemContractRegistry,
+            &empty_path,
+        ) {
+            Ok(StoredValue::CLValue(cl_registry)) => {
+                match CLValue::into_t::<BTreeMap<String, ContractHash>>(cl_registry) {
+                    Ok(registry) => registry,
+                    Err(error) => panic!("{:?}", error),
+                }
```
mpapierski:
```suggestion
                CLValue::into_t::<BTreeMap<String, ContractHash>>(cl_registry).unwrap();
```

execution_engine_testing/tests/src/test/regression/ee_966.rs
```diff
@@ -272,10 +273,11 @@ fn should_run_ee_966_regression_when_growing_mem_after_upgrade() {
         .with_current_protocol_version(*DEFAULT_PROTOCOL_VERSION)
         .with_new_protocol_version(*NEW_PROTOCOL_VERSION)
         .with_activation_point(DEFAULT_ACTIVATION_POINT)
-        .with_new_wasm_config(*DOUBLED_WASM_MEMORY_LIMIT)
         .build();
 
-    builder.upgrade_with_upgrade_request(&mut upgrade_request);
+    let engine_config = EngineConfig::new(5, *DOUBLED_WASM_MEMORY_LIMIT, SystemConfig::default());
```
mpapierski:
```suggestion
    let engine_config = EngineConfig::new(DEFAULT_MAX_QUERY_DEPTH, *DOUBLED_WASM_MEMORY_LIMIT, SystemConfig::default());
```

execution_engine_testing/tests/src/test/regression/gh_1470.rs
```diff
@@ -614,44 +639,38 @@ fn gh_1470_call_contract_should_verify_wrong_optional_argument_types() {
...
-        .upgrade_with_upgrade_request(&mut upgrade_request)
+        .upgrade_with_upgrade_request(*builder.get_engine_state().config(), &mut upgrade_request)
         .expect_upgrade_success();
 
-    let new_protocol_data = builder
-        .get_engine_state()
-        .get_protocol_data(new_protocol_version)
-        .expect("should have result")
-        .expect("should have protocol data");
+    //assert_eq!(new_protocol_data.mint(), old_protocol_data.mint());
```
mpapierski:
```suggestion
```

execution_engine_testing/tests/src/test/regression/gh_1470.rs
```diff
@@ -771,41 +770,33 @@ fn should_add_bid_after_major_bump() {
...
         .expect_upgrade_success();
 
-    let new_protocol_data = builder
-        .get_engine_state()
-        .get_protocol_data(new_protocol_version)
-        .expect("should have result")
-        .expect("should have protocol data");
-
-    assert_eq!(new_protocol_data.mint(), old_protocol_data.mint());
+    //assert_eq!(new_protocol_data.mint(), old_protocol_data.mint());
```
mpapierski:
```suggestion
```

execution_engine_testing/tests/src/test/regression/gh_1470.rs
```diff
@@ -830,38 +821,30 @@ fn should_add_bid_after_minor_bump() {
...
         .expect_upgrade_success();
 
-    let new_protocol_data = builder
-        .get_engine_state()
-        .get_protocol_data(new_protocol_version)
-        .expect("should have result")
-        .expect("should have protocol data");
-
-    assert_eq!(new_protocol_data.mint(), old_protocol_data.mint());
+    // assert_eq!(new_protocol_data.mint(), old_protocol_data.mint());
```
mpapierski:
```suggestion
```

execution_engine_testing/tests/src/test/regression/gh_1470.rs
```diff
@@ -885,41 +868,33 @@ fn should_wasm_transfer_after_major_bump() {
...
         .expect_upgrade_success();
 
-    let new_protocol_data = builder
-        .get_engine_state()
-        .get_protocol_data(new_protocol_version)
-        .expect("should have result")
-        .expect("should have protocol data");
-
-    assert_eq!(new_protocol_data.mint(), old_protocol_data.mint());
+    //assert_eq!(new_protocol_data.mint(), old_protocol_data.mint());
```
mpapierski:
```suggestion
```

execution_engine_testing/tests/src/test/storage_costs.rs
```diff
@@ -154,10 +75,10 @@ fn initialize_isolated_storage_costs() -> InMemoryWasmTestBuilder {
         .with_current_protocol_version(*DEFAULT_PROTOCOL_VERSION)
         .with_new_protocol_version(*NEW_PROTOCOL_VERSION)
         .with_activation_point(DEFAULT_ACTIVATION_POINT)
-        .with_new_wasm_config(*STORAGE_COSTS_ONLY)
```
mpapierski, line 157: You need to use `STORAGE_COSTS_ONLY` in the upgrade request below rather than carry on existing config

execution_engine_testing/tests/src/test/system_contracts/upgrade.rs
```diff
@@ -117,23 +118,20 @@ fn should_allow_only_wasm_costs_patch_version() {
             .with_current_protocol_version(PROTOCOL_VERSION)
             .with_new_protocol_version(new_protocol_version)
             .with_activation_point(DEFAULT_ACTIVATION_POINT)
-            .with_new_wasm_config(new_wasm_config)
             .build()
     };
 
+    let engine_config = EngineConfig::new(5, new_wasm_config, SystemConfig::default());
```
mpapierski:
```suggestion
    let engine_config = EngineConfig::new(DEFAULT_MAX_QUERY_DEPTH, new_wasm_config, SystemConfig::default());
```

execution_engine_testing/tests/src/test/system_costs.rs
```diff
@@ -190,6 +192,8 @@ fn upgraded_add_bid_and_withdraw_bid_have_expected_costs() {
         new_standard_payment_costs,
     );
 
+    let new_engine_config = EngineConfig::new(5, WasmConfig::default(), new_system_config);
```
mpapierski:
```suggestion
    let new_engine_config = EngineConfig::new(DEFAULT_MAX_QUERY_DEPTH, WasmConfig::default(), new_system_config);
```

execution_engine_testing/tests/src/test/system_costs.rs
```diff
@@ -421,6 +424,8 @@ fn upgraded_delegate_and_undelegate_have_expected_costs() {
...
```
mpapierski:
```suggestion
    let new_engine_config = EngineConfig::new(DEFAULT_MAX_QUERY_DEPTH, WasmConfig::default(), new_system_config);
```

execution_engine_testing/tests/src/test/system_costs.rs
```diff
@@ -853,17 +857,17 @@ fn should_verify_wasm_add_bid_wasm_cost_is_not_recursive() {
         new_standard_payment_costs,
     );
 
+    let new_engine_config = EngineConfig::new(5, new_wasm_config, new_system_config);
```
mpapierski:
```suggestion
    let new_engine_config = EngineConfig::new(DEFAULT_MAX_QUERY_DEPTH, WasmConfig::default(), new_system_config);
```

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -117,15 +119,8 @@ where
...
-    pub fn get_protocol_data(
-        &self,
-        protocol_version: ProtocolVersion,
-    ) -> Result<Option<ProtocolData>, Error> {
-        match self.state.get_protocol_data(protocol_version) {
-            Ok(Some(protocol_data)) => Ok(Some(protocol_data)),
-            Err(error) => Err(Error::Exec(error.into())),
-            _ => Ok(None),
-        }
+    pub fn update_config(&mut self, new_config: EngineConfig) {
```
mpapierski, line 122: ~I think we should put `EngineConfig` (or `WasmConfig` and `SystemConfig`) as part of `UpgradeConfig` - this way we'd enforce code change in node's contract runtime component - currently, the node wouldn't pass new `EngineConfig` as part of the upgrade.~ EDIT: @darthsiroftardis clarified it for me, and we don't need to put it there

I also think by doing that we should be able to remove `new_validator_slots` `new_auction_delay` `new_locked_funds_period_millis` `new_round_seigniorage_rate` `new_unbonding_delay`, and just take those values from appropriate configs at upgrade point.

@Fraser999 @xcthulhu Should we still update the system contract's named keys for purposes of fast sync as part of upgrade? Would it be safe to wire up system contracts to access the EngineConfig properties directly?

mpapierski (commented):
Bunch of clean-up nits, we could also improve the genesis process by writing to `Key::ContractRegistry` only once at the end of genesis, but other than that, I think it looks good and definitely should unblock fast sync work.

~You should also wire it up to contract runtime's upgrade~, and I'm a bit concerned about the amount of expects/unwraps, although looking at those, they should be safe.

execution_engine/src/core/engine_state/executable_deploy_item.rs
fizyk20: `any()` does, too: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any

> `any()` is short-circuiting; in other words, it will stop processing as soon as it finds a `true`, given that no matter what else happens, the result will also be `true`.

execution_engine/src/core/engine_state/mod.rs
darthsiroftardis: I tried your suggestion @mpapierski however, I run into an issue, earlier the genesis installer would handover it's instance of protocol data to setup a new runtime context for creating purses, however it then `create_purse` through a series of contract calls, invokes the host mint function which will then do a lookup for the contracts  (specifically handle_payment) in global state, and it will error out if some default values haven't already been written into global state

execution_engine/src/core/engine_state/executable_deploy_item.rs
mpapierski: Right, it's the same thing, basically. Let's leave it with `any`

darthsiroftardis: I applied the suggestion, however, `find() -> Option<&ContractHash>` and tailing it with an `.is_some()` is returning a lint warning, the compiler suggests `any` in favor of find

mpapierski: TIL. Thanks, @darthsiroftardis

execution_engine_testing/tests/src/test/system_costs.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/system_costs.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/system_costs.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/system_contracts/upgrade.rs
darthsiroftardis: Applied suggestion

execution_engine/src/core/engine_state/executable_deploy_item.rs
darthsiroftardis: I now pass the whole registry itself

execution_engine_testing/test_support/src/internal/exec_with_return.rs
darthsiroftardis: Moved

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: Applied type alias

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: Applies suggestion

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: Applied suggestion

execution_engine/src/core/engine_state/engine_config.rs
darthsiroftardis: Moved to `system_config`

execution_engine/src/core/engine_state/mod.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/regression/ee_966.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/test_support/src/internal/wasm_test_builder.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/regression/gh_1470.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/regression/gh_1470.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/regression/gh_1470.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/storage_costs.rs
darthsiroftardis: The test now uses `STORAGE_COSTS_ONLY`

execution_engine/src/core/execution/executor.rs
```diff
@@ -22,11 +22,12 @@ use crate::{
...
         runtime::{extract_access_rights_from_keys, instance_and_memory, Runtime},
         runtime_context::{self, RuntimeContext},
-        tracking_copy::TrackingCopy,
+        tracking_copy::{TrackingCopy, TrackingCopyExt},
     },
     shared::{account::Account, gas::Gas, newtypes::CorrelationId, stored_value::StoredValue},
-    storage::{global_state::StateReader, protocol_data::ProtocolData},
+    storage::global_state::StateReader,
 };
+use casper_types::system::{AUCTION, HANDLE_PAYMENT, MINT};
```
Fraser999: This should be grouped with the other casper_types imports.

types/src/key.rs
```diff
@@ -52,6 +53,7 @@ pub const KEY_DICTIONARY_LENGTH: usize = 32;
 /// The maximum length for a `dictionary_item_key`.
 pub const DICTIONARY_ITEM_KEY_MAX_LENGTH: usize = 64;
 
+const SYSTEM_CONTRACT_REGISTRY: [u8; 32] = [0u8; 32];
```
Fraser999:
```suggestion
const SYSTEM_CONTRACT_REGISTRY_KEY: [u8; 32] = [0u8; 32];
```

execution_engine/src/core/execution/executor.rs
```diff
@@ -393,11 +390,26 @@ impl Executor {
...
     {
+        let system_contract_registry = match tracking_copy
+            .borrow_mut()
+            .get_system_contracts(correlation_id)
+        {
+            Ok(registry) => registry,
+            Err(error) => {
+                panic!("Could not retrieve system contracts: {:?}", error)
+            }
+        };
```
Fraser999: You could use `unwrap_or_else(|error| panic!(...))` here I think

execution_engine/src/core/engine_state/executable_deploy_item.rs
```diff
@@ -320,7 +322,9 @@ impl ExecutableDeployItem {
                 error::Error::Exec(execution::Error::NoSuchMethod(entry_point_name.to_owned()))
             })?;
 
-        if protocol_data.system_contracts().contains(&contract_hash) {
+        let mut system_hashes = system_contract_registry.into_values();
+
+        if system_hashes.any(|value| value == contract_hash) {
```
Fraser999: Maybe a little more canonical to not extract the iterator to a separate variable:
```suggestion
        if system_contract_registry
            .values()
            .any(|&value| value == contract_hash)
        {
```

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -51,11 +51,13 @@ use crate::{
...
         wasm_config::WasmConfig,
     },
-    storage::{global_state::StateProvider, protocol_data::ProtocolData},
+    storage::global_state::StateProvider,
 };
 
 pub const PLACEHOLDER_KEY: Key = Key::Hash([0u8; 32]);
 const TAG_LENGTH: usize = U8_SERIALIZED_LENGTH;
+const DEFAULT_ADDRESS: [u8; 32] = [0; 32];
+pub type SystemContractRegistry = BTreeMap<String, ContractHash>;
```
Fraser999, line 61: I'm not a fan of this alias.  Was there a discussion around using a newtype wrapper round a `BTreeMap`, or better still, just a struct with the 4 required members rather than a map?

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -799,7 +798,19 @@ where
 
         let (_, mint_hash) = self.store_contract(access_key, named_keys, entry_points);
 
-        self.protocol_data = ProtocolData::partial_with_mint(mint_hash);
+        {
+            let mut partial_registry = BTreeMap::<String, ContractHash>::new();
+            partial_registry.insert(MINT.to_string(), mint_hash);
+            partial_registry.insert(HANDLE_PAYMENT.to_string(), DEFAULT_ADDRESS.into());
+            partial_registry.insert(AUCTION.to_string(), DEFAULT_ADDRESS.into());
+            partial_registry.insert(STANDARD_PAYMENT.to_string(), DEFAULT_ADDRESS.into());
```
Fraser999: Why do we need to do this?  Looks like we could just avoid these three lines?

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -1135,7 +1152,16 @@ where
...
 
-        let base_key = Key::Hash(self.protocol_data.mint().value());
+        let registry = self
+            .tracking_copy
+            .borrow_mut()
+            .get_system_contracts(self.correlation_id)
+            .map_err(execution::Error::from)
+            .map_err(GenesisError::ExecutionError)?;
+
+        let mint_hash = registry.get(MINT).expect("should have mint contract hash");
```
Fraser999: Seems like we should return an error rather than panicking.

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -1240,6 +1265,33 @@ where
...
+        &self,
+        contract_name: &str,
+        contract_hash: ContractHash,
+    ) -> Result<(), GenesisError> {
+        let partial_cl_registry = self
+            .tracking_copy
+            .borrow_mut()
+            .read(self.correlation_id, &Key::SystemContractRegistry)
+            .map_err(|_| GenesisError::FailedToCreateRegistry)?
+            .expect("should have stored value")
```
Fraser999: Again, seems like we shouldn't be panicking here and just below.

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -267,6 +267,9 @@ where
                 self.named_keys.remove(name);
                 Ok(())
             }
+            Key::SystemContractRegistry => {
+                unreachable!("should not remove this key once written")
```
Fraser999: Should we really be panicking here?
```suggestion
                error!("should not remove the system contract registry key");
                Err(Error::RemoveKeyFailure(RemoveKeyFailure::PermissionDenied))
```

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -758,8 +764,16 @@ where
...
+                .borrow_mut()
+                .get_system_contracts(self.correlation_id)
+                .expect("should have registry");
+
+            if system_contracts_registry
+                .into_values()
+                .any(|system_hash| system_hash == hash.into())
+            {
                 return true;
             }
```
Fraser999: If you rename `fn get_system_contract_registry()` as suggested in a subsequent comment, you can change this to:
```suggestion
            let contract_hash = ContractHash::new(hash);
            return self
                .system_contract_registry()
                .values()
                .any(|&system_hash| system_hash == contract_hash);
```

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -1118,4 +1132,32 @@ where
...
+
+    pub fn system_contracts(&self) -> Vec<ContractHash> {
+        self.tracking_copy
+            .borrow_mut()
+            .get_system_contracts(self.correlation_id)
+            .expect("should have system contract registry")
+            .values()
+            .cloned()
+            .collect()
+    }
```
Fraser999: I don't believe we need this - there's only one call site and it could just as well use `get_system_contract_registry`.

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -1118,4 +1132,32 @@ where
...
+    pub fn get_system_contract(&self, name: &str) -> ContractHash {
+        *self
+            .tracking_copy
+            .borrow_mut()
+            .get_system_contracts(self.correlation_id)
+            .map_err(Error::from)
+            .expect("should have system contracts registry")
+            .get(name)
+            .expect("should have system contract registry")
+    }
```
Fraser999: This seems particularly fragile to me.  Let's remove the ability to pass arbitrary strings through here and have 4 separate getters, one per system contract.  Each should be `pub(super)` ideally or `pub(crate)` if really required.  They should be named like `mint_contract_hash(&self) -> ContractHash`.

I'd be inclined to add those as methods on the tracking copy itself and just delegate here, but if that causes problems, leave it for now.

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -1118,4 +1132,32 @@ where
...
+
+    pub fn get_system_contract_registry(&self) -> BTreeMap<String, ContractHash> {
```
Fraser999:
```suggestion
    pub fn system_contract_registry(&self) -> BTreeMap<String, ContractHash> {
```

Fraser999 (changes requested):
A partial review, paused as there are upcoming changes as discussed offline.  Blocked pending these changes.

execution_engine/src/core/engine_state/genesis.rs
Fraser999: Ignore this comment - covered in offline discussion.

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: We need to insert the `HANDLE_PAYMENT` hash into the registry as, `create_handle_payment`, calls the `charge_host_mint` function which needs a default value for the `handle_payment` contract hash. we used to this earlier by calling `protocol_data::partial_with_mint`

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: We now return an error like:
```
let mint_hash = registry
            .get(MINT)
            .ok_or_else(|| GenesisError::MissingSystemContractHash(AUCTION.to_string()))?;
            
```

darthsiroftardis: We now return an error like:
 ```
 let mint_hash = registry
            .get(MINT)
            .ok_or_else(|| GenesisError::MissingSystemContractHash(MINT.to_string()))?;
```

File: execution_engine/src/core/runtime_context/mod.rs (hunk 34 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/runtime_context/mod.rs]: Applied suggestion

File: execution_engine/src/core/runtime_context/mod.rs (hunk 32 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/runtime_context/mod.rs]: Removed function

File: execution_engine/src/core/runtime_context/mod.rs (hunk 31 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/runtime_context/mod.rs]: Applied suggestion

File: execution_engine/src/core/runtime_context/mod.rs (hunk 30 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/runtime_context/mod.rs]: Applied suggestion

File: execution_engine/src/core/engine_state/genesis.rs (hunk 29 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/engine_state/genesis.rs]: Returns an error

File: execution_engine/src/core/engine_state/executable_deploy_item.rs (hunk 25 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/engine_state/executable_deploy_item.rs]: Applied suggestion

File: execution_engine/src/core/execution/executor.rs (hunk 24 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/execution/executor.rs]: Used the suggestion

File: types/src/key.rs (hunk 23 above)
Comment (darthsiroftardis) [reply types/src/key.rs]: Applied suggestion

File: execution_engine/src/core/execution/executor.rs (hunk 22 above)
Comment (darthsiroftardis) [reply execution_engine/src/core/execution/executor.rs]: Moved

File: execution_engine/src/core/engine_state/executable_deploy_item.rs [hunk 35]
```diff
@@ -22,24 +22,25 @@ use casper_types::{
...
         tracking_copy::{TrackingCopy, TrackingCopyExt},
     },
     shared::{
         account::Account, newtypes::CorrelationId, stored_value::StoredValue, wasm, wasm_prep,
         wasm_prep::Preprocessor,
     },
-    storage::{global_state::StateReader, protocol_data::ProtocolData},
+    storage::global_state::StateReader,
 };
+use casper_types::system::HANDLE_PAYMENT;
```
Fraser999: Needs moved.

execution_engine/src/core/engine_state/genesis.rs
Fraser999: So let's document the `partial_registry.insert(HANDLE_PAYMENT.to_string(), DEFAULT_ADDRESS.into());` call, since that's highly unintuitive.  Please also document the `DEFAULT_ADDRESS` const too, and it seems like it might be better named `TEMP_ADDRESS` or something like that?

As for the other two `insert()` calls - if they're not required here, let's remove them.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -253,50 +221,72 @@ where
             return Err(Error::InvalidProtocolVersion(new_protocol_version));
         }
 
+        let registry = {
+            // Check if the registry exists in global state.
+            let correlation_id = CorrelationId::new();
+            let maybe_stored_registry = tracking_copy
+                .borrow_mut()
+                .get(correlation_id, &Key::SystemContractRegistry)
+                .map_err(Into::into)?;
```
Fraser999: This probably warrants an `error!` call.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -253,50 +221,72 @@ where
...
+            let stored_registry = match maybe_stored_registry {
+                Some(registry) => registry,
+                None => {
```
Fraser999: This also seems like an invalid code path.  How could there validly be no previous registry?  It seems almost equivalent to the `Ok(None)` path in the replaced code, and for that we returned `Err(Error::InvalidProtocolVersion(current_protocol_version));`

If you agree, I think we should log an error here and return an `Err`.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -253,50 +221,72 @@ where
...
+                    let upgrade_registry = upgrade_config
+                        .global_state_update()
+                        .get(&Key::SystemContractRegistry)
+                        .expect("should posses registry in upgrade config")
+                        .to_owned();
+                    upgrade_registry
+                }
+            };
+            if let StoredValue::CLValue(cl_registry) = stored_registry {
+                CLValue::into_t::<BTreeMap<String, ContractHash>>(cl_registry)
```
Fraser999: We have an alias for this now - let's be consistent and stick with that.  Makes refactoring easier too if we change the type of the alias in the future.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -253,50 +221,72 @@ where
...
+                    .map_err(|error| GenesisError::CLValue(format!("{:?}", error)))?
```
Fraser999: It doesn't seem appropriate to return `GenesisError`s outside of the `commit_genesis` method.

Same comment 2 lines below.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -253,50 +221,72 @@ where
             return Err(Error::InvalidProtocolVersion(new_protocol_version));
         }
 
+        let registry = {
+            // Check if the registry exists in global state.
+            let correlation_id = CorrelationId::new();
+            let maybe_stored_registry = tracking_copy
```
Fraser999: Can we not use `tracking_copy.borrow_mut().get_system_contracts()` here?

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1150,8 +1124,6 @@ where
         };
 
         // vestigial system_contract_cache
```
```suggestion
```

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -2128,4 +2125,91 @@ where
...
+    fn get_system_contract_registry(
+        &self,
+        correlation_id: CorrelationId,
+        state_root_hash: Blake2bHash,
+    ) -> Result<BTreeMap<String, ContractHash>, Error> {
+        let tracking_copy = match self.tracking_copy(state_root_hash)? {
+            None => return Err(Error::RootNotFound(state_root_hash)),
+            Some(tracking_copy) => Rc::new(RefCell::new(tracking_copy)),
+        };
+        let stored_registry = tracking_copy
```
Fraser999: Can't we use the `get_system_contracts` on tracking copy here?

execution_engine/src/core/execution/executor.rs
```diff
@@ -393,11 +389,21 @@ impl Executor {
         R::Error: Into<Error>,
         T: FromBytes + CLTyped,
     {
+        let system_contract_registry = tracking_copy
+            .borrow_mut()
+            .get_system_contracts(correlation_id)
+            .unwrap_or_else(|error| panic!("Could not retrieve system contracts: {:?}", error));
```
Fraser999, line 396: Let's get rid of the panics in this file too please.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -1308,15 +1305,33 @@ where
     }
 
     pub fn is_mint(&self, key: Key) -> bool {
-        key.into_hash() == Some(self.protocol_data().mint().value())
+        key.into_hash()
+            == Some(
+                self.context
+                    .get_system_contract(MINT)
+                    .expect("should have mint hash")
```
Fraser999: Shouldn't use `expect`.  Same comment for next 2 changes.

execution_engine/src/core/runtime/standard_payment_internal.rs
```diff
@@ -50,7 +50,13 @@ where
         target: URef,
         amount: U512,
     ) -> Result<(), ApiError> {
-        let mint_contract_hash = self.get_mint_contract();
+        let mint_contract_hash = match self.get_mint_contract() {
+            Ok(mint_hash) => mint_hash,
+            Err(exec_error) => {
+                let maybe_api_error: Option<ApiError> = exec_error.into();
```
Fraser999: This will always be `None` I believe - please double check.

Seems like this probably warrants a new `ApiError` variant: `MissingSystemContractHash` with doc comment making it clear this is a bug if returned.

execution_engine/src/core/runtime/standard_payment_internal.rs
```diff
@@ -68,7 +74,12 @@ where
     R::Error: Into<execution::Error>,
 {
     fn get_payment_purse(&mut self) -> Result<URef, ApiError> {
-        let handle_payment_contract_hash = self.get_handle_payment_contract();
+        let handle_payment_contract_hash =
+            self.get_handle_payment_contract().map_err(|exec_error| {
```
Fraser999: Same comment as above.

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -15,24 +15,26 @@ use casper_types::{
...
-        tracking_copy::{AddResult, TrackingCopy},
+        tracking_copy::{AddResult, TrackingCopy, TrackingCopyExt},
         Address,
     },
     shared::{account::Account, gas::Gas, newtypes::CorrelationId, stored_value::StoredValue},
-    storage::{global_state::StateReader, protocol_data::ProtocolData},
+    storage::global_state::StateReader,
 };
 
+use tracing::error;
```
Fraser999: This should be moved.

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -758,10 +767,12 @@ where
...
     pub(crate) fn is_system_contract(&self) -> bool {
         if let Some(hash) = self.base_key().into_hash() {
-            let system_contracts = self.protocol_data().system_contracts();
-            if system_contracts.contains(&hash.into()) {
-                return true;
-            }
+            let contract_hash = ContractHash::new(hash);
+            return self
+                .system_contract_registry()
+                .expect("must get system contract registry")
```
Fraser999: Should not use `expect()`.

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -1118,4 +1129,23 @@ where
...
+    pub fn get_system_contract(&self, name: &str) -> Result<ContractHash, Error> {
+        let registry = self.system_contract_registry()?;
+        let hash = registry.get(name).ok_or_else(|| {
+            error!("Missing system contract hash: {}", name);
+            Error::MissingSystemContractHash(name.to_string())
+        })?;
+        Ok(*hash)
+    }
+
+    pub fn system_contract_registry(&self) -> Result<BTreeMap<String, ContractHash>, Error> {
```
Fraser999: Let's use the alias please.

execution_engine/src/core/runtime_context/tests.rs
```diff
@@ -32,20 +33,18 @@ use crate::{
...
-            in_memory::{InMemoryGlobalState, InMemoryGlobalStateView},
-            StateProvider,
-        },
-        protocol_data::ProtocolData,
+    storage::global_state::{
+        in_memory::{InMemoryGlobalState, InMemoryGlobalStateView},
+        StateProvider,
     },
 };
+use std::collections::BTreeMap;
```
Fraser999: Wrong place.

execution_engine/src/core/tracking_copy/ext.rs
```diff
@@ -15,6 +15,7 @@ use crate::{
     },
     storage::{global_state::StateReader, trie::merkle_proof::TrieMerkleProof},
 };
+use std::collections::BTreeMap;
```
Fraser999: Wrong place.

execution_engine/src/core/runtime_context/tests.rs
```diff
@@ -350,6 +349,19 @@ fn contract_key_addable_valid() {
     )));
     tracking_copy.borrow_mut().write(contract_key, contract);
 
+    let default_system_registry = {
+        let mut registry = BTreeMap::<String, ContractHash>::new();
```
Fraser999: Use alias.

execution_engine/src/core/tracking_copy/ext.rs
```diff
@@ -84,6 +85,11 @@ pub trait TrackingCopyExt<R> {
     ) -> Result<ContractPackage, Self::Error>;
 
     fn get_system_module(&mut self, preprocessor: &Preprocessor) -> Result<Module, Self::Error>;
+
+    fn get_system_contracts(
+        &mut self,
+        correlation_id: CorrelationId,
+    ) -> Result<BTreeMap<String, ContractHash>, Self::Error>;
```
Fraser999: Use alias.

execution_engine/src/core/tracking_copy/ext.rs
```diff
@@ -247,4 +253,25 @@ where
     fn get_system_module(&mut self, preprocessor: &Preprocessor) -> Result<Module, Self::Error> {
         Ok(wasm::do_nothing_module(preprocessor)?)
     }
+
+    fn get_system_contracts(
+        &mut self,
+        correlation_id: CorrelationId,
+    ) -> Result<BTreeMap<String, ContractHash>, Self::Error> {
```
Fraser999: Use alias.

execution_engine/src/core/tracking_copy/ext.rs
```diff
@@ -247,4 +253,25 @@ where
...
+            .get(correlation_id, &Key::SystemContractRegistry)
+            .map_err(Into::into)?
+        {
+            Some(StoredValue::CLValue(registry)) => {
+                let registry: BTreeMap<String, ContractHash> =
+                    CLValue::into_t(registry).map_err(Self::Error::from)?;
+                Ok(registry)
+            }
+            Some(other) => Err(execution::Error::TypeMismatch(TypeMismatch::new(
+                "StoredValue".to_string(),
```
Fraser999:
```suggestion
                "CLValue".to_string(),
```

execution_engine/src/core/tracking_copy/ext.rs
```diff
@@ -247,4 +253,25 @@ where
...
+    fn get_system_contracts(
+        &mut self,
+        correlation_id: CorrelationId,
+    ) -> Result<BTreeMap<String, ContractHash>, Self::Error> {
+        match self
+            .get(correlation_id, &Key::SystemContractRegistry)
+            .map_err(Into::into)?
+        {
+            Some(StoredValue::CLValue(registry)) => {
+                let registry: BTreeMap<String, ContractHash> =
```
Fraser999: Use alias.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -52,26 +48,21 @@ impl LmdbGlobalState {
...
+        Ok(LmdbGlobalState::new(environment, trie_store, root_hash))
     }
 
     /// Creates a state from an existing environment, store, and root_hash.
     /// Intended to be used for testing.
     pub(crate) fn new(
         environment: Arc<LmdbEnvironment>,
         trie_store: Arc<LmdbTrieStore>,
-        protocol_data_store: Arc<LmdbProtocolDataStore>,
+
```
Fraser999:
```suggestion
```

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -52,26 +48,21 @@ impl LmdbGlobalState {
...
         empty_root_hash: Blake2bHash,
     ) -> Self {
         LmdbGlobalState {
             environment,
             trie_store,
-            protocol_data_store,
+
```
Fraser999:
```suggestion
```

execution_engine_testing/test_support/src/internal/wasm_test_builder.rs
```diff
@@ -66,6 +66,7 @@ use casper_types::{
 use crate::internal::{
     utils, ExecuteRequestBuilder, DEFAULT_PROPOSER_ADDR, DEFAULT_PROTOCOL_VERSION, SYSTEM_ADDR,
 };
+use casper_execution_engine::core::engine_state::genesis::SystemContractRegistry;
```
Fraser999: Wrong place.

execution_engine_testing/tests/src/lmdb_fixture.rs
```diff
@@ -99,6 +100,7 @@ pub fn generate_fixture(
 
     // You can customize the fixture post genesis with a callable.
     post_genesis_setup(&mut builder);
+    // Pass in a function taking the builder
```
Fraser999: I don't understand this comment.

execution_engine_testing/tests/src/test/deploy/receipts.rs
```diff
@@ -6,7 +6,8 @@ use casper_engine_test_support::{
     internal::{ExecuteRequestBuilder, InMemoryWasmTestBuilder, DEFAULT_RUN_GENESIS_REQUEST},
     DEFAULT_ACCOUNT_ADDR,
 };
-use casper_execution_engine::storage::protocol_data::DEFAULT_WASMLESS_TRANSFER_COST;
+
```
Fraser999:
```suggestion
```

execution_engine_testing/tests/src/test/regression/gh_1470.rs
```diff
@@ -1,26 +1,26 @@
...
-        mint,
-    },
-    CLTyped, ContractHash, ContractPackageHash, EraId, Key, ProtocolVersion, RuntimeArgs, U512,
+    system::{auction, auction::DelegationRate, mint},
+    AccessRights, CLTyped, CLValue, ContractHash, ContractPackageHash, EraId, Key, ProtocolVersion,
+    RuntimeArgs, URef, U512,
 };
 
-use crate::lmdb_fixture;
-
```
Fraser999: Order needs fixed.

execution_engine_testing/tests/src/test/storage_costs.rs
```diff
@@ -31,6 +25,15 @@ use casper_types::{
...
+use casper_execution_engine::{
+    core::engine_state::{EngineConfig, DEFAULT_MAX_QUERY_DEPTH},
+    shared::{
+        host_function_costs::{HostFunction, HostFunctionCosts},
+        opcode_costs::OpcodeCosts,
+        system_config::SystemConfig,
+        wasm_config::{WasmConfig, DEFAULT_MAX_STACK_HEIGHT, DEFAULT_WASM_MAX_MEMORY},
+    },
+};
 use num_rational::Ratio;
```
Fraser999: Order needs fixed.

execution_engine_testing/tests/src/test/system_contracts/upgrade.rs
Fraser999: Looks like you didn't apply the suggestion.

execution_engine_testing/tests/src/test/system_contracts/upgrade.rs
```diff
@@ -156,23 +154,20 @@ fn should_allow_only_wasm_costs_minor_version() {
...
```
Fraser999: `5` should be replaced here too.

types/src/key.rs
```diff
@@ -63,6 +65,8 @@ const KEY_BALANCE_SERIALIZED_LENGTH: usize = KEY_ID_SERIALIZED_LENGTH + UREF_ADD
 const KEY_BID_SERIALIZED_LENGTH: usize = KEY_ID_SERIALIZED_LENGTH + KEY_HASH_LENGTH;
 const KEY_WITHDRAW_SERIALIZED_LENGTH: usize = KEY_ID_SERIALIZED_LENGTH + KEY_HASH_LENGTH;
 const KEY_DICTIONARY_SERIALIZED_LENGTH: usize = KEY_ID_SERIALIZED_LENGTH + KEY_DICTIONARY_LENGTH;
+const KEY_SYSTEM_CONTRACT_REGISTRY_LENGTH: usize =
```
Fraser999:
```suggestion
const KEY_SYSTEM_CONTRACT_REGISTRY_SERIALIZED_LENGTH: usize =
```

types/src/key.rs
```diff
@@ -309,6 +324,13 @@ impl Key {
             return Ok(Key::Dictionary(addr));
         }
 
+        if input
+            .strip_prefix(SYSTEM_CONTRACT_REGISTRY_PREFIX)
+            .is_some()
+        {
+            return Ok(Key::SystemContractRegistry);
+        }
```
Fraser999: This should parse the full string, not just the prefix.

types/src/key.rs
```diff
@@ -561,6 +593,10 @@ impl FromBytes for Key {
                 let (addr, rem) = DictionaryAddr::from_bytes(remainder)?;
                 Ok((Key::Dictionary(addr), rem))
             }
+            tag if tag == KeyTag::SystemContractRegistry as u8 => {
+                let (_, rem): ([u8; 32], &[u8]) = FromBytes::from_bytes(remainder)?;
+                Ok((Key::SystemContractRegistry, rem))
+            }
             _ => Err(Error::Formatting),
         }
     }
```
Fraser999, line 616: I can't apply the comment on the appropriate line, but `impl Distribution<Key> for Standard` should be updated to allow the new key variant to be randomly generated.

types/src/key.rs
```diff
@@ -700,6 +746,7 @@ mod serde_helpers {
                 BinaryDeserHelper::Bid(account_hash) => Key::Bid(account_hash),
                 BinaryDeserHelper::Withdraw(account_hash) => Key::Withdraw(account_hash),
                 BinaryDeserHelper::Dictionary(addr) => Key::Dictionary(addr),
+                BinaryDeserHelper::SystemContractRegistry => Key::SystemContractRegistry,
             }
         }
     }
```
Fraser999, line 767: The following need to be updated to include coverage for the new key variant, and also please update them to cover the `Dictionary` variant as that seems to have been overlooked:
* `KEYS`
* `should_display_key()`
* `should_fail_to_parse_key_from_str()`
* `key_to_json()`
* `serialization_roundtrip_json()`

Fraser999 (changes requested):

execution_engine/src/core/engine_state/mod.rs
darthsiroftardis: So from what I can understand there are cases, one case in which there is an error, which is covered by logging the error in the above comment: https://github.com/casper-network/casper-node/pull/1814#discussion_r683021804

The `None` case is where we check for a registry find it unavailable which is expected in the case of a pre-existing network, i.e validly no registry present
in which case it must be present in the `UpgradeConfig` as part of a global state update. However, if it isn't an error must be logged and returned

darthsiroftardis: Actually I have it as an expect, which I believe can be retained as; if the update isn't present it isn't a recoverable failure, how should this failure be handled @Fraser999

execution_engine_testing/tests/src/lmdb_fixture.rs
darthsiroftardis: That was footnote I forgot to take out

execution_engine/src/core/execution/executor.rs
```diff
@@ -434,6 +438,7 @@ impl Executor {
                 }
             }
             DirectSystemContractCall::GetEraValidators => {
+                // TODO See if these panics can be removed.
```
dwerner, line 441: TODO in this PR or in a subsequent one?

dwerner (commented):
Looking good, noting the presence of some remaining TODOs.

~~Is there any plan to to remove `execution_engine/src/storage/protocol_data*` in this PR?~~, sorry, I was reviewing a partial diff.

execution_engine/src/core/execution/executor.rs
darthsiroftardis: Subsequent PR

dwerner (approved):
Ignore my previous question about `protocol_data*.rs` removal. LGTM

execution_engine/src/core/execution/executor.rs
Fraser999: Well spotted, but we're aiming for a subsequent one.

Fraser999 (approved):

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -689,6 +695,7 @@ where
                 // bits are verified from within API call.
                 false
             }
+            Key::SystemContractRegistry => true,
```
mpapierski: I'd make this key unreadable by normal read APIs. We have a dedicated set of functions `get_mint()` `get_auction()` etc. on the WASM side. As currently implemented users won't even know what type is that as the `SystemContractRegistry` type alias is in EE and that's fine as WASM doesn't need to know how system contracts are stored.

```suggestion
            Key::SystemContractRegistry => false,
```

mpapierski (approved):
One more detail to address regarding readability of new Key variant under WASM otherwise LGTM

darthsiroftardis:
bors try

darthsiroftardis:
bors try

darthsiroftardis:
bors merge

PR #1832, opened 2021-08-02
Fix dial outcome logic

### LINKED ISSUE #1831: High reconnection frequency on cross-network failed connections
There is an observed ineffectiveness of the block list in the case of accidental cross-network connections. This is caused due to a bug where a connection that is blocked while the connection is being established ends up resetting the blocked state of said connection.

marc-casperlabs:
Adds a unit-level test for the observed issue with reconnections on testnet and fixes it in the code.

Fixes #1831


sacherjj (approved):
That looks right, but want Fraser check before merging.

Fraser999 (approved):

PR #1836, opened 2021-08-03
Allow longer hex-strings to be displayed in jsons

### LINKED ISSUE #1819: Rewrite PR 1555 with tests passing

rafal-ch:
* Allow longer hex string to be presented in `json` files. Current maximum is increased from 100 to 150 characters.
* Extract the `find_hex_strings_longer_than()` function to be able to unit-test it

Closes #1819

### DIFF
--- a/types/src/json_pretty_printer.rs
+++ b/types/src/json_pretty_printer.rs
@@ -7,6 +7,24 @@ use serde_json::{json, Value};
 
 const MAX_STRING_LEN: usize = 150;
 
+/// Represents the information about a substring found in a string.
+#[derive(Debug)]
+struct SubstringSpec {
+    /// Index of the first character.
+    start_index: usize,
+    /// Length of the substring.
+    length: usize,
+}
+
+impl SubstringSpec {
+    fn new(start_index: usize, length: usize) -> Self {
+        Self {
+            start_index,
+            length,
+        }
+    }
+}
+
 /// Serializes the given data structure as a pretty-printed `String` of JSON using
 /// `serde_json::to_string_pretty()`, but after first reducing any large hex-string values.
 ///
@@ -23,13 +41,15 @@ where
     serde_json::to_string_pretty(&json_value)
 }
 
-fn find_hex_strings_longer_than(s: &str, max_len: usize) -> Vec<(usize, usize)> {
+/// Searches the given string for all occurrences of hex substrings
+/// that are longer than the specified `max_len`.
+fn find_hex_strings_longer_than(string: &str, max_len: usize) -> Vec<SubstringSpec> {
     let mut ranges_to_remove = Vec::new();
     let mut start_index = 0;
     let mut contiguous_hex_count = 0;
 
     // Record all large hex-strings' start positions and lengths.
-    for (index, char) in s.char_indices() {
+    for (index, char) in string.char_indices() {
         if char.is_ascii_hexdigit() {
             if contiguous_hex_count == 0 {
                 // This is the start of a new hex-string.
@@ -39,14 +59,14 @@ fn find_hex_strings_longer_than(s: &str, max_len: usize) -> Vec<(usize, usize)> 
         } else if contiguous_hex_count != 0 {
             // This is the end of a hex-string: if it's too long, record it.
             if contiguous_hex_count > max_len {
-                ranges_to_remove.push((start_index, contiguous_hex_count));
+                ranges_to_remove.push(SubstringSpec::new(start_index, contiguous_hex_count));
             }
             contiguous_hex_count = 0;
         }
     }
     // If the string contains a large hex-string at the end, record it now.
     if contiguous_hex_count > max_len {
-        ranges_to_remove.push((start_index, contiguous_hex_count));
+        ranges_to_remove.push(SubstringSpec::new(start_index, contiguous_hex_count));
     }
     ranges_to_remove
 }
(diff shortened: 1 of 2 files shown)


types/src/json_pretty_printer.rs
```diff
@@ -23,40 +23,46 @@ where
...
-                    }
-                    contiguous_hex_count += 1;
-                } else if contiguous_hex_count != 0 {
-                    // This is the end of a hex-string: if it's too long, record it.
-                    if contiguous_hex_count > MAX_STRING_LEN {
-                        ranges_to_remove.push((start_index, contiguous_hex_count));
-                    }
-                    contiguous_hex_count = 0;
-                }
+fn find_hex_strings_longer_than(s: &str, max_len: usize) -> Vec<(usize, usize)> {
```
goral09: Can you add a comment to the method? The return type is very cryptic and could mean a lot of different things.

rafal-ch: I'd prefer to add a self-explaining type instead of a comment, like so:
```rust
struct SubstringSpec {
    start_index: usize,
    length: usize
}

fn find_hex_strings_longer_than(s: &str, max_len: usize) -> Vec<SubstringSpec> {...
```

What do you think?

Fraser999: I certainly prefer the struct to a tuple.

nilz3ro (approved):

types/src/json_pretty_printer.rs
nilz3ro: I think the function signature reads well 👍🏼 .
The self documenting struct is nicer to read IMO.

goral09: Why not both? If that function was in a library (and it seems to be generic enough to potentially be there), it would have a comment. I think people get uncle Bob's "great" advices too serious sometimes.

Fraser999: I must admit, I'd be inclined to add a doc comment to the function myself, but I don't feel strongly enough to insist since the function is well-named, and if the return type is changed too, it becomes pretty self-documenting.

Your call @rafal-ch - go for what you prefer.

Fraser999 (commented):
LGTM, given that the return type of `find_hex_strings_longer_than()` is updated as discussed.

types/src/json_pretty_printer.rs
rafal-ch: I'll do both :)
Thanks for the insights.

rafal-ch: Resolved in 3cfc1105f6c4e1628dccaffb9d6cf2c9446f85a4

types/src/json_pretty_printer.rs
```diff
@@ -86,6 +111,12 @@ mod tests {
         "0123456789abcdef".chars().cycle().take(length).collect()
     }
 
+    impl PartialEq<(usize, usize)> for SubstringSpec {
```
Fraser999, line 115: Nice to see this restricted to the `test` module.

Fraser999 (approved):
LGTM

darthsiroftardis (approved):

types/src/json_pretty_printer.rs
```diff
@@ -5,7 +5,25 @@ use alloc::{format, string::String, vec::Vec};
...
+#[derive(Debug)]
+struct SubstringSpec {
+    /// Index of the first character.
+    start_index: usize,
+    /// Length of the substring.
+    length: usize,
+}
+
+impl SubstringSpec {
+    fn new(start_index: usize, length: usize) -> Self {
```
marc-casperlabs, line 21: The completionist in me wants to add a (admittedly trivial) docstring here. :thinking:

types/src/json_pretty_printer.rs
```diff
@@ -5,7 +5,25 @@ use alloc::{format, string::String, vec::Vec};
 use serde::Serialize;
 use serde_json::{json, Value};
 
-const MAX_STRING_LEN: usize = 100;
+const MAX_STRING_LEN: usize = 150;
+
+/// Represents the information about a substring found in a string.
+#[derive(Debug)]
+struct SubstringSpec {
```
marc-casperlabs, line 12: So this is actually just a slice. A `SubstringSpec` in memory will be the size of two pointers, as will a `&[str]` be. However using one would make it awkward, since it would borrow the source string, I get that :)

I think the approach here could be improved by restructuring the replacement and NOT doing the "find, then replace" way. The current approach has the drawback that it performs a lot of allocations on the heap that are unnecessary.

My suggestion would be something like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=66df1d5ba144bc9931b096c5e0388c14

By doing the repleacement as we go over the string, we only iterate over it once, and only allocate the output string, nothing else. This would also obviate the need for something like this `SubstringSpec`.

types/src/json_pretty_printer.rs
marc-casperlabs: I disagree and would prefer to just derive it, but oh well :)

marc-casperlabs (commented):
Many others have greenlit this already, I don't disagree with that. If we are changing the module, I would like to see it refactored as described in the comment, but this will service regardless :)

types/src/json_pretty_printer.rs
rafal-ch: Resolved in 34594446fa1a7275bea6322f2283ad9719071a91

types/src/json_pretty_printer.rs
rafal-ch: As agreed, we'll consider further refactoring in a separate PR.

PR #1843, opened 2021-08-03
Restricting access to pub(crate)

### LINKED ISSUE #1842: Node Modules Should Be `pub(crate)`
While the node project has types it exposes as a library, it has many implementation details that it leaks.

Enforcing that modules be `pub(crate)` allows the compiler to statically analyze these implementation details, identifying dead code.

xcthulhu:
All modules except for types, `cli`, `rpc` and `crypto` for node are now `pub(crate)`.  Dead code has been identified and removed.

Closes https://github.com/casper-network/casper-node/issues/1842


node/src/utils/external.rs
```diff
@@ -40,9 +40,11 @@ pub static RESOURCES_PATH: Lazy<PathBuf> =
...
 /// used instead.
+// TODO: `Loaded` is only ever used in tests, refactor this
+#[allow(unused)]
 #[derive(Clone, DataSize, Eq, Debug, Deserialize, PartialEq, Serialize)]
 #[serde(untagged)]
-pub enum External<T> {
+pub(crate) enum External<T> {
     /// Value that should be loaded from an external path.
     Path(PathBuf),
     /// Loaded or immediate value.
```
xcthulhu: `Loaded` is only used in test, we should get rid of it.  CC @marc-casperlabs

node/src/lib.rs
```diff
@@ -49,24 +35,33 @@ use signal_hook::{
...
+#[cfg(test)]
+pub(crate) mod testing;
+pub(crate) mod tls;
+pub(crate) mod utils;
+
+// Public modules
+pub use components::rpc_server::rpcs;
+pub mod cli;
+pub mod crypto;
+pub mod types;
```
Fraser999: Let's keep the `mod` statements together above all the imports please.

node/src/reactor.rs
```diff
@@ -419,20 +419,10 @@ where
...
-        // Instantiate a new registry for metrics for this reactor.
-        let registry = Registry::new();
-        Self::with_metrics(cfg, rng, &registry).await
-    }
-
     /// Creates a new runner from a given configuration, using existing metrics.
     #[inline]
     #[instrument("runner creation", level = "debug", skip(cfg, rng, registry))]
-    pub async fn with_metrics(
+    pub(crate) async fn with_metrics(
```
Fraser999, line 440: Might be as well renaming this to just `new()` now.

node/src/reactor/queue_kind.rs
```diff
@@ -4,7 +4,7 @@
 //! round-robin manner. This way, events are only competing for time within one queue, non-congested
 //! queues can always assume to be speedily processed.
 
-use std::{fmt::Display, num::NonZeroUsize};
+use std::fmt::Display;
```
Fraser999: Why was this change made?  I think it was correct before the change.

types/src/lib.rs
```diff
@@ -97,7 +97,3 @@ pub use crate::{
 compile_error!(
     "casper-types requires one of 'std' (enabled by default) or 'no-std' features to be enabled"
 );
-#[cfg(all(feature = "std", feature = "no-std"))]
```
Fraser999: Removing this seems incorrect.  The `std` and `no-std` features in this crate are meant to be mutually exclusive and due to the limitations of Rust manifests, this is the only way that is enforced now.

It has made me notice that the error message for this and the `compile_error!` above are both wrong - `no-std` is enabled by default.

node/src/components/contract_runtime/announcements.rs
```diff
@@ -0,0 +1,121 @@
+use std::{
```
Fraser999: Looks like this file has been included in the PR by mistake.

Fraser999 (commented):
This is highly satisfying to review - nice work!

node/src/utils/external.rs
xcthulhu: Removed in 7470699

node/src/lib.rs
xcthulhu: I've moved all of the `mod` declarations above the `use` declarations  in c2860ed92ab602884d3ac08b971ed8f05bc5ae71

I recognize it is important to have `mod` declaration before `use` declarations to avoid name collisions.

types/src/lib.rs
xcthulhu: This has been breaking several people's tooling (ie, `rust-analyzer`).  We are going to have to figure out something here that works for everyone.

node/src/reactor/queue_kind.rs
xcthulhu: Fixed in 6cd7b1ee536c98cd9ccfb093df88cd18355678a1

node/src/reactor.rs
xcthulhu: `new()` is a `#[cfg(test)]` (see [below][1])

[1]: https://github.com/casper-network/casper-node/pull/1843/files/7470699d39396ffc99ee770247e72d4531d1e70c#diff-f37f8af7d3f5387726dc2460b5d53b8a4092e62a97a62c6491c7265b5f496801R727-R730

node/src/components/contract_runtime/announcements.rs
xcthulhu: Good catch! Deleted in 7877d0e

types/src/lib.rs
Fraser999: I only know of one person's tooling.  However, even if it's affecting more, let's not make an unrelated change like that in this PR please?

node/src/components/small_network.rs
```diff
@@ -675,36 +670,6 @@ where
             .ignore()
     }
 
-    /// Sweeps across connection symmetry, enforcing symmetrical connections.
-    fn enforce_symmetric_connections(&mut self, now: Instant) -> Effects<Event<P>> {
```
marc-casperlabs, line 689: This function is currently being put "back into use" as part of #1844, similar to `SweepSymmetries`. How do we want to tackle this?

marc-casperlabs: (We will have a merge conflict, depending on which is merged first).

node/src/components/small_network/error.rs
```diff
@@ -21,15 +21,12 @@ pub enum Error {
     #[error("could not resolve at least one known host (or none provided)")]
     EmptyKnownHosts,
     /// Consensus signing during handshake was provided, but keys could not be loaded.
-    #[error("consensus keys provided, but could not be loaded")]
+    #[error("consensus keys provided, but could not be loaded: {0}")]
```
marc-casperlabs, line 24: All errors should be (in logs) displayed with `display_error`, which is at least the case inside of `small_network`. The function shows the complete chain of error sources. Adding the `{0}` here is redundant, as far as I know it'd make that output look like "consensus keys provided, but could not be loaded: foo load error: foo load error: bar error".

node/src/components/small_network/symmetry.rs
```diff
@@ -173,8 +168,13 @@ impl ConnectionSymmetry {
         }
     }
 
+    #[cfg(test)]
     /// Indicates whether or not a connection should be cleaned up.
-    pub(super) fn should_be_reaped(&self, now: Instant, max_time_asymmetric: Duration) -> bool {
+    pub(super) fn should_be_reaped(
+        &self,
+        now: Instant,
+        max_time_asymmetric: std::time::Duration,
```
marc-casperlabs: Was the change intentional here? I have been asked to change it from qualified to unqualified in the past and `Duration` is fairly unambiguous, I think. :)

node/src/components/small_network.rs
Fraser999: I noticed that too, but I think it's such a trivial conflict, it's better to just remove it here and re-add it in your PR.

node/src/components/small_network/symmetry.rs
Fraser999: Good point - `std::time::Duration` should be imported behind a `#[cfg(test)]` gate.

node/src/utils/external.rs
marc-casperlabs: If we're not doing `Loaded` at all, the whole type should go away. The sole purpose of the enum was to make it easy to supply custom values for configuration that may be in an external file in a test. Temporary files have to be used instead if this is removed. I'm not fundamentally opposed to that, at the time I just thought it would be nice if that was not necessary.

So maybe we want to go a step further and remove the logic from the configuration entirely. Components would then themselves be responsible for loading data from a file.

This does move the point of failure back a bit (from configuration load to execution), but maybe the reduced complexity is worth it.

marc-casperlabs (approved):
Looks good in general to me, except for some minor things that should be addressed.

node/src/components/small_network.rs
marc-casperlabs: Ha, easy for you to say, but fair enough I guess :)

types/src/lib.rs
xcthulhu: Okay, so chatting this over with @Fraser999 , the concern here is that smart-contract writers might get knackered if they have both `std` and `no-std` feature flags set.

node/src/components/small_network/error.rs
xcthulhu: Isn't that what it looks like everywhere else?

node/src/utils/external.rs
xcthulhu: As discussed, we should consider iterating on this later.

node/src/components/small_network/error.rs
marc-casperlabs: Yes, unfortunately. But I would really prefer to see it changed anywhere else than here :).

At the very least, let's keep the small_network component consistent internally?

types/src/lib.rs
xcthulhu: Reverted in d32751566c18a5b587030abbf4f664e740c75b54

node/src/components/small_network/error.rs
xcthulhu: Let's leave this for now and figure out a solution where everyone can be on the same page.

Fraser999 (approved):

node/src/utils/external.rs
xcthulhu: Removed spurious comment in 7f003c0b3a8ac58d47082a3a36ed50ab85d5c5d8

node/src/components/small_network/symmetry.rs
xcthulhu: Fixed in d9ae2b730b2572b58d76b49b5ce201f7b0dd1719

xcthulhu: Fixed in https://github.com/casper-network/casper-node/pull/1843/commits/dc2919bc75167f006ab5842afbdeb288f311f845

xcthulhu:
bors merge

xcthulhu:
bors merge

node/src/reactor.rs
```diff
@@ -674,7 +677,6 @@ where
 
     /// Runs the reactor until `maybe_exit()` returns `Some` or we get interrupted by a termination
     /// signal.
-    #[inline]
```
dwerner: Intentional? If so, then also maybe we should think about removing `#[inline]` from some of the other `async` methods here.

marc-casperlabs: IIRC it was removed because that method used to be very short and then grew, without anyone removing the `#[inline]` declaration.

It is in theory academical at this point, since the method is `pub(crate)` now.

xcthulhu: While inlining, loop unraveling, and caching are powerful optimization techniques, they should be guided by evidence IMO.  This was done with your deploy caching @marc-casperlabs .

xcthulhu:
bors merge

Issue #1768, opened 2021-07-23
Docs update: execution_engine/storage


mpapierski:
Add or update existing rust docs in a `storage` rust module in EE:

- [x] storage/error
- [x] storage/global_state
- [ ] storage/protocol_data_store
- [x] storage/store
- [x] storage/transaction_source
- [x] storage/trie
- [x] storage/trie_store
- [x] storage/*

dwerner:
Opting not to document `protocol_data` and `protocol_data_store` because those modules should be removed in #1814

dwerner:
merged to dev in https://github.com/casper-network/casper-node/pull/1848

cspramit:
@dwerner closing issue since merged please note though " storage/protocol_data_store" is unchecked

PR #1848, opened 2021-08-04
GH-1768 - add documentation to storage, leverage pub(crate) to prune …

dwerner:
The purpose of this PR is to address issue #1768 - add docs to the `storage` module in `execution_engine`. I've also included some cleanup of code that was unused that I found by marking it `pub(crate)`.


execution_engine/src/storage/error/mod.rs
```diff
@@ -1,4 +1,9 @@
+//! Storage errors.
+
+/// Errors for In-Memory storage implementation.
```
xcthulhu, line 3: As you and I discussed, perhaps this can be `#[cfg(test)]`?

execution_engine/src/storage/global_state/in_memory.rs
```diff
@@ -28,18 +30,20 @@ use crate::storage::{
...
+/// used for testing purposes.
 pub struct InMemoryGlobalState {
-    pub environment: Arc<InMemoryEnvironment>,
-    pub trie_store: Arc<InMemoryTrieStore>,
-    pub protocol_data_store: Arc<InMemoryProtocolDataStore>,
-    pub empty_root_hash: Blake2bHash,
+    pub(crate) environment: Arc<InMemoryEnvironment>,
+    pub(crate) trie_store: Arc<InMemoryTrieStore>,
+    pub(crate) protocol_data_store: Arc<InMemoryProtocolDataStore>,
+    pub(crate) empty_root_hash: Blake2bHash,
```
xcthulhu, line 38: Recently when I was doing this sort of thing, I just did `pub(crate)` for the whole file...

xcthulhu: Looking below, I see now that while these structs are `pub`, we are restricting access to these internal details to just this crate, which makes sense.

execution_engine/src/lib.rs
```diff
@@ -16,4 +16,6 @@
 pub mod config;
 pub mod core;
```
xcthulhu, line 17: Perhaps `core` can be `pub(crate)`?  CC @mpapierski @Fraser999

execution_engine/src/storage/protocol_data.rs
```diff
@@ -128,23 +134,6 @@ impl ProtocolData {
...
-            } else if old_hash == self.standard_payment {
-                self.standard_payment = new_hash;
-            } else if old_hash == self.auction {
-                self.auction = new_hash;
-            } else {
-                return false;
-            }
-        }
-        true
-    }
```
xcthulhu: 👍

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -24,18 +26,19 @@ use crate::storage::{
...
+/// Global state implemented against LMDB as a backing data store.
 pub struct LmdbGlobalState {
-    pub environment: Arc<LmdbEnvironment>,
-    pub trie_store: Arc<LmdbTrieStore>,
-    pub protocol_data_store: Arc<LmdbProtocolDataStore>,
-    pub empty_root_hash: Blake2bHash,
+    pub(crate) environment: Arc<LmdbEnvironment>,
+    pub(crate) trie_store: Arc<LmdbTrieStore>,
+    pub(crate) protocol_data_store: Arc<LmdbProtocolDataStore>,
+    pub(crate) empty_root_hash: Blake2bHash,
```
xcthulhu, line 33: This is always `blake2b(EmptyPointerBlock)` (same for the in-memory); it _could_ just be replaced by a lazy static.

execution_engine/src/storage/global_state/in_memory.rs
```diff
@@ -28,18 +30,20 @@ use crate::storage::{
...
 };
 
+/// Global state implemented purely in memory only. No state is saved to disk, and this is mostly
+/// used for testing purposes.
 pub struct InMemoryGlobalState {
-    pub environment: Arc<InMemoryEnvironment>,
-    pub trie_store: Arc<InMemoryTrieStore>,
-    pub protocol_data_store: Arc<InMemoryProtocolDataStore>,
-    pub empty_root_hash: Blake2bHash,
+    pub(crate) environment: Arc<InMemoryEnvironment>,
```
xcthulhu, line 34:
```suggestion
    /// Environment for `InMemoryGlobalState`.
    /// Basically empty because this global state does not support transactions.
    pub(crate) environment: Arc<InMemoryEnvironment>,
```

xcthulhu: Chatting with @dwerner , even though this is private we should document this for others who may review it.

execution_engine/src/storage/error/mod.rs
dwerner: We talked more, and in-memory global state is going to be useful for perf profiling purposes.

execution_engine/src/storage/global_state/lmdb.rs
dwerner: It doesn't seem to be _quite_ that simple in this case, and since it's a documentation-oriented ticket I'd prefer not to refactor this at this point.

execution_engine/src/lib.rs
Fraser999: I'm all for reducing the public interface to the minimum required, but just not in this PR

execution_engine/src/storage/error/in_memory.rs
```diff
@@ -4,11 +4,14 @@ use thiserror::Error;
 
 use casper_types::bytesrepr;
 
+/// Error enum encapsulating possible errors from
```
Fraser999: Incomplete?

execution_engine/src/storage/store/mod.rs
```diff
@@ -7,13 +7,20 @@ use casper_types::bytesrepr::{self, FromBytes, ToBytes};
 pub use self::store_ext::StoreExt;
 use crate::storage::transaction_source::{Readable, Writable};
 
+/// Store is reponsible for abstracting `get` and `put` operations over the underlying store
+/// specified by it's associated `Handle` type.
```
Fraser999:
```suggestion
/// specified by its associated `Handle` type.
```

execution_engine/src/storage/store/mod.rs
```diff
@@ -7,13 +7,20 @@ use casper_types::bytesrepr::{self, FromBytes, ToBytes};
...
+    /// Errors possible from this store.
     type Error: From<bytesrepr::Error>;
 
+    /// Underlying store type.
     type Handle;
 
+    /// `handle` returns the underlying store.
     fn handle(&self) -> Self::Handle;
 
+    /// `get` gets an optional value (may exist or not) as read through a transaction, or an error
```
Fraser999:
```suggestion
    /// Returns an optional value (may exist or not) as read through a transaction, or an error
```

execution_engine/src/storage/store/mod.rs
```diff
@@ -31,6 +38,8 @@ pub trait Store<K, V> {
         }
     }
 
+    /// `put` puts a `value` into the store at `key` within a transaction, potentially returning an
```
Fraser999:
```suggestion
    /// Puts a `value` into the store at `key` within a transaction, potentially returning an
```
Probably also worth stating what happens if the value already exists.

execution_engine/src/storage/store/store_ext.rs
```diff
@@ -1,11 +1,15 @@
...
 use casper_types::bytesrepr::{FromBytes, ToBytes};
 
 use crate::storage::{
     store::Store,
     transaction_source::{Readable, Writable},
 };
 
+/// Extension trait for Store.
 pub trait StoreExt<K, V>: Store<K, V> {
+    /// Get many key/value pairs from the store.
```
Fraser999:
```suggestion
    /// Returns multiple optional values (each may exist or not) from the store in one transaction.
```

execution_engine/src/storage/store/store_ext.rs
```diff
@@ -25,6 +29,7 @@ pub trait StoreExt<K, V>: Store<K, V> {
         Ok(ret)
     }
 
+    /// Put many key/value pairs into the store.
```
Fraser999:
```suggestion
    /// Puts multiple key/value pairs into the store in one transaction, potentially returning an
    /// error of type `Self::Error` if that fails.
```

execution_engine/src/storage/trie/mod.rs
```diff
@@ -161,10 +171,12 @@ impl<'de> Deserialize<'de> for PointerBlock {
 }
 
 impl PointerBlock {
+    /// No-arg contstructor for `PointerBlock`. Delegates to `Default::default()`.
```
Fraser999:
```suggestion
    /// No-arg constructor for `PointerBlock`. Delegates to `Default::default()`.
```

types/src/lib.rs
```diff
@@ -97,7 +97,3 @@ pub use crate::{
 compile_error!(
     "casper-types requires one of 'std' (enabled by default) or 'no-std' features to be enabled"
 );
-#[cfg(all(feature = "std", feature = "no-std"))]
-compile_error!(
-    "casper-types features 'std' (enabled by default) and 'no-std' should not both be enabled"
-);
```
Fraser999: This should be reverted.

dwerner: Didn't mean to commit this.

Fraser999 (approved):

PR #1862, opened 2021-08-06
Add more asserts to run_equivocator_network.

### LINKED ISSUE #1859: Non-det failure of `run_equivocator_network` test.

```
failures:

  left: `[]`,
```

source:

afck:
Assert that the equivocator is not a validator anymore after up to 20 eras. The probability of no equivocation happening should be negligible in this test, but if it happens, this would tell us why the test failed.

Also assert that the equivocator is only in the `equivocators` collection, and not e.g. in `inactive_validators`. It shouldn't be possible for doppelgänger detection to make both nodes appear inactive, but if that somehow were the case, this would alert us.

I couldn't reproduce #1859: It passed 10 times, even with the added assertions. I'd close it for now; if it happens again, hopefully these should tell us more about what went wrong.

Closes  #1859.

### DIFF
--- a/node/src/reactor/participating/tests.rs
+++ b/node/src/reactor/participating/tests.rs
@@ -267,7 +267,7 @@ async fn run_equivocator_network() {
         }
 
         let expected = [alice_pk.clone()];
-        // Returns true if alice is listed as an equivocator in that block.
+        // Returns true if Alice is listed as an equivocator in that block.
         let alice_is_equivocator = |header: &BlockHeader| {
             header.era_end().expect("missing era end").equivocators == expected
         };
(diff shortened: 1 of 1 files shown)


node/src/reactor/participating/tests.rs
```diff
@@ -301,6 +301,16 @@ async fn run_equivocator_network() {
...
     }
 
+    assert!(
+        !switch_blocks
+            .last()
+            .expect("missing switch block")
+            .next_era_validator_weights()
+            .expect("missing validator weights")
+            .contains_key(&alice_pk),
+        "Alice should have been evicted."
```
goral09, line 311: Shouldn't that be the other way around? If we expect Alice to be evicted then she shouldn't be in `next_era_validator_weights` which means we should fail here if she is - we don't need the negation.

goral09: NVM. I was wrong.

goral09 (approved):

fizyk20 (approved):

afck:
bors merge

PR #1872, opened 2021-08-09
Event tracing

### LINKED ISSUE #1849: Improve event tracing
We should be able to trace events further and possibly associate effects with the event that caused them.

marc-casperlabs:
This PR has two headlining features:

* Event ancestor tracing: Every event is now assigned a non-zero ID (before it used to start at zero) at the time it is dispatched, logged as `ev`. Any event resulting from an effect created by that event will have that event ID stored in the queue as its ancestor. When an event is dispatched, the ancestor is logged in the span as `a` as well. This allows following a single event "around" or tracing log message back to the event that originally triggered them.
* Single event log reduction: Before, events used to be logged twice, once using `Display` at `DEBUG` and `Debug` at `TRACE` levels. Now, they are only logged at the `TRACE` level as `Display`, leading in a potential reduction of log sizes of 50+% in many cases.

Closes #1849

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -450,7 +450,7 @@ where
 
     /// Creates a new runner from a given configuration, using existing metrics.
     #[inline]
-    #[instrument("runner creation", level = "debug", skip(cfg, rng, registry))]
+    #[instrument("init", level = "debug", skip(cfg, rng, registry))]
     pub async fn with_metrics(
         cfg: R::Config,
         rng: &mut NodeRng,
@@ -485,7 +485,7 @@ where
         Ok(Runner {
             scheduler,
             reactor,
-            current_event_id: 0,
+            current_event_id: 1,
             metrics: RunnerMetrics::new(registry)?,
             last_metrics: Instant::now(),
             event_metrics_min_delay: Duration::from_secs(30),
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -39,6 +39,7 @@ use std::{
     fmt::{Debug, Display},
     fs::File,
     mem,
+    num::NonZeroU64,
     str::FromStr,
     sync::atomic::Ordering,
 };
@@ -167,7 +168,10 @@ pub enum ReactorExit {
 /// is the central hook for any part of the program that schedules events directly.
 ///
 /// Components rarely use this, but use a bound `EventQueueHandle` instead.
-pub type Scheduler<Ev> = WeightedRoundRobin<Ev, QueueKind>;
+///
+/// Schedule tuples contain an optional ancestor ID and the actual event. The ancestor ID indicates
+/// which potential previous event resulted in the event being created.
+pub type Scheduler<Ev> = WeightedRoundRobin<(Option<NonZeroU64>, Ev), QueueKind>;
 
 /// Event queue handle
 ///
@@ -193,12 +197,27 @@ impl<REv> EventQueueHandle<REv> {
     }
 
     /// Schedule an event on a specific queue.
+    ///
+    /// The scheduled event will not have an ancestor.
     #[inline]
     pub(crate) async fn schedule<Ev>(self, event: Ev, queue_kind: QueueKind)
     where
         REv: From<Ev>,
     {
-        self.0.push(event.into(), queue_kind).await
+        self.schedule_with_ancestor(None, event, queue_kind).await
+    }
+
+    /// Schedule an event on a specific queue.
+    #[inline]
+    pub(crate) async fn schedule_with_ancestor<Ev>(
+        self,
+        ancestor: Option<NonZeroU64>,
+        event: Ev,
+        queue_kind: QueueKind,
+    ) where
+        REv: From<Ev>,
+    {
+        self.0.push((ancestor, event.into()), queue_kind).await
     }
 
     /// Returns number of events in each of the scheduler's queues.
@@ -552,7 +571,7 @@ where
             QUEUE_DUMP_REQUESTED.store(false, Ordering::SeqCst);
         }
 
-        let (event, queue) = self.scheduler.pop().await;
+        let ((_ancestor, event), queue) = self.scheduler.pop().await;
 
         // Create another span for tracing the processing of one event.
         let event_span = debug_span!("dispatch", event_count = self.event_count, %event, %queue);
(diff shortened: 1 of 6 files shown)


node/src/reactor.rs
```diff
@@ -552,50 +571,56 @@ where
             QUEUE_DUMP_REQUESTED.store(false, Ordering::SeqCst);
         }
 
-        let (event, queue) = self.scheduler.pop().await;
+        let ((ancestor, event), queue) = self.scheduler.pop().await;
+        trace!(%event, %queue, "current");
```
afck, line 575: Should we log the ancestor here?

node/src/reactor.rs
```diff
@@ -552,50 +571,56 @@ where
...
 
         // Create another span for tracing the processing of one event.
```
afck, line 577: So we're not creating a new span anymore, right? We're basically using the event IDs as a replacement?

node/src/reactor.rs
```diff
@@ -552,50 +571,56 @@ where
...
-                    true,
-                )
-            };
+        Span::current().record("ev", &self.current_event_id);
 
-            let end = self.clock.end();
+        // If we know the ancestor of an event, record it.
+        if let Some(ancestor) = ancestor {
+            Span::current().record("a", &ancestor.get());
+        }
```
afck, line 583: If this is `None`, could `a` still be set (incorrectly) from the previous event?

afck (approved):

node/src/reactor.rs
marc-casperlabs: No, it goes onto the span. However if we move the `trace!` call further down, the log message will have the ancestor ID already in there when it is logged. Very nice catch!

node/src/reactor.rs
marc-casperlabs: We already get a new span every time `crank` is called, as the `instrument` macro will create one for us. But instead of creating another nested span, we made room for an event ID on the parent.

node/src/reactor.rs
marc-casperlabs: No, as above - `instrument` will create a new span each time `crank` is entered.

marc-casperlabs: ... on second thought, let me make extra sure :)

goral09:
Can you show an example of that new tracing? For example an event that is created as a reaction to an incoming request from another node - like an incoming consensus' vertex that requires fetching its dependencies. 

Also, can you explain when does this "even thread" ends? When we run out of effects (`Effects::new`, empty vector of effects) in the return value?

node/src/reactor.rs
marc-casperlabs:
```rust
use tracing::{warn, Span};

#[tracing::instrument(skip(to_set), fields(x))]
fn instrumented_function(to_set: Option<u8>) {
    if let Some(val) = to_set {
        Span::current().record("x", &val);
    }
    warn!("this is a message");
}

fn main() {
    tracing_subscriber::fmt::init();

    instrumented_function(None);
    instrumented_function(None);
    instrumented_function(Some(3));
    instrumented_function(None);
    instrumented_function(Some(4));
    instrumented_function(Some(5));
    instrumented_function(None);
}
```

works as expected:

```
Aug 09 15:13:07.143  WARN instrumented_function: tracingtest: this is a message
Aug 09 15:13:07.143  WARN instrumented_function: tracingtest: this is a message
Aug 09 15:13:07.143  WARN instrumented_function{x=3}: tracingtest: this is a message
Aug 09 15:13:07.143  WARN instrumented_function: tracingtest: this is a message
Aug 09 15:13:07.144  WARN instrumented_function{x=4}: tracingtest: this is a message
Aug 09 15:13:07.144  WARN instrumented_function{x=5}: tracingtest: this is a message
Aug 09 15:13:07.144  WARN instrumented_function: tracingtest: this is a message
```

marc-casperlabs:
> Can you show an example of that new tracing? For example an event that is created as a reaction to an incoming request from another node - like an incoming consensus' vertex that requires fetching its dependencies.

Sorry, on short notice I can only do an approximation using an old log:

Here's two message that happened while event 236 was processed, e.g. "synchronize queue lengths" and "oldest seen unit per validator". IT shows that the ancestors for 236 is 217.

```
{"timestamp":"Aug 09 10:28:15.031","level":"DEBUG","fields":{"message":"synchronizer queue lengths","era_id":"ed40affc7a881f3d422d0c12b31563f5b7639d59826a285dfb75cdaec5ade8b0","vertices_to_be_added_later":0,"vertices_no_deps":0,"vertices_awaiting_deps":0},"target":"casper_node::components::consensus::highway_core::synchronizer","span":{"a":217,"ev":236,"name":"dispatch"},"spans":[{"a":217,"ev":236,"name":"dispatch"}]}
{"timestamp":"Aug 09 10:28:15.031","level":"DEBUG","fields":{"message":"oldest seen unit per validator","oldest_panorama":"ValidatorMap(0, 0, 0, N, 0)"},"target":"casper_node::components::consensus::highway_core::synchronizer","span":{"a":217,"ev":236,"name":"dispatch"},"spans":[{"a":217,"ev":236,"name":"dispatch"}]}
```

We can scroll up and see that 217 has also multiple events, one also including "oldest seen unit per validator":
```
{"timestamp":"Aug 09 10:28:10.029","level":"DEBUG","fields":{"message":"oldest seen unit per validator","oldest_panorama":"ValidatorMap(N, N, N, N, N)"},"target":"casper_node::components::consensus::highway_core::synchronizer","span":{"a":198,"ev":217,"name":"dispatch"},"spans":[{"a":198,"ev":217,"name":"dispatch"}]}
```

Here the ancestor is 198, which we can again follow. I picked these completely at random. If I look into the source code, the origin of that message is from a function called `log_len`, which in turn is only called once in this snipped:

```rust
            TIMER_ID_SYNCHRONIZER_LOG => {
                self.synchronizer.log_len();
                if !self.finalized_switch_block() {
                    let next_timer = Timestamp::now() + TimeDiff::from(5_000);
                    vec![ProtocolOutcome::ScheduleTimer(next_timer, timer_id)]
                } else {
                    vec![]
                }
            }
```

Not surprisingly, it is from a timer - from the first time the timer was fire we will get a chain, because the next timer is always has its predecessor as an ancestor.

There are better examples if you run with `TRACE` log level and the updated code:

> incoming request from another node - like an incoming consensus' vertex that requires fetching its dependencies.

This is a bit tricky, because I do not know if you can see this in the logs right away, as consensus messages are opaque to the rest of the system. But assuming you found a log line in consensus that corresponds to that, that line will have `ev` and `a` set -- it is present all log messages when that event is processed. So `ev` will be the "originating" network message. I will assume `ev` is 12345 here.

From there on out, just look for any log message with `"a":12345` and you're set.

> Also, can you explain when does this "even thread" end? When we run out of effects (Effects::new, empty vector of effects) in the return value?

Yes, only when an empty vector of effects is returned.

goral09:
> Sorry, on short notice I can only do an approximation using an old log:

Can't you run a network with the upgraded tracing and check how it looks like?

marc-casperlabs:
> Can't you run a network with the upgraded tracing and check how it looks like?

Sure, I'll have to defer you for a few hours though :)

PR #1877, opened 2021-08-10
Consistently use humantime in config files

### LINKED ISSUE #1821: Consistent use of durations in configs and chainspec
CC @Fraser999

rafal-ch:
* Use human-readable time units in all parameters that relate to duration in the `config.toml` file (`"30sec"`, `"120min"`, etc.)
* Code that handles these parameters no longer uses the raw integer data types, but the dedicated `TimeDiff` type

This PR introduces potentially breaking change - names of three parameters in config file have changed:
* `finished_entry_duration_secs` => `finished_entry_duration`
* `gossip_request_timeout_secs` => `gossip_request_timeout`
* `get_remainder_timeout_secs` => `get_remainder_timeout`

Closes #1821.


node/src/components/fetcher/tests.rs
```diff
@@ -7,13 +7,14 @@ use casper_node_macros::reactor;
...
 use thiserror::Error;
-use tokio::time;
 
 use casper_types::ProtocolVersion;
 
 use super::*;
 use crate::{
-    components::{deploy_acceptor, in_memory_network::NetworkController, storage},
+    components::{
+        deploy_acceptor, in_memory_network::NetworkController, storage, tests_common::advance_time,
```
Fraser999: Our convention is to not fully import functions, so this should be just `tests_common` and will become `tests_common::advance_time(time_to_advance).await;` at the callsite below.

node/src/components/gossiper/tests.rs
```diff
@@ -25,6 +25,7 @@ use crate::{
         deploy_acceptor::{self, DeployAcceptor},
         in_memory_network::{self, InMemoryNetwork, NetworkController},
         storage::{self, Storage},
+        tests_common::advance_time,
```
Fraser999: Same point here.

node/src/components/tests_common.rs
```diff
@@ -0,0 +1,11 @@
+use std::time::Duration;
```
Fraser999: We have a `testing` module in the root of node which is feature-gated behind `test` already.  I think this function should live there.

Fraser999 (commented):
We have a `resources/production/config-example.toml` which should be updated in this PR too.

Other than the nitpick comments, this LGTM.

Fraser999 (approved):
LGTM

darthsiroftardis (approved):
LGTM!!

nilz3ro (approved):

PR #1885, opened 2021-08-10
Add validity check for zero approvals in a Deploy

### LINKED ISSUE #1646: Check deploy for non-emptiness of the approvals collection
Node accepted and included a deploy with NO APPROVALS on it: https://integration.cspr.live/deploy/7c5a04272d47c0bfed48340280c969a5430d70f6262165e2619cdd7bdaa955ae

darthsiroftardis:
CHANGELOG:

- Added a condition during a deploy's validity check that the `approvals` field must be non-empty
- Added a new `DeployValidatorFailure` variant `EmptyApprovals` to be raised if a deploy's approvals field is empty
- Added a unit test to check that deploy's with empty approvals raise a `DeployValidationFailure`

Closes #1646

### DIFF
--- a/node/src/types/deploy.rs
+++ b/node/src/types/deploy.rs
@@ -689,6 +689,7 @@ impl Deploy {
     /// Returns true if and only if:
     ///   * the deploy hash is correct (should be the hash of the header), and
     ///   * the body hash is correct (should be the hash of the body), and
+    ///   * approvals are non empty, and
     ///   * all approvals are valid signatures of the deploy hash
     pub fn is_valid(&mut self) -> Result<(), DeployValidationFailure> {
         match self.is_valid.as_ref() {
--- a/node/src/types/deploy.rs
+++ b/node/src/types/deploy.rs
@@ -862,6 +862,7 @@ fn serialize_body(payment: &ExecutableDeployItem, session: &ExecutableDeployItem
 // asymmetric_key signing verification.
 fn validate_deploy(deploy: &Deploy) -> Result<(), DeployValidationFailure> {
     if deploy.approvals.is_empty() {
+        warn!(?deploy, "deploy has no approvals");
         return Err(DeployValidationFailure::EmptyApprovals);
     }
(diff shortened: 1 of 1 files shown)


rafal-ch (approved):
LGTM 👍 

One thing to consider, though. Depending on the current approach to testing, we might think about adding an integration test for a "deploy" with no approvals (similarly to tests in `preconditions.rs`).

Yet, this could be overkill.

node/src/types/deploy.rs
```diff
@@ -1123,6 +1130,15 @@ mod tests {
         check_is_not_valid(deploy, DeployValidationFailure::InvalidDeployHash);
     }
 
+    #[test]
```
darthsiroftardis, line 1132: @rafal-ch Is this what you meant by test for a deploy with no approvals

rafal-ch: Not exactly, sorry for not being clear enough :)

This is an unit tests that tests the validation on the type level. Meaning, it tests whether the `Deploy` can correctly validate itself. This is all good.

It, however, doesn't validate if the `is_valid()` function is at all used when the real deploy is being created. So the test I proposed to consider would have steps like:

1. setup and run test network
2. create deploy with no approvals
3. try to deploy the deploy :-)
4. observe an error

**But again, this could be too much of a hassle if we don't have a framework for such tests ready.**

PS. 
What I see the call chain to accept the deploy is the following:
* `DeployAcceptor::handle_event()`
* `DeployAcceptor::accept()`
* `Deploy::is_acceptable()`
* `Deploy::is_valid()`
and since I cannot locate any tests for `DeployAcceptor` I feel a bit anxious :)

rafal-ch:
Ah, also one thing that I missed, due to the following lines not being a part of the PR:
https://github.com/casper-network/casper-node/blob/0369311898cc7cf05593a4f4d96be732c801786e/node/src/types/deploy.rs#L689-L692

I think the comment is now a bit rotten and should also contain a bullet about approvals not being an empty set.

Fraser999 (approved):

PR #1894, opened 2021-08-11
Log backtrace when responder fails

dwerner:
Introduces a `backtrace::Backtrace` to logging when a responder fails. We're currently logging only:
```
could not send response to request down oneshot channel
```
Which does not provide any resolution as to where the responder was called from or what was happening when it failed.


marc-casperlabs (approved):
This seems like a very good idea to me, if the backtraces pan out. Certainly does not hurt, good work!

Fraser999 (approved):

PR #1895, opened 2021-08-11
Fixes to get nctl tests working again.

### LINKED ISSUE #1898: Fix NCTL For Fast-Sync
NCTL needs to be fixed for the `feat-fast-sync` branch, as well as compiling without the `casper-mainnet` flag.

afck:
* Remove `mem_pool_prune_interval`. That doesn't exist yet on this branch.
* Don't fail if the trusted block is in the era immediately after an emergency restart.
* Don't use the `casper-mainnet` feature: nctl compiles the client without it, and the node needs to be compatible with that.
* Make `HashingAlgorithmVersion::HASH_V2_PROTOCOL_VERSION` public also without the `casper-mainnet` feature, to fix the test build.

This doesn't fix `emergency_upgrade_test.sh` yet: That failure is related to how we handle emergency upgrades with fast sync in general, not just to #1666.

Closes #1898.

### DIFF
--- a/node/src/components/linear_chain_sync/operations.rs
+++ b/node/src/components/linear_chain_sync/operations.rs
@@ -429,7 +429,7 @@ where
     let maybe_last_emergency_restart_era_id = chainspec.protocol_config.last_emergency_restart;
     match maybe_last_emergency_restart_era_id {
         Some(last_emergency_restart_era)
-            if last_emergency_restart_era >= trusted_block_header.era_id() =>
+            if last_emergency_restart_era > trusted_block_header.era_id() =>
         {
             return Err(
                 LinearChainSyncError::TryingToJoinBeforeLastEmergencyRestartEra {
@@ -437,7 +437,7 @@ where
                     trusted_hash,
                     trusted_block_header,
                 },
-            )
+            );
         }
         _ => {}
     }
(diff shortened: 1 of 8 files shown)


node/src/components/linear_chain_sync/operations.rs
```diff
@@ -139,7 +139,7 @@ where
     let maybe_last_emergency_restart_era_id = chainspec.protocol_config.last_emergency_restart;
 
     let min_era = maybe_last_emergency_restart_era_id.unwrap_or_else(|| EraId::new(0));
-    if min_era >= trusted_header.era_id() {
+    if min_era > trusted_header.era_id() {
```
xcthulhu: This will be unreachable after we immediately create an switchblock after genesis/restart

xcthulhu: ...could we get rid of this now?

afck: And line 133 as well, right?

xcthulhu (approved):
LGTM, only change to consider is getting rid of this check: https://github.com/casper-network/casper-node/pull/1895/files/f05b9af9981c496ebe87cb937ea93090d5a67693#diff-be0c11c9a6fcff5e4229eae55fb52c24d0a59100e637a36eaeed02f1321394a9R142

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Yup!  Looks good to me...

xcthulhu (approved):

fizyk20 (approved):

PR #1901, opened 2021-08-12
Add backtrace to Responder

dwerner:
Followup to #1894 - but this time hoist the backtrace capture outside of the `async` context, allowing it to capture the call site.

Here's a playground showing how the output is improved:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=241f419c72dd01d5f0ab0aec70fce2a3

### DIFF
--- a/node/src/effect.rs
+++ b/node/src/effect.rs
@@ -176,17 +176,20 @@ impl<T: 'static + Send> Responder<T> {
 
 impl<T> Responder<T> {
     /// Send `data` to the origin of the request.
-    pub(crate) async fn respond(mut self, data: T) {
-        if let Some(sender) = self.0.take() {
-            if sender.send(data).is_err() {
-                let backtrace = backtrace::Backtrace::new();
-                error!(
-                    ?backtrace,
-                    "could not send response to request down oneshot channel"
-                );
+    pub(crate) fn respond(mut self, data: T) -> impl Future {
+        let caller_backtrace = backtrace::Backtrace::new();
+        async move {
+            if let Some(sender) = self.0.take() {
+                if sender.send(data).is_err() {
+                    error!(
+                        responder = ?self,
+                        ?caller_backtrace,
+                        "could not send response to request down oneshot channel"
+                    );
+                }
+            } else {
+                error!("tried to send a value down a responder channel, but it was already used");
             }
-        } else {
-            error!("tried to send a value down a responder channel, but it was already used");
         }
     }
 }
--- a/node/src/effect.rs
+++ b/node/src/effect.rs
@@ -176,7 +176,7 @@ impl<T: 'static + Send> Responder<T> {
 
 impl<T> Responder<T> {
     /// Send `data` to the origin of the request.
-    pub(crate) fn respond(mut self, data: T) -> impl Future {
+    pub(crate) fn respond(mut self, data: T) -> impl Future<Output = ()> {
         let caller_backtrace = backtrace::Backtrace::new();
         async move {
             if let Some(sender) = self.0.take() {
--- a/node/src/effect.rs
+++ b/node/src/effect.rs
@@ -188,7 +188,11 @@ impl<T> Responder<T> {
                     );
                 }
             } else {
-                error!("tried to send a value down a responder channel, but it was already used");
+                error!(
+                    responder = ?self,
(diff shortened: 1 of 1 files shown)


node/src/effect.rs
```diff
@@ -176,17 +176,20 @@ impl<T: 'static + Send> Responder<T> {
...
+            if let Some(sender) = self.0.take() {
+                if sender.send(data).is_err() {
+                    error!(
+                        responder = ?self,
+                        ?caller_backtrace,
+                        "could not send response to request down oneshot channel"
+                    );
+                }
+            } else {
+                error!("tried to send a value down a responder channel, but it was already used");
```
goral09: If we logged the type of `T` then it would already be giving us some ideas about what that channel was for.

node/src/effect.rs
```diff
@@ -176,17 +176,20 @@ impl<T: 'static + Send> Responder<T> {
...
-                );
+    pub(crate) fn respond(mut self, data: T) -> impl Future {
+        let caller_backtrace = backtrace::Backtrace::new();
+        async move {
+            if let Some(sender) = self.0.take() {
+                if sender.send(data).is_err() {
+                    error!(
+                        responder = ?self,
+                        ?caller_backtrace,
+                        "could not send response to request down oneshot channel"
```
goral09, line 187: And here, if we logged the type of `T` then it would most probably tell us who created the responder. Or at least give more clues.

dwerner: `T` here is logged through the responder's `Debug` impl below.

node/src/effect.rs
```diff
@@ -176,17 +176,20 @@ impl<T: 'static + Send> Responder<T> {
...
-        if let Some(sender) = self.0.take() {
-            if sender.send(data).is_err() {
-                let backtrace = backtrace::Backtrace::new();
-                error!(
-                    ?backtrace,
-                    "could not send response to request down oneshot channel"
-                );
+    pub(crate) fn respond(mut self, data: T) -> impl Future<Output = ()> {
+        let caller_backtrace = backtrace::Backtrace::new();
+        async move {
```
xcthulhu, line 181: We can just un `async` this

node/src/effect.rs
```diff
@@ -176,17 +176,20 @@ impl<T: 'static + Send> Responder<T> {
...
+            if let Some(sender) = self.0.take() {
+                if sender.send(data).is_err() {
+                    error!(
+                        responder = ?self,
```
xcthulhu, line 185: 👍

node/src/components/rpc_server.rs
```diff
@@ -271,44 +271,68 @@ where
...
+                main_responder.respond(result);
+                Effects::new()
+            }
             Event::GetBalanceResult {
                 result,
                 main_responder,
-            } => main_responder.respond(result).ignore(),
+            } => {
+                main_responder.respond(result);
+                Effects::new()
```
xcthulhu: Man did you put in a lot of work here.  Greatly appreciated! 😊

xcthulhu (approved):

node/src/components/storage.rs
```diff
@@ -531,71 +535,87 @@ impl Storage {
                     block.header().hash(),
                     block.body(),
                 )?;
-                responder.respond(true).ignore()
+                responder.respond(true);
+                return Ok(Effects::new());
```
xcthulhu:
```suggestion
                Effects::new()
```

node/src/components/fetcher.rs
```diff
@@ -142,30 +142,29 @@ pub(crate) trait ItemFetcher<T: Item + 'static> {
         result: Option<FetchResult<T, NodeId>>,
         peer: NodeId,
     ) -> Effects<Event<T>> {
```
xcthulhu:
```suggestion
    ) -> () {
```

node/src/effect.rs
dwerner: Turns out this is much harder than it would have appeared.

goral09: > Turns out this is much harder than it would have appeared.

I just want to add that, after having seen the pre-revert changes, I totally support this decision.

goral09 (approved):

PR #1903, opened 2021-08-12
Add `uptime` in status reponse for REST and RPC servers

### LINKED ISSUE #1716: Provide uptime as part of the node's status
@Fraser999 commented on [Mon Jul 19 2021](https://github.com/casper-network/Governance/issues/33)

The node's REST and JSON-RPC servers should update their status responses to include an `uptime_seconds` field, which will be a count of whole seconds since the node started.

Each server should have an additional member variable `node_start_instant` which is set via the constructor.  The same instant should be used for each server, which means that the REST server will need to provide a getter for the value so it can be retrieved in order to be passed in to the JSON-RPC server's constructor.

rafal-ch:
The REST and RPC requests for node status now contain the `uptime` value, like so:

```
{
   --- Some fields omitted ---
   "api_version": "1.0.0",
   "round_length": "4s 96ms",
   "uptime": "20s 576ms"
}
```

Closes #1716.

### DIFF
--- a/node/src/components/rest_server.rs
+++ b/node/src/components/rest_server.rs
@@ -85,7 +85,7 @@ pub(crate) struct RestServer {
     /// The task handle which will only join once the server loop has exited.
     #[data_size(skip)]
     server_join_handle: Option<JoinHandle<()>>,
-    /// The instant at which the node has started
+    /// The instant at which the node has started.
     node_startup_time: Instant,
 }
 
--- a/node/src/components/rest_server.rs
+++ b/node/src/components/rest_server.rs
@@ -86,7 +86,7 @@ pub(crate) struct RestServer {
     #[data_size(skip)]
     server_join_handle: Option<JoinHandle<()>>,
     /// The instant at which the node has started.
-    node_startup_time: Instant,
+    node_startup_instant: Instant,
 }
 
 impl RestServer {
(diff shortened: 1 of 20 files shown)


rafal-ch:
Converted back to draft as there will be quite a major rewrite.

node/src/components/rest_server.rs
```diff
@@ -85,13 +85,16 @@ pub(crate) struct RestServer {
     /// The task handle which will only join once the server loop has exited.
     #[data_size(skip)]
     server_join_handle: Option<JoinHandle<()>>,
+    /// The instant at which the node has started.
+    node_startup_time: Instant,
```
Fraser999: Can we rename this variable `node_startup_instant` throughout please?

Fraser999 (commented):
One minor nitpick remaining.

node/src/components/rest_server.rs
rafal-ch: Done.

nilz3ro (approved):

Fraser999 (approved):

goral09:
Shouldn't we be bumping API's minor version as we add new fields to it?

darthsiroftardis (approved):
LGTM!

Fraser999:
@goral09 
> Shouldn't we be bumping API's minor version as we add new fields to it?

Yes, but for now the servers' API versions are tied to the protocol version.  We're planning to move away from that in an upcoming PR, so it's out of scope for this one.

goral09:
> @goral09
> 
> > Shouldn't we be bumping API's minor version as we add new fields to it?
> 
> Yes, but for now the servers' API versions are tied to the protocol version. We're planning to move away from that in an upcoming PR, so it's out of scope for this one.

Thanks. I was confused with the example that shows `"api_version": 1.0.0.` but now that I think of it, it's probably b/c we don't update the example chainspec in the repository.

PR #1908, opened 2021-08-12
Enforce contract runtime concurrency limit

### LINKED ISSUE #1899: Network stalls for few min with load of 12K transfers
network stalled for a few mins when loaded with 12k transfers - with Error/Warn : "could not send response to request down oneshot channel event took very long to dispatch".
No issues with WASM transfers.

Details:
https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1628649696179300

dwerner:
Resolves #1899 

This PR introduces a number of tweaks to the `ContractRuntime`'s concurrency. The thinking is that the current thread ends up blocking the reactor, when many requests run simultaneously with the HTTP server producing many events.

Changes:

- All calls to `execute_finalized_block` are executed in a new `async` task. In order to do this, `ExecutionPreState` and the `exec_queue` need to be shared to the task.
- Calls to `execute_finalized_block` call into LMDB code. This PR introduces using `tokio::task::unconstrained` to prevent IO with LMDB from being continually interrupted by the `tokio` scheduler.

Empirically, I am able to reproduce the long long stalls on `dev` by hammering a node with transfers and simultaneously curl-ing the metrics endpoint

### Results

```
$~/D/casper-node (dev)> time curl -s localhost:14101/metrics | grep pending
________________________________________________________
Executed in    8.43 secs   fish           external 
   usr time    2.56 millis  1035.00 micros    1.53 millis 
   sys time    8.97 millis  460.00 micros    8.51 millis 

$~/D/casper-node (dev) [SIGINT|SIGINT]> time bash -c "curl -s localhost:14101/metrics | grep pending'
                                                  ^C
$~/D/casper-node (dev) [SIGINT|SIGINT]> time bash -c "curl -s localhost:14101/metrics | grep pending"
# HELP pending_deploy amount of pending deploys
# TYPE pending_deploy gauge
pending_deploy 2053

________________________________________________________
Executed in    5.16 secs   fish           external 
   usr time   12.72 millis    0.00 micros   12.72 millis 
   sys time    1.29 millis  912.00 micros    0.38 millis 

$~/D/casper-node (dev)> time bash -c "curl -s localhost:14101/metrics | grep pending"
# HELP pending_deploy amount of pending deploys
# TYPE pending_deploy gauge
pending_deploy 1601

________________________________________________________
Executed in    3.13 secs   fish           external 
   usr time    7.80 millis    0.00 micros    7.80 millis 
   sys time    5.12 millis  803.00 micros    4.32 millis 
```

With the changes in this PR, I am unable to reproduce a long stall.

```
$~/D/casper-node (contract-runtime-concurrency-limit)> time bash -c "curl -s localhost:14101/metrics | grep pending"
# HELP pending_deploy amount of pending deploys
# TYPE pending_deploy gauge
pending_deploy 5316

________________________________________________________
Executed in   21.41 millis    fish           external 
   usr time    8.51 millis  823.00 micros    7.68 millis 
   sys time    2.00 millis    0.00 micros    2.00 millis 

$~/D/casper-node (contract-runtime-concurrency-limit)> time bash -c "curl -s localhost:14101/metrics | grep pending"
# HELP pending_deploy amount of pending deploys
# TYPE pending_deploy gauge
pending_deploy 0

________________________________________________________
Executed in   11.49 millis    fish           external 
   usr time   10.10 millis  846.00 micros    9.25 millis 
   sys time    2.21 millis    0.00 micros    2.21 millis 

$~/D/casper-node (contract-runtime-concurrency-limit)> time bash -c "curl -s localhost:14101/metrics | grep pending"
# HELP pending_deploy amount of pending deploys
# TYPE pending_deploy gauge
pending_deploy 0
```


node/src/components/contract_runtime/operations.rs
```diff
@@ -29,14 +29,19 @@ use crate::{
     types::{Block, Deploy, DeployHash, DeployHeader, FinalizedBlock},
 };
 
-pub(super) fn execute_finalized_block(
+pub(super) async fn execute_finalized_block(
```
goral09: This method  doesn't have to be `async`. If you want to call it in the async context it should be lifted to an async one on the call-side rather than forcing all callers to handle it as async (even though it doesn't have to be).

node/src/components/contract_runtime.rs
```diff
@@ -101,16 +101,18 @@ impl From<&BlockHeader> for ExecutionPreState {
     }
 }
 
+type ExecQueue = Arc<Mutex<BTreeMap<u64, (FinalizedBlock, Vec<Deploy>)>>>;
```
xcthulhu, line 104: As discussed, I suggest we move this queue to `LinearChain`.

Moreover, it's not necessary for it to have its deploys, we can just get those from storage.

node/src/components/contract_runtime.rs
```diff
@@ -548,65 +567,67 @@ impl ContractRuntime {
...
+            engine_state.as_ref(),
+            metrics.as_ref(),
             protocol_version,
-            execution_pre_state,
+            current_execution_pre_state,
             finalized_block,
             deploys,
-        ) {
+        ))
+        .await
```
xcthulhu, line 604: To elaborate a little on @goral09 's point [below][1], we could have this be `async move { operations::execute_finalized_block(...) }` right?

[1]: https://github.com/casper-network/casper-node/pull/1908/files#r688164423

node/src/components/contract_runtime.rs
```diff
@@ -548,65 +567,67 @@ impl ContractRuntime {
...
         deploys: Vec<Deploy>,
-    ) -> Effects<ContractRuntimeRequest>
-    where
+    ) where
         REv: From<ContractRuntimeRequest>
             + From<ContractRuntimeAnnouncement>
             + From<ControlAnnouncement>
             + Send,
     {
+        let current_execution_pre_state = { execution_pre_state.lock().unwrap().clone() };
```
xcthulhu:
```suggestion
        let current_execution_pre_state = execution_pre_state.lock().unwrap().clone();
```

node/src/components/contract_runtime.rs
goral09: `LinearChain` deals with blocks that are already in ... linear chain. A `FinalizedBlock` is not yet a linear chain block - it's in a different bounded context, it's an internal data structure that has been only finalized by consensus and has to be executed and only then transformed into a linear chain block. This queue is in my opinion in a proper place - we used to have a `BlockExecutor` object, where that queue would fit better, but it has been merged with `ContractRuntime`.

node/src/components/contract_runtime/operations.rs
dwerner: Thanks, an earlier version had required it be async, but you are correct.

node/src/components/contract_runtime.rs
xcthulhu: You are right. I agree we should not couple anything unnecessary to the `LinearChain` component.  The issue with this queue is that it is not used in the `ContractRuntime` component when we are fast-syncing.

@dwerner and I have discussed this and we are thinking now it should become task-local state (in fast-sync the `ExecutionPreState` is similarly task-local).  Planning on doing this in follow-up work.

node/src/components/contract_runtime.rs
```diff
@@ -425,27 +427,43 @@ where
...
-                        self.execution_pre_state.clone(),
-                        finalized_block,
-                        deploys,
+                let mut effects = Effects::new();
+                let engine_state = Arc::clone(&self.engine_state);
+                let metrics = Arc::clone(&self.metrics);
+                let exec_queue = Arc::clone(&self.exec_queue);
+                let execution_pre_state = Arc::clone(&self.execution_pre_state);
+                let protocol_version = self.protocol_version;
+                if self.execution_pre_state.lock().unwrap().next_block_height
```
xcthulhu, line 444: We should probably `fatal!` here...

xcthulhu:
Looking good! I say we slap in `fatal!`s instead of those unwraps and this is good to go.

node/src/components/contract_runtime.rs
goral09: I think it did happen to us in the past that two _finalized events_ `A1, A2` arrived in the reversed order `A2, A1`. This ensures that we recover from that situation correctly. Also, Dan is not changing the behavior that was already here.

node/src/components/contract_runtime.rs
```diff
@@ -548,65 +567,70 @@ impl ContractRuntime {
...
+        .await
+        {
             Ok(block_and_execution_effects) => block_and_execution_effects,
-            Err(error) => return fatal!(effect_builder, "{}", error).ignore(),
+            Err(error) => return fatal!(effect_builder, "{}", error).await,
         };
 
-        self.execution_pre_state = ExecutionPreState::from(block.header());
+        let new_execution_pre_state = ExecutionPreState::from(block.header());
+        *execution_pre_state.lock().unwrap() = new_execution_pre_state.clone();
```
goral09, line 611: This looks like `set_initial_state`?

goral09 (approved):
Looks good. We need to do some more fuzzy/under load testing, just to make sure that the two `Arc<Mutex<...>>` don't turn around and bite us in unexpected ways.

node/src/components/contract_runtime.rs
dwerner: That's true that it looks similar, but here we are in an `async` task that has a different lifetime, and `set_initial_state` is a method on the component. We can't use that here.

node/src/components/contract_runtime.rs
dwerner: It's considered "normal" to unwrap a locked `std::sync::Mutex`, because the only time it will panic is if another thread panicked while holding the Mutex. Therefore if we handle other errors within the lifetime of the `MutexGuard`, this cannot occur. I'm happy to revisit but I think this probably obviated by a follow-up we're already planning.

xcthulhu (approved):

xcthulhu:
Talked about the `unwrap()`s; since this is a regression it is better to have this solution in place and remove the `Mutex` in a subsequent PR.

PR #1909, opened 2021-08-13
Network next era validators

marc-casperlabs:
This PR builds upon the work from https://github.com/xcthulhu/casper-node/tree/network-next-era-validators , making the necessary changes to the networking component to consume the new announcement for validator changes.

Some notes:

* `get_era_validators` lives on for now, as it is also used in the `EraSupervisor`.
* There are none of the new announcements in the joiner reactor, which I assume is intended.
* Like previous approaches, the code is written with guards against out-of-order events and other potential issues. The highest era seen "wins out".
* Tests for this specific behavior are currently missing, as the required integration test setup with multiple validator set changes is currently hard to implement in the Rust tests (historically SRE has handled this instead).


execution_engine/src/core/engine_state/mod.rs
```diff
@@ -2052,29 +2052,8 @@ where
...
             .map_err(Into::into)?;
 
-        let next_era_validators = {
-            let mut era_validators = match self.get_era_validators(
-                correlation_id,
-                GetEraValidatorsRequest::new(post_state_hash, step_request.protocol_version),
-            ) {
-                Ok(era_validators) => era_validators,
-                Err(error) => {
-                    return Err(StepError::GetEraValidatorsError(error));
```
xcthulhu, line 2060: Sorry for not catching this earlier, but we can get rid of this error variant now

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -2052,29 +2052,8 @@ where
...
-                }
-            };
-
-            let era_id = &step_request.next_era_id;
-            match era_validators.remove(era_id) {
-                Some(validator_weights) => validator_weights,
-                None => {
-                    return Err(StepError::EraValidatorsMissing(*era_id));
```
xcthulhu, line 2068: We can also get rid fo this error variant

node/src/components/contract_runtime.rs
```diff
@@ -518,9 +518,9 @@ impl ContractRuntime {
     }
 
     /// Commits a genesis using a chainspec
-    fn commit_genesis(
+    pub(crate) fn commit_genesis(
         &self,
-        chainspec: Arc<Chainspec>,
+        chainspec: &Arc<Chainspec>,
```
xcthulhu: Maybe just `&Chainspec`?

node/src/components/contract_runtime.rs
```diff
@@ -611,7 +621,7 @@ impl ContractRuntime {
 
     /// Read a [Trie<Key, StoredValue>] from the trie store.
     pub(crate) fn read_trie(
-        &mut self,
+        &self,
```
xcthulhu, line 647: Good catch!

node/src/components/small_network.rs
```diff
@@ -186,17 +177,19 @@ where
     /// This is not incoming bandwidth but an independent resource estimate.
     #[data_size(skip)]
     incoming_limiter: Box<dyn Limiter>,
+
+    /// The era that is considered the active era by the small network component.
+    active_era: EraId,
+
+    /// The validator set associated with the `active_era`.
+    active_era_validators: HashSet<PublicKey>,
```
xcthulhu: ~So, we might want two of these.~

~Specifically `active_era_validators` and `next_era_validators` (which are the validators for `active_era` + 1).~

~I did suggest having the `last_era_validators` (for `active_era` - 1).  But after some thought, I don't think they are necessary.~

~WDYT?~

~It doesn't matter *too* much in practice, since validators don't rotate very much, and this is KISS which I approve.~

[EDIT: Now that I read the code [below][1], the limiters are already doing this bookkeeping. We can just get rid of `active_era_validators`, right?]

[1]: https://github.com/casper-network/casper-node/pull/1909/files#r688460341

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+                ContractRuntimeAnnouncement::LinearChainBlock(_)
+                | ContractRuntimeAnnouncement::StepSuccess { .. },
+            ) => Effects::new(),
+            Event::ContractRuntimeAnnouncement(
+                ContractRuntimeAnnouncement::UpcomingEraValidators {
+                    era_that_is_ending,
+                    mut upcoming_era_validators,
+                },
+            ) => {
+                if era_that_is_ending < self.active_era {
```
xcthulhu, line 897:
```suggestion
                if era_that_is_ending =< self.active_era {
```

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+                    debug!("ignoring past era end announcement");
+                } else if era_that_is_ending != self.active_era {
```
xcthulhu:
```suggestion
                } else if era_that_is_ending >= self.active_era {
```

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+                    debug!("received future era end announcement, clearing limiters");
+                    self.active_era = era_that_is_ending + 1;
```
xcthulhu, line 901:
```suggestion
                    self.active_era = era_that_is_ending;
```

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+                    self.active_era_validators = Default::default();
+
+                    self.incoming_limiter
+                        .update_validators(Default::default(), Default::default());
+                    self.outgoing_limiter
+                        .update_validators(Default::default(), Default::default());
+                } else {
+                    // At this point we can rely on the `active_era` being the one that is ending.
+                    self.active_era = era_that_is_ending + 1;
```
xcthulhu:
```suggestion
```

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+                    }
+
+                    active_validators.extend(self.active_era_validators.iter().cloned());
+
+                    let upcoming_validators: HashSet<PublicKey> = upcoming_era_validators
+                        .remove(&(self.active_era + 1))
+                        .unwrap_or_default()
+                        .into_keys()
+                        .collect();
+
```
xcthulhu:
```suggestion
                    // Initialize the next set of active validators with the currently active one.
                    let mut active_validators: HashSet<PublicKey> =
                        mem::take(&mut self.active_era_validators);

                    self.active_era_validators = upcoming_era_validators
                        .remove(&self.active_era)
                        .unwrap_or_default()
                        .into_keys()
                        .collect();

                    if self.active_era_validators.is_empty() {
                        warn!("received an empty set of active era validators");
                    }

                    let active_validators = upcoming_era_validators
                        .remove(&(era_that_is_ending + 1))
                        .unwrap_or_default()
                        .into_keys()
                        .collect();;

                    let upcoming_validators: HashSet<PublicKey> = upcoming_era_validators
                        .remove(&(era_that_is_ending + 2))
                        .unwrap_or_default()
                        .into_keys()
                        .collect();
```

Looks like we don't need `self.active_validators`...

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+            Event::ContractRuntimeAnnouncement(
+                ContractRuntimeAnnouncement::UpcomingEraValidators {
+                    era_that_is_ending,
+                    mut upcoming_era_validators,
+                },
+            ) => {
+                if era_that_is_ending < self.active_era {
+                    debug!("ignoring past era end announcement");
+                } else if era_that_is_ending != self.active_era {
+                    debug!("received future era end announcement, clearing limiters");
```
xcthulhu:
```suggestion
                    debug!("received upcoming era validators announcement, clearing limiters");
```

node/src/components/small_network.rs
xcthulhu: Less confusing to me (and maybe others?) if we set this *after* getting the active/upcoming era validators from the map.

node/src/components/small_network.rs
marc-casperlabs: The limiters do not make the assumption that they are always going to be called "in order", so the cannot rely on the previous set of validators being the ones right before the new set. There would have to be more logic in the limiters to handle this and/or an active era tracked, but I would rather keep them "dumb".

node/src/components/small_network.rs
marc-casperlabs: The if else if else branch here looks like (or should look like)

```
IF ending era is past era
   ignore
ELSE IF ending era is current era
   do regular update
ELSE (implies "ending era is future era")
   clear all limiters, as we cannot rely on the ending era being accurate
   and only use the information in we just received
```

The proposed change at the very least makes the `else` block at the dead code, if I am not mistaken (since one of `era_that_is_ending < self.active_era` and `era_that_is_ending >= self.active_era` should always be true), but I believe we need this "special case that is unlikely to happen" code.

node/src/components/small_network.rs
marc-casperlabs: There is no `self.active_validators`, only `self.active_era_validators` and `active_validators`.

`self.active_era_validators` keeps the validators of the currently active era, because we want to include them in the preferred peers set for one era after their active era ends. This, of course, assumes that they are not included in the announcement.

`active_validators` is the union of this previous set of validators and the very-soon-to-be-actually-active validators from the announcement and is passed to the limiters.

I believe we need both.

node/src/components/small_network.rs
marc-casperlabs: I'd need a good name for a variable `era_that_is_ending + 1` then, do you have a suggestion?

node/src/components/contract_runtime/announcements.rs
```diff
@@ -0,0 +1,125 @@
...
+    UpcomingEraValidators {
+        /// The era id in which the step was committed to global state.
+        era_that_is_ending: EraId,
+        /// The validators for the eras after the `era_that_is_ending` era.
+        upcoming_era_validators: BTreeMap<EraId, BTreeMap<PublicKey, U512>>,
+    },
+}
+
+impl ContractRuntimeAnnouncement {
+    /// Create a ContractRuntimeAnnouncement::LinearChainBlock from it's parts.
```
afck:
```suggestion
    /// Creates a `ContractRuntimeAnnouncement::LinearChainBlock` from its parts.
```

node/src/components/contract_runtime/announcements.rs
```diff
@@ -0,0 +1,125 @@
...
+            }
+            ContractRuntimeAnnouncement::StepSuccess { era_id, .. } => {
+                write!(f, "step completed for {}", era_id)
+            }
+            ContractRuntimeAnnouncement::UpcomingEraValidators {
+                era_that_is_ending, ..
+            } => {
+                write!(
+                    f,
+                    "upcoming era validators after current era {}.",
```
afck:
```suggestion
                    "upcoming era validators after current {}.",
```
I think that will print: "`current era era 5`" otherwise.

node/src/components/contract_runtime/announcements.rs
```diff
@@ -0,0 +1,125 @@
...
+                    era_that_is_ending,
+                )
+            }
+        }
+    }
+}
+
+/// Announce a committed Step success.
```
afck:
```suggestion
/// Announces a committed Step success.
```
(Also below.)

node/src/components/contract_runtime/operations.rs
```diff
@@ -15,19 +15,23 @@ use casper_execution_engine::{
...
+            BlockAndExecutionEffects, ContractRuntimeMetrics, ExecutionPreState,
         },
     },
     crypto::hash::Digest,
     types::{Block, Deploy, DeployHash, DeployHeader, FinalizedBlock},
 };
+use casper_execution_engine::{
+    core::engine_state::GetEraValidatorsRequest, shared::newtypes::Blake2bHash,
+};
+use std::collections::BTreeMap;
```
afck: Should move up to the other `std` imports.

node/src/components/contract_runtime/operations.rs
```diff
@@ -80,57 +84,66 @@ pub(super) fn execute_finalized_block(
...
+    let next_era_validator_weights: Option<BTreeMap<PublicKey, U512>> =
+        maybe_step_effect_and_upcoming_era_validators
+            .as_ref()
+            .and_then(
+                |StepEffectAndUpcomingEraValidators {
+                     upcoming_era_validators,
+                     ..
+                 }| {
+                    upcoming_era_validators
+                        .get(&finalized_block.era_id().successor())
```
afck, line 128: This must be `Some`.
We should log and/or return an error if it's `None`.

node/src/components/contract_runtime/types.rs
```diff
@@ -81,6 +81,15 @@ impl From<EraValidatorsRequest> for GetEraValidatorsRequest {
     }
 }
 
+/// Effects from running step and the next era validators that are gathered when an era ends.
+#[derive(Debug)]
+pub struct StepEffectAndUpcomingEraValidators {
+    /// An [`ExecutionEffect`] created by an era ending.
+    pub step_execution_effect: ExecutionEffect,
+    /// Upcoming era validators.
```
afck: We could explain in a bit more detail here which validators exactly are included, e.g.:

```suggestion
    /// Validator sets for all upcoming eras that have already been determined.
```

node/src/components/contract_runtime/types.rs
```diff
@@ -90,5 +99,5 @@ pub struct BlockAndExecutionEffects {
     /// The results from executing the deploys in the block.
     pub execution_results: HashMap<DeployHash, (DeployHeader, ExecutionResult)>,
     /// An [`ExecutionEffect`] created if an era ended.
```
afck:
```suggestion
    /// The [`ExecutionEffect`] created and the upcoming validator sets determined by the `step`
    /// function, if this was a switch block.
```

node/src/components/small_network.rs
```diff
@@ -335,11 +329,7 @@ where
             .collect();
 
         // Initialize the known validator set with the active era, if given.
```
afck: I think that's outdated now?

```suggestion
```

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+                        mem::take(&mut self.active_era_validators);
+
+                    self.active_era_validators = upcoming_era_validators
+                        .remove(&self.active_era)
+                        .unwrap_or_default()
+                        .into_keys()
+                        .collect();
+
+                    if self.active_era_validators.is_empty() {
+                        warn!("received an empty set of active era validators");
```
afck: I'd make this error-level.

node/src/components/small_network.rs
```diff
@@ -953,6 +884,67 @@ where
...
+                        "updating active and upcoming validators"
+                    );
+                    self.incoming_limiter
+                        .update_validators(active_validators.clone(), upcoming_validators.clone());
+                    self.outgoing_limiter
+                        .update_validators(active_validators, upcoming_validators);
+                }
+
+                Effects::new()
+            }
```
afck, line 928: This is the part that will be simplified as discussed, is it? We'll just prioritize the union of the `upcoming_era_validators` values and that's it?

node/src/components/small_network.rs
marc-casperlabs: This is now obsolete, as the code has been rewritten in c5720c575

node/src/components/small_network.rs
marc-casperlabs: Obsolete with c5720c575

node/src/components/small_network.rs
marc-casperlabs: Obsolete with c5720c575

node/src/components/small_network.rs
marc-casperlabs: Yes, see c5720c575

node/src/components/small_network.rs
marc-casperlabs: Removed in c5720c5

node/src/components/small_network.rs
marc-casperlabs: Obsolete, as the code has been rewritten in c5720c5

afck (approved):
Looks good to me, but I really think we should log an error [here](https://github.com/casper-network/casper-node/pull/1909/files#r689546048).

node/src/components/contract_runtime/operations.rs
marc-casperlabs: I will put this us as a different change, because implementing it as described would cause a change in behavior - before, this function silently minted switch blocks with no new validator sets, after this change it will return an error instead. This is a change too big to tack onto this somewhat unrelated PR.

PR #1918, opened 2021-08-14
Introducing better `pull_request_template.md`

xcthulhu:
The following template provides a few tweaks to what we already have:

 - Leverages github's [task-lists][1] feature
 - Comments out the advice by default
 - Removes a run-on sentence

[1]: https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,5 +3,5 @@
 
 # Please consider the following when creating a PR:
 # 
-# * Provide a description of this PR in simple language. This will help reviewers and contributers.
+# * Provide a list of significant changes made in this PR. This will help reviewers and contributors.
 # * Identify if any downstream impact as in to, SDKs, SmartContracts etc
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,7 @@
-Please consider the following when creating a PR:
+- [ ] Closes <ISSUE>
+- [ ] Updated CHANGELOG.md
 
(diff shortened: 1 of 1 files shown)


xcthulhu:
I did not update the `CHANGELOG.md` nor did I link an issue for this.  I'm not sure it is necessary given an essentially cosmetic change.

Fraser999 (commented):
I don't personally think this is an improvement.

Maybe if you could say what is lacking about the current template we could iterate on it?

xcthulhu:
The two things missing from the current template are:

1. A checklist of things the submitter should take care of
2. Commented out advice so the submitter doesn't have to delete it; that way it can keep reminding them of best practices while they update their PR.

.github/pull_request_template.md
```diff
@@ -1,6 +1,7 @@
...
+- [ ] Closes <ISSUE>
+- [ ] Updated CHANGELOG.md
 
-* Provide a useful description of this PR, suitably verbose, aimed at helping reviewers and contributors
-* Update all relevant changelogs
-* Provide a link to the GitHub issue relating to this PR
-* Identify if any downstream impact as in to, SDKs, SmartContracts etc
+# Please consider the following when creating a PR:
+# 
+# * Provide a description of this PR in simple language. This will help reviewers and contributers.
```
xcthulhu: Chatting with @darthsiroftardis , he had the following feedback for the line above:

```suggestion
# * Provide a list of significant changes made in this PR. This will help reviewers and contributors.
```

PR #1921, opened 2021-08-16
Make run_equivocator_network test simpler and less flaky.

### LINKED ISSUE #1859: Non-det failure of `run_equivocator_network` test.

afck:
This wraps the `Participating` reactor in a `FilterReactor` in the test, and delays the first few messages to and from one of the two nodes with the same key. That way, the nodes won't notice each other and will always equivocate in the first era.

Closes #1859

### DIFF
--- a/node/src/reactor/participating/tests.rs
+++ b/node/src/reactor/participating/tests.rs
@@ -1,11 +1,13 @@
 use std::{collections::BTreeMap, sync::Arc, time::Duration};
 
 use anyhow::bail;
+use either::Either;
 use log::info;
 use num::Zero;
 use num_rational::Ratio;
 use rand::Rng;
 use tempfile::TempDir;
+use tokio::time;
 
 use casper_execution_engine::{core::engine_state::query::GetBidsRequest, shared::motes::Motes};
 use casper_types::{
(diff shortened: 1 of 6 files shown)


goral09 (commented):
The test now looks quite nice. I wonder, if we can write a unit test now that asserts when the bid can be re-activated, what's the delay for that, that activating bid while you're evicted is a noop, etc.

fizyk20 (approved):
LGTM :+1: And nice one with the `FilterReactor` :slightly_smiling_face:

afck:
> I wonder, if we can write a unit test now that asserts when the bid can be re-activated, what's the delay for that, that activating bid while you're evicted is a noop, etc.

Makes sense, but let's make that a separate ticket.

afck:
I created #1922 for extending the test.

node/src/reactor/participating/tests.rs
```diff
@@ -240,110 +259,100 @@ async fn run_equivocator_network() {
...
+                next_era_validators.push(
+                    header
+                        .next_era_validator_weights()
+                        .expect("validators")
+                        .clone(),
+                );
                 switch_blocks.push(header);
             }
         }
+    }
```
goral09: This piece could be a separate method returning new data structure(s).

goral09 (approved):

node/src/reactor/participating/tests.rs
```diff
@@ -240,110 +259,100 @@ async fn run_equivocator_network() {
...
+    let min_round_len = chain.chainspec.highway_config.min_round_length();
+    net.reactors_mut()
+        .find(|reactor| *reactor.inner().consensus().public_key() == alice_pk)
+        .unwrap()
+        .set_filter(move |event| match event {
+            ParticipatingEvent::NetworkRequest(_)
+            | ParticipatingEvent::Consensus(consensus::Event::MessageReceived { .. })
+                if Timestamp::now() < genesis_time + min_round_len * 2 =>
+            {
+                Either::Left(time::sleep(min_round_len.into()).event(move |_| event))
```
xcthulhu, line 332: Looks like this is where we introduce the delay...

xcthulhu (approved):
LGTM

afck:
bors merge

PR #1924, opened 2021-08-16
Remove libp2p

### LINKED ISSUE #1641: Remove libp2p

marc-casperlabs:
This PR removes libp2p and simplifies types like `NodeId` that have been introduced to support both networking implementations. It is still kept as a newtype though.

As a result, we can remove over 3000 LoC and 20-25% of our dependencies and build times have been improved by 13% for release builds. Since the libp2p was not maintained for some time now, there should be no negative impact on the system as a whole.

Metrics only used by libp2p-networking are removed as well. Additionally, a configuration option has been removed that had not been in use for a while, this is mentioned in the changelog.

Closes #1641.


Fraser999 (approved):
Very happy to see this going!

PR #1900, opened 2021-08-12
Check Summed Hex

nilz3ro:
# About These Changes
* Fixes broken unit tests from the original draft PR.
* Removes references to third party hex crates (except in test).
* Brings #1557 up to speed with dev.
* For more detail, check comprehensive issue #1415.

PR #1557, opened 2021-06-09
CEP-57: Introducing Check-sum encoding for hex representations

xcthulhu:
Implements https://github.com/casper-network/ceps/pull/57

Closes #1415

Issue #1415, opened 2021-05-10
Checksum for target account hash


piotr-dziubecki:
CEP equivalent of EIP-55

PR #1925, opened 2021-08-16
Check Summed Hex

nilz3ro:
# About These Changes
* Replaces #1900.
* Brings #1557 up to speed with dev.
* For more info, see comprehensive issue #1415


client/lib/deploy.rs
```diff
@@ -510,6 +587,16 @@ mod tests {
         assert!(matches!(Deploy::read_deploy(bytes), Ok(_)));
     }
 
+    #[test]
+    fn should_round_trip_check_summed_sample_deploy_json() {
+        let bytes = CHECK_SUMMED_SAMPLE_DEPLOY.as_bytes();
+        let reader = BufReader::new(bytes);
+        let deploy: Deploy = serde_json::from_reader(reader).unwrap();
```
afck: Why not just `from_str`?

```suggestion
        let deploy: Deploy = serde_json::from_str(CHECK_SUMMED_SAMPLE_DEPLOY).unwrap();
```

client/tests/integration_test.rs
```diff
@@ -403,9 +402,11 @@ mod get_balance {
             server_handle.get_balance("", "").await,
             Err(Error::CryptoError {
                 context: "state_root_hash",
-                error: CryptoError::FromHex(FromHexError::InvalidStringLength)
+                error: CryptoError::DigestMustBe32Bytes {
+                    actual_byte_length: 0
+                }
             })
-        ));
+        ))
```
afck: The other asserts do have a semicolon.

execution_engine/src/shared/newtypes/blake2b256.rs
```diff
@@ -116,10 +98,16 @@ impl FromBytes for Blake2bHash {
...
+    fn create_bytes(value: &Blake2bHash) -> Cow<[u8]> {
+        Cow::from(value.to_bytes().unwrap())
+    }
 
-    fn from_hex<T: AsRef<[u8]>>(hex: T) -> Result<Self, Self::Error> {
-        Ok(Blake2bHash(hex::FromHex::from_hex(hex)?))
+    fn from_bytes(bytes: &[u8]) -> Result<Blake2bHash, Self::Error> {
+        FromBytes::from_bytes(bytes)
+            .map(|(b, _rem)| b)
+            .map_err(|_error| "replace me with real error.")
```
afck: If we merge the above `TODO` into `dev`, let's at least replace the message here.

node/src/components/consensus.rs
```diff
@@ -156,7 +155,12 @@ impl Display for ConsensusMessage {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         match self {
             ConsensusMessage::Protocol { era_id, payload } => {
-                write!(f, "protocol message {:10} in {}", HexFmt(payload), era_id)
+                write!(
+                    f,
+                    "protocol message {:10} in {}",
+                    check_summed_hex::encode(payload),
```
afck: I wonder whether for potentially large data we should avoid computing the whole checksummed hex representation, just to truncate it to 10 characters?

node/src/crypto/hash.rs
```diff
@@ -67,9 +68,12 @@ impl Digest {
 
     /// Returns a `Digest` parsed from a hex-encoded `Digest`.
     pub fn from_hex<T: AsRef<[u8]>>(hex_input: T) -> Result<Self, Error> {
-        let mut inner = [0; Digest::LENGTH];
-        hex::decode_to_slice(hex_input, &mut inner)?;
-        Ok(Digest(inner))
+        let bytes = check_summed_hex::decode(&hex_input)?;
```
afck: We could also check the length before `decode`, so that if someone sends us terabytes of data we don't bother decoding them.

node/src/protocol.rs
```diff
@@ -121,15 +121,18 @@ impl Debug for Message {
...
             Message::GetResponse {
                 tag,
                 serialized_item,
             } => f
                 .debug_struct("GetResponse")
                 .field("tag", tag)
-                .field("serialized_item", &HexFmt(serialized_item))
+                .field(
+                    "serialized_item",
+                    &check_summed_hex::encode(serialized_item),
```
afck: Some item types can be large; I think it's wasteful to checksum all that data. Or even to log e.g. a full hex-encoded deploy body at all.

types/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "casper-types"
-version = "1.0.0" # when updating, also update 'html_root_url' in lib.rs
+version = "1.0.0"                                                                                       # when updating, also update 'html_root_url' in lib.rs
```
afck:
```suggestion
version = "1.0.0" # when updating, also update 'html_root_url' in lib.rs
```

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+/// Takes a slice of bytes and outputs an infinite cyclic stream of bits for those bytes.
+fn bytes_to_bits_cycle(bytes: Vec<u8>) -> impl Iterator<Item = bool> {
+    bytes
+        .into_iter()
+        .cycle()
+        .flat_map(move |byte| (0..8usize).map(move |offset| ((byte >> offset) & 0x01) == 0x01))
+}
+
+/// Computes a Blake2b hash outputs it as a vector of *nibbles* (ie, 4-bit values) represented as
+/// `usize`s.
```
afck: I don't think it does?

```suggestion
/// Computes a Blake2b hash.
```

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+            deserializer.deserialize_str(CheckSummedHexVisitor)
+        } else {
+            deserializer.deserialize_bytes(BytesVisitor)
+        };
+        maybe_bytes.and_then(|bytes| Self::from_bytes(&bytes).map_err(D::Error::custom))
+    }
+}
+
+/// A dummy container for use inside `#[serde(with)]` attribute if the underlying type
+/// implements [`CheckSummedHex`].
```
afck:
```suggestion
/// implements `AsRef<[u8]>` and `TryFrom<&[u8], _>`.
```

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+            let err = serde_json::from_value::<Test>(bogus_json)
+                .unwrap_err()
+                .to_string();
+            assert!(
+                err.contains("expected check-summed hex-encoded byte array"),
+                "{}",
+                err
+            );
+        }
+    }
```
afck: We should add a test case where the actual checksum doesn't match.

node/src/types/block.rs
```diff
@@ -261,8 +259,8 @@ impl Display for BlockPayload {
         write!(
             formatter,
             "block payload: deploys {}, transfers {}, accusations {:?}, random bit {}",
-            HexList(&self.deploy_hashes),
-            HexList(&self.transfer_hashes),
+            format!("[{}]", itertools::join(&self.deploy_hashes, ", ")),
+            format!("[{}]", itertools::join(&self.transfer_hashes, ", ")),
```
afck: If I'm reading this correctly, this will now print `deploy-hash(01234567..)` instead of just `01234567..`. I'd prefer keeping the logs as concise as possible.

afck (commented):
I'm not sure it's a good idea to apply the checksumming in _all_ those places. We should at least make sure that checksumming large binary blobs like Wasm doesn't hurt performance.

(And I'd actually spell it "checksum" everywhere, not "check sum", e.g. `checksummed_hex`. But it's not important and I should have thought of that earlier.)

execution_engine/src/shared/newtypes/blake2b256.rs
xcthulhu: I pretty sure bytesrepr is not appropriate here… (it’s not how the old code worked…?)

node/src/crypto/hash.rs
```diff
@@ -183,15 +168,15 @@ mod test {
         let inputs_and_digests = [
             (
                 "",
-                "0e5751c026e543b2e8ab2eb06099daa1d1e5df47778f7787faab45cdf12fe3a8",
+                "0E5751C026E543B2E8aB2eb06099dAA1d1e5dF47778F7787fAAB45Cdf12fE3A8",
```
xcthulhu: These test vectors should work identically for Blake2bHash and Digest.  We probably should just get rid of one of the since they are the same thing…

node/src/protocol.rs
xcthulhu: I’ve operated under the assumption that checksumming them introduces insignificant overhead.  We should make a benchmark and check this.

execution_engine/src/shared/newtypes/blake2b256.rs
nilz3ro: The old PR did not have an `impl CheckSummedHex<Blake2bHash> for Blake2bHash` any recommendations would help

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+/// `usize`s.
+fn blake2b_hash(data: impl AsRef<[u8]>) -> Vec<u8> {
+    let mut hasher = Blake2b::new();
+    hasher.update(data);
+    hasher.finalize().to_vec()
+}
+
+const HEX_CHARS: [char; 16] = [
+    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f',
+];
```
xcthulhu:
```suggestion
const HEX_CHARS: [char; 16] = [
    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f', 'A', 'B', 'C', 'D', 'E', 'F',
];
```

This way could avoid `to_uppercase`, and `to_lowercase` with proper indexing.

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+///   - Uses hash bits rather than nibbles
+///
+/// [1]: https://eips.ethereum.org/EIPS/eip-55
+pub fn encode(input: &(impl AsRef<[u8]> + ?Sized)) -> String {
+    let input_bytes = input.as_ref();
+    let mut hash_bits = bytes_to_bits_cycle(blake2b_hash(input));
+    let mut hex_output_string = String::with_capacity(input_bytes.len() * 2);
+    for nibble in bytes_to_nibbles(input_bytes) {
+        let c = HEX_CHARS[nibble as usize];
+        if c.is_alphabetic() && hash_bits.next().unwrap_or(true) {
```
xcthulhu: Reimplement `is_alphabetic` using a nibble (not a char) as `|nibble| nibble >= 10`

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+            hex_output_string.extend(c.to_uppercase())
```
xcthulhu: Following the optimization suggested [here][1], this gets replaced with +6.

[1]: https://github.com/casper-network/casper-node/pull/1925/files/60bcba64a403d520b9604bac1f9729e4899cc3e6#r693932190

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+/// Encodes bytes as hexadecimal with mixed-case based checksums following a scheme similar to
+/// [EIP-55][1].
+///
+/// Key differences:
+///   - Works on any length of data, not just 20-byte addresses
+///   - Uses Blake2b hashes rather than Keccak
+///   - Uses hash bits rather than nibbles
+///
+/// [1]: https://eips.ethereum.org/EIPS/eip-55
+pub fn encode(input: &(impl AsRef<[u8]> + ?Sized)) -> String {
```
xcthulhu: Maybe we should have this be lazy, like

```suggestion
pub fn encode(input: &(impl AsRef<[u8]> + ?Sized)) -> impl Iterator<Item=char> {
```

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+/// For backward compatibility: if the hex string is all uppercase or all lowercase, the check is
+/// skipped.
+///
+/// [1]: https://eips.ethereum.org/EIPS/eip-55
+pub fn decode(input: &(impl AsRef<[u8]> + ?Sized)) -> Result<Vec<u8>, base16::DecodeError> {
+    let bytes = base16::decode(input)?;
+    // If the string was all uppercase or all lower case, don't verify the checksum.
+    // This is to support legacy clients.
+    // Otherwise perform the check as below.
+    if !(string_is_uppercase(input) || string_is_lowercase(input)) {
```
xcthulhu: Reduce two passes to one pass

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+///
+/// Key differences:
+///   - Works on any length of data, not just 20-byte addresses
+///   - Uses Blake2b hashes rather than Keccak
+///   - Uses hash bits rather than nibbles
+///
+/// [1]: https://eips.ethereum.org/EIPS/eip-55
+pub fn encode(input: &(impl AsRef<[u8]> + ?Sized)) -> String {
+    let input_bytes = input.as_ref();
+    let mut hash_bits = bytes_to_bits_cycle(blake2b_hash(input));
```
xcthulhu: Would be nice if there was a stack-only `blake2b_hash` rather than this one that uses the heap

types/src/check_summed_hex.rs
xcthulhu: We might want two of these: one lazy and one not.  Or maybe we can use some hints with sizes for the iterator...

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,436 @@
...
+/// skipped.
+///
+/// [1]: https://eips.ethereum.org/EIPS/eip-55
+pub fn decode(input: &(impl AsRef<[u8]> + ?Sized)) -> Result<Vec<u8>, base16::DecodeError> {
+    let bytes = base16::decode(input)?;
+    // If the string was all uppercase or all lower case, don't verify the checksum.
+    // This is to support legacy clients.
+    // Otherwise perform the check as below.
+    if !(string_is_uppercase(input) || string_is_lowercase(input)) {
+        let checksum_hex_bytes = encode(&bytes).into_bytes();
```
xcthulhu: We could use a lazy `encode` here...

types/src/check_summed_hex.rs
goral09: If that's the case then we should also update the original CEP (https://github.com/xcthulhu/ceps/blob/checksummed-addresses/text/0057-checksummed-addresses.md) which didn't use the uppercased letters.

types/src/check_summed_hex.rs
```diff
@@ -57,7 +61,8 @@ pub fn encode(input: &(impl AsRef<[u8]> + ?Sized)) -> String {
     let mut hex_output_string = String::with_capacity(input_bytes.len() * 2);
     for nibble in bytes_to_nibbles(input_bytes) {
         let c = HEX_CHARS[nibble as usize];
-        if c.is_alphabetic() && hash_bits.next().unwrap_or(true) {
+        let hash_bit = hash_bits.next().unwrap_or(true);
+        if c.is_alphabetic() && hash_bit {
```
afck: The old code deliberately did not advance the iterator if `c` is a number, because then we don't need the bit.
That way we always make use of as many bits from the hash as possible.

types/src/check_summed_hex.rs
```diff
@@ -86,22 +86,26 @@ pub fn encode_iter(input: &(impl AsRef<[u8]> + ?Sized)) -> impl Iterator<Item =
...
-        .iter()
-        .map(|byte| *byte as char)
-        .filter(|c| c.is_alphabetic())
-        .all(|c| c.is_lowercase())
+    match case_counts {
+        (0, 0) => true,
+        (0, 0..=u64::MAX) => true,
+        (0..=u64::MAX, 0) => true,
+        (0..=u64::MAX, 0..=u64::MAX) => false,
+    }
```
afck: No need to count; and we should return early as soon as we found one upper- and one lowercase letter.
Or use `Itertools`:

```rust
bytes.iter()
    .filter_map(|byte| {
        let c = *byte as char;
        c.is_alphabetic().then(|| c.is_uppercase())
    })
    .tuple_windows()
    .all(|(upper0, upper1)| upper0 == upper1)
```

types/src/check_summed_hex.rs
```diff
@@ -0,0 +1,492 @@
...
+    // If the string was all uppercase or all lower case, don't verify the checksum.
+    // This is to support legacy clients.
+    // Otherwise perform the check as below.
+    if !string_is_same_case(&bytes) {
+        let input_string_bytes = input.as_ref();
+
+        return encode_iter(&bytes)
+            .zip(input_string_bytes.iter())
+            .enumerate()
+            .map(|(idx, (chr, &byt))| {
```
xcthulhu:
```suggestion
            .map(|(index, (input_hex_char, &expected_cased_hex_char))| {
```

types/src/check_summed_hex.rs
nilz3ro: Note: We talked about this and decided not to change hashing in this PR.

node/src/crypto/hash.rs
nilz3ro: marking as resolved since there is another PR that will unify `Blake2b` and `Digest`.

xcthulhu (approved):

xcthulhu:
LGTM, we just need to get rid of `checksum_encode_if_small` (since the code always does that now)

types/src/checksummed_hex.rs
```diff
@@ -53,33 +57,35 @@ const HEX_CHARS: [char; 22] = [
...
-        if nibble >= 10 && hash_bits.next().unwrap_or(true) {
-            nibble += 6;
-        }
-        hex_output_string.push(HEX_CHARS[nibble as usize])
-    }
-    hex_output_string
+    encode_iter(input).collect()
 }
 
 /// `encode` but it returns an iterator.
```
afck, line 66: So this automatically checksums only for small inputs now, and otherwise outputs lowercase? That feels like somewhat surprising behavior. At least it should be explained in the documentation.

afck (approved):

client/tests/integration_test.rs
```diff
@@ -606,15 +605,19 @@ mod get_item {
...
+        println!("{:?}", blah);
+        // assert!(matches!(
+        //     server_handle
+        //         .get_item("<invalid state root hash>", "", "")
+        //         .await,
+        //     Err(Error::CryptoError {
+        //         context: "state_root_hash",
+        //         error: CryptoError::FromHex(base16::DecodeError::InvalidLength { length: 25 })
+        //     })
+        // ));
```
xcthulhu: @nilz3ro What happened here?

types/src/checksummed_hex.rs
xcthulhu: @Fraser999 wanted this behavior for `bytesrepr::Bytes` so for the sake of consistency we're doing it everywhere now.

types/src/checksummed_hex.rs
```diff
@@ -0,0 +1,584 @@
...
+
+/// Provides checksummed hex-encoded (de)serialization for `serde` when used with human-readable
+/// (de/en)coders, following an [EIP-55][1]-like scheme.  For non-human-readable (de/en)coders, the
+/// value is (de)serialized as a byte array.
+///
+/// Note that the trait is automatically implemented for types that implement [`AsRef`]`<[u8]>` and
+/// [`TryFrom`]`<&[u8]>`.
+///
+/// [1]: https://eips.ethereum.org/EIPS/eip-55
+pub trait CheckSummedHex<T> {
```
xcthulhu:
```suggestion
pub trait ChecksummedHex<T> {
```

My bad for messing up the capitalization here :(

types/src/checksummed_hex.rs
```diff
@@ -0,0 +1,584 @@
...
+        };
+        maybe_bytes.and_then(|bytes| Self::from_bytes(&bytes).map_err(D::Error::custom))
+    }
+}
+
+/// A dummy container for use inside `#[serde(with)]` attribute if the underlying type
+/// implements `AsRef<[u8]>` and `TryFrom<&[u8], _>`.
+#[cfg_attr(docsrs, doc(cfg(feature = "alloc")))]
+#[derive(Debug)]
+pub struct CheckSummedHexForm<T>(PhantomData<T>);
```
xcthulhu:
```suggestion
pub struct ChecksummedHexForm<T>(PhantomData<T>);
```

types/src/checksummed_hex.rs
```diff
@@ -0,0 +1,584 @@
...
+
+impl<T, E> CheckSummedHex<T> for CheckSummedHexForm<T>
```
xcthulhu:
```suggestion
impl<T, E> ChecksummedHex<T> for ChecksummedHexForm<T>
```

client/tests/integration_test.rs
nilz3ro: Oh, I was having a look around and I forgot to remove this.

types/src/checksummed_hex.rs
```diff
@@ -0,0 +1,584 @@
...
+            }
+        }
+        HEX_CHARS[nibble as usize]
+    })
+}
+
+/// Returns true if all chars in a string are uppercase or lowercase.
+/// Returns false if the string is mixed case or if there are no alphabetic chars.
+fn string_is_same_case<T: AsRef<[u8]> + ?Sized>(s: &T) -> bool {
+    let mut chars = s.as_ref().iter().filter(|c| c.is_ascii_alphabetic());
```
xcthulhu: cheaper to check if member of `[ 'A', 'B', 'C', 'D', 'E', 'F', 'a', 'b', 'c', 'd', 'e', 'f']` I think.

types/src/checksummed_hex.rs
```diff
@@ -0,0 +1,584 @@
...
+
+    match chars.next() {
+        Some(first) => {
+            let is_upper = first.is_ascii_uppercase();
```
xcthulhu: cheaper to check if member of `['A', 'B', 'C', 'D', 'E', 'F']`

types/src/checksummed_hex.rs
goral09: We can cast `char` to an integer and check if it's in expected range (`A-F` or `a-f`).

xcthulhu: Looks like it's already a `u8`...

xcthulhu:
```suggestion
    let mut chars = s.as_ref().iter().filter(|c| c >= 65); // 65 is the ascii code for 'A', '0'-'9' is 48-57
```

types/src/checksummed_hex.rs
xcthulhu:
```suggestion
            let is_upper = first < 97; // 97 is the ascii code for 'a', 'A'-'F' is 65-70
```

types/src/checksummed_hex.rs
```diff
@@ -0,0 +1,584 @@
...
+            chars.all(|c| c.is_ascii_uppercase() == is_upper)
```
xcthulhu:
```suggestion
            chars.all(|c| c < 97 == is_upper)
```

types/src/checksummed_hex.rs
nilz3ro: @xcthulhu I tried this but it breaks in our proptests. I went with @goral09's suggestion.

xcthulhu (approved):
LGTM

dwerner (approved):

dwerner:
Approved but let's hold off merging until we're out of code freeze.

types/src/checksummed_hex.rs
```diff
@@ -82,18 +83,18 @@ pub fn encode_iter(input: &(impl AsRef<[u8]> + ?Sized)) -> impl Iterator<Item =
 /// Returns true if all chars in a string are uppercase or lowercase.
 /// Returns false if the string is mixed case or if there are no alphabetic chars.
 fn string_is_same_case<T: AsRef<[u8]> + ?Sized>(s: &T) -> bool {
-    let lower_range = 97..=102u8; // a..=f
-    let upper_range = 65..=70u8; // A..=F
+    const LOWER_RANGE: RangeInclusive<u8> = 'a' as u8..='f' as u8;
+    const UPPER_RANGE: RangeInclusive<u8> = 'A' as u8..='F' as u8;
```
goral09: I'm not an expert in Rust optimization but this looks like it could be extract to be a `const` outside of this function _and_ have the "containst" methods also be `const`. What do you think?

xcthulhu: Disclosure: I wrote this with @nilz3ro 

#### _Placement of `const` declarations_

The generated assembly is exactly the same whether you move those consts outside of the function or not: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=71e395247fd3b803e383097c1cfe6587

So that's a matter of taste IMO.

 #### _Helper functions_

Breaking out a helper function probably _decreases_ performance slightly.  If you look at the assembly for that, you can see there's more `callq` instructions, which means it's using more stack frames:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1f859aa42039f690fe0e32493adc168c

Slapping an `#[inline(always)]` makes the `callq`s go away (for `debug` builds at any rate):

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=58501e1332f610c29b6703e293aadde4

We could play around with `release` builds, but rust-playground shows everything gets optimized away.  I'd have to change `main` I think.  Rather than looking at the assembly we probably want to run the benches IMO.

goral09: Nice! Thanks for taking the time to explain!

(BTW, all the Rust playground links are exactly the same).

xcthulhu: 😖 I went back and fixed the links.

PR #1926, opened 2021-08-16
ApiError documentation

### LINKED ISSUE #1727: Update the rustdocs for api_error
We need to get those over to docs comments in api_types.rs so they appear in https://docs.rs/casper-types/1.2.0/casper_types/enum.ApiError.html#mappings

dwerner:
Resolves #1727

The purpose of this PR is to improve documentation of `ApiError` and it's sub error types, and the error code mappings for them. For each variant, an integer value mapping is shown in a doctest.

ApiError's `Debug` impl also now `try_from`s the underlying error code to try to get more debugging for the user.


goral09:
We have multiple, module-specific, errors each documented individually. Would it be possible to have it (error documentation) statically-checked/generated and available in a single place? Products like https://cspr.live/ could integrate that "table" into their flow and present users with the human-readable interpretation of the error codes without having to parsing multiple files.

Fraser999 (approved):

PR #1935, opened 2021-08-18
Remove backtrace from Responder - keep logging self

dwerner:
The purpose of this PR is to remove the backtrace from `Responder`, which was unconditionally created and is expensive. We keep logging `self`, however, so there's at least a hint in this error case.

### DIFF
--- a/node/src/effect.rs
+++ b/node/src/effect.rs
@@ -176,24 +176,19 @@ impl<T: 'static + Send> Responder<T> {
 
 impl<T> Responder<T> {
     /// Send `data` to the origin of the request.
-    pub(crate) fn respond(mut self, data: T) -> impl Future<Output = ()> {
-        let caller_backtrace = backtrace::Backtrace::new();
-        async move {
-            if let Some(sender) = self.0.take() {
-                if sender.send(data).is_err() {
-                    error!(
-                        responder = ?self,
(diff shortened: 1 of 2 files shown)


goral09 (approved):

node/src/effect.rs
```diff
@@ -174,26 +174,26 @@ impl<T: 'static + Send> Responder<T> {
...
-                }
-            } else {
+    pub(crate) async fn respond(mut self, data: T) {
+        if let Some(sender) = self.0.take() {
+            if let Err(data) = sender.send(data) {
                 error!(
                     responder = ?self,
-                    ?caller_backtrace,
-                    "tried to send a value down a responder channel, but it was already used"
+                    ?data,
```
goral09, line 186: If we log the `data` we don't have to log the `responder` as that will be a duplicated information: `self` will log the type of `data` effectively. Same below.

dwerner: Addressed.

afck (approved):

afck:
bors merge

dwerner:
bors try

dwerner:
bors merge

afck:
bors merge

dwerner:
bors merge

PR #1938, opened 2021-08-18
Genesis should not clobber an existing blockchain

### LINKED ISSUE #1937: Check For Existing Block In Genesis to Avoid Clobbering an Existing Blockchain
When running genesis in order to create a switch block, we must not overwrite an existing blockchain.

xcthulhu:
Before running genesis in order to create a switch block, we first check to see if we have any blocks so we don't clobber an existing blockchain.

Closes #1937

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1143,12 +1143,10 @@ impl Storage {
 
     /// Retrieves the highest block header from the storage, if one exists.
     pub fn read_highest_block_header(&self) -> Result<Option<BlockHeader>, Error> {
-        let highest_block_hash = match self.block_height_index
-            .iter()
-            .last() {
-                Some((_, highest_block_hash)) => highest_block_hash,
-                None => return Ok(None),
-            };
+        let highest_block_hash = match self.block_height_index.iter().last() {
+            Some((_, highest_block_hash)) => highest_block_hash,
+            None => return Ok(None),
+        };
         self.read_block_header_by_hash(highest_block_hash)
     }
 
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1141,6 +1141,17 @@ impl Storage {
             .transpose()
     }
 
+    /// Retrieves the highest block header from the storage, if one exists.
+    pub fn read_highest_block_header(&self) -> Result<Option<BlockHeader>, Error> {
+        let highest_block_hash = match self.block_height_index
+            .iter()
+            .last() {
+                Some((_, highest_block_hash)) => highest_block_hash,
+                None => return Ok(None),
+            };
+        self.read_block_header_by_hash(highest_block_hash)
+    }
+
     /// Returns vector blocks that satisfy the predicate, starting from the latest one and following
     /// the ancestry chain.
     fn get_blocks_while<F, Tx: Transaction>(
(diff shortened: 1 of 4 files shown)


node/src/components/storage.rs
```diff
@@ -1141,6 +1141,17 @@ impl Storage {
             .transpose()
     }
 
+    /// Retrieves the highest block header from the storage, if one exists.
+    pub fn read_highest_block_header(&self) -> Result<Option<BlockHeader>, Error> {
+        let highest_block_hash = match self.block_height_index
+            .iter()
+            .last() {
```
goral09: Seems like having a separate value just for tracking the highest block would be useful. I can imagine a query like `GetHighestBlock`.

node/src/reactor/participating.rs
```diff
@@ -586,10 +586,10 @@ impl reactor::Reactor for Reactor {
             match chainspec.protocol_config.activation_point {
                 ActivationPoint::Genesis(genesis_timestamp) => {
                     // Check that no blocks exist in storage so we don't try to run genesis again on
-                    // an existing blockchain
-                    if let Some(first_block_header) = storage.read_block_header_by_height(1)? {
+                    // an existing blockchain.
+                    if let Some(highest_block_header) = storage.read_highest_block_header()? {
```
goral09, line 590: Isn't that equivalent though? If we have `n-th` block or `1st` we would be overwriting a chain regardless. It doesn't matter how many blocks there are.

node/src/components/storage.rs
afck: I don't think it needs a separate field: Accessing the last entry of a `BTreeMap` is fast.

afck (approved):

node/src/components/storage.rs
goral09: Ok.

node/src/reactor/participating.rs
goral09: If it's equivalent then this PR is unnecessary.

node/src/components/storage.rs
```diff
@@ -1329,7 +1340,7 @@ impl Storage {
     }
 
     // Retrieves a block header by hash.
-    pub(crate) fn get_block_header_by_hash(
+    pub(crate) fn read_block_header_by_hash(
```
goral09, line 1341: Is this rename unnecessary? I don't understand why you're making these decision unilaterally. There's `get_single_block`, `get_blocks_while`, `get_block_body_for_*`, etc. so there's definitely a pattern here to use `get_*`.

xcthulhu: Over time, the convention in this file has become that `get_` methods are private and take a transaction and `read_` methods are public and do not take a transaction.

node/src/reactor/participating.rs
xcthulhu: The node may not necessarily have the first block but it could still have a highest block.  For instance if someone uses fast sync with the trusted hash of the 100th switchblock, then he or she will never download the 0th or 1st block-header (according to the flow in `run_fast_sync_task`).

goral09: That makes sense, you're right. I didn't think of it. Thanks.

node/src/components/storage.rs
goral09: Thanks for the explanation. This is a weird design decision. I don't see any advantage of it - no one will memorize that rule anyway. Developers will type in `get_` or `read_` and then look for what the IDE is suggesting. It's easier to remember one rule (`get_` or `read_`) than thinking whether I have a transaction or not and whether I want to call public or private endpoint.

PR #1957, opened 2021-08-19
GH-1930 transform journal

### LINKED ISSUE #1930: Tracking copy transforms should be in insertion order
to be consistent ops also.

CC @EdHastingsCasperLabs @goral09

dwerner:
Resolves #1930

The purpose of this PR is to introduce an `ExecutionJournal` and related plumbing, allowing `Transforms` to be ordered by insertion. This then allows the values stored to reflect that insertion order.


node/src/effect/announcements.rs
```diff
@@ -224,16 +224,16 @@ pub(crate) enum ContractRuntimeAnnouncement {
     StepSuccess {
         /// The era id in which the step was committed to global state.
         era_id: EraId,
-        /// The operations and transforms committed to global state.
-        execution_effect: ExecutionEffect,
+        /// The JSON API format journal of transforms committed to global state.
+        json_execution_journal: JsonExecutionJournal,
```
goral09: We shouldn't be using JSON for internal structures. It's only a representation type so internally we should be using domain types like `ExecutionJournal` and map it to a JSON only before presentation (here, even stream). This comment applies to everything in this file and as a general rule.

types/src/execution_result.rs
```diff
@@ -287,39 +266,54 @@ impl FromBytes for ExecutionResult {
...
 }
 
-impl ToBytes for ExecutionEffect {
+impl JsonExecutionJournal {
+    /// Constructor for [`JsonExecutionJournal`].
+    #[allow(deprecated)]
+    pub fn new(transforms: Vec<TransformEntry>) -> Self {
+        Self {
+            transforms,
+            operations: Default::default(),
```
goral09, line 287: We were suppossed to discuss it before making the decision - i.e. we can still recover `Op`s from the `Transform`s. If we all agree to make it empty that's fine but let's have that discussion first.

types/src/execution_result.rs
```diff
@@ -287,39 +266,54 @@ impl FromBytes for ExecutionResult {
     }
 }
 
-/// The effect of executing a single deploy.
+/// The journal of execution transforms from a single deploy.
 #[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Default, Debug)]
 #[cfg_attr(feature = "std", derive(JsonSchema))]
 #[serde(deny_unknown_fields)]
-pub struct ExecutionEffect {
+pub struct JsonExecutionJournal {
```
goral09: I realize that it seems to be the approach we have now, where the `Json*` representations of the specific type/ADT are defined next to the original type but I think these presentation types should be living in another module. The reasoning being that they are from a different bounded context - namely API/event stream/etc. and can evolve independently from the types they map from.

/cc @Fraser999 @marc-casperlabs

types/src/execution_result.rs
```diff
@@ -114,11 +96,11 @@ static EXECUTION_RESULT: Lazy<ExecutionResult> = Lazy::new(|| {
 #[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
 #[cfg_attr(feature = "std", derive(JsonSchema))]
 #[serde(deny_unknown_fields)]
-pub enum ExecutionResult {
+pub enum JsonExecutionResult {
```
goral09: Why can't it still be called `ExecutionResult`? Just because it _can_ be mapped to JSON in the presentation layer doesn't really mean it _is only_ a JSON wrapper. I would keep the old name, current one confuses me. When I see a type/method that starts with `JSON*` I immediately think _"Oh, that is for the API where we use JSON"_.

node/src/components/rpc_server/rpcs/info.rs
```diff
@@ -60,11 +60,11 @@ impl DocExample for GetDeployParams {
 /// The execution result of a single deploy.
 #[derive(Serialize, Deserialize, Debug, JsonSchema)]
 #[serde(deny_unknown_fields)]
-pub struct JsonExecutionResult {
+pub struct JsonBlockHashAndExecutionResult {
```
goral09: If we rename the _new_ `JsonExecutionResult` back to `ExecutionResult` then this can remain being `JsonExecutionResult`. In this case it does make sense - it's a type for JSON API only.

goral09:
I don't have any blockers for the PR but if we update the names then this PR will get _much_ smaller.

types/src/execution_result.rs
goral09: cc @EdHastingsCasperLabs @Fraser999 . This change seems unharmful as probably no one uses the `Ops` offchain.

types/src/execution_result.rs
Fraser999: While I agree with @goral09's point, a stronger argument is that this is actually a breaking API change for the JSON-RPC server and the event stream server.

We could look to manually implement `JsonSchema` for this type and `ExecutionEffect` to work around the schema change, but I think the better and simpler approach is to just revert the name changes.

types/src/execution_result.rs
Fraser999: Maybe.  But addressing that is outside the scope of this PR though.

types/src/execution_result.rs
Fraser999: I believe we made that decision now.

types/src/execution_result.rs
```diff
@@ -287,39 +266,54 @@ impl FromBytes for ExecutionResult {
...
 
-/// The effect of executing a single deploy.
+/// The journal of execution transforms from a single deploy.
 #[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Default, Debug)]
 #[cfg_attr(feature = "std", derive(JsonSchema))]
 #[serde(deny_unknown_fields)]
-pub struct ExecutionEffect {
+pub struct JsonExecutionJournal {
     /// The resulting operations.
+    #[deprecated(since = "1.4.0")]
```
Fraser999, line 275: Shouldn't we just deprecate the `Operation` and `OpKind` types instead?

Even if that's not practicable for some reason, I think we should update the doc comment here, and the doc comments on those two types to indicate they're deprecated.  The doc comments appear in the generated JSON schema.

node/src/components/rpc_server/rpcs/info.rs
Fraser999: I agree with @goral09.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -215,6 +216,7 @@ pub struct TrackingCopy<R> {
     cache: TrackingCopyCache<HeapSize>,
     ops: AdditiveMap<Key, Op>,
     fns: AdditiveMap<Key, Transform>,
+    journal: Vec<(Key, Transform)>,
```
Fraser999, line 219: I think we should add test coverage for this new member in the existing unit tests.

Fraser999, line 219: Seems like we could replace `fns` with `journal` and just create the `AdditiveMap<Key, Transform>` on the fly inside `fn effect(&self)`.  Not sure if it'd be a better option, but was it considered?

execution_engine/src/core/engine_state/execution_result.rs
```diff
@@ -53,21 +43,24 @@ pub enum ExecutionResult {
         execution_effect: ExecutionEffect,
         transfers: Vec<TransferAddr>,
         cost: Gas,
+        execution_journal: ExecutionJournal,
```
Fraser999, line 47: Was the possibility of not duplicating the transforms across the `ExecutionJournal` and `ExecutionEffect` considered here?  Again, seems like we could dedup by swapping `execution_effect` for just the `ops`, and using that and the `execution_journal` we could create an `ExecutionEffect` on the fly as required.

execution_engine/src/core/engine_state/execution_effect.rs
```diff
@@ -15,25 +17,38 @@ impl ExecutionEffect {
...
+            Transform::Identity => Op::NoOp,
+            Transform::Write(_) => Op::Write,
+            Transform::AddInt32(_)
+            | Transform::AddUInt64(_)
+            | Transform::AddUInt128(_)
+            | Transform::AddUInt256(_)
+            | Transform::AddUInt512(_)
+            | Transform::AddKeys(_) => Op::Add,
+
+            // should be unreachable.
```
Fraser999, line 50: It seems like this is true because it's currently only called from `impl From<ExecutionJournal> for ExecutionEffect` above.  That seems fairly risky to me.

I think it'd be better to remove this impl and just match on the `transform` at line 26 above.

execution_engine/src/core/tracking_copy/mod.rs
Fraser999, line 219: Shouldn't we favour using the newtype here over a bare `Vec`?

types/src/execution_result.rs
xcthulhu: @goral09 We discussed this with @EdHastingsCasperLabs on Monday... you had to step out but he said it was alright.

execution_engine/src/core/engine_state/execution_effect.rs
```diff
@@ -15,25 +17,38 @@ impl ExecutionEffect {
...
+        let mut ops = AdditiveMap::new();
+        let mut transforms = AdditiveMap::new();
+        let journal: Vec<(Key, Transform)> = journal.into();
+        for (key, transform) in journal.into_iter() {
+            if let Transform::Failure(_) = transform {
+                continue;
+            }
+            let op: Op = (&transform).into();
+            ops.insert_add(key, op);
+            transforms.insert_add(key, transform.clone());
```
Fraser999:
```suggestion
            transforms.insert_add(key, transform);
```

types/src/execution_result.rs
xcthulhu: Well... this is for the API where we use JSON.

The ABI type is in the execution engine.

See https://github.com/casper-network/casper-node/pull/1957/files#diff-341faf4a5b1379f6974e6fc81be534dd7b03b6f9d9b8be78b87a6c71ce662ba5R310

Fraser999 (changes requested):
Main blocker is the non-backwards-compatible change to the names of the types in `casper-types`.

node/src/components/rpc_server/rpcs/info.rs
xcthulhu: Sadly, the other `JsonExecutionResult` is in fact a JSON representing of an ABI type... the name for this thing appears to be apropos...

types/src/execution_result.rs
goral09: Just because this particular type is used in contexts where it is being mapped to JSON doesn't mean it has to include `Json` in its name.

execution_engine/CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ All notable changes to this project will be documented in this file.  The format
 * Increase price of `create_purse` to 2.5CSPR.
 * Improve doc comments to clarify behavior of the bidding functionality.
 * Changed parameters to `LmdbEnvironment`'s constructor enabling manual flushing to disk.
+* Execution transforms are ordered by insertion.
```
goral09: Maybe
```suggestion
* Execution transforms are returned in their insertion order.
```
?

node/src/components/rpc_server/rpcs/info.rs
goral09: I see that you reverted the rename of `ExecutionResult` @dwerner , can we also revert _this_ renaming?

resources/test/sse_data_schema.json
```diff
@@ -1047,10 +1047,11 @@
           "type": "array",
           "items": {
             "$ref": "#/definitions/Operation"
-          }
+          },
+          "deprecated": true
```
Fraser999, line 1051: This addition to the schema has presumably come about due to the addition of the `deprecated` attribute, so seems like we'll need to remove that attribute in order to be strictly backwards compatible.  We'll need to handle this via updating the doc comments on the deprecated members/types only.

xcthulhu (changes requested):
I am concerned that this will introduce a performance regression in the EE.  CC @luis-casper

luis-casper:
After me taking over this PR from @dwerner, there is a new PR where I am landing my new commits.
https://github.com/casper-network/casper-node/pull/2174

luis-casper:
> Seems like we could replace fns with journal and just create the AdditiveMap<Key, Transform> on the fly inside fn effect(&self). Not sure if it'd be a better option, but was it considered?

Done.

luis-casper:
> Was the possibility of not duplicating the transforms across the ExecutionJournal and ExecutionEffect considered here? Again, seems like we could dedup by swapping execution_effect for just the ops, and using that and the execution_journal we could create an ExecutionEffect on the fly as required.

@Fraser999: I just tried to get rid of `execution_effect`. The problem is that field is no mere compilation of `execution_journal`. For instance, the `build` function in `execution_engine/src/core/engine_state/execution_result.rs` will gladly produce an `ExecutionResult` with an empty journal but non-empty effects.

luis-casper:
> I am concerned that this will introduce a performance regression in the EE. CC @luis-casper

@xcthulhu: There will be a performance impact, for sure. I have requested a ticket with a long-term plan to mitigate this.
https://github.com/casper-network/casper-node/issues/2173

luis-casper:
> This addition to the schema has presumably come about due to the addition of the deprecated attribute, so seems like we'll need to remove that attribute in order to be strictly backwards compatible. We'll need to handle this via updating the doc comments on the deprecated members/types only.

@Fraser999: I have added the documentation and removed the attribute. I don't know how to regenerate the schema to check, though.

luis-casper:
> It seems like this is true because it's currently only called from impl From<ExecutionJournal> for ExecutionEffect above. That seems fairly risky to me.

> I think it'd be better to remove this impl and just match on the transform at line 26 above.

I agree. Furthermore, there is a bug in the following line, as the type of the identity transform should be `Op::Read` (discussed with @xcthulhu and @dwerner).
```rust
            Transform::Identity => Op::NoOp,
```

Both changes have been implemented.

luis-casper:
> I think we should add test coverage for this new member in the existing unit tests.

Done.

luis-casper:
> Shouldn't we favour using the newtype here over a bare Vec?

Done.

luis-casper:
> Shouldn't we just deprecate the Operation and OpKind types instead?

> Even if that's not practicable for some reason, I think we should update the doc comment here, and the doc comments on those two types to indicate they're deprecated. The doc comments appear in the generated JSON schema.

I have updated the comments on the types, as I couldn't get it to compile with the `deprecated` attribute.

luis-casper:
> Was the possibility of not duplicating the transforms across the ExecutionJournal and ExecutionEffect considered here? Again, seems like we could dedup by swapping execution_effect for just the ops, and using that and the execution_journal we could create an ExecutionEffect on the fly as required.

The execution effect was removed entirely.

luis-casper:
> This addition to the schema has presumably come about due to the addition of the deprecated attribute, so seems like we'll need to remove that attribute in order to be strictly backwards compatible. We'll need to handle this via updating the doc comments on the deprecated members/types only.

It was removed.

luis-casper:
> I see that you reverted the rename of ExecutionResult @dwerner , can we also revert this renaming?

It is reverted.

goral09:
Is this PR superseeded by https://github.com/casper-network/casper-node/pull/2174 ?

luis-casper:
> Is this PR superseeded by #2174 ?

@goral09: Yes, I've been recommended to create it to keep my work separate from @dwerner's.

PR #1961, opened 2021-08-20
Build test suite for Deploy Acceptor

darthsiroftardis:
CHANGELOG:

- Add a test fixture to the deploy acceptor
- Added tests to assert the intended behavior of the deploy acceptor

Closes: #1782


node/src/components/deploy_acceptor/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+/// Error type returned by the test reactor.
+#[derive(Debug, Error)]
+enum Error {
+    #[error("prometheus (metrics) error: {0}")]
+    Metrics(#[from] prometheus::Error),
+}
+
+#[allow(clippy::enum_variant_names)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum TestScenario {
```
rafal-ch, line 102: Can you consider extracting this enum to a separate file? The `tests.rs` is already quite long and the `TestScenario` comes with a moderately long `impl` block.

node/src/components/deploy_acceptor/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+    let deploy_responder = Responder::create(deploy_sender);
+
+    // Create a deploy specific to the test scenario
+    let deploy = test_scenario.deploy(&mut rng);
+    // Mark the source as either a peer or a client depending on the scenario.
+    let source = test_scenario.source(&mut rng);
+
+    {
+        // Inject the deploy artificially into storage to simulate a previously seen deploy.
+        if test_scenario == TestScenario::FromClientRepeatedValidDeploy
```
rafal-ch: I would prefer keeping the "is injection needed?" logic inside the `TestScenario` implementation. So, for example, in addition to functions like `is_valid_deploy_case()` you can consider adding smth like:
```
    fn requires_injected_deploy(&self) -> bool {
        match self {
            TestScenario::FromPeerInvalidDeploy
            | TestScenario::FromPeerValidDeploy
            | TestScenario::FromPeerRegression
            | TestScenario::FromClientInvalidDeploy
            | TestScenario::FromClientMissingAccount
            | TestScenario::FromClientInsufficientBalance
            | TestScenario::FromClientValidDeploy => false,
            TestScenario::FromPeerRepeatedValidDeploy
            | TestScenario::FromClientRepeatedValidDeploy => true,
        }
    }
```
and then just use `if test_scenario.requires_injected_deploy() ...`

rafal-ch:
This might be out of scope of this PR, but the unit tests for `Deploy::is_acceptable()` function could be hardened while we're taking care of `DeployAcceptor` tests.

For now I see that I can comment the line below and all tests still pass.
`self.is_valid_size(config.max_deploy_size)?;`

node/src/components/deploy_acceptor/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+            Box::new(deploy.clone()),
+            source,
+            deploy_responder,
+        ))
+        .await;
+
+    // Tests where the deploy is already in storage will not trigger any deploy acceptor
+    // announcement, so use the deploy acceptor `PutToStorage` event as the condition.
+    let stopping_condition = move |event: &Event| -> bool {
+        match test_scenario {
```
rafal-ch, line 428: Similar to the above, I'd rather see this logic hidden inside the `TestScenario` implementation. Consider adding a function to the `TestScenario` impl, for example `fn is_stop_condition(&self, event: &Event) -> bool` and move the match there.
On the caller site you can just use:

```
    runner
        .reactor_mut()
        .set_condition_checker(Box::new(move |event: &Event| -> bool {
            test_scenario.is_stop_condition(event)
        }));
```

node/src/components/deploy_acceptor/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+
+#[tokio::test]
+async fn should_reject_invalid_deploy_from_client() {
+    let result = run_deploy_acceptor(TestScenario::FromClientInvalidDeploy).await;
+    assert!(matches!(result, Err(super::Error::InvalidDeploy(_))))
+}
+
+#[tokio::test]
+async fn should_reject_valid_deploy_from_client_for_invalid_account() {
+    let result = run_deploy_acceptor(TestScenario::FromClientMissingAccount).await;
```
rafal-ch: Minor nit, consider using either `invalid` or `missing` word, not both.

node/src/components/deploy_acceptor/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+                // Contract runtime requests should not be made in the case of a deploy sent
+                // by a peer.
+                assert!(!self.test_scenario.is_from_peer());
+                match event {
+                    ContractRuntimeRequest::Query {
+                        query_request,
+                        responder,
+                    } => {
+                        let query_result =
+                            if self.test_scenario == TestScenario::FromClientMissingAccount {
```
rafal-ch, line 268: There are several similar comments below :)

I think we should keep such logic within the `impl` of `TestScenario`, like so:
```
    fn query_result(&self, query_request: &QueryRequest) -> QueryResult {
        match self {
            TestScenario::FromClientMissingAccount => QueryResult::ValueNotFound(String::new()),
            TestScenario::FromPeerInvalidDeploy
            | TestScenario::FromPeerValidDeploy
            | TestScenario::FromPeerRepeatedValidDeploy
            | TestScenario::FromPeerRegression
            | TestScenario::FromClientInvalidDeploy
            | TestScenario::FromClientInsufficientBalance
            | TestScenario::FromClientValidDeploy
            | TestScenario::FromClientRepeatedValidDeploy => {
                if let Key::Account(account_hash) = query_request.key() {
                    let preset_account =
                        Account::create(account_hash, BTreeMap::new(), URef::default());
                    QueryResult::Success {
                        value: Box::new(StoredValue::Account(preset_account)),
                        proofs: vec![],
                    }
                } else {
                    panic!("expect only queries using Key::Account variant");
                }
            }
        }
    }
```

and just use it on the caller side:
`let query_result = self.test_scenario.query_result(&query_request);`

node/src/components/deploy_acceptor/tests.rs
```diff
@@ -0,0 +1,584 @@
...
+            &storage_withdir,
+            None,
+            ProtocolVersion::from_parts(1, 0, 0),
+            false,
+            "test",
+        )
+        .unwrap();
+
+        let deploy_acceptor = DeployAcceptor::new(
+            super::Config::new(true),
```
rafal-ch: * Would it be reasonable to test the `false` variant as well? If I read the `deploy_acceptor.rs` correctly, we can have a deploy with invalid account which could still be accepted.

* Also, consider using a named const here: `super::Config::new(VERIFY_ACCOUNTS)` or similar

darthsiroftardis: No, it doesn't make sense to have a `false` variant cause that means it won't perform account verification, which is what we need to test here

node/src/components/deploy_acceptor/tests.rs
darthsiroftardis: No, I don't want to extract this into a `TestScenario` cause we are mocking behavior of the reactor, and we should have it present in the implementation of the trait itself

node/src/components/deploy_acceptor/tests.rs
darthsiroftardis: I agree with your point on the `VERIFY_ACCOUNTS` const

node/src/components/deploy_acceptor/tests.rs
darthsiroftardis: So both in this case and the above case, I want to have this present in the deploy acceptor function itself, cause I feel its better to define the stopping conditions for the `deploy_acceptor` in this function itself, as opposed to have it extracted away into a utility enum. Moreover, I think this makes it more consistent as it follows the same pattern in the gossiper tests. I am not going to plant my flag on this issue, however. So if @Fraser999 over-rules me here, I am okay with extracting them away

node/src/components/deploy_acceptor/tests.rs
darthsiroftardis: I am currently fine with having it being present in the same file, however, I feel with follow up work on the acceptor, the test suite may grow and then it may make sense to revisit this concern

node/src/components/deploy_acceptor/tests.rs
rafal-ch: Sure. I just observed that in the `deploy_acceptor.rs` we can replace the following line

`verify_accounts: config.verify_accounts()`

with

`verify_accounts: true`

and all `deploy_acceptor` tests still pass. If this is tested in another place (in gossiper, probably) though, I'm fine.

node/src/components/deploy_acceptor/tests.rs
darthsiroftardis: Applied suggestion

node/src/components/deploy_acceptor/tests.rs
darthsiroftardis: Applied suggestion

node/src/components/deploy_acceptor/tests.rs
darthsiroftardis: Discussed with @rafal-ch , the `false` scenario contains a subset of code paths in the `true` case, i.e account verification

node/src/components/deploy_acceptor/tests.rs
Fraser999: I'm in agreement with Karan here, although I wouldn't have objected if it were implemented in `TestScenario` either.  I slightly prefer just having it inline in the test here as I think it makes the test more understandable.

Fraser999 (approved):

node/src/types/deploy.rs
```diff
@@ -833,6 +833,11 @@ impl Deploy {
             None,
         )
     }
+
+    #[cfg(test)]
+    pub(crate) fn invalidate(&mut self) {
```
goral09, line 838: IMO this should be a function and not a method in the test file. How do you know that there's no chain whose `chain_name == ""`? It's a rather unique way to invalidate a deploy and is probably use-case specific.

Issue #1818, opened 2021-07-30
Express gas counter as u64


mpapierski:
We can safely assume that the gas limit will not exceed `u64::MAX`, and therefore we can make the gas counter represented in `u64` instead of the current `U512`. This also means we can express `Gas` new type as `u64` to simplify the code further.

This topic came up while discussing #1241 (https://github.com/casper-network/casper-node/pull/1241#issuecomment-885503490).

goral09:
I suggest that instead of jumping between Wasm & host on every Wasm opcode, we inject a local (global) variable into the smart contract, in the preprocess phase, equal to the gas bought, and modify that contract so that instead of calling `contract_ffi::charge_gas` (that forwards to the host) would modify that local, contract variable and fail if that value goes below 0 we fail the execution.

PR #1968, opened 2021-08-20
GH-1818: Express gas counter as u64

mpapierski:
Ref: #1818

This PR expresses `Gas` as `u64` therefore making the gas accounting much simpler, and possibly faster, as we don't need to jump around and convert `u32` from `gas` function into `U512` and do arithmetic on that.

(WIP benchmark results)


execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1459,17 +1461,27 @@ where
...
+            let session_gas_limit: Gas = match Motes::from_gas(payment_result_cost, gas_price)
+                .and_then(|payment_result_motes| {
+                    payment_purse_balance.checked_sub(payment_result_motes)
+                })
+                .and_then(|motes| Gas::from_motes(motes, gas_price))
+            {
+                Some(gas) => gas,
+                None => {
+                    return Ok(ExecutionResult::precondition_failure(
+                        Error::GasConversionOverflow,
```
afck, line 1493: Should there be a different error variant for the case where `payment_result_motes > payment_purse_balance`?

execution_engine/src/shared/gas.rs
```diff
@@ -1,36 +1,38 @@
...
+    pub const MAX: Gas = Gas::new(u64::MAX);
+    pub const MIN: Gas = Gas::new(u64::MIN);
+
+    pub const fn new(value: u64) -> Self {
         Gas(value)
     }
 
-    pub fn value(&self) -> U512 {
-        self.0
+    pub fn from_u512(value: U512) -> Option<Gas> {
```
afck: We could `impl TryFrom<U512> for Gas` instead.

execution_engine/src/shared/gas.rs
```diff
@@ -40,170 +42,211 @@ impl fmt::Display for Gas {
...
+    fn checked_add(&self, v: &Self) -> Option<Self> {
+        self.0.checked_add(v.0).map(Gas)
     }
 }
 
-impl std::ops::Mul for Gas {
-    type Output = Gas;
+impl CheckedMul for Gas {
+    fn checked_mul(&self, v: &Self) -> Option<Self> {
+        self.0.checked_add(v.0).map(Gas)
```
afck:
```suggestion
        self.0.checked_mul(v.0).map(Gas)
```

execution_engine/src/shared/gas.rs
```diff
@@ -40,170 +42,211 @@ impl fmt::Display for Gas {
...
     fn should_be_able_to_get_instance_of_gas() {
         let initial_value = 1;
-        let gas = Gas::new(U512::from(initial_value));
-        assert_eq!(
-            initial_value,
-            gas.value().as_u64(),
-            "should have equal value"
-        )
+        let gas = Gas::new(initial_value);
+        assert_eq!(gas, Gas::one(), "should have equal value")
```
afck: That comment is a bit redundant in an `assert_eq!`.

node/src/components/block_proposer/tests.rs
```diff
@@ -17,23 +20,21 @@ use crate::{
...
-}
+static DEFAULT_PAYMENT_AMOUNT: Lazy<Motes> = Lazy::new(|| Motes::new(U512::from(1u64)));
 
 fn generate_transfer(
     rng: &mut TestRng,
     timestamp: Timestamp,
     ttl: TimeDiff,
     dependencies: Vec<DeployHash>,
-    payment_amount: Gas,
+    payment_amount: Motes,
```
afck, line 28: Is the conversion rate 1 currently hardcoded here anyway? Otherwise it's probably not valid to just change this from `Gas` to `Motes`?

execution_engine/src/shared/storage_costs.rs
```diff
@@ -27,8 +24,12 @@ impl StorageCosts {
     }
 
     /// Calculates gas cost for storing `bytes`.
+    ///
+    /// This method will never overflow
     pub fn calculate_gas_cost(&self, bytes: usize) -> Gas {
-        let value = U512::from(self.gas_per_byte) * U512::from(bytes);
+        let gas_per_byte = self.gas_per_byte as u64;
+        let bytes_size = bytes as u64;
+        let value = gas_per_byte.saturating_mul(bytes_size);
```
afck: I wonder whether this should return `Option<Gas>` now:
If it returns `Some(u64::MAX.into())`, the cost was _exactly_ the maximum limit, which is fine. If it returns `None`, the gas limit was exceeded. (Probably not really relevant in practice, but still…)

afck (changes requested):

node/src/components/block_proposer/tests.rs
mpapierski: So this argument is used to generate runtime args for payment code that accepts tokens (aka Motes or U512). If it's used only for this, I think it's a valid change (considering this argument's meaning), although this is not related to the problem I was solving. I can happily revert changes here if you prefer

execution_engine/src/shared/gas.rs
mpapierski: `Gas::from_u512` is used only in `Gas::from_motes` - As `TryFrom` would be a public interface, I decided to make `from_u512` private, and we can always revisit this in the future if we need to convert U512 -> Gas in other places. But I think we shouldn't allow API to do that considering that U512 usually means Motes and it's a different unit - `Gas::from_motes` should be the preferred constructor

execution_engine/src/shared/gas.rs
mpapierski: Ouch! Thanks for pointing this out

execution_engine/src/shared/storage_costs.rs
mpapierski: I think you're right. Let's consider the current gas_limit is 0, and I paid `u64::MAX` in the payment code. Then somehow, I tried to overflow the gas counter by writing gigabytes of data (let's assume it's possible), then technically, storage will already write the data as we'd end up in this branch of `charge_gas`: https://github.com/casper-network/casper-node/blob/b84d3909d7d05d8b3865dff692dbbb9ea86f1c11/execution_engine/src/core/runtime_context/mod.rs#L760-L765. The system will return the gas limit on the next charged operation (even a wasm NOP), but the data is already on disk.

I think the same can apply for `HostFunction::calculate_gas_cost` where I'm saturating the calculation, rather than exit early with None if such calculation would overflow.

So I guess saturating gas counter in operations that can overflow now could be considered risky? Maybe it always was 🤔

mpapierski: Addressed in 50f8d34e95d92e401c8caf04332f9a7a77803e62 where I'm doing `checked_*` ops on Gas when charging for host function, and when charging for storage.

execution_engine/src/shared/gas.rs
mpapierski: Addressed in 50f8d34 where I'm doing checked_* ops on Gas when charging for host function, and when charging for storage.

node/src/components/block_proposer/tests.rs
afck: No, I'm fine with it if the other reviewer(s) are/is; I was just confused.

afck (approved):
Looks good to me, but also needs a review from someone more familiar with EE.

execution_engine/src/core/engine_state/mod.rs
mpapierski: It's worth mentioning we validate the account's main_purse to possess at least the `payment_result_cost` in https://github.com/casper-network/casper-node/blob/b84d3909d7d05d8b3865dff692dbbb9ea86f1c11/execution_engine/src/core/engine_state/execution_result.rs#L252-L258.

I believe by rewriting this expression to avoid overflow if the user pays Gas::MAX+1, I'm not changing the existing behavior

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3496,8 +3496,8 @@ where
     where
         T: AsRef<[Cost]> + Copy,
     {
-        let cost = host_function.calculate_gas_cost(weights);
-        self.gas(cost)?;
+        self.context
+            .charge_host_function_call(host_function, weights)?;
```
xcthulhu, line 3514: This is probably a dumb question but why did we remove `self.gas(cost)` here?

execution_engine/src/shared/host_function_costs.rs
```diff
@@ -130,14 +131,17 @@ where
...
-    pub fn calculate_gas_cost(&self, weights: T) -> Gas {
+    pub fn calculate_gas_cost(&self, weights: T) -> Option<Gas> {
         let mut gas = Gas::new(self.cost.into());
+
+        // sum i->n weight_i * arg_i
         for (argument, weight) in self.arguments.as_ref().iter().zip(weights.as_ref()) {
             let lhs = Gas::new((*argument).into());
             let rhs = Gas::new((*weight).into());
-            gas += lhs * rhs;
+            let pn = lhs.checked_mul(&rhs)?;
```
xcthulhu, line 141: Rather than an return an option, shouldn't we error with `Err(Error::GasLimit)` and return a `Result` instead?

execution_engine/src/shared/storage_costs.rs
```diff
@@ -27,9 +24,15 @@ impl StorageCosts {
...
 
     /// Calculates gas cost for storing `bytes`.
-    pub fn calculate_gas_cost(&self, bytes: usize) -> Gas {
-        let value = U512::from(self.gas_per_byte) * U512::from(bytes);
-        Gas::new(value)
+    ///
+    /// Returns wrapped [`Gas`] if calculation did not overflow. This method will return [`None`] if
+    /// the result would overflow [`Gas`] range - in this case the contract execution logic should
+    /// exit early with a gas limit error.
+    pub fn calculate_gas_cost(&self, bytes: usize) -> Option<Gas> {
```
xcthulhu, line 32: This looks like duplicate code [from above][1].  Here too, shouldn't we be returning a `Result` rather than an option?

[1]: https://github.com/casper-network/casper-node/pull/1968/files#diff-7201cf01479c2c4eecedce0e5537db3472585761a185dc2aa994dc2258855060R134

execution_engine/src/shared/gas.rs
```diff
@@ -40,170 +42,195 @@ impl fmt::Display for Gas {
     }
 }
 
-impl std::ops::Add for Gas {
-    type Output = Gas;
+impl CheckedAdd for Gas {
```
xcthulhu: 👍

execution_engine/src/shared/gas.rs
```diff
@@ -1,36 +1,38 @@
...
 
 use casper_types::U512;
 
 use crate::shared::motes::Motes;
 
-#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
-pub struct Gas(U512);
+const GAS_MAX_AS_U512: U512 = U512([u64::MAX, 0, 0, 0, 0, 0, 0, 0]);
+
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Num, Zero, One, NumOps)]
```
xcthulhu: Why would we ever want to do a regular add/sub?  Can we get rid of `NumOps` here?

execution_engine/src/shared/gas.rs
```diff
@@ -40,170 +42,195 @@ impl fmt::Display for Gas {
...
+impl CheckedSub for Gas {
+    fn checked_sub(&self, v: &Self) -> Option<Self> {
+        self.0.checked_sub(v.0).map(Gas)
+    }
+}
 
-    fn sub(self, rhs: Self) -> Self::Output {
-        let val = self.value() - rhs.value();
-        Gas::new(val)
+impl Sum for Gas {
```
xcthulhu: As I [mention above][1], I don't understand why we would ever want to add gas without checking for overflow...

[1]: https://github.com/casper-network/casper-node/pull/1968/files#r700620753

execution_engine/src/shared/gas.rs
```diff
@@ -40,170 +42,195 @@ impl fmt::Display for Gas {
...
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+        iter.fold(Gas::zero(), std::ops::Add::add)
     }
 }
 
-impl std::ops::Div for Gas {
-    type Output = Gas;
+impl Integer for Gas {
```
xcthulhu: Why do we need an instance for `Integer`?  Every time you construct `Gas` I see you are doing `Gas::from(1u64)`...

execution_engine/src/shared/gas.rs
```diff
@@ -40,170 +42,195 @@ impl fmt::Display for Gas {
...
+        self.0.is_odd()
+    }
+
+    fn div_rem(&self, other: &Self) -> (Self, Self) {
+        let (div, rem) = self.0.div_rem(&other.0);
+        (Gas(div), Gas(rem))
     }
 }
 
 impl From<u32> for Gas {
```
xcthulhu: When do we cast a `u32` into `Gas`?  Maybe we don't need this `impl`?

execution_engine/src/shared/gas.rs
```diff
@@ -40,170 +42,195 @@ impl fmt::Display for Gas {
...
 
 impl From<u64> for Gas {
     fn from(gas: u64) -> Self {
-        let gas_u512: U512 = gas.into();
-        Gas::new(gas_u512)
+        Gas::new(gas)
+    }
+}
+
+impl From<usize> for Gas {
```
xcthulhu: Like I mentioned [above][1], when do we cast a `usize` into `Gas`?  

[1]: https://github.com/casper-network/casper-node/pull/1968/files#r700622879

execution_engine/src/core/runtime/mod.rs
mpapierski: I removed the direct `self.gas` call here in favor of delegating into `RuntimeContext` to handle gas overflow properly. Here's the method I'm calling: 
https://github.com/casper-network/casper-node/blob/50f8d34e95d92e401c8caf04332f9a7a77803e62/execution_engine/src/core/runtime_context/mod.rs#L805-L822

Before, when the gas limit was U512 overflow (incorrectly) wasn't a concern in the math, so now, having it as a u64, I'm halting execution if a host function call would overflow

execution_engine/src/shared/host_function_costs.rs
mpapierski: Whenever we'd return `Error::GasLimit,` we should saturate the gas counter (set the counter to gas limit), and host_function_costs.rs doesn't have access to the gas counter. 

edit: Sure I could match the result and check for GasLimit, then saturate the counter, then return GasLimit again, although my concern would be safety as accidentally using "?" operator would give incorrect results

execution_engine/src/shared/gas.rs
mpapierski: That would be great to prevent using unchecked arithmetic, although opting out of NumOps would also require opting out of Zero (and others) as this one requires Add to be implemented etc. I don't know - I think we should either go all-in into num traits or opt-out of those completely and implement only checked ops using methods and no traits. What do you think?

execution_engine/src/shared/gas.rs
mpapierski: Actually, only a single test used the `Sum` trait. I removed this and replaced the code with `try_fold` that allows `checked_add` to be used

execution_engine/src/shared/gas.rs
mpapierski: At least one test used Ratio<> to calculate gas costs precisely, although I could refactor it. Fair point - we shouldn't add extra traits just for tests

execution_engine/src/shared/gas.rs
mpapierski: All costs on our platform are represented in u32 (host functions, opcodes, storage, etc.). I think it's safe to implement this trait for any integer type T that's smaller or equal to u64 in size, then create a Gas instance from a specific cost and continue math using Gas objects.

execution_engine/src/shared/gas.rs
mpapierski: https://github.com/mpapierski/casper-node/blob/2e08ea52fa54fcabe3c3c9b3badd655a7dc9f350/execution_engine/src/shared/storage_costs.rs#L32-L37 whenever we charge for storing arbitrary number of bytes

execution_engine/src/shared/gas.rs
mpapierski: Addressed further in https://github.com/mpapierski/casper-node/commit/761b14992e257610481ca7c90da305ec55d67665 where I'm shadowing num_traits::Checked* ops with methods, and Add/Sub traits aren't implemented at all.

EdHastingsCasperAssociation (changes requested):
I don't agree with the scope of this change. Will discuss directly.

PR #1983, opened 2021-08-23
New outgoing metrics

marc-casperlabs:
This adds new metrics for outgoing connection states.


goral09:
Is outgoing symmetric with incoming? If not, will there be a followup PR to add the incoming metrics or do you think they're less important?

node/src/components/small_network/outgoing.rs
```diff
@@ -426,6 +426,14 @@ where
         self.routes.keys().cloned()
     }
 
+    /// Iterates over all outgoing connections.
+    #[allow(clippy::needless_lifetimes)]
```
goral09: Curiosity box: Why do you need this annotation? Would `clippy` complain that you should remove the lifetime but it wouldn't actually work?

goral09 (approved):
Code looks correct, I hope you remembered about all the places where the metrics have to be updated :)

node/src/components/small_network/outgoing.rs
Fraser999: It compiles for me if I just remove all the lifetimes.

Fraser999 (commented):
I think it'd be safer to mutate the metrics inside the `OutgoingManager`'s mutable methods.  This seems more bulletproof to me, and less importantly, it'll allow us to avoid exposing the `outgoing` module to the whole crate.

It'll make it harder to forget to update callsites when `redeem_addr()` starts getting used.

If you disagree, then I believe we have a couple of places where we should also consider updating the metrics:
* in the constructor, we call `component.outgoing_manager.learn_addr`
* while handling `OutgoingConnection::Loopback` we call `outgoing_manager.handle_dial_outcome`
* while handling `OutgoingConnection::Established` we call `outgoing_manager.handle_dial_outcome`
* in `handle_outgoing_dropped` we call `outgoing_manager.handle_connection_drop`

marc-casperlabs:
@Fraser999 Thanks. I was able to also remove the other disabling of the clippy lint by introducing more `'_` lifetimes. I am not sure if this has to do something with the newer compiler version or if I just missed it the first time around.

goral09:
@marc-casperlabs @Fraser999 what's the status of this PR?

marc-casperlabs:
Closed until reworked.

marc-casperlabs:
This PR has been fundamentally changed, now the metrics are updated inside the outgoing manager, in a single spot, solving the issue of potentially forgetting to add an update call. It also improves efficiency, as only the affected metrics are changed (before all connections were iterated on every change).

node/src/components/small_network/metrics.rs
```diff
@@ -212,6 +255,17 @@ impl Metrics {
             debug!("not recording metrics, component already shut down");
         }
     }
+
+    /// Creates a set of outgoing metrics that is connected to this set of metrics.
+    pub(super) fn create_outgoing_metrics(&self) -> OutgoingMetrics {
```
goral09, line 260: So we now have the same metrics in two places? Can it happen that the same event will update them twice b/c it will be handled on multiple layers that each can modify it?

goral09, line 260: So we now have the same metrics in two places? Can it happen that the same event will update them twice b/c it will be handled on multiple layers that each can modify it?

goral09, line 260: So we now have the same metrics in two places? Can it happen that the same event will update them twice b/c it will be handled on multiple layers that each can modify it?

goral09: Sorry, GH was down and unknowningly I've created 3 copies of the same comment.

marc-casperlabs: The metrics struct just hold a reference to a metric - it's a `Arc` around an `AtomicI64`, essentially.

The metrics are only updated inside the `outgoing` module.

goral09: I understand that `Metrics` holds an `Arc` but it means it can also be incremented/decremented there, correct? Which means that the same metric can be changed in two places?

marc-casperlabs: Yes, but that's true for all metrics?

goral09: Hmm, other metrics seem to be reachable only from a single place in the code.

marc-casperlabs: That's more of a coincidence :)

goral09: Not really, at least I was doing that on purpose and thought that it was designed that way as well :) Not a blocker, just something to be aware of and watch out for.

marc-casperlabs: Fair enough!

I'll wait for Fraser to greenlight this PR still, as he was the original reviewer that had concerns.

Fraser999 (approved):

PR #1985, opened 2021-08-23
Make run_equivocator_network test more reliable.

### LINKED ISSUE #1859: Non-det failure of `run_equivocator_network` test.

afck:
This is still flaky, possibly because sometimes the nodes only start a bit after the genesis timestamp. With this change, we delay messages for three rounds after the first message has been sent instead.

Locally, I cannot reproduce the failure at all.

Closes #1859


goral09 (approved):
🤞

xcthulhu (approved):
LGTM

PR #1990, opened 2021-08-24
Support PublicKey, AccountHash or URef as Client Transfer Target

### LINKED ISSUE #1920: Support public key for transfer target [client]
* remove parse_public_key and use a match inside `Deploy::new_transfer`
* update the help string for transfer: target account can be a uref, account hash or a public key.
* Update the docs in client/lib.rs to 
* all of this applies to `transfer` and `make-transfer` command verbs.

nilz3ro:
# About These Changes
* Client now supports formatted strings of `PublicKey`, `AccountHash` and `URef` for transfer target.
* Closes #1920


goral09:
IIRC, execution engine would also accept a `ByteArray(u32)` that should be an address. Yes, if you look [here](https://github.com/casper-network/casper-node/blob/release-1.3.2/execution_engine/src/core/engine_state/transfer.rs#L193), you will find the allowed types.

nilz3ro:
@goral09, that is a good point, could a `[u8; 32]` be validated as an address by the client, or would this validation happen in the code that you linked?

goral09:
> @goral09, that is a good point, could a `[u8; 32]` be validated as an address by the client, or would this validation happen in the code that you linked?

I'm not really sure why we even allow a type like that to be used (cc @EdHastingsCasperLabs @mpapierski @darthsiroftardis ?) but I think putting the validation in the client would be a good idea. No point sending a deploy that would fail anyway. Although, that can be done in a separate PR as well, I just wanted to point out the gap.

client/lib/deploy.rs
```diff
@@ -273,9 +272,23 @@ impl DeployExt for Deploy {
...
         }
 
-        let target_account_hash = target_account.to_account_hash().value();
-        transfer_args.insert(TRANSFER_ARG_TARGET, target_account_hash)?;
+        if let Ok(public_key) = PublicKey::from_hex(target_account) {
+            transfer_args.insert(TRANSFER_ARG_TARGET, public_key)?;
+        } else if let Ok(account_hash) = AccountHash::from_formatted_str(target_account) {
+            transfer_args.insert(TRANSFER_ARG_TARGET, account_hash)?;
+        } else if let Ok(uref) = URef::from_formatted_str(target_account) {
+            transfer_args.insert(TRANSFER_ARG_TARGET, uref)?;
```
goral09, line 280: This asymmetricity is something that I don't like personally:
* for `AccountHash` we expect the input to follow the pattern of `account-hash-XYZ`,
* for `URef` `uref-XYZ`,
* but for `PublicKey` just `xyz`.

I would really consider adding `public-key-XYZ` prefix to make it consistent with other options.

client/lib/deploy.rs
```diff
@@ -273,9 +272,23 @@ impl DeployExt for Deploy {
...
+        } else {
+            return Err(Error::InvalidArgument {
+                context: "target_account",
+                error: format!(
+                    "Could not parse value as a PublicKey, AccountHash or URef: {}",
```
goral09:
```suggestion
                    "Allowed types: PublicKey, AccountHash or URef, got {}",
```
I would just make the error message more surgical/sterile.

goral09 (commented):
We also need to update a CHANGELOG.

client/lib/deploy.rs
nilz3ro: @goral09 I agree with you, I think that would be more consistent and clearer to users.

Fraser999: I don't think we should do that in this PR I'm afraid.

I'm not against a change like that, but I think it should be planned out so we know we're being consistent across all our APIs.  For example, we'd probably want to change the format of the public_key_hex to match this.  It has a pretty far-reaching impact.

goral09: We are just adding support for public key so we're not changing anything yet.

Fraser999 (approved):

Fraser999 (approved):

PR #1991, opened 2021-08-25
Fix 1980 sse event for expired deploys

### LINKED ISSUE #1980: Emit event to SSE when deploy expires
This needs a few changes:
* create a new announcement enum in node/src/effect/announcements.rs named `BlockProposerAnnouncement` with one variant `DeploysExpired(Vec<DeployHash>)`
* create a new `EffectBuilder` method `announce_expired_deploys()` which takes a `Vec<DeployHash>` and schedules a `BlockProposerAnnouncement::DeploysExpired` (see other effect builder `announce_` methods for ref.)
* change [`deploy_sets::prune_deploys()` and `deploy_sets::prune_pending_deploys`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/block_proposer/deploy_sets.rs#L68-L88) to return the `DeployHash`es of all pruned deploys.  (Also reduce the functions' visibility to private and update the doc comments.)
* change [`BlockProposerDeploySets::prune()`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/block_proposer/deploy_sets.rs#L60-L65) to return the `DeployHash`es of all pruned deploys.  (Also reduce the function's visibility to `pub(super)` and update the doc comment.)
* change the [`BlockProposerReady::handle_event()` for `Event::Prune`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/block_proposer.rs#L266) to take the full list of pruned deploy hashes returned by `self.prune()` and call `effect_builder.announce_expired_deploys()`
* add a new variant `DeployExpired(DeployHash)` to [`event_stream_server::Event`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server/event.rs#L7-L27) immediately after `DeployProcessed`
* add a new variant `DeployExpired { deploy_hash: DeployHash }` to [`sse_server::SseData`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server/sse_server.rs#L76-L117) immediately after `DeployProcessed`
* update [`EventStreamServer::handle_event`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server.rs#L182) to have an `Event::DeployExpired` trigger `self.broadcast(SseData::DeployExpired ...)`
* update [`sse_server::EventFilter`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server/sse_server.rs#L254-L263) to have a `DeployExpired` variant and update `SseData::should_include()` accordingly
* update [`sse_server::MAIN_FILTER`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server/sse_server.rs#L58-L64) to include an instance of `EventFilter::DeployExpired`
* handle [as per `DeployProcessed` in `sse_server::filter_map_server_sent_event()`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server/sse_server.rs#L301)
* update the unit tests `should_filter_events_with_valid_ids` and `should_filter_events_with_invalid_ids` in sse_server.rs to cover the new event variant.  This will include provision of a test-only function similar to [`SseData::random_deploy_accepted()`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server/sse_server.rs#L154-L161)  
* update the unit tests in event_stream_server/tests.rs to have the new event variant available.  This should simply be a matter of updating [`TestFixture::new()`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/components/event_stream_server/tests.rs#L208-L219) to include the new variant in `events`
* finally, wire up the new announcement to trigger the required event by adding a new variant to [`ParticipatingEvent`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/reactor/participating.rs#L187), and handling it similarly to the [`LinearChainAnnouncement`](https://github.com/casper-network/casper-node/blob/8c06cf8394c1aeada6cfaaa0288df1a9454b4219/node/src/reactor/participating.rs#L1116-L1123)

rafal-ch:
Added new event to the main SSE server stream accessed via `<IP:Port>/events/main` which emits hashes of expired deploys.

Example:
```
data:{"DeployExpired":{"deploy_hash":"ba2ab9fdf43ed7a40c2a813f505fdb61a16f04a9a866409ac9b7c6d88af7c70e"}}
```

Closes #1980.

### DIFF
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -101,6 +101,8 @@ pub enum SseData {
         #[data_size(skip)]
         execution_result: Box<ExecutionResult>,
     },
+    /// TODO[RC]
+    DeployExpired { deploy_hash: Box<DeployHash> },
     /// Generic representation of validator's fault in an era.
     Fault {
         era_id: EraId,
@@ -123,6 +125,7 @@ impl SseData {
             SseData::BlockAdded { .. } => filter.contains(&EventFilter::BlockAdded),
             SseData::DeployAccepted { .. } => filter.contains(&EventFilter::DeployAccepted),
             SseData::DeployProcessed { .. } => filter.contains(&EventFilter::DeployProcessed),
+            SseData::DeployExpired { .. } => filter.contains(&EventFilter::DeployExpired),
             SseData::Fault { .. } => filter.contains(&EventFilter::Fault),
             SseData::FinalitySignature(_) => filter.contains(&EventFilter::FinalitySignature),
             SseData::Step { .. } => filter.contains(&EventFilter::Step),
@@ -171,6 +174,14 @@ impl SseData {
             dependencies: deploy.header().dependencies().clone(),
             block_hash: Box::new(BlockHash::random(rng)),
             execution_result: Box::new(rng.gen()),
+        }
+    }
+
+    /// TODO[RC]
+    pub(super) fn random_deploy_expired(rng: &mut TestRng) -> Self {
+        let deploy = Deploy::random(rng);
+        SseData::DeployExpired {
+            deploy_hash: Box::new(*deploy.id()), // TODO[RC] Is the hash enough?
         }
     }
 
@@ -257,6 +268,7 @@ pub(super) enum EventFilter {
     BlockAdded,
     DeployAccepted,
     DeployProcessed,
+    DeployExpired,
     Fault,
     FinalitySignature,
     Step,
@@ -299,6 +311,7 @@ async fn filter_map_server_sent_event(
 
         &SseData::BlockAdded { .. }
         | &SseData::DeployProcessed { .. }
+        | &SseData::DeployExpired { .. }
         | &SseData::Fault { .. }
         | &SseData::Step { .. }
         | &SseData::FinalitySignature(_) => Some(Ok(WarpServerSentEvent::default()
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -41,7 +41,7 @@ use crate::types::{
     BlockHash, Deploy, DeployHash, FinalitySignature, JsonBlock, TimeDiff, Timestamp,
 };
 #[cfg(test)]
-use crate::{crypto::AsymmetricKeyExt, testing::TestRng, types::Block};
+use crate::{crypto::AsymmetricKeyExt, testing, testing::TestRng, types::Block};
 
 /// The URL root path.
 pub const SSE_API_ROOT_PATH: &str = "events";
@@ -180,7 +180,7 @@ impl SseData {
 
     /// Returns a random `SseData::DeployExpired`
     pub(super) fn random_deploy_expired(rng: &mut TestRng) -> Self {
-        let deploy = Deploy::random(rng);
+        let deploy = testing::create_expired_deploy(Timestamp::now(), rng);
         SseData::DeployExpired {
             deploy_hash: *deploy.id(),
         }
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -102,7 +102,7 @@ pub enum SseData {
         #[data_size(skip)]
         execution_result: Box<ExecutionResult>,
     },
-    /// TODO[RC]
+    /// The given deploy has expired.
     DeployExpired { deploy_hash: DeployHash },
     /// Generic representation of validator's fault in an era.
     Fault {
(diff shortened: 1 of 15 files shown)


rafal-ch:
Moving back to Draft because there will be additional work to cover some potentially missing expiration accouncements.

rafal-ch:
Reopening the PR for review. Additional scenarios for sending the expiration event will by handled by separate PRs.

node/src/components/event_stream_server/sse_server.rs
```diff
@@ -174,6 +178,14 @@ impl SseData {
         }
     }
 
+    /// Returns a random `SseData::DeployExpired`
+    pub(super) fn random_deploy_expired(rng: &mut TestRng) -> Self {
+        let deploy = Deploy::random(rng);
```
goral09: Maybe we can make sure that the example deploy would have indeed expired: `timestamp + ttl < now`.

rafal-ch: Nice catch ;)
Addressed in 832961befc4a230029dcf15dac84a5407c9433f8

node/src/components/event_stream_server/tests.rs
```diff
@@ -590,6 +593,7 @@ async fn should_serve_events_with_no_query(path: &str) {
     let url = url(server_address, path, None);
     let (expected_events, final_id) = fixture.all_filtered_events(path);
     let received_events = subscribe(&url, barrier, final_id, "client").await.unwrap();
+
```
darthsiroftardis: Spurious new line???

node/src/effect.rs
```diff
@@ -603,6 +603,19 @@ impl<REv> EffectBuilder<REv> {
         .await
     }
 
+    /// Announces that deploys have expired.
```
darthsiroftardis:
```suggestion
    /// Announces which deploys have expired.
```

node/src/reactor/participating.rs
```diff
@@ -1113,6 +1120,18 @@ impl reactor::Reactor for Reactor {
...
+            ) => hashes
+                .into_iter()
+                .map(|hash| {
+                    let reactor_event = ParticipatingEvent::EventStreamServer(
+                        event_stream_server::Event::DeployExpired(hash),
+                    );
+                    self.dispatch_event(effect_builder, rng, reactor_event)
+                })
+                .flatten()
+                .collect(),
```
darthsiroftardis: So, I am wondering if its possible to have the event stream server, do the flattening of the deploy hashes on its side. Here I feel like we're pumping the event queue with a bunch of events, whereas I feel like we should have the Event server collect all the hashes and then handle the individual hashes on its part of the component boundary. 

@Fraser999 thoughts???

darthsiroftardis (commented):
I have one point about how we dispatch the `event_stream_server` Event itself. Idm approving once thats addressed

node/src/reactor/participating.rs
rafal-ch: @Fraser999 
Although we initially agreed that we can live with the current implementation, I'm keen to change it, given we have the comment from @darthsiroftardis .

It all boils down to how many such events we expect to get and if it's better to have a single, potentially big event or several smaller ones.

node/src/components/event_stream_server/tests.rs
rafal-ch: Yup, removed.

node/src/effect.rs
Fraser999: Looks like Karan's suggestion didn't get actioned?

node/src/reactor/participating.rs
Fraser999: Yes, I agree with Karan.

We want to minimise the work being done in `dispatch_event` and offload it to the event stream server.

node/src/effect.rs
rafal-ch: It did, hence resolved.
Sorry for the confusion, next time I'll refrain from resolving until all changes are actually pushed to the repo.

node/src/reactor/participating.rs
rafal-ch: Fixed.
The participating reactor now flushes a vector of expired hashes to the event stream server, which then creates appropriate events on its own.

Fraser999 (approved):

darthsiroftardis (approved):

PR #1994, opened 2021-08-25
Add read_trie JSON-RPC endpoint

### LINKED ISSUE #1993: Create Block Execution Benchmark System
We need a way of benchmarking and testing block execution.

This would enable future capability to test executing deploys before submitting them to mainnet.

dwerner:
Resolves part of #1993 (1/2)

The purpose of this PR is to:

- Add a JSON-RPC endpoint to read a trie as a base64 encoded blob (for backward compatibility).


dwerner:
Closing in favor of https://github.com/casper-network/casper-node/pull/2078

PR #1995, opened 2021-08-25
Archival Sync

### LINKED ISSUE #1984: Gap filling strategy
We need to have a way of having a node securely download all state going back to genesis (as well as moving forward).

This ticket implements https://github.com/casper-network/ceps/issues/62

xcthulhu:
Introduces archival node functionality for fast sync.

Closes #1984

### DIFF
--- a/node/src/components/linear_chain_sync/error.rs
+++ b/node/src/components/linear_chain_sync/error.rs
@@ -7,7 +7,7 @@ use casper_execution_engine::{
     shared::stored_value::StoredValue,
     storage::trie::Trie,
 };
-use casper_types::{system::auction::EraValidators, EraId, Key, ProtocolVersion};
+use casper_types::{EraId, Key, ProtocolVersion};
 
 use crate::{
     components::{
@@ -123,15 +123,8 @@ pub(crate) enum LinearChainSyncError {
         trusted_block_header: Box<BlockHeader>,
     },
 
-    #[error(
-        "Genesis block global state did not contain genesis validators. \
-         Genesis block header: {genesis_block_header} \
-         Genesis era validators: {era_validators:?}"
-    )]
-    GenesisGlobalStateDidNotHaveGenesisValidators {
-        genesis_block_header: Box<BlockHeader>,
-        era_validators: EraValidators,
-    },
+    #[error("Hit genesis block trying to get trusted era validators.")]
+    HitGenesisBlockTryingToGetTrustedEraValidators { trusted_header: BlockHeader },
 
     /// Error getting era validators from the execution engine.
     #[error(transparent)]
--- a/node/src/components/linear_chain_sync/error.rs
+++ b/node/src/components/linear_chain_sync/error.rs
@@ -3,14 +3,16 @@ use std::fmt::Debug;
 use thiserror::Error;
 
 use casper_execution_engine::{
-    core::engine_state, shared::stored_value::StoredValue, storage::trie::Trie,
+    core::{engine_state, engine_state::GetEraValidatorsError},
+    shared::stored_value::StoredValue,
+    storage::trie::Trie,
 };
-use casper_types::{EraId, Key, ProtocolVersion};
+use casper_types::{system::auction::EraValidators, EraId, Key, ProtocolVersion};
 
 use crate::{
     components::{
         consensus::error::FinalitySignatureError, contract_runtime::BlockExecutionError,
-        fetcher::FetcherError,
+        fetcher::FetcherError, linear_chain_sync::operations::KeyBlockInfo,
     },
     crypto,
     types::{
@@ -31,6 +33,16 @@ pub(crate) enum SignatureValidationError {
         block_header: Box<BlockHeader>,
         block_hash: Box<BlockHash>,
         block_signatures: Box<BlockSignatures>,
+    },
+
+    #[error(
+        "Block header is in wrong era. \
+         block header: {block_header:?} \
+         trusted key block info: {trusted_key_block_info:?}"
+    )]
+    HeaderIsInWrongEra {
+        block_header: Box<BlockHeader>,
+        trusted_key_block_info: Box<KeyBlockInfo>,
     },
 
     #[error(transparent)]
@@ -112,10 +124,16 @@ pub(crate) enum LinearChainSyncError {
     },
 
     #[error(
-        "Can't download block before genesis. \
-         Genesis block header: {genesis_block_header}"
+        "Genesis block global state did not contain genesis validators. \
+         Genesis block header: {genesis_block_header} \
+         Genesis era validators: {era_validators:?}"
     )]
-    CantDownloadBlockBeforeGenesis {
+    GenesisGlobalStateDidNotHaveGenesisValidators {
         genesis_block_header: Box<BlockHeader>,
+        era_validators: EraValidators,
     },
+
+    /// Error getting era validators from the execution engine.
+    #[error(transparent)]
+    GetEraValidatorsError(#[from] GetEraValidatorsError),
 }
--- a/node/src/components/linear_chain_sync/error.rs
+++ b/node/src/components/linear_chain_sync/error.rs
@@ -17,7 +17,7 @@ use crate::{
     crypto,
     types::{
         Block, BlockHash, BlockHeader, BlockHeaderWithMetadata, BlockSignatures, BlockWithMetadata,
-        Deploy, NodeId, TimeDiff, Timestamp,
+        Deploy, NodeId,
     },
 };
 
@@ -129,19 +129,4 @@ pub(crate) enum LinearChainSyncError {
     /// Error getting era validators from the execution engine.
     #[error(transparent)]
     GetEraValidatorsError(#[from] GetEraValidatorsError),
-
-    #[error(
-        "Trusted block header has expired. \
-         You will need to get a recent block hash to join the network. \
-         Current time: {current_time}, \
-         Trusted hash time to expiration: {trusted_hash_time_to_expiration:?}, \
-         Trusted hash: {trusted_hash:?}, \
-         Trusted block header: {trusted_block_header:?}"
-    )]
-    TrustedBlockHeaderHasExpired {
-        current_time: Timestamp,
-        trusted_hash_time_to_expiration: TimeDiff,
-        trusted_hash: BlockHash,
-        trusted_block_header: Box<BlockHeader>,
-    },
 }
--- a/node/src/components/linear_chain_sync/error.rs
+++ b/node/src/components/linear_chain_sync/error.rs
@@ -17,7 +17,7 @@ use crate::{
     crypto,
     types::{
         Block, BlockHash, BlockHeader, BlockHeaderWithMetadata, BlockSignatures, BlockWithMetadata,
-        Deploy, NodeId,
+        Deploy, NodeId, TimeDiff, Timestamp,
     },
 };
 
@@ -129,4 +129,19 @@ pub(crate) enum LinearChainSyncError {
     /// Error getting era validators from the execution engine.
     #[error(transparent)]
     GetEraValidatorsError(#[from] GetEraValidatorsError),
+
+    #[error(
+        "Trusted block header has expired. \
+         You will need to get an recent block hash to join the network. \
+         Current time: {current_time}, \
+         Trusted hash time to expiration: {trusted_hash_time_to_expiration:?}, \
+         Trusted hash: {trusted_hash:?}, \
+         Trusted block header: {trusted_block_header:?}"
+    )]
+    TrustedBlockHeaderHasExpired {
+        current_time: Timestamp,
+        trusted_hash_time_to_expiration: TimeDiff,
+        trusted_hash: BlockHash,
+        trusted_block_header: Box<BlockHeader>,
+    },
 }
--- a/node/src/components/linear_chain_sync/error.rs
+++ b/node/src/components/linear_chain_sync/error.rs
@@ -132,7 +132,7 @@ pub(crate) enum LinearChainSyncError {
 
     #[error(
         "Trusted block header has expired. \
-         You will need to get an recent block hash to join the network. \
+         You will need to get a recent block hash to join the network. \
          Current time: {current_time}, \
          Trusted hash time to expiration: {trusted_hash_time_to_expiration:?}, \
          Trusted hash: {trusted_hash:?}, \
--- a/node/src/components/linear_chain_sync/operations.rs
+++ b/node/src/components/linear_chain_sync/operations.rs
@@ -22,7 +22,8 @@ use crate::{
     reactor::joiner::JoinerEvent,
     types::{
         Block, BlockHash, BlockHeader, BlockHeaderWithMetadata, BlockSignatures, BlockWithMetadata,
-        Chainspec, Deploy, DeployHash, FinalizedBlock, Item, NodeConfig, NodeId, Timestamp,
+        Chainspec, Deploy, DeployHash, FinalizedBlock, Item, NodeConfig, NodeId, TimeDiff,
+        Timestamp,
     },
 };
 
@@ -618,16 +619,20 @@ pub(crate) async fn run_fast_sync_task(
     // Fetch the trusted header
     let trusted_block_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
 
-    if let Some(trusted_hash_time_to_expiration) = node_config.trusted_hash_time_to_expiration {
-        let current_time = Timestamp::now();
-        if trusted_block_header.timestamp() + trusted_hash_time_to_expiration < current_time {
-            return Err(LinearChainSyncError::TrustedBlockHeaderHasExpired {
-                current_time,
-                trusted_hash_time_to_expiration,
-                trusted_hash,
-                trusted_block_header,
-            });
-        }
+    let era_duration: TimeDiff = std::cmp::max(
+        chainspec.highway_config.min_round_length() * chainspec.core_config.minimum_era_height,
+        chainspec.core_config.era_duration,
+    );
+
+    if trusted_block_header.timestamp()
+        + era_duration
+            * chainspec
+                .core_config
+                .unbonding_delay
+                .saturating_sub(chainspec.core_config.auction_delay)
+        < Timestamp::now()
+    {
+        warn!(?trusted_block_header, "Timestamp of trusted hash is older than era_duration * (auction_delay - auction_delay)");
     }
 
     let maybe_last_emergency_restart_era_id = chainspec.protocol_config.last_emergency_restart;
--- a/node/src/components/linear_chain_sync/operations.rs
+++ b/node/src/components/linear_chain_sync/operations.rs
@@ -1,4 +1,4 @@
-use std::{collections::BTreeMap, time::Duration};
+use std::{collections::BTreeMap, sync::Arc, time::Duration};
 
 use num::rational::Ratio;
 use tracing::{info, trace, warn};
@@ -7,25 +7,26 @@ use casper_execution_engine::{
     shared::{newtypes::Blake2bHash, stored_value::StoredValue},
     storage::trie::Trie,
 };
-use casper_types::{Key, PublicKey, U512};
+use casper_types::{EraId, Key, PublicKey, U512};
 
 use crate::{
     components::{
         consensus,
         consensus::check_sufficient_finality_signatures,
-        contract_runtime::ExecutionPreState,
+        contract_runtime::{EraValidatorsRequest, ExecutionPreState},
         fetcher::{FetchedData, FetcherError},
         linear_chain_sync::error::{LinearChainSyncError, SignatureValidationError},
     },
+    crypto::hash::Digest,
     effect::{requests::FetcherRequest, EffectBuilder},
     reactor::joiner::JoinerEvent,
     types::{
         Block, BlockHash, BlockHeader, BlockHeaderWithMetadata, BlockSignatures, BlockWithMetadata,
-        Chainspec, Deploy, DeployHash, FinalizedBlock, Item, NodeId, Timestamp,
+        Chainspec, Deploy, DeployHash, FinalizedBlock, Item, NodeConfig, NodeId, Timestamp,
     },
 };
 
-const TIMEOUT_DURATION: Duration = Duration::from_millis(100);
+const SLEEP_DURATION_SO_WE_DONT_SPAM: Duration = Duration::from_millis(100);
 
 /// Fetches an item. Keeps retrying to fetch until it is successful. Assumes no integrity check is
 /// necessary for the item. Not suited to fetching a block header or block by height, which require
@@ -79,7 +80,7 @@ where
                 Err(error) => return Err(error),
             }
         }
-        tokio::time::sleep(TIMEOUT_DURATION).await
+        tokio::time::sleep(SLEEP_DURATION_SO_WE_DONT_SPAM).await
     }
 }
 
@@ -114,10 +115,17 @@ async fn fetch_and_store_deploy(
 /// Verifies finality signatures for a block header
 fn validate_finality_signatures(
     block_header: &BlockHeader,
-    trusted_validator_weights: &BTreeMap<PublicKey, U512>,
+    trusted_key_block_info: &KeyBlockInfo,
     finality_threshold_fraction: Ratio<u64>,
     block_signatures: &BlockSignatures,
 ) -> Result<(), SignatureValidationError> {
+    if block_header.era_id() != trusted_key_block_info.era_id {
+        return Err(SignatureValidationError::HeaderIsInWrongEra {
+            block_header: Box::new(block_header.clone()),
+            trusted_key_block_info: Box::new(trusted_key_block_info.clone()),
+        });
+    }
+
     // Check the signatures' block hash is the header's block hash
     let block_hash = block_header.hash();
     if block_signatures.block_hash != block_hash {
@@ -134,7 +142,7 @@ fn validate_finality_signatures(
     block_signatures.verify()?;
 
     check_sufficient_finality_signatures(
-        trusted_validator_weights,
+        &trusted_key_block_info.validator_weights,
         finality_threshold_fraction,
         block_signatures,
     )
@@ -145,7 +153,7 @@ fn validate_finality_signatures(
 async fn fetch_and_store_block_header_by_height(
     effect_builder: EffectBuilder<JoinerEvent>,
     height: u64,
-    trusted_validator_weights: &BTreeMap<PublicKey, U512>,
+    trusted_key_block_info: &KeyBlockInfo,
     finality_threshold_fraction: Ratio<u64>,
 ) -> Result<Option<Box<BlockHeaderWithMetadata>>, FetcherError<BlockHeaderWithMetadata, NodeId>> {
     for peer in effect_builder.get_peers_in_random_order().await {
@@ -162,7 +170,7 @@ async fn fetch_and_store_block_header_by_height(
 
                 if let Err(error) = validate_finality_signatures(
                     &block_header,
-                    trusted_validator_weights,
+                    trusted_key_block_info,
                     finality_threshold_fraction,
                     &block_signatures,
                 ) {
@@ -213,26 +221,85 @@ async fn fetch_and_store_block_header_by_height(
     Ok(None)
 }
 
-pub(super) struct KeyBlockInfo {
+/// Key block info used for verifying finality signatures.
+///
+/// Can come from either:
+///   - A switch block
+///   - The global state under a genesis block
+///
+/// If the data was scraped from genesis, then `era_id` is 0.
+/// Otherwise if it came from a switch block it is that switch's block `era_id + 1`.
+#[derive(Clone, Debug)]
+pub(crate) struct KeyBlockInfo {
+    /// The block hash of the key block
+    key_block_hash: BlockHash,
+    /// The validator weights for checking finality signatures.
     validator_weights: BTreeMap<PublicKey, U512>,
+    /// The time the era started.
     era_start: Timestamp,
+    /// The height of the switch block that started this era.
     height: u64,
+    /// The era in which the validators are operating
+    era_id: EraId,
 }
 
 impl KeyBlockInfo {
-    pub(super) fn maybe_from_block_header(block_header: &BlockHeader) -> Option<KeyBlockInfo> {
+    fn maybe_from_block_header(block_header: &BlockHeader) -> Option<KeyBlockInfo> {
         block_header
             .next_era_validator_weights()
             .map(|next_era_validator_weights| KeyBlockInfo {
+                key_block_hash: block_header.hash(),
                 validator_weights: next_era_validator_weights.clone(),
                 era_start: block_header.timestamp(),
                 height: block_header.height(),
+                era_id: block_header.era_id() + 1,
             })
     }
 }
 
-/// Get trusted switch block.
-/// Returns an error if we are still in the first era, or the era after an emergency restart.
+/// Get the trusted key block info for genesis.
+async fn get_genesis_trusted_key_block_info(
+    effect_builder: EffectBuilder<JoinerEvent>,
+    mut trusted_header: BlockHeader,
+) -> Result<KeyBlockInfo, LinearChainSyncError> {
+    // Get the genesis block header
+    while trusted_header.height() != 0 {
+        let block =
+            fetch_and_store_block_by_hash(effect_builder, *trusted_header.parent_hash()).await?;
+        trusted_header = block.take_header();
+    }
+
+    // Otherwise, sync the global trie state under genesis to get the genesis validators
+    let state_root_hash = trusted_header.state_root_hash();
+    sync_trie_store(effect_builder, *state_root_hash).await?;
+    let mut era_validators = effect_builder
+        .get_era_validators_from_contract_runtime(EraValidatorsRequest::new(
+            Blake2bHash::from(*state_root_hash),
+            trusted_header.protocol_version(),
+        ))
+        .await?;
+    let genesis_validators = match era_validators.remove(&EraId::new(0)) {
+        Some(genesis_validators) => genesis_validators,
+        None => {
+            return Err(
+                LinearChainSyncError::GenesisGlobalStateDidNotHaveGenesisValidators {
+                    genesis_block_header: Box::new(trusted_header),
+                    era_validators,
+                },
+            )
+        }
+    };
+
+    Ok(KeyBlockInfo {
+        key_block_hash: trusted_header.hash(),
+        validator_weights: genesis_validators,
+        era_start: trusted_header.timestamp(),
+        height: 0,
+        era_id: EraId::new(0),
+    })
+}
+
+/// Get the trusted key block info for a trusted block header.
 async fn get_trusted_key_block_info(
     effect_builder: EffectBuilder<JoinerEvent>,
     chainspec: &Chainspec,
@@ -264,10 +331,13 @@ async fn get_trusted_key_block_info(
         }
 
         if current_header_to_walk_back_from.height() == 0 {
-            return Err(LinearChainSyncError::CantDownloadBlockBeforeGenesis {
-                genesis_block_header: Box::new(current_header_to_walk_back_from),
-            });
+            return get_genesis_trusted_key_block_info(
+                effect_builder,
+                current_header_to_walk_back_from,
+            )
+            .await;
         }
+
         current_header_to_walk_back_from = *fetch_and_store_block_header(
             effect_builder,
             *current_header_to_walk_back_from.parent_hash(),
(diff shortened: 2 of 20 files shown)


goral09:
This PR has 20 commits (including **Sunset legacy sync**), are they all required for _archival node_ feature? Can this PR be chopped up into smaller ones to make the reviewing easier?

xcthulhu:
> This PR has 20 commits (including **Sunset legacy sync**), are they all required for _archival node_ feature? Can this PR be chopped up into smaller ones to make the reviewing easier?

In the course of this PR, I identified an issue with the fetcher.  It must always check if it has a sufficient amount of finality signatures (even when fetching from storage).  That could be broken out.

The changes to the `ChainspecLoader` properly belong to the #1917 .  I will move them.

xcthulhu:
@goral09 I created #1995 with the changes to the fetchers/storage (where we enforce we get sufficient finality signatures)

node/src/types/node_config.rs
```diff
@@ -3,11 +3,22 @@ use serde::{Deserialize, Serialize};
...
+
+    /// Whether to run in archival-sync mode. Archival-sync mode captures all data (blocks, deploys
+    /// and global state) back to genesis. Defaults to `false`.
+    #[serde(default = "NodeConfig::default_archival_sync_status")]
+    pub archival_sync: bool,
+}
+
+impl NodeConfig {
+    fn default_archival_sync_status() -> bool {
+        false
```
xcthulhu: ~This should default to `true` as discussed.~

As mentioned [below][1], this config option is not longer optional as of ba89b36.

[1]: https://github.com/casper-network/casper-node/pull/1995#discussion_r713357873

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -229,35 +221,90 @@ where
...
 
-pub(super) struct KeyBlockInfo {
+/// Key block info used for verifying finality signatures.
+///
+/// Can come from either:
+///   - A switch block
+///   - The global state under a genesis block
+///
+/// If the data was scraped from genesis, then `era_id` is 0.
+/// Otherwise if it came from a switch block it is that switch's block `era_id + 1`.
```
Fraser999:
```suggestion
/// Otherwise if it came from a switch block it is that switch block's `era_id + 1`.
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -229,35 +221,90 @@ where
...
+    mut trusted_header: BlockHeader,
+) -> Result<KeyBlockInfo, LinearChainSyncError> {
+    // Get the genesis block header
+    while trusted_header.height() != 0 {
+        let block =
+            fetch_and_store_block_by_hash(effect_builder, *trusted_header.parent_hash()).await?;
+        trusted_header = block.take_header();
+    }
+
+    // Otherwise, sync the global trie state under genesis to get the genesis validators
```
Fraser999:
```suggestion
    // Sync the global trie state under genesis to get the genesis validators
```

node/src/types/node_config.rs
```diff
@@ -3,11 +3,22 @@ use serde::{Deserialize, Serialize};
 
 use crate::types::BlockHash;
 
-/// Node configuration.
-#[derive(Default, DataSize, Debug, Deserialize, Serialize)]
+/// Node fast-sync configuration.
```
Fraser999, line 6: This edit has made me realise there _are_ only fast-sync related settings in `NodeConfig`.  I think for consistency, it'd make sense to move this code to the `linear_chain_sync` module, rename `NodeConfig` to just `Config`, export it in lib.rs as `LinearChainSyncConfig` (as per the other configs) and rename the field in the reactor config.

Happy for that to happen as a follow-up PR though.

node/src/types/node_config.rs
```diff
@@ -3,11 +3,22 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Default, DataSize, Debug, Deserialize, Serialize, Clone)]
 // Disallow unknown fields to ensure config files and command-line overrides contain valid keys.
 #[serde(deny_unknown_fields)]
 pub struct NodeConfig {
-    /// Hash used as a trust anchor when joining, if any.
+    /// Hash used as a trust anchor when joining. If no hash is provided, attempt to run genesis.
```
Fraser999: This could be painful for operators upgrading.  It's convenient to not have to provide a trusted hash after the initial version has run.

Again, possibly not something to change in this PR, but worth speaking to SRE about at least?

node/src/types/node_config.rs
```diff
@@ -3,11 +3,22 @@ use serde::{Deserialize, Serialize};
...
     pub trusted_hash: Option<BlockHash>,
+
+    /// Whether to run in archival-sync mode. Archival-sync mode captures all data (blocks, deploys
+    /// and global state) back to genesis. Defaults to `false`.
+    #[serde(default = "NodeConfig::default_archival_sync_status")]
```
Fraser999: I know we have a mixture of optional and non-optional fields in our config.toml files, but I think the vast majority are non-optional, so we should generally follow that approach.

One advantage is that we can't forget to add the field to the config.toml using that approach.

Fraser999 (commented):
Needs an update to the changelog to describe the new functionality and the associated config setting..

We also need to add `archival_sync` to all the config files except `utils/nctl/sh/scenarios/configs/gov96.config.toml`, even if it's commented out.

xcthulhu:
> Needs an update to the changelog to describe the new functionality and the associated config setting..
> 
> We also need to add `archival_sync` to all the config files except `utils/nctl/sh/scenarios/configs/gov96.config.toml`, even if it's commented out.

Should be fixed in dc54297a33984eda9d1c13fabe29beaebaad5d39

goral09:
This build is failing on `cargo fmt` Matt (@xcthulhu).

node/src/components/storage.rs
```diff
@@ -863,7 +863,10 @@ impl Storage {
                     &block_header,
                     false,
                 )? {
-                    error!("Could not insert block header: {}", block_header);
+                    error!(
+                        "Could not insert block header (maybe already inserted?): {}",
+                        block_header
+                    );
```
goral09:
```rust
                    error!(
                    ?block_header
                        "Could not insert block header (maybe already inserted?)",
                    );
```

otherwise it's not going to be indexed/displayed properly in Kibana.

resources/local/config.toml
```diff
@@ -6,6 +6,8 @@
 # If set, use this hash as a trust anchor when joining an existing network.
 #trusted_hash = 'HEX-FORMATTED BLOCK HASH'
 
+archival_sync = false
```
goral09: This should be `true` as well (as per https://github.com/casper-network/casper-node/pull/1995/files#r697618855).

resources/production/config-example.toml
```diff
@@ -5,7 +5,7 @@
 
 # If set, use this hash as a trust anchor when joining an existing network.
 #trusted_hash = 'HEX-FORMATTED BLOCK HASH'
-
+archival_sync = false
```
goral09:
```suggestion
archival_sync = true
```

utils/nctl/sh/scenarios/configs/bond_its.config.toml
```diff
@@ -6,6 +6,8 @@
 # If set, use this hash as a trust anchor when joining an existing network.
 #trusted_hash = 'HEX-FORMATTED BLOCK HASH'
 
+archival_sync = false
```
goral09:
```suggestion
archival_sync = true
```

utils/nctl/sh/scenarios/configs/itst13.config.toml
```diff
@@ -6,6 +6,8 @@
 # If set, use this hash as a trust anchor when joining an existing network.
 #trusted_hash = 'HEX-FORMATTED BLOCK HASH'
 
+archival_sync = false
```
goral09:
```suggestion
archival_sync = true
```

utils/nctl/sh/scenarios/configs/gov96.config.toml
```diff
@@ -10,6 +10,8 @@
 # If set, use this hash as a trust anchor when joining an existing network.
 #trusted_hash = 'HEX-FORMATTED BLOCK HASH'
 
+#archival_sync = false
```
goral09: Why did you choose to comment this one out?

node/src/types/node_config.rs
```diff
@@ -1,13 +1,22 @@
...
+    /// Hash used as a trust anchor when joining. If no hash is provided, attempt to run genesis.
     pub trusted_hash: Option<BlockHash>,
+
+    /// An (optional) time to expiration for the header for the trusted hash. If the timestamp of
+    /// the trusted hash is less than the current time minus this expiration time, the node will not
+    /// join the network.
+    pub trusted_hash_time_to_expiration: Option<TimeDiff>,
+
+    /// Whether to run in archival-sync mode. Archival-sync mode captures all data (blocks, deploys
+    /// and global state) back to genesis. Defaults to `false`.
```
xcthulhu:
```suggestion
    /// and global state) back to genesis.
```

node/src/components/linear_chain_sync/error.rs
```diff
@@ -111,11 +123,25 @@ pub(crate) enum LinearChainSyncError {
...
+
+    /// Error getting era validators from the execution engine.
+    #[error(transparent)]
+    GetEraValidatorsError(#[from] GetEraValidatorsError),
+
     #[error(
-        "Can't download block before genesis. \
-         Genesis block header: {genesis_block_header}"
+        "Trusted block header has expired. \
+         You will need to get an recent block hash to join the network. \
```
afck:
```suggestion
         You will need to get a recent block hash to join the network. \
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -371,39 +399,48 @@ async fn fetch_and_store_block_by_hash(
...
+    let mut outstanding_trie_keys = vec![Blake2bHash::from(state_root_hash)];
+    while let Some(trie_key) = outstanding_trie_keys.pop() {
+        let missing_descendant_trie_keys =
+            fetch_trie_and_insert_into_trie_store(effect_builder, trie_key).await?;
+        outstanding_trie_keys.extend(missing_descendant_trie_keys);
     }
+    Ok(())
+}
 
+/// Fetches the current header and fast-syncs too it.
```
afck:
```suggestion
/// Fetches the current header and fast-syncs to it.
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -371,39 +399,48 @@ async fn fetch_and_store_block_by_hash(
...
+///
+/// Performs the following:
+///
+///  1. Fetches the most recent block header.
+///  2. Fetches deploys for replay attacks and historical switch block info.
```
afck:
```suggestion
///  2. Fetches deploys for replay detection and historical switch block info.
```

(Or "prevention"?)

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +521,162 @@ pub(crate) async fn run_fast_sync_task(
...
+pub(crate) async fn run_fast_sync_task(
+    effect_builder: EffectBuilder<JoinerEvent>,
+    trusted_hash: BlockHash,
+    chainspec: Arc<Chainspec>,
+    node_config: NodeConfig,
+) -> Result<BlockHeader, LinearChainSyncError> {
+    // Fetch the trusted header
+    let trusted_block_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
+
+    if let Some(trusted_hash_time_to_expiration) = node_config.trusted_hash_time_to_expiration {
```
afck: This should be computed from the unbonding delay and auction delay, otherwise the configuration could go out of sync with the actual delay for withdrawing stake.

node/src/testing/multi_stage_test_reactor/test_chain.rs
```diff
@@ -383,6 +398,72 @@ async fn test_joiner_at_genesis() {
...
+            Some(switch_block_hash),
+            true,
+            &mut rng,
+        )
+        .await;
+
+    assert_eq!(
+        chain.network.nodes().len(),
+        2,
+        "There should be two validators in the network (one bonded and one read only)"
```
afck:
```suggestion
        "There should be two nodes in the network (one bonded validator and one read only)"
```

node/src/testing/multi_stage_test_reactor/test_chain.rs
```diff
@@ -383,6 +398,72 @@ async fn test_joiner_at_genesis() {
...
+    let synchronized_era = era_to_join + 1;
+    info!("Waiting for Era {} to end", synchronized_era);
+    chain
+        .network
+        .settle_on(
+            &mut rng,
+            has_passed_by_era(synchronized_era),
+            Duration::from_secs(600),
+        )
+        .await;
```
afck, line 492: Should we check at this point whether the added node has stored the first block?

node/src/types/node_config.rs
```diff
@@ -1,13 +1,22 @@
...
 #[serde(deny_unknown_fields)]
 pub struct NodeConfig {
-    /// Hash used as a trust anchor when joining, if any.
+    /// Hash used as a trust anchor when joining. If no hash is provided, attempt to run genesis.
     pub trusted_hash: Option<BlockHash>,
+
+    /// An (optional) time to expiration for the header for the trusted hash. If the timestamp of
+    /// the trusted hash is less than the current time minus this expiration time, the node will not
+    /// join the network.
+    pub trusted_hash_time_to_expiration: Option<TimeDiff>,
```
afck: Can we compute that from the unbonding delay instead?

node/src/types/node_config.rs
xcthulhu: I made this no longer optional in ba89b36

node/src/components/storage.rs
xcthulhu: Should be fixed in 752c1770d

resources/local/config.toml
xcthulhu: With the current implementation, this will cause nodes to take too long to join.

After implementing your proposal I have outlined in https://github.com/casper-network/casper-node/issues/2083 we can set these flags to `true`.

resources/production/config-example.toml
xcthulhu: With the current implementation, this will cause nodes to take too long to join.

After implementing your proposal I have outlined in https://github.com/casper-network/casper-node/issues/2083 we can set these flags to `true`.

utils/nctl/sh/scenarios/configs/bond_its.config.toml
xcthulhu: With the current implementation, this will cause nodes to take too long to join.

After implementing your proposal I have outlined in https://github.com/casper-network/casper-node/issues/2083 we can set these flags to `true`.

utils/nctl/sh/scenarios/configs/itst13.config.toml
xcthulhu: With the current implementation, this will cause nodes to take too long to join.

After implementing your proposal I have outlined in https://github.com/casper-network/casper-node/issues/2083 we can set these flags to `true`.

utils/nctl/sh/scenarios/configs/gov96.config.toml
xcthulhu: Didn't @Fraser999 say I should comment this out [here][1]?

[1]: https://github.com/casper-network/casper-node/pull/1995#pullrequestreview-750135646

node/src/types/node_config.rs
xcthulhu: I've reverted this comment in 5f69a6797.

In addition, in 7b2f376c2 we query storage for a block header if no hash is provided.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Hmmm... I'm not sure how to do this exactly.  Suppose I have a block hash.  How do I know if it's too old?  Do I trust it, get blocks by height to the tip, and then check if the block hash had expired?

node/src/types/node_config.rs
xcthulhu: You mentioned this above, and I am not sure how implement this.

Is the idea to set this to `era_length * unbonding_delay` rather than just a flat time diff?

node/src/testing/multi_stage_test_reactor/test_chain.rs
xcthulhu: I went ahead and wrote a check that the added node has the genesis state root in 3d06704826d7dc9a0ed492a72ebb8ce0b6565266

node/src/types/node_config.rs
afck: Or `minimum_era_length * (unbonding_delay - auction_delay)`… or, to be completely precise:

Let's try to find out whether a validator's signature in era `n` is still trustworthy. To do that, we'd have to find out the `start_time` of era `n`.
Since they were a validator in era `n` that means they can't have made an unbonding request before era `n - auction_delay`. If they made one in `n - auction_delay`, they will be able to withdraw their stake at the end of era `n - auction_delay + unbonding_delay`. (Is that correct?)
So the earliest time they could withdraw would be `start_time + minimum_era_length * (unbonding_delay + 1 - auction_delay)`.

That definitely needs to be double-checked to make sure I didn't make any off-by-one errors, and that I correctly understand how the unbonding delay works!

But I still think it would be safer to compute the trusted time this way, or to use some lower bound like `minimum_era_length * (unbonding_delay - auction_delay)`.

goral09: Technically, this doesn't have to be part of this PR - this one is *only* about archival nodes. We can solve that problem later in a different PR.

afck: Sure, it's not a blocker; it's just that we'd introduce a configuration option that we'd end up removing/deprecating again.

goral09: I meant to remove that change from this PR entirely so to not block otherwise correct PR.

node/src/testing/multi_stage_test_reactor/test_chain.rs
```diff
@@ -383,6 +426,77 @@ async fn test_joiner_at_genesis() {
...
+
+    chain
+        .network
+        .settle_on(&mut rng, all_have_genesis_state, Duration::from_secs(600))
```
goral09, line 496: Looks to me that this test assumes (maybe correctly) that joining node's state will be synchronized from the highest block until Genesis so this `all_have_genesis_state` is sufficient to confirm that archival functionality works. I am not sure that's enough though, what if the strategy was to start from Genesis and work the way towards the tip? Or worse, what if the joining node downloaded only the Genesis state? I think the assertion should be that the joining node has _all_ of the blocks and _all_ of the state referenced from these blocks.

resources/local/config.toml
goral09: Still, the archival sync should be enabled (`true`) by default.

node/src/types/node_config.rs
xcthulhu: This was removed in https://github.com/casper-network/casper-node/pull/1995/commits/fe6cd1be96a832d660077068a208a5eaa5f1135e

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -618,16 +619,20 @@ pub(crate) async fn run_fast_sync_task(
...
+
+    if trusted_block_header.timestamp()
+        + era_duration
+            * chainspec
+                .core_config
+                .unbonding_delay
+                .saturating_sub(chainspec.core_config.auction_delay)
+        < Timestamp::now()
+    {
+        warn!(?trusted_block_header, "Timestamp of trusted hash is older than era_duration * (auction_delay - auction_delay)");
```
afck:
```suggestion
        warn!(?trusted_block_header, "Timestamp of trusted hash is older than era_duration * (unbonding_delay - auction_delay)");
```
(Also: That line is too long.)

afck (approved):
`trusted_hash_time_to_expiration` also needs to be removed from the config files.

Otherwise LGTM.

node/src/testing/multi_stage_test_reactor/test_chain.rs
xcthulhu: Should be fixed in ceffeeea5489aa8d2dc50de60c30adcbac260108

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Should be fixed in https://github.com/casper-network/casper-node/pull/1995/commits/5f0f6fb9ebb706eb4a751c1e1c504e91791a0fb7

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Should be fixed in https://github.com/casper-network/casper-node/commit/fe6cd1be96a832d660077068a208a5eaa5f1135e

resources/local/config.toml
goral09: This should be set to `true`.

resources/production/config-example.toml
goral09: Same here, archival syncing should be enabled by default.

xcthulhu:
> `trusted_hash_time_to_expiration` also needs to be removed from the config files.
> 
> Otherwise LGTM.

Removed in https://github.com/casper-network/casper-node/pull/1995/commits/e30159c60f1f78c42129b8b26699187f9cf56afc

resources/production/config-example.toml
xcthulhu: Done in 209c9e95933dd452f349c0b9c1380487ab221d16

resources/local/config.toml
xcthulhu: Done in 209c9e95933dd452f349c0b9c1380487ab221d16

utils/nctl/sh/scenarios/configs/bond_its.config.toml
xcthulhu: Done in 209c9e95933dd452f349c0b9c1380487ab221d16

utils/nctl/sh/scenarios/configs/itst13.config.toml
xcthulhu: Done in 209c9e95933dd452f349c0b9c1380487ab221d16

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -371,39 +400,48 @@ async fn fetch_and_store_block_by_hash(
...
-                    trusted_hash,
-                    trusted_block_header,
-                },
-            );
-        }
-        _ => {}
+    state_root_hash: Digest,
+) -> Result<(), LinearChainSyncError> {
+    info!(
+        state_root_hash = ?state_root_hash,
```
goral09:
```suggestion
        ?state_root_hash,
```
that will work

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -371,39 +400,48 @@ async fn fetch_and_store_block_by_hash(
...
+        "Syncing trie store",
```
goral09:
```suggestion
        "syncing trie store",
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -455,7 +493,7 @@ pub(crate) async fn run_fast_sync_task(
         .era_start
         .saturating_diff(current_header.timestamp())
         < chainspec.deploy_config.max_ttl
-        && !current_header.is_genesis_child()
+        && current_header.height() != 0
```
goral09: What if we're just at the beginning of an era/shortly after the key block - we might miss some deploys as we will never go past the key block's era start timestamp. We will never download deploys that were included in blocks before that key block.

xcthulhu: @afck is the original author of this...

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +522,170 @@ pub(crate) async fn run_fast_sync_task(
...
+async fn sync_deploys_and_transfers_and_state(
+    effect_builder: EffectBuilder<JoinerEvent>,
+    block: &Block,
+) -> Result<(), LinearChainSyncError> {
+    for deploy_hash in block.deploy_hashes() {
+        fetch_and_store_deploy(effect_builder, *deploy_hash).await?;
+    }
+    for transfer_hash in block.transfer_hashes() {
+        fetch_and_store_deploy(effect_builder, *transfer_hash).await?;
+    }
```
goral09, line 563: This code will make 2500 (+ 100 for Wasm) sequential requests to random peers, correct? This will be *very* slow.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +522,170 @@ pub(crate) async fn run_fast_sync_task(
...
+        } else {
+            walkback_block = *fetch_and_store_block_by_hash(
+                effect_builder,
+                *walkback_block.header().parent_hash(),
+            )
+            .await?;
+        }
+    }
+
+    // Sync forward until we are at the current version.
```
goral09, line 607: Nitpick: not sure if syncing _forward_ from the trusted hash should be called/captured under "archival sync". Maybe splitting it out would make it easier to split the functionality of archival vs non-archival nodes where the latter have to walk forward but not sync the previous states.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +522,170 @@ pub(crate) async fn run_fast_sync_task(
...
-        let missing_descendant_trie_keys =
-            fetch_trie_and_insert_into_trie_store(effect_builder, trie_key).await?;
-        outstanding_trie_keys.extend(missing_descendant_trie_keys);
+
+    if trusted_block_header.timestamp()
+        + era_duration
+            * chainspec
+                .core_config
+                .unbonding_delay
+                .saturating_sub(chainspec.core_config.auction_delay)
```
goral09, line 655: Maybe we can extract it to a method on `chainspec.core_config`? `bonded_eras_num`? @afck

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +522,170 @@ pub(crate) async fn run_fast_sync_task(
...
+    chainspec: Arc<Chainspec>,
+    node_config: NodeConfig,
+) -> Result<BlockHeader, LinearChainSyncError> {
+    // Fetch the trusted header
+    let trusted_block_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
+
+    let era_duration: TimeDiff = std::cmp::max(
+        chainspec.highway_config.min_round_length() * chainspec.core_config.minimum_era_height,
+        chainspec.core_config.era_duration,
     );
```
goral09, line 648: Not necessarily this PR but it's been a long time coming - we should have this as a method/value on `chainspec`.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: My plan for #2083 is to split this functionality out as you suggest and run it in a background task.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -438,13 +476,13 @@ pub(crate) async fn run_fast_sync_task(
                 && is_current_era(
                     &most_recent_block_header,
                     &trusted_key_block_info,
-                    &chainspec,
+                    chainspec,
```
goral09, line 500: Under what circumstances `is_current_era` would return `true` but we would still be rocking a different version from the most recent block's header 🤔 ? Maybe we should log a warning in that case? So that we don't miss them as I think it would mean there's an error somewhere.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +522,170 @@ pub(crate) async fn run_fast_sync_task(
...
+                effect_builder,
+                *walkback_block.header().parent_hash(),
+            )
+            .await?;
+        }
+    }
+
+    // Sync forward until we are at the current version.
+    let mut most_recent_block = trusted_block;
+    while most_recent_block.header().protocol_version() < chainspec.protocol_config.version {
```
goral09, line 609: We walk forward here, just like we do in `fast_sync_to_most_recent`, but the condition to stop looks different:
* there, we return an error on `if most_recent_block_header.protocol_version() > current_version`
* we finish when we couldn't download any more blocks AND we're in the current era AND we're on the same protocol version as the most recent block while here we keep downloading as long as our protocol version is higher than the most recent block's.

Other than that, the two pieces look very similar. We should extract it and reuse. Not only we get rid of code duplication but also we would have consistent logic. Unless I'm missing something.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +522,170 @@ pub(crate) async fn run_fast_sync_task(
...
+    // If we are at an upgrade:
+    // 1. Sync the trie store
+    // 2. Get the trusted era validators from the last switchblock
+    // 3. Try to get the next block by height; if there is `None` then switch to the participating
+    //    reactor.
+    if trusted_block_header.is_switch_block()
+        && trusted_block_header.era_id().successor()
+            == chainspec.protocol_config.activation_point.era_id()
+    {
+        sync_trie_store(effect_builder, *trusted_block_header.state_root_hash()).await?;
```
goral09: Why do we have to download the trie store at the upgrade? Is that b/c Chainspec params are stored there?

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -213,26 +222,43 @@ async fn fetch_and_store_block_header_by_height(
...
                 era_start: block_header.timestamp(),
                 height: block_header.height(),
+                era_id: block_header.era_id() + 1,
             })
     }
 }
 
-/// Get trusted switch block.
-/// Returns an error if we are still in the first era, or the era after an emergency restart.
+/// Get the trusted key block info for a trusted block header.
```
afck:
```suggestion
/// Gets the trusted key block info for a trusted block header.
```

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -280,7 +309,7 @@ async fn get_trusted_key_block_info(
 async fn fetch_and_store_block_by_height(
```
afck, line 323: It would be nice to deduplicate this and `fetch_and_store_block_header_by_height` at some point. (Not necessarily in this PR.)

node/src/components/linear_chain_sync/operations.rs
afck: I don't understand the question.
The idea is that the current era began at `trusted_key_block_info.era_start`, so any blocks in it that we might need to validate and check for replays will have a timestamp greater or equal than that. For replay detection in a block with timestamp `t`, all deploys in blocks with timestamps greater or equal than `t - max_ttl` need to be considered. So we keep downloading old blocks until they are older than `trusted_key_block_info.era_start - max_ttl`.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -455,7 +493,7 @@ pub(crate) async fn run_fast_sync_task(
         .era_start
         .saturating_diff(current_header.timestamp())
         < chainspec.deploy_config.max_ttl
-        && !current_header.is_genesis_child()
+        && current_header.height() != 0
     {
         current_header =
             fetch_and_store_block_by_hash(effect_builder, *current_header.parent_hash())
```
afck: I just realized that this won't even check storage — it will always download the block again, even if it's already there? We should skip the blocks that we already have. (But if we only have the header, we do need to download the body, of course.)

node/src/components/linear_chain_sync/operations.rs
afck: But the main part of the common functionality is already extracted in `fetch_and_store_block_by_height`; otherwise this is a rather small method and the differences you're listing (plus the call to `sync_deploys_and_transfers_and_state`, which `fast_sync_to_most_recent` doesn't do) covers pretty much all of it. I'm not sure whether trying to deduplicate this further would actually make the code simpler?

node/src/components/linear_chain_sync/operations.rs
afck: I think it's so that the participating reactor can actually apply the upgrade: That requires the pre-upgrade state. (Please correct me if I'm wrong, @xcthulhu!)

On the other hand, if the next block _does_ exist, we don't need to do this, do we? So…

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -484,19 +522,170 @@ pub(crate) async fn run_fast_sync_task(
...
+            get_trusted_key_block_info(effect_builder, &*chainspec, &trusted_block_header).await?;
+        if fetch_and_store_block_by_height(
+            effect_builder,
+            trusted_block_header.height() + 1,
+            &trusted_key_block_info,
+            chainspec.highway_config.finality_threshold_fraction,
+        )
+        .await?
+        .is_none()
+        {
```
afck, line 701: …we could

```rust
sync_trie_store(effect_builder, *trusted_block_header.state_root_hash()).await?;
```

here instead of above?

node/src/components/linear_chain_sync/operations.rs
goral09: At the very least, the fact that they both have different stopping conditions is suspicious.

node/src/components/linear_chain_sync/operations.rs
goral09: I think it will - it uses `effect_builder.fetch` internally, which calls to `Fetcher` which in turns does check the storage firstly. The problem here is that `fetch_retry_forever` also calls `put_block_to_storage` which shouldn't be necessary if we found that block in the storage in the first place.

node/src/components/linear_chain_sync/operations.rs
goral09: Right, thanks. I confused myself with the usage of single key block throughout.

goral09 (approved):

node/src/components/linear_chain_sync/operations.rs
afck: That's a good point: Won't we fail to archive all those blocks that are at the current version but weren't fully downloaded by fast sync, @xcthulhu?

node/src/components/linear_chain_sync/operations.rs
xcthulhu: In an emergency restart, a clean-slate node might have to join the network and become a validator.  In this case, the node would need to have the global state before running upgrade.

@fizyk20 is out for a while but this is what we worked out last week.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Should be fixed in afa4a647170b3c161cfa706919d1f5132c8ea7b6

node/src/components/linear_chain_sync/operations.rs
xcthulhu: Operating under the assumption that executing blocks is faster than downloading tries, this loop stops downloading tries after we reach a block with the current version.  After that, we download and execute blocks.

node/src/components/linear_chain_sync/operations.rs
xcthulhu: I've thought about this.  We'd want:

- A `HasFinalitySignatures` trait
- A `Storable` trait and a generic `effect_builder::put::<T :: Storable>` function.

node/src/effect.rs
```diff
@@ -945,6 +945,24 @@ impl<REv> EffectBuilder<REv> {
         .await
     }
 
+    /// Puts a trie into the trie store and asynchronously returns any missing descendant trie keys.
```
afck: > Puts a trie into the trie store

Does it? Wouldn't it fail if `trie_key` isn't stored yet?

xcthulhu: I believe if `trie_key` isn't present it will just output `vec![trie_key]`.

See the [`missing_trie_keys`][1] implementation.

[1]: https://github.com/xcthulhu/casper-node/blob/afa4a647170b3c161cfa706919d1f5132c8ea7b6/execution_engine/src/storage/trie_store/operations/mod.rs#L235

afck: Makes sense. But then this documentation is still wrong, is it? It never puts anything into the trie store?

node/src/components/linear_chain_sync/operations.rs
xcthulhu: You are right we should do this.

xcthulhu: Should be fixed in c75bef9fb

node/src/effect.rs
xcthulhu: You are right, the documentation is wrong.

xcthulhu:
```suggestion
    /// Asynchronously returns any missing descendant trie keys given an ancestor.
```

PR #1998, opened 2021-08-26
Sufficient Signatures

### LINKED ISSUE #1997: Block/BlockHeader-by-height fetchers must check for sufficient finality signatures when checking storage
When a fetcher is fetching a block or block header by height, it must make sure there are sufficient finality signatures when checking storage first.

xcthulhu:
Ensure that when we fetch blocks or block headers by height, we make sure we have a sufficient amount of finality signatures for consensus when checking local storage.

Closes #1997

### DIFF
--- a/node/src/components/fetcher.rs
+++ b/node/src/components/fetcher.rs
@@ -384,7 +384,7 @@ impl ItemFetcher<BlockHeaderWithMetadata> for Fetcher<BlockHeaderWithMetadata> {
         peer: NodeId,
     ) -> Effects<Event<BlockHeaderWithMetadata>> {
         effect_builder
-            .get_block_header_at_height_with_metadata_from_storage(id)
+            .get_block_header_and_sufficient_finality_signatures_by_height_from_storage(id)
             .event(move |result| Event::GetFromStorageResult {
                 id,
                 peer,
--- a/node/src/components/fetcher.rs
+++ b/node/src/components/fetcher.rs
@@ -351,7 +351,7 @@ impl ItemFetcher<BlockWithMetadata> for Fetcher<BlockWithMetadata> {
         peer: NodeId,
     ) -> Effects<Event<BlockWithMetadata>> {
         effect_builder
-            .get_block_at_height_with_metadata_from_storage(id)
+            .get_block_and_sufficient_finality_signatures_by_height_from_storage(id)
             .event(move |result| Event::GetFromStorageResult {
                 id,
                 peer,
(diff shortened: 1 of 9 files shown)


goral09 (changes requested):
This PR changes the behavior of RPC server as well. The `get_block_at_height_with_metadata_from_storage` method (that currently only returns if we have sufficiently finality signatures) is used in the component that answers user queries. We should still return the block, even if we don't have the required finality signatures.

We need use the _sufficient finality signatures_ constraint for intra-node requests only. Otherwise, this PR looks fine.

xcthulhu:
@goral09 Thank you for the speedy review!

You are absolutely right. It would be a regression if that JSON RPC did not call `EffectBuilder::get_block_at_height_with_metadata_from_storage`.  But I took care not to delete it ([here][1] I have left the `EffectBuilder` method intact and [here][2] I have not changed the effect handler in `Storage`).

Confusingly, I *did* delete a related function.  That function is `Storage::get_block_header_and_metadata_by_height` in `storage.rs`.  This function has `_header` in its kind of long name.  It just returns headers, not entire blocks. I changed also changed effects related to it.

As you can see from [here][2], the `Storage` effect handler did not call `get_block_header_and_metadata_by_height`.

[1]: https://github.com/casper-network/casper-node/pull/1998/files#diff-f5624dd7b06a065363101765c50b65e17a21bb6c096b56fe57f0248faa3c13baR1034
[2]: https://github.com/casper-network/casper-node/blob/94a057819560eb179e2ae99b2050efb63cb65ef0/node/src/components/storage.rs#L803

goral09 (approved):
Oh, you're right @xcthulhu . I think I must have misread something or checked out an incorrect branch. Apologies for that. Looks good :+1:

node/src/components/fetcher.rs
```diff
@@ -351,7 +351,7 @@ impl ItemFetcher<BlockWithMetadata> for Fetcher<BlockWithMetadata> {
         peer: NodeId,
     ) -> Effects<Event<BlockWithMetadata>> {
         effect_builder
-            .get_block_at_height_with_metadata_from_storage(id)
+            .get_block_and_sufficient_finality_signatures_by_height_from_storage(id)
```
afck, line 354: Maybe we could introduce the term "fully signed block" or even "signed block" for "block with sufficient finality signatures"
→ `get_signed_block_by_height_from_storage`?

afck (approved):

fizyk20 (approved):

node/src/components/fetcher.rs
xcthulhu: Sure!  But I think we should fix this in another PR...

Issue #1677, opened 2021-07-12
High payment amount results in a stuck deploy


AlexanderLimonov:
PROBLEM:
Deploys with gas payment amounts set to anything above 10,000 CSPR (block gas limit) get stuck forever without failing or succeeding, with no errors or warnings for the user.

TO REPRODUCE:
Launch an NCTL network built from 1.2 or the dev branch, submit a deploy with payment amount set to 10,000,000,000,001 motes.

cspramit:
@AlexanderLimonov  does this kill the network??

AlexanderLimonov:
> @AlexanderLimonov  does this kill the network??

No, everything continues working normally, but the deploy is stuck forever (other deploys will still be processed, it doesn't block anything)

cspramit:
Add to deploy acceptor Epic @cspramit @piotr-dziubecki

PR #2002, opened 2021-08-26
Reject deploys with excessive payment amounts

darthsiroftardis:
CHANGELOG:

- Updated `is_acceptable` to check the payment object's `amount` arg and statically determine if the amount exceeds the block gas limit
- Added three new `DeployValidationFailure` variants to account for the change
- Added a test to assert the check for payment amounts exceeding `block_gas_limits`
- Updated `Deploy::random` function to generate random deploys with standard payment as their payment object

Closes: #1677


node/src/types/deploy.rs
```diff
@@ -1260,4 +1305,51 @@ mod tests {
             "deploy should not have run expensive `is_valid` call"
         );
     }
+
+    #[test]
+    fn not_acceptable_due_to_excessive_payment_amount() {
+        let mut rng = crate::new_rng();
+        let chain_name = "net-1";
+        let deploy_config = DeployConfig::default();
+        let amount = U512::from(deploy_config.block_gas_limit + 1);
```
rafal-ch, line 1344: I like that you check it juuust above the limit 👍

node/src/types/deploy.rs
```diff
@@ -1260,4 +1305,51 @@ mod tests {
             "deploy should not have run expensive `is_valid` call"
         );
     }
+
+    #[test]
+    fn not_acceptable_due_to_excessive_payment_amount() {
```
rafal-ch, line 1340: Please consider adding a test like `transfer_acceptable_regardless_the_excessive_payment_amount()`, just to assure that transfers are excluded from the check. We've hardcoded an exception to standard flow, so it may be worth checking it explicitly.

rafal-ch (approved):
LGTM with 2 small comments.

node/src/types/deploy.rs
```diff
@@ -170,6 +172,19 @@ pub enum DeployValidationFailure {
         got: usize,
     },
 
+    /// Missing payment amount.
+    #[error("missing payment amount")]
+    MissingPaymentAmount,
```
goral09:
```suggestion
    #[error("missing payment argument amount")]
    MissingPaymentAmountArgument,
```

I think it will make it clearer to the users.

node/src/types/deploy.rs
```diff
@@ -755,6 +770,31 @@ impl Deploy {
...
+        if !self.session().is_transfer() {
+            let value = self
+                .payment()
+                .args()
+                .get(ARG_AMOUNT)
+                .ok_or(DeployValidationFailure::MissingPaymentAmount)?;
+            let payment_amount = value
+                .clone()
+                .into_t::<U512>()
+                .map_err(|_| DeployValidationFailure::MissingPaymentAmount)?;
```
goral09: In that case, maybe we should include the expected type of the argument in the error message (or, alternatively, add a new one).

node/src/types/deploy.rs
```diff
@@ -21,6 +21,8 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tracing::{info, warn};
 
+#[cfg(test)]
+use casper_execution_engine::core::engine_state::MAX_PAYMENT;
```
mpapierski: You can put it under

```
#[cfg(test)]
mod tests {
```

below

node/src/types/deploy.rs
mpapierski: ~We technically support `ARG_AMOUNT` as a `u64` type just so you know.~ sorry - I think this comment doesn't apply for custom payments

node/src/types/deploy.rs
darthsiroftardis: Great idea!

node/src/types/deploy.rs
darthsiroftardis: I tried that, but its used in the `random` function which is part of the `impl Deploy` which is why its being imported the way it is

node/src/types/deploy.rs
darthsiroftardis: I decided to add a new error variant altogether

node/src/types/deploy.rs
darthsiroftardis: Applied suggestion

node/src/types/deploy.rs
```diff
@@ -169,6 +171,23 @@ pub enum DeployValidationFailure {
...
+    /// Failed to parse payment amount
+    #[error("failed to parse payment amount as U512")]
+    FailedToParsePaymentAmount,
+
+    /// The payment amount associated with the deploy exceeds the block gas limit.
+    #[error("payment amount of {got} exceeds the block gas limit of {block_gas_limit}")]
+    ExceededBlockGasLimit {
+        /// Configured block gas limit
+        block_gas_limit: u64,
+        /// The payment amount received
```
Fraser999:
```suggestion
    /// Failed to parse payment amount.
    #[error("failed to parse payment amount as U512")]
    FailedToParsePaymentAmount,

    /// The payment amount associated with the deploy exceeds the block gas limit.
    #[error("payment amount of {got} exceeds the block gas limit of {block_gas_limit}")]
    ExceededBlockGasLimit {
        /// Configured block gas limit.
        block_gas_limit: u64,
        /// The payment amount received.
```

node/src/types/deploy.rs
```diff
@@ -815,7 +859,12 @@ impl Deploy {
         ];
         let chain_name = String::from("casper-example");
 
-        let payment = rng.gen();
+        let payment = ExecutableDeployItem::ModuleBytes {
+            module_bytes: casper_types::bytesrepr::Bytes::new(),
```
Fraser999: Let's import `casper_types::bytesrepr::Bytes`

Fraser999 (commented):
Couple of nitpicks, but good to go once @rafal-ch's comment re. the missing test is actioned.

node/src/types/deploy.rs
darthsiroftardis: I moved it up to the imports under `cfg(test)`

node/src/types/deploy.rs
darthsiroftardis: Applied suggestion

node/src/types/deploy.rs
darthsiroftardis: Added test

PR #2009, opened 2021-08-27
Unify Digest and Blake2bHash

### LINKED ISSUE #2008: Unify `Digest` and `Blake2bHash` Types
The `Digest` and `Blake2bHash` types duplicate functionality and should be unified.

dwerner:
Resolves #2008 

This PR unifies the Digest and Blake2bHash.


types/Cargo.toml
```diff
@@ -62,8 +64,17 @@ std = [
...
+    "hashing/std",
+]
+no-std = [
+    "base16/alloc",
+    "hex/alloc",
+    "serde/alloc",
+    "serde_json/alloc",
+    "serde_bytes/alloc",
+    "displaydoc",
+    "hashing/alloc"
```
xcthulhu: @Fraser999 We ended up calling the `no-std` flag for this hashing library `alloc`, since we saw the other libraries are doing that.

Not for this PR, but maybe we follow suite and rename `no-std` -> `alloc` here for the 1.4.0 release?

hashing/src/lib.rs
```diff
@@ -1,174 +1,165 @@
...
+        Digest(result)
+    }
+
+    /// Returns a reference to the underlying value.
+    pub fn inner(&self) -> &[u8; Digest::LENGTH] {
+        &self.0
+    }
+
+    /// Returns the underlying value.
+    pub fn value(&self) -> [u8; Digest::LENGTH] {
```
xcthulhu:
```suggestion
    pub fn value(self) -> [u8; Digest::LENGTH] {
```

xcthulhu: @dwerner when we were investigating this together this appeared to not break anything...

hashing/src/lib.rs
```diff
@@ -1,174 +1,165 @@
...
         self.0
     }
 
-    /// Returns a copy of the wrapped `u8` array as a `Vec`
+    /// Converts the underlying hash digest array to a `Vec`
+    #[cfg(feature = "std")]
     pub fn into_vec(self) -> Vec<u8> {
         self.0.to_vec()
     }
```
xcthulhu: This is only used in 3 places, I say we just inline it...

client/lib/lib.rs
```diff
@@ -1160,9 +1160,8 @@ impl<'a> TryInto<GlobalStateIdentifier> for GlobalStateStrParams<'a> {
     fn try_into(self) -> Result<GlobalStateIdentifier> {
         let hash = Digest::from_hex(self.hash_value).map_err(|error| Error::CryptoError {
```
afck, line 1164: Would it be better to have a separate `HashError` variant now, instead of wrapping it in a `crypto::Error` everywhere? (Not sure; I'm not very familiar with the client code.)

(Also, I wouldn't repeat the word `Error` in the error variant names: just `Error::Crypto`.)

execution_engine/src/core/runtime_context/tests.rs
```diff
@@ -228,6 +228,7 @@ fn account_key_not_writeable() {
 
 #[test]
 fn account_key_readable_valid() {
+    println!("evil shit");
```
afck: :supervillain:

hashing/src/lib.rs
```diff
@@ -1,174 +1,165 @@
-//! Cryptographic hash type and function.
+#![cfg_attr(not(feature = "std"), no_std)]
 
-// TODO - remove once schemars stops causing warning.
-#![allow(clippy::field_reassign_with_default)]
+/// The number of bytes in a Blake2b hash
```
afck: That probably belongs to line 42?
And maybe

```rust
/// The number of bytes in a digest
```

would be more accurate: blake2b allows different sizes; 32 bytes is our choice.

hashing/src/lib.rs
```diff
@@ -1,174 +1,165 @@
...
+    /// Hashes a pair of byte slices into a single [`Digest`]
+    pub fn hash_pair<T: AsRef<[u8]>, U: AsRef<[u8]>>(data1: T, data2: U) -> Digest {
+        let mut result = [0; Digest::LENGTH];
+        let mut hasher = VarBlake2b::new(Digest::LENGTH).unwrap();
+        hasher.update(data1);
+        hasher.update(data2);
+        hasher.finalize_variable(|slice| {
+            result.copy_from_slice(slice);
+        });
+        Digest(result)
```
afck, line 76: ~Is `hash_pair(&[1, 2, 3], &[4]) == hash_pair(&[1], &[2, 3, 4])`? Is that intentional?~

It looks like we're only using this in places where both inputs have a fixed size. (Also, in those places we can't change anymore how we compute the hash.)

I'd still give the method a different name, so nobody uses it assuming that it gives different results for different variable-length pairs.

Maybe `hash_concatenated`?

hashing/src/lib.rs
```diff
@@ -1,174 +1,165 @@
...
-    }
-}
-
-impl From<[u8; Digest::LENGTH]> for Digest {
-    fn from(inner: [u8; Digest::LENGTH]) -> Self {
-        Digest(inner)
+#[cfg(feature = "std")]
+impl std::fmt::LowerHex for Digest {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let hex_string = base16::encode_lower(self.inner());
```
afck, line 118: If we just iterate over `self.inner()` and write each byte we could avoid the string allocation.

hashing/src/lib.rs
```diff
@@ -1,174 +1,165 @@
...
     }
 }
 
-impl Debug for Digest {
-    fn fmt(&self, formatter: &mut Formatter<'_>) -> fmt::Result {
-        write!(formatter, "{}", HexFmt(&self.0))
+#[cfg(feature = "std")]
+impl std::fmt::Display for Digest {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{:10}", base16::encode_lower(&self.0))
```
afck, line 142: …and here we wouldn't even have to encode all the bytes, since we're printing only five of them.

node/src/components/chainspec_loader.rs
```diff
@@ -165,6 +164,7 @@ pub(crate) struct ChainspecLoader {
     /// If `Some`, we're finished loading and committing the chainspec.
     reactor_exit: Option<ReactorExit>,
     /// The initial state root hash for this session.
+    #[data_size(skip)]
```
afck: Why not implement `DataSize` for `Digest`?

node/src/types/deploy.rs
```diff
@@ -69,11 +63,11 @@ static DEPLOY: Lazy<Deploy> = Lazy::new(|| {
         ttl: TimeDiff::from(3_600_000),
         gas_price: 1,
         body_hash,
-        dependencies: vec![DeployHash::new(Digest::from([1u8; Digest::LENGTH]))],
+        dependencies: vec![DeployHash::new(Digest::hash([1u8; Digest::LENGTH]))],
```
afck, line 70: `Digest::from` didn't hash the input, did it?

execution_engine_testing/cargo_casper/tests/integration_tests.rs
```diff
@@ -73,6 +73,8 @@ fn run_tool_and_resulting_tests() {
         .arg(nightly_version.trim())
         .arg("cargo")
         .arg("test")
+        .arg("--")
+        .arg("--nocapture")
```
afck: Not sure how spammy that output is; do we always want that?

node/src/components/small_network/tests.rs
```diff
@@ -350,6 +350,7 @@ async fn run_two_node_network_five_times() {
 /// Sanity check that we can bind to a real network.
 ///
 /// Very unlikely to ever fail on a real machine.
+#[cfg(not(target_os = "macos"))]
```
afck, line 352: Is this something that needs to be ticketed? Otherwise please add a comment why it doesn't work on MacOS.

node/src/components/small_network/tests.rs
```diff
@@ -350,6 +350,7 @@ async fn run_two_node_network_five_times() {
...
 #[tokio::test]
 async fn bind_to_real_network_interface() {
     init_logging();
```
afck, line 355: Please also put `#[cfg(not(target_os = "macos"))]` on `use datalink`.

client/lib/validation.rs
```diff
@@ -110,14 +109,8 @@ pub(crate) fn validate_get_era_info_response(
...
 
-            core::validate_query_proof(
-                &state_root_hash.to_owned().into(),
-                &proofs,
-                &key,
-                path,
-                &proof_value,
-            )
-            .map_err(Into::into)
+            core::validate_query_proof(&state_root_hash, &proofs, &key, path, &proof_value)
```
nilz3ro, line 113: Why did you change the whitespace here?

client/lib/validation.rs
```diff
@@ -173,14 +166,7 @@ pub(crate) fn validate_query_response(
...
 
-    core::validate_query_proof(
-        &state_root_hash.to_owned().into(),
-        &proofs,
-        key,
-        path,
-        proof_value,
-    )
-    .map_err(Into::into)
+    core::validate_query_proof(state_root_hash, &proofs, key, path, proof_value).map_err(Into::into)
```
nilz3ro, line 170: ^

client/lib/validation.rs
```diff
@@ -234,14 +220,8 @@ pub(crate) fn validate_query_global_state(
...
 
-    core::validate_query_proof(
-        &state_root_hash.to_owned().into(),
-        &proofs,
-        key,
-        path,
-        proof_value,
-    )
-    .map_err(Into::into)
+    core::validate_query_proof(&state_root_hash, &proofs, key, path, proof_value)
```
nilz3ro, line 224: ^

client/lib/validation.rs
```diff
@@ -280,13 +260,8 @@ pub(crate) fn validate_get_balance_response(
...
     };
 
-    core::validate_balance_proof(
-        &state_root_hash.to_owned().into(),
-        &balance_proof,
-        *key,
-        &balance,
-    )
-    .map_err(Into::into)
+    core::validate_balance_proof(state_root_hash, &balance_proof, *key, &balance)
```
nilz3ro, line 264: ^

execution_engine/src/core/runtime_context/tests.rs
nilz3ro: We will want to remove any curse words from our public source code

nilz3ro (commented):
I marked a few places where I noticed white space changes. You can ignore these comments if it was intentional, I just wanted to make sure we don't have conflicting editor config / cargo fmt config, otherwise future PRs could be polluted with unnecessary white space changes.

node/src/components/small_network/tests.rs
afck: Or better: just qualify `pnet::datalink` inside this test and remove the import.

node/src/types/deploy.rs
dwerner: No, it didn't. I think the hashed version is more appropriate in the deploy example above.

node/src/components/small_network/tests.rs
dwerner: I believe @xcthulhu was having port binding issues on while developing on osx.

execution_engine_testing/cargo_casper/tests/integration_tests.rs
dwerner: Nope, thanks.

client/lib/validation.rs
dwerner: This was `rustfmt` not an intentional change.

client/lib/validation.rs
dwerner: rustfmt

client/lib/validation.rs
dwerner: rustfmt

types/Cargo.toml
```diff
@@ -30,14 +33,15 @@ once_cell = { version = "1.5.2", default-features = false }
 proptest = { version = "1.0.0", optional = true }
 rand = { version = "0.8.3", default-features = false, features = ["small_rng"] }
 schemars = { version = "0.8.0", features = ["preserve_order"], optional = true }
-serde = { version = "1", default-features = false, features = ["derive"] }
+serde = { version = "1", default-features = false, features = ["derive"], optional=true}
```
mpapierski:
```suggestion
serde = { version = "1", default-features = false, features = ["derive"], optional = true}
```

dwerner:
Closing in favor of https://github.com/casper-network/casper-node/pull/2057

PR #2013, opened 2021-08-30
Reinstate `query-state` command to client binary

darthsiroftardis:
CHANGELOG:

- Added the alias `query-state` to the `query-global-state` command in the client binary to maintain backwards compatibility


rafal-ch (approved):
LGTM 👍

Issue #2022, opened 2021-09-01
Dev branch 1.4.0 - joiner nodes very slow/stalls for 100 val set


cspramit:
Issue:  With a Validator set of 75 , Blocktime =32 sec (check other in link) , joining nodes did not slow down but with all other parameters same  with 100 Validator sync time for joining nodes is very slow.

75 Validator - https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1630054232056700

100 Validator - https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1630308327068700

Other important discussions on the topic

- https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1630407144101700

- https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1630488755147100

- https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1630499150182400

- https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1630512347200900

PR #2024, opened 2021-09-02
Don't request latest Highway protocol state in old eras.

afck:
Fully synchronizing the state of all recent eras is wasteful in practice: Most of it won't be relevant anymore and we should use our bandwidth to synchronize the current era instead.
This changes the state request timer so that it's only initialized in `handle_is_current`, i.e. only in the current era.

Also contains minor improvements to logging.

This should improve #2022.


goral09 (approved):
Good catch.

fizyk20 (approved):

afck:
bors merge

afck:
bors merge

PR #2031, opened 2021-09-03
Better long running event logging

### LINKED ISSUE #2030: Allow for finding long-running events outside of `TRACE` debug levels

marc-casperlabs:
Closes #2030 

Adds a fixed string to the log of long running events.

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -283,6 +283,12 @@ pub(crate) trait ReactorEvent: Send + Debug + From<ControlAnnouncement> + 'stati
     /// [`ControlAnnouncement`](`crate::effect::announcements::ControlAnnouncement`) if the event
     /// is indeed a control announcement variant.
     fn as_control(&self) -> Option<&ControlAnnouncement>;
+
+    /// Returns a cheap but human-readable description of the event.
+    #[inline]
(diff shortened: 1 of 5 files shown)


node/src/reactor.rs
```diff
@@ -283,6 +283,11 @@ pub(crate) trait ReactorEvent: Send + Debug + From<ControlAnnouncement> + 'stati
     /// [`ControlAnnouncement`](`crate::effect::announcements::ControlAnnouncement`) if the event
     /// is indeed a control announcement variant.
     fn as_control(&self) -> Option<&ControlAnnouncement>;
+
+    /// Returns a cheap but human-readable description of the event.
+    fn description(&self) -> &'static str {
+        "anonymous event"
```
goral09, line 289:
```suggestion
        "unknown event type"
```
?

goral09 (approved):
Welcomed improvement.

node/src/reactor.rs
marc-casperlabs: It's not exactly unknown though? :)

marc-casperlabs: I wanted something that does not make the user give up immediately, but rather realize that there might be something that one can do about it!

afck (approved):

PR #2046, opened 2021-09-08
Adjustable consensus traffic controls

marc-casperlabs:
This PR adds configuration settings for the weight applied when throttling. The default configuration values are chosen to mirror the current production behavior.

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.  The format
 * Added new event to the main SSE server stream accessed via `<IP:Port>/events/main` which emits hashes of expired deploys.
 * Added `contract_runtime_execute_block` histogram tracking execution time of a whole block.
 * Long running events now log their event type.
+* Individual weights for traffic throttling can now be set through the configuration.
 
 ### Changed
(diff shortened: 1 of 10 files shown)


node/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.  The format
 * Added new event to the main SSE server stream accessed via `<IP:Port>/events/main` which emits hashes of expired deploys.
 * Added `contract_runtime_execute_block` histogram tracking execution time of a whole block.
 * Long running events now log their event type.
+* Individual weights for traffic throttling can now be set through the configuration.
```
Fraser999: We should be explicit about the new config field name.

Fraser999 (commented):
We need to update two of the three config.toml files in utils/nctl/sh/scenarios/configs too: bond_its.config.toml and itst13.config.toml, but not gov96.config.toml.

node/CHANGELOG.md
marc-casperlabs: Fair enough. Let's hope no one shoots themselves in the foot with this :)

Fraser999 (approved):

Issue #2051, opened 2021-09-10
Add UUID to consensus dependency requests.


goral09:
For better observability of synchronization process, we should add a random number (UUID) to `RequestDependency` messages and log (at `TRACE` level) on creation (sending node) and reception (receiving node).

Issue #2052, opened 2021-09-10
Improve logging in consensus component.


goral09:
We're not consistent w.r.t. to keywords we use for certain data types in the log messages. In some places we use `dep` while others `dependency`, `unit` vs `hash`, we don't log enough information about units to simplify debugging, etc. All of this is useful when we have to read through logs.

PR #2055, opened 2021-09-10
Improve tracking in Consensus component. Unify log messages.

goral09:
For better observability of synchronization process, we're adding a random number (UUID) to `RequestDependency` messages and log (at TRACE level) on creation (sending node) and reception (receiving node).

Also, we're not consistent w.r.t. to keywords we use for certain data types in the log messages. In some places we use dep while others dependency, unit vs hash, we don't log enough information about units to simplify debugging, etc. All of this is useful when we have to read through logs. This is also being addressed in this PR.

Closes https://github.com/casper-network/casper-node/issues/2051 and https://github.com/casper-network/casper-node/issues/2052

### DIFF
--- a/node/src/components/consensus/highway_core/synchronizer.rs
+++ b/node/src/components/consensus/highway_core/synchronizer.rs
@@ -427,7 +427,7 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
                 }
                 // Otherwise request the missing dependency from the sender.
                 let uuid = thread_rng().next_u64();
-                trace!(?uuid, dependency = ?transitive_dependency, %sender, "requesting dependency");
+                debug!(?uuid, dependency = ?transitive_dependency, %sender, "requesting dependency");
                 let ser_msg =
                     HighwayMessage::RequestDependency(uuid, transitive_dependency).serialize();
                 outcomes.push(ProtocolOutcome::CreatedTargetedMessage(ser_msg, sender));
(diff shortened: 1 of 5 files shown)


fizyk20 (approved):
:+1:

goral09:
bors merge

node/src/components/consensus/highway_core/synchronizer.rs
```diff
@@ -409,8 +426,10 @@ impl<I: NodeIdT, C: Context + 'static> Synchronizer<I, C> {
                     continue;
                 }
                 // Otherwise request the missing dependency from the sender.
-                debug!(dependency = ?transitive_dependency, %sender, "requesting dependency");
-                let ser_msg = HighwayMessage::RequestDependency(transitive_dependency).serialize();
+                let uuid = thread_rng().next_u64();
```
Fraser999, line 429: Using `thread_rng()` is undesirable, especially in production code.  Components get passed an RNG via `handle_event()` to avoid using a variety of different RNGs.  It also means in tests we generally get reproducibility.

Same comment applies to a couple of places in highway.rs too.

goral09: I did consider that but didn't see any real advantage of doing it. It's adding quite a lot of complexity to the code (currently, none of the highway/consensus top-level methods are accepting it) and wasn't useful anywhere else (we don't need that kind of determinism in the test code).

PR #2076, opened 2021-09-15
Also set `estimator_weights` on gov96 nightly tests

marc-casperlabs:
No changelog entry, as it was an addendum to the previous g31 traffic controls issue.


TomVasile (approved):
👍

PR #2077, opened 2021-09-15
gzip content negotiation

dwerner:
This PR brings in content-negotiation using warp's `compression` feature. This should be considered an interim fix until something like https://github.com/seanmonstar/warp/pull/513 lands.


node/src/components/rpc_server/http_server.rs
```diff
@@ -90,27 +90,35 @@ pub(super) async fn run<REv: ReactorEventT>(
...
+        .or(rpc_get_dictionary_item)
+        .or(rpc_query_global_state)
+        .or(unknown_method)
+        .or(parse_failure);
+
+    // Supports content negotiation for gzip responses. This is an interim fix until
+    // https://github.com/seanmonstar/warp/pull/513 moves forward.
+    let service_routes_gzip = warp::header::exact("accept-encoding", "gzip")
+        .and(service_routes.clone())
+        .with(warp::compression::gzip());
```
goral09, line 115: Is this adding the `Content-Encoding` header automatically?

dwerner: Yes, `warp`'s `compression` adds this to the response. Warp doesn't yet support automatically inspecting the `accept-encoding` header, so this does the check manually.

node/src/components/rpc_server/http_server.rs
```diff
@@ -90,27 +90,35 @@ pub(super) async fn run<REv: ReactorEventT>(
...
-            .or(rpc_get_peers)
-            .or(rpc_get_status)
-            .or(rpc_get_era_info)
-            .or(rpc_get_auction_info)
-            .or(rpc_get_account_info)
-            .or(rpc_get_rpcs)
-            .or(rpc_get_dictionary_item)
-            .or(unknown_method)
-            .or(parse_failure),
+        service_routes_gzip.or(service_routes)
```
goral09: Did you verify that we can have two set of the exact same endpoints that differ only in that new header? I.e. that we still can handle the requests without the `Accept-Encoding: gzip` header.

dwerner: They are the same endpoints, they all just run this filter to inspect the accept-encoding. I manually tested that this works to support both variations using my new `read_trie` endpoint and tool (with gzip enabled and with it disabled, via the client-provided header).

goral09 (approved):
LGTM but please wait with the final approval for Fraser.

xcthulhu (approved):

Fraser999 (approved):

dwerner:
Test plan for this is to hit any JSON-RPC endpoint with 
`accept-encoding: gzip` header set, as well as not set. I'd recommend using `curl --compressed` when set and just `curl` to verify that both gzip and plaintext responses are created.

PR #2078, opened 2021-09-15
Add `read_trie` rpc and state download tool

dwerner:
Resolves part of #1993 (1/2)

The purpose of this PR is to:
- Add a JSON-RPC endpoint to read a trie as a base16 encoded blob.
- Add a state downloader that will use this endpoint to download the trie, and also any blocks+deploys.

Replaces https://github.com/casper-network/casper-node/pull/1994

### DIFF
--- a/node/src/components/fetcher.rs
+++ b/node/src/components/fetcher.rs
@@ -294,10 +294,10 @@ impl ItemFetcher<GlobalStorageTrie> for Fetcher<GlobalStorageTrie> {
         peer: NodeId,
     ) -> Effects<Event<GlobalStorageTrie>> {
         async move {
-            let maybe_trie = match effect_builder.read_trie(id).await {
+            let maybe_trie = match effect_builder.get_trie(id).await {
                 Ok(maybe_trie) => maybe_trie,
                 Err(error) => {
-                    error!(?error, "read_trie_request");
+                    error!(?error, "get_trie_request");
                     None
                 }
             };
--- a/node/src/components/fetcher.rs
+++ b/node/src/components/fetcher.rs
@@ -293,13 +293,21 @@ impl ItemFetcher<GlobalStorageTrie> for Fetcher<GlobalStorageTrie> {
         id: Digest,
         peer: NodeId,
     ) -> Effects<Event<GlobalStorageTrie>> {
-        effect_builder
-            .read_trie(id)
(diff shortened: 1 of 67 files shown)


resources/test/sse_data_schema.json
```diff
@@ -231,15 +231,31 @@
       ],
       "properties": {
         "hash": {
-          "$ref": "#/definitions/BlockHash"
+          "description": "`BlockHash`",
```
goral09, line 234: Changes in this file seem unrelated to the PR. Did they end up here accidentally?

node/src/components/fetcher.rs
```diff
@@ -293,13 +293,21 @@ impl ItemFetcher<GlobalStorageTrie> for Fetcher<GlobalStorageTrie> {
...
-        effect_builder
-            .read_trie(id)
-            .event(move |maybe_trie| Event::GetFromStorageResult {
+        async move {
+            let maybe_trie = match effect_builder.read_trie(id).await {
+                Ok(maybe_trie) => maybe_trie,
+                Err(error) => {
+                    error!(?error, "read_trie_request");
+                    None
+                }
```
goral09: Why not return the result to the caller and let it decide what to do about the error?

dwerner: The type changed from `Option<_>`, to `Result<Option<_>>` and I think this was done here to minimize changes to the caller.

resources/test/sse_data_schema.json
dwerner: This PR does add doc comments to several types. I made these changes to pass the sse schema test.

dwerner:
Schema tests brought in for the JSON-RPC need updated here.

Fraser999 (commented):
Most of my points are addressed in https://github.com/dwerner/casper-node/pull/13.

However, the behaviour of the `retrieve-state` tool seems a little off in the `download-blocks` mode.  When the folder has even a single block downloaded, the tool doesn't support downloading any blocks at a greater height than that, even if we specify a higher one using the `--download-height` arg.

I didn't check if this behaviour is mirrored in the `download-tries` mode.

The lib could use a few doc comments on the more complex functions to explain what they do.

Fraser999 (approved):

Fraser999:
bors merge

PR #2081, opened 2021-09-15
Expand Deploy Acceptor

### LINKED ISSUE #2085: Expand deploy acceptor checks
A lot of the checks that EE does on deploys can be also performed in the deploy acceptor to reject the deploy earlier.

darthsiroftardis:
Co-authored-by: Ed Hastings <ed@casperlabs.io>

CHANGELOG:

- Expanded the checks the `DeployAcceptor` makes on a single `Deploy`
     - Account related checks
          1. Checks if the `AccountHashes` associated with the `Deploy` are valid  associated keys
          2. Checks if the `Weight` of the `AccountHashes` meet the threshold to perform a deployment
          3. Checks if the balance of the `Account` in the case of a `Deploy` received by from a `Client` is sufficient 
     - ExecutableDeployItem
         1. Checks if the ExecutableDeployItem present in the session or payment fields are valid and present in the case of stored contracts and contract packages
 - Moved the cryptographic validity check of a `Deploy` to be the last step once all other prerequisites are met.
 - Expanded the test suite to cover the newly added checks
         
Closes #2085


goral09:
Hey @darthsiroftardis, could you include in the description of the PR exactly _how_ the checks were expanded? i.e. what was added/changed/removed.

darthsiroftardis:
@goral09 Yes, I am working on polishing the PR before I mark it ready, at that point I will amend the PR description

goral09:
> Checks if the ExecutableDeployItem present in the session or payment fields are valid and present in the case of stored contracts and contract packages

In the spirit of doing only the minimal set of required checks prior to accepting a deploy, I think we shouldn't be validating session code for correctness. Once we know that the deploy will pay for the execution, we should accept any deploy even incorrect.

node/src/types/deploy.rs
```diff
@@ -827,27 +828,380 @@ impl Deploy {
...
+    }
+
+    #[cfg(test)]
+    pub(crate) fn with_missing_contract_version_in_session_package(
+        &mut self,
+        rng: &mut TestRng,
+    ) -> Self {
+        let session = ExecutableDeployItem::StoredVersionedContractByName {
+            name: "Test".to_string(),
+            version: Some(6u32),
```
rafal-ch, line 1205: `version` in `session` is `Some`, yet the function name says `with_missing_contract_version_in_session...`, maybe it's just a naming issue.

node/src/types/deploy.rs
```diff
@@ -827,27 +828,380 @@ impl Deploy {
...
+    }
+
+    #[cfg(test)]
+    pub(crate) fn with_missing_contract_version_in_payment_package(
+        &mut self,
+        rng: &mut TestRng,
+    ) -> Self {
+        let payment = ExecutableDeployItem::StoredVersionedContractByName {
+            name: "Test".to_string(),
+            version: Some(6u32),
```
rafal-ch, line 1056: `version` in `payment` is `Some`, yet the function name says `with_missing_contract_version_in_payment...`, maybe it's just a naming issue.

node/src/components/deploy_acceptor.rs
```diff
@@ -5,41 +5,126 @@ mod tests;
...
+    #[error("Insufficient balance")]
     InsufficientBalance,
+    /// A deploy was sent from account with an unknown balance.
+    #[error("Unable to determine balance")]
+    UnknownBalance,
+    /// Transfer is not valid for payment code.
+    #[error("Transfer is not valid for payment code")]
+    InvalidPaymentVariant,
+    /// Missing payment amount.
+    #[error("missing payment argument amount")]
```
rafal-ch, line 110: Minor nit, but still... Some of the error messages start with captial letter while others don't.

node/src/components/deploy_acceptor.rs
```diff
@@ -101,120 +184,586 @@ impl DeployAcceptor {
...
-            );
-            return effects;
-        }
-
-        let account_key = deploy.header().account().to_account_hash().into();
-
-        // Verify account if deploy received from client and node is configured to do so.
-        if source.from_client() && self.verify_accounts {
+        let acceptable_result =
+            cloned_deploy.is_config_compliant(&self.chain_name, &self.deploy_config);
```
rafal-ch: I think `is_config_compliant()` can be called directly on the `deploy`, because the `self` parameter of `is_config_compliant()` doesn't actually need to be mutable.

Therefore, the possibly expensive `clone` in line #186 might not be needed as well.

node/src/effect.rs
```diff
@@ -1464,29 +1463,90 @@ impl<REv> EffectBuilder<REv> {
...
-                    }
+                    return Some(account);
+                }
+                None
+            }
+            Ok(_) | Err(_) => None,
+        }
+    }
+
+    /// Retrieves the balance of a purse, returns `None` is no purse is present.
```
goral09:
```suggestion
    /// Retrieves the balance of a purse, returns `None` if no purse is present.
```

node/src/types/deploy.rs
```diff
@@ -707,7 +711,7 @@ impl Deploy {
     ///   * the body hash is correct (should be the hash of the body), and
     ///   * approvals are non empty, and
     ///   * all approvals are valid signatures of the deploy hash
-    pub fn is_valid(&mut self) -> Result<(), DeployValidationFailure> {
+    pub fn is_valid(&mut self) -> Result<(), DeployConfigurationFailure> {
```
goral09, line 723: Why did we change the name? I think the `Validation` was more fitting.

node/src/types/deploy.rs
```diff
@@ -827,27 +828,380 @@ impl Deploy {
...
+        Ok(())
     }
 
     #[cfg(test)]
     pub(crate) fn invalidate(&mut self) {
         self.header.chain_name.clear();
     }
 
+    #[cfg(test)]
+    pub(crate) fn valid_deploy(rng: &mut TestRng) -> Self {
```
goral09: This method always creates an instance of native transfer. Maybe rename it to `valid_native_transfer` so that no one tries to use it when a non-transfer valid deploy is expected.

node/src/components/deploy_acceptor.rs
```diff
@@ -5,41 +5,126 @@ mod tests;
...
+        contract_package_hash: ContractPackageHash,
+    },
+    #[error("ContractPackage named {name} does not exist in Account's NamedKeys")]
+    NonexistentContractPackageAtName { name: String },
+    /// Invalid contract at given version.
+    #[error("Invalid contract at version: {contract_version}")]
+    InvalidContractAtVersion { contract_version: ContractVersion },
+    /// Invalid associated keys
+    #[error("Account authorization invalid")]
+    InvalidAssociatedKeys,
```
goral09, line 96: Why no account (hash) in this error variant?

node/src/types/deploy.rs
darthsiroftardis: Naming issue, yes, its meant to be Some version that doesn't exist in a contract package

node/src/components/deploy_acceptor.rs
```diff
@@ -5,41 +5,126 @@ mod tests;
...
+    MissingTransferAmount,
+    /// Missing transfer target argument.
+    #[error("missing transfer target argument")]
+    MissingTransferTargetArgument,
+    /// Missing transfer source argument.
+    #[error("missing transfer source argument")]
+    MissingTransferSourceArgument,
+    /// Module bytes for session code cannot be empty.
+    #[error("module bytes for session code cannot be empty")]
+    MissingModuleBytes,
```
goral09, line 129: From experience working with Kibana - putting the details into the error message is making debugging more difficult as the account hash/prestate/public key/etc. are now part of the generic `"message"` field and not a separate keyword in the log that can be indexed by ElasticSearch. I would suggest _not_ doing that and on the logging side do that manually:
```rust
debug!(?account_hash, ?prestate_hash, "insufficient balance");
```
will produce a log looking like:
```json
{"account_hash":"<account_hash>", "prestate_hash":"<prestate_hash>", "message": "insufficient balance"}
```
vs
```json
{"message": "insufficient balance for <account_hash> at <prestate_hash>"}
```

node/src/components/deploy_acceptor.rs
```diff
@@ -101,120 +184,586 @@ impl DeployAcceptor {
...
+                    })
+            }
+            Some(account) => {
+                let authorization_keys = event_metadata
+                    .deploy()
+                    .approvals()
+                    .iter()
+                    .map(|approval| approval.signer().to_account_hash())
+                    .collect();
+                if !account.can_authorize(&authorization_keys) {
```
goral09, line 296: I think the logic of `can_authorize` is invalid. It requires that _all_ keys that signed the deploy are associated with the account. It should verify that there's enough of them to authorize the deploy execution - i.e. rather than _all_ it should verify if there _exists_ a key in the collection that is associated with the account. Looking at `can_deploy_with` that's exactly what's happening there - meaning we don't have to call `can_authorize` at all.

nilz3ro (approved):

node/src/types/deploy.rs
```diff
@@ -721,14 +725,11 @@ impl Deploy {
     /// Returns true if and only if:
     ///   * the chain_name is correct,
     ///   * the configured parameters are complied with,
-    ///   * the deploy is valid
-    ///
-    /// Note: if everything else checks out, calls the computationally expensive `is_valid` method.
-    pub fn is_acceptable(
+    pub fn is_config_compliant(
```
goral09, line 737: I've noticed that we're not checking if deploy's associated keys are within the chainspec limit. I think we should add it to `is_config_compliant` as it's an easy and cheap check.

goral09 (commented):
All of the `effect_builder.immediately().event(...)` constructs in `deploy_acceptor.rs` can be replaced with calls to the respective methods. It's not going to be bad for performance as it's we were just continuing a synchronous method call.  Performance-wise it will actually be better as we won't be creating additional messages that have to be routed through the reactor.

node/src/components/deploy_acceptor.rs
```diff
@@ -101,120 +184,586 @@ impl DeployAcceptor {
...
+                            prestate_hash,
+                            failure,
+                        },
+                        verification_start_timestamp,
+                    })
+            }
+            None => {
+                let executable_deploy_item_identifier = executable_deploy_item.identifier();
+                let account_hash = event_metadata.deploy().header().account().to_account_hash();
+                match executable_deploy_item_identifier {
```
goral09: ~Isn't `executable_deploy_item_identifier` coming from the payment code? Few lines above we've identified that `Transfer` is invalid variant for the payment while `Module` with empty bytes has to have a payment argument. Why do we pattern match on it here again but not check their content and proceed to verification of the session logic?~

~Or rather - how does it work for the cases below? We're loading contract (packages) for verification but we're using payment part of the deploy.~

**EDIT:** I see what's happening now - we're just doing some more payment code validation for these cases.

node/src/components/deploy_acceptor.rs
```diff
@@ -101,120 +184,586 @@ impl DeployAcceptor {
...
+                                    verification_start_timestamp,
+                                },
+                            )
+                    }
                 }
             }
+        }
+    }
+
+    fn verify_session_logic<REv: ReactorEventT>(
```
goral09, line 507: I don't think we should be verifying the session logic at all. If the deploy can be paid for we shouldn't really care whether deploy's creator made sure it can also be executed.

node/src/components/deploy_acceptor.rs
darthsiroftardis: @goral09 would it be possible to do *both*? That way the error message still contains plenty information for the end user, via the responder. I don't believe it has to be one or the other, please correct me if I am wrong here

node/src/types/deploy.rs
darthsiroftardis: Renamed the tests around it

node/src/types/deploy.rs
darthsiroftardis: Renamed the function to say `non-existent` instead of missing

node/src/components/deploy_acceptor.rs
darthsiroftardis: Thank you, they all start with lower case now

node/src/components/deploy_acceptor.rs
darthsiroftardis: Yep, clone removed

node/src/effect.rs
darthsiroftardis: Applied suggestion

node/src/types/deploy.rs
darthsiroftardis: Renamed as suggested

darthsiroftardis:
> All of the `effect_builder.immediately().event(...)` constructs in `deploy_acceptor.rs` can be replaced with calls to the respective methods. It's not going to be bad for performance as it's we were just continuing a synchronous method call. Performance-wise it will actually be better as we won't be creating additional messages that have to be routed through the reactor.

Applied suggestion and removed most of the `immediately().event` calls

node/src/components/deploy_acceptor.rs
darthsiroftardis: @Fraser999 What is your thought here?

node/src/components/deploy_acceptor.rs
darthsiroftardis: We report this error when the `approvals` on the `Deploy` aren't part of the account's associated keys. It would be more appropriate to have the `BTreeSet<AccountHash>`. However, I don't know if that is something we want to send over the wire given how large that set can be

node/src/types/deploy.rs
darthsiroftardis: We changed it cause a lot of these errors raised when a `Deploy` is checked against a specified `DeployConfig` , which is why it felt more apropos to rename it `ConfigurationFailure`

node/src/components/deploy_acceptor.rs
goral09: I meant an account hash for which that error was raised.

node/src/components/deploy_acceptor.rs
goral09: I guess it will be possible to have both but will require some dilligence on the logging side - developer will have to remember about this and construct a correct log format. 

I think you're right - we shouldn't be changing our code b/c _a_ tool we use for logs works this way.

node/src/components/deploy_acceptor.rs
```diff
@@ -100,121 +183,614 @@ impl DeployAcceptor {
...
+                    };
+                    debug!(?authorization_keys, "insufficient deploy signature weight");
+                    return self.handle_invalid_deploy_result(
+                        effect_builder,
+                        event_metadata,
+                        error,
+                        verification_start_timestamp,
+                    );
+                }
+                return if event_metadata.source().from_client() {
```
goral09: Can we get rid of all `return if`s in this file ? They're confusing to read.

node/src/components/deploy_acceptor.rs
```diff
@@ -224,12 +800,68 @@ impl DeployAcceptor {
...
     }
+
+    fn validate_deploy_cryptography<REv: ReactorEventT>(
+        &self,
+        effect_builder: EffectBuilder<REv>,
+        event_metadata: EventMetadata,
+        verification_start_timestamp: Timestamp,
+    ) -> Effects<Event> {
+        let mut cloned_deploy = event_metadata.deploy().clone();
+        if let Err(deploy_configuration_failure) = cloned_deploy.is_valid() {
```
goral09: I am not sure if everything that happens inside the `is_valid` method is actually more computationally expensive than all of this back-and-forth between `deploy_acceptor`, `storage` and `EE`. It's only the signature verification that _could_ be expensive and I still wouldn't be sure that's true - if there's a single signature to verify (which is probably 99% of cases) it will be faster than creating dozens of events, calling various components etc.

node/src/components/deploy_acceptor.rs
darthsiroftardis: Given that, it can be more than one such account hash or even the entire set, it doesn't make sense to report that information. Its easy enough for clients to validate this themselves off chain.

goral09: Not the account hash of the signing key but the account hash of the deploy that failed due to this error.

node/src/types/deploy.rs
goral09: My understanding is that we're _validating against config_ so it fails during _validation_ phase.

node/src/components/deploy_acceptor.rs
darthsiroftardis: In conversation with @Fraser999 , the functionality seems valid, and it seems outside of the scope of PR anyway

node/src/types/deploy.rs
darthsiroftardis: Added  check

node/src/components/deploy_acceptor.rs
darthsiroftardis: Removed

Fraser999 (approved):

darthsiroftardis:
bors try

goral09 (approved):

PR #2095, opened 2021-09-17
Stop silently ignoring parse errors for complex runtime args in client

### LINKED ISSUE #2094: Casper-client silently ignores errors while parsing complex runtime arg files
If a file is passed via `--session/payment-args-complex` which fails to parse, the client silently swallows the error and creates the deploy with empty runtime args.

Fraser999:
The client currently silently ignores errors which occur when passing a value via `--session-args-complex` or `--payment-args-complex`.  This PR fixes that, causing the client to fail and report the error to the user.  It also adds tests for these two cases.

Closes #2094.


dwerner (approved):

nilz3ro (approved):

Fraser999:
bors merge

PR #2096, opened 2021-09-19
Update feature sets on casper-types and casper-contract crates

### LINKED ISSUE #2088: Redo features on casper-types and casper-contract crates
After discussion with @marc-casperlabs @dwerner and @xcthulhu we agreed to move away from the mutually-exclusive `std` and `no-std` features recently added to these crates.

The intent was to split `casper-types` into smaller crates, one of which would be unconditionally `no_std` and removing the need for `std` or `no-std` features.

Given recent findings on the size and cost of compiled Wasm after splitting hashing and bytesrepr into new crates, and also given that we don't have a resolution for that as yet, I'll revert the changes to the casper-types and casper-contract features to make them more compatible with current (1.3.3) consumers' usage.

They will both be effectively `no_std` under most circumstances, and will contain neither a `std` feature nor a `no-std` feature.

Fraser999:
This PR changes the `casper-types` and `casper-contract` to be `no_std` by default.

The types crate has individual features to enable support for `datasize`, `schemars` and `proptest`.  If any of these are enabled, the std library is included, but it's anticipated that smart contract authors will never need to enable these features.

Using the code in this PR locally I checked that the full set of tests run via `make check-rs` is unchanged.

The PR also adds `codegen-units = 1` and `lto = true` to the release profile in order to reduce the size of the Wasm contracts.  Due to our choice to use a single workspace, these settings are applied to all targets, even though they're only really useful for some of the targets.  However, it didn't significantly affect the overall time to execute `make check-rs` and it did reduce the size of the compiled Wasm from 5.3MB to 3.2MB.

Closes #2088.

### DIFF
--- a/execution_engine/src/core/engine_state/step.rs
+++ b/execution_engine/src/core/engine_state/step.rs
@@ -129,14 +129,26 @@ pub enum StepError {
     TypeMismatch(StoredValueTypeMismatch),
     #[error("Era validators missing: {0}")]
     EraValidatorsMissing(EraId),
-    #[error(transparent)]
-    BytesRepr(#[from] bytesrepr::Error),
-    #[error(transparent)]
-    CLValueError(#[from] CLValueError),
+    #[error("{0}")]
+    BytesRepr(bytesrepr::Error),
+    #[error("{0}")]
+    CLValueError(CLValueError),
     #[error("Other engine state error: {0}")]
     OtherEngineStateError(#[from] Error),
     #[error(transparent)]
     ExecutionError(#[from] execution::Error),
+}
+
+impl From<bytesrepr::Error> for StepError {
+    fn from(error: bytesrepr::Error) -> Self {
+        StepError::BytesRepr(error)
+    }
+}
+
+impl From<CLValueError> for StepError {
+    fn from(error: CLValueError) -> Self {
+        StepError::CLValueError(error)
+    }
 }
 
 #[derive(Debug)]
--- a/execution_engine/src/core/engine_state/upgrade.rs
+++ b/execution_engine/src/core/engine_state/upgrade.rs
@@ -132,10 +132,16 @@ pub enum ProtocolUpgradeError {
     UnableToRetrieveSystemContractPackage(String),
     #[error("Failed to disable previous version of system contract: {0}")]
     FailedToDisablePreviousVersion(String),
-    #[error(transparent)]
-    Bytesrepr(#[from] bytesrepr::Error),
+    #[error("{0}")]
+    Bytesrepr(bytesrepr::Error),
     #[error("Failed to insert system contract registry")]
     FailedToCreateSystemRegistry,
+}
+
+impl From<bytesrepr::Error> for ProtocolUpgradeError {
+    fn from(error: bytesrepr::Error) -> Self {
+        ProtocolUpgradeError::Bytesrepr(error)
+    }
 }
 
 pub(crate) struct SystemUpgrader<S>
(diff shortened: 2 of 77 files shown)


dwerner (approved):
Looks good to me.

execution_engine/src/core/engine_state/step.rs
```diff
@@ -129,16 +129,28 @@ pub enum StepError {
     TypeMismatch(StoredValueTypeMismatch),
     #[error("Era validators missing: {0}")]
     EraValidatorsMissing(EraId),
-    #[error(transparent)]
```
marc-casperlabs, line 132: Hmm, why change this? The type still derives `thiserror` (see `#[derive(Error)]` above, so `#[from]` and `#[transparent]` still work. I believe they also set source correctly.

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -132,12 +132,18 @@ pub enum ProtocolUpgradeError {
     UnableToRetrieveSystemContractPackage(String),
     #[error("Failed to disable previous version of system contract: {0}")]
     FailedToDisablePreviousVersion(String),
-    #[error(transparent)]
-    Bytesrepr(#[from] bytesrepr::Error),
+    #[error("{0}")]
```
marc-casperlabs, line 135: Same _spiel_ as above? Still derives `thiserror`?

execution_engine/src/shared/transform.rs
```diff
@@ -23,10 +23,16 @@ use casper_types::{
 /// cause an overflow).
 #[derive(PartialEq, Eq, Debug, Clone, thiserror::Error)]
 pub enum Error {
-    #[error(transparent)]
+    #[error("{0}")]
```
marc-casperlabs, line 26: (etc. see also below, I guess :))

node/src/components/storage/lmdb_ext.rs
```diff
@@ -40,6 +40,10 @@ pub enum LmdbExtError {
     Other(Box<dyn std::error::Error + Send + Sync>),
 }
 
+#[derive(Debug, Error)]
+#[error("{0}")]
+pub struct BytesreprError(pub bytesrepr::Error);
```
marc-casperlabs, line 45: Is there an option to have a "std" feature on `cargo-types` that implements `Error` on `bytesrepr::Error`, instead of adding this extra wrapper here?

types/src/cl_type.rs
```diff
@@ -79,25 +82,32 @@ pub enum CLType {
     /// [`PublicKey`](crate::PublicKey) system type.
     PublicKey,
     /// `Option` of a `CLType`.
+    #[cfg_attr(feature = "datasize", data_size(skip))]
```
marc-casperlabs, line 85: This is probably minor and/or wrong, I'd consider a manual `impl DataSize for CLType` here for clarity, since we're always skipping anyway.

marc-casperlabs (approved):
Some somewhat bikesheddy comments from my end, I'll leave it up to you whether or not these are worth adressing. Otherwise LGTM. :)

execution_engine/src/core/engine_state/step.rs
Fraser999: It requires the wrapped error to implement `std::error::Error`.  Since this isn't available in a `no_std` environment, the errors defined in `casper-types` can't be used in this way.

node/src/components/storage/lmdb_ext.rs
Fraser999: I very strongly don't want to provide that functionality.  `casper-types` is meant to be consumed by smart-contracts, and I don't want to make it easy for those to accidentally include the std lib again.

Fraser999:
bors merge

types/src/cl_type.rs
Fraser999: As discussed offline, we can address this later if required.

Fraser999:
bors merge

Fraser999:
bors merge

PR #2098, opened 2021-09-20
Deduplicate requested deploys in fast-sync branch

marc-casperlabs:
The core of this PR is the re-introduction of memory deduplication of deploys (and now other types) being sent to other nodes by adding an object pool to the storage component. It contains a few other changes that were necessary refactorings made on the way, mainly on the path of networking messages being handled, to make this possible:

* New terminology is introduced in the form of "Incoming" events. "Incoming" refers to announcements that indicate the arrival of a new network message. Many components have had their internal event variant renamed, e.g. `consensus::Event::MessageReceived` is now `consensus::Event::Incoming`.
* Incoming network messages are standardized on the `MessageIncoming<I, M>` data type, which combines an incoming message with a sender. Various incoming events are just type aliases, e.g. `type ConsensusMessageIncoming<I> = MessageIncoming<I, consensus::ConsensusMessage>`. This replaces the copied old code where every event on different components would have the same or almost same fields of `sender` and `message`.
* The storage component now handles more than just storage requests and thus received its own associated event (it was a type alias on the fast sync branch before).
* Some of the logic hidden in the routing of various reactors relating to the handling of `GetRequest`s has been deduplicated and refactored.
* The handling of incoming network messages has also partially been cleaned up, instead of having a big, often copied `match` block in every reactor, the protocol message type itself is responsible for splitting itself into more fine-grained `Incoming` announcements, which can then be handled like any other announcements.
* Components can now advertise that they handle a specific kind of networking message by support `From<*Incoming>` conversions on their event type.
* The gossiper now has a proper effect builder method and a related event for the process of beginning the gossip of a new item (`EffectBuilder::begin_gossip` and `gossiper::Event::BeginGossipRequest`); these were built ad-hoc before.
* The `NetworkingAnnouncement` has been removed, as its functionality has been completely taken over by `*Incoming` and `begin_gossip`. The `NewPeer` variant was already unused.
* `GetRequest`s are the same on the networking layer, but instead of replicating `enum`s manually via explicit tags, requests are turned into proper enums, `NetRequest` and `TrieRequest` higher up. For now, the split between `Net`- and `Trie` requests is made pragmatically because the former are handled by the storage component, while the latter is handled by the contract runtime.
* Errors in storage have been split into fatal errors and potentially non-fatal ones. Before, all storage errors were fatal, but since the storage component now answers `GetRequest`s, non-fatal errors such as a malformed request, need to be handled differently.
* An object pool has been added (sometimes called a "cache", which is a slight misnomer) and is used by the storage component to deduplicate requested items, not just deploys. Any item retrieved from the storage component for a `GetRequest` will be deduplicated.

The related work of overhauling response message in the same way has not been included in this PR to keep the changes focused around the memory pool effort.


Fraser999 (approved):

node/src/components/fetcher/tests.rs
```diff
@@ -107,91 +111,80 @@ reactor!(Reactor {
     announcements: {
         // The deploy fetcher needs to be notified about new deploys.
         DeployAcceptorAnnouncement<NodeId> -> [deploy_fetcher];
-        NetworkAnnouncement<NodeId, Message> -> [fn handle_message];
         // Currently the RpcServerAnnouncement is misnamed - it solely tells of new deploys arriving
         // from a client.
         RpcServerAnnouncement -> [deploy_acceptor];
         ChainspecLoaderAnnouncement -> [!];
+
+        // The `handle_net_request` function implements the entire "custom" logic found in this test
```
afck:
```suggestion
        // The `handle_net_response` function implements the entire "custom" logic found in this test
```

node/src/components/small_network.rs
```diff
@@ -708,21 +697,16 @@ where
...
     /// Emits an announcement that a connection has been completed.
-    fn connection_completed(
-        &self,
-        effect_builder: EffectBuilder<REv>,
-        peer_id: NodeId,
-    ) -> Effects<Event<P>> {
+    fn connection_completed(&self, peer_id: NodeId) {
         trace!(num_peers = self.peers().len(), new_peer=%peer_id, "connection complete");
         self.net_metrics.peers.set(self.peers().len() as i64);
-        effect_builder.announce_new_peer(peer_id).ignore()
```
afck, line 725: Was that announcement unused?

node_macros/src/rust_type.rs
```diff
@@ -39,27 +49,45 @@ impl RustType {
             .clone()
     }
 
-    /// Returns the type without the path, but with type arguments, e.g. `SmallNet<NodeId>`.
-    pub fn ty(&self) -> TokenStream {
-        let ident = self.ident();
-        let args = &self
+    /// Returns the types identifier with type arguments attached as strings, i.e.
+    /// `GossipIncomingDeploy`'.
```
afck: Not sure if I understand correctly:

```suggestion
    /// Returns the type's identifier with type arguments concatenated into a single camel case
    /// identifier, e.g. `GossipIncomingDeploy`.
```

node_macros/src/rust_type.rs
```diff
@@ -39,27 +49,45 @@ impl RustType {
...
-            .arguments;
+            .ident
+            .clone();
+
+        // This is a slightly ugly hack to get a string out of the type.
+        let final_arg = self.final_type_arg_ident();
+        let args: String = quote!(#final_arg)
+            .to_string()
+            .chars()
+            .filter(|c| c.is_alphabetic())
```
afck, line 68: Should this say "alphanumeric", since type identifiers can contain numbers?

node/src/testing.rs
```diff
@@ -303,19 +307,23 @@ impl<REv: 'static> Default for ComponentHarness<REv> {
 
 /// A special event for unit tests.
 ///
-/// Essentially discards all event (they are not even processed by the unit testing hardness),
+/// Essentially discards most events (they are not even processed by the unit testing hardness),
```
afck:
```suggestion
/// Essentially discards most events (they are not even processed by the unit testing harness),
```

node/src/effect/incoming.rs
```diff
@@ -0,0 +1,186 @@
...
+/// A new message from a gossiper arrived.
+pub type GossiperIncoming<T> = MessageIncoming<NodeId, gossiper::Message<T>>;
+
+/// A new message requesting various objects arrived.
+pub(crate) type NetRequestIncoming = MessageIncoming<NodeId, NetRequest>;
+
+/// A new message requesting a trie arrived.
+pub(crate) type TrieRequestIncoming = MessageIncoming<NodeId, TrieRequest>;
+
+/// A new message requesting various objects arrived.
```
afck:
```suggestion
/// A new message responding to a request arrived.
```

node/src/effect/incoming.rs
```diff
@@ -0,0 +1,186 @@
...
+pub(crate) type NetResponseIncoming = MessageIncoming<NodeId, NetResponse>;
+
+/// A new message requesting a trie arrived.
```
afck:
```suggestion
/// A new message responding to a trie request arrived.
```

node/src/effect/incoming.rs
```diff
@@ -0,0 +1,186 @@
...
+pub(crate) enum NetRequest {
+    /// Request for a deploy.
+    Deploy(Vec<u8>),
+    /// Request for a block.
+    Block(Vec<u8>),
+    /// Request for a gossiped public listening address.
+    // TODO: Move this out of `NetRequest` into its own type, it is never valid.
+    GossipedAddress(Vec<u8>),
+    /// Request for a block by its height in the linear chain.
+    BlockAndMetadataByHeight(Vec<u8>),
```
afck, line 69: The "metadata" is just the finality signatures, I think? Maybe we should also call this `BlockAndFinalitySignaturesByHeight` then (also in the `Display` impl and `NetResponse`).

node/src/components/storage/tests.rs
```diff
@@ -257,10 +266,13 @@ fn put_block_signatures(
...
     let response = harness.send_request(storage, move |responder| {
-        StorageRequest::PutBlockSignatures {
-            signatures,
-            responder,
+        {
+            StorageRequest::PutBlockSignatures {
+                signatures,
+                responder,
+            }
         }
```
```suggestion
        StorageRequest::PutBlockSignatures {
            signatures,
            responder,
        }
```

node/src/components/storage/tests.rs
```diff
@@ -288,11 +299,14 @@ fn put_execution_results(
...
-            block_hash: Box::new(block_hash),
-            execution_results,
-            responder,
+        {
+            StorageRequest::PutExecutionResults {
+                block_hash: Box::new(block_hash),
+                execution_results,
+                responder,
+            }
         }
```
```suggestion
        StorageRequest::PutExecutionResults {
            block_hash: Box::new(block_hash),
            execution_results,
            responder,
        }
```

node/src/reactor/participating.rs
```diff
@@ -444,44 +496,12 @@ impl Reactor {
...
-                            block_height,
-                        )
-                })
-            }
             Tag::Trie => {
                 Self::respond_to_fetch(effect_builder, serialized_id, sender, |trie_key| {
                     self.contract_runtime.get_trie(trie_key)
                 })
             }
+            _ => todo!("already handled"),
```
afck: Should this method be removed then? Or called `handle_trie_request` and not have a `tag` argument?

node/src/reactor/joiner.rs
```diff
@@ -146,6 +152,10 @@ pub(crate) enum JoinerEvent {
     AddressGossiper(gossiper::Event<GossipedAddress>),
 
     /// Requests.
```
afck:
```suggestion
    // Requests.


```

node/src/reactor/joiner.rs
```diff
@@ -926,6 +867,153 @@ impl reactor::Reactor for Reactor {
...
+                        ) {
+                            Some(fetcher_event) => self.dispatch_event(
+                                effect_builder,
+                                rng,
+                                JoinerEvent::BlockByHeightFetcher(fetcher_event),
+                            ),
+                            None => {
+                                info!(
+                                    "{} sent us a block with metadata we couldn't parse!
+                Banning",
```
afck:
```suggestion
                                    "{} sent us a block with metadata we couldn't parse! Banning",
```

node/src/reactor/joiner.rs
```diff
@@ -926,6 +867,153 @@ impl reactor::Reactor for Reactor {
...
+                        serialized_item,
+                        ) {
+                            Some(fetcher_event) => {
+                                self.dispatch_event(effect_builder, rng,
+                JoinerEvent::BlockHeaderByHeightFetcher(fetcher_event))             } ,
+                            None => {
+                                  info!("{} sent us a block header with finality signatures we
+                couldn't parse! Banning", sender);
+                effect_builder.announce_disconnect_from_peer(sender).ignore()
+                }         }
```
afck: That could use some manual formatting… but I can understand why rustfmt didn't feel like doing it. :sweat_smile:

afck (commented):
(My comments so far. Haven't reviewed all of it yet.)

node/src/reactor/participating.rs
goral09: This will cause panic if hit.

node/src/components/storage.rs
```diff
@@ -121,7 +133,7 @@ const OS_FLAGS: EnvironmentFlags = EnvironmentFlags::WRITE_MAP;
 /// Mac OS X exhibits performance regressions when `WRITE_MAP` is used.
 #[cfg(target_os = "macos")]
 const OS_FLAGS: EnvironmentFlags = EnvironmentFlags::empty();
-const _STORAGE_EVENT_SIZE: usize = mem::size_of::<StorageRequest>();
+const _STORAGE_EVENT_SIZE: usize = mem::size_of::<Event>();
```
afck, line 136: Is this unused? Should the leading underscore be removed?

node/src/components/storage.rs
```diff
@@ -1487,6 +1689,48 @@ impl Storage {
...
+    {
+        let fetched_or_not_found = FetchedOrNotFound::from_opt(id, opt_item);
+        let serialized = fetched_or_not_found
+            .to_serialized()
+            .map_err(FatalStorageError::StoredItemSerializationFailure)?;
+        let shared: Arc<[u8]> = serialized.into();
+
+        if self.enable_mem_deduplication && fetched_or_not_found.was_found() {
+            self.serialized_item_pool
+                .put(serialized_id.into(), Arc::downgrade(&shared));
```
afck, line 1776: I still wonder whether the key should include the tag.
I don't think it's possible right now, but if there were two items with the same hash but different tags, that could return the wrong thing, could it?

node/src/components/storage.rs
marc-casperlabs: No, it is used in the assert below. But that's out of scope for this PR, IMO.

node/src/components/storage.rs
marc-casperlabs: Yes, but that'd be a hash collision, I don't think we are coding to defend against these.

The other case is that roughly the same item is valid as both, e.g. in the case of both being 0 or empty or something like that. For this to be an issue, they would have to hash to the same hash, but serialize to different representations. If that's the case, we may have a bigger issue on our hands?

node/src/reactor/joiner.rs
marc-casperlabs: Sometimes it just gives up, it seems. :man_shrugging:

node/src/components/small_network.rs
marc-casperlabs: Yes, as far as I now it was only used in the linear chain sync.

marc-casperlabs: `NetworkAnnouncement::NewPeer` was handled by just returning `Effects::new` in both participating and joining reactor.

node_macros/src/rust_type.rs
marc-casperlabs: I hope we're not using the "alphanumericability", I am fine with this error'ing. If it was the case though, we'd have to check the first letter separately, since idents cannot _start_ with numbers.

For simplicity, I just disallowed numbers here.

node/src/effect/incoming.rs
marc-casperlabs: Probably a good idea, but out of scope for this PR in my opinion. I'd rather not change these here.

node/src/reactor/participating.rs
marc-casperlabs: True. It should in theory be `unreachable!`, but I will replace it with a `warn!` instead, just in case.

node/src/reactor/joiner.rs
marc-casperlabs: Did the best I could...

node/src/components/storage.rs
afck: Actually, not all items are content-addressed. E.g. there's a block-with-signatures-by-height item type and a block-header-with-signatures-by-height.
Won't we store the serialized block _header_ under the same key as the serialized full block?

marc-casperlabs: Hmm, you are right, I will double check.

marc-casperlabs: Yes, sadly IDs are domain specific. I've added the tag to disambiguate.

afck (approved):

PR #2032, opened 2021-09-03
Retrieve state tools

dwerner:
Resolves part of #1993 (2/2)

The purpose of this PR is to:

- Add a `retrieve-state` that downloads the trie and blocks from the JSON-RPC to a local LMDB and JSON files respectively.
- Add a `dry-run-deploys` that runs block files generated by the above tool against an isolated execution engine.

PR #2102, opened 2021-09-20
Fix behavior of retrieve state tool

### LINKED ISSUE #2097: Fix behaviour of Retrieve State tool
We need to ensure that the tool downloads blocks higher than the currently stored ones if requested to do so.  This also applies to tries, although it's unconfirmed whether this is actually a problem for tries.

The tool also needs to be enhanced in a few ways:
- [x] ensure that the tool downloads blocks higher than the currently stored ones if requested to do so
- [x] address the todo comment re. making hard-coded values runtime args
- [x] address the review comment regarding enabling gzip on the client
- [x] add doc comments around the more complex functions to give a high-level explanation of their purpose
- [x] add help strings to the tool to explain the args and usage

dwerner:
Resolves #2097

New behavior is: 

Parse a BlockIdentifier from the command line argument for height.
Check on disk for each block file, download only when appropriate.

### DIFF
--- a/utils/retrieve-state/src/lib.rs
+++ b/utils/retrieve-state/src/lib.rs
@@ -176,7 +176,7 @@ pub async fn download_block_with_deploys(
     })
 }
 
-/// Download block files from the highest (provided) block hash to genesis.
+/// Download block files from the block hash or height to genesis.
 pub async fn download_blocks(
     client: &mut Client,
     url: &str,
--- a/utils/retrieve-state/src/lib.rs
+++ b/utils/retrieve-state/src/lib.rs
@@ -35,12 +35,12 @@ use casper_node::{
 };
 use casper_types::{bytesrepr::FromBytes, Key, StoredValue};
 
-// TODO: make these parameters
-const RPC_SERVER: &str = "http://localhost:11101/rpc";
+use crate::offline::{get_block_file, read_block_file};
+
 pub const LMDB_PATH: &str = "lmdb-data";
 pub const CHAIN_DOWNLOAD_PATH: &str = "chain-download";
-pub const DEFAULT_TEST_MAX_DB_SIZE: usize = 483_183_820_800; // 450 gb
-pub const DEFAULT_TEST_MAX_READERS: u32 = 512;
+pub const DEFAULT_MAX_DB_SIZE: usize = 483_183_820_800; // 450 gb
+pub const DEFAULT_MAX_READERS: u32 = 512;
 
 async fn rpc<'de, R, P>(
     client: &mut Client,
@@ -64,6 +64,7 @@ where
     Ok(deserialized)
 }
 
+/// Get a block with optional parameters.
 pub async fn get_block(
     client: &mut Client,
     url: &str,
@@ -72,6 +73,7 @@ pub async fn get_block(
     rpc(client, url, "chain_get_block", params).await
 }
 
+/// Get the genesis block.
 pub async fn get_genesis_block(
     client: &mut Client,
     url: &str,
@@ -103,6 +105,7 @@ async fn get_deploy(
     rpc(client, url, "info_get_deploy", params).await
 }
 
+/// Composes a [`JsonBlock`] along with all of it's transfers and deploys.
 #[derive(Debug, Deserialize, Serialize)]
 pub struct BlockWithDeploys {
     pub block: JsonBlock,
@@ -111,6 +114,7 @@ pub struct BlockWithDeploys {
 }
 
 impl BlockWithDeploys {
+    /// Save [`BlockWithDeploys`] to disk, in the provided file path.
     pub async fn save(&self, path: impl AsRef<Path>) -> Result<(), anyhow::Error> {
         let path = PathBuf::from(path.as_ref());
         let file_path = path.join(format!(
@@ -125,6 +129,7 @@ impl BlockWithDeploys {
     }
 }
 
+/// Download a block, along with all it's deploys.
 pub async fn download_block_with_deploys(
     client: &mut Client,
     url: &str,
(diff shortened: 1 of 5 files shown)


goral09:
How does this work if https://github.com/casper-network/casper-node/pull/2032 hasn't been merged yet 🤔

dwerner:
> How does this work if #2032 hasn't been merged yet thinking

That was split up, so the `retrieve-state` tool itself merged with #2078 and `dry-run-deploys` will follow.

goral09:
> > How does this work if #2032 hasn't been merged yet thinking
> 
> That was split up, so the `retrieve-state` tool itself merged with #2078 and `dry-run-deploys` will follow.

Can you update the description and title of that PR then? It still mentiones the "retrieve state tool".

utils/retrieve-state/src/lib.rs
```diff
@@ -171,38 +176,94 @@ pub async fn download_block_with_deploys(
     })
 }
 
+/// Download block files from the highest (provided) block hash to genesis.
```
xcthulhu: But `BlockIdentifier` isn't a hash...?

utils/retrieve-state/src/lib.rs
```diff
@@ -244,57 +305,69 @@ pub async fn download_trie(
...
+        chain_download_path: impl AsRef<Path>,
+    ) -> Result<Option<u64>, anyhow::Error> {
+        get_block_downloaded(chain_download_path, |a, b| a.max(b))
+    }
+
+    fn get_block_downloaded(
+        chain_download_path: impl AsRef<Path>,
+        comparator: impl Fn(u64, u64) -> u64,
     ) -> Result<Option<u64>, anyhow::Error> {
         let lowest = if chain_download_path.as_ref().exists() {
```
goral09: Not necessarily the lowest though, right? It depends on the comparator now.

utils/retrieve-state/src/lib.rs
dwerner: BlockIdentifier can be a hash or a height. If it's a height we need to download the block and get the hash from there.

utils/retrieve-state/src/lib.rs
dwerner: Thanks, renamed it.

utils/retrieve-state/src/lib.rs
```diff
@@ -244,57 +305,69 @@ pub async fn download_trie(
...
+    }
+
+    /// Returns the height of the highest block already downloaded.
+    pub fn get_highest_block_downloaded(
+        chain_download_path: impl AsRef<Path>,
+    ) -> Result<Option<u64>, anyhow::Error> {
+        get_block_downloaded(chain_download_path, |a, b| a.max(b))
+    }
+
+    fn get_block_downloaded(
```
goral09, line 345: If the module is already called `offline` then we don't really need the _downloaded_ part. `offline::get_block` should suffice, or `local::` or `storage::`.

utils/retrieve-state/src/lib.rs
```diff
@@ -314,8 +387,37 @@ pub mod offline {
         block_files
     }
 
+    /// Look for a block file matching the given [`BlockIdentifier`], returns `None` if not found.
+    pub fn get_block_file(
+        chain_path: impl AsRef<Path>,
```
goral09, line 391: In the other method this argument is called `chain_download_path`. ALternatively, can you wrap this with a struct and expose these as methods on the `LocalStorage`?

utils/retrieve-state/src/lib.rs
```diff
@@ -314,8 +387,37 @@ pub mod offline {
...
+        walkdir::WalkDir::new(chain_path)
+            .into_iter()
+            .filter_map(|entry| entry.ok())
+            .find(|entry| {
+                let file_name = match entry.file_name().to_str() {
+                    Some(file_name) => file_name,
+                    None => return false,
+                };
+                let split = file_name.split('-').collect::<Vec<&str>>();
+                if let ["block", height, hash] = &split[..] {
```
goral09, line 403: Detection of the block's height and hash is duplicated across a couple methods. It would be a more declarative approach if there was a method that took a file name and returned an object wrapping the file but that exposes the underlying block's hash and height. It could be reused in all of the places where we manually split the file name and pattern match on the parts.

utils/retrieve-state/src/lib.rs
xcthulhu:
```suggestion
/// Download block files from the block hash or height to genesis.
```

Maybe say this?

utils/retrieve-state/src/lib.rs
```diff
@@ -171,38 +176,94 @@ pub async fn download_block_with_deploys(
...
+            } else {
+                Ok(None)
+            }
         }
     }
-    println!("finished downloading blocks");
-    Ok(offline::get_block_files(chain_download_path))
 }
 
+async fn get_block_and_download(
```
xcthulhu, line 239: As discussed, this kind of kills my current system where I use an SD Card for storing these files.

Should we ticket using LMDB for this?

xcthulhu: Ticketed here: https://github.com/casper-network/casper-node/issues/2121

utils/retrieve-state/src/lib.rs
dwerner: I don't disagree with this perspective or with other comments in the vein of factoring this stuff out, but I think I'd like to do that work as a part of #2121.

utils/retrieve-state/src/lib.rs
dwerner: I want to defer this to #2121.

utils/retrieve-state/src/lib.rs
dwerner: Let's do this in #2121

utils/retrieve-state/src/lib.rs
dwerner: #2121 as well

xcthulhu (approved):
LGTM

dwerner:
bors merge

dwerner:
bors merge

dwerner:
bors try

dwerner:
Closing in favor of #2182

PR #2110, opened 2021-09-22
GH-1765: Add docs to the shared module.

### LINKED ISSUE #1765: Docs update: execution_engine/shared

mpapierski:
Closes #1765

This PR adds rust docs to the shared module in EE and its children.

### DIFF
--- a/execution_engine/CHANGELOG.md
+++ b/execution_engine/CHANGELOG.md
@@ -26,9 +26,8 @@ All notable changes to this project will be documented in this file.  The format
 * Support building and testing using stable Rust.
 * Increase price of `create_purse` to 2.5CSPR.
 * Improve doc comments to clarify behavior of the bidding functionality.
-* Documented `core` module and children.
-* Changed parameters to `LmdbEnvironment`'s constructor enabling manual flushing to disk.
-* Documented `shared` module and children.
+* Document `core` and `shared` modules and their children.
+* Change parameters to `LmdbEnvironment`'s constructor enabling manual flushing to disk.
 
 ### Fixed
 * Fix a case where user could potentially supply a refund purse as a payment purse.
--- a/execution_engine/CHANGELOG.md
+++ b/execution_engine/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file.  The format
 * Improve doc comments to clarify behavior of the bidding functionality.
 * Documented `core` module and children.
 * Changed parameters to `LmdbEnvironment`'s constructor enabling manual flushing to disk.
+* Documented `shared` module and children.
 
 ### Fixed
 * Fix a case where user could potentially supply a refund purse as a payment purse.
--- a/execution_engine/src/shared/newtypes/mod.rs
+++ b/execution_engine/src/shared/newtypes/mod.rs
@@ -5,13 +5,13 @@ use std::fmt::{self, Display, Formatter};
 use serde::Serialize;
 use uuid::Uuid;
 
-/// A correlation id is a unique representation of an event occuring
-/// in the system as it passes through different components.
+/// A correlation id is a unique identifier which can be used to track the progress of a given
+/// execution engine operation.
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Serialize)]
 pub struct CorrelationId(Uuid);
 
 impl CorrelationId {
-    /// Creates new unique `CorrelationId`
+    /// Creates new unique `CorrelationId`.
     pub fn new() -> CorrelationId {
         CorrelationId(Uuid::new_v4())
     }
--- a/execution_engine/src/shared/newtypes/mod.rs
+++ b/execution_engine/src/shared/newtypes/mod.rs
@@ -5,7 +5,8 @@ use std::fmt::{self, Display, Formatter};
 use serde::Serialize;
 use uuid::Uuid;
 
-/// A correlation id that is an unique represenatation of an event occuring     in the system passes through different components.
+/// A correlation id is a unique representation of an event occuring
+/// in the system as it passes through different components.
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Serialize)]
 pub struct CorrelationId(Uuid);
 
@@ -15,7 +16,7 @@ impl CorrelationId {
         CorrelationId(Uuid::new_v4())
     }
 
-    /// Returns true if given unique identifier is empty.
+    /// Returns true if the given unique identifier is empty.
     pub fn is_empty(&self) -> bool {
         self.0.is_nil()
     }
--- a/execution_engine/src/shared/newtypes/mod.rs
+++ b/execution_engine/src/shared/newtypes/mod.rs
@@ -5,14 +5,17 @@ use std::fmt::{self, Display, Formatter};
 use serde::Serialize;
 use uuid::Uuid;
 
+/// A correlation id that is an unique represenatation of an event occuring     in the system passes through different components.
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Serialize)]
 pub struct CorrelationId(Uuid);
 
 impl CorrelationId {
+    /// Creates new unique `CorrelationId`
     pub fn new() -> CorrelationId {
         CorrelationId(Uuid::new_v4())
     }
 
+    /// Returns true if given unique identifier is empty.
     pub fn is_empty(&self) -> bool {
         self.0.is_nil()
     }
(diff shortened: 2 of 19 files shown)


execution_engine/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file.  The format
 * Improve doc comments to clarify behavior of the bidding functionality.
 * Documented `core` module and children.
 * Changed parameters to `LmdbEnvironment`'s constructor enabling manual flushing to disk.
+* Documented `shared` module and children.
```
```suggestion
* Document `core` and `shared` modules and their children.
* Change parameters to `LmdbEnvironment`'s constructor enabling manual flushing to disk.
```

execution_engine/src/shared/newtypes/mod.rs
```diff
@@ -5,14 +5,18 @@ use std::fmt::{self, Display, Formatter};
 use serde::Serialize;
 use uuid::Uuid;
 
+/// A correlation id is a unique representation of an event occuring
+/// in the system as it passes through different components.
```
Fraser999: I think this type is only used inside EE as opposed to across different components.
```suggestion
/// A correlation id is a unique identifier which can be used to track the progress of a given
/// execution engine operation.
```

execution_engine/src/shared/newtypes/mod.rs
```diff
@@ -5,14 +5,18 @@ use std::fmt::{self, Display, Formatter};
...
 #[derive(Clone, Copy, Debug, Default, Hash, PartialEq, Serialize)]
 pub struct CorrelationId(Uuid);
 
 impl CorrelationId {
+    /// Creates new unique `CorrelationId`
```
Fraser999:
```suggestion
    /// Creates new unique `CorrelationId`.
```

execution_engine/src/shared/opcode_costs.rs
```diff
@@ -1,3 +1,4 @@
+//! Support for WASM opcode costs.
```
Fraser999: WASM should be Wasm throughout the changes in this PR.

execution_engine/src/shared/opcode_costs.rs
```diff
@@ -7,28 +8,48 @@ use serde::{Deserialize, Serialize};
...
 pub const DEFAULT_INTEGER_COMPARISON_COST: u32 = 250;
+/// Default cost of the `conversion` WASM opcode.
 pub const DEFAULT_CONVERSION_COST: u32 = 420;
+/// Default cost of the `unreachable` WASM opcode.
 pub const DEFAULT_UNREACHABLE_COST: u32 = 270;
-pub const DEFAULT_NOP_COST: u32 = 200; // TODO: This value is not researched
+/// Default cost of the `nop` WASM opcode.
+pub const DEFAULT_NOP_COST: u32 = 200;
+// TODO: This value is not researched.
+/// Default cost of the `current_memory` WASM opcode.
```
Fraser999:
```suggestion
/// Default cost of the `nop` WASM opcode.
// TODO: This value is not researched.
pub const DEFAULT_NOP_COST: u32 = 200;
/// Default cost of the `current_memory` WASM opcode.
```

execution_engine/src/shared/transform.rs
```diff
@@ -49,17 +52,34 @@ impl From<CLValueError> for Error {
...
+    /// Produced by add operations.
     AddInt32(i32),
+    /// Adds an unsigned 64-bit integer to an existing entry in the global state.
     AddUInt64(u64),
+    /// Adds an unsigned 128-bit integer to an existing entry in the global state.
     AddUInt128(U128),
+    /// Adds an unsigned 256-bit integer to an existing entry in the global state.
     AddUInt256(U256),
+    /// Adds an unsigned 512-bit integer to an existing entry in the global state.
     AddUInt512(U512),
```
Fraser999:
```suggestion
    /// A wrapping addition of an `i32` to an existing numeric value (not necessarily an `i32`) in
    /// the global state.
    AddInt32(i32),
    /// A wrapping addition of a `u64` to an existing numeric value (not necessarily an `u64`) in
    /// the global state.
    AddUInt64(u64),
    /// A wrapping addition of a `U128` to an existing numeric value (not necessarily an `U128`) in
    /// the global state.
    AddUInt128(U128),
    /// A wrapping addition of a `U256` to an existing numeric value (not necessarily an `U256`) in
    /// the global state.
    AddUInt256(U256),
    /// A wrapping addition of a `U512` to an existing numeric value (not necessarily an `U512`) in
    /// the global state.
    AddUInt512(U512),
```

execution_engine/src/shared/transform.rs
```diff
@@ -137,6 +157,12 @@ where
 }
 
 impl Transform {
+    /// Applies the transformation on a specified stored value instance.
+    ///
+    /// This method produces a new [`StoredValue`] instance based on the [`Transform`] variant.
+    ///
+    /// It is important to note that all arithmetic operations are deterministic and produce
+    /// wrapped values.
```
Fraser999: I think maybe the point which we were aiming to make here is that all variants (not just numeric ones) of `Transform` are commutative, meaning that a given collection of them can be executed in any order to produce the same end result.  I believe the idea was that this would allow parallelism when applying transformations.

We should probably document that as part of the doc comment on the `Transform` type itself rather than on this method.

execution_engine/src/shared/wasm_config.rs
```diff
@@ -8,24 +9,31 @@ use super::{
...
 pub const DEFAULT_MAX_STACK_HEIGHT: u32 = 64 * 1024;
 
+/// Configuration of the WASM execution environment.
+///
+/// This structure contains various WASM execution configuration options, such as memory limits,
+/// stack limits and costs.
 #[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Debug, DataSize)]
 pub struct WasmConfig {
-    /// Maximum amount of a heap memory (represented in 64kb pages) each contract can use.
+    /// Maximum amount of heap memory (represented in 64kb pages) each contract can use.
```
Fraser999:
```suggestion
    /// Maximum amount of heap memory (represented in 64kB pages) each contract can use.
```

execution_engine/src/shared/wasm_config.rs
```diff
@@ -42,14 +50,17 @@ impl WasmConfig {
         }
     }
 
+    /// Return opcode costs.
```
Fraser999:
```suggestion
    /// Returns opcode costs.
```

Fraser999 (commented):
Some nitpicks, but this is looking really good.

execution_engine/src/shared/opcode_costs.rs
mpapierski: Addressed in 08e23c47b6692c20d4ed700d455535051d913777

execution_engine/src/shared/transform.rs
mpapierski: Addressed

execution_engine/src/shared/transform.rs
```diff
@@ -49,17 +52,41 @@ impl From<CLValueError> for Error {
     }
 }
 
+/// Representation of a single transformation ocurring during execution.
+///
+/// It is important to note that all arithmetic operations represented by different varaints are
+/// deterministic and produce wrapped values. In other words we're aiming to make all variants of
+/// [`Transform`] commutative which means that a given collection of them can be executed in any
```
luis-casper: I don't see how `Write` variants can be commuted, so I think this comment needs to be fixed.

execution_engine/src/shared/transform.rs
```diff
@@ -49,17 +52,39 @@ impl From<CLValueError> for Error {
     }
 }
 
+/// Representation of a single transformation ocurring during execution.
+///
+/// Note that all variants of [`Transform`] are commutative which means that a given collection of
```
luis-casper: After discussion with @Fraser999 and @goral09, we concluded the `Write` variant isn't commutative and others, such as `Failure`, are not as well.
Thus I believe we should remove or clarify it, i.e., change the term “all variants” to “all `Add*` variants”.

Fraser999 (approved):

luis-casper (approved):

execution_engine/src/shared/transform.rs
```diff
@@ -54,8 +54,8 @@ impl From<CLValueError> for Error {
 
 /// Representation of a single transformation ocurring during execution.
 ///
-/// Note that all variants of [`Transform`] are commutative which means that a given collection of
-/// them can be executed in any order to produce the same end result.
+/// Note that all arithmetic variants of [`Transform`] are commutative which means that a given
```
goral09, line 57: This still isn't precise IMO - why not simply say that all `Add*` variants are commutative? What does "arithmetic" stands for? is `AddNamedKeys` arithmetic? Also, if there is no other "arithmetic" operation that `Add` (like there's no `Subtract`) we shouldn't try to be too smart about it.

PR #2118, opened 2021-09-22
Issue 2047 asymmetric type from hex

### LINKED ISSUE #2047: Fix AsymmetricType::from_hex for system variant
Should be updated to
```rust
match tag[0] {
    SYSTEM_TAG => {
        if key_bytes.is_empty() {
            return Ok(Self::system());
        }
        Err(Error::AsymmetricKey(format!("failed to decode from hex: invalid system variant")))
    }
    ED25519_TAG => {
        let bytes = hex::decode(key_bytes)?;
        Self::ed25519_from_bytes(&bytes)
    }
    SECP256K1_TAG => {
        let bytes = hex::decode(key_bytes)?;
        Self::secp256k1_from_bytes(&bytes)
    }
    _ => Err(Error::AsymmetricKey(format!(
        "failed to decode from hex: invalid tag: expected {} or {}, got {}",
        ED25519_TAG, SECP256K1_TAG, tag[0]
    ))),
}

```
and tests added

luis-casper:
Fully addresses the following issue:
https://app.zenhub.com/workspaces/engineering-60953fafb1945f0011a3592d/issues/casper-network/casper-node/2047


types/src/crypto/asymmetric_key.rs
```diff
@@ -743,3 +749,104 @@ mod detail {
...
+
+        #[test]
+        fn too_short() {
+            assert!(PublicKey::from_hex([0u8; 0]).is_err());
+            assert!(PublicKey::from_hex([0u8; 1]).is_err());
+        }
+
+        // The following atrocity is due to the crates `rand_core` and
+        // `k256::elliptic_curve::rand_core` not necessarily being the same.
+        mod myrng {
```
xcthulhu: Maybe use a [proptest][1] instead of this? We already have `crate::crypto::assymetric_key::public_key_arb()`...

[1]: https://docs.rs/proptest/1.0.0/proptest/

luis-casper: That's an awesome review and I learned a lot, thanks.
I think I should start over with a new PR.

luis-casper: I backtracked this PR and added new commits from the start.

xcthulhu (approved):
LGTM

luis-casper:
bors merge

Fraser999 (approved):

Fraser999 (approved):

Fraser999 (dismissed):
After further discussion, I think we should revert the behaviour of `AsymmetricType::from_hex()` so that it remains an error to try to parse a system key/sig from hex.  There doesn't appear to be a valid use-case for such functionality.

Let's keep the additional tests added in this PR, but update them as required to match the original behaviour.

xcthulhu:
>After further discussion, I think we should revert the behaviour of AsymmetricType::from_hex() so that it remains an error to try to parse a system key/sig from hex. There doesn't appear to be a valid use-case for such functionality.

@Fraser999 So actually 🤓  we may need this for `fast-sync`.  There's some parts here you know better than I, though. 

In the fast sync branch we make a switch block after each upgrade.  The block proposer for that block is the system contract.  I don't know the client code that well.  Will it be the case that the client will need to parse the hex for the system contract?  Perhaps there is some other encoding it uses...

luis-casper:
> After further discussion, I think we should revert the behaviour of `AsymmetricType::from_hex()` so that it remains an error to try to parse a system key/sig from hex. There doesn't appear to be a valid use-case for such functionality.
> 
> Let's keep the additional tests added in this PR, but update them as required to match the original behaviour.

Done.

types/src/crypto/asymmetric_key.rs
```diff
@@ -81,17 +81,20 @@ where
...
 
-        let tag = checksummed_hex::decode(&tag_bytes)?;
-        let key_bytes = checksummed_hex::decode(&key_bytes)?;
+        let tag = checksummed_hex::decode(&tag_hex)?;
+        let key_bytes = checksummed_hex::decode(&key_hex)?;
 
         match tag[0] {
+            SYSTEM_TAG => Err(Error::AsymmetricKey(
+                "failed to decode from hex: cannot decode system variant".to_string(),
+            )),
```
goral09: I think we should allow this. If someone tries to use the system key, as a _signing key_, then it should fail on signature validation later. In Ethereum, a common practice for burning tokens is to send them to an address that no key exists for - 0x00000... I don't see a reason with using it in other contexts (like an argument).

afck:
Yes, for fast sync we need to decide how we handle the automatic switch block. In the current implementation [its proposer is `PublicKey::System`](https://github.com/casper-network/casper-node/blob/be209cb863a911d3b5f590bf762a99acd637a17f/node/src/reactor/participating.rs#L649). An alternative would be to introduce a new special public key value instead. :confused:

luis-casper:
Even though we might drop this one, I've left it complete in a usable state.

Fraser999 (approved):

goral09 (approved):

Issue #1915, opened 2021-08-13
Keep Only The Last Two Eras In Consensus


xcthulhu:

afck:
We should disentangle these questions:

#### How many earlier switch blocks do you need to initialize an era?

To initialize era `n`, you need at least:
* The era's booking block, which is in era `n - auction_delay - 1`.
* The era's key block, from era `n - 1`: for the random seed, and so that we can exclude inactive validators from becoming leaders.
* All blocks in between those, so we can ban all validators that equivocated in the meantime.

#### Across how many eras can you cite evidence?

This is currently the maximum: You can put an accusation into a new block in era `N`, if any bonded era (i.e. no older than `unbonding_delay - auction_delay`) contains evidence.

We are going to reduce this value to 1, assuming that in practice an equivocation in an era older than the previous one will have no effect anyway.

#### How many past eras does the era supervisor need to keep in memory?

Based on the previous point, we won't be able to punish validators who equivocate in any era earlier than the previous one anymore anyway, so for equivocation detection there's no point in keeping more than that.
However, with each _full_ era (including units) we keep, we also need to keep enough information — the validator set and who was banned and excluded from the leader sequence — about the preceding era to validate evidence, since that could be cited by units from the full era.

The other reason to keep old eras around is to help other nodes to catch up; with [CEP 36](https://github.com/casper-network/ceps/blob/master/text/0036-consensus-memory-usage.md) this won't be necessary anymore, and it's already questionable whether we need to help nodes that fell behind by more than one era: they should restart instead, and sync again.

#### In summary:

While in era `N`:
* The era supervisor will fully keep eras `N` and `N - 1` in memory.
* It keeps basic information about era `N - 2` for validating evidence.
* For initialization it needs all switch blocks starting at era `N - 3 - auction_delay`.

piotr-dziubecki:
@goral09 @afck - is this already addressed?

afck:
No, not done yet.

afck:
Note that this has only been merged to the fast sync feature branch, not to `dev` yet.

Issue #1588, opened 2021-06-14
Unify era initialization code.


afck:
The separate methods for era initialization in a newly joining vs. already running node caused #1587 and #1558. Make sure the same code is used in both cases, for retrieving the switch blocks and booking blocks, and computing the validators and their states (active, banned, excluded from proposing, …).

afck:
This is being done as part of https://github.com/casper-network/casper-node/issues/1915.

PR #2127, opened 2021-09-24
Eras in memory

afck:
Keep only two full eras in memory in the era supervisor, and one additional era for evidence validation only. See https://github.com/casper-network/casper-node/issues/1915 for details.

Closes https://github.com/casper-network/casper-node/issues/1915, https://github.com/casper-network/casper-node/issues/1588.

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -343,25 +343,12 @@ where
             Ok((era_id, outcomes)) => {
                 self.handle_consensus_outcomes(effect_builder, rng, era_id, outcomes)
             }
-            Err(CreateNewEraError::AttemptedToCreateEraWithNoSwitchBlocks) => {
-                return fatal!(
-                    effect_builder,
-                    "attempted to create era with no switch blocks; this is a bug",
-                )
-                .ignore();
-            }
-            Err(CreateNewEraError::LastBlockHeaderNotASwitchBlock {
-                era_id,
-                last_block_header,
-            }) => {
-                return fatal!(
-                    effect_builder,
-                    "attempted to create {} with non-switch block {}; this is a bug",
-                    era_id,
-                    last_block_header,
-                )
-                .ignore();
-            }
+            Err(err) => fatal!(
+                effect_builder,
+                "failed to create era; this is a bug: {:?}",
+                err,
+            )
+            .ignore(),
         }
     }
 
@@ -381,6 +368,24 @@ where
                 last_block_header: Box::new(key_block.clone()),
             }
         })?;
+
+        let earliest_era = self.chainspec.earliest_switch_block_needed(era_id);
+        let switch_blocks_needed = era_id.value().saturating_sub(earliest_era.value()) as usize;
+        let first_idx = switch_blocks
+            .len()
+            .checked_sub(switch_blocks_needed)
+            .ok_or_else(|| CreateNewEraError::InsufficientSwitchBlocks {
+                era_id,
+                switch_blocks: switch_blocks.to_vec(),
+            })?;
+        for (i, switch_block) in switch_blocks[first_idx..].iter().enumerate() {
+            if switch_block.era_id() != earliest_era.saturating_add(i as u64) {
+                return Err(CreateNewEraError::WrongSwitchBlockEra {
+                    era_id,
+                    switch_blocks: switch_blocks.to_vec(),
+                });
+            }
+        }
 
         let report = era_end.era_report();
         let validators = era_end.next_era_validator_weights();
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -87,10 +87,17 @@ type ConsensusConstructor<I> = dyn Fn(
 
 #[derive(DataSize)]
 pub struct EraSupervisor<I> {
-    /// A map of active consensus protocols.
-    /// A value is a trait so that we can run different consensus protocol instances per era.
+    /// A map of active consensus protocol instances.
+    /// A value is a trait so that we can run different consensus protocols per era.
     ///
-    /// This map always contains exactly three entries, with the last one being the current one.
+    /// This map contains three consecutive entries, with the last one being the current era N. Era
+    /// N - 1 is also kept in memory so that we would still detect any equivocations there and use
+    /// them in era N to get the equivocator banned. And era N - 2 one is in a frozen state: It
+    /// doesn't accept any new Highway units anymore, but we keep the instance in memory so we can
+    /// evaluate evidence that units in era N - 1 might cite.
+    ///
+    /// Since eras at or before the most recent activation point are never instantiated, shortly
+    /// after that there can temporarily be fewer than three entries in the map.
     active_eras: HashMap<EraId, Era<I>>,
     secret_signing_key: Arc<SecretKey>,
     public_signing_key: PublicKey,
@@ -184,12 +191,17 @@ where
         let auction_delay = era_supervisor.protocol_config.auction_delay;
         let last_activation_point = era_supervisor.protocol_config.last_activation_point;
         let mut switch_blocks = Vec::new();
-        // We need to initialize current_era, current_era - 1 and current_era - 2. To initialize an
-        // era, all switch blocks between its booking block and its key block are required. The
-        // booking block for era N is in N - auction_delay - 1, and the key block in N - 1.
-        // So we need all switch blocks between (including) current_era - 2 - auction_delay - 1
-        // and (excluding) current_era. However, we never use any block from before the last
-        // activation point.
+
+        // We need to initialize current_era, current_era - 1 and (frozen) current_era - 2. To
+        // initialize an era, all switch blocks between its booking block and its key block are
+        // required. The booking block for era N is in N - auction_delay - 1, and the key block in
+        // N - 1. So we need all switch blocks between:
+        // (including) current_era - 2 - auction_delay - 1 and (excluding) current_era.
+        // However, we never use any block from before the last activation point.
+        //
+        // Example: If auction_delay is 1, to initialize era N we need the switch blocks from era N
+        // and N - 1. If current_era is 10, we will initialize eras 10, 9 and 8. So we need the
+        // switch blocks from eras 9, 8, 7 and 6.
         let earliest_era =
             last_activation_point.max(current_era.saturating_sub(auction_delay).saturating_sub(3));
         for era_id in (earliest_era.value()..current_era.value()).map(EraId::from) {
@@ -198,10 +210,15 @@ where
                 .ok_or_else(|| anyhow::Error::msg("No such switch block"))?;
             switch_blocks.push(switch_block);
         }
+
         // The create_new_era method initializes the era that the slice's last block is the key
         // block for. We want to initialize the three latest eras, so we have to pass in the whole
         // slice for the current era, and omit one or two elements for the other two. We never
         // initialize the last_activation_point or an earlier era, however.
+        //
+        // In the example above, we would call create_new_era with the switch blocks from eras
+        // 8 and 9 (to initialize 10) then 7 and 8 (for era 9), and finally 6 and 7 (for era 8).
+        // (We don't truncate the slice at the start since unneeded blocks are ignored.)
         let mut effects = Effects::new();
         for i in (switch_blocks.len().saturating_sub(2).max(1)..=switch_blocks.len()).rev() {
             effects.extend(era_supervisor.create_new_era(effect_builder, rng, &switch_blocks[..i]));
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -67,8 +67,12 @@ pub use self::era::Era;
 /// fault tolerance threshold.
 const FTT_EXCEEDED_SHUTDOWN_DELAY_MILLIS: u64 = 60 * 1000;
 
+/// The number of eras across which evidence can be cited.
+/// If this is 1, you can cite evidence from the previous era, but not the one before that.
+/// To be able to detect that evidence, we also keep that number of full past eras in memory.
+const PAST_EVIDENCE_ERAS: u64 = 1;
 /// The number of active eras that are kept in memory in addition to the current one.
-const PAST_ACTIVE_ERAS: u64 = 2;
+const PAST_ACTIVE_ERAS: u64 = 2 * PAST_EVIDENCE_ERAS;
 
 type ConsensusConstructor<I> = dyn Fn(
         Digest,                    // the era's unique instance ID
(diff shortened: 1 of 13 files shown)


node/src/types/block.rs
```diff
@@ -717,7 +717,7 @@ impl BlockHeader {
     }
 
     /// Returns reward and slashing information if this is the era's last block.
-    pub fn era_end(&self) -> Option<&EraReport> {
+    pub fn era_report(&self) -> Option<&EraReport> {
```
xcthulhu: 👍

node/src/components/consensus/era_supervisor.rs
```diff
@@ -181,47 +180,32 @@ where
...
-                .protocol_config
-                .last_emergency_restart
-                .unwrap_or_else(|| EraId::from(0u64));
-            let key_block_era_id = era_id.saturating_sub(1);
-            let block_header = storage
-                .read_switch_block_header_by_era_id(key_block_era_id)?
+        let auction_delay = era_supervisor.protocol_config.auction_delay;
+        let last_activation_point = era_supervisor.protocol_config.last_activation_point;
+        let mut switch_blocks = Vec::new();
+        // We need to initialize current_era, current_era - 1 and current_era - 2. To initialize an
```
xcthulhu: Consider changing to:

`// We need to initialize current_era, current_era - 1 and (frozen) current_era - 2. To initialize an`

node/src/components/consensus/era_supervisor.rs
```diff
@@ -181,47 +180,32 @@ where
...
-                    .ok_or_else(|| anyhow::Error::msg("No such switch block"))?
-                    .hash()
-            };
-            booking_blocks.insert(era_id, booking_block_hash);
+            switch_blocks.push(switch_block);
+        }
+        // The create_new_era method initializes the era that the slice's last block is the key
+        // block for. We want to initialize the three latest eras, so we have to pass in the whole
+        // slice for the current era, and omit one or two elements for the other two. We never
+        // initialize the last_activation_point or an earlier era, however.
```
xcthulhu: Consider a worked example with `Era(5)`, `Era(4)` and `Era(3)`

node/src/components/consensus/era_supervisor.rs
```diff
@@ -278,124 +262,210 @@ where
...
+                // (b/c it would have been from before Genesis, upgrade or emergency restart),
+                // use a "zero" block hash. This should not hurt the security of the leader
+                // selection algorithm.
+                BlockHash::default()
+            };
+        let seed = Self::era_seed(booking_block_hash, key_block.accumulated_seed());
+
+        // The beginning of the new era is marked by the key block.
+        #[allow(clippy::integer_arithmetic)] // Block height should never reach u64::MAX.
+        let start_height = key_block.height() + 1;
```
xcthulhu, line 418: Add to `Block::verify` (or maybe `BlockHeader::verify`?) a check for `height < u64::MAX / 2`.

node/src/components/linear_chain_sync/operations.rs
```diff
@@ -463,12 +462,9 @@ pub(crate) async fn run_fast_sync_task(
...
 
-    // The era supervisor needs validator information from previous eras.
-    // The number of previous eras is determined by a *delay* in which consensus participants become
-    // bonded validators or unbond.
-    let delay = consensus::bonded_eras(&(&chainspec).into());
-    // The era supervisor requires at least to 3*delay + 1 eras back to be stored in the database.
-    let historical_eras_needed = delay.saturating_mul(3).saturating_add(1);
+    // The era supervisor requires at least auction_delay + 3 eras back to be stored in the
+    // database.
+    let historical_eras_needed = chainspec.core_config.auction_delay.saturating_add(3);
```
goral09: Could we hide it behind `consensus::bonded_eras` (or similar) as well? I'm not a fan of hardcoding these assumptions all over the code. It's easy to miss and get out of sync.

node/src/components/consensus/era_supervisor.rs
afck: I'm still not sure about this: It's a very different kind of "invalid" than the hashes not matching.

E.g. if we asked for a block by hash, and its hashes don't match (current `verify` would fail), the fetcher should continue asking other peers for the block. But if the hashes do match and only the height is invalid, that means it should fail, and we should drop all blocks that referred to this one by hash.

node/src/components/consensus/era_supervisor.rs
goral09: Not sure if usage of word "frozen" makes it more clear. What does it even mean?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -181,47 +180,32 @@ where
...
+        let last_activation_point = era_supervisor.protocol_config.last_activation_point;
+        let mut switch_blocks = Vec::new();
+        // We need to initialize current_era, current_era - 1 and current_era - 2. To initialize an
+        // era, all switch blocks between its booking block and its key block are required. The
+        // booking block for era N is in N - auction_delay - 1, and the key block in N - 1.
+        // So we need all switch blocks between (including) current_era - 2 - auction_delay - 1
+        // and (excluding) current_era. However, we never use any block from before the last
+        // activation point.
+        let earliest_era =
+            last_activation_point.max(current_era.saturating_sub(auction_delay).saturating_sub(3));
```
goral09: Can we extract the `3` from the `saturating_sub` to a constant and give it a descriptive name?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -181,47 +180,32 @@ where
...
+        for era_id in (earliest_era.value()..current_era.value()).map(EraId::from) {
+            let switch_block = storage
+                .read_switch_block_header_by_era_id(era_id)?
                 .ok_or_else(|| anyhow::Error::msg("No such switch block"))?;
```
goral09: Can you add an `era_id` to the error message please?

node/src/components/consensus/era_supervisor.rs
afck: I'll add an explaining comment to the `active_eras` field:
So far I've used the term "evidence only" for this; it means that we deleted all the era's units and won't accept any new units, because we just keep the era around so we can verify evidence.

Not sure if "frozen" or "evidence only" is better?

goral09: Definitely prefer "evidence-only".

node/src/components/linear_chain_sync/operations.rs
afck: I'm working on it. But I'll have to change the `EraSupervisor` to keep a `Chainspec` instead of a `ProtocolConfig`… that type should probably be removed anyway at some point.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/273073a47ece41dfdb75419b614c945b9fb11868.

node/src/components/linear_chain_sync/operations.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/429a2b1af6acf011c0a141bc198e9bb52ecb75b8.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/429a2b1af6acf011c0a141bc198e9bb52ecb75b8.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/429a2b1af6acf011c0a141bc198e9bb52ecb75b8.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/c8b83472b3c9051e2e38ac016d940b33ecb2e37e.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -181,68 +190,47 @@ where
...
+        // We need to initialize current_era, current_era - 1 and (evidence-only) current_era - 2.
+        // To initialize an era, all switch blocks between its booking block and its key block are
+        // required. The booking block for era N is in N - auction_delay - 1, and the key block in
+        // N - 1. So we need all switch blocks between:
+        // (including) current_era - 2 - auction_delay - 1 and (excluding) current_era.
+        // However, we never use any block from before the last activation point.
+        //
+        // Example: If auction_delay is 1, to initialize era N we need the switch blocks from era N
+        // and N - 1. If current_era is 10, we will initialize eras 10, 9 and 8. So we need the
+        // switch blocks from eras 9, 8, 7 and 6.
```
goral09: This is a really good explanation that will definitely make it easier for people to follow the code. Thanks 👍

node/src/components/consensus/era_supervisor.rs
```diff
@@ -298,124 +286,210 @@ where
...
+                .ignore();
+            }
+        };
+
         if self.active_eras.contains_key(&era_id) {
-            panic!("{} already exists", era_id);
+            warn!(era = era_id.value(), "era already exists");
+            return Effects::new();
+        }
+        if self.current_era > era_id.saturating_add(2) {
```
goral09:
```suggestion
        if self.current_era > era_id.saturating_add(PAST_ACTIVE_ERAS) {
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -298,124 +286,210 @@ where
...
                 era.consensus.set_evidence_only();
             }
         }
-        // Remove the era that has become obsolete now: The oldest bonded era could still receive
-        // units that refer to evidence from any era that was bonded when it was the current one.
-        let oldest_evidence_era_id = oldest_bonded_era(&self.protocol_config, oldest_bonded_era_id);
-        if let Some(obsolete_era_id) = oldest_evidence_era_id.checked_sub(1) {
+
+        // Remove the era that has become obsolete now: We keep only three in memory.
+        if let Some(obsolete_era_id) = self.current_era.checked_sub(3) {
```
goral09: Can we use `PAST_ACTIVE_ERAS + 1` here, instead of `3`?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -723,32 +599,38 @@ where
...
+        rng: &mut NodeRng,
         new_block_payload: NewBlockPayload,
     ) -> Effects<Event<I>> {
         let NewBlockPayload {
             era_id,
             block_payload,
             block_context,
         } = new_block_payload;
-        if !self.era_supervisor.is_bonded(era_id) {
+        if !self.active_eras.contains_key(&era_id) || era_id.successor() < self.current_era {
```
goral09: If we switched places of these statements, then the check could be slightly cheaper as comparing integers will be faster than checking set membership.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -881,60 +701,68 @@ where
...
+            .flat_map(|result| self.handle_consensus_outcome(effect_builder, rng, era_id, result))
             .collect()
     }
 
     /// Returns `true` if any of the most recent eras has evidence against the validator with key
     /// `pub_key`.
     fn has_evidence(&self, era_id: EraId, pub_key: PublicKey) -> bool {
-        self.era_supervisor
-            .iter_past(era_id, self.era_supervisor.bonded_eras())
+        self.iter_past(era_id, 1)
```
goral09: Can this `1` be extracted to a constant as well?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1116,53 +935,48 @@ where
...
-                    .era_supervisor
-                    .iter_future(era_id, self.era_supervisor.bonded_eras())
-                {
-                    let proposed_blocks =
-                        if let Some(era) = self.era_supervisor.active_eras.get_mut(&e_id) {
-                            era.resolve_evidence_and_mark_faulty(&pub_key)
-                        } else {
-                            continue;
-                        };
+                for e_id in self.iter_future(era_id, 1) {
```
goral09: Here we would use that new constant I left a comment about earlier.

node/src/components/consensus/era_supervisor.rs
```diff
@@ -1177,37 +991,96 @@ where
...
+        self.active_eras.keys()
+    }
+}
+
+/// Returns all switch blocks needed to initialize `era_id`.
+///
+/// Those are the booking block, i.e. the switch block in `era_id - auction_delay - 1`,
+/// the key block, i.e. the switch block in `era_id - 1`, and all switch blocks in between.
+async fn get_switch_blocks<REv>(
+    chainspec: Arc<Chainspec>,
```
goral09, line 1098: Do we need a whole `Chainspec` or would that new type you had added suffice?

afck: I _removed_ `ProtocolConfig`, and added a _trait_ `ChainspecConsensusExt`.

goral09 (approved):
Ńo blockers.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/f417ff94706385eda1d3d134cadcc7fec3f6dfd0.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/f417ff94706385eda1d3d134cadcc7fec3f6dfd0.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/f417ff94706385eda1d3d134cadcc7fec3f6dfd0.

node/src/components/consensus/era_supervisor.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2127/commits/f417ff94706385eda1d3d134cadcc7fec3f6dfd0.

afck:
@goral09: Please double-check the latest commit; just want to make sure I didn't introduce any bugs with the constants. :sweat_smile:

node/src/components/consensus/era_supervisor.rs
```diff
@@ -67,8 +67,12 @@ pub use self::era::Era;
 /// fault tolerance threshold.
 const FTT_EXCEEDED_SHUTDOWN_DELAY_MILLIS: u64 = 60 * 1000;
 
+/// The number of eras across which evidence can be cited.
+/// If this is 1, you can cite evidence from the previous era, but not the one before that.
+/// To be able to detect that evidence, we also keep that number of full past eras in memory.
+const PAST_EVIDENCE_ERAS: u64 = 1;
 /// The number of active eras that are kept in memory in addition to the current one.
-const PAST_ACTIVE_ERAS: u64 = 2;
+const PAST_ACTIVE_ERAS: u64 = 2 * PAST_EVIDENCE_ERAS;
```
goral09: Is it really related?

afck: I think so, yes:
We want to keep the past `PAST_EVIDENCE_ERAS` around as _full_ eras so that we can detect all the equivocations we would be allowed to cite.
Then, we have to keep a further `PAST_EVIDENCE_ERAS` in evidence-only mode because the earliest full era could cite evidence from them.

xcthulhu (approved):

afck:
bors merge

afck:
(Failed due to `lsof` issue. retrying.)

afck:
bors merge

afck:
bors merge

afck:
Bors currently doesn't work with the `feat-fast-sync` branch. Merging manually.

PR #2131, opened 2021-09-24
Perform account checks on deploys from clients only.

### LINKED ISSUE #2130: Deploy Acceptor should not validate account when deploy received from peer

darthsiroftardis:
CHANGELOG:

- Changed the behavior of the deploy acceptor to never perform account checks on deploys sent by peers.
- Added two tests to ensure account checks for deploys sent by peers don't occur

Closes #2130

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,9 +331,9 @@ dependencies = [
 
 [[package]]
 name = "bstr"
-version = "0.2.16"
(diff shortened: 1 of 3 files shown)


Fraser999 (approved):
Optimistic approval given https://github.com/darthsiroftardis/casper-node/pull/13 is merged first.

PR #2132, opened 2021-09-25
Fix minimum penalty charge

mpapierski:
Co-authored-by: Ed Hastings <ed@casperlabs.io>
Co-authored-by: Fraser Hutchison <fraser@casperlabs.io>
Co-authored-by: Karan Dhareshwar <karan@casperlabs.io>

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -26,6 +26,7 @@ use std::{
     rc::Rc,
 };
 
+use num::Zero;
 use num_rational::Ratio;
 use once_cell::sync::Lazy;
 use tracing::{debug, error};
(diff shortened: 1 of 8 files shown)


smart_contracts/contracts/test/do-nothing-stored-upgrader/src/main.rs
```diff
@@ -24,6 +24,7 @@ const ARG_PURSE_NAME: &str = "purse_name";
 
 #[no_mangle]
 pub extern "C" fn delegate() {
+    let _named_keys = runtime::list_named_keys();
```
Fraser999, line 27: Should comment why this is necessary while the contract is still named "do-nothing".  Same for other two similar contracts.

execution_engine_testing/tests/src/test/regression/regression_20210924.rs
```diff
@@ -0,0 +1,209 @@
...
+        proposer_balance_after
+    );
+}
+
+#[ignore]
+#[test]
+fn should_execute_do_minimum_session() {
+    let minimum_deploy_payment = U512::from(DEFAULT_NOP_COST);
+
+    let do_nothing_request = {
```
Fraser999:
```suggestion
    let do_minimum_request = {
```

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1576,6 +1602,29 @@ where
             );
         }
 
+        // Session execution was zero cost.
+        // Check if the payment purse can cover the minimum floor for session execution.
+        if session_result.cost().is_zero() && payment_purse_balance < max_payment_cost {
+            // When session code structure is invalid but still has 0 cost we should propagate the
```
```suggestion
            // When session code structure is valid but still has 0 cost we should propagate the
```

smart_contracts/contracts/test/do-nothing-stored-upgrader/src/main.rs
EdHastingsCasperAssociation: should rename the do-nothing family of contracts to "do-nothing-useful..." or "exercise-lower-gas-bounds..." or something like that

EdHastingsCasperAssociation (approved):
optimistic approval, once nitpicks have been considered

Fraser999 (approved):
Optimistic approval given the nitpicks are addressed.

smart_contracts/contracts/test/do-nothing-stored-upgrader/src/main.rs
mpapierski: I agree - I'll address it adequately in a separate ticket

PR #2133, opened 2021-09-27
Improve and add logging to run_equivocator_network test.

### LINKED ISSUE #1859: Non-det failure of `run_equivocator_network` test.

afck:
The test is flaky on CI but always passes locally. This adds some logging so we can see whether it's a timing issue.

We also start isolating one of the two clone nodes for three rounds starting from the first _consensus_ message it sends now, instead of the first message it receives: If other messages are exchanged before the genesis timestamp (address gossiping?), the three round lengths could expire before the first consensus message is even sent.

See issue #1859.

### DIFF
--- a/node/src/reactor/participating/tests.rs
+++ b/node/src/reactor/participating/tests.rs
@@ -326,25 +326,25 @@ async fn run_equivocator_network() {
         .await
         .expect("network initialization failed");
     let min_round_len = chain.chainspec.highway_config.min_round_length();
-    let mut maybe_first_message = None;
+    let mut maybe_first_message_time = None;
     net.reactors_mut()
         .find(|reactor| *reactor.inner().consensus().public_key() == alice_pk)
         .unwrap()
         .set_filter(move |event| {
             let now = Timestamp::now();
-            let first_message = match (&event, maybe_first_message) {
+            let first_message_time = match (&event, maybe_first_message_time) {
                 (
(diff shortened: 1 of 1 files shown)


node/src/reactor/participating/tests.rs
```diff
@@ -325,19 +331,18 @@ async fn run_equivocator_network() {
...
-            ) {
-                return Either::Right(event);
-            }
             let now = Timestamp::now();
-            let first_message = if let Some(first_message) = maybe_first_message {
-                first_message
-            } else {
-                maybe_first_message = Some(now);
-                now
+            let first_message = match (&event, maybe_first_message) {
```
goral09: It's first message's timestamp, right? Can we change the name?

goral09 (approved):

xcthulhu (approved):
LGTM

node/src/reactor/participating/tests.rs
afck: Done in https://github.com/casper-network/casper-node/pull/2133/commits/afaeaa465e49f1878bf902f5bdb415bd6b972df7.

afck:
bors merge

afck:
bors cancel

afck:
bors merge

afck:
bors merge

afck:
bors merge

PR #2168, opened 2021-09-28
dry-run-deploys tool

### LINKED ISSUE #2121: Retrieve State Tool Should Not Do A Directory Walk to Check For Existing Blocks
The proposed implementation of the retrieve state tool uses a directory walk to check for existing blocks before resuming (see [here][1]).

On slow disks this does not have great performance.

[1]: https://github.com/casper-network/casper-node/pull/2102/files#r713313001

dwerner:
Resolves #2121 and brings in `dry-run-deploys` cli tool.

Should be applied after #2102 and replaces #2032.


execution_engine/src/storage.rs
```diff
@@ -11,7 +11,8 @@ pub mod trie;
 /// Merkle Trie storage.
 pub mod trie_store;
 
-const MAX_DBS: u32 = 2;
+/// Minimum number of databases needed for construction of an LMDB environment that the EE can use.
+const MIN_DBS: u32 = 2;
```
xcthulhu, line 15: 👍

dwerner:
Closing in favor of #2182

PR #2169, opened 2021-09-28
Add `ChecksummedHexEncode` trait.

nilz3ro:
# About These Changes

* Introduces a new trait `ChecksummedHexEncode<const N: usize>` with a `checksummed_hex_encode()` method.
* Implements `ChecksummedHexEncode` on a few `[u8; _]` types.


xcthulhu:
This trait does not appear to be necessary.  Is there an analogous trait that the `hex` crate introduces that we are copying here?  If not let's try to be API compatible with the `hex` crate.

types/src/checksummed_hex.rs
```diff
@@ -140,6 +139,18 @@ pub fn decode(input: &(impl AsRef<[u8]> + ?Sized)) -> Result<Vec<u8>, base16::De
...
+    /// More stuff
+    fn checksummed_hex_encode(&self) -> String {
+        encode_iter(self.as_ref()).collect()
+    }
+}
+
+impl ChecksummedHexEncode<8> for [u8; 8] {}
+impl ChecksummedHexEncode<16> for [u8; 16] {}
+impl ChecksummedHexEncode<32> for [u8; 32] {}
+impl ChecksummedHexEncode<64> for [u8; 64] {}
```
xcthulhu, line 153: Something like this should work

```suggestion
impl<N> ChecksummedHexEncode<N> for [u8; N] {}
```

That way it only monomorphizes over the array lengths we actually use.

types/src/checksummed_hex.rs
```diff
@@ -140,6 +139,18 @@ pub fn decode(input: &(impl AsRef<[u8]> + ?Sized)) -> Result<Vec<u8>, base16::De
     Ok(bytes)
 }
 
+/// Stuff
+pub trait ChecksummedHexEncode<const N: usize>: AsRef<[u8]> + AsMut<[u8]> {
+    /// More stuff
```
goral09, line 144: These cannot be the comments we want to put into the repo, correct? :)

PR #2174, opened 2021-09-29
GH-1930 Transform Journal

### LINKED ISSUE #1930: Tracking copy transforms should be in insertion order

luis-casper:
This is me taking over the following PR from @dwerner.
<https://github.com/casper-network/casper-node/pull/1957>

Related to Issue #1930.
<https://github.com/casper-network/casper-node/issues/1930>


goral09:
@luis-casper , if you open a PR and ask for a review it means it's ready to be reviewed. It looks like you're still working on this one and it's got a lot of merge conflicts. Either close a PR and re-open it once it's ready or put it back to a draft.

execution_engine/src/core/engine_state/execution_effect.rs
```diff
@@ -15,25 +17,27 @@ impl ExecutionEffect {
...
+                Transform::Write(_) => Op::Write,
+                Transform::AddInt32(_)
+                | Transform::AddUInt64(_)
+                | Transform::AddUInt128(_)
+                | Transform::AddUInt256(_)
+                | Transform::AddUInt512(_)
+                | Transform::AddKeys(_) => Op::Add,
+            };
+            ops.insert_add(key, op);
+            transforms.insert_add(key, transform);
```
mpapierski, line 35: I don't think this will preserve insertion order. Let's change the backing store inside AdditiveMap into https://crates.io/crates/indexmap and add tests to ensure transforms are now in insertion order. Collapsing a Vec into AdditiveMap (aka HashMap) will break the insertion order.

luis-casper:
Reopening.

types/src/execution_result.rs
```diff
@@ -287,39 +284,53 @@ impl FromBytes for ExecutionResult {
...
+    pub fn new(transforms: Vec<TransformEntry>) -> Self {
+        Self {
+            transforms,
+            operations: Default::default(),
+        }
+    }
+}
+
 impl ToBytes for ExecutionEffect {
+    #[allow(deprecated)]
```
dwerner: I thought we removed the `deprecated` attribute because it was causing schema changes. Is that not the case?

execution_engine/src/core/engine_state/execution_result.rs
```diff
@@ -472,110 +464,63 @@ impl ExecutionResultBuilder {
...
-
-        for (k, old_value) in kvs {
-            if let Some(Transform::Write(new_value)) = transforms.remove(&k) {
-                if new_value == old_value {
-                    transforms.insert(k, Transform::Identity);
-                    ops.insert(k, Op::Read);
-                } else {
-                    transforms.insert(k, Transform::Write(new_value));
-                }
-            }
```
dwerner, line 576: This feels wrong to me to simply remove this. Can you provide justification for it?

luis-casper: The main justification is that the code becomes much simpler. We are investigating the purpose of this, but so far it looks to me like an “optimization” that tries to avoid writes in the commit by reading from global state. It also looks like a vestige from when deploys were meant to commute, trying to reduce the strength of transforms to facilitate that.

types/src/execution_result.rs
luis-casper: This is a residue from the rebase. Should I remove that?

luis-casper: Removed.

execution_engine/src/core/engine_state/execution_effect.rs
dwerner: (As long as we are talking about types in execution_engine, as it isn't true for the casper_types ExecutionEffect), ExecutionEffect isn't used to preserve order. In those cases where order preservation is desirable, an ExecutionJournal is to be used.

dwerner (approved):

execution_engine/src/core/engine_state/execution_result.rs
luis-casper: For further evidence, see [the call site](https://github.com/luis-casper/casper-node/blob/release-1.3.2/execution_engine/src/core/engine_state/execution_result.rs#L438).

execution_engine/src/core/tracking_copy/tests.rs
```diff
@@ -282,11 +258,9 @@ fn tracking_copy_rw() {
     let value = StoredValue::CLValue(CLValue::from_t(3_i32).unwrap());
     let _ = tc.read(correlation_id, &k);
     tc.write(k, value.clone());
-    assert_eq!(tc.ops.len(), 1);
-    assert_eq!(tc.ops.get(&k), Some(&Op::Write));
```
goral09, line 269: @Fraser999 why these two asserts got removed? if they are no longer true then I would prefer to update the rules instead.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -329,9 +322,7 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
     pub fn write(&mut self, key: Key, value: StoredValue) {
         let normalized_key = key.normalize();
         self.cache.insert_write(normalized_key, value.clone());
-        self.ops.insert_add(normalized_key, Op::Write);
-        let transform = Transform::Write(value);
-        self.journal.push((normalized_key, transform));
```
goral09: Is this correct? If we create a `Write` transform then we should also add it to the `ops` collection.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -400,7 +391,6 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
         match transform.clone().apply(current_value) {
             Ok(new_value) => {
                 self.cache.insert_write(normalized_key, new_value);
-                self.ops.insert_add(normalized_key, Op::Add);
```
goral09, line 401: Why is this getting removed?

execution_engine/src/core/tracking_copy/mod.rs
goral09: I see - `ops` collection is removed entirely.

execution_engine/src/core/tracking_copy/tests.rs
luis-casper: The `TrackingCopy` type no longer has the `ops` field.

node/src/components/contract_runtime/announcements.rs
```diff
@@ -78,14 +77,14 @@ impl Display for ContractRuntimeAnnouncement {
...
-    execution_effect: ExecutionEngineExecutionEffect,
+    execution_journal: ExecutionJournal,
 ) where
     REv: From<ContractRuntimeAnnouncement>,
 {
     effect_builder
         .schedule_regular(ContractRuntimeAnnouncement::StepSuccess {
             era_id,
-            execution_effect: (&execution_effect).into(),
+            execution_effect: (&execution_journal).into(),
```
goral09: This go soo confusing now - there are two `ExecutionEffect` types: one in `types/src/execution_result.rs` and another in `execution_engine/src/core/engine_state/execution_effect.rs`. There exists `impl From<ExecutionJournal> for ...` for both of them, for the former one it preserves the order of transforms while for the latter it does not 😵 And the fact that the former one is renamed when importing to `JsonExecutionEffect` adds to this confusion.

goral09: Can we at least use the `From` impl here more explicitly - like `casper_types::ExecutionEffect::from(&execution_journal)`?

execution_engine_testing/tests/src/test/system_contracts/standard_payment.rs
```diff
@@ -535,15 +538,25 @@ fn independent_standard_payments_should_not_write_the_same_keys() {
...
-            (Some(Transform::Write(_)), Some(Transform::Write(_)))
-        )
+        *k != &Key::Balance(payment_purse.addr())
+            && matches!(
+                (
+                    transforms_from_genesis.get(k),
+                    transforms_from_account_1.get(k),
+                ),
+                (Some(Transform::Write(_)), Some(Transform::Write(_)))
+            )
```
goral09, line 559: Wouldn't it be clearer if we took an intersection of both sets (`transforms_from_genesis.keys()` and `transforms_from_account_1.keys()`) and verified that it has a single element that is a write to payment purse balance?

node/src/components/contract_runtime/announcements.rs
luis-casper: Agreed, very confusing.

execution_engine_testing/tests/src/test/system_contracts/standard_payment.rs
luis-casper: Yes, but the keys in a `HashMap` are not necessarily sorted, so taking that intersection would require making a new collection (collect into a `HashSet`) or sorting them (collect into a `Vec`), and at that point I don't know which is clearer.

goral09: To me, definitely `assert_eq!(x.keys().into::<HashSet<_>>() && y.keys.into()::<HashSet<_>>().len(), 1)`.

goral09 (commented):
PR looks good. The only thing that's missing, in my opinion, is a test that verifies the order of entries in the final `ExecutionEffect` structure.

execution_engine_testing/tests/src/test/system_contracts/standard_payment.rs
luis-casper: It turned out not to be so simple, as we must filter keys that correspond to `Write` transforms.
Overall I think this isn't worth pursuing, as it would make it more complicated, but I'm open to PRs.

luis-casper:
Fixed a bug found by Fraser regarding the inclusion of journal changes in error variants.

execution_engine/src/core/execution/executor.rs
```diff
@@ -143,6 +143,10 @@ impl Executor {
         let gas_counter: Gas = Gas::default();
         let transfers = Vec::default();
 
+        // Snapshot of the journal before execution, so in case of error
+        // only nonce update can be returned.
```
goral09: We don't have nonces...

luis-casper: That comment was originally there.
This PR had removed it and I just brought it back.
Should I remove it?

goral09: Yes, let's remove it and replace with something like:
```
so in case of an error we don't apply changes of the execution but keep the payment code effects.
```

luis-casper: Done.

luis-casper:
> PR looks good. The only thing that's missing, in my opinion, is a test that verifies the order of entries in the final `ExecutionEffect` structure.

Done.

smart_contracts/contracts/test/ordered-transforms/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+        )));
+
+    let (contract_hash, _contract_version) =
+        storage::new_locked_contract(entry_points, Some(named_keys.collect()), None, None);
+    runtime::put_key("ordered-transforms-contract-hash", contract_hash.into());
+}
+
+#[no_mangle]
+pub extern "C" fn perform_operations() {
+    let operations: Vec<(u8, u32, i32)> = runtime::get_named_arg("operations");
```
goral09, line 51: What are the expected values here? Can you just add a comment explaining what each of the parameters is? You can simply copy the comment from the `transforms_must_be_ordered.rs` file.

goral09 (approved):

.drone.yml
```diff
@@ -107,6 +107,7 @@ steps:
   image: casperlabs/node-build-u1804
   commands:
   - make setup
+  - make build-contracts
```
Fraser999: This should not be required.  Rather, the new unit test should be decorated with `#[ignore]`, and it will thus not be executed as part of the `make test` call, but will be executed as part of the `make test-contracts` call, where `make test-contracts` already invokes `make build-contracts`.

luis-casper: Aha! Thanks for pointing that out.

Fraser999 (approved):

Fraser999:
bors merge

PR #2177, opened 2021-09-30
Send Shutdown event to event stream

### LINKED ISSUE #2153: Emit an event when node is shutting down
The event stream server should emit a new event `Shutdown` via every endpoint when it is about to shut down.

darthsiroftardis:
CHANGELOG: 

- Added a new `SseData` event called `Shutdown` that will emit an event to every `event-stream-server` endpoint when the node itself shutsdown
- Expanded the current SSE tests to include this new variant

Closes #2153


node/src/components/event_stream_server/sse_server.rs
```diff
@@ -678,14 +675,16 @@ mod tests {
...
         should_not_filter_out(&fault, &MAIN_FILTER[..], getter.clone()).await;
         should_not_filter_out(&step, &MAIN_FILTER[..], getter.clone()).await;
+        should_not_filter_out(&shutdown, &MAIN_FILTER, getter.clone()).await;
 
         should_filter_out(&deploy_accepted, &MAIN_FILTER[..], getter.clone()).await;
         should_filter_out(&finality_signature, &MAIN_FILTER[..], getter.clone()).await;
 
         // `EventFilter::DeployAccepted` should filter out everything except `ApiVersion`s and
-        // `DeployAccepted`s.
+        // `DeployAccepted`s and Shutdown's.
```
rafal-ch:
```
        // `EventFilter::DeployAccepted` should filter out everything except `ApiVersion`s,
        // `DeployAccepted`s and `Shutdown`s.
```

node/src/components/event_stream_server/tests.rs
```diff
@@ -1171,13 +1171,19 @@ async fn should_limit_concurrent_subscribers() {
 /// `resources/test/sse_data_schema.json` across different versions of the codebase.
 #[test]
 fn schema() {
+    // To generate the contents to replace the input JSON files, run the test
+    // and print the `actual_schema by uncommenting the `println!`
```
rafal-ch:
```suggestion
    // and print the `actual_schema` by uncommenting the `println!`
```

rafal-ch (approved):
LGTM with two minor nits.

node/src/components/event_stream_server/tests.rs
darthsiroftardis: Applied suggestion

node/src/components/event_stream_server/sse_server.rs
darthsiroftardis: Applied suggestion

node/src/components/event_stream_server/tests.rs
```diff
@@ -1171,13 +1171,19 @@ async fn should_limit_concurrent_subscribers() {
...
+    // towards the end of the test.
     let schema_path = format!(
         "{}/../resources/test/sse_data_schema.json",
         env!("CARGO_MANIFEST_DIR")
     );
     let expected_schema = fs::read_to_string(schema_path).unwrap();
     let schema = schema_for!(SseData);
     let actual_schema = serde_json::to_string_pretty(&schema).unwrap();
+
+    // println!("{}", actual_schema);
```
nilz3ro, line 1200: Do we need this?

darthsiroftardis: Yes, its for people to uncomment, when they change the schema and have to backfill the schema to account for the change

goral09: Then maybe add a comment explaining that :) I would remove that comment if I had seen it :)

rafal-ch: There is a comment explaining that just above, at the beginning of the test.

goral09: My brain would not go to the beginning of the test in search of explanation why a commented out `println!()` should not be removed.

Fraser999 (changes requested):
I can approve after https://github.com/darthsiroftardis/casper-node/pull/15 is merged to this branch.  This currently has an issue whereby a shutdown event will be given the same ID as the first event emitted after re-starting the server.

Fraser999 (approved):

PR #2178, opened 2021-09-30
Check the immediate caller in system contracts.

mpapierski:
Checks if immediate caller is of session type of the same account as the provided account hash.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3580,6 +3580,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "regression-20210831"
+version = "0.1.0"
+dependencies = [
+ "casper-contract",
+ "casper-types",
+]
+
+[[package]]
 name = "remove-associated-key"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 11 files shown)


darthsiroftardis (approved):

luis-casper (approved):
Looks good to me.
It should be noted that each of the affected auction functions now have a public key parameter that is unneeded: it could be understood that the account controlling the session is what the parameter used to represent.

sacherjj:
> Looks good to me. It should be noted that each of the affected auction functions now have a public key parameter that is unneeded: it could be understood that the account controlling the session is what the parameter used to represent.

Should capture this as a tech debt ticket?

mpapierski:
> 
> 
> > Looks good to me. It should be noted that each of the affected auction functions now have a public key parameter that is unneeded: it could be understood that the account controlling the session is what the parameter used to represent.
> 
> Should capture this as a tech debt ticket?

I don't think we can realistically address this as long as we either use account hashes on EE side, or public keys on the consensus side. Will be ticketed to not forget about this

sacherjj:
bors try

PR #2205, opened 2021-10-08
GH-1902: Fix misleading system contract gas charges

### LINKED ISSUE #1902: Direct auction contract call charges for other system contract calls are misleading.
Calling auction contract's `add_bid` directly with a standard payment with the amount of 2.5CSPR results in a GasLimit error even though `add_bid` costs 2.5CSPR. 

The "hidden cost" here is that the native auction contract calls for "create_purse," which also calls host mint's implementation that charges for `create` entry point. Also, reported gas cost is for 2.5CSPR (add_bid) but we technically charge for 2.5CSPR + 10k motes (add_bid + create)

Workaround: Currently on mainnet and testnet `create` costs 10k motes, then users should be able to call add_bid by calling payment code with 2.5CSPR + 10k motes.

mpapierski:
Closes #1902 

This PR introduces an improved system contract charge method that charges depending on the immediate caller being a system contract or user.

The problem was that `auction::ADD_BID` entry point calls `mint::CREATE_PURSE,` which misled gas cost reporting.

This change also fills the gap where the call stack was inconsistent across the system - for example, in genesis, the installer didn't set it correctly, and the improved charging method wouldn't be reliable. This PR also ensures that invalid (the first frame that's not a session) or empty call stack is a programming error.


execution_engine/src/core/runtime/auction_internal.rs
```diff
@@ -195,7 +195,12 @@ where
         .map_err(|_| Error::CLValue)?;
 
         let gas_counter = self.gas_counter();
-        let call_stack = self.call_stack().clone();
+        let mut call_stack = self.call_stack().clone();
```
luis-casper, line 198: Why do we need to clone call stacks? Wouldn't it be more natural to have the call functions push or pop from them on demand?
This is, of course, out of the scope of this PR.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -2893,6 +2904,22 @@ where
         self.context.get_system_contract(MINT)
     }
 
+    fn get_system_contract_stack_frame(&mut self, name: &str) -> Result<CallStackElement, Error> {
+        let contract_hash = self.context.get_system_contract(name)?;
+        let key = Key::from(contract_hash);
+        let contract = match self.context.read_gs(&key)? {
```
luis-casper, line 2903: We shouldn't have to read from global state to get the hash of system contracts because they are part of the node software.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3639,6 +3666,50 @@ where
         }
         Ok(Ok(()))
     }
+
+    /// Checks if immediate caller is a system contract or account.
+    ///
+    /// For cases where call stack is only the session code, then this method returns `true` if the
+    /// caller is system, or `false` otherwise.
+    fn is_system_immediate_caller(&self) -> Result<bool, Error> {
```
luis-casper, line 3667: I'm concerned with this code for a few reasons:
* First, doesn't `get_immediate_caller` already consult the call stack? Why is there a need to do it twice?
* Second, despite the name of the method being `is_system_immediate_caller`, it looks at the first entry of the stack, not the last, as does `get_immediate_caller`.

execution_engine_testing/tests/src/test/regression/gh_1902.rs
```diff
@@ -0,0 +1,158 @@
...
+static ACCOUNT_1_SECRET_KEY: Lazy<SecretKey> =
+    Lazy::new(|| SecretKey::secp256k1_from_bytes([99; 32]).unwrap());
+static ACCOUNT_1_PUBLIC_KEY: Lazy<PublicKey> =
+    Lazy::new(|| PublicKey::from(&*ACCOUNT_1_SECRET_KEY));
+static ACCOUNT_1_ADDR: Lazy<AccountHash> = Lazy::new(|| AccountHash::from(&*ACCOUNT_1_PUBLIC_KEY));
+
+fn setup() -> InMemoryWasmTestBuilder {
+    let mut builder = InMemoryWasmTestBuilder::default();
+    builder.run_genesis(&*DEFAULT_RUN_GENESIS_REQUEST);
+    let id: Option<u64> = None;
```
luis-casper, line 34: Why not just put `Option::<u64>::None` below directly?

luis-casper, line 36: Why not just put `Option::<u64>::None` below directly?

luis-casper (changes requested):

execution_engine/src/core/runtime/mod.rs
darthsiroftardis: @luis-casper We write the system contract hashes to global state, and the node retrieves it by reading it from there. Do you have an example where the node has an alternate source?

luis-casper: We have `SystemContractRegistry`, for instance.

execution_engine/src/core/runtime/auction_internal.rs
mpapierski: This is because `call_host_mint` assumes the call stack is already updated. We do push to call stack before creating a new Runtime object. I.e. in `Executor::exec` it already runs with an initial call stack (session + variant for stored contracts). In `Runtime::execute_contract` we're pushing a new element before `call_host_mint` is called. Auction internals are only outlier and without this change implemented checks wouldn't work consistently

execution_engine/src/core/runtime/mod.rs
mpapierski: System contract hashes are, in fact, part of node software, but we need to be able to query them off-chain, so we've moved them into a global state. (previously, we've had a separate database for "ProtocolData," which was complicating the design of fast sync

execution_engine/src/core/runtime/mod.rs
mpapierski: > First, doesn't get_immediate_caller already consult the call stack? Why is there a need to do it twice?

Get immediate caller checks only the one element before the end of the stack. This method verifies that the caller is a system one. 

> Second, despite the name of the method being is_system_immediate_caller, it looks at the first entry of the stack, not the last, as does get_immediate_caller.

Fair point on naming. I re-checked the necessity to consult the first element in case of a lack of an immediate caller. The immediate caller is assumed to be present when `call_host_*` methods are calling this, and there's no further check needed for the first entry. Addressed in 7d5b9a79d59498d9b59f3853f8afa10b7ec3ec75

execution_engine_testing/tests/src/test/regression/gh_1902.rs
mpapierski: Consistency - all of the testing code uses this pattern, so I went with it.

execution_engine/src/core/runtime/mod.rs
luis-casper: Sure, but this code is part of the node software, so we could avoid that read, right?

luis-casper (approved):

execution_engine/src/core/runtime/mod.rs
luis-casper: So the bottom line is that this isn't really possible since we need the package hash also.

luis-casper (approved):

execution_engine/src/core/runtime/mod.rs
```diff
@@ -987,6 +987,17 @@ where
...
+        // programming errors.
+        debug_assert!(
+            !call_stack.is_empty(),
+            "Call stack should not be empty while creating a new Runtime instance"
+        );
+        debug_assert!(
+            call_stack.first().unwrap().contract_hash().is_none(),
+            "First element should always represent a Session call"
+        );
+
```
Fraser999: We should log errors here for each of these `debug_assert!` cases.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -1032,6 +1043,9 @@ where
     where
         T: Into<Gas>,
     {
+        if self.is_system_immediate_caller()? {
```
Fraser999, line 1048: Please update doc comment.

execution_engine_testing/tests/src/test/regression/gh_1902.rs
```diff
@@ -0,0 +1,158 @@
...
+use casper_types::{
+    account::Account,
+    runtime_args,
+    system::{
+        auction::{self, DelegationRate},
+        mint, standard_payment,
+    },
+    Gas, PublicKey, RuntimeArgs, SecretKey, U512,
+};
+use once_cell::sync::Lazy;
```
Fraser999: Should be in its own group.

execution_engine_testing/tests/src/test/regression/gh_1902.rs
```diff
@@ -0,0 +1,158 @@
...
+    let transaction_fee = builder.get_proposer_purse_balance() - proposer_reward_starting_balance;
+    assert_eq!(transaction_fee, expected_payment_charge);
+
+    let expected = balance_before - expected_tokens_paid - transaction_fee;
+
+    assert_eq!(
+        balance_after,
+        expected,
+        "before and after should match; off by: {}",
+        expected - balance_after
```
Fraser999: Could overflow?

Fraser999 (approved):
Approved pending nitpicks being handled.

PR #2211, opened 2021-10-14
fix chainspec loader test and add a further test

### LINKED ISSUE #2210: Failing test in chainspec loader
[This PR](https://github.com/casper-network/casper-node/pull/2101/files) changed the behaviour of the chainspec loader such that the `should_keep_running_if_unplanned_shutdown_and_future_upgrade_scheduled` test fails if the randomly-generated block happens to be a the last switch block before the next scheduled upgrade.

Before the linked PR, such a case would cause the chainspec loader to indicate that the node should keep running.  After the PR, the chainspec loader indicates the node should stop for an upgrade.

Fraser999:
This PR fixes a chainspec loader test which should have been updated as part of #2101.

It also adds a further test case which explicitly exercises the new code path introduced in that same PR and which was causing the other test to spuriously fail.

Closes #2210.

### DIFF
--- a/node/src/components/chainspec_loader.rs
+++ b/node/src/components/chainspec_loader.rs
@@ -1016,14 +1016,12 @@ mod tests {
         assert!(maybe_next_point(&current).is_none());
     }
 
-    #[cfg(test)]
     struct TestFixture {
         chainspec_loader: ChainspecLoader,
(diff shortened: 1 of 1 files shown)


darthsiroftardis (approved):

Fraser999:
bors merge

Fraser999:
bors merge

PR #2217, opened 2021-10-15
Purge deploys using latest block timestamp

### LINKED ISSUE #2155: Purge expired deploys from block proposer using latest block timestamp
The `BlockProposerDeploySets::prune` takes a current timestamp as means of deciding which deploys have expired.  Rather than using `Timestamp::now()` to provide this value, we should use the `timestamp` from the latest finalized block.

darthsiroftardis:
CHANGELOG:

- Changed the pruning functionality in the `block_proposer` to use the timestamp of the latest finalized block as seen by the block proposer

Closes #2155


node/src/components/block_proposer.rs
```diff
@@ -269,7 +274,7 @@ impl BlockProposerReady {
                     .event(|_| Event::Prune);
 
                 // Announce pruned hashes
-                let pruned_hashes = self.prune(Timestamp::now());
+                let pruned_hashes = self.prune(self.last_finalized_timestamp);
```
afck: Should we get rid of the `PRUNE_INTERVAL` and `Event::Prune` now, and just prune whenever `last_finalized_timestamp` changes?

afck (commented):
We should add checks to both call sites to make sure we only `propose_block_payload` if `request.context.timestamp() >= self.last_finalized_timestamp`, and otherwise `warn!`.
It's probably only relevant if our clock is wildly out of sync, but it's better to be safe and make sure we never propose an invalid block.

rafal-ch (approved):
LGTM

node/src/components/block_proposer.rs
darthsiroftardis: Removed `Event::Prune`

goral09 (approved):

node/src/components/block_proposer.rs
```diff
@@ -261,38 +249,33 @@ impl BlockProposerReady {
                             request.random_bit,
                         ))
                         .ignore()
+                } else {
+                    Effects::new()
```
afck, line 253: We should log a warning here. It probably means that a block request spent so much time in the reactor queue (or in the block proposer queue?) that in the meantime we finalized a block with a timestamp newer than this.

node/src/components/block_proposer.rs
```diff
@@ -261,38 +249,33 @@ impl BlockProposerReady {
...
                 Effects::new()
             }
             Event::FinalizedBlock(block) => {
+                let mut effects = Effects::new();
                 // We have received a finalized block, set the timestamp to
                 // allow for pruning.
                 self.last_finalized_timestamp = block.timestamp();
+                // Announce pruned hashes
+                let pruned_hashes = self.prune(self.last_finalized_timestamp);
+                let pruned_count = pruned_hashes.total_pruned;
```
afck, line 272: Do we know how fast pruning is? I wonder whether we should have a minimum delay, e.g. something like: if we already pruned within the last minute, don't prune again? But maybe it's not needed.

node/src/components/block_proposer.rs
darthsiroftardis: @afck Do you have an idea of what the log message should be like,

"block request timestamp is behind the timestamp of the last finalized block"?

afck: Yes, sounds good. Maybe add something like: "not proposing a block; relevant deploys might have been purged"

Fraser999 (changes requested):
As discussed, this PR should include tests for this change.

goral09:
@Fraser999 @darthsiroftardis , Do we plan on including this PR with 1.5.0 release?

PR #2218, opened 2021-10-18
Non-blocking `execute_finalized_block`

### LINKED ISSUE #2216: Offload execute_finalized_block to blocking task
In the contract runtime, there is a [specific site](https://github.com/casper-network/casper-node/blob/7d6100b334042822ff6098e6fa800230fd7ae626/node/src/components/contract_runtime.rs#L415-L428) running blocking code:

```rust
tokio::task::unconstrained(async move {
    let result = execute_finalized_block(
        engine_state.as_ref(),
        Some(metrics),
        protocol_version,
        execution_pre_state,
        finalized_block,
        deploys,
        transfers,
    );
    trace!(?result, "execute block response");
    responder.respond(result).await
})
.ignore()
```

This should be examined for two reasons:

1. It is _possible_ that `unconstrained` isn't doing anything here (see [example](https://docs.rs/tokio/1.12.0/tokio/task/index.html#unconstrained)), as there is only one `await` and it is right at the end of the execution of the block.

2. Furthermore, the work of `execute_finalized_block` is probably CPU heavy and will be done on a task reserved for `async` work. Blocking these is generally not a good idea, since there may be only one or two in total. We currently avoid any issues since it is rare if not impossible for more than one execution to run in parallel.

At the very least, something like `tokio::spawn_blocking` should be used. A better approach is adding a method to the effect builder (e.g. called `run_blocking`) that takes a closure (not a future!) and offers a unified API to deal with these things. This could also be extended with some preliminary throttling, e.g. by adding a semaphore that limits the parallelism of these spawned tasks.

luis-casper:
Implements function `run_intensive_task` that leverages a semaphore to run at most a certain number of resource intensive tasks asynchronously in `tokio`'s thread pool.

Uses this function to offload the execution of `execute_finalized_block` from the component reactor.

Closes #2216.

### DIFF
--- a/node/src/components/contract_runtime.rs
+++ b/node/src/components/contract_runtime.rs
@@ -52,6 +52,8 @@ pub use types::BlockAndExecutionEffects;
 pub(crate) use types::EraValidatorsRequest;
 
 /// Maximum number of resource intensive tasks that can be run in parallel.
+///
+/// TODO: Fine tune this constant to the machine executing the node.
 const MAX_PARALLEL_INTENSIVE_TASKS: usize = 4;
 
 /// Semaphore enforcing maximum number of parallel resource intensive tasks.
--- a/node/src/components/contract_runtime.rs
+++ b/node/src/components/contract_runtime.rs
@@ -439,7 +439,7 @@ where
                 );
                 let engine_state = Arc::clone(&self.engine_state);
                 let metrics = Arc::clone(&self.metrics);
-                tokio::task::spawn(async move {
+                async move {
                     let result = run_intensive_task(move || {
                         execute_finalized_block(
                             engine_state.as_ref(),
@@ -454,7 +454,7 @@ where
                     .await;
                     trace!(?result, "execute block response");
                     responder.respond(result).await
-                })
+                }
                 .ignore()
             }
             ContractRuntimeRequest::EnqueueBlockForExecution {
--- a/node/src/components/contract_runtime.rs
+++ b/node/src/components/contract_runtime.rs
@@ -5,6 +5,7 @@ mod error;
 mod operations;
 mod types;
 
+use once_cell::sync::Lazy;
 use std::{
     collections::BTreeMap,
     fmt::{self, Debug, Formatter},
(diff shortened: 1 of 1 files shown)


luis-casper:
I would like to note that, if we were to write `run_intensive_task` as a macro, we might be able to save the spawning of one asynchronous context and it might make the call sites more readable (at the expense of the definition).

node/src/components/contract_runtime.rs
```diff
@@ -412,16 +439,19 @@ where
...
-                    let result = execute_finalized_block(
-                        engine_state.as_ref(),
-                        Some(metrics),
-                        protocol_version,
-                        execution_pre_state,
-                        finalized_block,
-                        deploys,
-                        transfers,
-                    );
+                tokio::task::spawn(async move {
```
marc-casperlabs: I believe we can get rid of this `spawn`, the `.ignore` here is currently called on the `JoinHandle`, as far as I can tell, so just call it on the `async` block directly?

```rust
                async move {
                    let result = run_intensive_task(move || {
                        execute_finalized_block(
                            engine_state.as_ref(),
                            Some(metrics),
                            protocol_version,
                            execution_pre_state,
                            finalized_block,
                            deploys,
                            transfers,
                        )
                    })
                    .await;
                    trace!(?result, "execute block response");
                    responder.respond(result).await
                }
                .ignore()
```

should just work?

marc-casperlabs (approved):
Looks good to me, although I believe we can just get rid of the `spawn` anyway?

> I would like to note that, if we were to write run_intensive_task as a macro, we might be able to save the spawning of one asynchronous context and it might make the call sites more readable (at the expense of the definition).

I am not sure how a macro factors in here, do you have an example? Or was this just about getting rid of the `tokio::spawn`?

node/src/components/contract_runtime.rs
```diff
@@ -50,6 +51,32 @@ pub use operations::execute_finalized_block;
...
+/// Semaphore enforcing maximum number of parallel resource intensive tasks.
+static INTENSIVE_TASKS_SEMAPHORE: Lazy<tokio::sync::Semaphore> =
+    Lazy::new(|| tokio::sync::Semaphore::new(MAX_PARALLEL_INTENSIVE_TASKS));
+
+/// Asynchronously runs a resource intensive task.
+/// At most `MAX_PARALLEL_INTENSIVE_TASKS` are being run in parallel at any time.
+///
+/// The task is a closure that takes no arguments and returns a value.
+/// This function returns a future for that value.
+async fn run_intensive_task<T, V>(task: T) -> V
```
marc-casperlabs, line 68: As a note for later, we probably want to move this out of the contract runtime into a more general place, in case other components (like consensus) want to use it.

luis-casper: Agreed.

node/src/components/contract_runtime.rs
luis-casper: You are right. It's fixed.

Fraser999 (approved):
Approved, but we should follow up as @marc-casperlabs suggested by making this available to any component.  We should also make `MAX_PARALLEL_INTENSIVE_TASKS` a config value.

Fraser999:
bors merge

PR #2222, opened 2021-10-18
Only upgrade after making sure the switch block was stored.

### LINKED ISSUE #2209: 1.3.2>13.4>1.4.0 Upgrade failure when era duration short
Issue: When era duration is 15 mins and upgrade point era 3, upgrade failure .
Path - 1.3.2> 13.4 and 1.3.2>1.3.4>1.4.0

"// This is an invalid run where blocks are missing from storage.  Try exiting the
// process and downgrading the version to recover the missing blocks.
//
// TODO - if migrating data yields a new empty block as a means to store the
//        post-migration global state hash, we'll come to this code branch, and we
//        should not exit the process in that case."

https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1633999365206100
https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1634116042220500?thread_ts=1633999365.206100&cid=C01ULKL8G8J

afck:
The era supervisor stops voting in an era once the switch block has remained finalized without new equivocations occurring for a while, to make sure all other nodes can verify its finality in the consensus protocol. Once it stops voting it also checks whether it was the last era before an upgrade, and if that is the case, it shuts down the node so that the next software version can take over.

In extreme cases where block execution is delayed by a lot this could potentially happen before the block was executed and stored.

This fix moves the trigger for era deactivation to `handle_block_added`, which is only called after the block has been stored.

Closes #2209.

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -809,21 +809,37 @@ where
             trace!(era = era_id.value(), "executed block in old era");
             return effects;
         }
-        if block_header.is_switch_block() && !self.should_upgrade_after(&era_id) {
-            // if the block is a switch block, we have to get the validators for the new era and
-            // create it, before we can say we handled the block
-            let new_era_id = era_id.successor();
-            let effect = get_booking_block_hash(
-                self.effect_builder,
-                new_era_id,
-                self.era_supervisor.protocol_config.auction_delay,
-                self.era_supervisor.protocol_config.last_activation_point,
-            )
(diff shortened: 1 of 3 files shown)


goral09 (approved):

afck:
I added a test; without the fix the test fails as expected:
`thread 'reactor::participating::tests::dont_upgrade_without_switch_block' panicked at 'missing switch block', node/src/reactor/participating/tests.rs:477:14`

Fraser999 (approved):

afck:
bors merge

PR #2239, opened 2021-10-20
Only upgrade after making sure the switch block was stored. (1.3.4 backport)

afck:
Backport of https://github.com/casper-network/casper-node/pull/2222 (ticket https://github.com/casper-network/casper-node/issues/2209) to version 1.3.4.

The test was adapted because the contract runtime event structure had changed since 1.3.4. Also, the commit with the `FilterReactor` from https://github.com/casper-network/casper-node/pull/1921 was required for the test.

https://github.com/casper-network/casper-node/pull/2219 and https://github.com/casper-network/casper-node/pull/2229 were included, too, otherwise CI would fail on cargo audit.

### DIFF
--- a/node/src/reactor/participating/tests.rs
+++ b/node/src/reactor/participating/tests.rs
@@ -1,11 +1,13 @@
 use std::{collections::BTreeMap, sync::Arc, time::Duration};
 
 use anyhow::bail;
+use either::Either;
 use log::info;
 use num::Zero;
 use num_rational::Ratio;
 use rand::Rng;
 use tempfile::TempDir;
+use tokio::time;
 
 use casper_execution_engine::shared::motes::Motes;
 use casper_types::{system::auction::DelegationRate, EraId, PublicKey, SecretKey, U512};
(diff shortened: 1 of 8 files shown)


afck:
I also verified for the backport that without the fix the test fails as expected:
`thread 'reactor::participating::tests::dont_upgrade_without_switch_block' panicked at 'missing switch block', node/src/reactor/participating/tests.rs:423:14`

Fraser999 (approved):

goral09 (approved):

node/src/reactor/participating/tests.rs
```diff
@@ -332,3 +338,95 @@ async fn run_equivocator_network() {
...
+            .await;
+        let mut exec_request_received = false;
+        runner.reactor_mut().inner_mut().set_filter(move |event| {
+            if let ParticipatingEvent::ContractRuntime(contract_runtime::Event::Request(request)) =
+                &event
+            {
+                if let ContractRuntimeRequest::ExecuteBlock(finalized_block) = request.as_ref() {
+                    if finalized_block.era_report().is_some()
+                        && finalized_block.era_id() == EraId::from(1)
+                        && !exec_request_received
```
goral09, line 387: Do we need this flag at all? If the test is completed at the switch block in era-1, then there won't be any more `ExecuteBlock` requests that we would have to filter on.

afck: No, this is so that when the delay has passed and that _same_ event comes in again we don't delay it again (indefinitely).

afck:
bors merge

PR #2243, opened 2021-10-20
**Deprecated** Regression test for the size and gas cost of the `host_function_metrics` contract (Issue #2049)

### LINKED ISSUE #2049: Gas Cost and Deploy Size Regression Tests For Smart Contracts
Currently, when modifying the `casper-types` crate it's possible to increase gas costs and deploy sizes for smart contracts.

We need to instrument this so that we do not introduce regressions.

luis-casper:
Closes #2049

There are only two issues here:
* These tests cannot possibly be a comprehensive benchmark for other contracts; and
* For some reason, the second test outputs garbage to the terminal (if it fails or `cargo test` has been passed `-- --show-output`).

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -809,21 +809,37 @@ where
             trace!(era = era_id.value(), "executed block in old era");
             return effects;
         }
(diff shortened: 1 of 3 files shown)


luis-casper:
I've made a mistake on this PR trying to get Drone to work and it won't reopen.
I will make a duplicate soon, so this one is closed forever.

PR #2244, opened 2021-10-20
Regression test for the size and gas cost of the `host_function_metrics` contract (Issue #2049)

### LINKED ISSUE #2049: Gas Cost and Deploy Size Regression Tests For Smart Contracts

luis-casper:
Closes #2049

There are only two issues here:
* These tests cannot possibly be a comprehensive benchmark for other contracts; and
* For some reason, the second test outputs garbage to the terminal (if it fails or `cargo test` has been passed `-- --show-output`).


xcthulhu:
The problem @dwerner and I were have previously is that refactoring crates creates regressions in gas costs in the various test contracts.  I have the same concern regarding the hashing-crate refactor.

I'm not sure how this solves that issue, but this is desirable anyway...

goral09:
Maybe, instead of trying to instrument _contracts_, we should instrument all host API/FFI endpoints and primitive types for which we have `ToBytes`/`FromBytes` implementations. If cost and/or size of these change, then any more complex type (that builds on these) or smart contract (that uses the types/calls the FFI) will have its cost/size changed.

luis-casper:
Why not, resources permitting, instrument both?

goral09:
Sure but, like you noted in your comment, it cannot possibly cover all (external) contracts and, if not paired with instrumenting primitive types and FFI, may result in false positives (no regression would be detected b/c our internal contracts cover only a portion of the FFI/types).

xcthulhu:
> Maybe, instead of trying to instrument contracts, we should instrument all host API/FFI endpoints and primitive types for which we have ToBytes/FromBytes implementations. If cost and/or size of these change, then any more complex type (that builds on these) or smart contract (that uses the types/calls the FFI) will have its cost/size changed.

Sure, we should probably instrument all of the host API/FFI endpoints.  But we could do both because they complement one another.

To be specific, the regression @dwerner and I saw was when we moved out `bytesrepr` into its own crate, more gas was being consumed in the payment phase for one of the contracts (`get-call-stack-call-recursive-subcall` I think). This can be explained that the linker may not optimize across crates as well as it does inside a crate. So I had imagined we would instrument all of the contracts.  But the sizes of our data structures didn't change, and I don't think our wasm code was much bigger either.

Since this is @luis-casper 's ticket, I'll ask him: are you confident that `host_function_metrics.wasm` has enough coverage so we don't need to instrument the other smart contracts?

luis-casper:
> Sure but, like you noted in your comment, it cannot possibly cover all (external) contracts and, if not paired with instrumenting primitive types and FFI, may result with false positives (no regression would be detected b/c our internal contracts cover only a portion of the FFI/types).

Right, but the primitive types and FFI aren't the only culprits of contract performance regressions. Non-deterministic compilations, change of the toolchain and changes in the contracts themselves all could result in worsened performance.
I think your idea is awesome and should definitely be implemented, but I also see value in more extensive contract performance tests.

luis-casper:
> Since this is @luis-casper 's ticket, I'll ask him: are you confident that `host_function_metrics.wasm` has enough coverage so we don't need to instrument the other smart contracts?

I don't think so, as I mentioned in my comment. I think capturing contract interaction data, say from mainnet, and replaying those locally onto the new contracts would be fairly comprehensive.

smart_contracts/contracts/profiling/host-function-metrics/src/lib.rs
```diff
@@ -148,14 +149,13 @@ pub extern "C" fn call() {
...
     // Store large function with 10 named keys, then execute it.
-    runtime::call_contract::<NamedKeys>(contract_hash, entry_point_name, runtime_args);
+    runtime::call_contract::<NamedKeys>(contract_hash, entry_point_name, runtime_args.clone());
 
     // Small function
     let small_function_name =
         "s".repeat(rng.gen_range(MIN_FUNCTION_NAME_LENGTH..=MAX_FUNCTION_NAME_LENGTH));
 
     let entry_point_name = &small_function_name;
-    let runtime_args = runtime_args! {};
```
Fraser999, line 158: Was this change required?

execution_engine_testing/tests/src/test/regression/host_function_metrics_size_and_gas_cost.rs
```diff
@@ -0,0 +1,110 @@
...
+
+const HOST_FUNCTION_METRICS_MIN_SIZE: usize = HOST_FUNCTION_METRICS_STANDARD_SIZE * 95 / 100;
+const HOST_FUNCTION_METRICS_MAX_SIZE: usize = HOST_FUNCTION_METRICS_STANDARD_SIZE * 105 / 100;
+const HOST_FUNCTION_METRICS_MIN_GAS_COST: u64 = HOST_FUNCTION_METRICS_STANDARD_GAS_COST * 95 / 100;
+const HOST_FUNCTION_METRICS_MAX_GAS_COST: u64 = HOST_FUNCTION_METRICS_STANDARD_GAS_COST * 105 / 100;
+
+const ACCOUNT0_ADDR: AccountHash = AccountHash::new([42; ACCOUNT_HASH_LENGTH]);
+const ACCOUNT1_ADDR: AccountHash = AccountHash::new([43; ACCOUNT_HASH_LENGTH]);
+
+#[test]
```
Fraser999, line 51: Should have the `#[ignore]` attribute applied as per the other EE tests in this crate.  Same for the test below.

smart_contracts/contracts/profiling/host-function-metrics/src/lib.rs
luis-casper: Yes, otherwise I get the error `0: Exec(MissingArgument { name: "bytes" })` from a subsequent contract call.

Fraser999 (approved):

piotr-dziubecki:
@mpapierski could you take a look?

mpapierski (changes requested):
Needs more discussion to find possibly better approach

mpapierski (approved):
After discussing this issue it's certainly better than having nothing with a possible improvement to it being ticketed

PR #2245, opened 2021-10-20
Matching exp 15 speed block counts with current MainNet.

sacherjj:
Correcting wasm and native transfer limits to match existing at exp 15 block speed.
Adding approx mainnet activation point.


TomVasile (approved):

sacherjj:
bors  r+

PR #2247, opened 2021-10-21
GH-1444: Fix perf tools

### LINKED ISSUE #1444: system::transfer_to_account fails with a surprisingly low transfer amount
The largest U512 that system::transfer_to_account takes is 99_998_500_000_000_000 motes, or 99,998,500 CSPR. One mote more and it reverts.

To reproduce, follow instructions for running host-function-metrics here : https://github.com/casper-network/casper-node/tree/master/execution_engine_testing/tests/src/profiling

Note that the failing binary is the state-initializer, while the offending call is made in smart_contracts/contracts/profiling/state-initializer/src/main.rs

mpapierski:
Closes #1444

This PR closes the original bug report. The bug wasn't related to the system::transfer API, but the state-initializer contract. This contract was using `DEFAULT_ACCOUNT_INITIAL_BALANCE - X` where X was chosen arbitrarily. For creating new accounts, we have a special constant `MINIMUM_ACCOUNT_CREATION_BALANCE` in the testing framework that is guaranteed to be correct at the framework level.

Mint system contract or EE does not look at the amounts or depend on some maximum.

After changing the contract, we can rerun the profiling tools:

```
cargo run --bin state-initializer | cargo run --bin simple-transfer
```


Fraser999 (approved):

dwerner (approved):
Just need to satisfy clippy.

PR #2249, opened 2021-10-21
Do not crash node when AuctionState query can't be answered.

goral09:


darthsiroftardis (approved):

node/src/components/rpc_server/rpcs/state.rs
```diff
@@ -437,10 +437,24 @@ impl RpcWithOptionalParamsExt for GetAuctionInfo {
...
-            } else {
-                None
+            let bids = match get_bids_result {
+                Ok(get_bids_result) => match get_bids_result {
+                    GetBidsResult::RootNotFound => {
+                        warn!(?state_root_hash, "failed to get bids at root hash");
+                        return Ok(response_builder.error(warp_json_rpc::Error::custom(
+                            ErrorCode::QueryFailed as i64,
+                            "get-auction-info failed to get bids".to_string(),
+                        ))?);
```
Fraser999: I think this should be an error level message since it doesn't seem like the node could return anything but success here and still be in a valid operational state.  In the same way, I'd be inclined to return an RPC error of `-32603` ([internal server error](https://www.jsonrpc.org/specification#error_object)).

While the `warp_json_rpc` crate does provide `Error::INTERNAL_ERROR` for this, it doesn't make it easy to construct such an error with a custom message.  I'd maybe just add a new variant to our own `ErrorCode` enum like `InternalError = 32603` and use it via `warp_json_rpc::Error::custom()`.

Also, it might be worth including the `state_root_hash` in the error message of the response too, since we might not have access to the node where this triggers?

node/src/components/rpc_server/rpcs/state.rs
```diff
@@ -437,10 +437,24 @@ impl RpcWithOptionalParamsExt for GetAuctionInfo {
...
+                    }
+                    GetBidsResult::Success { bids } => bids,
+                },
+                Err(err) => {
+                    warn!(?err, "failed to get bids");
```
Fraser999: Same points as above, including providing the `state_root_hash` in our log error message and in the error message in the RPC response (the response should include `err` too probably?)

node/src/components/rpc_server/rpcs/state.rs
```diff
@@ -454,10 +468,19 @@ impl RpcWithOptionalParamsExt for GetAuctionInfo {
                 )
                 .await;
 
-            let era_validators = era_validators_result.ok();
+            let era_validators = match era_validators_result {
+                Ok(validators) => validators,
+                Err(err) => {
+                    warn!(?err, "failed to get era validators");
```
Fraser999: And again.

node/src/components/rpc_server/rpcs/state.rs
goral09: re using `ERROR` rather than `WARN` level - I thought we are supposed to use error only when the node is in unoperable/unrecoverable/incorrect state. It's not the case here. It might be an intermittent problem only.

I considered using `warp_json_rps::Error::INTERNAL_ERROR` initially but didn't for the reasons you mention - that it doesn't allow for providing a custom message. I am not convinced whether we should create our own variant for it with custom messages. Does `QueryFailed` not fit here?

Fraser999: I don't really think so.  It seemed like I misunderstood how we hit this unwrap, and it may indeed be some kind of startup timing issue.  But ultimately, the client is making the request by providing a block hash or height.  If they provide an invalid one, then they correctly get a `ErrorCode::NoSuchBlock` response.  But after that, the node should definitely get the state root hash from the requested stored block, and at that root hash, we should definitely be able to access the bids and validator set.  So, this seems like it's an internal server error if we fail in this way, and I'd say it does represent the node being in an invalid state.

It'd be interesting to see if we ever hit this point again and don't panic here, whether the node actually _can_ continue to run.

node/src/components/rpc_server/rpcs/state.rs
goral09: Resolved in [`84e7611` (#2249)](https://github.com/casper-network/casper-node/pull/2249/commits/84e76119a243b2379706ab323e8a1e886070ca96)

node/src/components/rpc_server/rpcs/state.rs
goral09: Resolved in [`84e7611` (#2249)](https://github.com/casper-network/casper-node/pull/2249/commits/84e76119a243b2379706ab323e8a1e886070ca96)

node/src/components/rpc_server/rpcs/state.rs
goral09: Resolved.

Fraser999 (approved):

goral09:
bors merge

PR #2250, opened 2021-10-21
Do not crash node when AuctionState query can't be answered. (Backport to release-1.4.1)

darthsiroftardis:
CHANGELOG:

- Backport of [PR](https://github.com/casper-network/casper-node/pull/2249)


Fraser999 (changes requested):
I made a couple of comments on the original PR, so I'll block this until we see how that shakes out.

Fraser999 (approved):

PR #2252, opened 2021-10-22
Have only `casper-hashing` depend on `blake2b`

### LINKED ISSUE #2082: Get rid of `blake2b` dependencies everywhere except for the `casper-hashing` crate
Currently `AccountHash` requires a call to `blake2b` in its constructor in `casper-types`.

There should be one data path for computing hashes.

luis-casper:
Closes #2082


afck:
I noticed there are still [variables `blake_hash` and `blake_hashes` in `components/storage.rs`](https://github.com/casper-network/casper-node/blob/450fe20ff51324bb2051e05ee60f750c715f2290/node/src/components/storage.rs#L1141). Maybe this would be the right PR to remove the `blake_`.

luis-casper:
> I noticed there are still [variables `blake_hash` and `blake_hashes` in `components/storage.rs`](https://github.com/casper-network/casper-node/blob/450fe20ff51324bb2051e05ee60f750c715f2290/node/src/components/storage.rs#L1141). Maybe this would be the right PR to remove the `blake_`.

I tend to agree.

xcthulhu:
So... there's a few snags with this PR.

I had mentioned to @luis-casper that we'd ideally want to get rid of `casper_types::blake2b()`.  The problem with making `casper-hashing` an upstream dependency and doing this is that it probably introduces a regression.  I'm not sure that https://github.com/casper-network/casper-node/pull/2244 will catch one here.

I looked into this myself. While I can get rid of `casper_types::blake2b()` with some elbow grease, we also have `casper_types::key::Key::dictionary()` that uses a blake2b hash. It's taking a hash of a pair there.

So I suggest:
   - Don't add `casper-hashing` as a dependency to `types`. We're already seeing creep in smart contract sizes and we want to optimize this crate.
   - Remove `casper_types::blake2b()` and either use the smart-contract-runtime's `runtime::blake2b` or and `|x| Digest::hash(x).value()` elsewhere.
   - Leave `casper_types::key::Key::dictionary()` alone (so not every data path goes through `Digest`, just all-but-one).

luis-casper:
> * Don't add `casper-hashing` as a dependency to `types`. We're already seeing creep in smart contract sizes and we want to optimize this crate.

Agreed.

> * Remove `casper_types::blake2b()` and either use the smart-contract-runtime's `runtime::blake2b` or and `|x| Digest::hash(x).value()` elsewhere.

The `runtime::blake2b` function is part of the contract api, but the code in question is native code.
Also, using `Digest::hash` would make `casper-types` depend on `casper-hashing`.

> * Leave `casper_types::key::Key::dictionary()` alone (so not every data path goes through `Digest`, just all-but-one).

That isn't the only place left, by far.

xcthulhu:
So having chatted with @luis-casper about this, I'm going to recommend that we drop this and label **Won't Fix**. 

If we go the way this PR is currently, there's a high likelihood that introducing `casper-hashing` upstream will cause a regression if we replace `casper_types::blake2b()`.

We could try to have `runtime::blake2b()` be the only way a smart contract can construct a hash. Then we could remove `blake2b` from the `types` crate altogether. But that would make breaking changes to the system contracts.

goral09:
We should measure the impact of the regression first. If it's minimal then maybe we can consider doing it anyway.

luis-casper:
> We should measure the impact of the regression first. If it's minimal then maybe we can consider doing it anyway.

I don't see any viable way to implement this, that's the problem.

piotr-dziubecki:
@Fraser999 -> ticket closed as won't do - we are closing the PR as well.

PR #2263, opened 2021-10-22
Fix feature set on casper-contract

### LINKED ISSUE #2261: Fix broken features on casper-contract
The changes to the Rust features in `casper-contract` applied in #2096 broke backwards compatibility for some smart contracts.

We need to fix this and add a regression test.

Fraser999:
This PR fixes an accidental breaking change in the Rust features of the `casper-contract` crate, introduced in the 1.4.0 version.

The previous 1.3.x behavior has been reinstated, whereby if the `std` feature is enabled, `casper-contract` does not provide the `wee-alloc` allocator, and instead explicitly imports the std library with its allocator.

There is also a regression test added in the form of a test contract which enables the `std` feature of `casper-contract`.

Furthermore these changes were tested against the [Payment Contract Example](https://github.com/casper-ecosystem/payment-example) and [Casper Keys Manager](https://github.com/casper-ecosystem/keys-manager) with their recent changes ([#3](https://github.com/casper-ecosystem/payment-example/pull/3/files) and [#18](https://github.com/casper-ecosystem/keys-manager/pull/18/files) respectively) reverted.

Closes #2261.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4074,6 +4074,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
+name = "std-feature-regression"
+version = "0.1.0"
+dependencies = [
+ "casper-contract",
+]
+
+[[package]]
 name = "storage-costs"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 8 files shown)


darthsiroftardis (approved):

luis-casper (approved):

nilz3ro (approved):

Fraser999:
bors merge

Issue #1945, opened 2021-08-19
Commit Multiple Deploys At Once


piotr-dziubecki:

dwerner:
PR is up here: https://github.com/casper-network/casper-node/pull/2277

PR #2277, opened 2021-10-27
GH-1945 - Cache GlobalState changes during execution of a single block

dwerner:
This PR introduces changes to the `ContractRuntime` which allows many deploys to be executed by a temporary cache over `EngineState<_>`. The stored values that were read and modified are then written back in bulk to LMDB.

In my tests this results in blocks executed in approximately 2/3 wall clock time when compared to the old code, for a speedup of approximately 30%.

```
 ~/D/casper-node (many-deploys-per-commit-3)> critcmp multiple-requests one-request
group                                                                                              multiple-requests                      one-request
-----                                                                                              -----------------                      -----------
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1/purses:1         1.00    221.8±2.54µs 10.7 MElem/sec    1.03    229.2±9.64µs 10.4 MElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1/purses:100       1.03    231.1±1.94µs 10.3 MElem/sec    1.00    224.8±2.67µs 10.6 MElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1/purses:20        1.00    218.9±0.75µs 10.9 MElem/sec    1.04    227.1±3.86µs 10.5 MElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1/purses:50        1.00    228.7±5.66µs 10.4 MElem/sec    1.00    228.4±1.80µs 10.4 MElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1200/purses:1      1.47    180.1±1.11ms 13.6 KElem/sec    1.00    122.6±2.90ms 19.9 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1200/purses:100    1.41    176.9±0.68ms 13.8 KElem/sec    1.00    125.1±3.24ms 19.5 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1200/purses:20     1.51    183.8±1.11ms 13.3 KElem/sec    1.00    122.0±4.77ms 20.0 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:1200/purses:50     1.45    181.4±4.73ms 13.5 KElem/sec    1.00    124.8±3.87ms 19.6 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:2500/purses:1      1.38    363.3±9.10ms  6.7 KElem/sec    1.00    262.6±9.16ms  9.3 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:2500/purses:100    1.50   383.9±15.24ms  6.4 KElem/sec    1.00    255.8±4.55ms  9.5 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:2500/purses:20     1.39    366.0±5.85ms  6.7 KElem/sec    1.00    263.8±5.18ms  9.3 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:2500/purses:50     1.45    361.9±9.17ms  6.7 KElem/sec    1.00    249.5±3.56ms  9.8 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:500/purses:1       1.52     76.4±0.82ms 31.9 KElem/sec    1.00     50.4±1.27ms 48.4 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:500/purses:100     1.46     77.3±1.74ms 31.6 KElem/sec    1.00     52.9±1.76ms 46.2 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:500/purses:20      1.40     75.2±0.83ms 32.5 KElem/sec    1.00     53.8±0.97ms 45.4 KElem/sec
tps_native/multiple_native_transfers/criterion::measurement::WallTime/transfers:500/purses:50      1.47     77.0±2.92ms 31.7 KElem/sec    1.00     52.5±1.02ms 46.5 KElem/sec
```


node/src/components/contract_runtime/error.rs
```diff
@@ -25,10 +25,6 @@ pub(crate) enum ConfigError {
 /// An error raised by a contract runtime variant.
 #[derive(Debug, thiserror::Error)]
 pub enum BlockExecutionError {
-    /// Currently the contract runtime can only execute one commit at a time, so we cannot handle
-    /// more than one execution result.
-    #[error("More than one execution result")]
-    MoreThanOneExecutionResult,
```
goral09: Can we extract this change to a separate PR? Assuming that it has far reaching consequences - i.e. we do assume in the other layers of the node that there can be multiple blocks in which a deploy is included in.

execution_engine_testing/tests/benches/transfer_bench.rs
```diff
@@ -408,6 +503,17 @@ pub fn transfer_bench(c: &mut Criterion) {
...
+);
+criterion_group!(
+    name = benches;
+    config = Criterion::default().with_measurement(WallTime);
+    targets = transfer_bench
+);
+criterion_main!(
+    //   benches,
+    native_transfer_benches
+);
```
goral09: Can we extract the extension of transfer benchmarks to a separate PR?

execution_engine_testing/tests/benches/transfer_bench.rs
```diff
@@ -247,7 +249,13 @@ pub fn transfer_to_existing_accounts(group: &mut BenchmarkGroup<WallTime>, shoul
     );
 }
 
-pub fn multiple_native_transfers(group: &mut BenchmarkGroup<WallTime>) {
+pub fn multiple_native_transfers_many_requests<M>(
+    group: &mut BenchmarkGroup<M>,
+    transfer_count: usize,
+    purse_count: usize,
```
goral09, line 256: Can you add a comment to this public method and explain how the `purse_count` and `transfer_count` affect the test?

execution_engine_testing/tests/benches/transfer_bench.rs
```diff
@@ -290,8 +303,79 @@ pub fn multiple_native_transfers(group: &mut BenchmarkGroup<WallTime>) {
...
+
+/// Creates one exec request with BLOCK_TRANSFER_COUNT transfers in it.
+fn multiple_native_transfers_one_request<M>(
+    group: &mut BenchmarkGroup<M>,
+    transfer_count: usize,
+    purse_count: usize,
+) where
+    M: Measurement,
+{
+    let target_account = TARGET_ADDR;
```
goral09: **[idea]:** In the future, we should add a test where transfers are made to mulitple, different accounts. This way we're eliminating any potential skews of the results where LMDB would use "hot" path and have the account memory mapped.

execution_engine_testing/tests/benches/transfer_bench.rs
dwerner: Well I think it's appropriate for this PR as it measures the performance changes in the PR.

node/src/components/contract_runtime/error.rs
dwerner: This is relevant to this PR - we are now ending up with many more execution results, now separated by deploy hash, so we never create this variant.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -71,6 +74,32 @@ impl LmdbGlobalState {
             empty_root_hash,
         }
     }
+
+    /// Creates an in-memory cache for changes written.
+    pub fn create_scratch(&self) -> ScratchGlobalState {
+        ScratchGlobalState::new(
+            Arc::clone(&self.environment),
+            Arc::clone(&self.trie_store),
+            self.empty_root_hash,
```
xcthulhu, line 83: So I'm going to recommend that we make this a constant rather than carry it around.

It's the blake2b digest of the empty pointer block.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -104,6 +106,9 @@ pub struct WasmTestBuilder<S> {
     standard_payment_hash: Option<ContractHash>,
     /// Auction contract key
     auction_contract_hash: Option<ContractHash>,
+
+    /// KLUDGE: for testing multiple-execution...
```
xcthulhu: Hmm... why is this a kludge?

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -231,6 +250,23 @@ impl LmdbWasmTestBuilder {
...
+    /// Creates new instance of builder and applies values only which allows the engine state to be
+    /// swapped with a new one, possibly after running genesis once and reusing existing database
+    /// (i.e. LMDB).
+    pub fn new_with_config_and_result<T: AsRef<OsStr> + ?Sized>(
+        data_dir: &T,
+        engine_config: EngineConfig,
+        result: &WasmTestResult<LmdbGlobalState>,
+    ) -> Self {
+        let mut builder = Self::new_with_config(data_dir, engine_config);
+        // Applies existing properties from gi
```
xcthulhu: I'm a little confused, what's gi?

execution_engine/src/storage/global_state/lmdb.rs
dwerner: Well this field may be named inappropriately. An empty state root hash (a single pointer block `Trie::Node` with no entries) which is hashed is written to LMDB, and then that hash passed along inside LmdbGlobalState. This makes this task more than just lifting a constant out - we would need to ensure that the state root exists in the database. That said I understand that the hash would always be the same - however I'd argue that this current path leads developers back to the source of truth, without having a magic constant and an assumed-to-be-present database record.

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: I'll be removing this.

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: I'll be removing this method entirely, it snuck in during a rebase but it's not in dev.

execution_engine/src/storage/global_state/lmdb.rs
xcthulhu: It’s a confusing wart, so I guess there’s no nice way about this.

execution_engine/src/storage/global_state/in_memory.rs
```diff
@@ -197,6 +197,24 @@ impl StateReader<Key, StoredValue> for InMemoryGlobalStateView {
     }
 }
 
+impl CommitProvider for InMemoryGlobalState {
```
xcthulhu, line 200: Maybe cfg(test)?  This is an interface we will want to deprecate eventually…

dwerner: `CommitProvider` trait provides a `commit` method, which is an extracted capacity of all global state implementations (except `ScratchGlobalState`). Removing or marking this `cfg(test)` would make `InMemoryGlobalState` useless and stop `execution_engine_testing` from being able to reference it.

utils/retrieve-state/src/lib.rs
```diff
@@ -26,6 +26,7 @@ pub mod storage;
 pub const LMDB_PATH: &str = "lmdb-data";
 pub const CHAIN_DOWNLOAD_PATH: &str = "chain-download";
 pub const DEFAULT_MAX_DB_SIZE: usize = 483_183_820_800; // 450 gb
+pub const VALGRIND_MAX_DB_SIZE: usize = 483_183_820; // .4 gb
```
goral09: Why is this part of the PR?

utils/dry-run-deploys/src/main.rs
```diff
@@ -129,6 +140,12 @@ async fn main() -> Result<(), anyhow::Error> {
             .map_err(anyhow::Error::msg)?;
 
         let header = block_and_execution_effects.block.take_header();
+        let expected = block.take_header();
+        assert_eq!(
+            header.state_root_hash(),
+            expected.state_root_hash(),
+            "state root hash mismatch"
+        );
```
goral09, line 148: Are these changes necessary for the PR?

node/src/components/contract_runtime.rs
```diff
@@ -150,6 +150,7 @@ impl Debug for ContractRuntime {
 pub struct ContractRuntimeMetrics {
     run_execute: Histogram,
     apply_effect: Histogram,
+    commit_cache: Histogram,
```
goral09: This histogram doesn't seem to be used anywhere.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -298,11 +311,54 @@ impl LmdbWasmTestBuilder {
...
+                transforms
+            },
+        ));
+        self.exec_results.push(
+            exec_results
+                .into_iter()
+                .map(|(_deploy_hash, res)| res)
+                .map(Rc::new)
+                .collect(),
+        );
```
goral09: We should be indexing transforms and execution results by the deploy they originated from. Otherwise, we're risking loosing the data (when we overwrite duplicates).

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +133,111 @@ pub fn execute_finalized_block(
...
     metrics: Option<Arc<ContractRuntimeMetrics>>,
+    protocol_version: ProtocolVersion,
+    finalized_block: &FinalizedBlock,
+    deploys: Vec<Deploy>,
     state_root_hash: Digest,
-    deploy_hash: DeployHash,
-    execution_results: ExecutionResults,
-) -> Result<(Digest, ExecutionResult), BlockExecutionError> {
-    let ee_execution_result = execution_results
+    block_time: u64,
```
goral09: `block_time` comes from the `finalized_block` that is passed in as well. Let's remove the `block_time` argument.

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +133,111 @@ pub fn execute_finalized_block(
...
-        } => {
-            // Failure to execute a contract is a user error, not a system error.
-            // We do want to see the deploy hash, error, and cost in the logs.
-            // We don't need to see the effects in the logs.
-            debug!(?deploy_hash, ?error, %cost, "execution failure");
-            execution_journal
+        let journal = ee_execution_result.execution_journal().clone();
+        let journal: AdditiveMap<Key, Transform> = journal.into();
+        for (key, transform) in journal.iter() {
+            transforms.insert(*key, transform.clone());
```
goral09: **[blocker]:** Isn't this going to overwrite the transform under the `key`? For `Write`s that's fine but not for `Add`s - we should merge the results.

goral09 (changes requested):
Not done with the review completely but it seems to me that there's a bug in `extract_transforms_from_execution_results` that is a blocker.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -460,19 +496,20 @@ where
...
         correlation_id: CorrelationId,
         mut exec_request: ExecuteRequest,
-    ) -> Result<ExecutionResults, Error> {
+    ) -> Result<Execution, Error> {
         let executor = Executor::new(*self.config());
 
-        let deploys = exec_request.take_deploys();
-        let mut results = ExecutionResults::with_capacity(deploys.len());
+        let deploy_items = exec_request.take_deploys();
+        let mut exec_results = BTreeMap::new();
```
goral09: **[blocker]:** This needs to be a `Vec` to retain the order of transforms. Otherwise, we might `Write` a transform that's not issued by the last transaction in a block 😬 The order has to be kept and reported upstream as well - `Execution` should hold a `Vec` instead of `BTreeMap`.

node/src/components/contract_runtime/operations.rs
goral09: We need a test similar to [this one](https://github.com/dwerner/casper-node/blob/many-deploys-per-commit-3/execution_engine_testing/tests/src/test/regression/transforms_must_be_ordered.rs) but one that will include multiple deploys trying to modify the same keys (create transforms for the same key) and validate that the last comitted value is what we expect to be.

goral09 (commented):
So it looks to me that `ScratchGlobalState` is similar to `TrackingCopy` in that it's caching eventual changes made to the underlying GlobalState but while `TrackingCopy` does so within the context of a single deploy execution, `ScratchGlobalState` does it across the whole block execution. I think that similar mechanisms (that have been applied to `TrackingCopy` cache) should be applied to `ScratchGlobalState` cache to limit its size - namely, we can always invalidate what was _read_ and keep cached the mutations (`Write` and `Add`). Actually, I am not convinced anymore that we need to cache `READ` values as LMDB will memory map it anyway.

utils/dry-run-deploys/src/main.rs
dwerner: These are part of the manual test I ran to ensure that the state roots produced were the expected values.

utils/retrieve-state/src/lib.rs
dwerner: This doesn't need to be included. Removed.

node/src/components/contract_runtime.rs
dwerner: Removed

node/src/components/contract_runtime/operations.rs
dwerner: Done

execution_engine/src/core/engine_state/mod.rs
dwerner: Done.

node/src/components/contract_runtime/operations.rs
dwerner: Completed the ordering here, will look into the test you're suggesting.

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: Done.

node/src/components/contract_runtime/operations.rs
dwerner: This is completed now.

dwerner:
Update on this:

I've tried to address all the review comments - ended up force-pushing a single commit over due to a mixup with a few merges. This is a read-tree out of the old branch for brevity's sake.

New performance tests show this is closer to a 75% improvement in terms of wall time.

```
   Compiling casper-engine-tests v0.1.0 (/home/dan/Development/casper-node/execution_engine_testing/tests)
    Finished bench [optimized] target(s) in 2m 13s
     Running unittests (target/release/deps/transfer_bench-b13e06b7cc517c43)
tps_native/type:scratch/transfers:500/purses:50/metric:criterion::measurement::WallTime                                                                            
                        time:   [56.671 ms 57.731 ms 58.839 ms]
                        thrpt:  [42.489 Kelem/s 43.304 Kelem/s 44.114 Kelem/s]
tps_native/type:lmdb/transfers:500/purses:50/metric:criterion::measurement::WallTime                                                                           
                        time:   [208.16 ms 223.33 ms 230.10 ms]
                        thrpt:  [10.865 Kelem/s 11.194 Kelem/s 12.010 Kelem/s]
tps_native/type:scratch/transfers:1200/purses:50/metric:criterion::measurement::WallTime                                                                           
                        time:   [150.52 ms 151.87 ms 154.49 ms]
                        thrpt:  [16.182 Kelem/s 16.461 Kelem/s 16.609 Kelem/s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
tps_native/type:lmdb/transfers:1200/purses:50/metric:criterion::measurement::WallTime                                                                           
                        time:   [533.50 ms 545.04 ms 557.69 ms]
                        thrpt:  [4.4828 Kelem/s 4.5869 Kelem/s 4.6861 Kelem/s]
tps_native/type:scratch/transfers:2500/purses:50/metric:criterion::measurement::WallTime                                                                           
                        time:   [288.39 ms 290.31 ms 293.73 ms]
                        thrpt:  [8.5111 Kelem/s 8.6115 Kelem/s 8.6687 Kelem/s]
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
tps_native/type:lmdb/transfers:2500/purses:50/metric:criterion::measurement::WallTime                                                                           
                        time:   [1.0565 s 1.0875 s 1.1064 s]
                        thrpt:  [2.2597 Kelem/s 2.2988 Kelem/s 2.3663 Kelem/s]
tps_native/type:scratch/transfers:500/purses:100/metric:criterion::measurement::WallTime                                                                           
                        time:   [76.025 ms 77.032 ms 77.962 ms]
                        thrpt:  [32.067 Kelem/s 32.454 Kelem/s 32.884 Kelem/s]
tps_native/type:lmdb/transfers:500/purses:100/metric:criterion::measurement::WallTime                                                                           
                        time:   [218.51 ms 229.84 ms 235.69 ms]
                        thrpt:  [10.607 Kelem/s 10.877 Kelem/s 11.441 Kelem/s]
tps_native/type:scratch/transfers:1200/purses:100/metric:criterion::measurement::WallTime                                                                           
                        time:   [145.28 ms 146.57 ms 149.37 ms]
                        thrpt:  [16.736 Kelem/s 17.056 Kelem/s 17.208 Kelem/s]
tps_native/type:lmdb/transfers:1200/purses:100/metric:criterion::measurement::WallTime                                                                           
                        time:   [512.41 ms 538.90 ms 554.26 ms]
                        thrpt:  [4.5105 Kelem/s 4.6390 Kelem/s 4.8789 Kelem/s]
tps_native/type:scratch/transfers:2500/purses:100/metric:criterion::measurement::WallTime                                                                           
                        time:   [304.40 ms 312.26 ms 318.48 ms]
                        thrpt:  [7.8497 Kelem/s 8.0061 Kelem/s 8.2128 Kelem/s]
tps_native/type:lmdb/transfers:2500/purses:100/metric:criterion::measurement::WallTime                                                                           
                        time:   [1.0003 s 1.0313 s 1.0470 s]
                        thrpt:  [2.3878 Kelem/s 2.4241 Kelem/s 2.4993 Kelem/s]
```

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -104,13 +110,38 @@ pub struct WasmTestBuilder<S> {
...
+                    .iter()
+                    .cloned()
+                    .map(|(_deploy_hash, result)| result)
+                    .collect()
+            })
+            .collect()
+    }
+
+    /// Returns transforms.
+    pub fn transforms(&self) -> Vec<AdditiveMap<Key, Transform>> {
```
goral09: I would return `DeployHash`-es as well, so that the caller can decide whether to use it/correlate with an order of the submitted txns.

Same comment aplies for `fn exec_results` above.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -155,10 +187,15 @@ impl<S> Clone for WasmTestBuilder<S> {
...
             standard_payment_hash: self.standard_payment_hash,
             auction_contract_hash: self.auction_contract_hash,
+            scratch_engine_state: None,
         }
     }
 }
 
+/// A wrapper type to disambiguate builder from an actual result
+#[derive(Clone)]
+pub struct WasmTestResult<S>(WasmTestBuilder<S>);
```
goral09: This doesn't seem to be used.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -281,6 +319,7 @@ impl LmdbWasmTestBuilder {
             handle_payment_contract_hash: None,
             standard_payment_hash: None,
             auction_contract_hash: None,
+            scratch_engine_state: None,
```
goral09, line 293: Why do we not create it immediately in the constructor?

goral09: I've got my partial answer - the `get_scratch_global_state` is not defined on `InMemGlobalState`.

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: Refactored this to return the deploys, however the above `exec_results` method is not needed anymore.

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: Removed.

execution_engine_testing/tests/benches/transfer_bench.rs
```diff
@@ -377,7 +424,13 @@ pub fn native_transfer_bench(c: &mut Criterion) {
     // Measure by elements where one element per second is one transaction per second
     group.throughput(Throughput::Elements(BLOCK_TRANSFER_COUNT as u64));
 
-    multiple_native_transfers(&mut group);
+    for purse_count in [50, 100] {
+        for transfer_count in [500, 1200, 2500usize] {
```
goral09: IIRC, on MainNet we allow for 1500 native transfers - it would be good to get the performance figures when all 1500 transfers were made between different accounts. To do that we would have to create 1500 unique purses and make 1500 transfers where each purse is used only once.

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +131,121 @@ pub fn execute_finalized_block(
...
+    // Create a new EngineState that reads from LMDB but only caches changes in memory.
+    let scratch_state = engine_state.get_scratch_engine_state();
+
+    let deploy_items = deploys
+        .iter()
+        .cloned()
+        .map(DeployItem::from)
+        .collect::<Vec<_>>();
+
+    let mut execution_results = BTreeMap::new();
```
goral09: Why not make it a `mut Vec` and push in line 191 rather than insert into a tree? We execute in order so we should collect the results in order as well.

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +131,121 @@ pub fn execute_finalized_block(
...
+
+    for deploy_item in deploy_items {
+        // Note that we generate a request to execute all deploys in the block.
```
goral09: The comment says that we generate a request to execute all deploys in the block but it uses only a single `deploy_jtem`.

I had actually thought that we would put all of the deploys in a single `ExecuteRequest` and the `scratch_state` would take care of executing them in order and "committing" the results (applying the transforms).

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +131,121 @@ pub fn execute_finalized_block(
...
+                .find(|deploy| node_deploy_hash == *deploy.id())
+            {
+                execution_results.insert(
+                    node_deploy_hash,
+                    (
+                        deploy.header().clone(),
+                        ExecutionResult::from(&ee_exec_result),
+                    ),
+                );
+            }
```
goral09: Did you actually mean to pass in all `deploy_items` to `ExecRequest`? If not, all of these iterations over the results are unnecessary as there will always be just one `execution result`.

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +131,121 @@ pub fn execute_finalized_block(
...
+        let execute_request = ExecuteRequest::new(
+            state_root_hash,
+            block_time,
+            vec![deploy_item],
+            protocol_version,
+            finalized_block.proposer(),
+        );
+
+        // Execute our request against the in-memory cache.
+        let Execution { exec_results } = execute(&scratch_state, metrics.clone(), execute_request)?;
```
goral09: This is a good question - why did we not implement this feature to be self-contained within the `EngineState`? It uses the `S: StateProvider + CommitProvider,` which should be sufficient to provide the `ScratchGlobalState` and we would not need to do this whole "dance" around deploys and effects here.

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +131,121 @@ pub fn execute_finalized_block(
...
-            // We don't need to see the effects in the logs.
-            debug!(?deploy_hash, %cost, "execution succeeded");
-            execution_journal
+) -> Result<(ExecResultMap, Digest), BlockExecutionError> {
+    let block_time = finalized_block.timestamp().millis();
+
+    // Create a new EngineState that reads from LMDB but only caches changes in memory.
+    let scratch_state = engine_state.get_scratch_engine_state();
+
+    let deploy_items = deploys
```
goral09: You can simplify the following code like
```rust

    let execution_results = deploys
        .into_iter()
        .flat_map(|deploy| {
            let deploy_item = DeployItem::from(deploy);

            let execute_request = ExecuteRequest::new(
                state_root_hash,
                block_time,
                vec![deploy_item],
                protocol_version,
                finalized_block.proposer(),
            );

            // Execute our request against the in-memory cache.
            let Execution { exec_results } =
                execute(&scratch_state, metrics.clone(), execute_request)?;

            let transforms_by_deploy = extract_transforms_from_execution_results(&exec_results);

            // This ensures that transforms are applied in the order they were executed.
            for (_deploy_hash, transforms) in transforms_by_deploy.into_iter() {
                // Because we are applying the transforms to the ScratchGlobalState, no new
                // state-root-hash is generated here.
                let _same_root_hash =
                    apply_transforms(&scratch_state, metrics.clone(), state_root_hash, transforms)?;
            }

            exec_results
                .into_iter()
                .filter_map(|(deploy_hash, ee_exec_result)| {
                    let node_deploy_hash: NodeDeployHash = deploy_hash.into();
                    if node_deploy_hash == *deploy.id() {
                        Some((deploy.header(), ExecutionResult::from(&ee_exec_result)))
                    } else {
                        warn!("");
                        None
                    }
                })
        })
        .collect();
```

This avoid cloning of `deploys` into `deploy_items` or the `deploy.header()`. It doesn't compile yet but the fixes should be easy.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,498 @@
...
+
+    fn insert(&mut self, key: Key, value: StoredValue) {
+        self.stored_values.insert(key, value);
+    }
+
+    fn get(&self, key: &Key) -> Option<&StoredValue> {
+        self.stored_values
+            .iter()
+            .find(|(cached_key, _)| *cached_key == key)
+            .map(|(_, value)| value)
```
goral09:
```suggestion
        self.stored_values.get(key)
```

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,498 @@
...
+    pub(crate) empty_root_hash: Digest,
+}
+
+/// Represents a "view" of global state at a particular root hash.
+pub struct ScratchGlobalStateView {
+    cache: SharedCache,
+    /// Environment for LMDB.
+    pub(crate) environment: Arc<LmdbEnvironment>,
+    /// Trie store held within LMDB.
+    pub(crate) store: Arc<LmdbTrieStore>,
```
goral09: Earlier you called it `trie_store`, let's be consistent with the naming.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,498 @@
...
+        for result in keys_iter {
+            match result {
+                Ok(key) => ret.push(key),
+                Err(error) => return Err(error),
+            }
+        }
+        txn.commit()?;
+        Ok(ret)
+    }
+}
```
goral09, line 183: I hope we can get rid of this code duplication when we implement the trie-level caching.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,498 @@
...
+            missing_trie_keys::<Key, StoredValue, lmdb::RoTransaction, LmdbTrieStore, Self::Error>(
+                correlation_id,
+                &txn,
+                self.trie_store.deref(),
+                trie_keys,
+            )?;
+        txn.commit()?;
+        Ok(missing_descendants)
+    }
+}
```
goral09, line 324: If `ScratchGlobalState` struct contained `LmdbGlobalState` then all of these could just forward to the underlying DB and only `StateReader::read` and `CommitProvider::commit` would reimplement their logic on top of it.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -85,13 +88,16 @@ pub struct WasmTestBuilder<S> {
     /// [`EngineState`] is wrapped in [`Rc`] to work around a missing [`Clone`] implementation
     engine_state: Rc<EngineState<S>>,
     /// [`ExecutionResult`] is wrapped in [`Rc`] to work around a missing [`Clone`] implementation
-    exec_results: Vec<Vec<Rc<ExecutionResult>>>,
+    exec_results: Vec<Vec<(DeployHash, Rc<ExecutionResult>)>>,
+    /// Upgrade results.
     upgrade_results: Vec<Result<UpgradeSuccess, engine_state::Error>>,
+    /// .
     genesis_hash: Option<Digest>,
+    ///
```
goral09: Can you fill in these comments with actual content?

dwerner: Done

dwerner:
Update on this: I've reverted much of the contract_runtime and related operations code that changed to accommodate _multiple deploy items per exec_, which should never have been a part of this PR. The primary change is now simply making use of `ScratchGlobalState` in our exec/commit cycle. @goral09

execution_engine/src/storage/global_state/scratch.rs
dwerner: Ah, I had changed the underlying store to a `Vec<(Key, StoredValue)>` and neglecte to put this right.

goral09:
> Update on this: I've reverted much of the contract_runtime and related operations code that changed to accommodate _multiple deploy items per exec_, which should never have been a part of this PR. The primary change is now simply making use of `ScratchGlobalState` in our exec/commit cycle. @goral09

Ah, I have to start review from the start! 😬

dwerner:
> Ah, I have to start review from the start! grimacing

Perhaps, but many of your review comments were pointing me towards this.

> If ScratchGlobalState struct contained LmdbGlobalState then all of these could just forward to the underlying DB and only StateReader::read and CommitProvider::commit would reimplement their logic on top of it.

I tried this, however it led to refactoring the `LmdbGlobalState` and `LmdbGlobalStateView` as they require a `StateReader::checkout()` and therefore a transaction to construct the view. This is not needed because we are doing things in memory at this point.

execution_engine/src/storage/global_state/scratch.rs
goral09: Still the issue.

node/src/components/contract_runtime/operations.rs
```diff
@@ -155,61 +131,121 @@ pub fn execute_finalized_block(
...
+        .iter()
+        .cloned()
+        .map(DeployItem::from)
+        .collect::<Vec<_>>();
```
goral09: You're still cloning the items unnecessarily.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,495 @@
...
+    /// State hash returned is the one provided, as we do not write to lmdb with this kind of global
+    /// state.
+    fn commit(
+        &self,
+        _correlation_id: CorrelationId,
+        state_hash: Digest,
+        effects: AdditiveMap<Key, Transform>,
+    ) -> Result<Digest, Self::Error> {
+        for (key, transform) in effects.into_iter() {
+            let read_result = self.cache.read().unwrap().get(&key).cloned();
```
goral09:
```suggestion
            let cached_value = self.cache.read().unwrap().get(&key).cloned();
```

b/c it's not `READ` and not really a result.

node/src/components/contract_runtime/operations.rs
goral09: We don't have to clone the data - we can simply iterate over the input `deploys`. I had created a PR to your repo with that approach - it avoid cloning the data which you do twice in here.

node/src/components/contract_runtime/error.rs
goral09: We could be creating this variant if we used the PR I made against your repo.

goral09:
@dwerner This PR fails the compilation step.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,495 @@
...
+        }
+    }
+
+    #[test]
+    fn checkout_fails_if_unknown_hash_is_given() {
+        let TestState { state, .. } = create_test_state();
+        let fake_hash: Digest = Digest::hash(&[1u8; 32]);
+        let result = state.checkout(fake_hash).unwrap();
+        assert!(result.is_none());
+    }
```
goral09: This test doesn't check anything new added in this PR as far as I can tell - the `TestState.state` is an `LmdbGlobalState` so the `state.checkout` unit tests the LMDB. Let's either move it to the correct module or delete it from here as it's creating a false sense of security about the `ScratchState`.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,495 @@
...
+        for (key, transform) in effects.into_iter() {
+            let cached_value = self.cache.read().unwrap().get(&key).cloned();
+            let value = match (cached_value, transform) {
+                (None, Transform::Write(new_value)) => new_value,
+                (None, transform) => {
+                    error!(
+                        ?key,
+                        ?transform,
+                        "Key not found while attempting to apply transform"
+                    );
```
goral09: Ideally, in this case, we would read from the underlying LMDB and apply the transform on top if it.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,495 @@
...
+        let test_pairs_updated = create_test_pairs_updated();
+
+        let TestState { state, root_hash } = create_test_state();
+
+        let scratch = state.create_scratch();
+
+        let effects: AdditiveMap<Key, Transform> = {
+            let mut tmp = AdditiveMap::new();
+            for TestPair { key, value } in &test_pairs_updated {
+                tmp.insert(*key, Transform::Write(value.to_owned()));
```
goral09, line 460: We should have a test that uses `Transform::Add` as well, otherwise the unit tests so far don't verify that we're actually reading from LMDB into the Scratch* as all of the effects are `Write`s and those don't need a "starting" value.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,495 @@
...
+        let stored_values = scratch.into_inner();
+        let updated_checkout = state.checkout(updated_hash).unwrap().unwrap();
+        let all_keys = updated_checkout
+            .keys_with_prefix(correlation_id, &[])
+            .unwrap();
+
+        assert_eq!(all_keys.len(), stored_values.len());
+
+        for key in all_keys {
+            println!("key {}", key);
```
goral09: Leftover `println!`.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,495 @@
...
+                original_checkout.read(correlation_id, &key).unwrap()
+            );
+        }
+        assert_eq!(
+            None,
+            original_checkout
+                .read(correlation_id, &test_pairs_updated[2].key)
+                .unwrap()
+        );
+    }
```
goral09, line 606: This test doesn't create a `ScratchGlobalState` so it's testing LMDB. We should rewrite it so that it tests the new logic.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,495 @@
...
+}
+
+impl StateReader<Key, StoredValue> for ScratchGlobalStateView {
+    type Error = error::Error;
+
+    fn read(
+        &self,
+        correlation_id: CorrelationId,
+        key: &Key,
+    ) -> Result<Option<StoredValue>, Self::Error> {
```
goral09, line 110: We need a unit test for this method - i.e. when reading a not-yet-cached key from `ScratchGlobalStateView` it should read from the LMDB and update the cache. Since in practice we can't mock the underlying `trie_store` at that `root_hash`, it should be OK to simply verify that reading from `ScratchGlobalStateView` returns correct values.

execution_engine/src/storage/global_state/scratch.rs
dwerner: Deleted it, thanks

execution_engine/src/storage/global_state/scratch.rs
dwerner: This code was lifted from the top-level `commit` method where effects are collapsed. I believe this is correct because we would have had to read the value into cache in order to write any changes to it. So somehow if we have a transform that doesn't correspond to a stored value that we already have cached, we're doing something wrong and this should be an error.

mpapierski (approved):
I'll go ahead and give optimistic approval as there are some comments yet to be answered. I played with this code before, and I saw the benefits. What could be said in review to improve it was told already (i.e., extra test for transfers) but please wait for more approvals as it has a significant impact

execution_engine/src/storage/global_state/scratch.rs
dwerner: We do test this, in `transfer_cached`.

execution_engine/src/storage/global_state/scratch.rs
dwerner: Fixed

execution_engine/src/storage/global_state/scratch.rs
goral09: `transfer_cached` is a higher level test. We could test it as a unit test here as well. The other tests are very similar and almost do that already.

dwerner: This is now covered.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -298,11 +307,65 @@ impl LmdbWasmTestBuilder {
         path.push(GLOBAL_STATE_DIR);
         path
     }
+
+    /// Execute and commit transforms from an ExecuteRequest into a scratch global state.
+    /// You MUST call scratch_flush to flush these changes to LmdbGlobalState.
+    pub fn scratch_exec(&mut self, mut exec_request: ExecuteRequest) -> &mut Self {
```
goral09: Since we use `ScratchGlobalState` in the contract runtime to execute all deploys, it should be the default (and the only option) in the tests as well - we should modify the `WasmTestBuilder::exec` to use `ScratchGlobalState`.

dwerner: I agree that `WasmTestBuilder` needs some work, but it's not true today that all `exec` calls go through `ScratchGlobalState`.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,564 @@
...
+                        }
+                        CLType::String => {}
+                        _ => unreachable!(
+                            "remember to update the test pairs if new cl_types are added"
+                        ),
+                    }
+                }
+            }
+            tmp
+        };
```
goral09: This might be much simpler and cleaner if instead of having `CLValue`s in `test_pairs_updated` we used `Transform`s.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,564 @@
...
+        scratch
+            .commit(correlation_id, root_hash, add_effects.clone())
+            .unwrap();
+        let updated_hash = state2
+            .commit(correlation_id, updated_hash, add_effects)
+            .unwrap();
+
+        let scratch_checkout = scratch.checkout(root_hash).unwrap().unwrap();
```
goral09, line 542: This is a bit weird UX - we commit effects on top of `root_hash` and then get the _new_ values at that same root hash.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,564 @@
...
+            let mut tmp = AdditiveMap::new();
+            for TestPair { key, value } in &test_pairs_updated {
+                tmp.insert(*key, Transform::Write(value.to_owned()));
+            }
+            tmp
+        };
+
+        let updated_hash = scratch.commit(correlation_id, root_hash, effects).unwrap();
+
+        let updated_checkout = scratch.checkout(updated_hash).unwrap().unwrap();
```
goral09, line 584: The test passes with `scratch.checkout(root_hash)` as well. b/c of how the `ScratchGlobalState::commit` works, I think we should be ignoring the returned hash.

Or even better, make sure it doesn't change the original hash! 
```rust
assert_eq!(updated_hash, root_hash);
```

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,564 @@
...
+                Some(value),
+                updated_checkout.read(correlation_id, &key).unwrap()
+            );
+        }
+
+        let original_checkout = state.checkout(root_hash).unwrap().unwrap();
+        for TestPair { key, value } in create_test_pairs().iter().cloned() {
+            assert_eq!(
+                Some(value),
+                original_checkout.read(correlation_id, &key).unwrap()
```
goral09, line 597: Can you add an assertion message for the error case? IIUC, it would fail if the original state gets update via `scratch` manipulation.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,564 @@
...
+            );
+        }
+        assert_eq!(
+            None,
+            original_checkout
+                .read(correlation_id, &test_pairs_updated[2].key)
```
goral09, line 603: Why not iterate over all `test_pairs_updated` keys and verify that the `original_checkout` was not updated.

execution_engine_testing/test_support/src/wasm_test_builder.rs
goral09: All calls to `operations::execute_finalized_block` use `ScratchGlobalState`. Which calls do not use it (but use a non-`ScratchGlobalState`)?

execution_engine_testing/tests/benches/transfer_bench.rs
```diff
@@ -309,10 +341,22 @@ fn transfer_to_account_multiple_native_transfers(
...
-        builder.commit();
+        if use_scratch {
+            builder.scratch_exec(request).expect_success();
+        } else {
+            builder.exec(request).expect_success();
+            builder.commit();
+        }
+    }
+    if use_scratch {
+        builder.write_scratch_to_lmdb();
```
goral09, line 352: Why not include that in the previous `if use_scratch { ... }`?

execution_engine_testing/tests/src/test/contract_api/transfer_cached.rs
```diff
@@ -0,0 +1,191 @@
...
+
+static ACCOUNT_2_SECRET_KEY: Lazy<SecretKey> =
+    Lazy::new(|| SecretKey::secp256k1_from_bytes(&[210u8; 32]).unwrap());
+static ACCOUNT_2_PUBLIC_KEY: Lazy<PublicKey> =
+    Lazy::new(|| PublicKey::from(&*ACCOUNT_2_SECRET_KEY));
+static ACCOUNT_2_ADDR: Lazy<AccountHash> = Lazy::new(|| ACCOUNT_2_PUBLIC_KEY.to_account_hash());
+
+const ARG_TARGET: &str = "target";
+const ARG_AMOUNT: &str = "amount";
+const ARG_ID: &str = "id";
```
goral09: These can be imported from the `mint` crate.

execution_engine_testing/tests/src/test/contract_api/transfer_cached.rs
```diff
@@ -0,0 +1,191 @@
...
+    let pre_state_hash = builder.get_post_state_hash();
+
+    // Default account to account 1
+    let mut exec_builder = ExecuteRequestBuilder::new();
+    exec_builder = exec_builder.push_deploy(transfer(
+        *DEFAULT_ACCOUNT_ADDR,
+        runtime_args! {
+           ARG_TARGET => *ACCOUNT_1_ADDR,
+           ARG_AMOUNT => U512::one(),
+           ARG_ID => <Option<u64>>::None,
```
goral09: Let's extract it to `const ID_ARG: Option<u64> = None` and reuse throughout.

execution_engine_testing/tests/src/test/contract_api/transfer_cached.rs
```diff
@@ -0,0 +1,191 @@
...
+        U512::zero(),
+        "account 1 balance should have been completely consumed"
+    );
+
+    let account_2_balance = builder.get_purse_balance(account2.main_purse());
+    assert_eq!(
+        account_2_balance, *TRANSFER_AMOUNT,
+        "account 2 balance should have changed"
+    );
+}
```
goral09, line 189: Could we have used the `operations::execute_finalized_block` that hides the whole complexity of creation of `ScratchGlobalState`, commit and flushing its effects into the underlying LMDB?

dwerner: `contract_runtime::operations::execute_finalized_block` exists in the node, and relies on an input parameter `FinalizedBlock` which exists in the node, and we are upstream. I agree that `execute_finalized_block` should be lifted out of the node, however I don't think that should be done in this PR.

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: Step is called through an `ExecuteRequest`, and uses `WasmTestBuilder::exec`. `execute_finalized_block` exists in the node and we don't call it directly.

goral09: I didn't realize that `Step` is executed via `WasmTestBuilder::exec`.

execution_engine_testing/tests/benches/transfer_bench.rs
dwerner: The previous `if` was in a loop - this needs to be run after the "block" has been completed and the cache is full of values ready to be written back.

execution_engine/src/storage/global_state/scratch.rs
dwerner: Yeah, I'll make it more clear that `ScratchGlobalState` doesn't do any hashing here.

dwerner:
> Come on :)

I love how intentional that commit looks.

execution_engine/src/storage/global_state/scratch.rs
dwerner: Actually, the two sets intersect, and this last assertion says "make sure that the value written to scratch does not exist in the lmdb state". Both original and updated should have the other two values.

goral09: Can you add that message then? :)

execution_engine/src/storage/global_state/scratch.rs
goral09: GitHub still doesn't show me the error message in this `assert_eq!`.

Fraser999 (approved):

goral09:
blocked by https://github.com/casper-network/casper-node/issues/2506

goral09 (changes requested):
Blocked until https://github.com/casper-network/casper-node/issues/2506 is done.

goral09 (approved):

goral09 (changes requested):
Blocking for the time being.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -0,0 +1,572 @@
...
+
+/// Global state implemented against LMDB as a backing data store.
+pub struct ScratchGlobalState {
+    /// Underlying, cached stored values.
+    cache: SharedCache,
+    /// Environment for LMDB.
+    pub(crate) environment: Arc<LmdbEnvironment>,
+    /// Trie store held within LMDB.
+    pub(crate) trie_store: Arc<LmdbTrieStore>,
+    // TODO: make this a lazy-static
```
xcthulhu, line 64: Might as well drop this TODO

execution_engine/src/storage/global_state/scratch.rs
xcthulhu: Isn't trie-level caching obviated by bulk-writing the trie?

Fraser999 (approved):

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -196,12 +196,45 @@ impl CommitProvider for ScratchGlobalState {
...
+                            );
+                            return Err(CommitError::KeyNotFound(key).into());
+                        }
+                        ReadResult::RootNotFound => {
+                            error!(root_hash=?state_hash, "root not found");
+                            return Err(CommitError::ReadRootNotFound(state_hash).into());
+                        }
+                    };
+                    txn.commit()?;
+                    updated_value
```
goral09, line 238: I copied (with slight modifications) this piece from another place. Maybe there's an option to avoid that but I didn't think about it when I was testing the quick-fix.

goral09 (approved):
No blockers but it might be possible to extract bug fixing code and reuse.

PR #2282, opened 2021-10-29
Reinstate caching of block proposer state

### LINKED ISSUE #2248: Make BlockProposer re-populate list of pending deploys on startup

Fraser999:
This PR reinstates caching of the pending deploys and transfers by the block proposer, so that this information can be repopulated on startup.

It reverts some of [#1216](https://github.com/casper-network/casper-node/pull/1216).  Before that PR, the block proposer cached finalized deploys too.  This PR does not re-introduce that behavior.  Finalized deploys are still provided by the storage component on startup.

However, the pending deploys and transfers _are_ stored regularly (every 10 seconds, just after the state is purged) and are read from storage on startup.  Only those deploys (read from storage) which are also not now in the finalized deploys list are retained in the pending collections.

There is still plenty of room for improving this mechanism, but alternative approaches could be significantly more complex and should be planned out accordingly.

Closes #2248.

### DIFF
--- a/node/src/components/block_proposer.rs
+++ b/node/src/components/block_proposer.rs
@@ -175,17 +175,35 @@ where
                     cached_state,
                 },
             ) => {
+                let (sets, pruned_hashes) = BlockProposerDeploySets::new(
+                    finalized_deploys,
+                    next_finalized_block,
+                    cached_state,
+                );
+
                 let mut new_ready_state = BlockProposerReady {
-                    sets: BlockProposerDeploySets::new(
-                        finalized_deploys,
-                        next_finalized_block,
-                        cached_state,
-                    ),
+                    sets,
                     unhandled_finalized: Default::default(),
                     deploy_config: *deploy_config,
                     request_queue: Default::default(),
                     local_config: local_config.clone(),
                 };
+
+                // Announce pruned hashes.
+                let pruned_count = pruned_hashes.total_pruned;
+                debug!(%pruned_count, "pruned deploys from buffer on loading");
+                effects.extend(
+                    effect_builder
+                        .announce_expired_deploys(pruned_hashes.expired_hashes_to_be_announced)
+                        .ignore(),
+                );
+
+                // After pruning, we store a state snapshot.
+                effects.extend(
+                    effect_builder
+                        .save_state(STATE_KEY.into(), CachedState::from(&new_ready_state.sets))
+                        .ignore(),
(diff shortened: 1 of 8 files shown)


fizyk20 (approved):
I think it _should_ be fine to do that, so looks okay :+1:

node/src/components/block_proposer.rs
```diff
@@ -174,12 +194,8 @@ where
...
                 self.state = BlockProposerState::Ready(new_ready_state);
 
-                // Start pruning deploys after delay.
-                effects.extend(
-                    effect_builder
-                        .set_timeout(PRUNE_INTERVAL)
-                        .event(|_| Event::Prune),
-                );
+                // Start pruning deploys.
+                effects.extend(effect_builder.immediately().event(|_| Event::Prune));
```
goral09: Why make this change?

goral09 (approved):
Looks good. I'm only worried how this implementation behaves when there are tens of thousands of pending transactions in the block proposer (need to be persisted, loaded, pruned...).

node/src/components/block_proposer.rs
Fraser999: I wanted to prune any potentially expired pending deploys as soon as we transition to the ready state so we don't accidentally propose them.

goral09: That is a good catch - but this doesn't guarantee that `Event::Prune` gets handled before a request to propose a block. Maybe instead, we should prune in the constructor of `BlockProposerDeploySets `?

dwerner:
> Looks good. I'm only worried how this implementation behaves when there are tens of thousands of pending transactions in the block proposer (need to be persisted, loaded, pruned...).

This is really my only concern as well. I'd like to see how this behaves when the node is shut down under load.

dwerner (approved):

Fraser999:
bors merge

Fraser999:
bors merge-

node/src/components/block_proposer.rs
Fraser999: > That is a good catch

Right back atcha :smile: 

Fixed in [1cb2ead](https://github.com/casper-network/casper-node/pull/2282/commits/1cb2ead9a4bb47b486ac87a7310a8bead1397bb6).

Fraser999:
bors merge

Fraser999:
bors merge

PR #2283, opened 2021-10-29
GH-2280: Fix misleading host functions gas charges

### LINKED ISSUE #2280: Gas Cost Issues with Faucet
Purse Create updated to 2.5 CSPR from 0.000_170 and we assume this has changed the gas needed for the faucet.  Prior to 1.4.1 the faucet payment was set to 0.2 CSPR  (mint and create both 10_000 motes -> 2.5 CSPR as well)

Call with 0.2 CSPR as payment failed with Out of Gas and 11_960_190 cost.   https://testnet.cspr.live/deploy/c54e1781b87a0f43222f04dab2792a2e503af1cbf42a3df9d8e20ce8f09c8f67

Call with 3.0 CSPR as payment succeeded with 168_507_540 cost listed.  The assumption is that the 2.5 CSPR account creation charge was part of the gas cost but not shown as cost of execution in the results.  https://testnet.cspr.live/deploy/7cec7ce433a714b6de88082ede6b79fcb4baca0e1ec7ce82422feafad68ebfc2

mpapierski:
Closes #2280

This PR fixes a problem where `transfer_to_account` and `transfer_from_purse_to_account` (and also `create_purse`) host functions are charging not only for host function cost, but also for a system contract charge.

Using those APIs leads to a situation where the user who uses the absolute minimum payment amount to cover the host function cost will receive a `GasLimit` error. In this case, the system will also incorrectly attempt to charge for system contracts.


execution_engine/src/shared/scoped_guard.rs
```diff
@@ -0,0 +1,49 @@
...
+
+/// When created it increases a counter, and decreases it when dropped.
+pub(crate) struct ScopedCountingGuard {
+    shared_value: Rc<RefCell<i32>>,
+}
+
+impl ScopedCountingGuard {
+    #[must_use]
+    pub(crate) fn new(shared_value: &Rc<RefCell<i32>>) -> Self {
+        shared_value.replace_with(|&mut value| value + 1);
```
luis-casper: Why do we use `replace_with` here instead of `borrow_mut` as in L20?

luis-casper: We might also want to use a checked addition here, even if we panic on overflow.

execution_engine_testing/tests/src/test/regression/gh_2280.rs
```diff
@@ -0,0 +1,359 @@
...
+    let gas_cost_diff = gas_cost_3.checked_sub(gas_cost_2).unwrap_or_default();
+    assert_eq!(
+        gas_cost_diff,
+        Gas::new(U512::from(HOST_FUNCTION_COST_CHANGE))
+    );
+}
+
+#[ignore]
+#[test]
+fn gh_2280_create_purse_should_always_cost_the_same_gas() {
```
luis-casper, line 195: This function is largely repeated code from the function above.
We should refactor these if possible.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -64,6 +67,9 @@ pub struct Runtime<'a, R> {
     host_buffer: Option<CLValue>,
     context: RuntimeContext<'a, R>,
     call_stack: Vec<CallStackElement>,
+    /// Value larger than 0 indicates executing under host function context, so we don't need to
+    /// charge for system contracts.
+    host_function_counter: Rc<RefCell<i32>>,
```
luis-casper: We might want to consider giving this type a name.

luis-casper (changes requested):

execution_engine/src/shared/scoped_guard.rs
mpapierski: Addressed

mpapierski: I addressed this. After thinking, I decided not to panic but have another error in case of overflow. I don't see any scenario where it would hit this error other than a programming error.

execution_engine_testing/tests/src/test/regression/gh_2280.rs
mpapierski: Refactored as much of duplicated code as possible out of it

execution_engine/src/core/runtime/mod.rs
mpapierski: Addressed

execution_engine/src/shared/scoped_guard.rs
```diff
@@ -1,23 +1,75 @@
...
+/// A generic counting flag used by the [`ScopedCountingGuard`].
+#[derive(Default, Clone)]
+pub struct ScopedCounter<T>(Rc<RefCell<T>>);
+
+impl<T: PartialOrd + Copy + Zero> ScopedCounter<T> {
+    pub(crate) fn count(&self) -> T {
+        *self.0.borrow()
+    }
+
+    pub(crate) fn is_set(&self) -> bool {
```
luis-casper: Maybe we should name this `is_zero` or `is_positive`, since counters can be set to zero.

execution_engine/src/shared/scoped_guard.rs
```diff
@@ -27,23 +79,36 @@ mod tests {
 
     #[test]
     fn scoping_test() {
-        let value = Rc::new(RefCell::new(0i32));
+        let counter = ScopedCounter::<i64>::default();
+
         {
-            assert_eq!(*value.borrow(), 0);
+            assert_eq!(counter.count(), 0);
```
luis-casper: The `is_*` method could be used.

execution_engine/src/core/execution/error.rs
```diff
@@ -156,6 +156,9 @@ pub enum Error {
     /// Missing system contract hash.
     #[error("Missing system contract hash: {0}")]
     MissingSystemContractHash(String),
+    /// Host function reached depth limit.
```
luis-casper: Since this limit is `2^64-1`, it doesn't make sense to have this error variant.
Reaching the limit (with the system contracts only) is clearly a bug and panicking would be preferable.

luis-casper (changes requested):

execution_engine/src/shared/scoped_guard.rs
luis-casper: But then I think we should panic rather than having the error variant, as it would be a programming error.

execution_engine/src/core/execution/error.rs
mpapierski: Fair - I removed this error variant, restored `#[must_use]` and added a `// SAFETY:` comment around `.expect()` call.

execution_engine/src/shared/scoped_guard.rs
mpapierski: Addressed

luis-casper (approved):

dwerner (approved):

Fraser999 (approved):

PR #2301, opened 2021-11-02
Query Global State by Contract Package Hash.

### LINKED ISSUE #1931: Query state via package hash
CC @EdHastingsCasperLabs

nilz3ro:
# About These Changes

This PR makes it possible to use contract package hashes when querying global state.
If you use a query path based on a contract package, the query will look for named keys in the latest version of the contract inside the contract package.

* Modifies the `query` function in `tracking_copy/mod.rs`.
* Introduces a new smart contract for testing: `set-named-key-stored`.
* closes #1931


execution_engine_testing/tests/src/test/regression/gh_1931.rs
```diff
@@ -0,0 +1,77 @@
...
+    let exec_request = ExecuteRequestBuilder::from_deploy_item(deploy_item).build();
+
+    let mut builder = InMemoryWasmTestBuilder::default();
+    builder
+        .run_genesis(&DEFAULT_RUN_GENESIS_REQUEST)
+        .exec(exec_request)
+        .expect_success()
+        .commit();
+
+    let _stored_contract_hash = builder
```
nilz3ro: remove this

execution_engine_testing/tests/src/test/regression/gh_1931.rs
```diff
@@ -0,0 +1,77 @@
+use casper_engine_test_support::{
+    DeployItemBuilder, ExecuteRequestBuilder, InMemoryWasmTestBuilder, ARG_AMOUNT,
+    DEFAULT_ACCOUNT_ADDR, DEFAULT_PAYMENT, DEFAULT_RUN_GENESIS_REQUEST,
+};
+use casper_types::{runtime_args, CLValue, Key, RuntimeArgs};
+use core::convert::TryFrom;
+
+const CONTRACT_SET_NAMED_KEY_STORED: &str = "set_named_key_stored.wasm";
```
nilz3ro, line 8: make these constant names less horrible

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -633,14 +639,23 @@ pub fn validate_query_proof(
     path: &[String],
```
luis-casper, line 650: The comment for this function needs to be fixed.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -652,7 +667,20 @@ pub fn validate_query_proof(
 
     let mut proof_value = first_proof.value();
 
-    for (proof, path_component) in proofs_iter.zip(path.iter()) {
+    for proof in proofs_iter {
```
luis-casper: The purpose of `proofs_iter2` isn't clear.
Also, why do we have this convoluted iteration stepping rather than zipping them together?
Explanations to this would fit nicely as code comments.

nilz3ro: I'll write this up in the comments but the reason I decided to use a secondary iterator is because when we are iterating through the proofs and we get to a proof of a `ContractPackage`, we need to skip it and assign the contract (which should always be the next proof in the iterator) to the `proof_value` variable.

Calling `iter.next()` from inside of a loop that consumes that iterator does not work, so I decided to use a copy.

nilz3ro: I can probably simplify this by indexing into `proofs` directly: `for 0..proofs_length { ... }`

dwerner (approved):

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -625,19 +631,46 @@ impl From<bytesrepr::Error> for ValidationError {
...
 /// Validates proof of the query.
 ///
-/// Returns [`ValidationError`] for any of
+/// Returns [`ValidationError`] in the following cases:
+/// - The number of proofs[^note] in `proofs` is different than the path length less one.
+/// - The first proof's [`Key`] does not match the `expected_first_key`.
+/// - Any proof's [`Key`] does not match its expected [`Key`].
+/// - The first proof's state root hash does not match `hash`.
+/// - The state root hash is invalid for any proof.
+/// - A expected named key cannot be found in a set of named keys for an [`Account`] or a
```
rafal-ch:
```suggestion
/// - An expected named key cannot be found in a set of named keys for an [`Account`] or a
```

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -625,19 +631,46 @@ impl From<bytesrepr::Error> for ValidationError {
...
+        for proof in proofs {
+            if let StoredValue::Contract(_) = proof.value() {
+                if let Some(StoredValue::ContractPackage(_)) = previous_proof_value {
+                    count += 1usize;
+                }
+            }
+            previous_proof_value = Some(proof.value().to_owned());
+        }
+        count
+    };
```
rafal-ch, line 666: Please check if these changes to `fn validate_query_proof()` don't require update of the `validate_query_proof_should_work()` test.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -652,7 +685,19 @@ pub fn validate_query_proof(
...
+                .get(index + 1)
+                .expect("expected a contract after this contract package.")
+                .value();
+            continue;
+        }
+
+        let path_component = path_iter
+            .next()
+            .expect("expected path component but none found.");
+
```
rafal-ch: Please check if these changes to `fn validate_query_proof()` don't require update of the `validate_query_proof_should_work()` test.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -625,19 +631,46 @@ impl From<bytesrepr::Error> for ValidationError {
...
     hash: &Digest,
     proofs: &[TrieMerkleProof<Key, StoredValue>],
     expected_first_key: &Key,
     path: &[String],
     expected_value: &StoredValue,
 ) -> Result<(), ValidationError> {
-    if proofs.len() != path.len() + 1 {
+    let proof_len_adjustment = {
+        let mut count = 0;
+        let mut previous_proof_value = None;
```
rafal-ch, line 655: Maybe we can avoid these two mutants by using smth like this:

```
    let proof_len_adjustment = {
        proofs
            .into_iter()
            .fold((0, None), |(acc, previous_proof_value), proof| {
                (
                    acc + match (proof.value(), previous_proof_value.as_ref()) {
                        (StoredValue::Contract(_), Some(StoredValue::ContractPackage(_))) => 1,
                        _ => 0,
                    },
                    Some(proof.value().to_owned()),
                )
            })
            .0
    };
```

But I'm not sure w/o unit test.

execution_engine_testing/tests/src/test/regression/gh_1931.rs
rafal-ch: Is the file `set_named_key_stored.wasm` missing from the PR? I can't run this test due to:
```
thread 'test::regression::gh_1931::main' panicked at '
Failed to open compiled Wasm file.  Tried the following locations:
    - /home/magister/Casper/casper-node/target/wasm32-unknown-unknown/release/set_named_key_stored.wasm
    - /home/magister/Casper/casper-node/execution_engine_testing/tests/wasm/set_named_key_stored.wasm
    - set_named_key_stored.wasm
```

rafal-ch: Unresolving due to the new comment for the same line.

execution_engine/src/core/tracking_copy/mod.rs
nilz3ro: Confirmed that this still passes.

execution_engine_testing/tests/src/test/regression/gh_1931.rs
nilz3ro: You need to `make build-contracts-rs` to make sure that the compiled WASM file exists.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -653,7 +685,19 @@ pub fn validate_query_proof(
 
     let mut proof_value = first_proof.value();
 
-    for (proof, path_component) in proofs_iter.zip(path.iter()) {
+    for (index, proof) in proofs_iter.enumerate() {
+        if let StoredValue::ContractPackage(_) = proof_value {
+            proof_value = proofs
+                .get(index + 1)
+                .expect("expected a contract after this contract package.")
```
xcthulhu: Don't expect here

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -653,7 +685,19 @@ pub fn validate_query_proof(
...
+                .value();
+            continue;
+        }
+
+        let path_component = path_iter
+            .next()
+            .expect("expected path component but none found.");
```
xcthulhu: Don't expect here either

xcthulhu (changes requested):
We can't introduce any `expect`s or `panic`s.

execution_engine/src/core/tracking_copy/mod.rs
```diff
@@ -626,19 +632,45 @@ impl From<bytesrepr::Error> for ValidationError {
...
+/// - The first proof's [`Key`] does not match the `expected_first_key`.
+/// - Any proof's [`Key`] does not match its expected [`Key`].
+/// - The first proof's state root hash does not match `hash`.
+/// - The state root hash is invalid for any proof.
+/// - An expected named key cannot be found in a set of named keys for an `Account` or a `Contract`.
+/// - The value returned by the query does not match the expected value.
+///
+/// [^note]: In cases where a query traverses through a `ContractPackage` and into a `Contract`
+/// using a named key, the `ContractPackage` will not be counted for path length validation.
 pub fn validate_query_proof(
```
xcthulhu, line 646: There should be no need to modify this function

xcthulhu (changes requested):
Upon further review this PR makes unnecessary changes to Merkle proof verification in general.  There should be a protocol for querying contract hashes where the Merkle proof verification code works as-is.

piotr-dziubecki:
@EdHastingsCasperLabs : working as intended.

Issue #2307, opened 2021-11-05
Fix incorrect condition and log message for protocol state batch sync


goral09:
When peer sends a request for batch protocol sync, it will specify what are the *next* sequence number it expects to receive for each validator. A receiving node will compare that with it's local view and response with either higher elements (if received more units than the sender) or a request for missing dependency (if lagging behind the sender). 

Currently, we would log an error if neither sender nor our node have any unit for a validator. That's b/c the *next sequence* is 0 and our local state would be empty. That's incorrect - if both our states (for a particular validator) are equivalent we shuold not respond with anything.

PR #2308, opened 2021-11-05
Fix invalid state batch sync condition

goral09:
Closes https://github.com/casper-network/casper-node/issues/2307

### DIFF
--- a/node/src/components/consensus/protocols/highway.rs
+++ b/node/src/components/consensus/protocols/highway.rs
@@ -629,6 +629,9 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
         their_next_seq: u64,
     ) -> Vec<HighwayMessage<C>> {
         let state = self.highway.state();
+        if our_next_seq == their_next_seq {
+            return vec![];
+        }
         if our_next_seq < their_next_seq {
             // We're behind. Request missing vertices.
             (our_next_seq..their_next_seq)
--- a/node/src/components/consensus/protocols/highway.rs
+++ b/node/src/components/consensus/protocols/highway.rs
@@ -668,7 +668,10 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
                         let ev = match state.maybe_evidence(vid) {
                             Some(ev) => ev.clone(),
                             None => {
-                                error!(?vid, instance_id=?self.highway.instance_id(), "panorama marked validator as faulty but no evidence was found");
+                                warn!(
+                                    ?vid, instance_id=?self.highway.instance_id(),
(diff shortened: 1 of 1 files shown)


node/src/components/consensus/protocols/highway.rs
```diff
@@ -648,19 +651,29 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
             // We're ahead.
             match state.panorama().get(vid) {
                 None => {
-                    error!(?vid, "received a request for non-existing validator");
```
afck, line 651: Shouldn't that be really impossible now, with the `==` check? (Same for the next error below.)

node/src/components/consensus/protocols/highway.rs
```diff
@@ -648,19 +651,29 @@ impl<I: NodeIdT, C: Context + 'static> HighwayProtocol<I, C> {
...
-                            "`IndexPanorama` doesn't match `state.panorama` for validator"
+                            "expected unit for validator but found none"
                         );
                         vec![]
                     }
+                    Observation::Faulty => {
+                        let ev = match state.maybe_evidence(vid) {
+                            Some(ev) => ev.clone(),
+                            None => {
+                                error!(?vid, instance_id=?self.highway.instance_id(), "panorama marked validator as faulty but no evidence was found");
```
afck:
```suggestion
                                error!(
                                    ?vid, instance_id = ?self.highway.instance_id(),
                                    "panorama marked validator as faulty but no evidence \
                                    was found"
                                 );
```

afck: This one might actualy be just a warning: What if they are marked as faulty based on evidence from the previous era?
(Ideally we'd send _that_ to them instead.)

afck (approved):
No blockers.

node/src/components/consensus/protocols/highway.rs
goral09: It should be impossible but I have to pattern match on it to avoid compiler errors. Also, I didn't want to leave it on `ERROR` as we have Prometheus alerts on this that wake up our SRE. Finally, it's only `WARN` as it's non-critical and its occurence can't stall (or crash) the network/node.

node/src/components/consensus/protocols/highway.rs
goral09: Do we change panorama's entries based on evidences from previous eras? I will dial the logging level down to WARN.

goral09:
bors merge

node/src/components/consensus/protocols/highway.rs
afck: Yes, we do. :+1:

PR #2310, opened 2021-11-05
Do not call auction::slash contract when no validator was slashed.

goral09:
This PR:
* removes `run_auction` boolean from the `StepRequest`. It was unnecessary as we always run it (it was always set to `true` as well).
* does not call `Auction::slash` endpoint if there are no validators to slash.

Closes https://github.com/casper-network/casper-node/issues/2309

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -2058,54 +2058,51 @@ where
             }
         }
 
-        if step_request.run_auction {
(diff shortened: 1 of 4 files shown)


Fraser999 (approved):

mpapierski (approved):
Looks good. We do also have #2316, where part of that work is a benchmark to measure auction improvements

goral09:
bors merge

PR #2311, opened 2021-11-05
move casper-engine-test-support crate out of monorepo

### LINKED ISSUE #2285: move casper-engine-test-support to own repo

Fraser999:
This PR removes the `casper-engine-test-support` crate from the monorepo, and replaces its usage with the published one now living in its own repo.

It also moves the `execution_engine_testing/tests` to `execution_engine_testing` inanticipation of `cargo-casper` also being moved out of the monorepo.

Finally, the version of `schemars` is pinned to 0.8.5 temporarily to avoid minor changes to the schema output caused by upgrading to 0.8.6.  Upgrading `schemars` will be addressed in a separate PR.

Closes #2285.


Cargo.toml
```diff
@@ -45,3 +43,11 @@ lto = true
 [profile.bench]
 codegen-units = 1
 lto = true
+
+# This section is required by any of the crates in this workspace which depend upon `casper-engine-test-support`.  It
+# causes `casper-engine-test-support` to depend upon these local copies of the three `casper-` crates rather than the
+# published ones, and hence avoids a compiler error where two versions of the same crate are consumed in a single crate.
+[patch.crates-io]
```
rafal-ch, line 50: Does this mean that we have these 3 crates published to `crates.io`, but we prefer to use to local copy? Are the published versions outdated?

rafal-ch (approved):
LGTM

PR #2318, opened 2021-11-08
Use `era_end_timestamp_millis` as seed

darthsiroftardis:
Co-authored-by: michal@casperlabs.io

CHANGELOG:

- Replaced `ProtocolVersion` in favor of `era_end_timestamp_millis` as the seed
- Added a regression test that calls `commit_step` twice, and ensures that there
are no common Key::Balance's with Write of 0. There shouldn't be common
keys in this scenario if the generator is seeded properly.

Closes Governance [89](https://app.zenhub.com/workspaces/engineering-60953fafb1945f0011a3592d/issues/casper-network/governance/89)


execution_engine_testing/tests/src/test/regression/step_regression.rs
```diff
@@ -0,0 +1,196 @@
...
+const ACCOUNT_1_BOND: u64 = 100_000_000;
+
+static ACCOUNT_2_PK: Lazy<PublicKey> = Lazy::new(|| {
+    let secret_key = SecretKey::ed25519_from_bytes([202; SecretKey::ED25519_LENGTH]).unwrap();
+    PublicKey::from(&secret_key)
+});
+const ACCOUNT_2_BALANCE: u64 = 200_000_000;
+const ACCOUNT_2_BOND: u64 = 200_000_000;
+
+fn initialize_builder() -> InMemoryWasmTestBuilder {
```
goral09: **[idea/not blocker]** Can we have this method somewhere in the `test/utils`? I've seen it used so many times already. It could accept a list of Genesis accounts as an argument.

execution_engine_testing/tests/src/test/regression/step_regression.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    assert_eq!(
+        balances_2.len(),
+        4,
+        "distribute should create 4 purses and zero out their balance"
+    );
+
+    let common_keys: BTreeSet<_> = balances_1.intersection(&balances_2).collect();
+    assert_eq!(common_keys.len(), 0, "there should be no commmon Key::Balance keys with Transfer::Write(0) in two distinct step requests");
+}
```
goral09: I haven't read the whole test but I'm guessing that what we're looking for is a list of execution effects in each of the step requests and then verify that they don't overlap. If that's true, then we should be able to extract this code to a single method `fn run_step(...) -> ExecutionEffects`, call it twice and compare the keys that were written.

goral09 (commented):
I think we can simplify the test with rather low cost.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1960,8 +1960,8 @@ where
         let mut named_keys = auction_contract.named_keys().to_owned();
         let gas_limit = Gas::new(U512::from(std::u64::MAX));
         let deploy_hash = {
-            // seeds address generator w/ protocol version
-            let bytes: Vec<u8> = step_request.protocol_version.value().into_bytes()?.to_vec();
+            // seeds address generator w/ era_end_timestamp_millis
+            let bytes: Vec<u8> = step_request.era_end_timestamp_millis.into_bytes()?.to_vec();
```
```suggestion
            let bytes = step_request.era_end_timestamp_millis.into_bytes()?;
```

execution_engine_testing/tests/src/test/regression/step_regression.rs
darthsiroftardis: @mpapierski Could you provide some comments here, and on the other comment Mateusz has made?

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1961,7 +1961,8 @@ where
         let gas_limit = Gas::new(U512::from(std::u64::MAX));
         let deploy_hash = {
             // seeds address generator w/ era_end_timestamp_millis
-            let bytes = step_request.era_end_timestamp_millis.into_bytes()?;
+            let mut bytes = step_request.era_end_timestamp_millis.into_bytes()?;
+            bytes.append(&mut step_request.next_era_id.into_bytes()?);
```
goral09, line 1966: I think this is unnecessary, I don't see how it's adding more entropy to the seed.

darthsiroftardis: Well, the EE could receive a `StepRequest` with the same `era_end_timestamp_millis` and it could also receive two `StepRequests` within the same `era_id`. This was we ensure that unless its the same `StepRequest` we will always have a safe seed

execution_engine_testing/tests/src/test/regression/step_regression.rs
mpapierski: Since TestBuilder is not usable without genesis it probably makes sense to have two methods: `InMemoryTestBuilder::default()` that calls `run_genesis` with `DEFAULT_RUN_GENESIS_REQUEST`, and `InMemoryTestBuilder::new(&custom_request)`. We already have a helper function to create a genesis request from supplied accounts.

execution_engine/src/core/engine_state/mod.rs
goral09: It's not possible to receive two `StepRequests` for the same era - we only have it when it's a switch block and we're guaranteed to have only one of these.

darthsiroftardis: Well thats how its currently implemented, you could imagine two nodes making `StepRequests` to the same EE instance given that the EE has no notion of the blockchain itself. This IMO makes the seed generation more robust.

goral09: Those two blockchains would have to have exactly same performance, down to a milliseconds of block productions since the genesis.

darthsiroftardis: Sure, what I am trying to say here is that how we pass in the `StepRequest` to the EE itself is part of how the node interacts with it, and that may/may not change. This adds additional variability in the seed.

@Fraser999 or @mpapierski do you guys have any opinions here?

goral09: It's still not guaranteed that adding a `next_era_id` increases security as there's no way to guarantee that it will be different from "the other `StepRequest`'s".

goral09: If you want to provide more randomness, use `step_request.pre_state_hash`. If that's _still_ equivalent then it should be fine b/c what we have is basically a function `f(x)` where `f` is the logic and `x` is the prestate hash. If both `f` and `x` are equivalent then the result also will be. Meaning it will be a no-op from the external pov.

darthsiroftardis: Yes, not by itself, however, the combination of both `next_era_id` and the timestamp for two different `StepRequests` is very unlikely to be the same.

Fraser999: I'm fine leaving the code as per authored here and continuing this discussion offline.

execution_engine_testing/tests/src/test/regression/step_regression.rs
Fraser999: Let's also look to action this idea outside of this PR please.  Seems worthwhile, but outside the scope of this PR.

Fraser999 (approved):

execution_engine_testing/tests/src/test/regression/mod.rs
```diff
@@ -37,3 +37,4 @@ mod gh_1902;
 mod regression_20210707;
 mod regression_20210831;
 mod regression_20210924;
+mod step_regression;
```
mpapierski: @darthsiroftardis, can you change this to `gov_89_regression` before merging? I wasn't aware of this would be fixed privately or not, so I came up with this generic name

mpapierski (approved):
Looks good to me only one nitpick to keep regression test names consistent

execution_engine_testing/tests/src/test/regression/mod.rs
darthsiroftardis: Understood!

Issue #2207, opened 2021-10-12
Do not warn about harmless dropped responders


piotr-dziubecki:
A dropped Responder is like an asynchronous function call that never returned. It's a decision whether to handle failures here or not, currently we log potentially confusing error messages even in cases where the dropping is most without meaningful consequence. While from a component's perspective, changing reactors is a "crash" of everything around it, this usually is not an issue if the component itself is also shutdown or pending requests are not otherwise especially "valuable".
This ticket is for adding a "global" (per-reactor, not truly global) variable that indicates we are considering dropped responders harmless at the moment, for example while shutting down the reactor. It can be stored on the Runner or similar as an Arc<AtomicBool> with a clone to each created EffectBuilder. Once shutting down, we can set the flag, having it change how the running request tasks are handling dropped responders.

Fraser999:
Please triage #2240 while working on this ticket to investigate whether the node in question was shutting down or not.  If it was, please close #2240 as a duplicate of this ticket.  If not, leave a comment to that effect on that ticket, since we'll need to investigate it further.

The triage should be fairly quick - if that turns out to not be the case, please let me know.

PR #2321, opened 2021-11-09
Downgrade loglevel of responder drops during shutdown

### LINKED ISSUE #2204: Fix misleading component error messages
Issue: Below error messages were reported under 2 different scenarios.

Scenario1: Small network 5Validator+2 Non, - post upgrade(1.3.4->1.4.0) one node was stuck and error messages as below reported - node had to be restarted to sync back, however in subsequent runs - unable to replicate.

Scenario2: Large network test(1.3.4->1.4.0)) to intentionally kill nodes(sigint) and restart to simulate data corruption, nodes were not corrupt and continued to work properly - but issued error messages as below - replication of issue pending

{"timestamp":"Oct 07 01:25:53.572","level":"ERROR","fields":{"message":"dropped without being responded to --- this is always a bug and will likely cause another component to be stuck!","responder":"Responder<casper_node::types::status_feed::StatusFeed<casper_node::types::node_id::NodeId>>"},"target":"casper_node::effect"}
{"timestamp":"Oct 07 01:25:53.572","level":"ERROR","fields":{"message":"request for channel closed, this may be a bug? check if a component is stuck from now on ","err":"oneshot canceled","queue_kind":"Api","channel":"\"casper_node::types::status_feed::StatusFeed<casper_node::types::node_id::NodeId>\""},"target":"casper_node::effect"}

https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1633602241187400
https://casperlabs-team.slack.com/archives/C01ULKL8G8J/p1633676668197200

marc-casperlabs:
This PR makes the node keep track of when it shuts down a reactor and adjusts the logging behavior accordingly, only outputting debug level messages if this occurs. Testing is unaffected and will still warn as before.

As a result, there should be no false positive error messages in the log due to shutting down a reactor related to requests/responders anymore.

Additionally, pushing items into a sealed queue has been downgraded to debug level logging as well.

Closes #2204, #2207.


node/CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ All notable changes to this project will be documented in this file.  The format
   * `[gossip][gossip_request_timeout]`
   * `[gossip][get_remainder_timeout]`
   * `[fetcher][get_from_peer_timeout]`
+* There are now less false warnings/errors regarding dropped responders or closed channels during a shutdown, where they are expected and harmless.
```
goral09, line 45: This line looks to me misaligned with the rest of the list.

node/src/utils.rs
```diff
@@ -120,6 +125,45 @@ pub(crate) fn leak<T>(value: T) -> &'static T {
...
+        SharedFlag(leak(AtomicBool::new(false)))
+    }
+
+    /// Checks whether the flag is set.
+    pub(crate) fn is_set(self) -> bool {
+        self.0.load(Ordering::SeqCst)
+    }
+
+    /// Set the flag.
+    pub(crate) fn set(self) {
```
goral09, line 146: Should we have `unset` method for completeness?

goral09 (commented):
Nice, clean. I'm wondering how does it affect general performance if we have this single flag that get's checked in multiple places.

node/src/utils.rs
mbr: No, I left it out for now, since these flags should be "set once", otherwise we may have to think about scenarios where some clients miss a set/unset cycle.

node/CHANGELOG.md
mbr: Only in this limited context, above is a sub-enumeration :)

```
* The following parameters in config now follow the humantime convention ('30sec', '120min', etc.):
  * `[network][gossip_interval]`
  * `[gossip][finished_entry_duration]`
  * `[gossip][gossip_request_timeout]`
  * `[gossip][get_remainder_timeout]`
  * `[fetcher][get_from_peer_timeout]`
* There are now less false warnings/errors regarding dropped responders or closed channels during a shutdown, where they are expected and harmless.
```

mbr:
> I'm wondering how does it affect general performance if we have this single flag that gets checked in multiple places.

Barely at all. We only check it if we have to log specific messages right now, which is a couple of times per day, at most. And even if that wasn't the case, I assume checking a single atomic boolean is fairly negligible.

node/CHANGELOG.md
goral09: Ah! Thanks :)

node/src/reactor.rs
```diff
@@ -713,16 +748,26 @@ where
                         break ReactorExit::ProcessShouldExit(ExitCode::Abort);
```
Fraser999, line 749: Shouldn't we call `self.is_shutting_down.set();` just before this line too?

marc-casperlabs: Yes, good point. I missed that occurrence, overall I tried not to accidentally add new logic to the code outside of the intended changes, but this was an oversight.

Fraser999 (approved):

PR #2331, opened 2021-11-11
Reject expired deploys sent from a client

### LINKED ISSUE #2330: DeployAcceptor should reject expired deploys sent from a client

darthsiroftardis:
CHANGELOG:

- Modified the DeployAcceptor to reject expired deploys received from a client.
- Added two tests to ensure the correct behavior

Closes #2330

### DIFF
--- a/node/src/components/deploy_acceptor.rs
+++ b/node/src/components/deploy_acceptor.rs
@@ -64,8 +64,13 @@ pub(crate) enum Error {
     },
 
     /// The deploy received by the node from the client has expired.
-    #[error("deploy received by the node expired at {0}")]
-    ExpiredDeploy(Timestamp),
+    #[error("deploy received by the node expired at {deploy_expiry_timestamp} with node's time at {node_current_timestamp}")]
+    ExpiredDeploy {
+        /// The timestamp when the deploy expires.
+        deploy_expiry_timestamp: Timestamp,
+        /// The timestamp when the node validated the expiry timestamp.
+        node_current_timestamp: Timestamp,
+    },
 }
 
 /// A representation of the way in which a deploy failed validation checks.
@@ -203,13 +208,17 @@ impl DeployAcceptor {
         }
 
         // We only perform expiry checks on deploys received from the client.
-        if source.from_client() && deploy.header().expired(Timestamp::now()) {
+        let node_current_timestamp = Timestamp::now();
+        if source.from_client() && deploy.header().expired(node_current_timestamp) {
             let time_of_expiry = deploy.header().expires();
             debug!(%deploy, "deploy has expired");
             return self.handle_invalid_deploy_result(
                 effect_builder,
                 EventMetadata::new(deploy, source, maybe_responder),
-                Error::ExpiredDeploy(time_of_expiry),
+                Error::ExpiredDeploy {
+                    deploy_expiry_timestamp: time_of_expiry,
+                    node_current_timestamp,
+                },
                 verification_start_timestamp,
             );
         }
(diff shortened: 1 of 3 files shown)


node/src/types/deploy.rs
```diff
@@ -1145,6 +1145,23 @@ impl Deploy {
         Self::random_transfer_with_session(rng, session)
     }
 
+    pub(crate) fn random_expired_deploy(rng: &mut TestRng) -> Self {
```
goral09, line 1148: I don't think there's a need to use randomized deploy generator for this particular test, especially that what we're testing - TTL expiration - is hardcoded.

node/src/components/deploy_acceptor.rs
```diff
@@ -62,6 +62,10 @@ pub(crate) enum Error {
         prestate_hash: Digest,
         failure: DeployParameterFailure,
     },
+
+    /// The deploy received by the node from the client has expired.
+    #[error("deploy received by the node from the client has expired")]
```
goral09: Let's mention deploy TTL somewhere - it should make it easier for the clients to understand what they did wrong and what to fix.

goral09 (approved):
No blockers.

node/src/types/deploy.rs
darthsiroftardis: Yes, but its just really trying to keep in step with the convention thats emerged around tests related for a DA.

goral09: We don't have to follow the convention if it's not helpful but only adds complexity.

Fraser999: I prefer using random (reproducible) data for unit tests like this, since it occasionally catches bugs which hard-coded values wouldn't.

goral09: I meant that the test we're adding here is only verifying that a too small TTL will be rejected and this particular property of the "random" deploys is hardcoded.

Fraser999 (approved):
Approval given that @goral09's comment on TTL being mentioned in the error is addressed.

node/src/types/deploy.rs
Fraser999: Yup - but aside from liking the randomness (even though the particular part being tested isn't random), I think we'd just be moving the complexity to the test, since it's not geared up at all just now to create a "real" deploy.

goral09: What I was suggesting was to remove the `random_` from the method name and the `rng` argument and just have a deploy that is already expired and use that in a test. Currently, this method might be giving a false impression of testing corner cases, multiple possible way a deploy can expire, etc. while in reality is not.

node/src/components/deploy_acceptor.rs
darthsiroftardis: The error message now includes the timestamp of expiry to let the user know when their deploy expired. @goral09 , is that okay?

darthsiroftardis: And the timestamp of the node when it checked the expiry

Fraser999:
bors merge

node/src/components/deploy_acceptor/tests.rs
```diff
@@ -1279,6 +1291,20 @@ async fn should_reject_deploy_with_mangled_transfer_amount() {
     ))
 }
 
+#[tokio::test]
+async fn should_reject_expired_deploy_from_client() {
+    let test_scenario = TestScenario::ShouldNotAcceptExpiredDeploySentByClient;
```
rafal-ch, line 1296: I'm too late to the party with my nit, but... :)
I'd prefer consistent naming, either:

```
should_reject_expired_deploy_from_client
ShouldRejectExpiredDeploySentByClient
```

or

```
should_not_accept_expired_deploy_from_client
ShouldNotAcceptExpiredDeploySentByClient
```

PR #2350, opened 2021-11-15
Do not set `no_std` when `std` feature is enabled for `casper-types`

xcthulhu:
Avoid setting `no_std` if the `std` feature is enabled for `casper-types`.

This is because we `derive(thiserror::Error)` on `bytesrepr::Error` when the `std` feature is set, and `thiserror::Error` relies on the standard library.

### DIFF
--- a/types/src/lib.rs
+++ b/types/src/lib.rs
@@ -1,7 +1,13 @@
 //! Types used to allow creation of Wasm contracts and tests for use on the Casper Platform.
 
 #![cfg_attr(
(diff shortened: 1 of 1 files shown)


Fraser999 (approved):

afck (approved):

xcthulhu:
bors try

xcthulhu:
bors try

xcthulhu:
bors merge

PR #2353, opened 2021-11-15
`retrieve-state` tool fanout to peers

dwerner:
This PR introduces some parallel downloading of tries in the `retrieve-state` tool, which reduces it's runtime significantly from about 10 hours to approximately 15 mins in my own case.


utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+                        engine_state
+                            .put_trie_and_find_missing_descendant_trie_keys(
+                                CorrelationId::new(),
+                                &trie,
+                            )
+                            .unwrap()
+                    })
+                    .await
+                    .unwrap();
+                    */
```
xcthulhu: Did you want to leave this comment in?

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1754,6 +1754,19 @@ where
         Ok(missing_descendant_trie_keys)
     }
 
+    /// Puts a trie.
+    pub fn put_trie(
```
xcthulhu: This function does not appear to be used...

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+        }
+
+        if let Some(msg) = base_recv.next().await {
+            match msg {
+                Ok(PeerMsg::TrieDownloaded {
+                    trie,
+                    peer,
+                    elapsed: _,
+                }) => {
+                    let engine_state = Arc::clone(&engine_state);
```
xcthulhu:
```suggestion
```

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+                    if let Some(counter) = in_flight_counters.get_mut(&peer) {
+                        *counter = counter.saturating_sub(1);
+                    }
+
+                    // count of all tries downloaded
+                    tries_downloaded += 1;
+
+                    if tries_downloaded % 1000 == 0 {
+                        println!(
+                            "tries downloaded: {downloaded}/{total}",
```
xcthulhu: Tries or MB / sec might be helpful here...

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+        .iter()
+        .map(|addr| (addr, 0usize))
+        .collect::<HashMap<_, _>>();
+
+    loop {
+        // Top level task keeps track of missing descendants and work distribution.
+        for (address, peer) in peer_map.iter() {
+            let counter = in_flight_counters.get_mut(&address).unwrap();
+            // TODO: understand what's happening when requests get backed up, causing starvation of
+            // base_recv
```
xcthulhu: Bit of a show-stopper... I get some serious hiccups because of this

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+                    trie_key: next_trie_key,
+                },
+            )
+            .await
+            {
+                Ok(GetTrieResult {
+                    maybe_trie_bytes: Some(blob),
+                    ..
+                }) => {
+                    let bytes: Vec<u8> = blob.into();
```
xcthulhu, line 664:
```suggestion
                    let bytes: Vec<u8> = blob.into();
                    if Digest::hash(bytes) != next_trie_key {
                        <...ban peer somehow...>
                        base_send
                        .send(Ok(PeerMsg::NoBytes {
                            trie_key: next_trie_key,
                            peer: address,
                        }))
                        .await
                        .expect("should send");
                    }
```

utils/retrieve-state/src/lib.rs
xcthulhu: Also, this is stalling for me... and it would be nice to have timestamps.

Maybe switch use logging like we have in the node/execution engine?

utils/retrieve-state/src/main.rs
```diff
@@ -79,14 +83,26 @@ struct Opts {
     use_gzip: bool,
 
     #[structopt(
-        short,
+        short = "b",
         long,
         about = "Specify the path to the folder to be used for loading block files to be converted."
```
xcthulhu: I thought we were using LMDB now?

It might be a little easier if we just had a `--data-dir`, and the subdirectories were the same as they are for the node.  I know a planned use case for this tool is to clone main-net for dry-run testing of upgrades...

utils/retrieve-state/src/main.rs
```diff
@@ -79,14 +83,26 @@ struct Opts {
     use_gzip: bool,
```
xcthulhu: I suggest we just do this by default rather than have a flag.

utils/retrieve-state/src/main.rs
```diff
@@ -195,20 +219,18 @@ async fn main() -> Result<(), anyhow::Error> {
                 fs::create_dir_all(&chain_download_path)?;
             }
             let highest_block: JsonBlock =
```
xcthulhu: This isn't used...

utils/retrieve-state/src/main.rs
```diff
@@ -79,14 +83,26 @@ struct Opts {
...
         long,
         about = "Specify the path to the folder to be used for loading block files to be converted."
     )]
     block_file_download_path: Option<PathBuf>,
 
-    #[structopt(short, long, about = "Enable manual syncing after each block to LMDB")]
-    manual_sync_enabled: bool,
+    #[structopt(long, about = "Enable manual syncing after each block to LMDB")]
+    manual_sync_enabled: Option<bool>,
+
```
xcthulhu, line 90: It might be a good idea to have an option for how many workers. But we'd have to drop the `tokio::main` macro and use the builder in that case.

utils/retrieve-state/src/main.rs
```diff
@@ -23,27 +23,34 @@ use retrieve_state::storage::create_storage;
...
+        about = "Specifies the host address for the node to be
+         reached. e.g. \"127.0.0.1\""
     )]
-    server_host: String,
+    server_ip: Ipv4Addr,
 
     #[structopt(
-        short,
+        long = "port",
+        about = "Specifies the port to reach peer RPC endpoints at."
```
afck: Should either be "Specify" or "Specifies" everywhere.
Or maybe better: None of them? Just "the port to reach peer RPC endpoints at"?

utils/retrieve-state/src/main.rs
```diff
@@ -217,14 +239,40 @@ async fn main() -> Result<(), anyhow::Error> {
...
+                    } else {
+                        // the peer's list gives us the port for the node, and instead we want to
+                        // use the user-provided RPC port
+                        SocketAddr::V4(SocketAddrV4::new(*address.ip(), port))
+                    };
+                    Some(address)
+                })
+                .collect::<Vec<_>>();
+
+            // also use the server we provided inititially that peers were gathered from.
```
afck:
```suggestion
            // also use the server we provided initially that peers were gathered from.
```

utils/retrieve-state/src/main.rs
```diff
@@ -217,14 +239,40 @@ async fn main() -> Result<(), anyhow::Error> {
...
-                retrieve_state::get_block(&mut client, &url, maybe_download_block)
+                retrieve_state::get_block(&client, &url, maybe_download_block)
                     .await?
                     .block
-                    .unwrap();
-            let progress = ProgressBar::new(highest_block.header.height);
-            println!(
+                    .expect("unable to download highest block");
+
+            eprintln!(
```
afck: We could also print the block hash here, so that a careful user can e.g. compare with a block explorer.

utils/retrieve-state/src/main.rs
```diff
@@ -233,18 +281,20 @@ async fn main() -> Result<(), anyhow::Error> {
...
-                &mut client,
-                &url,
-                &engine_state,
+                &client,
+                &peers_list,
+                engine_state,
                 highest_block.header.state_root_hash,
-                || progress.inc(1),
+                opts.concurrent_requests_per_peer
+                    .unwrap_or(DEFAULT_MAX_REQUESTS_PER_PEER),
```
afck: Would it be better to configure that default with `structop` above instead?

utils/retrieve-state/src/main.rs
```diff
@@ -233,18 +281,20 @@ async fn main() -> Result<(), anyhow::Error> {
                 lmdb_path,
                 opts.max_db_size
                     .unwrap_or(retrieve_state::DEFAULT_MAX_DB_SIZE),
-                opts.manual_sync_enabled,
-            )?;
+                opts.manual_sync_enabled.unwrap_or(true),
```
afck: Would it be better to call this `automatic_sync` and make it just a `bool` (not an option) instead? Then omitting it would make it `false`, i.e. manual sync.

utils/retrieve-state/src/lib.rs
```diff
@@ -43,9 +60,34 @@ pub enum GetBlockError {
...
+}
+
+/// Query global state for StoredValue
+pub async fn query_global_state(
+    client: &Client,
+    url: &str,
+    params: QueryGlobalStateParams,
+) -> Result<QueryGlobalStateResult, rpc::Error> {
+    let result = rpc::rpc(client, url, "query_global_state", params).await?;
+    Ok(result)
```
afck, line 82:
```suggestion
    rpc::rpc(client, url, "query_global_state", params).await
```

utils/retrieve-state/src/lib.rs
```diff
@@ -43,9 +60,34 @@ pub enum GetBlockError {
...
+}
+
+/// Query global state for EraInfo
+pub async fn get_era_info(
+    client: &Client,
+    url: &str,
+    params: GetEraInfoParams,
+) -> Result<GetEraInfoResult, rpc::Error> {
+    let result = rpc::rpc(client, url, "chain_get_era_info", params).await?;
+    Ok(result)
```
afck, line 92:
```suggestion
    rpc::rpc(client, url, "chain_get_era_info", params).await
```

utils/retrieve-state/src/lib.rs
```diff
@@ -70,10 +112,7 @@ pub async fn get_block(
 }
 
 /// Get the genesis block.
-pub async fn get_genesis_block(
-    client: &mut Client,
-    url: &str,
-) -> Result<GetBlockResult, rpc::Error> {
+pub async fn get_genesis_block(client: &Client, url: &str) -> Result<GetBlockResult, rpc::Error> {
```
afck, line 122: Why do we need that function? Shouldn't we always either specify a block or use the latest block?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+    #[error("BadData {trie_key:?} from {peer_address:?} due to {error:?}")]
+    BadData {
+        peer_address: SocketAddr,
+        trie_key: Digest,
+        error: bytesrepr::Error,
+    },
+}
+
+pub enum PeerMsg {
+    /// Download and save the trie under this
```
afck, line 376:
```suggestion
    /// Download and save the trie under this hash
```
?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+        peer: SocketAddr,
+        elapsed: Duration,
+    },
+    NoBytes {
+        trie_key: Digest,
+        peer: SocketAddr,
+    },
+}
+
 /// Download the trie from a node to the provided lmdb path.
```
afck, line 390:
```suggestion
/// Downloads the trie from the peers to the provided LMDB path.
```

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+        let base_send = base_send.clone();
+        let peer_future = async move {
+            let maybe_peer =
+                match maybe_construct_peer(base_send.clone(), *peer, client, state_root_hash).await
+                {
+                    Ok(peer) => Some(peer),
+                    Err(_err) => {
+                        // just skip those that can't be communicated with
+                        // println!("error constructing peer with address {:?} {:?}", peer,
+                        // err);
```
afck, line 410: Why not actually print that error here?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+                    .unwrap();
+                    */
+                    let mut missing_trie_descendants = engine_state
+                        .put_trie_and_find_missing_descendant_trie_keys(CorrelationId::new(), &trie)
+                        .unwrap();
+
+                    // count of all tries we know about
+                    total_tries_count += missing_trie_descendants.len();
+
+                    outstanding_trie_keys.append(&mut missing_trie_descendants);
```
afck, line 476: Not sure how likely that is in practice, but that could add a key for which a request is already in flight.

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+    TrieDownloaded {
+        trie: Trie<Key, StoredValue>,
+        peer: SocketAddr,
+        elapsed: Duration,
+    },
+    NoBytes {
+        trie_key: Digest,
+        peer: SocketAddr,
+    },
+}
```
afck, line 388: Is this a single message enum for both directions? Aren't the variants sent in one direction disjoint from the ones sent in the other direction? Could this be two separate enums?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
-            tries_downloaded += 1;
-            progress_fn()
-        } else {
-            return Err(anyhow::anyhow!(
-                "unable to download trie at {:?}",
-                next_trie_key
-            ));
+    println!("got {}/{} valid peers", peer_map.len(), peers.len());
+
+    let mut total_tries_count = 0;
```
afck, line 427: Why not also count the root right away, instead of adding `1` at the end?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+                    missing.len(),
+                    in_flight_counters,
+                );
+            }
+        }
+    }
+
+    for (_address, peer) in peer_map.drain() {
+        peer.shutdown.send(()).unwrap();
+        peer.join_handle.await.unwrap();
```
afck, line 583: Would it be a bit faster to first send to _all_ of them, and then await all of them?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+    address: SocketAddr,
+    client: Client,
+) -> Peer {
+    let (sender, peer_recv) = futures_channel::mpsc::unbounded::<PeerMsg>();
+    let peer_recv = peer_recv.map(|msg| Either::Left(msg));
+
+    let (shutdown, shutdown_recv) = futures_channel::oneshot::channel::<()>();
+    let shutdown_recv = shutdown_recv.into_stream().map(|_| Either::Right(()));
+
+    let mut next_peer_msg_or_shutdown_stream = futures::stream::select(peer_recv, shutdown_recv);
```
afck, line 641: Why not just a single stream, and a `Shutdown` message variant?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+            let mut base_send = base_send.clone();
+            let url = format!("http://{}:{}/rpc", address.ip(), address.port());
+            match get_trie(
+                &client,
+                &url,
+                GetTrieParams {
+                    trie_key: next_trie_key,
+                },
+            )
+            .await
```
afck, line 658: If every peer is only a single task and always immediately `await`s the result of `get_trie` before making the next request, are we really ever making multiple requests to the same peer at once, as suggested by `concurrent_requests_per_peer`? Or are we just putting `concurrent_requests_per_peer` items into the channel at once?

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,369 @@ pub fn put_block_with_deploys(
...
+                println!("state root has no missing descendants.");
+                println!(
+                    "tries downloaded: {downloaded}/{total}",
+                    downloaded = tries_downloaded,
+                    total = total_tries_count,
+                );
+                break;
+            } else {
+                println!(
+                    "state root still has {} missing descendants. {:#?}",
```
afck, line 573: Can this happen if `tries_downloaded == total_tries_count + 1`?

utils/retrieve-state/src/main.rs
xcthulhu: I suggest that this code avoid trusted RPC calls and instead require a trusted hash.  It'd then get blocks to current using finality signatures to verify them.

@dwerner WDYT?

(might be good to leave in the current behavior under an `--unsafe` flag)

utils/retrieve-state/src/main.rs
dwerner: This was a migration to load blocks from their old files into storage, and can be removed now.

execution_engine/src/core/engine_state/mod.rs
dwerner: Thanks, I was trying out a mode where I decoupled the missing check, but this is unused now. Removed.

utils/retrieve-state/src/lib.rs
dwerner: I was trying out spawn_blocking, but this actually led to some performance issues where we overuse the threadpool. Removed.

utils/retrieve-state/src/lib.rs
dwerner: Removed

utils/retrieve-state/src/lib.rs
dwerner: Stale comment, not related to the pauses we see where the trie is backed up on writes.

utils/retrieve-state/src/main.rs
dwerner: Done

utils/retrieve-state/src/main.rs
dwerner: You can do this now since the `highest_block` parameter is a `BlockIdentifier`. I'll add a note that asking for the highest block by height is considered insecure.

utils/retrieve-state/src/lib.rs
dwerner: I'm using the tries downloaded to determine when the downloading has completed, whereas if I counted it right away, I would lose that resolution.

utils/retrieve-state/src/lib.rs
dwerner: The idiom here is adapting a oneshot into a stream that is selected over with the response from the other stream. Warp uses this sort of api here https://docs.rs/warp/0.3.2/warp/struct.Server.html#method.serve_incoming_with_graceful_shutdown

utils/retrieve-state/src/lib.rs
dwerner: Correct, this is badly named and instead can be thought of as the `peer`'s mailbox in an actor model sense. I've renamed it appropriately.

utils/retrieve-state/src/lib.rs
dwerner: Logically, no. `tries_downloaded == total_tries_count + 1` is used here as the completion check, but we still rely on `missing_trie_keys` to tell us if we've downloaded the _correct_ tries. This shouldn't be a problem, but I wanted a check in place.

utils/retrieve-state/src/lib.rs
afck: But then why do we have the `concurrent_requests_per_peer` setting at all?
I think we _do_ want to make that many requests at the same time, instead of `await`ing here immediately?

utils/retrieve-state/src/main.rs
xcthulhu: I'd still prefer an `--unsafe` so they have to acknowledge they are doing something dangerous... like `kill -9` or `sudo`.

As discussed there is a security protocol for getting the highest block.  I'll see if I can quick, prototype something for you...

utils/retrieve-state/src/lib.rs
xcthulhu: Since the Merkle trie truly is a tree, I don't think that is possible @afck

utils/retrieve-state/src/lib.rs
afck: Is `peer_mailbox_size` really useful? I think we'd really want to make concurrent requests instead.

utils/retrieve-state/src/lib.rs
afck: Yes, I'm not suggesting adding the root to `tries_downloaded` right away, just to `total_tries_count`, so that the `tries_downloaded == total_tries_count + 1` check below doesn't need the `+ 1` anymore.

xcthulhu:
@afck Can you run this tool locally?

```bash
cargo run --release --bin retrieve-state -- -a download-tries -n 3.14.161.135
```

@dwerner says it should take 5 minutes...

goral09:
Let's extract [`d9e15bf` (#2353)](https://github.com/casper-network/casper-node/pull/2353/commits/d9e15bf8747aefdd9f311411e498492a6870e55c) from this PR to a separate one. It's not inherent to this ticket and should not be part of the changes. Extracting it to a separate PR will make it simpler to review, without having the `retrieve-state` tool context in mind.

xcthulhu:
> It's not inherent to this ticket and should not be part of the changes.

While it could be a separate PR as you say, the tool doesn't work on my machine without this...

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +337,380 @@ pub fn put_block_with_deploys(
...
+            }
+        }
+    }
+
+    for (_address, peer) in peer_map.drain() {
+        peer.shutdown.send(()).unwrap();
+        peer.join_handle.await.unwrap();
+    }
+
+    Ok(0)
```
xcthulhu:
```suggestion
    Ok(())
```

dwerner:
Update on this:

I've integrated @marc-casperlabs 's `work_queue` here now. I had to adapt it a bit but tried to stay true to what other parts are being used by fast-sync.

```
RUST_LOG=info cargo run --release --bin retrieve-state -- -a download-tries -n <node_ip_address>
```

utils/retrieve-state/src/lib.rs
```diff
@@ -714,3 +721,227 @@ fn spawn_peer(
...
+                    queue.push_job(child_job);
+                }
+            }
+            Ok(None) => {
+                warn!(
+                    "got no bytes back for trie, requeuing job and killing worker {}",
+                    worker_id
+                );
+                queue.push_job(*job.inner());
+                return;
```
afck, line 878: Why kill the worker in that case?

dwerner: In my tests, keeping the worker around led to a lot more attempts to request other tries from the same worker that failed with zero bytes.

afck: Ah, I just realized that each worker corresponds to one peer.
Isn't that inefficient? We should easily be able to make multiple parallel requests to the same peer.

dwerner: Have you tried running the tool?

resources/test/rest_schema_status.json
```diff
@@ -1,218 +1,221 @@
...
-        "type": "string"
-      },
-      "uptime": {
-        "description": "Time that passed since the node has started.",
-        "allOf": [
-          {
-            "$ref": "#/definitions/TimeDiff"
-          }
-        ]
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
Fraser999, line 2: It seems like the bulk of the change here is formatting, making spotting the actual diff fairly hard.

If the schema is generated by just using `serde_json::to_string_pretty` we can avoid the large diff.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1759,12 +1759,13 @@ where
         &self,
         correlation_id: CorrelationId,
         trie_keys: Vec<Digest>,
+        check_integrity: bool,
```
Fraser999, line 1793: We should document exactly what effect the `check_integrity` flag has.

execution_engine/src/storage/global_state/in_memory.rs
```diff
@@ -268,16 +268,22 @@ impl StateProvider for InMemoryGlobalState {
         &self,
         correlation_id: CorrelationId,
         trie_keys: Vec<Digest>,
+        check_integrity: bool,
```
Fraser999, line 272: Let's document the new arg please.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -219,6 +219,7 @@ impl StateProvider for LmdbGlobalState {
         &self,
         correlation_id: CorrelationId,
         trie_keys: Vec<Digest>,
+        check_integrity: bool,
```
Fraser999, line 223: Let's document the new arg please.

execution_engine/src/storage/global_state/mod.rs
```diff
@@ -114,6 +114,7 @@ pub trait StateProvider {
         &self,
         correlation_id: CorrelationId,
         trie_keys: Vec<Digest>,
+        check_integrity: bool,
```
Fraser999, line 118: Let's document the new arg please.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -238,6 +238,7 @@ pub fn missing_trie_keys<K, V, T, S, E>(
     txn: &T,
     store: &S,
     mut trie_keys_to_visit: Vec<Digest>,
+    check_integrity: bool,
```
Fraser999, line 242: The doc comment needs updated as point 2 is conditional now.

utils/retrieve-state/src/lib.rs
```diff
@@ -289,50 +344,604 @@ pub fn put_block_with_deploys(
...
+        if time_chunk.elapsed() > Duration::from_secs(1) {
+            let chunk_elapsed = time_chunk.elapsed().as_millis() as usize;
+            let bytes_per_sec = (bytes_downloaded_since_last_iteration / chunk_elapsed) * 1000;
+            let kbytes_per_sec = bytes_per_sec / 1024;
+            let timestamp = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap()
+                .as_millis();
+            info!(
+                "({timestamp}) tries: {downloaded}/{total} - {kbytes_per_sec} kbps",
```
Fraser999: `kbps` is generally an acronym for kilobits per second.  Can we rename all "bps" output strings and variables to be explicit please?

Fraser999 (commented):
The work queue is due to get unit tests added in the `feat-fast-sync` PR - let's incorporate that here too please.

The tool's `--help` command doesn't really explain all the various options very well.  I wouldn't block the PR for that, but it seems like this tool could get used a fair bit, and I'd really like better help output, or even a comprehensive README.

dwerner:
> The work queue is due to get unit tests added in the feat-fast-sync PR - let's incorporate that here too please.

@marc-casperlabs could you please port over the unit tests here when they land in `feat-fast-sync`? Should be a simple copy into this branch (which is based on dev btw).

> The tool's --help command doesn't really explain all the various options very well. I wouldn't block the PR for that, but it seems like this tool could get used a fair bit, and I'd really like better help output, or even a comprehensive README.

I agree, but following this PR I'd like to move this to a separate repo, so I'd like to prioritize that work first. @Fraser999

resources/test/rest_schema_status.json
dwerner: Done, hopefully it isn't too controversial that I added a file write of that `actual` value to disk, so that a quick copy can be performed, manually, once the diff is inspected.

execution_engine/src/storage/trie_store/operations/mod.rs
dwerner: @xcthulhu can you once-over my doc comments here to see if more detail is needed? I believe this came in from your PR to my branch to improve the performance.

Fraser999 (approved):
Optimistic approval given that the WorkQueue tests are ported over.

I only reviewed the production code changes in depth, so it'd be worth waiting until you get approval from someone more familiar with the tool than me.  I'm also fine to leave the help/readme improvements until a subsequent PR.

afck:
I tried it with testnet and get a throughput rate of ~260 kilobytes per second. I didn't wait for the full download to finish, though.

afck (approved):
Approving since none of my remaining comments are blockers.

PR #2366, opened 2021-11-17
Revert std feature behavior in casper-types

### LINKED ISSUE #2363: Revert std feature in casper-types
The `std` feature was deprecated in `casper-types` in 1.4.0, but this was accidentally partially reverted in 1.4.2, resulting in a broken crate: it does not compile with `std` enabled.

The issue is that enabling `std` causes the `bytesrepr::Error` enum to derive `thiserror::Error` which in turn depends upon the std library being available.  However, enabling the `std` feature does not cause `no_std` to be disabled, and thus the std library is unavailable.

We should revert to `std` being a no-op, with no support for deriving `thiserror::Error` for `bytesrepr::Error`, as per the API at 1.4.0 and 1.4.1.

Fraser999:
This PR fixes the broken `std` feature on `casper-types`, reverting to the behavior from 1.4.0 onwards.

It also includes a version bump for `casper-types` so that 1.4.3 can be published and 1.4.2 yanked.

Closes #2363.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -675,7 +675,6 @@ dependencies = [
  "serde_json",
  "serde_test",
  "strum",
- "thiserror",
  "uint",
  "version-sync",
 ]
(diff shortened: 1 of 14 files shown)


afck (approved):

afck (approved):

mbr (approved):
LGTM - should we yank the current version?

Fraser999:
bors merge

Issue #2369, opened 2021-11-17
Disable read-ahead on all LMDB storages.


cspramit:

goral09:
PR merged to 1.4.3

cspramit:
https://github.com/casper-network/casper-node-private/pull/16

PR #2372, opened 2021-11-17
Add no read ahead flag to EE and storage.

goral09:
Disables the read-ahead option in LMDB. Our access patterns are not sequential so the data that LMDB decided to cache in memory were usually a miss which caused a lot of page swapping and polluted the RAM unnecessarily.

Closes https://github.com/casper-network/casper-node/issues/2369


dwerner (approved):

goral09:
bors merge

PR #2374, opened 2021-11-18
flush_environment after step, upgrade and genesis

dwerner:


node/src/components/contract_runtime/operations.rs
```diff
@@ -112,6 +112,8 @@ pub fn execute_finalized_block(
                 finalized_block.timestamp().millis(),
                 finalized_block.era_id().successor(),
             )?;
+            // Flush if we ran commit_step
+            engine_state.flush_environment()?;
```
goral09: Wouldn't it be equivalent if we called `flush_environment()?` once, after this code block? We flush after deploy execution a couple of lines above this (before we enter into the auction code block).

dwerner: Done

goral09 (approved):

Fraser999 (approved):

dwerner:
bors merge

Issue #2070, opened 2021-09-13
Make fast-sync work with `nctl`


xcthulhu:
We want to test the following using something like the nctl upgrade tests:

* start a network using 1.4 binaries
* upgrade the network to binaries using fast-sync
* join a fresh fast-sync node using a post-upgrade trusted hash
* ~join a fresh 1.4 node after the upgrade (should sync to upgrade point then upgrade)~

rafal-ch:
Note: `1.5.0` is built from `feat_fast_sync` branch with two cherry-picked commits from the following PR:
https://github.com/casper-network/casper-node/pull/2323/commits

Clearly, the update from `1.4.1` to `1.5.0` doesn't work:
```
Activation point: 4
starting node version 1.5.0
creating era supervisor
node panicking: current_era [era 4] <= chainspec.activation_era() [era 4]
```

Debugging of the upgrade process continues...

rafal-ch:
The 1.5.0 node falls into the "Downgrade <-> Upgrade" loop:
```
starting node version 1.5.0
linear chain sync state: Syncing
run_fast_sync_task
joiner DowngradeVersion
starting node version 1.5.0
...
```
which, after ~250 cycles finishes with the panic mentioned in the comment above.

So, the `run_fast_sync_task` keeps returning `LinearChainSyncError::CurrentBlockHeaderHasOldVersion` until it finally returns the `Ok(block_header)`.

Checking what is the reason of `CurrentBlockHeaderHasOldVersion`...

rafal-ch:
The sequence that causes the loop:

- `is_current_era()` is invoked
  - it calculates the diff between `Timestamp::now()` and `era_start` from `trusted_key_block_info`
    - the `trusted_key_block_info.era_start` is taken from `block_header.timestamp()`
    - and the `block_header.timestamp()` comes from storage (read block header with `block_height_index.iter().last()`)
  - the diff is compared against `chainspec.core_config.era_duration`

So, we calculate the diff:
```
2021-11-10T11:30:37.828Z - 2021-11-10T11:29:59.296Z = 38s 532ms, which is less than era duration (41s)
    current_timestamp                era_start
```
The above causes the `check_block_version()` to return `CurrentBlockHeaderHasOldVersion` and trigger the downgrade.

We repeat the above every few milliseconds, and finally we get:
```
2021-11-10T11:30:40.396Z - 2021-11-10T11:29:59.296Z = 41s 100ms (more than era duration)
[...]

PR #2376, opened 2021-11-19
Add test that tests upgrade of joining node

rafal-ch:
Updated `upgrade_scenario_01.sh`, but it could potentially be a separate test scenario.


rafal-ch:
There will be another PR prepared for #2070.

Replacement PR: https://github.com/casper-network/casper-node/pull/2456

PR #2377, opened 2021-11-22
Add missing roundtrip serialization tests.

goral09:
This includes adding missing proptest strategies.

Closes https://github.com/casper-network/casper-node/issues/2382.

### DIFF
--- a/types/src/contracts.rs
+++ b/types/src/contracts.rs
@@ -1675,3 +1675,27 @@ mod tests {
         assert_eq!(contract_hash, decoded)
     }
 }
+
+#[cfg(test)]
+mod prop_tests {
+    use proptest::prelude::*;
+
+    use crate::{bytesrepr, gens};
+
+    proptest! {
+        // #![proptest_config(ProptestConfig {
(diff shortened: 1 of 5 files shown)


types/src/contracts.rs
```diff
@@ -1675,3 +1675,27 @@ mod tests {
...
 }
+
+#[cfg(test)]
+mod prop_tests {
+    use proptest::prelude::*;
+
+    use crate::{bytesrepr, gens};
+
+    proptest! {
+        // #![proptest_config(ProptestConfig {
```
mpapierski, line 1686: Do we need this?

mpapierski (approved):

types/src/contracts.rs
goral09: It's just very handy - easy to re-introduce to code and navigate into relevant `proptest` code.

goral09:
@Fraser999 , merging this but if you have any comments I will follow-up. It's a test code only so is not breaking anything that already exists.

goral09:
bors merge

PR #2383, opened 2021-11-22
GH-2398: Online map resize reintroduced

### LINKED ISSUE #2398: Online LMDB map resize of global state
We can resize the LMDB map before an attempt to write data is made and only when the space left is below a configurable threshold. Contrary to an old PR from old repo it’s a change much easier to test and does not affect other parts of EE.

mpapierski:
Closes #2398 

This PR reintroduces the idea from CasperLabs/CasperLabs#1740, which tries to resize the map once a commit returns a `MapFull` error, and then it will retry with the larger size until the transaction succeeds. Unfortunately, too many parts of the codebase were changed to be reviewed and tested.

In this PR, I'm introducing a more straightforward method where a global state map resizes happens before a write transaction starts and the free space in the map falls below the configured threshold. In this case, a map resizes by adding a fixed amount of configured bytes.

I also measured new calls to LMDB:

- A check to obtain map size info always falls below ~1000us
- Map resizing operation is proportional to the new map size

Defaults chosen are:

- initial map size: 750GiB
- grow map threshold: 100GiB
- grow map bytes: 150GiB

The above means once global state free space is below 100GiB, it is automatically resized by adding 150GiB.

### DIFF
--- a/execution_engine/src/storage/global_state/lmdb.rs
+++ b/execution_engine/src/storage/global_state/lmdb.rs
@@ -244,7 +244,8 @@ mod tests {
     use super::*;
     use crate::storage::{
         trie_store::operations::{write, WriteResult},
-        DEFAULT_TEST_MAX_DB_SIZE, DEFAULT_TEST_MAX_READERS,
+        DEFAULT_GROW_SIZE_BYTES, DEFAULT_GROW_SIZE_THRESHOLD, DEFAULT_TEST_MAX_DB_SIZE,
+        DEFAULT_TEST_MAX_READERS,
     };
 
     #[derive(Debug, Clone)]
@@ -309,12 +310,21 @@ mod tests {
     fn create_global_state(
         map_size: usize,
         max_readers: u32,
+        grow_size_threshold: usize,
+        grow_size_bytes: usize,
     ) -> (LmdbGlobalState, Digest, TempDir) {
         let temp_dir = tempdir().unwrap();
 
         let environment = Arc::new(
-            LmdbEnvironment::new(&temp_dir.path().to_path_buf(), map_size, max_readers, true)
-                .unwrap(),
+            LmdbEnvironment::new(
+                &temp_dir.path().to_path_buf(),
+                map_size,
+                max_readers,
+                true,
+                grow_size_threshold,
+                grow_size_bytes,
+            )
+            .unwrap(),
         );
         let trie_store =
             Arc::new(LmdbTrieStore::new(&environment, None, DatabaseFlags::empty()).unwrap());
@@ -328,8 +338,12 @@ mod tests {
     fn create_test_state() -> (LmdbGlobalState, Digest) {
         let correlation_id = CorrelationId::new();
 
-        let (ret, mut current_root, _temp_dir) =
-            create_global_state(DEFAULT_TEST_MAX_DB_SIZE, DEFAULT_TEST_MAX_READERS);
+        let (ret, mut current_root, _temp_dir) = create_global_state(
+            DEFAULT_TEST_MAX_DB_SIZE,
+            DEFAULT_TEST_MAX_READERS,
+            DEFAULT_GROW_SIZE_THRESHOLD,
+            DEFAULT_GROW_SIZE_BYTES,
+        );
 
         {
             let mut txn = ret.environment.create_read_write_txn().unwrap();
@@ -362,14 +376,20 @@ mod tests {
     fn should_not_run_out_of_space() {
         const BYTES_IN_MB: usize = 1024 * 1024;
         const INIT_MAP_SIZE: usize = BYTES_IN_MB * 5;
+        const GROW_THRESHOLD: usize = BYTES_IN_MB * 2;
+        const GROW_SIZE: usize = BYTES_IN_MB * 5;
 
         const TEST_PAIRS: usize = 1000;
         const STEP_BY: usize = 100;
 
         let test_pairs = create_n_test_pairs(TEST_PAIRS);
 
-        let (global_state, mut current_root, _temp_dir) =
-            create_global_state(INIT_MAP_SIZE, DEFAULT_TEST_MAX_READERS);
+        let (global_state, mut current_root, _temp_dir) = create_global_state(
+            INIT_MAP_SIZE,
+            DEFAULT_TEST_MAX_READERS,
+            GROW_THRESHOLD,
+            GROW_SIZE,
+        );
 
         for position in (0..test_pairs.len()).step_by(STEP_BY) {
             let mut txn = global_state.environment.create_read_write_txn().unwrap();
--- a/execution_engine/src/storage/global_state/lmdb.rs
+++ b/execution_engine/src/storage/global_state/lmdb.rs
@@ -236,10 +236,10 @@ impl StateProvider for LmdbGlobalState {
 #[cfg(test)]
 mod tests {
     use lmdb::DatabaseFlags;
-    use tempfile::tempdir;
+    use tempfile::{tempdir, TempDir};
 
     use casper_hashing::Digest;
-    use casper_types::{account::AccountHash, CLValue};
+    use casper_types::{account::AccountHash, CLValue, U256, U512};
 
     use super::*;
     use crate::storage::{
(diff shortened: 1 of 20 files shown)


execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -266,6 +267,29 @@ mod tests {
...
     }
 
+    fn create_n_test_pairs(n: usize) -> Vec<TestPair> {
+        let mut vec = Vec::with_capacity(n);
+
+        for i in 1..=n {
+            let key = {
+                let u256_val = U256::from(i);
+                let mut account_hash_bytes: [u8; 32] = Default::default();
+                u256_val.to_big_endian(&mut account_hash_bytes);
```
goral09, line 277: Out of curiosity: we use little endian everywhere so why did you choose to use big endian here?

execution_engine/src/storage/transaction_source/lmdb.rs
```diff
@@ -116,6 +155,39 @@ impl LmdbEnvironment {
...
+
+    fn resize_if_required(&self) -> Result<(), lmdb::Error> {
+        let env_info = self.info()?;
+        let stat = self.env.stat()?;
+        let size_used_bytes = stat.page_size() as usize * env_info.last_pgno();
+        let size_left_bytes = env_info.mapsize() - size_used_bytes;
+        if size_left_bytes <= self.grow_size_threshold {
+            let new_map_size = size_used_bytes + self.grow_size_bytes;
+            self.set_map_size(new_map_size)?;
+            debug!(
```
goral09, line 184: I think `info` here is fine, it should be happening very rarely but when it does happen it would be nice to know it. We could also measure the time it took and log it. Again, it should be happening very rarely so won't pollute the logs.

execution_engine/src/storage/transaction_source/lmdb.rs
```diff
@@ -131,7 +203,8 @@ impl<'a> TransactionSource<'a> for LmdbEnvironment {
         self.env.begin_ro_txn()
     }
 
-    fn create_read_write_txn(&'a self) -> Result<RwTransaction<'a>, Self::Error> {
+    fn create_read_write_txn(&'a self) -> Result<Self::ReadWriteTransaction, Self::Error> {
+        self.resize_if_required()?;
```
goral09, line 207: I don't think we should be resizing on every transaction creation. Moreover, it would check the DB size many times per second in our case and it's not needed. The buffer (and if necessary the resize) is big enough to do it less frequently.

resources/local/config.toml
```diff
@@ -326,6 +326,14 @@ verify_accounts = true
...
 #enable_manual_sync = false
 
+# Threshold for global state size that will trigger a resize upon next write transaction.
+#
+# Defaults to 107,374,182,400 == 100 GiB
+#grow_size_threshold = 107374182400
+
+# After global state will exceed `grow_size_threshold` bytes it will be resized by adding
+# extra space. Defaults to 161,061,273,600 == 150 GiB
+#grow_size_bytes = 161061273600
```
goral09, line 336: I think that for local tests it's far too much. Maybe 1/100th of it will be OK.

utils/nctl/sh/scenarios/configs/bond_its.config.toml
```diff
@@ -325,6 +325,14 @@ verify_accounts = true
...
 #enable_manual_sync = false
 
+# Threshold for global state size that will trigger a resize upon next write transaction.
+#
+# Defaults to 107,374,182,400 == 100 GiB
+#grow_size_threshold = 107374182400
+
+# After global state will exceed `grow_size_threshold` bytes it will be resized by adding
+# extra space. Defaults to 161,061,273,600 == 150 GiB
+#grow_size_bytes = 161061273600
```
goral09, line 335: Same for the CI tests - let's make it 1/100th of it.

utils/nctl/sh/scenarios/configs/gov96.config.toml
```diff
@@ -329,6 +329,14 @@ verify_accounts = true
...
 #enable_manual_sync = false
 
+# Threshold for global state size that will trigger a resize upon next write transaction.
+#
+# Defaults to 107,374,182,400 == 100 GiB
+#grow_size_threshold = 107374182400
+
+# After global state will exceed `grow_size_threshold` bytes it will be resized by adding
+# extra space. Defaults to 161,061,273,600 == 150 GiB
+#grow_size_bytes = 161061273600
```
goral09, line 339: And here.

utils/nctl/sh/scenarios/configs/itst13.config.toml
```diff
@@ -325,6 +325,14 @@ verify_accounts = true
...
 #enable_manual_sync = false
 
+# Threshold for global state size that will trigger a resize upon next write transaction.
+#
+# Defaults to 107,374,182,400 == 100 GiB
+#grow_size_threshold = 107374182400
+
+# After global state will exceed `grow_size_threshold` bytes it will be resized by adding
+# extra space. Defaults to 161,061,273,600 == 150 GiB
+#grow_size_bytes = 161061273600
```
goral09, line 335: Here.

utils/retrieve-state/src/lib.rs
```diff
@@ -27,6 +27,8 @@ pub const LMDB_PATH: &str = "lmdb-data";
 pub const CHAIN_DOWNLOAD_PATH: &str = "chain-download";
 pub const DEFAULT_MAX_DB_SIZE: usize = 483_183_820_800; // 450 gb
 pub const DEFAULT_MAX_READERS: u32 = 512;
+pub const DEFAULT_GROW_SIZE_THRESHOLD: usize = 386_547_056_640; // 360 gb
```
goral09, line 30: This will resize a DB by 360GB if it gets close to the limit within 50GB range. Why values so big?

goral09 (changes requested):
Let's revisit the defaults but we, mainly, I don't think we should be checking and resizing the DB on _every_ new transaction. IMO, we should expose an API in the `EngineState` and call it at the end of a block execution, after storing a deploy or a block. Maybe even in `flush_environment`.

marc-casperlabs:
Hmm, so I am not sure what this feature is for, because I may have an insufficient understanding how LMDB works. Here's my confusion explained though:

The map size of LMDB should have almost no effect except that it limits the size of the db -- but it can be safely set to an extremely large value. See the Python-LMDB[^1] docs:

> map_size:
>
>    Maximum size database may grow to; used to size the memory mapping. If database grows larger than map_size, an exception will be raised and the user must close and reopen Environment. On 64-bit there is no penalty for making this huge (say 1TB). Must be <2GB on 32-bit.
>
>    Note
>
>    The default map size is set low to encourage a crash, so users can figure out a good value before learning about this option too late.

So what are the drawbacks of setting a large (say, 100 TB map)? As far as I know, on some OS constellations there may be an issue if sparse files cannot be allocated on the filesystem, but I want to discount those, as I consider them out of scope.

The one advantage I can see of having a smaller limit is a more controlled crash if we run out of disk space - but it would be better to leave this up to the operator and only enable it by manually changing the config, as it is machine specific. This one advantage gets negated by an automatic increase in size. 

Overall, for this functionality to make any sense over just having a larger map, there would need to an advantage of having a smaller map that gradually grows bigger. But I can't see one, if there is one, I would very like to know, as mentioned above :).

If there is none though, we could consider just remove/not-merging this feature and changing the default settings in the configuration files to a larger limits, solving the problem. This would avoid adding a lot of code as well.

[^1]: I can't say how valid the information in the Python LMDB docs is, sadly, as I have not figured out where they get their information from. I will assume it is accurate at least for now.

mpapierski:
Thanks, @marc-casperlabs. Maybe the issue is simpler than we thought it was. I don't have a counterargument, and I don't know which systems support sparse files, but that shouldn't be an issue on Linux that we support officially.

Perhaps we should document that node operates should use a filesystem that supports sparse files - there aren't many https://en.wikipedia.org/wiki/Comparison_of_file_systems#Allocation_and_layout_policies. HFS+ doesn't have it from the most popular ones, but it's not our officially supported platform. And AFAIK, it's not a default filesystem on macOS anymore.

We can revisit this behavior when we get to support other OSes (i.e., macOS with HFS+ volumes)

marc-casperlabs:
Awesome, @mpapierski !

Well, technically it is not a must to support sparse files, if you're willing to allocate the entirety of the drive (or space you want to allow) upfront. If you are dedicating the machine to a casper-node or are okay with it crashing once that space is exhausted, that's fine, too.

So I see two use-cases:

* Production use - you will probably want to dedicate a large portion of the available drive space to the node and run almost nothing else on the machine. In this case, it doesn't matter if sparse files are supported or not, as your disk drive is the physical limit. If sparse files are supported, the size should be very large (using the default). If sparse files are not supported, the limit must be set to something that makes all storages fit on disk, as they will be preallocated.
* Test/local/development use - smaller databases are fine when not having to store the entire mainnet on a local node. If sparse file are supported, again, no changes from the defaults are necessary. If sparse files are _not_ supported, pick a reasonable size (a hundred megs?) for testing as the limit for each DB, and everything is fine as well. The node will crash in the same way as if it ran out of disk space if it hits the limit.

At the end of the day, running out of disk space means you cannot continue running the node :).

piotr-dziubecki:
@Fraser999 not required

PR #2385, opened 2021-11-22
Disable expensive LNC check.

afck:


goral09 (approved):

fizyk20 (approved):

node/src/components/consensus/highway_core/state.rs
```diff
@@ -52,6 +52,9 @@ use tallies::Tallies;
 // endorsements again.
 pub(super) const TODO_ENDORSEMENT_EVIDENCE_DISABLED: bool = true;
 
+// Disables checking the limited naivete criterion for endorsements.
```
GuybrushX, line 55: Typo or what is "naivete criterion"?

GuybrushX: I checked the Highway protocol PDF and it should be "naivetY" I guess.

afck: You're right, we should make it consistent. My English dictionary has both "naivety" and "naïveté", though.

afck:
bors merge

afck:
bors merge

afck:
bors merge

PR #2394, opened 2021-11-24
GH-2401: Deserialize a slice instead of allocating a vec.

### LINKED ISSUE #2401: Do not allocate temporary vector to deserialize data from Global State
We allocate a Vec from data read from lmdb just to consume it later in deserializer. Profiling 50k dictionary writes showed that about 7% probed time was spent in from_vec (which is the deserializing routine we used). After this simple change new flamegraph showed only 0.1% time for the same test run.

mpapierski:
Closes #2401 

This PR avoids allocating temporary Vec to deserialize data from the data returned by `lmdb::Transaction::get.`


goral09 (approved):
I support the change made in this PR but let's give @Fraser999 an opportunity to review as well.

types/src/bytesrepr.rs
```diff
@@ -139,6 +139,19 @@ pub fn deserialize<T: FromBytes>(bytes: Vec<u8>) -> Result<T, Error> {
     }
 }
 
+/// Deserializes slice of bbytes into an instance of `T`.
```
Fraser999:
```suggestion
/// Deserializes a slice of bytes into an instance of `T`.
```

types/src/bytesrepr.rs
```diff
@@ -139,6 +139,19 @@ pub fn deserialize<T: FromBytes>(bytes: Vec<u8>) -> Result<T, Error> {
...
+///
+/// Returns an error if the bytes cannot be deserialized into `T` or if not all of the input bytes
+/// are consumed in the operation.
+pub fn deserialize_slice<T: FromBytes>(bytes: &[u8]) -> Result<T, Error> {
+    let (t, remainder) = T::from_bytes(bytes)?;
```
Fraser999:
```suggestion
pub fn deserialize_from_slice<I: AsRef<[u8]>, O: FromBytes>(bytes: I) -> Result<O, Error> {
    let (t, remainder) = O::from_bytes(bytes.as_ref())?;
```

Fraser999 (commented):
It looks like we could use the new `deserialize_from_slice` function in the following places:
* [InMemoryEnvironment::dump()](https://github.com/casper-network/casper-node/blob/dev/execution_engine/src/storage/trie_store/operations/tests/mod.rs#L1009-L1010)
* [benchmark](https://github.com/casper-network/casper-node/blob/dev/types/benches/bytesrepr_bench.rs#L317) to avoid cloning
* [`CLValue::into_t()`](https://github.com/casper-network/casper-node/blob/dev/types/src/cl_value.rs#L92) to avoid allocating a new vec

types/src/bytesrepr.rs
```diff
@@ -144,6 +144,19 @@ pub fn deserialize<T: FromBytes>(bytes: Vec<u8>) -> Result<T, Error> {
     }
 }
 
+/// Deserializes a slice of bytes into an instance of `T`.
+///
+/// Returns an error if the bytes cannot be deserialized into `T` or if not all of the input bytes
+/// are consumed in the operation.
+pub fn deserialize_from_slice<I: AsRef<[u8]>, O: FromBytes>(bytes: I) -> Result<O, Error> {
```
Fraser999, line 151: It'd be good to add this to the `test_serialization_roundtrip` function:
```rust
    let deserialized =
        deserialize_from_slice::<_, T>(&serialized).expect("Unable to deserialize data");
    assert!(*t == deserialized);
```

Fraser999 (approved):
Optimistic approval given the final comment is addressed.

types/src/bytesrepr.rs
mpapierski: Added

PR #2396, opened 2021-11-24
Do not create temporary purses during distribute

### LINKED ISSUE #2384: Avoid creating temporary purses in Auction::distribute

darthsiroftardis:
CHANGELOG:

- Added a new entry point `mint_into_existing_purse` that transfers newly minted token into an existing purse without having to create a temporary purse

Closes #2384


types/src/system/auction/providers.rs
```diff
@@ -72,6 +72,12 @@ pub trait MintProvider {
         id: Option<u64>,
     ) -> Result<Result<(), mint::Error>, Error>;
 
+    /// Mint new token with given `initial_balance` balance into
+    /// a existing purse. Returns unit on success, otherwise
+    /// an error.
```
EdHastingsCasperAssociation: /// Mint `amount` new token into `existing_purse`.
    /// Returns unit on success, otherwise an error.

types/src/system/mint/mod.rs
```diff
@@ -154,4 +152,40 @@ pub trait Mint: RuntimeProvider + StorageProvider + SystemProvider {
...
+        }
+        if amount.is_zero() {
+            // treat as noop
+            return Ok(());
+        }
+        if self.read_balance(existing_purse)?.is_none() {
+            return Err(Error::PurseNotFound);
+        }
+        self.add_balance(existing_purse, amount)?;
+        // get total supply uref if exists, otherwise create it.
```
EdHastingsCasperAssociation: // get total supply uref if exists, otherwise error.

types/src/system/mint/mod.rs
```diff
@@ -34,10 +34,8 @@ pub trait Mint: RuntimeProvider + StorageProvider + SystemProvider {
             // get total supply uref if exists, otherwise create it.
```
EdHastingsCasperAssociation: // get total supply uref if exists, otherwise error.

EdHastingsCasperAssociation (approved):
see nits vis a vis comments

types/src/system/mint/mod.rs
```diff
@@ -31,10 +31,10 @@ pub trait Mint: RuntimeProvider + StorageProvider + SystemProvider {
         self.write_balance(purse_uref, initial_balance)?;
 
         if !is_empty_purse {
-            // get total supply uref if exists, otherwise create it.
+            // get total supply uref if exists, otherwise error
             let total_supply_uref = match self.get_key(TOTAL_SUPPLY_KEY) {
                 None => {
-                    // total supply URef should exist due to genesis
+                    // get total supply uref if exists, otherwise error
```
EdHastingsCasperAssociation: the original comment on line 37 was correct

types/src/system/auction/providers.rs
```diff
@@ -72,6 +72,11 @@ pub trait MintProvider {
         id: Option<u64>,
     ) -> Result<Result<(), mint::Error>, Error>;
 
+    /// Mint `amount` new token into `existing_purse`.
+    /// Returns unit on success, otherwise an error.s
```
rafal-ch:
```suggestion
    /// Returns unit on success, otherwise an error.
```

types/src/system/auction/error.rs
```diff
@@ -415,6 +425,21 @@ impl From<bytesrepr::Error> for Error {
...
+    fn from(error: mint::Error) -> Self {
+        match error {
+            mint::Error::Storage => Self::Storage,
+            mint::Error::MissingKey => Self::MissingKey,
+            mint::Error::CLValue => Self::CLValue,
+            mint::Error::Serialize => Self::Serialization,
+            mint::Error::ArithmeticOverflow => Self::ArithmeticOverflow,
+            mint::Error::GasLimit => Self::GasLimit,
+            mint::Error::InvalidContext => Self::InvalidContext,
+            _ => Self::MintError,
```
goral09: This _catch-all_ is usually always dangerous. Why do you think it's not here?

types/src/system/mint/mod.rs
```diff
@@ -31,13 +31,11 @@ pub trait Mint: RuntimeProvider + StorageProvider + SystemProvider {
...
-            // get total supply uref if exists, otherwise create it.
+            // get total supply uref if exists, otherwise error
             let total_supply_uref = match self.get_key(TOTAL_SUPPLY_KEY) {
                 None => {
-                    // create total_supply value and track in mint context
-                    let uref: URef = self.new_uref(U512::zero())?;
-                    self.put_key(TOTAL_SUPPLY_KEY, uref.into())?;
-                    uref
+                    // total supply URef should exist due to genesis
+                    return Err(Error::TotalSupplyNotFound);
```
goral09, line 38: Is this change related to the ticket? It's changing the logic that already exists.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -941,6 +941,24 @@ where
         ret
     }
 
+    /// Gets all `[Key::Balance]`s in global state.
+    pub fn get_balances(&mut self) -> Vec<Key> {
```
goral09:
```suggestion
    pub fn get_balance_keys(&mut self) -> Vec<Key> {
```

Otherwise, `get_balances` makes it sound like it woud return _balances_ (`U512`) but this method actually returns keys.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -941,6 +941,24 @@ where
...
+        let tracking_copy = self
+            .engine_state
+            .tracking_copy(state_root_hash)
+            .unwrap()
+            .unwrap();
+
+        let reader = tracking_copy.reader();
+
+        reader
+            .keys_with_prefix(correlation_id, &[KeyTag::Withdraw as u8])
```
goral09: The method says it will return balances' keys but here you're looking for a `Withdraw` key tag. There exists a `Key::Balance` tag so why not use it? I'm suspecting it's because you're actually not looking for balances but unbonding purses but if that's the case we should change the method name and its comment.

execution_engine/src/core/runtime/auction_internal.rs
```diff
@@ -215,6 +215,69 @@ where
         cl_value.into_t().map_err(|_| Error::CLValue)
     }
 
+    fn mint_into_existing_purse(
+        &mut self,
+        amount: U512,
+        existing_purse: URef,
+    ) -> Result<(), Error> {
+        if self.context.get_caller() != PublicKey::System.to_account_hash()
```
goral09, line 223: Should we even allow for a non-system account to call `mint` endpoints? I don't think so.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -3571,6 +3571,192 @@ fn should_increase_total_supply_after_distribute() {
...
+    let mut builder = InMemoryWasmTestBuilder::default();
+
+    builder.run_genesis(&DEFAULT_RUN_GENESIS_REQUEST);
+
+    // initial token supply
+    let initial_supply = builder.total_supply(None);
+
+    for request in post_genesis_requests {
+        builder.exec(request).commit().expect_success();
+    }
```
goral09, line 3703: We could add an utility method on `builder` to execute and commit a collection of requests (and expect all of them succeed). It could/should return execution results as well so that the caller can validate them further:
```rust
fn run_requests(&mut self, requests: Vec<ExecuteRequest>) -> Vec<ExecutionResult> { ... }
```

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -3571,6 +3571,192 @@ fn should_increase_total_supply_after_distribute() {
...
+    let post_genesis_supply = builder.total_supply(None);
+
+    assert_eq!(
+        initial_supply, post_genesis_supply,
+        "total supply should remain unchanged prior to first distribution"
+    );
+
+    // run auction
+    for _ in 0..5 {
+        builder.run_auction(timestamp_millis, Vec::new());
```
goral09, line 3714: Can you explain why we need to run auction exactly 5 times?

types/src/system/auction/error.rs
```diff
@@ -262,6 +263,13 @@ pub enum Error {
     /// assert_eq!(39, Error::ArithmeticOverflow as u8);
     /// ```
     ArithmeticOverflow = 39,
+    /// An error that is raised when there is an error in the mint contract that cannot
+    /// be mapped to a specific auction error.
+    /// ```
+    /// # use casper_types::system::auction::Error;
+    ///  assert_eq!(40, Error::MintError as u8);
+    /// ```
+    MintError = 40,
```
mpapierski: We should move GasLimit up and assign it a number. We did something similar for other system contract recently iirc

```suggestion
    #[doc(hidden)]
    GasLimit = 40,
    MintError = 41,
```

types/src/system/auction/error.rs
mpapierski: My suggestion would be to do `.map_err(|_| Error::MintError)` at the call time rather than this wildcard. I think there are just two places you're calling `mint_into_existing_purse.` You could also have two separate variants for each call for better error reporting

types/src/system/mint/mod.rs
darthsiroftardis: Not really, but I think its a valid change to the logic. We shouldn't be creating the URef again anywhere else cause we know it should be present due to genesis, this change prevents that any value `TOTAL_SUPPLY` named key to modified unintientionally

execution_engine_testing/test_support/src/wasm_test_builder.rs
darthsiroftardis: No, that was a foolish mistake on my part, it is meant to be `Balance` cause that way I can track in my test if any new purses are being created. I re-ran the test with this correction and it seems to be working as intended.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
Fraser999: Let's not do that in this PR please, to avoid scope creep.

goral09: It's not a scope creep, it's just a good engineering practice. Adding this method won't take more than 5 minutes.

types/src/system/auction/error.rs
Fraser999: I agree with @mpapierski - this is a blocker as otherwise we're breaking the API.

Fraser999 (changes requested):

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
darthsiroftardis: Following the convention I've seen in other tests

goral09: I see, so we don't know :)

types/src/system/auction/providers.rs
darthsiroftardis: Applied suggestion

types/src/system/auction/error.rs
darthsiroftardis: Applied Michal's suggestion

execution_engine_testing/test_support/src/wasm_test_builder.rs
darthsiroftardis: Applied suggestion

types/src/system/auction/error.rs
darthsiroftardis: Applied @mpapierski suggestion

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
darthsiroftardis: I'm with Fraser on this one, whilst it won't take much to create the function, we'd have to go through all the pre-existing tests which use the "old" convention and refactor them. I prefer we ticket it separately as a refactoring story

execution_engine/src/core/runtime/auction_internal.rs
Fraser999: I think we should address that in a separate PR too if required - this PR matches the existing functionality I believe.

Fraser999 (approved):

Issue #2352, opened 2021-11-15
Benchmark auction


Fraser999:
Specifically we want instrumentation of the `distribute()` and `run_auction()` functions.

PR #2397, opened 2021-11-24
GH-2352 - Add auction benchmark

dwerner:
Resolves GH-2352 - adds a benchmark for the auction. For now we benchmark with 250, 500, and 1000 delegators and run into memory issues if we go higher.

Can be run with something like 
```
cargo bench --bench auction_bench -- --save-baseline 'auction-bench-1' 'auction_bench'
```


goral09:
> For now we benchmark with 250, 500, and 1000 delegators and run into memory issues if we go higher.

I think we should control that via a test parameter or an env variable. Otherwise, we might kill CI/people's machines. Let's default to some reasonable value and allow for fine tuning it if necessary.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -104,6 +107,8 @@ pub struct WasmTestBuilder<S> {
     standard_payment_hash: Option<ContractHash>,
     /// Auction contract key
     auction_contract_hash: Option<ContractHash>,
+    /// Current "timestamp" for use with the auction.
+    timestamp: u64,
```
goral09: It looks like this is used only on the caller's side to get the next era end timestamp. It can be tracked outside of the `WasmTestBuilder` as a mutable variable in the context of a unit test.

dwerner: Done

dwerner:
> > For now we benchmark with 250, 500, and 1000 delegators and run into memory issues if we go higher.
> 
> I think we should control that via a test parameter or an env variable. Otherwise, we might kill CI/people's machines. Let's default to some reasonable value and allow for fine tuning it if necessary.

I've set this to 500 as a reasonable default for now.

dwerner:
> I've set this to 500 as a reasonable default for now.

Updated: I've  added support for many validators now.

goral09:
> > > For now we benchmark with 250, 500, and 1000 delegators and run into memory issues if we go higher.
> > 
> > 
> > I think we should control that via a test parameter or an env variable. Otherwise, we might kill CI/people's machines. Let's default to some reasonable value and allow for fine tuning it if necessary.
> 
> I've set this to 500 as a reasonable default for now.

I think you've changed it back again to run multiple, quite resource-heavy tests. Can we set a single, reasonable default and let users tune it?

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+};
+use casper_types::{
+    account::AccountHash,
+    runtime_args,
+    system::auction::{self, Bids},
+    Motes, ProtocolVersion, PublicKey, RuntimeArgs, SecretKey, U512,
+};
+
+const ARG_AMOUNT: &str = "amount";
+const ARG_TARGET: &str = "target";
```
goral09, line 32: I think we already have these as public constants somewhere, might as well import them here instead of redefining.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+const ARG_ID: &str = "id";
```
goral09, line 33: This argument is optional IIRC.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+
+const DELEGATION_AMOUNT: u64 = 42;
+const DELEGATION_RATE: u8 = 1;
+const DELEGATOR_INITIAL_BALANCE: u64 = 500 * 1_000_000_000u64;
+
+const VALIDATOR_BID_AMOUNT: u64 = 100;
+const TIMESTAMP_INCREMENT_MILLIS: u64 = 30_000;
+
+fn run_genesis_and_create_initial_accounts(
```
goral09, line 44: Please, add a short comment explaining what this function does in the context of the benchmark.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+                Motes::new(U512::from(VALIDATOR_BID_AMOUNT)),
+                DELEGATION_RATE,
+            )),
+            // None,
+        ),
+        GenesisAccount::account(
+            DEFAULT_PROPOSER_PUBLIC_KEY.clone(),
+            Motes::new(DEFAULT_ACCOUNT_INITIAL_BALANCE.into()),
+            None,
+        ),
```
goral09, line 62: Do we still need this account if later we initialize `validator_keys` accounts?

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+    ];
+    for validator in validator_keys {
+        genesis_accounts.push(GenesisAccount::account(
+            validator.clone(),
+            Motes::new(U512::MAX), // all the monies
```
goral09: I think we should not seed all of the accounts with `U512::MAX` as on any transfer we will overflow the type - transferring anything between two `U512::MAX` accounts will overflow immediately.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+    genesis_accounts: Vec<GenesisAccount>,
+) -> RunGenesisRequest {
+    let exec_config = {
+        let wasm_config = *DEFAULT_WASM_CONFIG;
+        let system_config = *DEFAULT_SYSTEM_CONFIG;
+        let auction_delay = DEFAULT_AUCTION_DELAY;
+        let locked_funds_period_millis = DEFAULT_LOCKED_FUNDS_PERIOD_MILLIS;
+        let round_seigniorage_rate = DEFAULT_ROUND_SEIGNIORAGE_RATE;
+        let unbonding_delay = DEFAULT_UNBONDING_DELAY;
+        let genesis_timestamp_millis = DEFAULT_GENESIS_TIMESTAMP_MILLIS;
```
goral09: We don't need these, right? We can pass the constants to `ExecConfig::new`.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+            *DEFAULT_ACCOUNT_ADDR,
+            runtime_args! {
+                    ARG_TARGET => *SYSTEM_ADDR,
+                    ARG_AMOUNT => MINIMUM_ACCOUNT_CREATION_BALANCE,
+                    ARG_ID => <Option<u64>>::None,
+            },
+        )
+        .build();
+        builder.exec(transfer);
+        builder.expect_success().commit();
```
goral09: Why do we need to fund the system account at all? Can't we have a single _faucet account_ in the Genesis (here it seems to be `DEFAULT_ACCOUNT_ADDR`) and reuse it for funding all other validators' and delegators' accounts?

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+    let active_bid = bids.get(&DEFAULT_ACCOUNT_PUBLIC_KEY.clone()).unwrap();
+    assert_eq!(
+        builder.get_purse_balance(*active_bid.bonding_purse()),
+        U512::from(VALIDATOR_BID_AMOUNT),
+    );
+    assert_eq!(
+        *active_bid.delegation_rate(),
+        1,
+        "unexpected delegation rate"
+    );
```
goral09: Do we still need to keep this assertions? If so, can you add message to the first one about what's expected and why?

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+        ret.push(public_key);
+    }
+    ret
+}
+
+fn step_and_run_auction(builder: &mut LmdbWasmTestBuilder, validator_keys: &[PublicKey]) {
+    let mut step_request_builder = StepRequestBuilder::new()
+        .with_parent_state_hash(builder.get_post_state_hash())
+        .with_protocol_version(ProtocolVersion::V1_0_0)
+        .with_reward_item(RewardItem::new(DEFAULT_ACCOUNT_PUBLIC_KEY.clone(), 1));
```
goral09: Is this still needed? We're iterating over all `validator_keys` and adding rewards to each. We can make all validators from `validator_keys` the _genesis validators_.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+            .motes()
+            .cloned()
+            .unwrap();
+
+        assert_eq!(U512::from(DELEGATOR_INITIAL_BALANCE), balance);
+
+        let delegation_amount = U512::from(DELEGATION_AMOUNT);
+        let delegator_account_hash = delegator_public_key.to_account_hash();
+        let delegate = {
+            let contract_hash = builder.get_auction_contract_hash();
```
goral09: This can be extracted out of the loop.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+                .with_empty_payment_bytes(
+                    runtime_args! { ARG_AMOUNT => U512::from(DEFAULT_DELEGATE_COST), },
+                )
+                .with_authorization_keys(&[delegator_account_hash])
+                .with_deploy_hash(deploy_hash)
+                .build();
+
+            ExecuteRequestBuilder::new().push_deploy(deploy)
+        }
+        .build();
```
goral09: Please extract this into a method.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+        builder.exec(delegate);
+        builder.expect_success();
+        builder.commit();
+        builder.clear_results();
+    }
+
+    let mut era_end_timestamp = TIMESTAMP_INCREMENT_MILLIS;
+
+    // TODO: use add_bid to add non-genesis validator with stake
```
goral09: That would be a separate test, correct? If so, let's create a ticket for it and remove the TODO.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    // advance the auction past the auction delay so that the added validator will be present in the
+    // auction
+    for _ in 0..DEFAULT_AUCTION_DELAY {
+        step_and_run_auction(&mut builder, &validator_keys);
+    }
```
goral09: If we add validators as Genesis ones, we won't need this step.

goral09 (commented):
Some comments that will hopefully improve the readability of the test and simplify it a bit.

execution_engine_testing/tests/Cargo.toml
```diff
@@ -44,6 +45,17 @@ bench = false
...
 
+[[bench]]
+name = "auction_bench"
+harness = false
+
+[[bin]]
+name = "auction_bench_bin"
+path = "benches/auction_bench.rs"
+test = false
+bench = true
```
Fraser999: Why do we need two benchmark targets here?  Seems like just the `[[bench]]` one would do?

execution_engine_testing/tests/Cargo.toml
```diff
@@ -24,6 +24,7 @@ rand = "0.8.3"
 serde = "1"
 serde_json = "1"
 tempfile = "3"
+criterion = "0.3.5"
```
Fraser999: I don't think we need this change if the `auction_bench_bin` binary target is removed from the manifest.

If we do need to keep that target, then please move this dep into correct alphabetical order in the list, and remove it from the `dev-dependencies` section.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    let mut genesis_accounts = vec![
+        GenesisAccount::account(
+            DEFAULT_ACCOUNT_PUBLIC_KEY.clone(),
+            Motes::new(U512::MAX), // all the monies
+            Some(GenesisValidator::new(
+                Motes::new(U512::from(VALIDATOR_BID_AMOUNT)),
+                DELEGATION_RATE,
+            )),
+            // None,
```
Fraser999: Left in by mistake?

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+    )
+}
+
+fn setup_bench_run_auction(
+    group: &mut BenchmarkGroup<WallTime>,
+    validator_count: usize,
+    delegator_count: usize,
+) {
+    // Setup delegator public keys
+    let delegator_keys = generate_pks(delegator_count);
```
Fraser999: Can we avoid all uses of the abbreviation `pk` for public keys please?  It's too easily confused with "private key" in my opinion.

execution_engine_testing/tests/benches/auction_bench.rs
```diff
@@ -0,0 +1,297 @@
...
+    let validator_keys = generate_pks(validator_count);
+
+    let data_dir = TempDir::new().expect("should create temp dir");
+    let mut builder = run_genesis_and_create_initial_accounts(
+        data_dir.path(),
+        &validator_keys,
+        delegator_keys
+            .iter()
+            .cloned()
```
Fraser999:
```suggestion
```

execution_engine_testing/tests/Cargo.toml
dwerner: The binary target allows debugging of the resultant bench as a binary, as well as independent analysis with other tools. In particular, vscode integrates with the bench and actually provides better tooling.
image

execution_engine_testing/tests/Cargo.toml
dwerner: I can remove this but I'm finding it essential for quicker iteration on our benches. In particular, outside of IDE support this improves the tooling around profiling the benches.

dwerner: I can remove this but I'm finding it essential for quicker iteration on our benches. In particular, outside of IDE support this improves the tooling around profiling the benches.

execution_engine_testing/tests/benches/auction_bench.rs
dwerner: Ah yes, my intention was to only seed the system account with MAX.

execution_engine_testing/tests/benches/auction_bench.rs
dwerner: Artifacts of inlining the method, I'll clean up the temporaries.

execution_engine_testing/tests/benches/auction_bench.rs
dwerner: If we don't fund the system account, when it goes to run the auction there is an error of `InsufficientPayment` :shrug:

execution_engine_testing/tests/benches/auction_bench.rs
dwerner: No we're good, I'll remove them.

execution_engine_testing/tests/benches/auction_bench.rs
dwerner: You're right, we don't need to do this now that we have many validators.

execution_engine_testing/tests/Cargo.toml
dwerner: Ok I've commented this out with a note and relocated the `criterion` dependency to a local one (also commented out).

dwerner:
> I think you've changed it back again to run multiple, quite resource-heavy tests. Can we set a single, reasonable default and let users tune it?

I think it's worthwhile to settle on testing within the -range- of reasonable values, i.e. where values are now and where they may go in the future. Running multiple tests here (like we do with transfer_bench) allows us to see how things may change over time, rather than assume performance has not regressed because a given change didn't show a regression.

I've chosen the following values to represent the auction as it will grow in delegators into the near future:
```
 [(100, 8000), (150, 8000), (100, 10000), (150, 10000)];
```

execution_engine_testing/tests/benches/auction_bench.rs
dwerner: Only ARG_AMOUNT is provided by casper-test-support.

execution_engine_testing/tests/Cargo.toml
dwerner: ...And I've now removed the target altogether.

Fraser999 (approved):

Fraser999:
bors merge

goral09:
> Running multiple tests here (like we do with transfer_bench) allows us to see how things may change over time, rather than assume performance has not regressed because a given change didn't show a regression.

Benches are not run in the CI anyway. They have to be run implicitly. In that case, I would expect a person to understand this fact and adjust the values to the most appropriate ones at the time of testing the change.

PR #2417, opened 2021-11-30
Merge remote-tracking branch 'upstream/dev' into feat-fast-sync

rafal-ch:
This PR integrates the recent changes from the `dev` branch into the `feat-fast-sync` branch.


afck (commented):
~Looks good apart from `StateStoreRequest` and `SharedObject`: I vaguely remember that we didn't want to reintroduce them, but I don't know the details.~
(Edit: We _do_ need `StateStoreRequest` because the block proposer now stores its state again!)
(Edit 2: And `SharedObject` is needed for in-memory message deduplication.)

afck (approved):
Looks good! Could use another review, though.

Fraser999 (approved):

node/src/effect.rs
```diff
@@ -76,7 +77,9 @@ use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use smallvec::{smallvec, SmallVec};
 use tokio::{sync::Semaphore, time};
-use tracing::error;
+#[cfg(not(feature = "fast-sync"))]
```
Fraser999: This should be removed.

rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2417/commits/2b7c5ef9fdba9ab09d0587bfba42c5b9be0353a0

Fraser999 (approved):

rafal-ch:
Merging manually because the `nctl-upgrade-test` will fail when trying to update to the node based on the `feat-fast-sync` branch.

marc-casperlabs (commented):
Checked only storage related things, those LGTM.

PR #2428, opened 2021-12-01
Verify fetched items in fast sync.

### LINKED ISSUE #2189: Verify fetched items in fast sync
Fast sync needs to verify that the deploys and blocks received from peers are cryptographically valid and match the ones that were requested.

afck:
Makes `Item::id` return a `Result`, so that cryptographically invalid blocks and deploys don't even _have_ an ID. In that case, the fetcher will disconnect from the sending peer.

This is [xcthulhu/fetcher-checks](https://github.com/xcthulhu/casper-node/tree/fetcher-checks) rebased on current `feat-fast-sync`.

Closes #2189.


node/src/types/block.rs
```diff
@@ -936,6 +939,18 @@ impl Display for BlockHeaderWithMetadata {
...
 }
 
+impl Item for BlockHeaderWithMetadata {
+    type Id = u64;
+    type ValidationError = BlockHeaderWithMetadataValidationError;
+    const TAG: Tag = Tag::BlockHeaderAndFinalitySignaturesByHeight;
+    const ID_IS_COMPLETE_ITEM: bool = false;
+
+    fn id(&self) -> Result<Self::Id, Self::ValidationError> {
+        validate_block_header_and_finality_signatures(&self.block_header, &self.block_signatures)?;
```
goral09: Isn't it a case that every node can be configured with a different FTT and still participate in the same network? If my local node is configured with stronger safety requirements then the weight of finality signatures that I expect will have to greater than sender's. In that case we might start disconnecting from peers we don't have to.

I said it in the past and will only repeat here - it should be a responsibility of the _requesting_ node to collect signatures until it satisfies his local safety requirements.

goral09:
Looks correct but didn't we reach a consensus on the approach in that we would add "decorators" for _not-yet-validated_ data and use that to discriminate? It seems like an arbitrary decision that an `id` method would perform validation. Also, there's nothing in the type system that prevents me from calling it repeatedly which could result in high CPU consumption - `id`/`name`/`value` are such a commonly used method names that having it perform any kind of validation is, in my opinion, counterintuitive. If we don't want to change the approach then maybe we could at least add a `fn validate(&mut self)` that could memoize the result and call that. It might not make a big difference in practice now but it has a couple of advantages over the current approach:
* it avoid possibly expensive computation by memoizing the results,
* method name makes it clear what it does.

node/src/types/block.rs
afck: This function doesn't check that it's _enough_ finality signatures, just that the ones received are valid and belong to this block.

goral09: Ah sorry, my mistake. There's another one that does it then.

goral09 (approved):

node/src/types/block.rs
afck: Yes, but at least that's not introduced in this PR. :sweat_smile:

afck:
But we also decided against introducing the decorator approach on the fast sync branch, because it's a large change and would make the diff even bigger.

Not sure about memoizing: So we'd add a `Cell<Option<BlockHash>>` field to `BlockHeaderWithMetadata` and `BlockWithMetadata`, and a `Cell<Option<DeployHash>>` to `Deploy`? I'm happy to do that, but I'll wait for other reviewers' opinions first.

goral09:
Ah, that sounds much worse when you said it than in my mind. No, let's not do _that_.

node/src/reactor.rs
```diff
@@ -925,3 +929,32 @@ where
...
+    R: Reactor,
+    <R as Reactor>::Event: From<fetcher::Event<T>> + From<BlocklistAnnouncement<NodeId>>,
+{
+    match fetcher::Event::<T>::from_get_response_serialized_item(sender, &serialized_item) {
+        Some(fetcher_event) => {
+            Reactor::dispatch_event(reactor, effect_builder, rng, fetcher_event.into())
+        }
+        None => {
+            info!(
+                "{} sent us a {:?} item we couldn't parse! Banning",
```
rafal-ch: Just a minor nit, but this might be a good opportunity to make the ban-related messages consistent, even if just for the sake of easier log parsing.

Please consider picking one of:
* "Banning"
* "banning"
* "banning peer"

afck: Done in https://github.com/casper-network/casper-node/pull/2428/commits/fb0a0d55d0ae7af73c86fcc21bda9e37ec9aff38.

rafal-ch (approved):
LGTM :+1:

fizyk20 (approved):

afck:
bors merge

PR #2431, opened 2021-12-01
Avoid passing named keys into call_host_ functions.

nilz3ro:
In this PR I have removed named keys from the signatures of the following functions:
* `Runtime::call_host_mint()`
* `Runtime::call_host_auction()`
* `Runtime::call_host_handle_payment()`

Now we look up and pass in a `Contract`  instance, which these functions take named keys from.

### DIFF
--- a/execution_engine/src/core/execution/executor.rs
+++ b/execution_engine/src/core/execution/executor.rs
@@ -176,10 +176,39 @@ impl Executor {
         let call_stack = runtime.call_stack().to_owned();
 
         if runtime.is_mint(base_key) {
+            let mint_contract_hash = match runtime.context().get_system_contract(MINT) {
+                Ok(contract_hash) => contract_hash,
+                Err(error) => {
+                    return ExecutionResult::Failure {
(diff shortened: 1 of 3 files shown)


darthsiroftardis (changes requested):
We shouldn't be passing in contract objects as an arg to the call methods. You could with this pass in any contract object. Instead the contract object should be retrieved inside the `call_host` methods and not via an arg

nilz3ro:
@darthsiroftardis I did it that way because [we already have a contract in scope here](https://github.com/casper-network/casper-node/blob/621abde06d69cd9a071b18558e7e771797d59c32/execution_engine/src/core/runtime/mod.rs#L2096), so if we pass in a contract, it'd only have to be looked up once in every case. If we encapsulate the logic that gets the contract, it'd be looked up every time a `call_host_` method is called, and we'd waste the contract that we already have in `execute_contract`.

I am happy to make the requested change, but I'd like to hear from @mpapierski to see what he thinks.

PR #2433, opened 2021-12-02
Avoid passing named_keys into runtime::call_host_ methods.

nilz3ro:
In this PR I have removed named keys from the signatures of the following functions:
* `Runtime::call_host_mint()`
* `Runtime::call_host_auction()`
* `Runtime::call_host_handle_payment()`


execution_engine/src/core/execution/executor.rs
```diff
@@ -676,7 +673,7 @@ impl DirectSystemContractCall {
         &self,
         mut runtime: Runtime<R>,
         protocol_version: ProtocolVersion,
-        named_keys: &mut NamedKeys,
+        _named_keys: &mut NamedKeys,
```
darthsiroftardis: Why not remove this argument if its unused?

execution_engine/src/core/runtime/mod.rs
```diff
@@ -57,6 +57,8 @@ use crate::{
 
 use self::host_function_flag::HostFunctionFlag;
 
+use super::tracking_copy::TrackingCopyExt;
```
darthsiroftardis: Import in the wrong place, can be grouped alongside with the other core imports

darthsiroftardis (approved):
Looks good

mpapierski (approved):
LGTM but you need to fix clippy issues @nilz3ro

Fraser999 (approved):

Issue #2338, opened 2021-11-12
Never shutdown with a downgrade signal in feat-fast-sync.


piotr-dziubecki:
If we encounter a block with a lower version we should just shut down with an error message.
Because if we are starting a new node we probably don't even have the previous version staged. And if we upgraded from the previous version it is an error to find a lower-version block after that.

PR #2434, opened 2021-12-03
Never shutdown for downgrade

rafal-ch:
This PR will cause the node to never shutdown for Downgrade.

Closes https://github.com/casper-network/casper-node/issues/2338


node/src/reactor/joiner.rs
```diff
@@ -518,8 +518,8 @@ impl reactor::Reactor for Reactor {
                                 block_header_with_old_version,
                             }) => {
                                 let old_version = block_header_with_old_version.protocol_version();
-                                info!(%current_version, %old_version, "restarting for downgrade");
-                                Some(JoinerEvent::Shutdown(ExitCode::DowngradeVersion))
+                                error!(%current_version, %old_version, "downgrade is not allowed, exiting");
+                                panic!("downgrade is not allowed");
```
afck: We should use the `fatal!` macro/effect instead of panicking.

rafal-ch: Fixed here: https://github.com/casper-network/casper-node/pull/2434/commits/634c0c8f110c69fae8cb782356874ae7ce4fd5c8

node/src/reactor/joiner.rs
```diff
@@ -518,8 +518,8 @@ impl reactor::Reactor for Reactor {
                                 block_header_with_old_version,
                             }) => {
                                 let old_version = block_header_with_old_version.protocol_version();
-                                error!(%current_version, %old_version, "downgrade is not allowed, exiting");
-                                panic!("downgrade is not allowed");
+                                fatal!(effect_builder, "downgrade is not allowed, exiting; current_version={}; old_version={}", current_version, old_version).await;
```
afck: If rustfmt doesn't want to, let's fix the formatting manually.

afck: I think "downgrade is not allowed" is hard to understand for the user.
Maybe something like: "network is still running an older version"?

rafal-ch: * rustfmt did format the code when I temporarily moved it out of the closure :)
* message has been changed

Both fixed in https://github.com/casper-network/casper-node/pull/2434/commits/246344895b1ea995ae917d35ffdeef1c1b084e30

afck (approved):

node/src/reactor/joiner.rs
```diff
@@ -518,8 +518,14 @@ impl reactor::Reactor for Reactor {
...
                             }) => {
                                 let old_version = block_header_with_old_version.protocol_version();
-                                info!(%current_version, %old_version, "restarting for downgrade");
-                                Some(JoinerEvent::Shutdown(ExitCode::DowngradeVersion))
+                                fatal!(
+                                    effect_builder,
+                                    "network is still running an older version, exiting; current_version={}; old_version={}",
+                                    current_version,
+                                    old_version
+                                )
```
afck: Actually, is this much different than the general `Err(error)` case below? Maybe the whole `match` arm can just be removed?
(Either way, would be good to get a second approval, too.)

rafal-ch: Well, good catch. The error will be formatted by the `fatal!` macro with slightly different message:
```
    #[error(
        "Current version is {current_version}, but current block header has older version: \
         {block_header_with_old_version:?}"
    )]
    CurrentBlockHeaderHasOldVersion {
        current_version: ProtocolVersion,
        block_header_with_old_version: Box<BlockHeader>,
    },
```

My preference would be to delete the match arm (`LinearChainSyncError::CurrentBlockHeaderHasOldVersion`) and rephrase the error message in the code block above.

afck: I'd also delete this `match` arm.
Maybe the `Err(error)` case below should just use the errors' `Display` impl, not `Debug`. Then it would print a readable message in every case?

rafal-ch: Yes, `Display` would be required.
@Fraser999 @marc-casperlabs what do you think?

node/src/components/linear_chain_sync/error.rs
```diff
@@ -76,7 +76,7 @@ pub(crate) enum LinearChainSyncError {
     },
 
     #[error(
-        "Current version is {current_version}, but current block header has older version: \
+        "Network is still running an older version. Current version is {current_version}, but current block header has older version: \
```
afck: Line too long

afck (approved):
Looks good, just one more nit-pick.

node/src/components/linear_chain_sync/error.rs
rafal-ch: Addressed in https://github.com/casper-network/casper-node/pull/2434/commits/f89f8e38fe57b187ece6b627a8e76045c93e61e2

node/src/types/exit_code.rs
```diff
@@ -18,9 +18,6 @@ pub enum ExitCode {
     /// The process should exit with `101`, equivalent to panicking.  The launcher should not
     /// restart the node.
     Abort = 101,
-    /// The process should exit with `102`.  The launcher should proceed to run the previous
-    /// installed version of `casper-node`.
-    DowngradeVersion = 102,
```
goral09, line 23: The exit codes are used to integrate node with the launcher. If we remove this exit code, will the launcher be updated as well to not handle it? Maybe we should leave a comment so that we don't use `102` as an exit code in the future? Or maybe we should simply leave it but mark as `allow(unused)` so that we don't try to use it in the future for other purposes.

rafal-ch: Addressed in: cc389c66a0864b8b3a89f4c2dc1648dc241cb7fc

The fact that this enum is `pub` makes it unnecessary for the `#allow(unused)` flag to be added. I marked it with `#[doc(hidden)]` though.

goral09: Awesome, thanks.

Fraser999 (approved):

PR #2436, opened 2021-12-03
Temporarily disable the `run_equivocator_network` test

rafal-ch:
This PR temporarily disables the `run_equivocator_network` test as it is giving the non-deterministic results.

The root-cause will be addressed in the following ticket: https://github.com/casper-network/casper-node/issues/1859


afck (approved):
Approving, but I hope we can consider re-enabling this a high priority.

Even if it's just with a workaround: E.g. we could do the test in a loop and retry until line 374 is satisfied.

Fraser999 (approved):

rafal-ch:
> Approving, but I hope we can consider re-enabling this a high priority.

Thanks!
I added a note in the ticket: https://github.com/casper-network/casper-node/issues/1859#issuecomment-985623956
We should discuss this during next week or on the next planning meeting at the latest.

PR #2440, opened 2021-12-06
Redelegate without unbonding

### LINKED ISSUE #2345: Add functionality to redelegate without unbonding

darthsiroftardis:
CHANGELOG:

- Added a `new_validator` field to `UnbondingPurses` which allows redelegation without unbonding
- Added a new struct `WithdrawPurse` to represent unbonding purses in current/live networks
- Added a new `Key::Unbond` under which `UnbondingPurses` are written
- Amended the upgrade logic to transform previous purses that are still be paid out into `UnbondingPurses`

Closes #2345


goral09:
FYI, I had to unassign myself as the reviewer as I won't be able to do a proper review before the end of 1.5.0 development cycle.

execution_engine/src/core/engine_state/error.rs
```diff
@@ -78,6 +78,21 @@ pub enum Error {
     /// Missing system contract hash.
     #[error("Missing system contract hash: {0}")]
     MissingSystemContractHash(String),
+    /// Failed to get the set of Key::Withdraw from global state.
+    #[error("Failed to get withdraw keys")]
+    FailedToGetWithdrawsKeys,
```
Fraser999:
```suggestion
    FailedToGetWithdrawKeys,
```

execution_engine_testing/tests/src/test/regression/ee_1119.rs
```diff
@@ -133,6 +133,7 @@ fn should_run_ee_1119_dont_slash_delegated_validators() {
             ARG_AMOUNT => U512::from(UNDELEGATE_AMOUNT_1),
             ARG_VALIDATOR => VALIDATOR_1.clone(),
             ARG_DELEGATOR => DEFAULT_ACCOUNT_PUBLIC_KEY.clone(),
+            ARG_NEW_VALIDATOR_PUBLIC_KEY => Option::<PublicKey>::None,
```
mpapierski: If the intention is for this argument to be optional (as not specifying it wouldn't raise an error), we need to handle it at `call_host_auction.`. I think it would be a good UX to add parameters and make it required - exchanges etc. may not be ready with changing this

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -399,6 +399,78 @@ where
             tracking_copy.borrow_mut().write(*key, value.clone());
         }
 
+        // Run through global state and process the withdraw purses
```
Fraser999: I think we should be a lot more explicit about what's actually being done here.  I view this as a mini data-migration which can probably be removed in the next release.

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -399,6 +399,78 @@ where
...
+        {
```
mpapierski, line 435: Can we add a big if clause only to run this when protocol_version is less or equal to `1.5.0`?  With current setup we'd have to either remove this code post mainnet activation, or risk running it on each future upgrade

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -90,6 +94,16 @@ static ACCOUNT_2_ADDR: Lazy<AccountHash> = Lazy::new(|| AccountHash::from(&*ACCO
 const ACCOUNT_2_BALANCE: u64 = MINIMUM_ACCOUNT_CREATION_BALANCE;
 const ACCOUNT_2_BOND: u64 = 200_000;
 
+static GENESIS_VALIDATOR_ACCOUNT_1_PK: Lazy<PublicKey> = Lazy::new(|| {
```
Fraser999: Let's avoid using the abbreviation "pk" for public key please.  I understand it already exists in a few places, but let's not add any more.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3058,3 +3080,727 @@ fn should_fail_with_more_accounts_than_slots() {
...
+
+    let delegators = bids[&NON_FOUNDER_VALIDATOR_2_PK].delegators();
+    assert_eq!(delegators.len(), 1);
+    let redelegated_amount_1 = *delegators[&BID_ACCOUNT_1_PK].staked_amount();
+    assert_eq!(redelegated_amount_1, U512::from(UNDELEGATE_AMOUNT_1));
+}
+
+#[ignore]
+#[test]
+fn should_upgrade_ubonding_purses_from_rel_1_4_2() {
```
Fraser999:
```suggestion
fn should_upgrade_unbonding_purses_from_rel_1_4_2() {
```

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3058,3 +3080,727 @@ fn should_fail_with_more_accounts_than_slots() {
...
+    let (mut builder, lmdb_fixture_state, _temp_dir) =
+        lmdb_fixture::builder_from_global_state_fixture(lmdb_fixture::RELEASE_1_4_2);
```
Fraser999, line 3284: Please describe the state captured in the `lmdb_fixture::RELEASE_1_4_2` database: i.e. that it has a single withdraw key in the unbonding queue at the upgrade point.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3058,3 +3080,727 @@ fn should_fail_with_more_accounts_than_slots() {
...
+    let delegators = bids[&GENESIS_VALIDATOR_ACCOUNT_1_PK].delegators();
+    assert_eq!(delegators.len(), 1);
+    let redelegated_amount_1 = *delegators[&DELEGATOR_2].staked_amount();
+    assert_eq!(redelegated_amount_1, U512::from(UNDELEGATE_AMOUNT_1));
+}
+
+#[ignore]
+#[test]
+fn should_transfer_to_main_purse_when_validator_is_no_longer_active() {
+    let (mut builder, lmdb_fixture_state, _temp_dir) =
```
Fraser999, line 3801: What do we gain by starting this test from the `lmdb_fixture::RELEASE_1_4_3`?  Seems like the previous test ensures that the unbonding queue is preserved across an upgrade, which is all we need from the hard-coded LMDB test fixture.

I'd think any further tests of the redelegate functionality should be run using a programmatically-generated fixture to make the test more future-proof and less opaque.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3058,3 +3080,727 @@ fn should_fail_with_more_accounts_than_slots() {
...
+        .get(&NON_FOUNDER_VALIDATOR_1_ADDR)
+        .expect("should have unbonding purse for non founding validator");
+    assert_eq!(unbond_list.len(), 3);
+    assert_eq!(
+        unbond_list[0].validator_public_key(),
+        &*NON_FOUNDER_VALIDATOR_1_PK
+    );
+    assert!(unbond_list[0].new_validator_public_key().is_none());
+    assert!(unbond_list[1].new_validator_public_key().is_none());
+    assert!(unbond_list[2].new_validator_public_key().is_none());
```
Fraser999: Seems like it would be better to have the LMDB fixture contain at least one unbond representing a redelegate request?

execution_engine_testing/tests/src/test/system_costs.rs
```diff
@@ -385,6 +385,7 @@ fn delegate_and_undelegate_have_expected_costs() {
             auction::ARG_DELEGATOR => DEFAULT_ACCOUNT_PUBLIC_KEY.clone(),
             auction::ARG_VALIDATOR => VALIDATOR_1.clone(),
             auction::ARG_AMOUNT => U512::from(BID_AMOUNT),
+            auction::ARG_NEW_VALIDATOR_PUBLIC_KEY => Option::<PublicKey>::None
```
Fraser999: Let's use `Some` here since I imagine that would be the more expensive choice if we ever actually make the costs that fine-grained.  Same for other similar change below in this file.

types/src/system/auction/constants.rs
```diff
@@ -32,6 +32,8 @@ pub const ARG_VALIDATOR_PURSE: &str = "validator_purse";
 pub const ARG_VALIDATOR_KEYS: &str = "validator_keys";
 /// Named constant for `validator_public_keys`.
 pub const ARG_VALIDATOR_PUBLIC_KEYS: &str = "validator_public_keys";
+/// Named constant for `new_validator_public_key`.
+pub const ARG_NEW_VALIDATOR_PUBLIC_KEY: &str = "new_validator_public_key";
```
Fraser999: Seems like this should be just `ARG_NEW_VALIDATOR` and `"new_validator"` to be consistent with the other delegate args.  `ARG_VALIDATOR_PUBLIC_KEY` is only used by `activate_bid` from what I can see.

smart_contracts/contracts/client/undelegate/src/main.rs
```diff
@@ -9,13 +9,20 @@ use casper_types::{runtime_args, system::auction, PublicKey, RuntimeArgs, U512};
 const ARG_AMOUNT: &str = "amount";
 const ARG_DELEGATOR: &str = "delegator";
 const ARG_VALIDATOR: &str = "validator";
+const ARG_NEW_VALIDATOR_PUBLIC_KEY: &str = "new_validator_public_key";
```
Fraser999: Should be `const ARG_NEW_VALIDATOR: &str = "new_validator";` to be consistent.

smart_contracts/contracts_as/client/undelegate/assembly/index.ts
```diff
@@ -1,14 +1,16 @@
...
+import {CLType, CLTypeTag, CLValue} from "../../../../contract_as/assembly/clvalue";
 import {RuntimeArgs} from "../../../../contract_as/assembly/runtime_args";
 import {Pair} from "../../../../contract_as/assembly/pair";
 import {PublicKey} from "../../../../contract_as/assembly/public_key";
+import {Option} from "../../../../contract_as/assembly/option";
 
 const ARG_AMOUNT = "amount";
 const ARG_DELEGATOR = "delegator";
 const ARG_VALIDATOR = "validator";
+const ARG_NEW_VALIDATOR_PUBLIC_KEY = "new_validator_public_key";
```
Fraser999: Should be `const ARG_NEW_VALIDATOR = "new_validator";` to be consistent.

types/src/gens.rs
```diff
@@ -455,14 +455,31 @@ fn withdraw_arb() -> impl Strategy<Value = UnbondingPurse> {
...
+fn unbonding_arb() -> impl Strategy<Value = UnbondingPurse> {
+    (
+        uref_arb(),
+        public_key_arb_no_system(),
+        public_key_arb_no_system(),
+        era_id_arb(),
+        u512_arb(),
+    )
+        .prop_map(|(bonding_purse, validator_pk, unbonder_pk, era, amount)| {
+            UnbondingPurse::new(bonding_purse, validator_pk, unbonder_pk, era, amount, None)
```
Fraser999: Let's not always have the `Option` as `None`.  Also, further instances of "pk" abbreviation.

execution_engine/src/core/runtime/auction_internal.rs
```diff
@@ -313,6 +313,68 @@ where
...
+                )
+                .map_err(|_| Error::TransferToDelegatorPurse)?
+                .map_err(|_| Error::TransferToDelegatorPurse)?;
+                delegator.increase_stake(amount)?;
+                *delegator.staked_amount()
+            }
+            None => {
+                let bonding_purse = self.create_purse().map_err(|exec_error| {
+                    <Option<Error>>::from(exec_error).unwrap_or(Error::CreatePurseFailed)
+                })?;
```
Fraser999:
```suggestion
                let bonding_purse = self.create_purse().map_err(|_| Error::CreatePurseFailed)?;
```

node/src/types/json_compatibility/stored_value.rs
```diff
@@ -44,7 +44,9 @@ pub enum StoredValue {
     /// A bid
     Bid(Box<Bid>),
     /// A withdraw
-    Withdraw(Vec<UnbondingPurse>),
+    Withdraw(Vec<WithdrawPurse>),
```
Fraser999, line 47: This is a breaking API change, but it could probably be deemed low impact enough to ignore.

types/src/system/auction/detail.rs
```diff
@@ -73,16 +73,16 @@ pub fn get_unbonding_purses<P>(provider: &mut P) -> Result<UnbondingPurses, Erro
 where
     P: StorageProvider + RuntimeProvider + ?Sized,
 {
-    let withdraws_keys = provider.get_keys(&KeyTag::Withdraw)?;
+    let withdraws_keys = provider.get_keys(&KeyTag::Unbond)?;
```
Fraser999:
```suggestion
    let unbond_keys = provider.get_keys(&KeyTag::Unbond)?;
```

types/src/system/auction/unbonding_purse.rs
```diff
@@ -37,13 +39,15 @@ impl UnbondingPurse {
...
         amount: U512,
+        new_validator_public_key: Option<PublicKey>,
     ) -> Self {
         Self {
             bonding_purse,
             validator_public_key,
             unbonder_public_key,
             era_of_creation,
             amount,
+            new_validator_public_key,
```
Fraser999: I think we should change the return type of this to `Option<Self>`, and have it return `None` if the `new_validator_public_key` is the same public key as the `validator_public_key`.

Then in `create_unbonding_purse()`, if `UnbondingPurse::new()` returns `None` we can exit early with `Ok(())` as the redelegation is effectively a no-op.

types/src/system/auction/detail.rs
```diff
@@ -228,23 +252,25 @@ pub(crate) fn create_unbonding_purse<P: Auction + ?Sized>(
     unbonder_public_key: PublicKey,
     bonding_purse: URef,
     amount: U512,
+    new_validator_public_key: Option<PublicKey>,
 ) -> Result<(), Error> {
     if provider.get_balance(bonding_purse)?.unwrap_or_default() < amount {
         return Err(Error::UnbondTooLarge);
     }
```
Fraser999, line 265: If we wanted to be even more efficient, we could leave `UnbondingPurse::new()` as is, and do the check for `validator_public_key` being the same public key as `new_validator_public_key` here.  That way, we'd exit having done the minimum amount of work, i.e. before creating the account hash and reading from global state.

types/src/system/auction/detail.rs
```diff
@@ -205,10 +205,34 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
             // current era id + unbonding delay is equal or greater than the `era_of_creation` that
             // was calculated on `unbond` attempt.
             if current_era_id >= unbonding_purse.era_of_creation() + unbonding_delay {
-                // Move funds from bid purse to unbonding purse
-                provider
-                    .unbond(unbonding_purse)
-                    .map_err(|_| Error::TransferToUnbondingPurse)?;
+                match unbonding_purse.new_validator_public_key() {
+                    Some(public_key) => {
```
Fraser999: Might be clearer to name this variable `new_validator`.

types/src/system/auction/detail.rs
```diff
@@ -228,23 +252,25 @@ pub(crate) fn create_unbonding_purse<P: Auction + ?Sized>(
...
         bonding_purse,
         validator_public_key,
         unbonder_public_key,
         era_of_creation,
         amount,
+        new_validator_public_key,
     );
     unbonding_purses.push(new_unbonding_purse);
-    provider.write_withdraw(validator_account_hash, unbonding_purses)?;
+    provider.write_unbond(validator_account_hash, unbonding_purses)?;
```
Fraser999, line 279: We maybe want to check that if `new_validator_public_key` is `Some` that it represents an actual validator.  It's not clear to me yet whether it would be possible for that check to succeed at the point it's made here, but by the time the unbonding is actioned, the specified new validator is no longer a validator, rendering the check here pointless.

types/src/system/auction/detail.rs
```diff
@@ -205,10 +205,34 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
...
             if current_era_id >= unbonding_purse.era_of_creation() + unbonding_delay {
-                // Move funds from bid purse to unbonding purse
-                provider
-                    .unbond(unbonding_purse)
-                    .map_err(|_| Error::TransferToUnbondingPurse)?;
+                match unbonding_purse.new_validator_public_key() {
+                    Some(public_key) => {
+                        let bid = provider
+                            .read_bid(&public_key.to_account_hash())?
+                            .ok_or(Error::ValidatorNotFound)?;
```
Fraser999: We need to be absolutely certain that this can't fail outside of some local data corruption or similar.  Same for the `provider.handle_delegation()` call immediately after this.

Seems like if any of these calls return errors we should log an error and continue.  If the `read_bid` call returns `Ok(None)` we should probably just unbond.

Currently if an invalid validator is specified in the `new_validator_public_key`, this will cause the `process_unbond_requests` to return an error which would cause none of the unbonding requests to be actioned, and the whole `commit_step()` to fail.

Fraser999 (changes requested):
There is one critical issue in `process_unbond_requests()` which must be resolved.

I'd like to see tests for
* redelegate to an invalid validator
* redelegate to the same validator (should be a no-op resulting in no unbonding purse entry)

The PR also needs to include an entry to each of the EE and types changelogs.

utils/validation/src/generators.rs
```diff
@@ -370,7 +372,7 @@ pub fn make_abi_test_fixtures() -> Result<TestFixtures, Error> {
         );
         stored_value.insert(
             "Withdraw".to_string(),
-            ABITestCase::from_inputs(vec![StoredValue::Withdraw(vec![
+            ABITestCase::from_inputs(vec![StoredValue::Unbonding(vec![
```
Fraser999: Let's rename this test case from `"Withdraw"` to `"Unbonding"` (in the line above here) and add a test case for `StoredValue::Withdraw` please.

types/src/system/auction/detail.rs
EdHastingsCasperAssociation: > Ok(None) we should probably just unbond.

correct

EdHastingsCasperAssociation: @darthsiroftardis fix this as soon as you can, thx

execution_engine_testing/tests/src/test/regression/ee_1119.rs
EdHastingsCasperAssociation: the intention is indeed for it to be optional

execution_engine/src/core/engine_state/mod.rs
EdHastingsCasperAssociation: we'll remove the migration after the fact; it only matters to extant network and means nothing to new networks

execution_engine/src/core/engine_state/mod.rs
EdHastingsCasperAssociation: i'd remove the one liner comment altogether and replace it with a link to a new issue to remove the block for upgrade after this one
// TODO: CRef={link to issue to remove this one time migration logic}

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
EdHastingsCasperAssociation: i think the various tests using a data capture from an earlier binary are proving that unbonds put into the queue BEFORE the upgrade is applied get processed correctly in the upgraded binary, and permutations around that basic idea

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
EdHastingsCasperAssociation: redelegate didn't exist in the older binary tho... 

or am i misunderstanding you?

execution_engine_testing/tests/src/test/system_costs.rs
EdHastingsCasperAssociation: better, do both

node/src/types/json_compatibility/stored_value.rs
EdHastingsCasperAssociation: not breaking, merely an extension

types/src/system/auction/unbonding_purse.rs
EdHastingsCasperAssociation: do what the signer(s) of the deploy said to do

execution_engine/src/core/engine_state/error.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
darthsiroftardis: Using `public_key` instead of `pk`

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
darthsiroftardis: Provided a description

execution_engine_testing/tests/src/test/system_costs.rs
darthsiroftardis: Both are now used and tested

types/src/system/auction/constants.rs
darthsiroftardis: Applied suggestion

smart_contracts/contracts_as/client/undelegate/assembly/index.ts
darthsiroftardis: Applied suggestion

types/src/gens.rs
darthsiroftardis: Removed instances of abbreviation and the test now creates an `Option` instead of just `None`

execution_engine/src/core/runtime/auction_internal.rs
darthsiroftardis: Applied suggestion

types/src/system/auction/detail.rs
darthsiroftardis: Applied suggestion

types/src/system/auction/detail.rs
darthsiroftardis: Renamed as such

types/src/system/auction/detail.rs
darthsiroftardis: The fix is present and a corresponding test has been added

utils/validation/src/generators.rs
darthsiroftardis: Renamed, and added a test case

execution_engine_testing/tests/src/test/regression/ee_1119.rs
mpapierski: Apologies for not making myself clearer in the comment above. It's a **required** optional argument and it requires all integrations who use direct auction contract calls to be updated with the new argument so it's backward incompatible. If we're fine with this we should communicate it better in the changelog and release notes to give them time to update.

Alternatively we can modify `call_host_auction` to ignore `ARG_NEW_VALIDATOR_PUBLIC_KEY` if it's not passed and it will be backwards compatible

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
Fraser999: I believe the snapshot in `lmdb_fixture::RELEASE_1_4_3` is generated from the code in this PR, so the `new_validator` field of `UnbondingPurse` should be available.

That highlights an issue whereby we should probably record a lot more info about the snapshots.  At the very least, the Git commit used, but also a copy of the code used for `post_genesis_setup` in `lmdb_fixture::generate_fixture` would be very helpful.

types/src/system/auction/detail.rs
Fraser999: Rename doesn't seem to have been committed?

resources/test/rpc_schema_hashing_V2.json
```diff
@@ -41,7 +41,7 @@
             "type": "object"
           },
           "AccountHash": {
-            "description": "Hex-encoded account hash.",
+            "description": "Checksummed hex-encoded account hash.",
```
Fraser999: Seems like many of the changes in this file are invalid.  However, turns out we're not running the test using this input file in CI unfortunately.

Please fix up this file by running
```
cargo t -p casper-node --lib components::rpc_server::tests::schema --no-default-features
```

Fraser999 (commented):
I made a few further comments, but the main blocker has been resolved now thanks.

resources/test/rpc_schema_hashing_V2.json
darthsiroftardis: Fixed file as requested

types/src/system/auction/detail.rs
darthsiroftardis: Renamed, correctly now

Fraser999 (approved):

goral09:
bors retry

types/CHANGELOG.md
```diff
@@ -13,6 +13,12 @@ All notable changes to this project will be documented in this file.  The format
 
 ## [Unreleased]
 
+### Changed
+* Extended `UnbondingPurses` to take a new field `new_validator` which represents the validator to whom tokens will be re-delegated.
```
goral09, line 17: You forgot to mention that we also changed:
* RPC API
* SSE data
* clients (in this PR only Rust client but the change forces the update to all the other SDKs)

cspramit:
> There is one critical issue in `process_unbond_requests()` which must be resolved.
> 
> I'd like to see tests for
> 
> * redelegate to an invalid validator
> * redelegate to the same validator (should be a no-op resulting in no unbonding purse entry)
> 
> The PR also needs to include an entry to each of the EE and types changelogs.

@Fraser999 @darthsiroftardis were these tests included?

PR #2442, opened 2021-12-06
Remove jemalloc, swap for stats_alloc simple wrapper

dwerner:
As @TomVasile has pointed out, we don't report the correct values for `consumed`, @thamps-casp has previously noted that `CL_MEM_DUMP_THRESHOLD` doesn't result in a memory dump, and we don't use the related metrics reported by the node. This does not affect our node-internal `DataSize` related memory metrics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -586,7 +586,9 @@ dependencies = [
  "signature",
  "smallvec",
  "static_assertions",
+ "stats_alloc",
  "structopt",
+ "sys-info",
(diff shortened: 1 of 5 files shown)


dwerner:
Adding this to 1.5.0 - I'd like to bring it in then or after.

marc-casperlabs (approved):
Looks good to me!

Fraser999 (approved):

PR #2452, opened 2021-12-09
Set up `enable_mem_deduplication` and `mem_pool_prune_interval` in co…

rafal-ch:
This PR adds the following parameters to the configs that require them and also sets the values as follows:
```
enable_mem_deduplication = true
mem_pool_prune_interval = 4096
```

### DIFF
--- a/resources/local/config.toml
+++ b/resources/local/config.toml
@@ -244,7 +244,12 @@ max_state_store_size = 10_737_418_240
 # Memory deduplication.
 #
 # If enabled, nodes will attempt to share loaded objects if possible.
-enable_mem_deduplication = false
+enable_mem_deduplication = true
+
+# Memory duplication garbage collection.
+#
+# Sets the frequency how often the memory pool cache is swept for free references.
+mem_pool_prune_interval = 4096
 
 
 # ===================================
--- a/utils/nctl/sh/scenarios/configs/bond_its.config.toml
+++ b/utils/nctl/sh/scenarios/configs/bond_its.config.toml
@@ -243,12 +243,12 @@ max_state_store_size = 10_737_418_240
 # Memory deduplication.
 #
 # If enabled, nodes will attempt to share loaded objects if possible.
-enable_mem_deduplication = false
+enable_mem_deduplication = true
 
 # Memory duplication garbage collection.
 #
 # Sets the frequency how often the memory pool cache is swept for free references.
-mem_pool_prune_interval = 1024
+mem_pool_prune_interval = 4096
 
 
 # ===================================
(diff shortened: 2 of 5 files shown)


resources/local/config.toml
```diff
@@ -244,7 +244,12 @@ max_state_store_size = 10_737_418_240
 # Memory deduplication.
 #
 # If enabled, nodes will attempt to share loaded objects if possible.
-enable_mem_deduplication = false
+enable_mem_deduplication = true
+
+# Memory duplication garbage collection.
+#
+# Sets the frequency how often the memory pool cache is swept for free references.
```
afck, line 251: Would be good to clarify that "5" means "every 5th time something is put in the pool the cache is swept". (@marc-casperlabs: Please confirm that's correct!)

utils/nctl/sh/scenarios/configs/bond_its.config.toml
```diff
@@ -243,12 +243,12 @@ max_state_store_size = 10_737_418_240
...
 #
 # If enabled, nodes will attempt to share loaded objects if possible.
-enable_mem_deduplication = false
+enable_mem_deduplication = true
 
 # Memory duplication garbage collection.
 #
 # Sets the frequency how often the memory pool cache is swept for free references.
-mem_pool_prune_interval = 1024
+mem_pool_prune_interval = 4096
```
afck, line 251: Or was it intentionally lower in the tests? (Maybe because we're running more nodes, which therefore already use up more memory?)

afck (approved):
Looks good to me, but please also wait for @marc-casperlabs's approval.

Fraser999 (approved):

resources/local/config.toml
marc-casperlabs: Yes, that's correct! :)

utils/nctl/sh/scenarios/configs/bond_its.config.toml
marc-casperlabs: I just don't know what a good value is yet. There's three things that this number mainly affects: Latency, CPU usage and memory usage.

Latency because higher numbers mean that the potential pause for pruning is longer the more items can be accumulated, so lower numbers mean less latency. On the other hand, if set too low, the CPU usage becomes closer to `O(n^2)`, which would be checking on every single request. Finally lower numbers are obviously better for memory usage.

To be honest, I think with these numbers, we can discount the memory usage as a non-issue - it is 8 or 16 bytes per entry currently, so 4096 is 64 KB of maximum memory usage. Latency wise, iterating over 4096 entries is also next to nothing, I think, and I believe the CPU time to be negligble as well. We could possibly go higher, but if 4096 checks the boxes, why not use it?

As for why this was at 1024 before, I believe it was lowered in tests mainly to have it triggered faster - remember, you have to request 4096 things first for the code to be run the first time :)

PR #2456, opened 2021-12-10
Add NCTL upgrade test for fast-sync nodes

rafal-ch:
This PR adds `nctl` test for testing upgrade from the "classic" nodes to the "fast sync" ones.

Scenario:
1. Start nodes 1-5 running version V1
2. Execute some deploys to populate global state a little
3. Upgrade nodes 1-5 to V2
4. Assert V2 nodes run & the chain advances (new blocks are generated)
5. Stage upgrade of node 6 to V2
6. Start the node 6, assert that it upgrades
7. Assert all nodes run & the chain advances (new blocks are generated)
8. Clean-up

Closes https://github.com/casper-network/casper-node/issues/2070


TomVasile (changes requested):
Blocking for now. Resolving questions OOB in slack.

rafal-ch:
Closing the PR as the test it adds will be replaced by the one prepared by @TomVasile here: https://github.com/TomVasile/casper-node/tree/nctl_fast_sync_testing

PR #2457, opened 2021-12-10
Reject incompatible versions during handshake

### LINKED ISSUE #2402: Make networking incompatible in fast sync branch
With fast sync, networking should terminate & block peers whose versions don't match.
Additionally we want to tarpit outdated nodes with low reconnection delays.

marc-casperlabs:
Closes #2402.

This adds an optional (enabled by default) rejection for incompatible versions after the handshake has been established. It can be disabled through the configuration to help unstick potentially sticky situation if the nodes are by accident still compatible.

It also adds a "tarpit" functionality that will delay processing of a handshake (and communication) with a node if it is on a very old version. As these old versions will only reconnect once all their connections are gone, even one node delaying them should be sufficient to stop them in their tracks and ease their impact on the network, specifically the known node subset.

### DIFF
--- a/node/src/components/small_network/tasks.rs
+++ b/node/src/components/small_network/tasks.rs
@@ -184,6 +184,8 @@ where
     pub(super) consensus_keys: Option<ConsensusKeyPair>,
     /// Weights to estimate payloads with.
     pub(super) payload_weights: PayloadWeights,
+    /// Whether or not to reject incompatible versions during handshake.
+    pub(super) reject_incompatible_versions: bool,
 }
 
 /// Handles an incoming connection.
@@ -359,7 +361,9 @@ where
         //
         // Since we are not using SemVer for versioning, we cannot make any assumptions about
         // compatibility, so we allow only exact version matches.
-        if protocol_version != context.chain_info.protocol_version {
+        if context.reject_incompatible_versions
+            && protocol_version != context.chain_info.protocol_version
+        {
             return Err(ConnectionError::IncompatibleVersion(protocol_version));
         }
 
--- a/node/src/components/small_network/tasks.rs
+++ b/node/src/components/small_network/tasks.rs
@@ -21,6 +21,7 @@ use openssl::{
     ssl::Ssl,
 };
 use prometheus::IntGauge;
+use rand::{rngs::OsRng, Rng};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 use tokio::{
     net::TcpStream,
@@ -190,6 +191,8 @@ where
     pub(super) tarpit_version_threshold: Option<ProtocolVersion>,
     /// If tarpitting is enabled, duration for which connections should be kept open.
     pub(super) tarpit_duration: Duration,
+    /// The chance, expressed as a number between 0.0 and 1.0, of triggering the tarpit.
+    pub(super) tarpit_chance: f32,
 }
 
 /// Handles an incoming connection.
@@ -370,10 +373,18 @@ where
         {
             if let Some(threshold) = context.tarpit_version_threshold {
                 if protocol_version <= threshold {
-                    // If tarpitting is enabled, we hold open the connection for a specific amount
-                    // of time, to reduce load on other nodes and keep them from reconnecting.
-                    info!(duration=?context.tarpit_duration, "tarpitting node");
-                    tokio::time::sleep(context.tarpit_duration).await;
+                    let mut rng = OsRng;
+
+                    let sample = rng.gen_range(0.0f32..1.0f32);
+                    if context.tarpit_chance > sample {
+                        // If tarpitting is enabled, we hold open the connection for a specific
+                        // amount of time, to reduce load on other nodes and keep them from
+                        // reconnecting.
+                        info!(duration=?context.tarpit_duration, "tarpitting node");
+                        tokio::time::sleep(context.tarpit_duration).await;
+                    } else {
+                        debug!(p = context.tarpit_chance, "randomly not tarpitting node");
+                    }
                 }
             }
             return Err(ConnectionError::IncompatibleVersion(protocol_version));
--- a/node/src/components/small_network/tasks.rs
+++ b/node/src/components/small_network/tasks.rs
@@ -375,8 +375,7 @@ where
                 if protocol_version <= threshold {
                     let mut rng = OsRng;
 
-                    let sample = rng.gen_range(0.0f32..1.0f32);
-                    if context.tarpit_chance > sample {
+                    if rng.gen_bool(context.tarpit_chance as f64) {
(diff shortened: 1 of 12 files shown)


node/src/components/small_network/tasks.rs
```diff
@@ -353,6 +362,34 @@ where
...
+        // compatibility, so we allow only exact version matches.
+        if context.reject_incompatible_versions
+            && protocol_version != context.chain_info.protocol_version
+        {
+            if let Some(threshold) = context.tarpit_version_threshold {
+                if protocol_version <= threshold {
+                    let mut rng = OsRng;
+
+                    let sample = rng.gen_range(0.0f32..1.0f32);
+                    if context.tarpit_chance > sample {
```
Fraser999: We could use [`gen_bool`](https://rust-random.github.io/rand/rand/trait.Rng.html#method.gen_bool):
```suggestion
                    if rng.gen_bool(context.tarpit_chance as f64) {
```

node/src/components/small_network/tasks.rs
```diff
@@ -353,6 +362,34 @@ where
...
+        // during upgrades where nodes may have a legitimate reason for differing versions.
+        //
+        // Since we are not using SemVer for versioning, we cannot make any assumptions about
+        // compatibility, so we allow only exact version matches.
+        if context.reject_incompatible_versions
+            && protocol_version != context.chain_info.protocol_version
+        {
+            if let Some(threshold) = context.tarpit_version_threshold {
+                if protocol_version <= threshold {
+                    let mut rng = OsRng;
```
Fraser999:
```suggestion
                    let mut rng = crate::new_rng();
```

rafal-ch (commented):
LGTM, but can we prove the new mechanism is working correctly by covering it with a test?

mbr:
> LGTM, but can we prove the new mechanism is working correctly by covering it with a test?

That's a bit tough, since we'd need to run an older version in parallel? The handshake negotiation also requires an actual connection, as it has not been factored out, so tests solely for the returned error are also tricky :/

node/src/components/small_network/tasks.rs
mbr: TIL! Thanks.

rafal-ch:
> > LGTM, but can we prove the new mechanism is working correctly by covering it with a test?
> 
> That's a bit tough, since we'd need to run an older version in parallel? The handshake negotiation also requires an actual connection, as it has not been factored out, so tests solely for the returned error are also tricky :/

Consider testing it on the UT level and request a ticket for some more sophisticated `nctl` test?

You may feed some dummy `context` and `remote_handshake` into the `fn negotiate_handshake()`. Or better, extract the `if` that starts at line `371` to a separate function and just test this one, as we can assume that the rest of `fn negotiate_handshake()` works correctly.

Anyway, leaving it up to you to decide.

node/src/components/small_network/tasks.rs
```diff
@@ -353,6 +362,33 @@ where
...
+        {
+            if let Some(threshold) = context.tarpit_version_threshold {
+                if protocol_version <= threshold {
+                    let mut rng = crate::new_rng();
+
+                    if rng.gen_bool(context.tarpit_chance as f64) {
+                        // If tarpitting is enabled, we hold open the connection for a specific
+                        // amount of time, to reduce load on other nodes and keep them from
+                        // reconnecting.
+                        info!(duration=?context.tarpit_duration, "tarpitting node");
```
afck:
```suggestion
                        info!(duration = ?context.tarpit_duration, "tarpitting node");
```

node/src/components/small_network/tasks.rs
```diff
@@ -353,6 +362,33 @@ where
...
+                        tokio::time::sleep(context.tarpit_duration).await;
+                    } else {
+                        debug!(p = context.tarpit_chance, "randomly not tarpitting node");
```
afck, line 385: I'd add the `randomly` ~`not`~ to the other message, too, or remove it from this one.

Fraser999 (approved):

marc-casperlabs:
> Consider testing it on the UT level and request a ticket for some more sophisticated nctl test?

nctl is probably the better option.

> You may feed some dummy context and remote_handshake into the fn negotiate_handshake(). Or better, extract the if that starts at line 371 to a separate function and just test this one, as we can assume that the rest of fn negotiate_handshake() works correctly.

It's annoying to get a transport instantiated though. But I think we can factor out the decision-making process once the handshake has been received. I'll try that.

Edit: After the required refactoring I feel like the actual unit-testable logic is so small that it becomes almost impossible to get wrong. On the other hand, the IO part is still untested and much more likely to contain bugs. So we will have to defer to nctl.

afck (approved):

node/src/components/small_network/tasks.rs
marc-casperlabs: We usually do no spaces around these arguments, as far as I can tell. However, for some reason `rustfmt` reformats the `debug!` message, but not the `info!` message :angry:

node/src/components/small_network/tasks.rs
marc-casperlabs: But we are tarpitting it on one instance and not in the other? I'm not sure I understand this suggestion :)

node/src/components/small_network/tasks.rs
afck: I've seen both styles; not sure whether we've agreed on any of them.
But the `tracing` docs do use spaces.

node/src/components/small_network/tasks.rs
afck: Ah, sorry, not the `not`, of course, just the `randomly`. :smile: 
Anyway, certainly not a blocker.

rafal-ch:
> On the other hand, the IO part is still untested and much more likely to contain bugs. So we will have to defer to nctl.

ACK.

Can you request appropriate ticket so we don't forget to address this in the future?

node/src/components/small_network/tasks.rs
goral09: Agree with Andreas - let's keep the messages consistent.

goral09 (approved):

node/src/components/small_network/tasks.rs
marc-casperlabs: Ah, makes more sense now. Got it :)

cspramit:
@rafal-ch @marc do we @TomVasile to address this test?

rafal-ch:
> @rafal-ch @marc do we @TomVasile to address this test?

Not sure, as we plan to do it with `casper-test`. There is a ticket request on the channel, so the ticket could be planned and distributed on the next grooming meeting.

PR #2460, opened 2021-12-13
2404 USR1/USR2: Different formats for queue dumps.

### LINKED ISSUE #2404: Use both USR1/USR2 for dumping queue data
Currently, dumping the queue via the USR1 signal dumps the larget JSON representation as well as the smaller Debug representation, in that order. Since the JSON representation may be prohibitively large, it should be possible to dump one or the other. This ticket is about making it so that USR1 dumps JSON, USR2 dumps Debug output.

marc-casperlabs:
This PR causes `USR1` to dump only the debug representation of the event queue. `USR2` can be used to obtain the JSON format.

Closes #2404

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -711,7 +711,6 @@ where
                 );
                 if let Err(error) = self.scheduler.snapshot(&mut serializer).await {
                     warn!(%error, "could not serialize snapshot to {}", output_fn);
-                    return;
                 }
             }
             DumpFormat::Debug => {
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -569,7 +569,7 @@ where
                             %threshold_bytes,
                             "node has allocated enough memory to trigger queue dump"
                         );
-                        self.dump_queues().await;
+                        self.dump_queues(DumpFormat::Debug).await;
                     }
                 }
             }
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -64,7 +64,7 @@ use crate::{
     types::{ExitCode, Timestamp},
     unregister_metric,
     utils::{self, SharedFlag, WeightedRoundRobin},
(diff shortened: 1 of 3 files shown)


node/src/reactor.rs
```diff
@@ -678,33 +693,40 @@ where
...
 
-        let debug_dump_filename = format!("/tmp/queue_dump_debug-{}.txt", timestamp);
-        let mut file = match File::create(&debug_dump_filename) {
-            Ok(file) => file,
-            Err(error) => {
-                warn!(%error, "could not create debug output file ({}) for queue snapshot", debug_dump_filename);
-                return;
+        match format {
+            DumpFormat::Json => {
+                let output_fn = format!("/tmp/queue_dump-{}.json", timestamp);
```
goral09, line 702: **[nitpick]:** In my mind `fn` will forever be _function_.

node/src/reactor.rs
```diff
@@ -678,33 +693,40 @@ where
...
+                    warn!(%error, "could not serialize snapshot to {}", output_fn);
+                    return;
+                }
+            }
+            DumpFormat::Debug => {
+                let output_fn = format!("/tmp/queue_dump_debug-{}.txt", timestamp);
+                let mut file = match File::create(&output_fn) {
+                    Ok(file) => file,
+                    Err(error) => {
+                        warn!(%error, "could not create debug output file ({}) for queue snapshot", output_fn);
```
goral09, line 721: Do we need _debug_ here? The file name will already indicate the type of the format.

goral09 (approved):

node/src/reactor.rs
marc-casperlabs: But "filename" is so loooooooooong...

node/src/reactor.rs
marc-casperlabs: I'd rather be explicit, TBH :/

PR #2465, opened 2021-12-14
Re-enable run_equivocator_network test.

### LINKED ISSUE #2475: Re-enable run_equivocator_network test with temporary workaround.
Remove the #[ignore] and instead return early without failing the test if we failed to produce an equivocation. This is a temporary workaround until #1859 is fixed and the test reliably simulates an equivocation.

afck:
This re-enables the test, but doesn't fail if we didn't manage to equivocate.

The check for equivcation should be re-enabled once #1859 is fixed.

Closes  #2475.

### DIFF
--- a/node/src/reactor/participating/tests.rs
+++ b/node/src/reactor/participating/tests.rs
@@ -8,6 +8,7 @@ use num_rational::Ratio;
 use rand::Rng;
 use tempfile::TempDir;
 use tokio::time;
+use tracing::error;
 
 use casper_execution_engine::core::engine_state::GetBidsRequest;
 use casper_types::{
@@ -302,7 +303,6 @@ async fn run_participating_network() {
 }
 
 #[tokio::test]
-#[ignore = "This test fails randomly - it'll be reenabled after this issue is fixed: https://github.com/casper-network/casper-node/issues/1859"]
 async fn run_equivocator_network() {
     testing::init_logging();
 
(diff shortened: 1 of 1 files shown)


node/src/reactor/participating/tests.rs
```diff
@@ -369,6 +369,13 @@ async fn run_equivocator_network() {
         .map(|era_number| switch_blocks.bids(net.nodes(), era_number))
         .collect();
 
+    // Since this setup sometimes fails to produce an equivocation we return early here.
+    // TODO: Remove this once https://github.com/casper-network/casper-node/issues/1859 is fixed.
+    if switch_blocks.equivocators(0).is_empty() {
+        error!("Failed to equivocate in the first era.");
```
goral09, line 375: People will have to remember to check the logs for occurences of `ERROR`-level entries before merging?

afck: If they don't that's not a big issue, though: Most of the time we do manage to create an equivocation here, even in CI, and then the test works correctly, and would actually _fail_ if something unexpected happened.

Anyway, I don't have a better idea currently, and this is only a temporary workaround. I think it's better than ignoring the test entirely.

rafal-ch (approved):
Approved as this is the solution we agreed on during the standup.

node/src/reactor/participating/tests.rs
goral09: I agree about not having a better idea. I was just wondering how we would catch a case when we break something in the code and never produce/detect an equivocation. Then, the only way to find out about it would be to run the test repeatadly and check if it logs error every time?

afck: Yes, that's the danger; that's why this PR doesn't close https://github.com/casper-network/casper-node/issues/1859.
In the meantime, as a temporary workaround, this is still better than having the complete test ignored.

Fraser999 (approved):

afck:
bors merge

PR #2468, opened 2021-12-14
GH-2463: Provide smart contract access to associated keys that signed a deploy 

### LINKED ISSUE #2463: Make deploy signatures available to executing smart contract
Add contract_api // ffi // runtime support to allow smart contracts to load the already present authorized_keys across the wasm boundary, similarly to how load_named_keys currently works.

This feature is necessary for certain types of more advanced smart contracts that utilize multi signature accounts and need to impose some form of role based security in their app logic. For instance, a smart contract that has defined a "clerk", a "supervisor", and "registrar" roles and contains certain entrypoints or operations that are only allowed if one or more signatories of specific role(s) have signed a deploy.

mpapierski:
Closes #2463 

This PR adds `list_authorization_keys` to the contract_api / wasm ffi, allowing smart contracts access to the set of associated `AccountHash`es that signed a deploy.

This is necessary for certain types of more advanced smart contracts that utilize multi signature accounts and need to impose some form of role based security in their app logic. For instance, a smart contract that has defined a "clerk", a "supervisor", and "registrar" roles and contains certain entrypoints or operations that are only allowed if one or more signatories of specific role(s) have signed a deploy.

### DIFF
--- a/execution_engine/src/core/runtime/mod.rs
+++ b/execution_engine/src/core/runtime/mod.rs
@@ -11,7 +11,7 @@ use std::{
     cmp,
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     convert::TryFrom,
-    iter::IntoIterator,
+    iter::{FromIterator, IntoIterator},
 };
 
 use itertools::Itertools;
@@ -3649,6 +3649,46 @@ where
                 Ok(self.context.is_system_contract(contract_hash)?)
             }
         }
+    }
+
+    fn load_authorization_keys(
+        &mut self,
+        len_ptr: u32,
+        result_size_ptr: u32,
+    ) -> Result<Result<(), ApiError>, Trap> {
+        if !self.can_write_to_host_buffer() {
+            // Exit early if the host buffer is already occupied
+            return Ok(Err(ApiError::HostBufferFull));
+        }
+
+        // A set of keys is converted into a vector so it can be written to a host buffer
+        let authorization_keys =
+            Vec::from_iter(self.context.authorization_keys().clone().into_iter());
+
+        let total_keys = authorization_keys.len() as u32;
+        let total_keys_bytes = total_keys.to_le_bytes();
+        if let Err(error) = self.memory.set(len_ptr, &total_keys_bytes) {
+            return Err(Error::Interpreter(error.into()).into());
+        }
+
+        if total_keys == 0 {
+            // No need to do anything else, we leave host buffer empty.
+            return Ok(Ok(()));
+        }
+
+        let named_keys = CLValue::from_t(authorization_keys).map_err(Error::CLValue)?;
+
+        let length = named_keys.inner_bytes().len() as u32;
+        if let Err(error) = self.write_host_buffer(named_keys) {
+            return Ok(Err(error));
+        }
+
+        let length_bytes = length.to_le_bytes();
+        if let Err(error) = self.memory.set(result_size_ptr, &length_bytes) {
+            return Err(Error::Interpreter(error.into()).into());
+        }
+
+        Ok(Ok(()))
     }
 }
 
--- a/execution_engine/src/core/runtime/mod.rs
+++ b/execution_engine/src/core/runtime/mod.rs
@@ -3676,10 +3676,10 @@ where
             return Ok(Ok(()));
         }
 
-        let named_keys = CLValue::from_t(authorization_keys).map_err(Error::CLValue)?;
+        let authorization_keys = CLValue::from_t(authorization_keys).map_err(Error::CLValue)?;
 
-        let length = named_keys.inner_bytes().len() as u32;
-        if let Err(error) = self.write_host_buffer(named_keys) {
+        let length = authorization_keys.inner_bytes().len() as u32;
+        if let Err(error) = self.write_host_buffer(authorization_keys) {
             return Ok(Err(error));
         }
 
--- a/smart_contracts/contract/src/contract_api/runtime.rs
+++ b/smart_contracts/contract/src/contract_api/runtime.rs
@@ -1,6 +1,6 @@
 //! Functions for interacting with the current runtime.
 
-use alloc::{vec, vec::Vec};
+use alloc::{collections::BTreeSet, vec, vec::Vec};
 use core::mem::MaybeUninit;
 
 use casper_types::{
@@ -272,6 +272,31 @@ pub fn put_key(name: &str, key: Key) {
 pub fn remove_key(name: &str) {
     let (name_ptr, name_size, _bytes) = contract_api::to_ptr(name);
     unsafe { ext_ffi::casper_remove_key(name_ptr, name_size) }
+}
+
+/// Returns the set of [`AccountHash`] from the calling account's context `authorization_keys`.
+pub fn list_authorization_keys() -> BTreeSet<AccountHash> {
+    let (total_authorization_keys, result_size) = {
+        let mut authorization_keys = MaybeUninit::uninit();
+        let mut result_size = MaybeUninit::uninit();
+        let ret = unsafe {
+            ext_ffi::casper_load_authorization_keys(
+                authorization_keys.as_mut_ptr(),
+                result_size.as_mut_ptr(),
+            )
+        };
+        api_error::result_from(ret).unwrap_or_revert();
+        let total_authorization_keys = unsafe { authorization_keys.assume_init() };
+        let result_size = unsafe { result_size.assume_init() };
+        (total_authorization_keys, result_size)
+    };
+
+    if total_authorization_keys == 0 {
+        return BTreeSet::new();
+    }
+
+    let bytes = read_host_buffer(result_size).unwrap_or_revert();
+    bytesrepr::deserialize(bytes).unwrap_or_revert()
 }
 
 /// Returns the named keys of the current context.
--- a/smart_contracts/contract_as/assembly/key.ts
+++ b/smart_contracts/contract_as/assembly/key.ts
@@ -44,12 +44,32 @@ export class AccountHash {
         return !this.equalsTo(other);
     }
 
+    @operator(">")
+    graterThan(other: AccountHash): bool {
+        for (let i = 0; i < 32; i++) {
+            if (this.bytes[0] > other.bytes[0]) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    @operator("<")
+    lowerThan(other: AccountHash): bool {
+        for (let i = 0; i < 32; i++) {
+            if (this.bytes[0] < other.bytes[0]) {
+                return true;
+            }
+        }
+        return false;
+    }
+
     static fromPublicKey(publicKey: PublicKey) : AccountHash {
         let algorithmName = publicKey.getAlgorithmName();
         let algorithmNameBytes = encodeUTF8(algorithmName);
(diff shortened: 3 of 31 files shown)


execution_engine/src/core/runtime/mod.rs
```diff
@@ -3650,6 +3650,46 @@ where
...
+        if let Err(error) = self.memory.set(len_ptr, &total_keys_bytes) {
+            return Err(Error::Interpreter(error.into()).into());
+        }
+
+        if total_keys == 0 {
+            // No need to do anything else, we leave host buffer empty.
+            return Ok(Ok(()));
+        }
+
+        let named_keys = CLValue::from_t(authorization_keys).map_err(Error::CLValue)?;
```
EdHastingsCasperAssociation: let `authorization_keys` not let `named_keys`

smart_contracts/contract/src/contract_api/runtime.rs
```diff
@@ -274,6 +274,31 @@ pub fn remove_key(name: &str) {
     unsafe { ext_ffi::casper_remove_key(name_ptr, name_size) }
 }
 
+/// Returns the set of [`AccountHash`] from the calling account's context `authorization_keys`.
+pub fn list_authorization_keys() -> BTreeSet<AccountHash> {
```
EdHastingsCasperAssociation, line 278: I know why it is the case, but on reflection it seems confusing to me that the helper method is called `list_authorization_**keys**` but it returns a set of `AccountHash`.

would it make sense to turn the AccountHash's into Key::Account to align the naming w the type?

smart_contracts/contracts/test/add-associated-key/src/main.rs
```diff
@@ -0,0 +1,18 @@
...
+use casper_types::account::{AccountHash, Weight};
+
+const ARG_ACCOUNT: &str = "account";
+const ARG_WEIGHT: &str = "weight";
+
+#[no_mangle]
+pub extern "C" fn call() {
+    let account: AccountHash = runtime::get_named_arg(ARG_ACCOUNT);
+    let weight: Weight = runtime::get_named_arg(ARG_WEIGHT);
+    account::add_associated_key(account, weight).unwrap_or_revert();
```
EdHastingsCasperAssociation, line 17: would it make sense to remove `authorized-keys` test contract in favor of this one (reworking the tests that use `authorized-keys`), to avoid having two very similar / redundant test contracts?

smart_contracts/contracts/test/list-authorization-keys/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl From<UserError> for ApiError {
+    fn from(error: UserError) -> ApiError {
+        ApiError::User(error as u16)
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn call() {
```
EdHastingsCasperAssociation, line 25: i think we should also add a more complex test contract that sets up the userland use case and proves it works as expected

such a contract would define two roles...lets say role A and role B, with an associated account hash assigned to each, and have two entrypoints...A and B. 

entrypoint A passes if account hash associated to role A signed the deploy calling it else fails

entrypoint B passes if account hash associated to role B signed the deploy calling it else fails

types/src/account/weight.rs
```diff
@@ -17,7 +17,7 @@ pub struct Weight(u8);
 
 impl Weight {
     /// Constructs a new `Weight`.
-    pub fn new(weight: u8) -> Weight {
+    pub const fn new(weight: u8) -> Weight {
```
EdHastingsCasperAssociation, line 20: ?

smart_contracts/contracts/test/list-authorization-keys/src/main.rs
EdHastingsCasperAssociation: and by fail I mean reverts with a custom user error, in case that wasn't obvious

EdHastingsCasperAssociation:
good start, quick turn around, but see notes

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3650,6 +3650,46 @@ where
...
+            // Exit early if the host buffer is already occupied
+            return Ok(Err(ApiError::HostBufferFull));
+        }
+
+        // A set of keys is converted into a vector so it can be written to a host buffer
+        let authorization_keys =
+            Vec::from_iter(self.context.authorization_keys().clone().into_iter());
+
+        let total_keys = authorization_keys.len() as u32;
+        let total_keys_bytes = total_keys.to_le_bytes();
```
goral09, line 3669: Shouldn't we be using `to_bytes()` instead? It will call `to_le_bytes()` on `u32` as well but we would be using a higher-level API that we had defined ourselves.

smart_contracts/contract/src/contract_api/runtime.rs
goral09: OK, this would be a BIG change but I think we should be returning `BTreeSet<PublicKey>` here. Mostly b/c we can derive an `AccountHash` from `PublicKey` but not the other way around. It technically is possible b/c `deploy` [uses `PublicKey` type](https://github.com/casper-network/casper-node/blob/dev/node/src/types/deploy.rs#L1302) that gets mapped to an `AccountHash` when creating a `DeployItem`. The change would be mostly internal and we can keep the interfaces we have so far intact and add new methods that return `PublicKey`s instead. I think contract authors would be more interested in the keys rather than hashes.

execution_engine/src/core/runtime/mod.rs
mpapierski: Currently, `bytesrepr` represents u32 as four little-endian bytes, but it may not be the case in the future. The intention is to write u32 as LE bytes under a pointer, rather than 'ABI serialized u32 integer', which is different (and more expensive). This approach also saves us from calling `bytesrepr::deserialize` on this memory.

goral09: So what are the rules about this? When do we choose `ToBytes/FromBytes` and when we do not?

mpapierski: When you need to pass a single value of a type statically known on the Wasm side like a pointer to u32, then write `to_le_bytes()`, but if you need something more complex like `Vec<AccountHash>`, you should consider `bytesrepr.` Historically we weren't always careful, and we overused `bytesrepr` a lot on the Wasm side, but I hope to address this in 2.0 - for now, I'm trying to stay consistent and pass plain values as LE bytes

smart_contracts/contract/src/contract_api/runtime.rs
mpapierski: > I know why it is the case, but on reflection it seems confusing to me that the helper method is called `list_authorization_**keys**` but it returns a set of `AccountHash`.
> 
> would it make sense to turn the AccountHash's into Key::Account to align the naming w the type?

So this approach would require extra work to unwrap `AccountHash` on the Wasm side. Before making a decision, it's worth pointing out that we also have existing APIs with `_keys` suffix: `add_associated_key,` `remove_associated_key,` `update_associated_key` all operate on `AccountHash`es.

smart_contracts/contracts/test/list-authorization-keys/src/main.rs
mpapierski: Addressed in https://github.com/mpapierski/casper-node/blob/4001974acc48f423064671c1feab4a24583c63d0/smart_contracts/contracts_as/test/multisig-authorization/src/main.rs#L1-L104

smart_contracts/contracts/test/list-authorization-keys/src/main.rs
```diff
@@ -0,0 +1,34 @@
...
+#![no_main]
+
+extern crate alloc;
+
+use alloc::collections::BTreeSet;
+
+use casper_contract::contract_api::runtime;
+use casper_types::{account::AccountHash, ApiError};
+
+const ARG_EXPECTED_AUTHORIZATION_KEYS: &str = "expected_authorized_keys";
```
EdHastingsCasperAssociation: `"expected_authorization_keys"`

types/src/account/weight.rs
mpapierski: Purely convenience to be able to write:

```
const DEFAULT_WEIGHT: Weight = Weight::new(1);
```

EdHastingsCasperAssociation (approved):
lgtm, but will ping you w/ a nuance question

goral09:
My comment (https://github.com/casper-network/casper-node/pull/2468#discussion_r769119376) still stands. Why are we not exposing *keys* instead? They're more valuable to the smart contracts than hashes (one can always go *key -> hash* but not the other way around).

mpapierski:
> My comment ([#2468 (comment)](https://github.com/casper-network/casper-node/pull/2468#discussion_r769119376)) still stands. Why are we not exposing _keys_ instead? They're more valuable to the smart contracts than hashes (one can always go _key -> hash_ but not the other way around).


Consistency with other associated keys APIs with `_key` suffix. There was also an idea to expose `Key::Account`s (as the function has _keys suffix) but it's the same issue

goral09:
> > My comment ([#2468 (comment)](https://github.com/casper-network/casper-node/pull/2468#discussion_r769119376)) still stands. Why are we not exposing _keys_ instead? They're more valuable to the smart contracts than hashes (one can always go _key -> hash_ but not the other way around).
> 
> Consistency with other associated keys APIs with `_key` suffix. There was also an idea to expose `Key::Account`s (as the function has _keys suffix) but it's the same issue

If exposing keys is superior functionality we shouldn't disregard it b/c we've exposed hashes until now.

mpapierski:
> > > My comment ([#2468 (comment)](https://github.com/casper-network/casper-node/pull/2468#discussion_r769119376)) still stands. Why are we not exposing _keys_ instead? They're more valuable to the smart contracts than hashes (one can always go _key -> hash_ but not the other way around).
> > 
> > 
> > Consistency with other associated keys APIs with `_key` suffix. There was also an idea to expose `Key::Account`s (as the function has _keys suffix) but it's the same issue
> 
> If exposing keys is superior functionality we shouldn't disregard it b/c we've exposed hashes until now.

Can we address this is in separate work perhaps? I have some concerns with exposing `casper_types::PublicKey` in current shape to Wasm

darthsiroftardis (approved):

ashok-cl:
I agree with @goral09 here: https://github.com/casper-network/casper-node/pull/2468#issuecomment-998931886

@mpapierski @piotr-dziubecki can we create a ticket and plan for this?

Fraser999 (approved):

goral09:
@mpapierski any actions you need to take before merging the PR?

mpapierski:
> @mpapierski, any actions you need to take before merging the PR?

I just notified Piotr about exposing PublicKeys in future work (2.0.0 perhaps). Merging now

smart_contracts/contract_as/assembly/key.ts
```diff
@@ -44,12 +44,32 @@ export class AccountHash {
         return !this.equalsTo(other);
     }
 
+    @operator(">")
+    graterThan(other: AccountHash): bool {
+        for (let i = 0; i < 32; i++) {
+            if (this.bytes[0] > other.bytes[0]) {
```
afck, line 50: Should that be `i` instead of `0`? (Also in `lowerThan`.)

PR #2493, opened 2021-12-15
Change pid-file behaviour and provide env var to trigger integrity checks

### LINKED ISSUE #2268: trie-store integrity check is running for 9h already - Please reconsider integrity checks after a crash
I'm in the situation that I migrated a testnet node to another server which went well. As you probably know the testnet DB has ~500GB already because of some "load tests". After the migration I started the node on the new server and everything seemed to be fine -> it was syncing. Unfortunately another "load test" was started and the new node seemed to be stuck for about 2h (no log entries anymore, iostat also didn't show any activity anymore) -> I decided to restart the casper-node-launcher via systemctl restart casper-node-launcher. That seemed to kill the node without a clean shutdown and the LMDB was in a dirty state.

That was now 9h ago and it is still "running trie-store integrity check, this may take a while". iotop shows almost constantly 2.15 G/s DISK READ.

I found the following comment:
https://github.com/casper-network/casper-node/blob/release-1.4.1/node/src/reactor/initializer.rs#L234

Are there any plans/possibilities to improve that kind of checks in the new future? Maybe split the DB into smaller chunks? According to @sacherjj 1.4.1 will flush the DB more often (if I understood him correctly) but not sure if that will help with such checks.

Besides the unacceptable long time it would also be very nice to see some kind of progress since I have absolutely no idea what the node is doing and if it checked 95% already or only 3% or something in between.

If I imagine that something like this can happen easily to a mainnet validator I'm really afraid since that means losing lots of delegators, rewards and generally big trouble.

Fraser999:
This PR changes the node startup behaviour to no longer trigger database integrity checks when a crash has been detected.

Instead, the integrity checks can be triggered by the operator setting the environment variable `CL_RUN_INTEGRITY_CHECKS` or by providing non-numeric content in the initializer.pid file before node startup.

A previous crash being detected will still cause a warn-level log line to be created.

Closes #2268.

Tagging @sacherjj for info.


nilz3ro (approved):

marc-casperlabs (approved):
Overall, the code seems fine to me. I'd like to point out that "pidfile" is one word in my mind, I present <https://linux.die.net/man/3/pidfile> as evidence :)

There's only two things that bother me, but I will leave that up for @Fraser999 to decide:

1. Should we really be using the pidfile for this at all? I think specify the envvar is enough mechanism to trigger storage integrity checks.
2. It could have been nice to keep the `Pidfile` more generic. The previous version was something you could have stuffed into a crate without issues, but we have some business logic leaking in with `PidFileOutcome::ForceIntegrityChecks`. If this was `PidFileOutcome::NonNumeric` instead, we would leave the interpretation up to the caller and keep it clean, so to speak.

I don't think any of these are show stoppers though.

Fraser999:
@marc-casperlabs 
> I'd like to point out that "pidfile" is one word in my mind

I just wanted consistency, and since the existing module name (`pid_file`), and type names (`PidFile`, `PidFileError` and `PidFileOutcome`) all pointed to two words, I went with that :)

> Should we really be using the pidfile for this at all?

That's a reasonable point.  This was discussed with @EdHastingsCasperLabs and we decided to retain the pidfile functionality (albeit in a slightly modified form) for now, pending upcoming design discussions around integrity checks.

> It could have been nice to keep the Pidfile more generic.

I tend to favour code starting specific and becoming more generic as it evolves (unless there's a clear need for generic from the outset).  I find that this leads to clearer, simpler code and often the expectation that there will be a need for generic functionality doesn't come to fruition.

Fraser999:
bors merge

PR #2494, opened 2021-12-15
Test Undelegation Timing

### LINKED ISSUE #2241: Add functionality to undelegate everything
Is there a way to undelegate everything from a validator?  We are observing that there’s a little bit that’s left over because we can’t accurately calculate rewards. This will eventually become a problem because we will have dangling delegators in the auction state with small amounts of cspr delegated.

nilz3ro:
# About These Changes
* Tests undelegating during the auction delay.
* Tests undelegating during auction delay + 1.
* closes #2241.


mpapierski (approved):
The test looks solid but you need to fix CI errors

goral09:
The ticket this PR is closing claims that there's some leftover dust (motes) in the account's balance after undelegating but the PR itself adds only a test. Does it mean we don't have a bug in the software that leaves few motes after undelegating? Also, is there an endpoint on Auction contract that allows for undelegating everything? Without having to specify the actual amount we want to undelegate?

nilz3ro:
@goral09 that's correct, I am not able to reproduce the bug in the tests. I'm tagging @EdHastingsCasperLabs because we talked about adding a new Auction method that checks an account's balance at runtime and undelegates everything, but we decided not to do that right now.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -4024,3 +4026,339 @@ fn should_distribute_delegation_rate_full_after_upgrading() {
     assert!(total_payout_before > total_payout_after); // expected total payout after decreasing
                                                        // rate is lower than the first payout
 }
+
+#[test]
```
Fraser999, line 4036:
```suggestion
#[ignore]
#[test]
```

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -4024,3 +4026,339 @@ fn should_distribute_delegation_rate_full_after_upgrading() {
...
+
+    builder.step(step_request).expect("should step");
+
+    // Delegator should not remain delegated even though they were eligible for rewards in the
+    // second era.
+    let delegator = get_delegator_bid(&mut builder, VALIDATOR_1.clone(), DELEGATOR_1.clone());
+    assert!(delegator.is_none());
+}
+
+#[test]
```
Fraser999, line 4206:
```suggestion
#[ignore]
#[test]
```

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -4024,3 +4026,339 @@ fn should_distribute_delegation_rate_full_after_upgrading() {
...
+fn delegator_full_unbond_during_first_reward_era() {
```
Fraser999, line 4207: There's only a very slight difference between these two test cases.  It'd be good to briefly document the two tests, highlighting the difference between them.

PR #2497, opened 2021-12-16
`WasmTestBuilder` should use `ExecutionJournal` for its transforms.

### LINKED ISSUE #2496: Update WasmTestBuilder 's transforms type to Vec<ExecutionJournal> .
Since we started ordering transforms within a block we should also update WasmTestBuilder to track transforms in the correct order.

goral09:
Closes https://github.com/casper-network/casper-node/issues/2496


goral09:
Opened against wrong base branch.

PR #2498, opened 2021-12-16
Update `WasmTestBuilder` to use `ExecutionJournal`.

goral09:
Since we started ordering transforms within a block we should also update WasmTestBuilder to track transforms in the correct order.

Closes https://github.com/casper-network/casper-node/issues/2496.


mpapierski (approved):

goral09:
bors merge

PR #2501, opened 2021-12-16
Upping open file limit on macos

xcthulhu:
- Allows `cargo test` to run successfully


marc-casperlabs (approved):
Looks good to me, thanks for fixing that!

Small nitpick: We could consider making the ~target_os~ target_family `"unix"` instead of removing it I believe, but as long as we're not supporting Windows, it is a non-issue I think.

xcthulhu:
bors merge

PR #2502, opened 2021-12-16
Record `Identity` transform, with a correct contract address, when deploy calls it directly.

goral09:
For deploy types that call stored contract directly (`ExecutableDeployItem::Stored*Contract*`) we were not recording the `Transform::Identity` that we do when those contracts are called from within the body of another contract.

This PR fixes that by using `TrackingCopy::read`, method instead of `TrackingCopy::get`, that adds a proper `Transform::Identity` to the `ExecutionJournal`.

Closes https://github.com/casper-network/casper-node/issues/2470.


mpapierski (approved):

goral09:
bors merge

PR #2511, opened 2021-12-17
Use only hashes, not full deploys, to initialize block proposer.

### LINKED ISSUE #2472: Make replay prevention work with just the deploy hashes.
In fast sync we download all past blocks that according to their timestamps could still have not-yet-expired deploys. For replay detection and prevention the deploy hashes in the block body should suffice; no need to download the actual deploys. However, the current code fails with Non-existent deploy or transfer in block because get_finalized_deploys tries to return the deploys, not just the hashes.

afck:
With fast sync, we don't download all deploys for past blocks. For replay prevention, the block proposer only needs the hashes.

Closes #2472.


node/src/components/block_proposer.rs
```diff
@@ -384,16 +361,11 @@ impl BlockProposerReady {
             trace!(%hash, "expired deploy rejected from the buffer");
             return;
         }
-        if self.unhandled_finalized.remove(hash.deploy_hash()) {
```
goral09, line 387: I don't really understand why `undhandled_finalized` was remove as part of this change. Can you explain why we don't have that collection anymore?

node/src/effect.rs
```diff
@@ -1362,16 +1362,13 @@ impl<REv> EffectBuilder<REv> {
         .await
     }
 
-    /// Retrieves finalized deploys from blocks that were created more recently than the TTL.
-    pub(crate) async fn get_finalized_deploys(
-        self,
-        ttl: TimeDiff,
-    ) -> Vec<(DeployHash, DeployHeader)>
+    /// Retrieves finalized blocks that were created more recently than the deploy TTL.
```
goral09: Since there's no `Deploy` argument to this method I don't think we should be mentioning it in the comment.

node/src/components/storage.rs
```diff
@@ -1352,42 +1352,12 @@ impl Storage {
...
-                // It may happen that deploy was not expired at the time of proposing a block but it
-                // is now.
-                if deploy_header.timestamp().elapsed() > ttl {
-                    continue;
-                }
-                deploys.push((*deploy_hash, deploy_header));
-            }
-        }
-        Ok(deploys)
+        let ttl_not_expired = |block: &Block| block.timestamp().elapsed() < ttl;
```
goral09, line 1359: Not all deploys returned will have their TTL non-expired yet. Is that intentional and OK?

node/src/components/block_proposer/deploy_sets.rs
```diff
@@ -92,6 +105,14 @@ impl BlockProposerDeploySets {
             [pending_deploys, pending_transfers].concat(),
         )
     }
+
+    /// Adds a finalized deploy hash.
+    pub(super) fn add_finalized(&mut self, deploy_hash: DeployHash, new_expiry: Timestamp) {
+        self.finalized_deploys
+            .entry(deploy_hash)
+            .and_modify(|expiry| *expiry = new_expiry.min(*expiry))
```
goral09, line 134: Why would a deploy be finalized more than once?

node/src/components/block_proposer.rs
afck: We needed that collection to track finalized deploy hashes for which we were still missing the deploy itself: We couldn't add them to `finalized_deploys` yet, because that needed the header.

Now that we're only tracking the _hashes_ anyway, we can add them to `self.sets.finalized_deploys` right away.

node/src/components/block_proposer/deploy_sets.rs
afck: In the situation where we would have used `unhandled_finalized`, we would now add the deploy twice:

If we learn about a finalized block containing a deploy for which we are missing the header, we'll add it to `finalized_deploys` with the latest possible expiry date based on the block timestamp and max TTL.

If we then learn about the deploy's header later, we can update its expiry date and use the correct one, based on the deploy's timestamp and actual TTL.

node/src/components/storage.rs
afck: Yes, that's fine:
`finalized_deploys` is a filter to prevent replays. It wouldn't hurt if it contained _all_ finalized deploys, but to preserve space we can remove the ones where we are sure that they are expired.

node/src/effect.rs
afck: Right, that should say "maximum deploy TTL". We should also explain that that means it will contain the hashes of all deploys that are finalized but not yet expired.

node/src/components/storage.rs
goral09: > but to preserve space we can remove the ones where we are sure that they are expired.

Yes, that's what I was thinking about. Not a blocker but a nice-to-have.

node/src/components/block_proposer.rs
goral09: This makes sense. Thanks.

node/src/components/block_proposer.rs
```diff
@@ -410,47 +382,26 @@ impl BlockProposerReady {
...
-                DeployOrTransferHash::Deploy(hash) => {
-                    (hash, self.sets.pending_deploys.remove(&hash))
-                }
-                DeployOrTransferHash::Transfer(hash) => {
-                    (hash, self.sets.pending_transfers.remove(&hash))
-                }
+    /// Handles finalization of a block.
+    fn handle_finalized_block(&mut self, block: &FinalizedBlock) -> Effects<Event> {
+        for deploy_hash in block.deploy_hashes() {
+            let expiry = match self.sets.pending_deploys.remove(deploy_hash) {
```
goral09, line 399: Why do we need to keep the collections for `pending_deploys` and `pending_transfers` separate?

node/src/components/storage.rs
afck: But we are already doing that! When pruning, we remove the expired entries.

node/src/components/block_proposer.rs
afck: The deploy and transfer hashes are separate entries in the block, so it's convenient for proposing blocks if they're already separate. (We could also merge them, and only separate them when proposing… either way, it's not something that was introduced in this PR.)

node/src/components/block_proposer.rs
```diff
@@ -384,16 +361,11 @@ impl BlockProposerReady {
...
-                .finalized_deploys
-                .insert(hash.into(), deploy_info.header);
-            return;
-        }
         // only add the deploy if it isn't contained in a finalized block
         if self.sets.finalized_deploys.contains_key(hash.deploy_hash()) {
-            info!(%hash, "deploy rejected from the buffer");
+            info!(%hash, "finalized deploy rejected from the buffer");
+            self.sets
+                .add_finalized(*hash.deploy_hash(), deploy_info.header.expires());
```
goral09: Seems weird that `add_deploy` adds a deploy to the _finalized_ collection.

afck:
(I won't be able to update the PR before 2022-01-05, but if nothing big is missing maybe it could get merged earlier than that.)

goral09 (approved):

node/src/components/block_proposer.rs
afck: It's only if the deploy is already in that collection.
We add it again to update its expiry date: We probably added it earlier because we learned about the finalized block before we learned about the deploy itself, so we inserted it with the maximum possible expiry date (_block_ timestamp plus _maximum_ deploy TTL). Now we can replace it with the actual expiry date (_deploy_ timestamp plus _actual_ deploy TTL), so we'll be able to purge it from the collection earlier.

goral09: Yes, I understand it. I just wanted to note that it seems weird that we had known about deploy being finalized but didn't have the exact expiry date (b/c we based our knowledge only on the block's timestamp even though we _must_ have had the deploy already - since we can't finalize without downloading all dependencies) and now update it incidentally via `add_deploy` method which should be used for different purposes - like adding _new_ deploys to the block proposer to include in the future.

afck: Yes, I think it's just for the edge case where one event arrives before the other. I'm not sure it can actually happen, but it seems safer to avoid relying on event order?

node/src/components/storage.rs
goral09: Seems incidental. We are interested in deploy hashes (as PR description says) but we are pulling in whole blocks instead. Maybe there should be `get_finalized_deploy_hashes` endpoint instead.

afck: To me the current approach seems more natural: Returning whole blocks instead of just a list of deploy hashes together with an expiry date estimate. But if you prefer it I'd be fine with the alternative, too.

goral09: No, it's your PR and it does the job so I'm fine. I was just trying to understand why it wasn't clear to me what's going on since the beginning.

fizyk20 (approved):
Looks good with the exception of the comment on `get_finalized_blocks` noted by Mateusz, but that's not a blocker.

afck:
bors merge

afck:
bors merge

afck:
bors try

node/src/components/block_proposer/deploy_sets.rs
```diff
@@ -49,11 +50,21 @@ impl BlockProposerDeploySets {
...
     /// the cached state.
     pub(super) fn new(
-        finalized_deploys: Vec<(DeployHash, DeployHeader)>,
+        finalized_blocks: Vec<Block>,
         next_finalized_height: u64,
         cached_state: CachedState,
+        max_ttl: TimeDiff,
     ) -> (BlockProposerDeploySets, PruneResult) {
-        let finalized_deploys: HashMap<_, _> = finalized_deploys.into_iter().collect();
+        let mut finalized_deploys = HashMap::<DeployHash, Timestamp>::new();
```
rafal-ch, line 62: Would it work if we keep a separate collections for deploys and transfers here?

Then instead of 
```
        pending_deploys.retain(|hash, _| !finalized_deploys.contains_key(hash));
        pending_transfers.retain(|hash, _| !finalized_deploys.contains_key(hash));
```
we can have
```
        pending_deploys.retain(|hash, _| !finalized_deploys.contains_key(hash));
        pending_transfers.retain(|hash, _| !finalized_transfers.contains_key(hash));
```
for added legibility :)

Fraser999 (approved):

PR #2516, opened 2021-12-18
Disable usage of checksummed-hex encoding in the node, but continue to support decoding

### LINKED ISSUE #2514: Disable usage of checksummed-hex encoding in the node, but continue to support decoding
Early feedback on the 1.4.3 mainnet release shows that some stakeholders are unable to work with the checksummed-hex encoding as currently implemented.

To resolve this, we should reduce the scope of this encoding to only `AccountHash` and `PublicKey` types.

Fraser999:
This PR limits the usage of checksummed-hex encoding and decoding to account hash and public key types.

The specific types which retain this form of encoding are:
* `AccountHash`
* `PublicKey`
* `Key::Account`
* `Key::Bid`
* `Key::Withdraw`

This will affect downstream consumers who expect checksummed-hex encoding and decoding to be applied to all hash or digest types.  While the node will still accept checksummed-hex encoded values for all such types, no checksum validation will be performed on them.

Closes #2514.


EdHastingsCasperAssociation (approved):

node/src/components/rest_server.rs
```diff
@@ -210,6 +210,8 @@ mod tests {
         let actual_schema = serde_json::to_string_pretty(&actual_schema).unwrap();
         let actual_schema: Value = serde_json::from_str(&actual_schema).unwrap();
 
+        // println!("{}", actual_schema);
```
nilz3ro, line 213: Do we need this?

nilz3ro (approved):

node/src/components/rest_server.rs
darthsiroftardis: Yes, its being used for when changes are made to the RPC schema

Fraser999:
Note that the ticket has been updated to cover disabling all usages of this encoding style to ensure we're providing output which all current consumers can use.

cspramit:
whats is the sdk impact? @hoffmannjan @Fraser999 @Adrian-Wrona

Fraser999:
Closing this PR as it will target a different branch and will disable all checksummed-hex encoding

PR #2531, opened 2021-12-22
Remove `casper-mainnet` feature flag

rafal-ch:
This PRs removes the `casper-mainnet` feature flag.


rafal-ch:
**Update**: tests are now updated and do not fail.

TODOs:
* ~Failing test: `block_with_deploys_round_trip_lmdb()`, with `Block has incorrect block hash.`~ - resolved
* ~Failing tests with a particular seed:~
```
CL_TEST_SEED=db116508126c89831CD6AA9da22d2925 cargo test components::storage::tests::can_put_and_get_blocks_v2

CL_TEST_SEED=1fE5231685e4cF5E486a395Cf113FD8D cargo test types::block::tests::block_body_merkle_proof_should_be_correct
```
caused by `era_id` to be randomized as `0`.

~It might require further work around this place:~
~https://github.com/afck/casper-node/commit/b2318bb43ae028b94097b461c20a54b1b8fdccc2#diff-f8f1402a94ae96f1e79dfff1d2af53d86a2114c73096a211873dbbb4fe4226e9R2143~

goral09:
@rafal-ch This PR has over 5k LOC touched. Is it really just removing `casper-mainnet` feature flag?

rafal-ch:
> @rafal-ch This PR has over 5k LOC touched. Is it really just removing `casper-mainnet` feature flag?

It removes the `resources/test/rpc_schema_hashing_V2.json` which contributes -4222 lines.

client/lib/validation.rs
```diff
@@ -287,23 +287,38 @@ pub(crate) fn validate_get_block_response(
...
+    // Currently, the consumer of this function (the client app) has no access to the chainspec,
+    // hence, it cannot give us the proper `merkle_tree_hash_activation` which is needed
+    // for block verification. Therefore, we try to validate the block with both legacy
+    // and new hashing algorithm. If one of the validation passes, we assume the block is correct.
+    // This is a temporary solution which should be removed once the client is given
+    // access to the proper `merkle_tree_hash_activation`.
+    let result_legacy_hash = block.verify(EraId::from(0u64));
+    let result_new_hash = block.verify(block.header().era_id() + 1);
+
+    match (result_legacy_hash, result_new_hash) {
```
darthsiroftardis: Would it possible to change to this to a if statement
if `result_legacy_hash.is_ok()` || `result_new_hash.is_ok()`

node/src/components/fetcher/event.rs
```diff
@@ -1,5 +1,6 @@
 use std::fmt::{self, Debug, Display, Formatter};
 
+use casper_types::EraId;
```
darthsiroftardis: Import in the wrong place

node/src/reactor.rs
```diff
@@ -44,6 +44,7 @@ use std::{
     sync::atomic::Ordering,
 };
 
+use casper_types::EraId;
```
darthsiroftardis: Import in the wrong place

node/src/types/block.rs
```diff
@@ -1376,21 +1382,23 @@ impl Block {
...
 
-    pub(crate) fn header(&self) -> &BlockHeader {
-        &self.header
-    }
-
     pub(crate) fn body(&self) -> &BlockBody {
         &self.body
     }
 
+    /// Returns the reference to the header/
```
darthsiroftardis:
```suggestion
    /// Returns the reference to the header
```

utils/dry-run-deploys/src/main.rs
```diff
@@ -1,5 +1,6 @@
 use std::{path::PathBuf, time::Instant};
 
+use casper_types::EraId;
```
darthsiroftardis: Import in the wrong place

utils/retrieve-state/src/main.rs
```diff
@@ -6,6 +6,7 @@ use std::{
     str::FromStr,
 };
 
+use casper_types::EraId;
```
darthsiroftardis: Import in the wrong place

darthsiroftardis (commented):
Some nitpicks

node/src/components/consensus.rs
```diff
@@ -302,9 +311,12 @@ where
...
                 delay,
             } => self.handle_deactivate_era(effect_builder, era_id, faulty_num, delay),
-            Event::CreateNewEra { switch_blocks } => {
-                self.create_new_era_effects(effect_builder, rng, &switch_blocks)
-            }
+            Event::CreateNewEra { switch_blocks } => self.create_new_era_effects(
+                effect_builder,
+                rng,
+                &switch_blocks,
+                self.merkle_tree_hash_activation(),
```
afck: Why pass that in if `self` has the chainspec anyway?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -334,8 +341,9 @@ where
         effect_builder: EffectBuilder<REv>,
         rng: &mut NodeRng,
         switch_blocks: &[BlockHeader],
+        merkle_tree_hash_activatio: EraId,
```
afck:
```suggestion
        merkle_tree_hash_activation: EraId,
```

But I think that can be removed anyway and taken from the chainspec:

```suggestion
```

node/src/components/consensus/era_supervisor.rs
```diff
@@ -234,12 +235,18 @@ where
...
                 &switch_blocks[..i],
+                merkle_tree_hash_activation,
             ));
         }
 
         Ok((era_supervisor, effects))
     }
 
+    /// Returns the merkle tree hash activation from the chainspec.
+    pub(crate) fn merkle_tree_hash_activation(&self) -> EraId {
```
afck: `pub(super)` would be enough; but I suspect we can even make it private.

node/src/components/fetcher.rs
```diff
@@ -524,15 +531,23 @@ where
                 }
                 None => self.failed_to_get_from_storage(effect_builder, id, peer),
             },
-            Event::GotRemotely { item, source } => {
+            Event::GotRemotely {
+                merkle_tree_hash_activation: _,
```
afck, line 535: That should be removed from the event, if we don't need it.

node/src/components/fetcher/tests.rs
```diff
@@ -89,9 +89,20 @@ reactor!(Reactor {
                 .chainspec()
                 .protocol_config
                 .last_emergency_restart,
+            chainspec_loader
+                .chainspec()
+                .protocol_config
+                .merkle_tree_hash_activation,
```
afck, line 95: Would it be easier to just pass a `&Chainspec` into `Storage::new`?

node/src/components/rpc_server.rs
```diff
@@ -335,28 +335,8 @@ mod tests {
 
     #[test]
     fn schema() {
-        // The expected schema depends on the hashing algorithm selected by the `casper-mainnet`
-        // feature.
-        //
-        // To generate the contents to replace the input JSON files, run the test with and without
```
afck, line 337: The comment starting from here, and the `println` at the end, are not made obsolete with this change.

node/src/components/storage/tests.rs
```diff
@@ -35,6 +35,8 @@ use crate::{
     utils::WithDir,
 };
 
+const MERKLE_TREE_HASH_ACTIVATION: u64 = 1;
```
afck: Since our test blocks (almost?) always have height ≥ 1, should we set this to e.g. 5, so we test both hashing versions?

node/src/lib.rs
```diff
@@ -75,16 +75,7 @@ pub(crate) use types::NodeRng;
...
-        "{}-{}{}",
-        env!("CARGO_PKG_VERSION"),
-        env!("VERGEN_SHA_SHORT"),
-        if cfg!(feature = "casper-mainnet") {
-            "-casper-mainnet"
-        } else {
-            ""
-        }
-    );
+    let mut version = format!("{}-{}", env!("CARGO_PKG_VERSION"), env!("VERGEN_SHA_SHORT"),);
```
afck:
```suggestion
    let mut version = format!("{}-{}", env!("CARGO_PKG_VERSION"), env!("VERGEN_SHA_SHORT"));
```

node/src/reactor/joiner.rs
```diff
@@ -566,6 +570,10 @@ impl reactor::Reactor for Reactor {
             }
         };
 
+        let merkle_tree_hash_activation = chainspec_loader
+            .chainspec()
```
afck, line 535:
```suggestion
        let merkle_tree_hash_activation = chainspec
```

(And probably in some more places in this function.)

node/src/testing/multi_stage_test_reactor/test_chain.rs
```diff
@@ -26,6 +26,10 @@ use crate::{
     NodeRng,
 };
 
+const ERA_TO_JOIN: u64 = 3;
+const MERKLE_TREE_HASH_ACTIVATION: u64 = ERA_TO_JOIN + 1;
+const MERKLE_TREE_HASH_ACTIVATION_FOR_ARCHIVAL_SYNC: u64 = ERA_TO_JOIN;
```
afck: Why use a different value for the archival sync test?

It also seems a bit dangerous; we'll end up accidentally using these in the wrong test. Should these constants be local to the individual tests instead?

node/src/types/block.rs
```diff
@@ -2000,6 +2017,11 @@ mod tests {
 
     use super::*;
 
+    impl HashingAlgorithmVersion {
+        pub(crate) const HASH_V2_PROTOCOL_VERSION: ProtocolVersion =
+            ProtocolVersion::from_parts(1, 5, 0);
+    }
+
```
afck: That should be unused now?

```suggestion
```

utils/dry-run-deploys/src/main.rs
```diff
@@ -48,8 +49,11 @@ async fn main() -> Result<(), anyhow::Error> {
     let chain_download_path = normalize_path(&opts.chain_download_path)?;
     let lmdb_path = normalize_path(&opts.lmdb_path)?;
 
+    let merkle_tree_hash_activation = EraId::from(0u64);
```
afck, line 60: I think eventually tools like this should just read a chainspec. Not in this PR, but maybe we should ticket it and add a `TODO` comment.

utils/retrieve-state/src/main.rs
```diff
@@ -137,6 +138,7 @@ async fn main() -> Result<(), anyhow::Error> {
 
     let chain_download_path = env::current_dir()?.join(&opts.chain_download_path);
     let url = format!("{}/rpc", opts.server_host);
+    let merkle_tree_hash_activation = EraId::from(0u64);
```
afck, line 166: Maybe also add a `TODO` here.

client/lib/validation.rs
rafal-ch: Thanks! Good point.
Fixed here with the "early exit" approach: https://github.com/casper-network/casper-node/pull/2531/commits/0d334a753fe3f3d86216a1d148e8f72325ab84c4

node/src/components/fetcher/event.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2531/commits/9a4261fae4aa6c6834052ff12514d12d47c9e858

node/src/reactor.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2531/commits/9a4261fae4aa6c6834052ff12514d12d47c9e858

node/src/types/block.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2531/commits/9a4261fae4aa6c6834052ff12514d12d47c9e858

utils/dry-run-deploys/src/main.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2531/commits/9a4261fae4aa6c6834052ff12514d12d47c9e858

utils/retrieve-state/src/main.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2531/commits/9a4261fae4aa6c6834052ff12514d12d47c9e858

node/src/components/consensus.rs
rafal-ch: Good catch, fixed in https://github.com/casper-network/casper-node/pull/2531/commits/70cc245e4930d1ed0e81d79d0d212d3a73bbe8a0

node/src/components/consensus/era_supervisor.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2531/commits/70cc245e4930d1ed0e81d79d0d212d3a73bbe8a0

node/src/components/consensus/era_supervisor.rs
rafal-ch: Fixed in 70cc245

node/src/components/fetcher.rs
rafal-ch: Unfortunately, it is needed for the sake of `impl Display`, like here: https://github.com/casper-network/casper-node/pull/2531/files#diff-eef5ec27d08d2fabcc4844fb7bb058c0e88ccb2d2f9e352facd842b7b9666554R134-R149

That was the simplest approach I came up with to be able to present the correct hash in the `Display`.

I will happily accept any other approach that doesn't require substantial refactoring. Alternatively, we may spawn a ticket to handle that in the future.

afck: Hmm… maybe adding an `id` field instead of `merkle_tree_hash_activation` would work? We'd just compute the hash before generating the event: Then we wouldn't need a field that's specific to one event type.
Anyway, I'm not sure about it. It's not a blocker.

node/src/components/rpc_server.rs
rafal-ch: Comment restored in https://github.com/casper-network/casper-node/pull/2531/commits/a2125d3ebda7cf16c4e0923504e7e2641a7eb3f4 and https://github.com/casper-network/casper-node/pull/2531/commits/ea3399cf46e774fc6f7d39fecc00decf45597b5b

node/src/components/fetcher/tests.rs
rafal-ch: Yes, that's a good idea. We should even use this approach in `ContractRuntime::new()`, but as agreed on a call, we should attack it in the separate PR mainly because it will require substantial changes to `retrieve-state` tool, for example [here](https://github.com/casper-network/casper-node/blob/release-1.4.3/utils/retrieve-state/src/storage.rs#L128-L139) we'd need to create an entire, dummy chainspec to pass it to the constructor.

It'll work nicely after the suggestion from [this comment](https://github.com/casper-network/casper-node/pull/2531#discussion_r778709175) is implemented, though.

I will leave the code as-is and request a `technical-debt` ticket.

node/src/reactor/joiner.rs
rafal-ch: The `chainspec` binding is consumed by the `move` closure above and cannot be used any longer.

Introducing a new binding will not improve clarity that much, imho.

utils/retrieve-state/src/main.rs
rafal-ch: TODO added [here](https://github.com/casper-network/casper-node/pull/2531/commits/2ddb40a696aa96983c34e0a7e1ed31b7c67d0570).
Resolving for now due to https://github.com/casper-network/casper-node/pull/2531#discussion_r778823863

utils/dry-run-deploys/src/main.rs
rafal-ch: TODO added [here](https://github.com/casper-network/casper-node/pull/2531/commits/2ddb40a696aa96983c34e0a7e1ed31b7c67d0570).
Resolving for now due to https://github.com/casper-network/casper-node/pull/2531#discussion_r778823863

node/src/components/storage/tests.rs
rafal-ch: Yes, actually we should explicitly test against both hashing versions starting at `block.rs`. I will update the tests there and then follow with the components that use the `Block`.

node/src/testing/multi_stage_test_reactor/test_chain.rs
rafal-ch: Inspired by this comment I changed the approach to the tests as follows:
- set the `MERKLE_TREE_HASH_ACTIVATION` individually for each test that involves this parameter
- if the test allows for the activation point to be arbitrarily chosen, I use random value
- otherwise, I execute the test using **both** legacy blocks and merkle hashed blocks

`block` tests are updated here: https://github.com/casper-network/casper-node/pull/2531/commits/f9e0ba78556fb1b2856974b241d758887da882f7

`storage` tests are updated here: https://github.com/casper-network/casper-node/pull/2531/commits/1c233d024bc5aa05a9e59a77a97a17e615fe7050

Tests still to be updated/reviewed:
- deploy_acceptor :heavy_check_mark:
- gossiper :heavy_check_mark: 
- linear_chain :heavy_check_mark:
- linear_chain_sync :heavy_check_mark:
- multi_stage_test_reactor :heavy_check_mark:

Please note that in the two commits above there is some temporary code (like: `random_with_specifics_1()`) which will be removed after all tests are updated/reviewed.

CC @Fraser999

node/src/components/storage/tests.rs
```diff
@@ -388,293 +357,346 @@ fn switch_block_for_block_header(
...
         Default::default(),
-        finalized_block,
+        finalized_block.clone(),
         Some(validator_weights),
         block_header.protocol_version(),
-        MERKLE_TREE_HASH_ACTIVATION.into(),
+        merkle_tree_hash_activation,
     )
     .expect("Could not create block")
 }
```
afck, line 368:
```suggestion
}

```

node/src/components/storage/tests.rs
```diff
@@ -388,293 +357,346 @@ fn switch_block_for_block_header(
...
-    assert!(get_block_header_at_height(&mut storage, 14).is_none());
-    assert!(get_block_at_height(&mut storage, 33).is_none());
-    assert!(get_block_header_at_height(&mut storage, 33).is_none());
-    assert!(get_block_at_height(&mut storage, 99).is_none());
-    assert!(get_block_header_at_height(&mut storage, 99).is_none());
-
-    // Inserting 33 changes this.
-    let was_new = put_block(&mut harness, &mut storage, block_33.clone());
-    assert!(was_new);
+    for block_spec in block_specs {
```
afck, line 544: Not sure if we really need to double all the tests.
For this one we could just set the activation at era 2, so one of the three example blocks uses the new scheme?

node/src/components/storage/tests.rs
```diff
@@ -942,22 +988,22 @@ fn store_execution_results_twice_for_same_block_deploy_pair() {
...
 
-    put_execution_results(&mut harness, &mut storage, block_hash, exec_result.clone());
+//     put_execution_results(&mut harness, &mut storage, block_hash, exec_result.clone());
 
-    // We should be fine storing the exact same result twice.
-    put_execution_results(&mut harness, &mut storage, block_hash, exec_result);
-}
+//     // We should be fine storing the exact same result twice.
+//     put_execution_results(&mut harness, &mut storage, block_hash, exec_result);
+// }
```
afck: ~Does this test break?~ (Ah, already fixed.)

node/src/components/storage/tests.rs
```diff
@@ -986,73 +1036,98 @@ fn test_legacy_interface() {
...
 
 #[test]
 fn should_hard_reset() {
     let blocks_count = 8_usize;
     let blocks_per_era = 3;
     let mut harness = ComponentHarness::default();
-    let mut storage = storage_fixture(&harness);
+
+    // `merkle_tree_hash_activation` can be chosen arbitrarily
+    let merkle_tree_hash_activation = harness.rng.gen();
```
afck: This test will create blocks in three eras, but `merkle_tree_hash_activation` will almost never be that small. Maybe use a random number from the range `0..4` instead?

node/src/components/storage/tests.rs
```diff
@@ -1319,19 +1398,25 @@ impl DatabaseEntriesSnapshot {
...
     let blocks: Vec<Block> = (0..blocks_count)
         .map(|height| {
             let is_switch = height % blocks_per_era == blocks_per_era - 1;
-            Block::random_with_specifics(
+
+            Block::random_with_specifics_1(
                 &mut harness.rng,
                 EraId::from((height / blocks_per_era) as u64),
                 height as u64,
                 ProtocolVersion::from_parts(1, 4, 0), // so that they are stored in v2 scheme
```
afck: This comment is obsolete.

node/src/components/storage/tests.rs
```diff
@@ -1395,27 +1481,66 @@ fn should_garbage_collect() {
...
     let mut harness = ComponentHarness::default();
-    let mut storage = storage_fixture(&harness);
 
-    let era_id = harness
-        .rng
-        .gen_range(MERKLE_TREE_HASH_ACTIVATION..MERKLE_TREE_HASH_ACTIVATION + 10)
-        .into();
+    let era_id = harness.rng.gen_range(0..10).into();
+
+    // Ensure the merkle hashing algorithm is used.
```
```suggestion
    // Ensure the Merkle hashing algorithm is used.
```

afck (approved):
My remaining comments aren't blockers.

node/src/testing/multi_stage_test_reactor/test_chain.rs
rafal-ch: The tests are now updated.

node/src/components/storage/tests.rs
rafal-ch: Yes, that could also work.

I'll keep it as it, since it's not a blocker and is also doing the job. The test is very quick to execute so we're not adding much to the total execution of the tests.

btw: I used the approach with specifically selected eras and activation points in the tests that are long to execute (mainly in `multi_stage_test_reactor/test_chain.rs`)

node/src/components/fetcher.rs
rafal-ch: Having the `id` in the event could work, but also has drawbacks:
1. event creators need to have access to `merkle_tree_hash_activation` - even if they don't care (we can solve it by having an `Option<T::Id>`
2. even with the Option, we'll end-up having duplicated information in the Event, so the `handle_event` can access both `item.id()` or `T::Id`. We have two items that should match, but we can't enforce that in the code.

I'll leave it like it is with a `TODO` comment. I have a feeling that sooner or later we'd need to refactor out this `merkle_tree_hash_activation` tramp data.

node/src/components/storage/tests.rs
rafal-ch: Yes, that's right, to some extent :)
Because `merkle_tree_hash_activation` is `u64` I'd still like this to be tested against huge values. I think I'll use the `LogNormal` distribution here with parameters strongly preferring the low numbers, but occasionally hitting the big ones.

node/src/components/fetcher.rs
rafal-ch: Comment added in https://github.com/casper-network/casper-node/pull/2531/commits/efabc750a6da1dc3b8d15d7e421b1bf2c5fee74a

node/src/components/storage/tests.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2531/commits/1b1ab632ad8a3b673597c109ba986ccfb5638d5e

node/src/components/storage/tests.rs
rafal-ch: Applied in https://github.com/casper-network/casper-node/pull/2531/commits/5bc3333778683c431c4666489acf3b2e21b12dcc

node/src/components/storage/tests.rs
rafal-ch: Removed in https://github.com/casper-network/casper-node/pull/2531/commits/15a212035887b1ced40ec3375626d8c251fef585

node/src/components/storage/tests.rs
rafal-ch: Updated (also in few other places) here https://github.com/casper-network/casper-node/pull/2531/commits/15a212035887b1ced40ec3375626d8c251fef585

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1030,7 +1030,8 @@ mod tests {
         chainspec.core_config.minimum_era_height = minimum_era_height;
 
         // `merkle_tree_hash_activation` can be chosen arbitrarily
-        let merkle_tree_hash_activation = EraId::from(rng.gen::<u64>());
+        let log_normal = LogNormal::new(2.5, 3.7).unwrap();
```
afck: Is it really worth adding another dependency for this?
Why not just a uniform distribution on `0..10` or something?

rafal-ch: Ok, updated in https://github.com/casper-network/casper-node/pull/2531/commits/718213235b419e3e44699eaf5682ee1e1fbff8bb.

I'm usually cool about additional `dev-dependencies`, especially the de-facto standard ones. I could achieve what I imagined with the already included `proptest` but thought it would be overkill.

Dependencies aside, what worries me a bit more is testing the system with input data that covers a tiny fraction of possible values (`1..=10` out of `u64::MAX`). Anyway, changes to that, if needed at all, should not be a part of this PR though, hence the revert.

afck (approved):

afck (approved):

Fraser999 (approved):

PR #2534, opened 2021-12-22
Disable checksummed-hex encoding

### LINKED ISSUE #2514: Disable usage of checksummed-hex encoding in the node, but continue to support decoding

Fraser999:
This PR disables all instances of checksummed-hex encoding, leaving checksummed-hex decoding unmodified.

This will affect downstream consumers who expect checksummed-hex encoding to be applied to all hash or digest types output by the node or client.

Despite the output types no longer using mixed-case hex, mixed-case hex as input for all such types will continue to have checksum validation performed on them.

Closes #2514.


nilz3ro (approved):

node/src/components/small_network/message.rs
```diff
@@ -352,6 +352,24 @@ mod tests {
         55, 55, 53, 48, 66, 67, 67, 69, 65, 69,
     ];
 
+    /// A "conserved" version 1.4.3 handshake.
+    ///
+    /// NEVER CHANGE THIS CONSTANT TO MAKE TESTS PASS, AS IT IS BASED ON MAINNET DATA.
+    const V1_4_3_HANDSHAKE: &[u8] = &[
```
marc-casperlabs, line 358: Is this table-to-farm, i.e. did you acquire this from a live 1.4.3 and dumping it?

dwerner (approved):
lgtm

darthsiroftardis (approved):

marc-casperlabs (commented):
The networking portions look good at first glance at least. Did not run a full cross version compatibility, I'll leave that to SRE :)

Fraser999:
bors merge

node/src/components/small_network/message.rs
Fraser999: No, I acquired it by running test code from this module, but while checked out to the commit from which 1.4.3 was cut.

xcthulhu:
@Fraser999 I went ahead and created a ticket to reenable this in 2.0

node/src/components/small_network/message.rs
marc-casperlabs: That's actually what I meant, oops. Thank you!

PR #2535, opened 2021-12-22
Pass chainspec values to EE test support crate

### LINKED ISSUE #2313: CasperEngine test support crate needs to be updated to support the use of chainspec files

darthsiroftardis:
CHANGELOG:

- Changed the constructor for the `WasmTestBuilder` to take a path pointing to a chainspec file
- Added a new struct `ChainspecConfig` to represent and load the chainspec values and pass the values to the `WasmTestBuilder`
- Refactored tests within for the execution engine to use the new functionality of the test support crate

Closes #2313


goral09:
If we change the constructor of `WasmTestBuilder` then this becomes a breaking change for all downstream consumers of the testing crate. Is it necessary for it to become the main way of constructing the `WasmTestBuilder`? I would prefer if we added a separate method and schedule deprecation for the future.

darthsiroftardis:
> If we change the constructor of `WasmTestBuilder` then this becomes a breaking change for all downstream consumers of the testing crate. Is it necessary for it to become the main way of constructing the `WasmTestBuilder`? I would prefer if we added a separate method and schedule deprecation for the future.

Point well made, I can reintroduce the more "default" methods trivially, @Fraser999 thoughts on this?

goral09 (changes requested):
Let me block this PR until my previous comment is resolved as it is a breaking change.

darthsiroftardis:
@goral09 I spoke to @Fraser999 and we agree with your comments, closing this PR in favor of a non-breaking change, which will be forthcoming

PR #2536, opened 2021-12-23
GH-2547: Test framework improvements

### LINKED ISSUE #2547: Test framework - handle missing deploy hash
DeployItemBuilder should panic when build() was called without specifying with_deploy_hash . Lack of deploy hash may affect addresses generated during contract execution which may lead to unexpected results. There is chance some tests doesn’t specify deploy hashes, and there’s also chance some tests are passing because of that, but shouldn’t with a valid deploy hash.
Points: 1-3
Components: test support
:ticket: Misleading error
We discovered that (and above ticket) with @Austin  yesterday that installing and upgrading a contract with a deploy hash that was same for both exec requests produced strange results: executing versioned contract by contract package hash and a version number raised TypeMismatch error here: https://github.com/mpapierski/casper-node/blob/984f445232370aa13b30530286a91c9869e0143f/execution_engine/src/core/tracking_copy/ext.rs#L198-L212
This is not exploitable on live network and happens only when a test doesn’t specify deploy hash.
We should change TypeMismatch to say ContractWasm but not ContractHeader

mpapierski:
Closes #2547 

Not specifying the `deploy_hash` field in `DeployItemBuilder` will use the default value. This may lead to a situation where multiple deploys are starting with the same seed value in PRNG that may cause unexpected results.

This PR also changes misleading error `TypeMismatch(ContractHeader, ...)` into `TypeMismatch(ContractWasm, ...)`


execution_engine_testing/test_support/src/deploy_item_builder.rs
```diff
@@ -18,7 +18,7 @@ struct DeployItemData {
     pub session_code: Option<ExecutableDeployItem>,
     pub gas_price: u64,
     pub authorization_keys: BTreeSet<AccountHash>,
-    pub deploy_hash: DeployHash,
```
goral09, line 21: So if this type was non-optional before, how was it NOT required? Was some default used?

goral09, line 21: Couldn't we calculate `DeployHash` from the `DeployItemData`? What's missing?

mpapierski: I think for the testing purpose, it's fine as it is. We're using RNG by default here through the high-level APIs, but if you want to construct deploy yourself, I think it's fair to require this field

nilz3ro (approved):

PR #2537, opened 2021-12-23
GH-2429: Change pointer block format

### LINKED ISSUE #2429: Optimize PointerBlock serialization format
Follow up for #2346. We serialize PointerBlocks as 256 Option's regardless if the value is Some or None and due to Option overhead PointerBlock size is at least 256 bytes. As an optimization we can serialize only Some variants and we can reuse option's tag to write u8 index. After running create_domains entrypoint that writes exactly 49500 dictionary writes I could save 40MB of storage (848 -> 808). The biggest obstacle implementing this is requirement of either a complete global state migration, or a clever mechanism to determine which deserialization format to use based on a protocol version.

mpapierski:
Closes #2429 

This PR decouples Merkle proofs from the serialization format which will allow us to freely change trie (and the global state) serialization format. With that in mind, this PR implements a new PointerBlock serialization format that achieves a smaller size on disk.

Before the PointerBlock was serialized with `N * (digest_length + pointer_tag_length + option_tag_length) + (256 - N) * option_tag_length`. This means at minimum for empty PointerBlock we always have overhead of 256 bytes (`option_tag_length == 1`).

New format serializes into `32 + (N + 7) / 8 + (N * digest_length)` bytes where first 32 bytes is a bitmap where each bit represents Some or None, then next `(N + 7) / 8` bytes represents variants of `Pointer`s (Leaf or Node), and remaining data are the raw pointers stored (multiplies of 32 bytes). Depending on the fill ratio in PointerBlock tries we can save [224..448] bytes per PointerBlock. (saves 224 for empty pointer block, 448 for full pointer block).

The new dense format is used always in ToBytes, but the legacy sparse PointerBlock is still readable. This is achieved by adding a new trie tag to the code.

This also changes ToBytes API to instead use a `Writer` trait rather than `Vec<u8>` to be able to generate a hash of a Trie without allocating memory and as a side effect, it achieves about ~3% gain in `bytesrepr` benchmarks for those involving arrays of bytes or other complex structures. Perf profiling wasn't done but we should observe much fewer ToBytes calls during storage operations.

WIP: I was not able to test it with mainnet global date despite efforts but this PR uses lmdb fixtures to confirm existing Merkle proofs are valid, and legacy data can be read, and also that newly written data is readable. This should give enough confidence to try syncing a node with new code.


dwerner:
image

dwerner (dismissed):

goral09:
Can we wait with this merge until after 1.5.0 release?

PR #2539, opened 2021-12-26
diagnostic console

marc-casperlabs:
This adds a diagnostic console, which right now primarily allows for dumping the state of consensus.

The console is configurable (see the configuration files) and accepts connections from a unix socket created (if enabled) next to the config files by default.


README.md
```diff
@@ -201,6 +221,23 @@ Count number of events in each queue:
 jq 'map_values(map(keys[0]))' queue_dump.json
 ```
 
+### Debug console
+
+If the configuration option `console.enabled` is set to `true`, a unix socket named `debug.socket` by default can be found next to the configuration while the node is running. It can be connected to by tools like `socat`:
+
+```sh
+$ sockat - unix:/path/to/debug.socket
```
dwerner:
```suggestion
$ socat - unix:/path/to/debug.socket
```

README.md
```diff
@@ -155,19 +155,39 @@ modules in `components`, and `warn` level for the remaining codebase:
 RUST_LOG=casper_node::components::small=trace,casper_node::comp=info,warn
 ```
 
+### Logging network messages and tracing events
+
+Special logging targets exist in `net_in` and `net_out` which can be used to log every single network message leaving or
```
goral09, line 168: Do we need to break this line? It will get wrapped correctly when displayed. These artifical break lines will mess up the presentation if you have wider screen.

Same comment applies for this whole file.

README.md
```diff
@@ -155,19 +155,39 @@ modules in `components`, and `warn` level for the remaining codebase:
...
+entering a node when set to trace level:
+
+```
+RUST_LOG=net_in::TRACE,net_out::TRACE
```
goral09, line 172: Should we make it 
```bash
RUST_LOG=$RUST_LOG:net_in::TRACE,net_out::TRACE
```
?
Otherwise someone might accidentally overwrite previous log levels when trying to add this special logging.

README.md
```diff
@@ -195,6 +215,23 @@ Count number of events in each queue:
 jq 'map_values(map(keys[0]))' queue_dump.json
 ```
 
+### Debug console
```
goral09:
```suggestion
### Diagnostic console
```
?
To match the PR description/name.

README.md
```diff
@@ -195,6 +215,23 @@ Count number of events in each queue:
...
+
+If the configuration option `console.enabled` is set to `true`, a unix socket named `debug.socket` by default can be found next to the configuration while the node is running. It can be connected to by tools like `socat`:
+
+```sh
+$ socat - unix:/path/to/debug.socket
+```
+
+Entering `help` will show available commands. The interface can also be scripted, it may be helpful to change some of the per-connection settings in this case:
```
goral09, line 234:
```suggestion
Entering `help` will show available commands. 

The interface can also be scripted, it may be helpful to change some of the per-connection settings in this case:
```

README.md
```diff
@@ -195,6 +215,23 @@ Count number of events in each queue:
...
+
+```sh
+echo -e 'set -q true -o bincode\nexample-command' | socat - unix-client:/path/to/debug.socket > output.dump
```
goral09, line 237: Should it be 
```suggestion
echo -e 'set -q true -o bincode\example-command' | socat - unix-client:/path/to/debug.socket > output.dump
```
?

node/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.  The format
 ### Added
 * Add new event to the main SSE server stream across all endpoints `<IP:PORT>/events/*` which emits a shutdown event when the node shuts down.
 * Add `SIGUSR2` signal handling to dump the queue in JSON format (see "Changed" section for `SIGUSR1`).
+* A debug console can now be enabled via the `[console]` section in the configuration file. See the `README.md` for details.
```
```suggestion
* A diagnostic console can now be enabled via the `[console]` section in the configuration file. See the `README.md` for details.
```

node/src/components/consensus.rs
```diff
@@ -357,6 +363,29 @@ where
...
+                    .open_eras
+                    .get(&requested_era)
+                    .ok_or_else(|| {
+                        Cow::Owned(format!(
+                            "could not dump consensus, {} not found",
+                            requested_era
+                        ))
+                    })
+                    .and_then(|era| EraDump::dump_era(era, requested_era));
+
```
goral09, line 383: Maybe log that we're done dumping the era? To let the node operator a chance to measure how long it took.

node/src/components/consensus.rs
```diff
@@ -357,6 +363,29 @@ where
                 let validator_changes = self.get_validator_changes();
                 responder.respond(validator_changes).ignore()
             }
+            Event::DumpState(req @ DumpConsensusStateRequest { era_id, .. }) => {
```
goral09, line 366: Do we need to bind the `req @` if it's not used later? Wouldn't
```suggestion
            Event::DumpState(DumpConsensusStateRequest { era_id, .. }) => {
```
suffice?

node/src/components/consensus.rs
```diff
@@ -357,6 +363,29 @@ where
...
+                            "could not dump consensus, {} not found",
+                            requested_era
+                        ))
+                    })
+                    .and_then(|era| EraDump::dump_era(era, requested_era));
+
+                match era_dump_result {
+                    Ok(dump) => req.answer(Ok(&dump)).ignore(),
+                    Err(err) => req.answer(Err(err)).ignore(),
+                }
```
goral09, line 387: If `era_dump_result` is of `Result` type already, can it be shortened to:
```suggestion
	req.answer(era_dump_result).ignore()
```
?

node/src/components/consensus/era_supervisor.rs
```diff
@@ -85,10 +86,10 @@ pub struct EraSupervisor<I> {
     ///
     /// This map always contains exactly `2 * bonded_eras + 1` entries, with the last one being the
     /// current one.
-    open_eras: HashMap<EraId, Era<I>>,
+    pub(super) open_eras: HashMap<EraId, Era<I>>,
     secret_signing_key: Arc<SecretKey>,
     public_signing_key: PublicKey,
-    current_era: EraId,
+    pub(super) current_era: EraId,
```
goral09: I'm not sure how @afck feels about it but I would prefer to stick to the approach where we have methods to access this rather than making fields accessible directly.

node/src/components/consensus/highway_core/state.rs
```diff
@@ -141,7 +142,7 @@ impl<C: Context> Fault<C> {
 /// Both observers and active validators must instantiate this, pass in all incoming vertices from
 /// peers, and use a [FinalityDetector](../finality_detector/struct.FinalityDetector.html) to
 /// determine the outcome of the consensus process.
-#[derive(Debug, Clone, DataSize)]
+#[derive(Debug, Deserialize, Clone, DataSize, Serialize)]
```
goral09: Why do we need `Deserialize` for these types? Do we make use of these automatically derived deserializers in the client that reads the dumped state of consensus?

node/src/effect/console.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+
+impl Display for DumpConsensusStateRequest {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "dump consensus state for ")?;
+        if let Some(ref era_id) = self.era_id {
+            Display::fmt(era_id, f)
+        } else {
+            f.write_str("for latest era")
```
goral09:
```suggestion
            f.write_str(" latest era")
```
Otherwise it will repeat `for`.

node/src/reactor/joiner.rs
```diff
@@ -352,6 +369,7 @@ pub(crate) struct Reactor {
     block_validator: BlockValidator<NodeId>,
     deploy_fetcher: Fetcher<Deploy>,
     linear_chain: linear_chain::LinearChainComponent<NodeId>,
+    console: Console,
```
goral09, line 372: Can we find a better name for it? `console` is too ambiguous. Maybe `DiagnosticConsole` or `DebuggingConsole`?

node/src/reactor/joiner.rs
```diff
@@ -407,6 +425,9 @@ impl reactor::Reactor for Reactor {
 
         let metrics = Metrics::new(registry.clone());
 
+        let (console, console_effects) =
+            Console::new(&WithDir::new(&root, config.console.clone()), event_queue)?;
```
goral09, line 429: Do we need to create an instance of it if we can't handle the requests? Can't we just _not_ create it?

node/src/effect/console.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(DataSize, Serialize)]
+pub(crate) struct DumpConsensusStateRequest {
+    /// Era to serialize.
+    ///
+    /// If not given, use active era.
+    pub(crate) era_id: Option<EraId>,
+    /// Serialization function to serialize the given era with.
+    #[data_size(skip)]
+    #[serde(skip)]
+    pub(crate) serialize: fn(&EraDump) -> Result<Vec<u8>, Cow<'static, str>>,
```
goral09, line 24: Can you explain to me why it has to be `Cow`? Can't it just be `String`? I think developer's clarity and understanding is more important than saving some bits and CPU cycles. Especially that these will be rather small strings.

node/src/utils/umask.rs
```diff
@@ -0,0 +1,30 @@
...
+pub(crate) struct UmaskGuard {
+    /// The mode stored in the guard.
+    prev: Mode,
+}
+
+impl Drop for UmaskGuard {
+    fn drop(&mut self) {
+        umask(self.prev);
+    }
+}
```
goral09, line 30: I don't understand why we need this type. Can you add a docstring at the top?

node/src/components/consensus/era_supervisor/debug.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl<'a> Display for EraDump<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "era {}: TBD", self.id)
+    }
+}
+
+impl<'a> EraDump<'a> {
+    /// Creates a new `EraDump` from a given era.
+    pub(crate) fn dump_era<I>(era: &'a Era<I>, era_id: EraId) -> Result<Self, Cow<'static, str>> {
```
goral09, line 56: Why did you choose to have a special method for it rather than `new` or even `TryFrom`?

goral09:
Just FYI, your branch is called _config hot reload_ .

node/src/components/console.rs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    // This is not thread-safe, as it will set the umask for the entire process, but we assume that
+    // initalization happens "sufficiently single-threaded".
+    let umask_guard = umask::temp_umask(socket_umask);
+    let listener = UnixListener::bind(socket_path)?;
+    drop(umask_guard);
+
+    debug!(local_addr=%ShowUnixAddr(&listener.local_addr()?), "console socket listening");
```
goral09, line 125:
```suggestion
    debug!(local_addr=%ShowUnixAddr(&listener.local_addr()?), "diagnostic console socket listening");
```
? To make it clearer to the log reader what's happening.

node/src/components/console.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// Creates a new Unix console component.
+    pub(crate) fn new<REv>(
+        cfg: &WithDir<Config>,
+        event_queue: EventQueueHandle<REv>,
+    ) -> Result<(Self, Effects<Event>), Error>
+    where
+        REv: From<DumpConsensusStateRequest> + Send,
+    {
+        let config = cfg.value();
+        let (shutdown_sender, shutdown_receiver) = watch::channel(());
```
goral09, line 72: Pardon me my incompetence, but how does this work? We're creating a channel here but we never use `sender` part. Is the design here that `shutdown_sender` will be dropped as the `Console` component gets dropped in the `joiner`/`participating` reactor which will then send a signal to `shutdown_receiver` that the corresponding `sender` was dropped? Which in turn will close the socket connection?

node/src/components/console.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// `umask` to apply before creating the socket.
+    socket_umask: u16,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            socket_path: "debug.socket".into(),
+            socket_umask: 0o077,
```
goral09, line 57: How is that `u16`? 😅 I haven't seen this format before. What does the `o` indicate to the compiler?

README.md
```diff
@@ -195,6 +215,23 @@ Count number of events in each queue:
...
+```
+
+This calls `example-command` on the console, but disables the command confirmation (`-q true`) and changes the output format to bincode (`-o bincode`) before doing so.
+
+
```
goral09, line 242: It's probably just me being dumb but I wanted to play with it. Connected just fine. Then I went ahead and tried `help`, which gave me a list of allowed commands. I wanted to see how to change a session so I typed:
```
set help

ERR  error: Found argument 'help' which wasn't expected, or isn't valid in this context

USAGE:
    casper-console set [OPTIONS]

For more information try --help
```
See under `USAGE`. It sounded like I should try `casper-console set` but that failed:
```
casper-console set
ERR  error: Found argument 'casper-console' which wasn't expected, or isn't valid in this context

USAGE:
    casper-console <SUBCOMMAND>

For more information try --help
```

What I'm trying to say is that we should try to change the `USAGE` output as it's misleading.

goral09: Actually, I'm not really sure how I should be using the console once I'm already connected. I can't see a README on it anywhere and the auto-generated descriptions aren't very helpful.

**EDIT:**
I got it. It didn't take long but I had to read the code and required understanding of how `structopt` generates the (sub)commands. I think we should add it to the README.

**EDIT 2:**
What is the expected behavior of
```
set -o interactive
dump-consensus
```
? Because I'm getting some weird behavior:
```
set -o interactive
OK   session updated
dump-consensus
OK   dumping consensus state
era era 45: TBD

ERR  casper-node 1.4.3
Action to perform

USAGE:
    casper-console <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    dump-consensus    Dump the state of the consensus component
    help              Prints this message or the help of the given subcommand(s)
    session           Retrieve the active console session information
    set               Set options on active console session
```

goral09 (approved):
This looks great. Very informative. I wish I will be able to write a code like that from the top of my head one day :) Great job 🎉

node/src/components/console/tasks.rs
```diff
@@ -0,0 +1,335 @@
...
+pub(super) async fn server<REv>(
+    effect_builder: EffectBuilder<REv>,
+    socket_path: PathBuf,
+    listener: UnixListener,
+    mut shutdown_receiver: watch::Receiver<()>,
+) where
+    REv: From<DumpConsensusStateRequest> + Send,
+{
+    let handling_shutdown_receiver = shutdown_receiver.clone();
+    let mut next_client_id: u64 = 0;
```
goral09, line 290: Shouldn't it be `AtomicU32` or sth?

node/src/components/console/tasks.rs
```diff
@@ -0,0 +1,335 @@
...
+
+                    span.in_scope(|| {
+                        info!(client_addr = %ShowUnixAddr(&client_addr), "accepted connection");
+                    });
+
+                    next_client_id += 1;
+
+                    tokio::spawn(
+                        handler(effect_builder, stream, handling_shutdown_receiver.clone())
+                            .instrument(span),
```
goral09, line 307: Why do you think it's necessary to intstrument it?

node/src/components/console/tasks.rs
```diff
@@ -0,0 +1,335 @@
...
+                    );
+                }
+                Err(err) => {
+                    info!(%err, "failed to accept incoming connection on console socket");
+                }
+            }
+        }
+    };
+
+    let shutdown_messages = async move { while shutdown_receiver.changed().await.is_ok() {} };
```
goral09, line 317: Just checking my understanding: this future will run until `shutdown_receiver.changed()` returns `Err`? Shouldn't we consider any signal sent on to that receiver as a shutdown signal?

node/src/components/console/tasks.rs
```diff
@@ -0,0 +1,335 @@
...
+
+                        if let Some(quiet) = quiet {
+                            self.quiet = quiet;
+                            changed = true;
+                        }
+
+                        if let Some(output) = output {
+                            self.output = output;
+                            changed = true;
+                        }
```
goral09: If we wanted to be too picky, then I would say we should check if that's a no-op (new values are equal to the old ones). But that's not really important.

node/src/components/console/tasks.rs
```diff
@@ -0,0 +1,335 @@
...
+        match Command::from_line(line) {
+            Ok(ref cmd) => {
+                info!(?cmd, "processing command");
+                match cmd.action {
+                    Action::Session => {
+                        self.send_outcome(writer, &Outcome::success("showing session info"))
+                            .await?;
+                        self.send_to_client(writer, &self).await?;
+                    }
+                    Action::Set { quiet, output } => {
```
goral09, line 136: In order to match the flow in other two actions, there should be something like `self.send_outcome(writer, &Outcome::success("changing session"));` here.

node/src/components/console/tasks.rs
```diff
@@ -0,0 +1,335 @@
...
+                            )
+                            .await;
+
+                        match output {
+                            Ok(ref data) => {
+                                self.send_outcome(
+                                    writer,
+                                    &Outcome::success("dumping consensus state"),
+                                )
+                                .await?;
```
goral09, line 171: IMO this entry should be right at the beginning of this clause and the error should be logged regardless - i.e. it looks to me that these first `send_outcome`s calls are indicating an action being taken regardless of the results later.

node/src/components/consensus.rs
goral09: Oh wait, there is `req` used down below 🙈

node/src/components/consensus.rs
goral09: Ah, I see you can't as `asnwer` expects `Result<&EraDump>` while we have here `Result<EraDump>`.

node/src/effect/console.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Responder to send the serialized representation into.
+    pub(crate) responder: Responder<Result<Vec<u8>, Cow<'static, str>>>,
```
goral09, line 26: Looks like neither this `pub(crate)`, nor the one on `serialize` are necessary.

node/src/reactor/joiner.rs
goral09: Oh, I think I see - we should be accepting new connections but we won't handle the command to dump consensus' state. But won't we drop all connections when transitioning from `joiner` to `participating` reactor? Since the `sender_shutdown` gets dropped.

goral09: And just for the sake of completion, when reactor transitions from `joining` to `participating`, it will drop the socket and print this to the console:
```
2022/01/17 01:18:00 socat[1881] E write(5, 0x55e4ef777f10, 8): Broken pipe
```

README.md
marc-casperlabs: > Same comment applies for this whole file.

I was faced with the same question and just went with what was there. I don't think this PR should address that. I also have no opinion on the matter :)

README.md
marc-casperlabs: I would keep it simple as is, this is just an example. I hope the user is competent enough to realize how to use `RUST_LOG` to compose things (otherwise, why are they setting these settings at all?)

Also, depending on how and where they set these settings, the `$RUST_LOG` might not even be the appropriate way to include previous vars.

README.md
marc-casperlabs: No, the `\n` is a linebreak, which terminates the previous commands (thus the `-e` on `echo`).

goral09: Ah, I thought it's a path 👍

README.md
marc-casperlabs: I didn't want the single line there to feel so lonely! In my mind, this makes it a little bit awkward. I can still change it if you insist though, I don't exactly have a strong opinion on this :P

node/src/components/consensus.rs
marc-casperlabs: They can just measure it client side (put a `time` before the `socat` command).

node/src/components/consensus/era_supervisor.rs
marc-casperlabs: My bad, that was an oversight. I initially had all these reverted to accessor methods once I saw the pattern, but these two slipped through. I'll fix it right away.

marc-casperlabs: Also, shoutout to rust-analyzer!

```rust
    /// Get a reference to the era supervisor's open eras.
    pub fn open_eras(&self) -> &HashMap<EraId, Era<I>> {
        &self.open_eras
    }
```

The above was fully autogenerated, including the comment. :exploding_head:

node/src/components/consensus/era_supervisor/debug.rs
marc-casperlabs: `TryFrom` is wrong IMHO, since it is not trivial code to create it. Also, I don't like `TryFrom` for reference types that much.

`dump_era` for one is more descriptive then `new`, I don't think it hurts for constructors to have more descriptive names.

I'm not attached to this though - but I have not seen sufficient evidence to change it just yet :)

See also: https://rust-lang.github.io/api-guidelines/naming.html

node/src/components/consensus/highway_core/state.rs
marc-casperlabs: That was the initial intention, but I have copied these types in the interim. I'll remove the `Deserialize`.

node/src/components/console.rs
marc-casperlabs: Octal notation, the `0o` is to octal what `0x` is to hexadecimal. Other languages use `0077` == `077` (leading zero indicates octal), which is confusing. Rust thus requires the explicit `o`.

Unix file permissions are usually written as octal, since every digit then indicates one set of permissions (world, group, user). Three bits per (read, write, execute), so read+write+execute is `4+2+1 = 7` (or `0b111` = `0o7`).

goral09: > Octal notation, the `0o` is to octal what `0x` is to hexadecimal.

Thanks! I didn't know that :)

node/src/components/console.rs
marc-casperlabs: That's exactly right :)

goral09: Awesome.

node/src/utils/umask.rs
marc-casperlabs: ~I will, but at the `UmaskGuard` type level.~

After reading all the docs, I'm not entirely sure how to make this better/more obvious. `temp_umask` allows setting a new umask (which is global) for the time the guard lives. We usually do not want to change it permanently, just for a single call/instance/limited block when creating a file (it's a deficiency in the POSIX/Linux APIs that makes this a thing at all). The guard helps to make it impossible to forget to reset it.

node/src/reactor/joiner.rs
marc-casperlabs: > Oh, I think I see - we should be accepting new connections but we won't handle the command to dump consensus' state.

Yes! We could just not create it of course, but not being able to handle any requests is temporary state of the feature - soon there should be requests that can be answered in the joiner as well: #2546. So while we could omit it for now, I think it's better to add it here straight away.

> But won't we drop all connections when transitioning from joiner to participating reactor? Since the sender_shutdown gets dropped.

Yes. This is a bit of a wrinkle, but I think I can live with that for now. It's the same with any other component that gets re-created.

> And just for the sake of completion, when reactor transitions from joining to participating, it will drop the socket and print this to the console:

That's just socat reporting the connection has been closed.

node/src/reactor/joiner.rs
marc-casperlabs: There shall only be one! I'm loathe to change it at this point, to be honest. If you strongly feel it is necessary, I will though :)

node/src/effect/console.rs
marc-casperlabs: I think `Cow<'static, str>` is pretty standard (or should be). If need be, a neat trick would be to `type Str = Cow<'static, str>' to make it more readable.

The main reason is to not have to write `"asdf".to_string()` further down.

node/src/effect/console.rs
marc-casperlabs: `rustc` seems to differ :)

```
error[E0451]: field `serialize` of struct `DumpConsensusStateRequest` is private
    --> node/src/effect.rs:1829:17
     |
1829 |                 serialize,
     |                 ^^^^^^^^^ private field

error[E0451]: field `responder` of struct `DumpConsensusStateRequest` is private
    --> node/src/effect.rs:1830:17
     |
1830 |                 responder,
     |                 ^^^^^^^^^ private field
```

goral09: Oh interesting. It passed `cargo check` for me.

node/src/components/console/tasks.rs
marc-casperlabs: There is, in lines 150ff? Do you want this to happen earlier?

node/src/effect/console.rs
goral09: Why not `Result<Vec<u8>, String>`?

node/src/components/console/tasks.rs
goral09: That's not the same as in line 150 you're confirming the action and my understanding is that the first output sent is a confirmation of action being taken.

node/src/components/console/tasks.rs
marc-casperlabs: I think 65005407e should make you happy though :)

node/src/components/console/tasks.rs
marc-casperlabs: There's no ~one else~ other task accessing this, so no atomic necessary!

node/src/components/console/tasks.rs
marc-casperlabs: `instrument` is a bit misleading here, it's just entering and exiting the span. This gives proper context on log messages if there are multiple console connections.

node/src/components/console/tasks.rs
marc-casperlabs: We never send anything down the receiver, `Err` means the other end was dropped, which is the signal to shutdown.

node/src/components/console/tasks.rs
goral09: Thanks :)

node/src/components/console/tasks.rs
marc-casperlabs: I'm not sure I understand this. To be clear, we are sending the outcome, which is the status of the consensus dumping operation first, then the actual data.

We cannot send down any errors encountered in `write_all` if that's what you mean, as we cannot send the client a sending error!

node/src/components/console/tasks.rs
goral09: I see! I think I thought that when a sender is dropped it will change the shared value. Thanks for the explanation.

node/src/components/console/tasks.rs
goral09: I meant that the approach here is that `Console` will send the first response that is a confirmation of an action being taken (ignored if `quiet = true`) and then the result of taking the action. `"dumping consensus state"` is the former and should be sent regardless of the actual outcome.

node/src/components/console/tasks.rs
goral09: Got it. Thanks.

node/src/effect/console.rs
marc-casperlabs: That's what I'm trying to avoid. It makes it so that you have to heap allocate a copy of every fixed string returned as an error (`return Err("missing foo".to_owned())`).

README.md
marc-casperlabs: You entered three commands, `set -o interactive`, `dump-consensus` and an empty command ``.

`era era 45: TBD` is the "interactive" description of the era dump. It is lacking in detail, because the `Display` implementation is still poor.

The data sent afterwards is what you get when you write `help` or are missing a command.

node/src/effect/console.rs
goral09: That's not really a big performance issue, isn't it? Short string only in case of errors which should be infrequent anyway.

node/src/components/console/tasks.rs
marc-casperlabs: It just sometimes happens to be sent early, but these are not intended as confirmations of reception, but rather outcomes.

README.md
goral09: > `era era 45: TBD` is the "interactive" description of the era dump. It is lacking in detail, because the Display implementation is still poor.

Ok. That solves the mystery :)

PR #2542, opened 2021-12-30
Merge `dev` to `feat-fast-sync` 30.12.2021

rafal-ch:
This PR integrates the recent changes from the `dev` (as of 30.12.2021 11:00am CEST) branch into the `feat-fast-sync` branch.


node/src/components/fetcher/metrics.rs
```diff
@@ -3,7 +3,7 @@ use prometheus::{IntCounter, Registry};
 use crate::unregister_metric;
 
 #[derive(Debug)]
-pub(crate) struct FetcherMetrics {
+pub(crate) struct Metrics {
```
Fraser999, line 6:
```suggestion
pub(super) struct Metrics {
```

Fraser999 (approved):
One nitpick.

node/src/components/fetcher/metrics.rs
rafal-ch: Actually, on the `feat-fast-sync` the `Fetcher` further exposes the `Metrics` type via the public `ItemFetcher` trait.

I'll see if I can tune the code structure in order not to expose `Metrics` on the crate level.

rafal-ch: As agreed on the call, we'll merge as-is.

PR #2549, opened 2022-01-06
Add LMDB file size to wasm_test_builder

### LINKED ISSUE #2481: Add on-disk file size checks to WasmTestBuilder
We need to watch data.lmdb and it's growth, so we should add the capacity to check the on-disk size of lmdb during a test. This would be useful to know, for example, the size growth after N transfers, purses, auction steps with N delegators, etc.

dwerner:
Resolves #2481 

This PR (which should be reviewed after https://github.com/casper-network/casper-node/pull/2277 merges to dev):

- adds `lmdb_on_disk_size` method to WasmTestBuilder.
- extracts a significant amount of the logic in `transfer_bench` and `auction_bench` into utility methods within execution_engine_testing, for reuse.
- `disk_use` binary target that will run a series of tests against the trie to measure disk use with transfers and auction code that was extracted in the previous point.

### DIFF
--- a/execution_engine_testing/test_support/src/wasm_test_builder.rs
+++ b/execution_engine_testing/test_support/src/wasm_test_builder.rs
@@ -10,7 +10,7 @@ use std::{
 };
 
 use filesize::PathExt;
-use lmdb::DatabaseFlags;
+use lmdb::{Database, DatabaseFlags};
 use log::LevelFilter;
 
 use bytesrepr::FromBytes;
@@ -326,9 +326,8 @@ impl LmdbWasmTestBuilder {
         None
     }
 
-
     /// Execute and commit transforms from an ExecuteRequest into a scratch global state.
-    /// You MUST call scratch_flush to flush these changes to LmdbGlobalState.
+    /// You MUST call write_scratch_to_lmdb to flush these changes to LmdbGlobalState.
     pub fn scratch_exec_and_commit(&mut self, mut exec_request: ExecuteRequest) -> &mut Self {
         if self.scratch_engine_state.is_none() {
             self.scratch_engine_state = Some(self.engine_state.get_scratch_engine_state());
@@ -379,6 +378,16 @@ impl LmdbWasmTestBuilder {
             );
         }
         self
+    }
+
+    /// Returns a handle to the underlying LMDB environment.
+    pub fn lmdb_environment(&self) -> Arc<LmdbEnvironment> {
+        self.engine_state.lmdb_environment()
+    }
+
+    /// Returns a handle to the underlying LMDB database.
+    pub fn lmdb_handle(&self) -> Database {
+        self.engine_state.lmdb_handle()
     }
 }
 
--- a/execution_engine_testing/test_support/src/wasm_test_builder.rs
+++ b/execution_engine_testing/test_support/src/wasm_test_builder.rs
@@ -9,6 +9,7 @@ use std::{
     sync::Arc,
 };
 
+use filesize::PathExt;
 use lmdb::DatabaseFlags;
 use log::LevelFilter;
 
@@ -110,6 +111,8 @@ pub struct WasmTestBuilder<S> {
     auction_contract_hash: Option<ContractHash>,
     /// Scratch global state used for in-memory execution and commit optimization.
     scratch_engine_state: Option<EngineState<ScratchGlobalState>>,
+    /// Global state dir, for implementations that define one.
+    global_state_dir: Option<PathBuf>,
 }
 
 impl<S> WasmTestBuilder<S> {
@@ -140,6 +143,7 @@ impl Default for InMemoryWasmTestBuilder {
             handle_payment_contract_hash: None,
             standard_payment_hash: None,
             auction_contract_hash: None,
+            global_state_dir: None,
             scratch_engine_state: None,
         }
     }
@@ -163,6 +167,7 @@ impl<S> Clone for WasmTestBuilder<S> {
             standard_payment_hash: self.standard_payment_hash,
             auction_contract_hash: self.auction_contract_hash,
             scratch_engine_state: None,
+            global_state_dir: self.global_state_dir.clone(),
         }
     }
 }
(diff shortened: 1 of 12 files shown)


goral09:
Can you merge `dev` to this branch, now that the ScratchGlobalState has been merged?

piotr-dziubecki:
@EdHastingsCasperLabs : rename and generalise the description.

dwerner:
@Fraser999 @goral09 can I please get a review on this? My next PR (rocksdb) will build on it.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -309,8 +316,18 @@ impl LmdbWasmTestBuilder {
         path
     }
 
+    /// Returns the file size on disk of the backing lmdb file behind LmdbGlobalState.
+    pub fn lmdb_on_disk_size(&self) -> Option<u64> {
```
goral09, line 320:
```suggestion
    pub fn db_on_disk_size(&self) -> Option<u64> {
```

We don't want to "leak" the implementation details which is the usage of LMDB. It's even more true when you replace it with RocksDB.

execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -0,0 +1,254 @@
...
+            println!("running auction");
+            auction::step_and_run_auction(&mut builder, &validator_keys);
+        }
+    }
+}
+
+fn main() {
+    transfer_disk_use(2500, 100).unwrap();
+    for purse_count in [100] {
+        for transfer_count in [2500usize] {
```
goral09, line 248: Let's make these two values configurable from the CLI.

execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+        if run_auction {
+            println!("running auction");
+            auction::step_and_run_auction(&mut builder, &validator_keys);
+        }
+    }
+}
+
+fn main() {
+    transfer_disk_use(2500, 100).unwrap();
```
goral09, line 246: We could make this configurable too.

goral09 (commented):
@dwerner can you include example output of the `disk_use` tool in the PR description?

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -362,6 +379,16 @@ impl LmdbWasmTestBuilder {
...
+
+    /// Returns a handle to the underlying LMDB environment.
+    pub fn lmdb_environment(&self) -> Arc<LmdbEnvironment> {
+        self.engine_state.lmdb_environment()
+    }
+
+    /// Returns a handle to the underlying LMDB database.
+    pub fn lmdb_handle(&self) -> Database {
+        self.engine_state.lmdb_handle()
+    }
```
goral09, line 391: These two also leak the implementation detail. From what it looks like later on the usage side (`disk_use`) the only reason we need those is to call `cursor.iter()`. How about instead this, we add a method `iter()` and return an iterator over elements in the DB? That should be implementable for all types of DB backends:
* LMDB
* in-memory
* RocksDB

execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -0,0 +1,254 @@
...
+            Trie::Extension {
+                affix: _,
+                pointer: _,
+            } => extension_node_lengths.push(byte_len),
+        }
+
+        record_count += 1;
+        let serialized_len = value.len();
+        if largest_record < serialized_len {
+            println!("found new largest DB entry with len {}", serialized_len);
```
goral09, line 107: Won't we print this multiple times - every time we find a new record that is bigger than the previous biggest one? Isn't that a bit too chatty? I think we can drop it and just report at the very end, after the loop, next to other statistics.

execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -0,0 +1,254 @@
...
+
+use casper_engine_test_support::{
+    auction::{self},
+    transfer, LmdbWasmTestBuilder, DEFAULT_ACCOUNT_ADDR, MINIMUM_ACCOUNT_CREATION_BALANCE,
+};
+use casper_types::{account::AccountHash, bytesrepr::FromBytes, Key, StoredValue, U512};
+
+const TARGET_ADDR: AccountHash = AccountHash::new([127; 32]);
+
+fn transfer_disk_use(transfer_count: usize, purse_count: usize) -> Result<(), Box<dyn Error>> {
```
goral09, line 16: Please add a comment describing what will be done and measured as part of this method call.

execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -0,0 +1,254 @@
...
+                &exec_requests,
+                use_scratch,
+            );
+            println!(
+                "{}, {}",
+                builder.lmdb_on_disk_size().unwrap(),
+                total_transfers
+            );
+        }
+        if run_auction {
```
goral09, line 238: Why would we _not_ want to run an auction?

execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -0,0 +1,254 @@
...
+}
+
+fn main() {
+    transfer_disk_use(2500, 100).unwrap();
+    for purse_count in [100] {
+        for transfer_count in [2500usize] {
+            // baseline, one deploy per exec request
+            multiple_native_transfers(transfer_count, purse_count, false, false);
+            multiple_native_transfers(transfer_count, purse_count, true, false);
+        }
```
goral09, line 252: This is a bit confusing to me. The `multiple_native_transfers` will output on-disk size via `builder.lmdb_on_disk_size()` but we also have another measurement of the Trie size in `transfer_disk_use`. How are they different? Is it a coincidence that `transfer_disk_suze` and `multiple_native_transfers` use same parameters (2500 transfers, 100 purses)? Or is it on purpose so that we have a measurements of both Trie structure and on-disk size?

goral09 (commented):
Suggestion fo the future, you could have made the commits a bit more granular - it would make the review much simpler. For example, the first commit contains both new `lmdb_on_disk_size()` method AND extraction of bench utils - it could have been made two separate commits.

PR #2552, opened 2022-01-07
Fix leaky faucet

### LINKED ISSUE #2556: Fix leaky faucet

nilz3ro:
# About These Changes

Previously, this contract accepted the parameters `target` and `amount` which were used to perform a transfer of `amount` tokens to the `target` account. This was always called in the faucet contract installer's context as a stored session.

This contract has been refactored to a stored contract (which runs in its own context) and keeps track of the installer, a distribution interval , an amount per interval, max distributions per interval, and a last distribution timestamp and internally, remaining amount.  A caller will receive `amount_per_interval / max_distributions_per_interval` while the `remaining_amount > 0`, but the installer may fund accounts as often as they like.


smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,493 @@
...
 };
-use casper_types::{account::AccountHash, ApiError, Key, U512};
+use casper_types::{
+    account::AccountHash,
+    api_error, bytesrepr,
+    bytesrepr::{FromBytes, ToBytes},
+    ApiError, BlockTime, CLTyped, CLValue, Key, URef, U512,
+};
+use core::{cmp::Ordering, mem::MaybeUninit};
+use num::rational::Ratio;
```
Fraser999: These should be grouped as
* std lib
* third party libs
* casper libs
* this crate's imports

So:
```rust
use alloc::{vec, vec::Vec};
use core::{cmp::Ordering, mem::MaybeUninit};

use num::rational::Ratio;

use casper_contract::{
    contract_api::{self, runtime, storage, system},
    ext_ffi,
    unwrap_or_revert::UnwrapOrRevert,
};
use casper_types::{
    account::AccountHash,
    api_error,
    bytesrepr::{self, FromBytes, ToBytes},
    ApiError, BlockTime, CLTyped, CLValue, Key, URef, U512,
};
```

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,493 @@
...
+    let last_distribution_time_uref = get_uref_with_user_errors(
+        LAST_DISTRIBUTION_TIME,
+        FaucetError::MissingDistributionTime,
+        FaucetError::InvalidDistributionTime,
+    );
+
+    storage::write::<u64>(last_distribution_time_uref, t.into());
+}
+
+fn get_named_arg_size(name: &str) -> Option<usize> {
```
Fraser999, line 349: I can't say I like the level of code duplication in this and the following functions, especially since they mostly call unsafe code.  These are in some cases identical to code in `casper-contract`.

I can see why it's done like this, but I think we should ticket improvements to the contract API to support this use case, since I don't think it's unique to the faucet contract.

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,493 @@
...
+        INSTALLER,
+        FaucetError::MissingInstaller,
+        FaucetError::InvalidInstaller,
+    );
+
+    if installer != runtime::get_caller() {
+        runtime::revert(FaucetError::InvalidAccount);
+    }
+
+    runtime::put_key(TIME_INTERVAL, storage::new_uref(7_200_000u64).into());
```
Fraser999: This value `7_200_000u64` could be a const with a name like `TWO_HOURS_AS_MILLISECS`.

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,493 @@
...
+        FAUCET_PURSE,
+        FaucetError::MissingFaucetPurse,
+        FaucetError::InvalidFaucetPurse,
+    );
+
+    system::transfer_from_purse_to_account(faucet_purse, target, amount, id)
+        .unwrap_or_revert_with(FaucetError::FailedToTransfer);
+}
+
+fn get_distribution_amount_debounced() -> U512 {
```
Fraser999, line 216: I believe this function could be simplified:
```rust
    let remaining_amount_uref = get_uref_with_user_errors(
        REMAINING_AMOUNT,
        FaucetError::MissingRemainingAmount,
        FaucetError::InvalidRemainingAmount,
    );
    let remaining_amount: U512 = read_with_user_errors(
        remaining_amount_uref,
        FaucetError::MissingRemainingAmount,
        FaucetError::InvalidRemainingAmount,
    );

    let distribution_amount = get_distribution_amount();

    if remaining_amount >= distribution_amount {
        distribution_amount
    } else {
        remaining_amount
    }
```

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,493 @@
...
+    storage::write(available_amount_uref, new_available_amount);
+
+    let remaining_amount_uref = get_uref_with_user_errors(
+        REMAINING_AMOUNT,
+        FaucetError::MissingRemainingAmount,
+        FaucetError::InvalidRemainingAmount,
+    );
+    storage::write(remaining_amount_uref, new_available_amount);
+
+    let new_distributions_per_interval: u64 =
```
Fraser999: I think we should either fail here if the caller sets the distributions per interval to `0`, or else we should handle it better in `get_distribution_amount()` where it's used as a denominator.

Maybe setting it to `0` should result in `get_distribution_amount()` returning `0`.

If not, then I don't think it's worth having `get_distribution_amount()` always reverting with an error if this is `0`.  Seems like it would effectively disable the main purpose of the faucet contract, in which case I think it'd be better to not allow it to be set to `0` in this function.

Whichever approach we take, we should add a test case for it.

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,493 @@
...
+        FaucetError::InvalidDistributionsPerInterval,
+    );
+
+    let distributions_per_interval: u64 = read_with_user_errors(
+        distributions_per_interval_uref,
+        FaucetError::MissingDistributionsPerInterval,
+        FaucetError::InvalidDistributionsPerInterval,
+    );
+
+    Ratio::new(available_amount, U512::from(distributions_per_interval)).to_integer()
```
Fraser999:
```suggestion
    available_amount / U512::from(distributions_per_interval)
```
then we can remove the `num` crate dependency.

smart_contracts/contracts/explorer/faucet-stored/src/main.rs
```diff
@@ -28,32 +42,74 @@ fn store() -> (ContractHash, ContractVersion) {
...
 
-        let entry_point = EntryPoint::new(
-            ENTRY_POINT_NAME,
+        let faucet = EntryPoint::new(
+            ENTRY_POINT_FAUCET,
+            vec![Parameter::new(
+                ARG_ID,
+                CLType::Option(Box::new(CLType::U64)),
+            )],
+            CLType::U512,
```
Fraser999:
```suggestion
            CLType::Unit,
```

smart_contracts/contracts/explorer/faucet-stored/src/main.rs
```diff
@@ -28,32 +42,74 @@ fn store() -> (ContractHash, ContractVersion) {
     let entry_points = {
         let mut entry_points = EntryPoints::new();
 
-        let entry_point = EntryPoint::new(
-            ENTRY_POINT_NAME,
+        let faucet = EntryPoint::new(
+            ENTRY_POINT_FAUCET,
+            vec![Parameter::new(
+                ARG_ID,
```
Fraser999: I think we should make the relevant consts public in the lib so we can use them here, e.g.
```suggestion
                faucet::ARG_ID,
```

smart_contracts/contracts/explorer/faucet-stored/src/main.rs
```diff
@@ -28,32 +42,74 @@ fn store() -> (ContractHash, ContractVersion) {
...
+            EntryPointType::Contract,
+        );
+
+        let set_variables = EntryPoint::new(
+            ENTRY_POINT_SET_VARIABLES,
             vec![
-                Parameter::new(ARG_TARGET, AccountHash::cl_type()),
-                Parameter::new(ARG_AMOUNT, CLType::U512),
+                Parameter::new(ARG_AVAILABLE_AMOUNT, CLType::U512),
+                Parameter::new(ARG_TIME_INTERVAL, CLType::U64),
```
Fraser999:
```suggestion
                Parameter::new(faucet::ARG_AVAILABLE_AMOUNT, CLType::U512),
                Parameter::new(faucet::ARG_TIME_INTERVAL, CLType::U64),
                Parameter::new(faucet::ARG_DISTRIBUTIONS_PER_INTERVAL, CLType::U64),
```

smart_contracts/contracts/explorer/faucet-stored/src/main.rs
```diff
@@ -3,20 +3,34 @@
...
 
+#[repr(u16)]
+enum InstallerSessionError {
+    FailedToTransfer = 101,
+}
+
+const ARG_ID: &str = "id";
+const ARG_AMOUNT: &str = "amount";
+const ARG_AVAILABLE_AMOUNT: &str = "available_amount";
+const ARG_TIME_INTERVAL: &str = "time_increment";
```
Fraser999: If we import these from the faucet lib, these should be deleted.  However, if we don't do that, then we need to fix `ARG_TIME_INTERVAL` to be `"time_interval"`.

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -1,77 +1,1210 @@
...
 };
-use casper_types::{account::AccountHash, runtime_args, ApiError, RuntimeArgs, U512};
+use casper_types::{
+    account::AccountHash, crypto, runtime_args, system::mint, ContractHash, Key, PublicKey,
+    RuntimeArgs, SecretKey, U512,
+};
+
+const FAUCET_INSTALLER_SESSION: &str = "faucet_stored.wasm";
+const FAUCET_CONTRACT_NAMED_KEY: &str = "faucet";
+// const FAUCET_CONTRACT_PACKAGE_NAMED_KEY: &str = "faucet_package";
```
Fraser999:
```suggestion
```

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -1,77 +1,1210 @@
...
+    let time_interval = builder
+        .query(
+            None,
+            faucet_contract_hash.into(),
+            &[TIME_INTERVAL_NAMED_KEY.to_string()],
+        )
+        .expect("failed to find time interval named key")
+        .as_cl_value()
+        .cloned()
+        .expect("failed to conert to cl value")
```
Fraser999:
```suggestion
        .expect("failed to convert to cl value")
```
There are another few similar typos - suggest you find and replace `conert`.

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -1,77 +1,1210 @@
...
+        let deploy_item = DeployItemBuilder::new()
+            .with_address(installer_account)
+            .with_authorization_keys(&[installer_account])
+            .with_stored_session_named_key(
+                FAUCET_CONTRACT_NAMED_KEY,
+                ENTRY_POINT_SET_VARIABLES,
+                runtime_args! {
+                    ARG_AVAILABLE_AMOUNT => U512::from(500_000u64),
+                    ARG_TIME_INTERVAL => 10_000u64,
+                    ARG_DISTRIBUTIONS_PER_INTERVAL => 2u64
```
Fraser999: Let's make these three values into variables so they're not duplicated in the subsequent assertions.

Applies to a few other tests here too.

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -1,77 +1,1210 @@
...
+    let faucet_call_by_installer = {
+        let deploy_item = DeployItemBuilder::new()
+            .with_address(installer_account)
+            .with_authorization_keys(&[installer_account])
+            .with_stored_session_named_key(
+                FAUCET_CONTRACT_NAMED_KEY,
+                ENTRY_POINT_FAUCET,
+                runtime_args! {ARG_TARGET => user_account, ARG_ID => <Option<u64>>::None},
+            )
+            // U512::from(25_000_000_000u64)
```
Fraser999:
```suggestion
```

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -1,77 +1,1210 @@
...
+    let faucet_call_by_user = {
+        let deploy_item = DeployItemBuilder::new()
+            .with_address(user_account)
+            .with_authorization_keys(&[user_account])
+            .with_stored_session_hash(
+                faucet_contract_hash,
+                ENTRY_POINT_FAUCET,
+                runtime_args! {ARG_TARGET => user_account, ARG_ID => <Option<u64>>::None},
+            )
+            // U512::from(25_000_000_000u64)
```
Fraser999:
```suggestion
```

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -1,77 +1,1210 @@
...
+        .expect("failed to convert into U512");
+
+    assert_eq!(remaining_amount, U512::from(100_000_000_000_000u64));
+    let user_main_purse_balance_after =
+        builder.get_purse_balance(builder.get_expected_account(user_account).main_purse());
+
+    assert_eq!(
+        user_main_purse_balance_after,
+        U512::from(200_000_000_000_000u64) - *DEFAULT_PAYMENT
     );
```
Fraser999, line 763: I guess we could also add
```rust
    let faucet_purse_balance = builder.get_purse_balance(faucet_purse);
    assert!(faucet_purse_balance.is_zero());
```

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -1,77 +1,1210 @@
...
+    // check the balance of the user's main purse
+    let user_main_purse_balance_after =
+        builder.get_purse_balance(builder.get_expected_account(user_account).main_purse());
+    //                 available_amount
+    //                 ________________
+    //                /
+    //               /  distributions_per_interval
+    //              /   __________________________
+    //  ___________/  _/
+    // (500_000_000 / 2 = 250_000_000) * 3 = 750_000_000
```
Fraser999: Given that this is the very end of the last test, it's a little odd to find the first fairly explicit comment here :)

Probably worth just removing this I think.

Fraser999 (commented):
A few small issues to sort, but no major problems.

goral09:
What's the cost of fauceting and does it increase with new calls to the contract?

smart_contracts/contracts/explorer/faucet/src/lib.rs
nilz3ro: I think it makes sense to have the `get_distribution_amount()` function return `0` if the max distributions per interval are `0`.

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,513 @@
...
+}
+
+impl From<FaucetError> for ApiError {
+    fn from(e: FaucetError) -> Self {
+        ApiError::User(e as u16)
+    }
+}
+
+#[no_mangle]
+pub fn init() {
```
mpapierski: You could move this to installer contract, and instead of making it an entry point it could be something like `fn init() -> NamedKeys.` you could use in `storage::new_contract`

nilz3ro: We can't do that because we need the contract to create its own purse and store it in its named keys, then return the purse `URef` to the installer. Do you mean there's a way to do that from the installer contract?

mpapierski: Installer contract can create a purse through `system::create_purse,` and then you can store it in named keys used for `storage::new_contract.`

nilz3ro:
> What's the cost of fauceting and does it increase with new calls to the contract?

@goral09 

| feature | cost |
|---------|------|
| faucet install | `35_129_051_410`|
| faucet set variables | `53_432_090`|
| faucet call by installer | `2_597_128_310`|
| faucet call by user | `2_546_174_240`|

smart_contracts/contracts/explorer/faucet/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Calling the Faucet
+
+The faucet will calculate a distribution amount as a ratio of the available amount per interval to the max distribution amount per interval. As an example, if the installer sets the available amount per interval to `100_000_000` and the max distributions per interval to `2`. When an existing user calls the faucet, `50_000_000` motes will be distributed to the caller. If a second user calls the faucet, they will also receive `50_000_000` motes. The remaining amount will now be `0` tokens. If a third user calls the faucet then they will not receive any token.
+After an interval passes after then last user was funded, the available amount will be replenished.
+
+`distributions_per_interval`, `available_amount`, `time_interval` and `max_distributions_per_interval`
+must be set and must be a number greater than `0` for the contract to run properly.
```
ihor, line 31: What is `max_distributions_per_interval`? I'm not sure I saw it in the README or the code.

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,498 @@
...
+pub const HASH_KEY_NAME: &str = "faucet_package";
+pub const ACCESS_KEY_NAME: &str = "faucet_package_access";
+pub const CONTRACT_VERSION: &str = "contract_version";
 
-#[repr(u32)]
-enum CustomError {
-    AlreadyFunded = 1,
+#[repr(u16)]
+enum FaucetError {
+    InvalidAccount = 1,
```
ihor, line 44: Based on the usage, a better name for this error would be `NotAuthorized` or something like that

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,498 @@
...
+    match api_error::result_from(ret) {
+        Ok(_) => Some(arg_size),
+        Err(ApiError::MissingArgument) => None,
+        Err(e) => runtime::revert(e),
+    }
+}
+
+fn get_optional_named_arg_with_user_errors<T: FromBytes>(
+    name: &str,
+    missing: FaucetError,
```
ihor, line 367: Is this argument needed here? It confused me because the method name says that the argument is supposed to be `optional`

smart_contracts/contracts/explorer/faucet/src/lib.rs
```diff
@@ -1,39 +1,498 @@
...
+
+    let distributions_per_interval: u64 = read_with_user_errors(
+        distributions_per_interval_uref,
+        FaucetError::MissingDistributionsPerInterval,
+        FaucetError::InvalidDistributionsPerInterval,
+    );
+
+    if distributions_per_interval == 0 {
+        runtime::revert(FaucetError::ZeroDistributionsPerInterval);
+    }
```
ihor, line 281: Shouldn't we do this validation here if this function is used only to get distribution amout to be transfered by the installer-account? 

Since, the limitations aren't applied to the installer, this piece of code doesn't seem relevant here, right?

ihor:
I started checking the code and noticed that the approach with `available_amount` and `distributions_per_interval` makes it harder to explain and follow the contract logic. Is there any reason to go with those configurations and not switch to:
- `credit_amount` - amount credited by the Faucet contract, always the same and predictable (!) to the user
- `credits_per_block` - number of credits per block, which is a known time-unit on the chain and should be easier to communicate to users

This way it will be easier to explain how the contract works, and it also will be easier to write the logic.

ihor:
Actually, nvm what I said before.

The Faucet is meant to give tokens to new accounts to be able to test something on the network. However, how do you call this Faucet contract if you don't have an account with tokens? cc @nilz3ro @EdHastingsCasperLabs @goral09

smart_contracts/contracts/explorer/faucet/README.md
nilz3ro: `max_distributions_per_interval` is the number of distributions that you'd like the faucet to make (maximum) per `time_interval`

smart_contracts/contracts/explorer/faucet/src/lib.rs
nilz3ro: @ihor I think we should allow the installer to pass an `amount` argument and then we don't need to calculate an amount for them at all.

PR #2553, opened 2022-01-07
Perform upgrade/genesis process in joiner reactor

### LINKED ISSUE #2548: Complete upgrade step before transitioning to participator

Fraser999:
This PR moves the logic for executing the `commit_upgrade`/`commit_genesis` and the subsequent switch block creation out of the constructor of the participating reactor.

The logic is moved to a new component `LinearChainSynchronizer` which has responsibility for running the fast sync process or for running the `commit_upgrade`/`commit_genesis` work as appropriate.  The existing fast sync async task is essentially unchanged.

I've tested this using the `upgrade_scenario_01.sh` script (modified with fast-sync updates from [this commit](https://github.com/TomVasile/casper-node/commit/3b774ce0b0dcbf8f71e616c6dc0aa6e9edc7cab6)) and it seems to perform the same as current `feat-fast-sync`.

Closes #2548.


node/src/components/contract_runtime/error.rs
```diff
@@ -23,7 +26,7 @@ pub(crate) enum ConfigError {
 }
 
 /// An error raised by a contract runtime variant.
```
afck: (Not your fault, but I don't understand that formulation; "contract runtime variant"? Maybe just: "An error during block execution"?)

node/src/components/linear_chain/state.rs
```diff
@@ -12,7 +12,7 @@ use super::{
     pending_signatures::PendingSignatures, signature::Signature, signature_cache::SignatureCache,
 };
 use crate::{
-    components::{consensus, linear_chain_sync::KeyBlockInfo},
```
afck, line 15: I wonder whether we should just call it `chain_synchronizer`?
I don't think we call the proposed blocks "chain" anywhere, so it's not really ambiguous.

node/src/reactor/participating.rs
```diff
@@ -575,22 +569,73 @@ impl reactor::Reactor for Reactor {
...
+                        maybe_step_effect_and_upcoming_era_validators,
+                    },
+            } => {
+                // The outcome of joining in this case caused a new switch block to be created, so
+                // we need to emit the effects which would have been created by that execution, but
+                // add them to the participating reactor's event queues so they don't get dropped as
+                // the joining reactor shuts down.
+                effects.extend(
+                    effect_builder
+                        .announce_new_linear_chain_block(block.clone(), execution_results)
```
afck, line 604: We should check (maybe as a separate ticket) whether the era supervisor will actually create a finality signature for an immediate switch block.

afck (approved):
Looks great! No blockers. :+1:

node/src/components/linear_chain_synchronizer/event.rs
```diff
@@ -0,0 +1,56 @@
...
+    contract_runtime::{BlockAndExecutionEffects, BlockExecutionError},
+    types::{BlockHash, BlockHeader},
+};
+
+#[derive(Debug, Serialize)]
+pub(crate) enum Event {
+    /// The result of getting the highest block from storage.
+    HighestBlockHash(Option<BlockHash>),
+    /// The result of the fast sync task.
+    SyncResult(Result<BlockHeader, Error>),
```
rafal-ch: Maybe `FastSyncResult`?

rafal-ch (approved):
LGTM with just one minor nit :+1:

node/src/reactor/participating.rs
Fraser999: Sure - let's do that in a separate ticket

node/src/components/linear_chain_synchronizer/event.rs
Fraser999: I'll probably go the other way and remove the `fast` from the doc comment.  I think it's clearer to just stick to calling this "chain sync" throughout (as per @afck feedback) rather than a mixture of "linear chain sync" and "fast sync".

node/src/components/contract_runtime/error.rs
Fraser999: Fixed in [`2d3671c2`](https://github.com/casper-network/casper-node/pull/2553/commits/2d3671c2e1cf45578eb1b096edba71d4fa347994).

node/src/components/linear_chain/state.rs
Fraser999: Fixed in [`2d3671c2`](https://github.com/casper-network/casper-node/pull/2553/commits/2d3671c2e1cf45578eb1b096edba71d4fa347994).

node/src/components/linear_chain_synchronizer/event.rs
Fraser999: Fixed in [`2d3671c2`](https://github.com/casper-network/casper-node/pull/2553/commits/2d3671c2e1cf45578eb1b096edba71d4fa347994).

PR #2563, opened 2022-01-11
Shut down if trusted block is from an old version.

### LINKED ISSUE #2253: Ensure graceful exit if provided with pre-upgrade trusted hash

afck:
Fast sync doesn't have the capability to sync forward across an upgrade boundary yet, so we shut down right away if the trusted block has a version lower than the current one and isn't the last block before the upgrade.

Closes #2253.

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -200,6 +200,16 @@ async fn get_trusted_key_block_info(
             block_header_with_future_version: Box::new(trusted_header.clone()),
         });
     }
+    // If the trusted block's version is older than ours we also return an error, except if we are
+    // at the current activation point, i.e. at an upgrade.
+    if trusted_header.protocol_version() < chainspec.protocol_config.version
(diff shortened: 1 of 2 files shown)


Fraser999 (approved):

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1081,31 +1072,5 @@ mod tests {
             }
             result => panic!("expected future block version error, got {:?}", result),
         }
-
-        // If the current version is 1.4.0 but the current block's is 1.3.0, we have to downgrade.
```
goral09, line 1085: If we're removing this unit test, shouldn't we also be removing some logic that was causing the downgrade?

goral09 (approved):

node/src/components/chain_synchronizer/operations.rs
afck: We are! That code is removed from `check_block_version` above.

goral09: 🙈 somehow I totally mised that

PR #2576, opened 2022-01-13
Limit incoming connections per peer

marc-casperlabs:
Limits incoming to a maximum of 3 (default setting) per peer, identified by unique node ID. A value of 0 disables the feature.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -445,6 +445,25 @@ where
                 peer_consensus_public_key,
                 stream,
             } => {
+                if self.cfg.max_incoming_peer_connections != 0 {
+                    if let Some(symmetries) = self.connection_symmetries.get(&peer_id) {
+                        let incoming_count = symmetries
+                            .incoming_addrs()
+                            .map(|addrs| addrs.len())
+                            .unwrap_or_default();
+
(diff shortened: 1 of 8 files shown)


afck (approved):

node/src/components/small_network.rs
```diff
@@ -445,6 +445,25 @@ where
                 peer_consensus_public_key,
                 stream,
             } => {
+                if self.cfg.max_incoming_peer_connections != 0 {
+                    if let Some(symmetries) = self.connection_symmetries.get(&peer_id) {
+                        let incoming_count = symmetries
+                            .incoming_addrs()
+                            .map(|addrs| addrs.len())
```
goral09, line 452: Just wondering, why would a single peer (IP) try to connect to us on multiple sockets?

goral09 (approved):

node/src/components/small_network.rs
marc-casperlabs: Bugs, attacks or race conditions due to shutdown/reconnection. All edge cases, basically, that do not occur on well-behaved nodes. Still, this needed an upper bound.

goral09:
bors retry

goral09:
bors retry

PR #2577, opened 2022-01-13
Add `nctl` test for disconnecting from nodes that speak different protocol version

rafal-ch:
This PR adds NCTL test for the following scenario:

- Start nodes 1-5 in V1
- Stage upgrade to V2 for node-6 only
- Run node-6
- Assert it logs error about the protocol version mismatch


rafal-ch:
This code has already been reviewed by @TomVasile here: https://github.com/TomVasile/casper-node/pull/11

TomVasile (approved):
tested locally and it ran fine

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_08.sh
```diff
@@ -0,0 +1,135 @@
...
+    log_step_upgrades 5 "asserting error messages in node log"
+
+    log "... allowing 15 seconds for the node to report errors"
+    sleep 15
+
+    COUNT=$(cat "$NCTL"/assets/net-1/nodes/node-"$NODE_ID"/logs/stdout.log 2>/dev/null | grep 'peer is running incompatible version' | wc -l)
+    if [ "$COUNT" -eq "0" ]; then
+        log "ERROR: We got 0 error messages, but expected some"
+        exit 1
+    fi
```
goral09: It would be good to verify that it not only logs an error but also not connects. That can be done by quering the `/status` endpoint and counting the # of entries under `.peers`.

rafal-ch: Took some time, but the issue is finally addressed here: https://github.com/casper-network/casper-node/pull/2577/commits/f26af83f206ea9b3526d17df657c3e737445f9e4

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_08.sh
```diff
@@ -98,17 +99,33 @@ function _step_05()
...
-# Step 06: Terminate.
+# Step 06: Assert joiner node is not connected to any peers
 function _step_06()
 {
-    log_step_upgrades 6 "test successful - tidying up"
+    local NODE_ID=${1}
+
+    log_step_upgrades 6 "asserting node has no peers connected"
+
+    local COUNT=$(curl -s "$(get_node_address_rest "$NODE_ID")/status" | jq '.peers' | jq length)
```
goral09: If it's not too difficult to move that up into NCTL tooling then it would be great - something like `get_node_peers` ?

rafal-ch: Fixed in https://github.com/casper-network/casper-node/pull/2577/commits/8b18fe0f1d33510790b27b37c86cae1165ebc17f

goral09 (approved):

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_08.sh
goral09: Thanks!

PR #2578, opened 2022-01-13
Random order symmetric peers

marc-casperlabs:
Ensure we only select symmetrically connected peers when syncing.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -860,7 +860,7 @@ where
                 NetworkInfoRequest::GetPeers { responder } => {
                     responder.respond(self.peers()).ignore()
                 }
-                NetworkInfoRequest::GetPeersInRandomOrder { responder } => {
+                NetworkInfoRequest::GetFullyConnectedPeers { responder } => {
                     let peers_vec: Vec<NodeId> = self
                         .connection_symmetries
                         .iter()
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -56,7 +56,7 @@ use futures::{future::BoxFuture, FutureExt};
 use openssl::{error::ErrorStack as OpenSslErrorStack, pkey};
 use pkey::{PKey, Private};
 use prometheus::Registry;
-use rand::seq::IteratorRandom;
+use rand::{seq::IteratorRandom, prelude::SliceRandom};
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 use tokio::{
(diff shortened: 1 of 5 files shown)


node/src/components/small_network.rs
```diff
@@ -860,10 +860,16 @@ where
...
-                    peers_vec.shuffle(rng);
-                    responder.respond(peers_vec).ignore()
+                NetworkInfoRequest::GetFullyConnectedPeers { responder } => {
+                    let symmetric_peers: Vec<NodeId> = self
+                        .connection_symmetries
+                        .iter()
+                        .filter_map(|(node_id, sym)| {
+                            matches!(sym, ConnectionSymmetry::Symmetric { .. }).then(|| *node_id)
+                        })
+                        .collect();
```
goral09, line 870: Why not add a method on `enum ConnectionSymmetries` to return a collection of symmetric connections?

node/src/effect.rs
```diff
@@ -624,15 +624,15 @@ impl<REv> EffectBuilder<REv> {
...
+    pub async fn get_fully_connected_peers<I>(self) -> Vec<I>
     where
         REv: From<NetworkInfoRequest<I>>,
         I: Send + 'static,
     {
         self.make_request(
-            |responder| NetworkInfoRequest::GetPeersInRandomOrder { responder },
-            QueueKind::Api,
+            |responder| NetworkInfoRequest::GetFullyConnectedPeers { responder },
+            QueueKind::Regular,
```
afck, line 635: Good catch! `Api` made the priority much lower, did it?

goral09 (approved):

afck (approved):

node/src/effect.rs
marc-casperlabs: In a way. It made it share resources with any external API requests.

node/src/components/small_network.rs
marc-casperlabs: There's no `ConnectionSymmetries`. The `ConnectionSymmetry` enum holds only connections for a single peer. The only method here would be something like `ConnectionSymmetries::is_symmetric`, which would contain the `matches!`. I opted against that for now, I'd add if I needed the functionality again.

goral09: Ah, b/c we want to extract the `node_id`s that are keys in the hashmap of `connection_symmetries` in which a value is of type `ConnectionSymmetry` 😵

afck (approved):

Fraser999 (approved):

afck (approved):
Also looks fine to me; but why did you switch to this approach?

Fraser999 (approved):

goral09:
bors retry

goral09:
bors retry

PR #2587, opened 2022-01-14
NO-TICKET: Bump rust toolchain in smart contracts

mpapierski:
This PR updates `rust-toolchain` in `./smart_contracts` to a date of recent stable release (2022-01-13) from an outdated one (2021-06)


goral09:
How can we verify that this won't cause problems downstream?

node/src/components/deploy_acceptor.rs
```diff
@@ -493,10 +493,7 @@ impl DeployAcceptor {
                     })
             }
             ExecutableDeployItemIdentifier::Package(
-                ref
-                contract_package_identifier
-                @
-                ContractPackageIdentifier::Hash {
+                ref contract_package_identifier @ ContractPackageIdentifier::Hash {
```
Fraser999: Since these two codeblocks happen to be the only blockers to using older versions of stable Rust, could you just change them to:
```rust
ExecutableDeployItemIdentifier::Package(ContractPackageIdentifier::Hash {
    contract_package_hash,
    version: maybe_package_version,
}) => {
```
and then remove the line
```rust
let maybe_package_version = contract_package_identifier.version();
```

types/src/block_time.rs
```diff
@@ -17,6 +17,7 @@ impl BlockTime {
 
     /// Saturating integer subtraction. Computes `self - other`, saturating at `0` instead of
     /// overflowing.
+    #[must_use]
```
Fraser999, line 20: Did the compiler suggest you add these `#[must_use]`s?

mpapierski:
```
error: missing `#[must_use]` attribute on a method returning `Self`
  --> types/src/block_time.rs:21:5
   |
21 | /     pub fn saturating_sub(self, other: BlockTime) -> Self {
22 | |         BlockTime(self.0.saturating_sub(other.0))
23 | |     }
   | |_____^
```

node/src/components/deploy_acceptor.rs
mpapierski: Addressed

Fraser999 (approved):

mpapierski:
@cspramit fyi this is something users should be aware of in next release

cspramit:
@mpapierski will be communicated, however is this a breaking change?

PR #2590, opened 2022-01-14
GH-2443: Get auction info backwards compatibility

### LINKED ISSUE #2443: The casper-client cannot retrieve get-auction-info for blocks from version 1.3.4 and before.
FWIW - I tried this both with casper-client v1.3.2  and casper-client v1.4.2 - so it’s not a backward compatibility issue in a newer client. The older client also wasn’t able to see further back than 1.4.1

Reporter Michael / Make

mpapierski:
Closes #2443 

This PR fixes a problem of the `get-auction-state` query for pre #1814 (introduction of `Key::SystemContractRegistry`) by passing down a state root hash to a ContractRuntime component based on the highest block (if found). 

The query requests a registry on the EE side using a user-provided state root hash. If there's no registry, the query uses a fallback state hash. Any further use of the system contract registry uses the cached value rather than always queried.

In a particular case, when block storage can't provide a fallback state root hash (i.e., new networks with no block) to create new ContractRuntime, it is internally set after genesis in EE or set after the upgrade.

### DIFF
--- a/execution_engine/src/core/execution/executor.rs
+++ b/execution_engine/src/core/execution/executor.rs
@@ -9,17 +9,17 @@ use casper_types::{
     bytesrepr::FromBytes,
     contracts::NamedKeys,
     system::{auction, handle_payment, mint, AUCTION, HANDLE_PAYMENT, MINT},
-    BlockTime, CLTyped, CLValue, ContractPackage, DeployHash, EntryPoint, EntryPointType, Gas, Key,
-    Phase, ProtocolVersion, RuntimeArgs, StoredValue,
+    BlockTime, CLTyped, ContractPackage, DeployHash, EntryPoint, EntryPointType, Gas, Key, Phase,
+    ProtocolVersion, RuntimeArgs, StoredValue,
 };
 
 use crate::{
     core::{
-        engine_state::{execution_result::ExecutionResult, EngineConfig},
+        engine_state::{execution_result::ExecutionResult, EngineConfig, SystemContractRegistry},
         execution::{address_generator::AddressGenerator, Error},
         runtime::{extract_access_rights_from_keys, instance_and_memory, Runtime, RuntimeStack},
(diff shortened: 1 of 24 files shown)


goral09:
Personally, I think this is a wrong approach. It unnecessarily couples the `ContractRuntime` component - we're changing a whole component (its fields) just to handle one old use-case. I think we should handle that in the higher layers where we know about protocol version and can use a correct logic for handling the request.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -986,24 +986,29 @@ where
...
+        era_id: EraId,
+    ) -> Option<ValidatorWeights> {
+        let mut result = self.get_era_validators(state_root_hash);
         result.remove(&era_id)
     }
 
     /// Gets [`Bids`].
-    pub fn get_bids(&mut self) -> Bids {
-        let get_bids_request = GetBidsRequest::new(self.get_post_state_hash());
+    pub fn get_bids(&mut self, state_root_hash: Option<Digest>) -> Bids {
```
goral09, line 1009: Please add comments to all three methods what the optional `state_root_hash` represents.

execution_engine/src/core/execution/executor.rs
```diff
@@ -485,86 +485,6 @@ impl Executor {
         ret
     }
 
-    /// Used to execute arbitrary wasm; necessary for running system contract installers / upgraders
-    /// This is not meant to be used for executing system contracts.
-    pub fn exec_wasm_direct<R, T>(
```
goral09, line 490: Was this method unused all along or is it possible to remove it only as part of this PR?

Fraser999 (commented):
I'd like to discuss the changeset in this PR soon to see if we can mitigate the breaking changes to the EE and test-support crates.

mpapierski:
Closing in favor of much simpler #2712

PR #2595, opened 2022-01-17
Handshake timeout

marc-casperlabs:
This PR makes the existing handshake timeout on both incoming and outgoing connections (and both directions on these) configurable, as the value was hardcoded to 20 seconds before.

Adding a timeout to incoming connections has been dropped, as with the merging of #2576, the amount of resources that can be occupied is limited anyway.


node/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.  The format
 * Detection of a crash no longer triggers DB integrity checks to run on node start; the checks can be triggered manually instead.
 * `SIGUSR1` now only dumps the queue in the debug text format.
 * Incoming connections from peers are rejected if they are exceeding the default incoming connections per peer limit of 3.
+* Connection handshake timeouts can now be configured (they were hardcoded at 20 seconds before).
```
goral09: Please add name of the config variable.

goral09 (approved):

PR #2599, opened 2022-01-19
Add Redelegate to auction

darthsiroftardis:
CHANGELOG:

- Added a new `Redelegate` entrypoint to the auction allowing users to re-delegate to another validator without having to unbond.
- Added a `new_validator` field to `UnbondingPurses` which allows redelegation without unbonding, this introduces extensions to the RPC API and SSE data.
-   Added a new struct `WithdrawPurse` to represent unbonding purses in current/live networks
-   Added a new `Key::Unbond` under which `UnbondingPurses` are written
-   Amended the upgrade logic to transform previous purses that are still be paid out into `UnbondingPurses`
-   Amended the upgrade logic to upgrade system contracts when there is a change in the system contract entrypoints


execution_engine/CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ All notable changes to this project will be documented in this file.  The format
 ## [Unreleased]
 
 ### Changed
-
+* Undelegate now takes an optional `new_validator` argument which will re-delegate to a validator without unbonding.
```
mpapierski, line 17: > Added a `redelegate` entry point to auction system contract to re-delegate without unbonding

And put it in `Added` section

types/src/system/auction/providers.rs
```diff
@@ -93,6 +93,16 @@ pub trait MintProvider {
     /// Reduce total supply by `amount`. Returns unit on success, otherwise
     /// an error.
     fn reduce_total_supply(&mut self, amount: U512) -> Result<(), Error>;
+
+    /// Transfers token from a given source URef to a bonding purse.
+    /// Returns the number of tokens currently delegated to a given validator.
+    fn handle_delegation(
```
mpapierski, line 99: This method should be a a free function in auction's `detail.rs` as MintProvider trait is meant to be a bridge between Auction contract and Mint system contract

types/src/system/auction/detail.rs
```diff
@@ -205,10 +205,39 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
...
+                                        .handle_delegation(
+                                            unbonding_purse.unbonder_public_key().clone(),
+                                            new_validator.clone(),
+                                            *unbonding_purse.bonding_purse(),
+                                            *unbonding_purse.amount(),
+                                        )
+                                        .map(|_| ())?
+                                } else {
+                                    // Move funds from bid purse to unbonding purse
+                                    provider
```
mpapierski, line 223: There must be a way to rewrite this fallback statement for better readability? This call is repeated 4 times

README.md
```diff
@@ -223,23 +203,6 @@ Count number of events in each queue:
...
-
-Entering `help` will show available commands. The interface can also be scripted, it may be helpful to change some of the per-connection settings in this case:
-
-```sh
-echo -e 'set -q true -o bincode\nexample-command' | socat - unix-client:/path/to/debug.socket > output.dump
-```
-
-This calls `example-command` on the console, but disables the command confirmation (`-q true`) and changes the output format to bincode (`-o bincode`) before doing so.
-
-
```
goral09, line 242: Not sure why changes to this file are part of the PR.

darthsiroftardis: It may have been introduced unintentionally as part of resolving merge conflicts

goral09: There are more changes that seem to be stemming from a bad merge.

types/src/system/auction/detail.rs
darthsiroftardis: I've asked for a better way to structure this, haven't found one. FWIW I am not a fan of how it looks either

darthsiroftardis:
Closes in favor of #2610

PR #2600, opened 2022-01-20
Rename chainspec setting to block_hash_v2_activation.

afck:
The `block_hash_v2_activation` setting will not be used for trie leaves, and continue to only affect block hashes.


rafal-ch (approved):
LGTM :+1:

afck:
Closing this, as the new block hashing scheme will not be enabled in 1.5.0.

PR #2270, opened 2021-10-26
Fixes integer overflow in `u8_slice_to_bytes`.

luis-casper:
Introduces a new variant of `ApiError`.

Closes #2260

PR #2602, opened 2022-01-20
GH-2260: Explicit overflow safety when casting

### LINKED ISSUE #2260: Fix integer overflow in u8_slice_to_bytes.
Instead of a conversion using as, prefer the TryFrom trait in https://github.com/casper-network/casper-node/blob/release-1.3.4/types/src/bytesrepr.rs#L1100

mpapierski:
Closes #2260
Closes #2270 

This commit adds explicit conversion from `usize` to `u32` whenever possible and handles the failure appropriately for each host function.

### DIFF
--- a/execution_engine/src/core/runtime/mod.rs
+++ b/execution_engine/src/core/runtime/mod.rs
@@ -11,7 +11,7 @@ mod standard_payment_internal;
 use std::{
     cmp,
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
-    convert::TryFrom,
+    convert::{TryFrom, TryInto},
     iter::{FromIterator, IntoIterator},
 };
 
@@ -1186,8 +1186,11 @@ where
             return Err(Error::Interpreter(error.into()).into());
         }
 
-        // For all practical purposes following cast is assumed to be safe
-        let bytes_size = key_bytes.len() as u32;
+        // For all practical purposes following conversion is assumed to be safe
+        let bytes_size: u32 = key_bytes
+            .len()
+            .try_into()
+            .expect("Keys should not serialize to many bytes");
         let size_bytes = bytes_size.to_le_bytes(); // Wasm is little-endian
         if let Err(error) = self.memory.set(bytes_written_ptr, &size_bytes) {
             return Err(Error::Interpreter(error.into()).into());
@@ -1303,7 +1306,10 @@ where
             return Ok(Err(ApiError::HostBufferFull));
         }
         let call_stack = self.stack.call_stack_elements();
-        let call_stack_len = call_stack.len() as u32;
+        let call_stack_len: u32 = match call_stack.len().try_into() {
+            Ok(value) => value,
+            Err(_) => return Ok(Err(ApiError::OutOfMemory)),
+        };
         let call_stack_len_bytes = call_stack_len.to_le_bytes();
 
         if let Err(error) = self.memory.set(call_stack_len_ptr, &call_stack_len_bytes) {
@@ -1316,7 +1322,12 @@ where
 
         let call_stack_cl_value = CLValue::from_t(call_stack.clone()).map_err(Error::CLValue)?;
 
-        let call_stack_cl_value_bytes_len = call_stack_cl_value.inner_bytes().len() as u32;
+        let call_stack_cl_value_bytes_len: u32 =
+            match call_stack_cl_value.inner_bytes().len().try_into() {
+                Ok(value) => value,
+                Err(_) => return Ok(Err(ApiError::OutOfMemory)),
+            };
+
         if let Err(error) = self.write_host_buffer(call_stack_cl_value) {
             return Ok(Err(error));
         }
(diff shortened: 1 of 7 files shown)


execution_engine/src/core/runtime/mod.rs
```diff
@@ -1186,8 +1186,11 @@ where
             return Err(Error::Interpreter(error.into()).into());
         }
 
-        // For all practical purposes following cast is assumed to be safe
-        let bytes_size = key_bytes.len() as u32;
+        // For all practical purposes following conversion is assumed to be safe
+        let bytes_size: u32 = key_bytes
+            .len()
+            .try_into()
+            .expect("Keys should not serialize to many bytes");
```
goral09, line 1193: Everywhere else in this file when the conversion fails we return `Ok(Err(ApiError::OutOfMemory))`. Why not do that here as well?

types/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.  The format
 ### Changed
 * Disable checksummed-hex encoding, but leave checksummed-hex decoding in place.
 * Increase `DICTIONARY_ITEM_KEY_MAX_LENGTH` to 128.
+* Fixed some integer casts.
```
goral09, line 22: We need to mention the new error variant and communicate that to the downstream clients.

execution_engine/src/core/runtime/mod.rs
mpapierski: If suddenly `impl ToBytes for Key` overflowed `u32::MAX` we'd know earlier through our test suite. This is  serialization output we control, so this is justified IMO

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3325,11 +3358,11 @@ where
         let name_bytes = self.bytes_from_mem(name_ptr, name_size)?;
         let name = String::from_utf8_lossy(&name_bytes);
 
-        let arg_size = match self.context.args().get(&name) {
+        let arg_size: u32 = match self.context.args().get(&name) {
             Some(arg) if arg.inner_bytes().len() > u32::max_value() as usize => {
                 return Ok(Err(ApiError::OutOfMemory));
             }
-            Some(arg) => arg.inner_bytes().len() as u32,
+            Some(arg) => arg.inner_bytes().len().try_into().unwrap(),
```
mpapierski: expect

execution_engine/src/core/runtime/mod.rs
```diff
@@ -2770,7 +2795,11 @@ where
             None => return Ok(Err(ApiError::ValueNotFound)),
         };
 
-        let value_size = cl_value.inner_bytes().len() as u32;
+        let value_size: u32 = match cl_value.inner_bytes().len().try_into() {
+            Ok(value) => value,
+            Err(_) => return Ok(Err(ApiError::BufferTooSmall)),
```
mpapierski, line 2800: Comment on safety

execution_engine/src/core/runtime/mod.rs
```diff
@@ -3325,11 +3358,17 @@ where
...
         let name = String::from_utf8_lossy(&name_bytes);
 
-        let arg_size = match self.context.args().get(&name) {
+        let arg_size: u32 = match self.context.args().get(&name) {
             Some(arg) if arg.inner_bytes().len() > u32::max_value() as usize => {
                 return Ok(Err(ApiError::OutOfMemory));
             }
-            Some(arg) => arg.inner_bytes().len() as u32,
+            Some(arg) => {
+                // SAFETY: Safe to unwrap as wee asserted length above
```
Fraser999:
```suggestion
                // SAFETY: Safe to unwrap as we asserted length above
```

node/src/components/block_proposer/tests.rs
```diff
@@ -5,6 +5,7 @@ use casper_types::{
     bytesrepr::Bytes, runtime_args, system::standard_payment::ARG_AMOUNT, Gas, RuntimeArgs,
     SecretKey,
 };
+use core::convert::TryInto;
```
Fraser999: We normally just use `std`.  Also, could the import order get fixed please?

types/CHANGELOG.md
Fraser999: Seems like this wasn't actioned?

Fraser999 (approved):
Optimistic approval given the few remaining comments are addressed.

types/CHANGELOG.md
mpapierski: Added an entry describing newly added variant

execution_engine/src/core/runtime/mod.rs
mpapierski: Done

node/src/components/block_proposer/tests.rs
mpapierski: Did this

PR #2616, opened 2022-01-21
Add Redelegate to auction

darthsiroftardis:
CHANGELOG:

- Added a new `Redelegate` entrypoint to the auction allowing users to re-delegate to another validator without having to unbond.
- Added a `new_validator` field to `UnbondingPurses` which allows redelegation without unbonding, this introduces extensions to the RPC API and SSE data.
-   Added a new struct `WithdrawPurse` to represent unbonding purses in current/live networks
-   Added a new `Key::Unbond` under which `UnbondingPurses` are written
-   Amended the upgrade logic to transform previous purses that are still be paid out into `UnbondingPurses`
-   Amended the upgrade logic to upgrade system contracts when there is a change in the system contract entrypoints


execution_engine_testing/tests/src/test/regression/ee_1152.rs
```diff
@@ -107,6 +107,7 @@ fn should_run_ee_1152_regression_test() {
             auction::ARG_DELEGATOR => DELEGATOR_1.clone(),
             auction::ARG_VALIDATOR => VALIDATOR_1.clone(),
             auction::ARG_AMOUNT => U512::from(DELEGATE_AMOUNT),
+            auction::ARG_NEW_VALIDATOR => Option::<PublicKey>::None
```
goral09: if we have to update all callsites, doesn't it mean users will also have to do that? Either upgrade version of the SDK or change the code manually themselve?

goral09: And it definitely means that if they don't do that, deploys produced by their SDKs/code will be incompatible with Ledger.

execution_engine/src/shared/transform.rs
```diff
@@ -360,9 +365,12 @@ impl From<&Transform> for casper_types::Transform {
             Transform::Write(StoredValue::Bid(bid)) => {
                 casper_types::Transform::WriteBid(bid.clone())
             }
-            Transform::Write(StoredValue::Withdraw(unbonding_purses)) => {
+            Transform::Write(StoredValue::Unbonding(unbonding_purses)) => {
                 casper_types::Transform::WriteWithdraw(unbonding_purses.clone())
             }
+            Transform::Write(StoredValue::Withdraw(_)) => casper_types::Transform::Failure(
+                "withdraw purses should not be be written to global state".to_string(),
```
goral09, line 372: `StoredValue::Withdraw` is the only variant that fails this transformation. Can you explain why it should not be written to global state?

resources/test/sse_data_schema.json
```diff
@@ -1873,6 +1874,17 @@
               "$ref": "#/definitions/U512"
             }
           ]
+        },
+        "new_validator": {
```
goral09, line 1878: This needs to be mentioned in the changelog. This is also backwards incompatible change as old clients cannot continue to parse this event without patching.

/cc @Fraser999

execution_engine/src/shared/transform.rs
darthsiroftardis: `WithdrawPurses` are the old representation of UnbondingPurses, therefore moving forward we shouldn't write those variants to global state

execution_engine_testing/tests/src/test/regression/ee_1152.rs
darthsiroftardis: No, this is incorrect, undelegate takes only three args: `validator`, `delegator` and `amount`. No one should need to update their SDKs to call `undelegate`. The test works because the extra argument passed here gets ignored, but thank you for pointing it out

smart_contracts/contracts_as/client/redelegate/assembly/index.ts
```diff
@@ -0,0 +1,78 @@
...
+        Error.fromErrorCode(ErrorCode.InvalidArgument).revert();
+        return;
+    }
+    let newValidator = newValidatorResult.value;
+
+    let runtimeArgs = RuntimeArgs.fromArray([
+        new Pair(ARG_AMOUNT, CLValue.fromU512(amount)),
+        new Pair(ARG_DELEGATOR, CLValue.fromPublicKey(delegator)),
+        new Pair(ARG_VALIDATOR, CLValue.fromPublicKey(validator)),
+        new Pair(ARG_NEW_VALIDATOR, CLValue.fromPublicKey(newValidator)),
```
goral09, line 75: Does it mean, this argument is also ignored? If it's ignored, why not remove it?

darthsiroftardis: This is in the call to redelegate and we do expect four arguments here (undelegate's three args + new_validator) so it is not ignored here

goral09: I see, so the other one was `delegate` which simply ignores the new argument and here is `redelegate` which does not do it 👍

execution_engine_testing/tests/src/test/regression/ee_1152.rs
goral09: Does it mean it can be safely removed from the list of arguments to `undelegate` endpoint? Meaning, all the regression tests.

types/src/system/auction/providers.rs
```diff
@@ -43,10 +43,10 @@ pub trait StorageProvider {
     fn write_bid(&mut self, account_hash: AccountHash, bid: Bid) -> Result<(), Error>;
 
     /// Reads collection of [`UnbondingPurse`]s at account hash derived from given public key
-    fn read_withdraw(&mut self, account_hash: &AccountHash) -> Result<Vec<UnbondingPurse>, Error>;
+    fn read_unbond(&mut self, account_hash: &AccountHash) -> Result<Vec<UnbondingPurse>, Error>;
```
goral09, line 46: This renaming (and `get_withdraw() -> get_unbonds()`) could have been introduced as a separate PR.

types/src/system/auction/providers.rs
```diff
@@ -93,6 +93,16 @@ pub trait MintProvider {
     /// Reduce total supply by `amount`. Returns unit on success, otherwise
     /// an error.
     fn reduce_total_supply(&mut self, amount: U512) -> Result<(), Error>;
+
+    /// Transfers token from a given source URef to a bonding purse.
+    /// Returns the number of tokens currently delegated to a given validator.
+    fn handle_delegation(
```
goral09: I don't think `Mint` should be coupled with auction-like logic (`delegation` is specific to auction and has nothing to do with minting/destroying/transferring tokens)

execution_engine_testing/tests/src/test/regression/ee_1152.rs
darthsiroftardis: Aye, I see that its still using them in places, will remove it from the list of arguments

execution_engine_testing/tests/src/test/system_costs.rs
```diff
@@ -386,7 +422,8 @@ fn delegate_and_undelegate_have_expected_costs() {
         runtime_args! {
             auction::ARG_DELEGATOR => DEFAULT_ACCOUNT_PUBLIC_KEY.clone(),
             auction::ARG_VALIDATOR => VALIDATOR_1.clone(),
-            auction::ARG_AMOUNT => U512::from(BID_AMOUNT),
+            auction::ARG_AMOUNT => U512::from(BID_AMOUNT - 10u64),
+            auction::ARG_NEW_VALIDATOR => Option::<PublicKey>::None,
```
goral09: Apparently, this argument is ignored and therefore unnecessary.

execution_engine_testing/tests/src/test/system_costs.rs
```diff
@@ -532,7 +603,8 @@ fn upgraded_delegate_and_undelegate_have_expected_costs() {
...
```
goral09: Same here, should be safe to remove?

node/src/types/json_compatibility/stored_value.rs
```diff
@@ -44,7 +44,9 @@ pub enum StoredValue {
     /// A bid
     Bid(Box<Bid>),
     /// A withdraw
-    Withdraw(Vec<UnbondingPurse>),
+    Withdraw(Vec<WithdrawPurse>),
```
goral09, line 47: Won't this change the already-existing events/data schemes? I would expect that JSON representation is derived via `serde-json` + `Serialize` and that would autogenerate the names from the types/variants?

/cc @Fraser999

smart_contracts/contracts/client/redelegate/src/main.rs
```diff
@@ -0,0 +1,37 @@
+#![no_std]
```
goral09, line 1: Just wondering, is this contract being used anywhere right now? I.e. to validate that it indeed does what it's expected to?

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -576,6 +577,37 @@ where
...
+                    *registry
+                        .get(STANDARD_PAYMENT)
+                        .expect("should have standard payment hash"),
+                )
+            }
+            if self.auction_contract_hash.is_none() {
+                self.auction_contract_hash =
+                    Some(*registry.get(AUCTION).expect("should have auction hash"))
+            }
+        }
```
goral09, line 609: It's not clear to me why it worked previously and doesn't do anymore. What has changed that we now have to populate `self.*` manually?

types/src/system/auction/entry_points.rs
```diff
@@ -76,6 +77,20 @@ pub fn auction_entry_points() -> EntryPoints {
     );
     entry_points.add_entry_point(entry_point);
 
+    let entry_point = EntryPoint::new(
+        METHOD_REDELEGATE,
+        vec![
+            Parameter::new(ARG_DELEGATOR, PublicKey::cl_type()),
+            Parameter::new(ARG_VALIDATOR, PublicKey::cl_type()),
+            Parameter::new(ARG_AMOUNT, U512::cl_type()),
+            Parameter::new(ARG_NEW_VALIDATOR, Option::<PublicKey>::cl_type()),
```
goral09: if someone calles `redelegate`, shouldn't the `ARG_NEW_VALIDATOR` argument be required?

resources/test/rpc_schema_hashing_V1.json
```diff
@@ -2245,7 +2245,7 @@
                 "properties": {
                   "Withdraw": {
                     "items": {
-                      "$ref": "#/components/schemas/UnbondingPurse"
+                      "$ref": "#/components/schemas/WithdrawPurse"
```
goral09, line 2248: I am not sure how people use these schemas and whether their pipeline is scripted enough to recognize that the path for the type's schema has changed.

/cc @Fraser999

types/src/system/auction/providers.rs
darthsiroftardis: Well, the handling of delegation requires that token be transferred to the necessary bonding purses and so I feel that its okay for it to be here

goral09: This method could be implemented on an `Auction` trait directly:
```rust
pub trait Auction:
    StorageProvider + RuntimeProvider + MintProvider + AccountProvider + Sized
{
```
which means that implementations of methods in that trait will have access to other methods available on `Mint*` and `Runtime*` and `Storage*`. That should be enough to implement correct logic and not burden `Mint` with auction-related logic and concepts.

smart_contracts/contracts/client/redelegate/src/main.rs
darthsiroftardis: Yes, it is being used in the tests, under the auction/bid.rs

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -429,6 +431,80 @@ where
...
+                let withdraw_purses = tracking_copy
+                    .borrow_mut()
+                    .read(correlation_id, &key)
+                    .map_err(|_| Error::FailedToGetWithdrawKeys)?
+                    .ok_or(Error::FailedToGetStoredWithdraws)?
+                    .as_withdraw()
+                    .ok_or(Error::FailedToGetWithdrawPurses)?
+                    .to_owned();
+
+                let unbonding_purses: Vec<UnbondingPurse> = withdraw_purses
```
goral09, line 488: I guess that suggests our design here is not ideal if we have all withdraw purses in the queue, even though it should contain only the ones that are still to be processed (i.e. we don't have to iterate over purses that had been already).

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -184,37 +185,49 @@ where
...
             new_protocol_version,
+            old_protocol_version,
             tracking_copy,
         }
     }
 
-    /// Bump major version for system contracts.
-    pub(crate) fn upgrade_system_contracts_major_version(
+    /// Check if the next version is a major version bump.
+    pub(crate) fn is_major_bump(&self) -> bool {
```
goral09: Instead of reimplementing this method, let's use ones defined on `ProtocolVersion`:
```rust
old_protocol_version.check_next_version(&new_protocol_version).is_major_version()
```

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -184,37 +185,49 @@ where
...
         &self,
         correlation_id: CorrelationId,
         mint_hash: &ContractHash,
         auction_hash: &ContractHash,
         handle_payment_hash: &ContractHash,
         standard_payment_hash: &ContractHash,
     ) -> Result<(), ProtocolUpgradeError> {
-        self.store_contract(correlation_id, *mint_hash, MINT, mint::mint_entry_points())?;
-        self.store_contract(
+        self.refresh_system_contract_entry_points(
```
goral09, line 203: Let's pattern match on `SystemContractType` and handle all of its variants so that we have additional check that makes sure we handle all the cases.

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -224,20 +237,46 @@ where
...
+                        return true;
+                    }
+                }
+            }
+        }
+        // Return true when Entry points have been removed.
+        !contract
+            .entry_points()
+            .keys()
+            .eq(entry_points_to_check.keys())
```
goral09: Since we already check that all `entry_points_to_check` exist in `contract.entry_points()` then it should be enough to compare lengths `contract.entry_points().len() == entry_points.to_check.keys().len()`.

execution_engine/src/core/runtime/auction_internal.rs
```diff
@@ -76,9 +76,9 @@ where
             .map_err(|exec_error| <Option<Error>>::from(exec_error).unwrap_or(Error::Storage))
     }
 
-    fn read_withdraw(&mut self, account_hash: &AccountHash) -> Result<Vec<UnbondingPurse>, Error> {
-        match self.context.read_gs(&Key::Withdraw(*account_hash)) {
-            Ok(Some(StoredValue::Withdraw(unbonding_purses))) => Ok(unbonding_purses),
+    fn read_unbond(&mut self, account_hash: &AccountHash) -> Result<Vec<UnbondingPurse>, Error> {
```
goral09, line 79: I'm now wondering whether it won't become a problem, similar to auction state, where after an upgrade new node won't be able to load auction state (in this case unbonding purses) from before it.

types/src/system/auction/detail.rs
```diff
@@ -205,10 +205,39 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
...
             // was calculated on `unbond` attempt.
             if current_era_id >= unbonding_purse.era_of_creation() + unbonding_delay {
-                // Move funds from bid purse to unbonding purse
-                provider
-                    .unbond(unbonding_purse)
-                    .map_err(|_| Error::TransferToUnbondingPurse)?;
+                match unbonding_purse.new_validator() {
+                    Some(new_validator) => {
+                        match provider.read_bid(&new_validator.to_account_hash()) {
+                            Ok(Some(bid)) => {
```
goral09:
```suggestion
                            Ok(Some(new_validator_bid)) => {
```

types/src/system/auction/detail.rs
```diff
@@ -205,10 +205,39 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
...
+                                            *unbonding_purse.bonding_purse(),
+                                            *unbonding_purse.amount(),
+                                        )
+                                        .map(|_| ())?
+                                } else {
+                                    // Move funds from bid purse to unbonding purse
+                                    provider
+                                        .unbond(unbonding_purse)
+                                        .map_err(|_| Error::TransferToUnbondingPurse)?
+                                }
```
goral09, line 227: So we unbond if a new delegator has a bid and that his current stake is zero?

types/src/system/auction/detail.rs
```diff
@@ -205,10 +205,39 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
...
+                            }
+                            // Move funds from bid purse to unbonding purse
+                            Ok(_) | Err(_) => provider
```
goral09:
```suggestion
                            Ok(None) | Err(_) => provider
```

goral09: We should also log the error.

goral09: Personally, I think if the unbonding request is invalid - in this case the new validator is not bonded or there's an error - we should make it a no-op and not do anything.

execution_engine/CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ All notable changes to this project will be documented in this file.  The format
 ## [Unreleased]
 
 ### Changed
-
+* Undelegate now takes an optional `new_validator` argument which will re-delegate to a validator without unbonding.
```
mpapierski: Isn't that more like

> Added new auction entrypoint `redelegate` that allows to move delegated funds to different validator

execution_engine/src/core/runtime/auction_internal.rs
```diff
@@ -292,6 +292,66 @@ where
         self.mint_reduce_total_supply(mint_contract, amount)
             .map_err(|exec_error| <Option<Error>>::from(exec_error).unwrap_or(Error::MintReward))
     }
+
+    fn handle_delegation(
```
mpapierski: I followed up offline re: MintProvider responsibilities

execution_engine_testing/tests/src/test/regression/ee_1120.rs
```diff
@@ -202,6 +204,7 @@ fn should_run_ee_1120_slash_delegators() {
             ARG_AMOUNT => U512::from(UNDELEGATE_AMOUNT_3),
             ARG_VALIDATOR => VALIDATOR_1.clone(),
             ARG_DELEGATOR => VALIDATOR_2.clone(),
+            ARG_NEW_VALIDATOR => Option::<PublicKey>::None
```
mpapierski: Is that still necessary though. It's an extra parameter to undelegate which is not handled there (merge conflicts?). I think you should revert ARG_NEW_VALIDATOR parameters from undelegate calls

types/src/system/auction/detail.rs
mpapierski: I think this is correct behavior, though. What if I'll redelegate to a validator, but my validator withdrew himself in the meantime. Not doing anything here would likely lose my stake? ~But I agree that we should log errors for the `Err` case~ not sure if logging is possible from within casper-types crate. @darthsiroftardis might be a good idea to write a test for this scenario if you didn't write it yet

types/src/system/auction/detail.rs
```diff
@@ -205,10 +205,39 @@ pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(provider: &mut P) ->
...
+                                .unbond(unbonding_purse)
+                                .map_err(|_| Error::TransferToUnbondingPurse)?,
+                        }
+                    }
+                    None => {
+                        // Move funds from bid purse to unbonding purse
+                        provider
+                            .unbond(unbonding_purse)
+                            .map_err(|_| Error::TransferToUnbondingPurse)?
```
mpapierski, line 239: It'd be great to reorganize error handling here as `unbond()` is called three times at different indentation levels.

types/src/system/auction/entry_points.rs
mpapierski:
```suggestion
            Parameter::new(ARG_NEW_VALIDATOR, PublicKey::cl_type()),
```

types/src/system/auction/providers.rs
mpapierski: I'd prefer a different approach where `handle_delegation` is in `detail.rs` IMO it neither belongs to Auction trait nor MintProvider

execution_engine_testing/test_support/src/wasm_test_builder.rs
darthsiroftardis: This is to compensate for introducing the system contract registry in an upgrade request, which we weren't running into our earlier test fixtures

types/src/system/auction/detail.rs
darthsiroftardis: Yes, if the stake is zero, the bid is marked inactive and we transfer the delegated tokens back to the delegator's main purse

execution_engine/src/core/engine_state/mod.rs
goral09: This isn't about changes in this PR - it's a general comment about auction design in general. If `withdraw_purses` were "keyed" (stored in the GlobalState) under a key that is specific to an era in which they should be paid out then that vector  would contain only those purses that needed to be paid out. No unnecessary `if` conditions would be required.

execution_engine/src/core/engine_state/upgrade.rs
darthsiroftardis: Keeping it as is, because `keys()` returns an iterator, so the `len` method isn't present

goral09: But `len()` should be available on `contract.entry_points()`.

execution_engine_testing/tests/src/test/regression/ee_1152.rs
darthsiroftardis: Removed unnecessary argument

types/src/system/auction/providers.rs
darthsiroftardis: Removed from the trait and used @mpapierski of putting it in `detail.rs`

execution_engine_testing/tests/src/test/system_costs.rs
darthsiroftardis: Removed unnecessary argument

execution_engine_testing/tests/src/test/system_costs.rs
darthsiroftardis: Removed unnecessary argument

types/src/system/auction/entry_points.rs
darthsiroftardis: It is required, and has been changed to reflect that

execution_engine/src/core/engine_state/upgrade.rs
darthsiroftardis: Using the method defined on `ProtocolVersion`

execution_engine/src/core/engine_state/upgrade.rs
darthsiroftardis: Using `SystemContractType`

types/src/system/auction/detail.rs
darthsiroftardis: Applied suggestion

types/src/system/auction/detail.rs
darthsiroftardis: Applied suggestion, and a test has been written to address @mpapierski 's concern

execution_engine/CHANGELOG.md
darthsiroftardis: Moved to the added section

execution_engine/src/core/runtime/auction_internal.rs
darthsiroftardis: Moved to `detail.rs`

execution_engine_testing/tests/src/test/regression/ee_1120.rs
darthsiroftardis: Removed unnecessary argument

types/src/system/auction/detail.rs
darthsiroftardis: Yeah, I can't seem to find a structure that can accomodate that

types/src/system/auction/entry_points.rs
darthsiroftardis: Applied suggestion

execution_engine/src/core/engine_state/upgrade.rs
darthsiroftardis: ``` no method named `len` found for reference `&EntryPoints` in the current scope```

goral09: `EntryPoints` is our type though and it's just a wrapper around `BTreeMap`, so we can add it. If it's equivalent to the current version then why not use the simpler and faster approach?

execution_engine/src/core/runtime/auction_internal.rs
darthsiroftardis: Discussed offline

resources/test/sse_data_schema.json
darthsiroftardis: The change has been noted here:
https://github.com/darthsiroftardis/casper-node/blob/Fix-redelegate/types/CHANGELOG.md#changed

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -224,20 +237,46 @@ where
         Ok(())
     }
 
-    /// Store new system contract.
-    fn store_contract(
+    /// Check if a contract's entry points differ from a given set.
+    fn has_different_entry_points(
```
Fraser999: It's not clear to me why we can't just avoid this method altogether and do something like `let entry_points_unchanged = contract.entry_points() == entry_points_to_check;`

types/src/system/auction/mod.rs
```diff
@@ -297,7 +296,65 @@ pub trait Auction:
...
+        };
+
+        self.write_bid(validator_account_hash, bid)?;
+
+        Ok(new_amount)
+    }
+
+    /// Removes specified amount of motes (or the value from the collection altogether, if the
+    /// remaining amount is 0) from the entry in delegators map for given validator and creates a
+    /// new unbonding request to the queue, which when proccessed will redelegate the
```
Fraser999: Typo: "proccessed" -> "processed".  Sentences should end in periods too.

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -6,12 +6,8 @@ use thiserror::Error;
...
 use casper_hashing::Digest;
 use casper_types::{
-    bytesrepr,
-    system::{
-        auction, handle_payment, mint, standard_payment, AUCTION, HANDLE_PAYMENT, MINT,
-        STANDARD_PAYMENT,
-    },
-    Contract, ContractHash, EntryPoints, EraId, Key, ProtocolVersion, StoredValue,
+    bytesrepr, system::SystemContractType, Contract, ContractHash, EntryPoints, EraId, Key,
+    ProtocolVersion, StoredValue,
```
Fraser999: There's a typo just below here on line 19: "successfuly" -> "successfully".

execution_engine/src/core/engine_state/upgrade.rs
darthsiroftardis: Applied suggestion

execution_engine/src/core/engine_state/upgrade.rs
darthsiroftardis: Fixed typo

types/src/system/auction/mod.rs
darthsiroftardis: Fixed typo + grammar

execution_engine/src/core/engine_state/upgrade.rs
darthsiroftardis: Used suggestion

Fraser999 (approved):

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -245,10 +245,7 @@ where
             }
         }
         // Return true when Entry points have been removed.
-        !contract
-            .entry_points()
-            .keys()
-            .eq(entry_points_to_check.keys())
+        !contract.entry_points().len() == entry_points_to_check.len()
```
goral09: Small optimization: Moving this comparison before the `for ... in ... { ... }` loop should improve performance. I.e. if lengths are different then it `has_different_entry_points == true`. If they're the same, we have to compare key-for-key.

execution_engine/src/core/engine_state/upgrade.rs
```diff
@@ -281,7 +257,8 @@ where
             .check_next_version(&self.new_protocol_version)
             .is_major_version();
 
-        if !self.has_different_entry_points(&contract, &entry_points) && !is_major_bump {
+        let entry_points_unchanged = *contract.entry_points() == entry_points;
```
goral09, line 260: Oh, I see you removed the `has_different_entry_points` entirely and make use of `impl Eq for EntryPoints` - I don't know what's the performance of that but hopefuly Rust also produces some smart code for the equality.

execution_engine/CHANGELOG.md
```diff
@@ -13,8 +13,10 @@ All notable changes to this project will be documented in this file.  The format
 
 ## [Unreleased]
 
-### Changed
+### Added
+* Added a new entry point `Redelegate` which allows users to redelegate to another validator without having to unbond.
```
goral09: This changelog feels incomplete - firstly, where did you add this entrypoint? I'm guessing `Auction` contract but that needs to be said here explicitly. That would be lowercase _redeegate_ as well. Please, include also the set of arguments and their type. Also, let's include the type signature of the new endpoint: 
```
redelegate(delegator: PublicKey, validator: PublicKey, amount: U512, new_validator: PublicKey)
```

goral09:
We should have a new integration test that asserts that unbond requests made across the new upgrade boundary (current protocol -> protocola after this PR) will work as expected. It would test both the migration logic the logic around handing unbonding queue.

Let's extend the docs (PR description) with a documentation about the API of the new endpoint - what are the arguments' names and types.

smart_contracts/contracts/client/redelegate/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+
+use casper_contract::contract_api::{runtime, system};
+use casper_types::{runtime_args, system::auction, PublicKey, RuntimeArgs, U512};
+
+const ARG_AMOUNT: &str = "amount";
+const ARG_DELEGATOR: &str = "delegator";
+const ARG_VALIDATOR: &str = "validator";
+const ARG_NEW_VALIDATOR: &str = "new_validator";
+
+fn redelegate(delegator: PublicKey, validator: PublicKey, amount: U512, new_validator: PublicKey) {
```
goral09, line 14: Is this necessary from the client PoV to redelegate? `ExecutableDeployItem::StoredContractByHash { auction_hash, "redelegate", runtime_args! { ... } }` should be equivalent but cheaper and simpler.

darthsiroftardis: Not necessarily, but this keeps the format as the other two client contracts of `delegate` and `undelegate` so I would prefer to keep this as is.

goral09: Yes, I understand we already do this for other endpoints. I was just making sure the cheaper version still works.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3068,3 +3082,999 @@ fn should_fail_with_more_accounts_than_slots() {
...
+fn should_delegate_and_redelegate() {
+    let system_fund_request = ExecuteRequestBuilder::standard(
+        *DEFAULT_ACCOUNT_ADDR,
+        CONTRACT_TRANSFER_TO_ACCOUNT,
+        runtime_args! {
+            ARG_TARGET => *SYSTEM_ADDR,
+            ARG_AMOUNT => U512::from(SYSTEM_TRANSFER_AMOUNT)
+        },
+    )
+    .build();
```
goral09, line 3097: We're writing this down so many times that I am surprised there's no helper method like
```
fund(from_account, to_account, amount)
```

goral09: If not in this PR, let's create a ticket to add most commonly-used helpers like `fund`, `bid` `(un)delegate`, `withdraw`, etc. and use them across all tests. If we keep postponing that, people will never change their habits.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3068,3 +3082,999 @@ fn should_fail_with_more_accounts_than_slots() {
...
+            .with_parent_state_hash(builder.get_post_state_hash())
+            .with_protocol_version(ProtocolVersion::V1_0_0)
+            .with_next_era_id(builder.get_era().successor())
+            .with_run_auction(true)
+            .build();
+
+        builder
+            .step(step_request)
+            .expect("must execute third step request post upgrade");
+    }
```
goral09, line 3190: It took a bit over 100 lines of code to setup the test. And honestly, when I get here I already forget how the test was set up.

It's not a problem with this particular test - just the process in general.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3068,3 +3082,999 @@ fn should_fail_with_more_accounts_than_slots() {
...
+            delegator_1_redelegate_purse_balance
+        );
+
+        let step_request = StepRequestBuilder::new()
+            .with_parent_state_hash(builder.get_post_state_hash())
+            .with_protocol_version(ProtocolVersion::V1_0_0)
+            .with_next_era_id(builder.get_era().successor())
+            .with_reward_item(RewardItem::new(NON_FOUNDER_VALIDATOR_1_PK.clone(), 1))
+            .with_reward_item(RewardItem::new(NON_FOUNDER_VALIDATOR_2_PK.clone(), 1))
+            .with_run_auction(true)
```
goral09: I can't imagine a situation where we wouldn't want to run an auction in `step` - can we make it `true` by default and not have to type it down explicitly?

execution_engine/CHANGELOG.md
darthsiroftardis: Added a clearer description to the CHANGELOG, including the signature

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3068,3 +3082,999 @@ fn should_fail_with_more_accounts_than_slots() {
...
 }
+
+#[ignore]
+#[test]
+fn should_delegate_and_redelegate() {
+    let system_fund_request = ExecuteRequestBuilder::standard(
+        *DEFAULT_ACCOUNT_ADDR,
+        CONTRACT_TRANSFER_TO_ACCOUNT,
+        runtime_args! {
+            ARG_TARGET => *SYSTEM_ADDR,
```
goral09, line 3093: Why not have the system account funded by default in tests. We do this in almost every single unit test.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3068,3 +3082,999 @@ fn should_fail_with_more_accounts_than_slots() {
...
+        },
+    )
+    .build();
+
+    let validator_2_fund_request = ExecuteRequestBuilder::standard(
+        *DEFAULT_ACCOUNT_ADDR,
+        CONTRACT_TRANSFER_TO_ACCOUNT,
+        runtime_args! {
+            ARG_TARGET => *NON_FOUNDER_VALIDATOR_2_ADDR,
+            ARG_AMOUNT => U512::from(TRANSFER_AMOUNT)
```
goral09, line 3114: Why not make `TRANSFER_AMOUNT` a `U512` type? Aren't you tired of writing `U512::from(TRANSFER_AMOUNT)` so many times?

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3068,3 +3082,999 @@ fn should_fail_with_more_accounts_than_slots() {
...
+
+    let unbond_list = unbonding_purses
+        .get(&NON_FOUNDER_VALIDATOR_1_ADDR)
+        .expect("should have unbonding purse for non founding validator");
+    assert_eq!(unbond_list.len(), 1);
+    assert_eq!(
+        unbond_list[0].validator_public_key(),
+        &*NON_FOUNDER_VALIDATOR_1_PK
+    );
+    assert!(unbond_list[0].new_validator().is_none())
```
goral09, line 3308: Personally I prefer an approach where we assert it something like:
```rust
match &unbond_list {
  &[UnbondingPurse { .. }] => { assertions }
  _ => { panic!("expected single purse in the unbonding queue")
}

```

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3068,3 +3082,999 @@ fn should_fail_with_more_accounts_than_slots() {
...
+}
+
+#[ignore]
+#[test]
+fn should_handle_redelegation_to_inactive_validator() {
```
goral09, line 3313: If my _inactive_ you mean not bonded then _inactive_ is a wrong term. Inactive validator is one that IS bonded but hasn't been producing blocks.

goral09 (approved):
The test are so freaking long. In some of them it takes hundreds of lines to setup the test and first `assert!`s are appearing after 200+ loc. They _seem_ to be OK.

I would prefer if we didn't change the old structures' names. Even if we now thing the names was not correct.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
darthsiroftardis: It is true by default actually, this was probably me being extra sure the auction was running in the early part of the testing. I can remove them from my tests

goral09: Is it? `new()` uses `default()` and default for bool is false.

darthsiroftardis: No, there is `default` implementation for the builder which sets it to true

Issue #1665, opened 2021-07-08
Fast Sync Must Work With Emergency Upgrades


xcthulhu:

xcthulhu:
For fast sync, we will need to download old blocks and deploys to avoid replay attacks.  Those blocks may not necessarily have been executed.

We can reasonably maintain the invariant that the _highest_ block in storage has been executed.  This may be sufficient. 

We should consider a way of storing deploys and correlating them to blocks without storing blocks, so we can preserve the invariant "If a block is in storage it has been executed"

xcthulhu:
Moreover, we have the following somewhat tricky logic that we might wish to sort out [here][1].

[1]: https://github.com/casper-network/casper-node/pull/1669/files#diff-212250b2151062263c01ec4511602e2ada0b8a83dd02b8727589737f3d3312ebR479-R482

xcthulhu:
The main outstanding issue with fast-sync right now is that it does not work with emergency restarts, so I've repurposed this ticket for that.

afck:
The following changes (suggested by @fizyk20) definitely seem necessary, but haven't fixed the test for me so far:

```diff
--- a/node/src/components/linear_chain_sync/operations.rs
+++ b/node/src/components/linear_chain_sync/operations.rs
@@ -206,10 +206,12 @@ async fn get_trusted_key_block_info(
     // This will crash if we try to get the parent hash of genesis, which is the default [0u8; 32]
     let mut current_header_to_walk_back_from = trusted_header.clone();
     loop {
-        // Check that we are not restarting right after an emergency restart, which is too early
+        // Blocks before an emergency restart should not be trusted based on their finality
+        // signatures. So the earliest allowed trusted hash is the last block before the upgrade.
         match chainspec.protocol_config.last_emergency_restart {
             Some(last_emergency_restart)
-                if last_emergency_restart > current_header_to_walk_back_from.era_id() =>
+                if last_emergency_restart
+                    > current_header_to_walk_back_from.next_block_era_id() =>
             {
                 return Err(LinearChainSyncError::TrustedHeaderEraTooEarly {
                     trusted_header: Box::new(trusted_header.clone()),
@@ -735,8 +737,10 @@ pub(crate) async fn run_fast_sync_task(
     }
 
     let maybe_last_emergency_restart_era_id = chainspec.protocol_config.last_emergency_restart;
+    // Blocks before an emergency restart should not be trusted based on their finality
[...]

PR #2619, opened 2022-01-22
chain synchronizer and storage update for emergency upgrades

Fraser999:
This PR updates the chain synchronizer to exit for an upgrade at the end of syncing if there is an installed higher version, and if the highest stored block indicates it is time to upgrade.

Please note that the health checks at the end of `utils/nctl/sh/scenarios/emergency_upgrade_test.sh` was modified to allow the test to pass.  The `restarts=20` was reduced to `10`.  It is worth investigating why this is the case.

Part of #1665.


node/src/components/chain_synchronizer.rs
```diff
@@ -110,6 +116,23 @@ impl ChainSynchronizer {
         match result {
             Ok(latest_block_header) => {
                 if latest_block_header.protocol_version() == self.chainspec.protocol_version() {
+                    if let Some(next_upgrade_activation_point) = self
+                        .maybe_next_upgrade
+                        .map(|next_upgrade| next_upgrade.era_id())
+                    {
+                        if latest_block_header.era_id() >= next_upgrade_activation_point {
```
afck: Shouldn't we rather stop before we've downloaded that block? Getting blocks for a future version seems unsafe.

Instead I'd make sure that we stop syncing if `latest_block_header.next_block_era_id() == next_upgrade_activation_point`, so that the new version creates the upgrade block and continues.

node/src/components/storage.rs
```diff
@@ -543,9 +542,7 @@ impl Storage {
                 // Remove blocks that are in to-be-upgraded eras, but have obsolete protocol
                 // versions - they were most likely created before the upgrade and should be
                 // reverted.
-                if block_header.era_id() >= invalid_era
-                    && block_header.protocol_version() < protocol_version
-                {
+                if block_header.era_id() >= invalid_era {
```
afck: But what if there was an emergency upgrade in the past and we've already correctly handled it and continued. If we then restart the node won't this remove a lot of perfectly valid blocks?

node/src/components/chainspec_loader.rs
```diff
@@ -210,7 +221,11 @@ impl ChainspecLoader {
         let mut effects = if should_stop {
             Effects::new()
         } else {
-            async { Some(Event::Initialize) }.map_some(std::convert::identity)
+            effect_builder
+                .get_highest_block_from_storage()
```
afck: Would it be enough to load the highest block _header_, without body?

node/src/components/chainspec_loader.rs
```diff
@@ -236,6 +247,49 @@ impl ChainspecLoader {
...
+    }
+
+    fn should_exit_for_upgrade(
+        maybe_highest_block: Option<Box<Block>>,
+        maybe_next_upgrade_activation_point: Option<ActivationPoint>,
+    ) -> bool {
+        maybe_highest_block.map_or(false, |highest_block| {
+            maybe_next_upgrade_activation_point.map_or(false, |next_upgrade_activation_point| {
+                let highest_block_header = highest_block.header();
+                if highest_block_header.era_id() >= next_upgrade_activation_point.era_id() {
```
afck: We could probably exit even if:

```suggestion
                if highest_block_header.next_block_era_id() >= next_upgrade_activation_point.era_id() {
```

node/src/components/chainspec_loader.rs
```diff
@@ -482,6 +540,85 @@ mod tests {
...
+            HEIGHT,
+            PROTOCOL_VERSION,
+            IS_SWITCH,
+            VERIFIABLE_CHUNKED_HASH_ACTIVATION.into(),
+        )));
+        let next_upgrade_activation_point = Some(ActivationPoint::EraId(2.into()));
+        assert!(ChainspecLoader::should_exit_for_upgrade(
+            highest_block,
+            next_upgrade_activation_point
+        ));
```
afck, line 640: I don't see the difference to the previous case.

afck (approved):
> The storage component's constructor is also modified to not consider protocol versions when deleting blocks due to a hard_reset.

That's not true anymore, is it?

node/src/components/chainspec_loader.rs
Fraser999: Yup - good point.  Same thing for the deploy acceptor - I'll make that change too since it's more frequently called and will be even more beneficial than this one which is only called once in a session.

node/src/components/chainspec_loader.rs
Fraser999: Yeah, fair point.  Would save us running the chain synchronizer in that case.

node/src/components/chainspec_loader.rs
Fraser999: Me neither - I'll update to check for the case of the highest block being the last switch block before the next upgrade activation point.

node/src/components/chainspec_loader.rs
Fraser999: Fixed in [148a037](https://github.com/casper-network/casper-node/pull/2619/commits/148a037bdc00338d76e60afd30ff1d4834a10b95).

node/src/components/chainspec_loader.rs
Fraser999: Fixed in [148a037](https://github.com/casper-network/casper-node/pull/2619/commits/148a037bdc00338d76e60afd30ff1d4834a10b95).

node/src/components/chainspec_loader.rs
Fraser999: Fixed in [148a037](https://github.com/casper-network/casper-node/pull/2619/commits/148a037bdc00338d76e60afd30ff1d4834a10b95).

node/src/components/chainspec_loader.rs
```diff
@@ -538,83 +539,103 @@ fn next_upgrade(dir: PathBuf, current_version: ProtocolVersion) -> Option<NextUp
...
+            Block::random_with_specifics(
+                &mut rng,
+                EraId::from(2),
+                HEIGHT,
+                PROTOCOL_VERSION,
+                IS_NOT_SWITCH,
+                VERIFIABLE_CHUNKED_HASH_ACTIVATION.into(),
+            )
+            .header()
+            .clone(),
```
afck, line 616: There's also a `Block::take_header` function.

afck (approved):

node/src/components/chainspec_loader.rs
Fraser999: Good tip - I'll remember for the future, but leave this as is, since it's only in a test.

Fraser999:
bors merge

Issue #2329, opened 2021-11-10
Prefer exhaustive pattern matching around ToBytes/FromBytes


piotr-dziubecki:
We should prefer exhaustive wildcard pattern matching and avoid wildcards (`_ => { something; }`) around ToBytes/FromBytes methods.

Recently we had to revert https://github.com/CasperLabs/casper-node/pull/1321 back to a state which includes `_ => U8_SERIALIZED_LENGTH`, which is potentially harmful.

Write new `proptests` and arb generators to validate this change.

Ideally, as part of this task, we should consider using a tag enum with #[derive(FromPrimitive)] over constants:

Instead of https://github.com/CasperLabs/casperlabs-node/blob/556a95a322641fbefdea67abf3da59c9d6ffb169/types/src/execution_result.rs#L50-L69

Something like:

```rust
#[derive(FromPrimitive)]
enum TransformTag {
  Identity = 0,
  CLValue = 1,
  // ...
}

/// ...
fn from_bytes(...) {
  let (tag_value, remainder) = u8::from_bytes(bytes)?;
  let tag = TransformTag::from_primitive(tag_value).ok_or(Error::Formatting)?;
  match tag {
    // ... avoid _ => and let your IDE fill in missing variants upon extending
  }
}
```

Fraser999:
@nilz3ro please get an update from @mpapierski before taking on this ticket

PR #2627, opened 2022-01-26
Exhaustive Pattern Matching in Deserialization

SamuelSchlesinger:
## Description
Improves ToBytes/FromBytes implementations so they are less likely to have bugs in them. In particular, use `enum`s and `FromPrimitive` to make sure we match on all cases of the `*Tag` constructor. This doesn't exactly fix everything, as implementors still need to add a `*Tag` variant, but I believe that seeing its lack in the `ToBytes` implementation will cause that. I had an idea like:

```rust
enum T<A, B, C> {
  A(A),
  B(B),
  C(C),
}
```

but I don't know how to make it `repr(u8)` _only_ when `A = B = C = ()`. Beyond that, there's the worse problem of not being able to specify the explicit numbers corresponding to each variant in this declaration. That may be fixed by the implicit derivation of `FromPrimitive`, haven't checked.

## Risks
Any faulty modification of the serialization layer could lead to serious bugs in the Casper Network, its users' code, or in any of our tooling.

## Issue
https://github.com/casper-network/casper-node/issues/2329

### DIFF
--- a/execution_engine/src/storage/trie/mod.rs
+++ b/execution_engine/src/storage/trie/mod.rs
@@ -1,11 +1,13 @@
 //! Core types for a Merkle Trie
 
 use std::{
-    convert::TryInto,
+    convert::{TryFrom, TryInto},
     fmt::{self, Debug, Display, Formatter},
     mem::MaybeUninit,
 };
 
+use num_derive::{FromPrimitive, ToPrimitive};
+use num_traits::cast::{FromPrimitive, ToPrimitive};
 use serde::{
     de::{self, MapAccess, Visitor},
     ser::SerializeMap,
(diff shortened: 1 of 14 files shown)


execution_engine/src/storage/trie/mod.rs
```diff
@@ -95,16 +104,18 @@ impl ToBytes for Pointer {
...
+        match FromPrimitive::from_u8(tag) {
+            Some(tag) => match tag {
+                0 => {
+                    let (hash, rem) = Digest::from_bytes(rem)?;
+                    Ok((Pointer::LeafPointer(hash), rem))
+                }
+                1 => {
+                    let (hash, rem) = Digest::from_bytes(rem)?;
+                    Ok((Pointer::NodePointer(hash), rem))
+                }
```
SamuelSchlesinger: Match on the actual constructors here.

execution_engine/src/storage/trie/mod.rs
```diff
@@ -460,27 +502,29 @@ where
...
-                    rem,
-                ))
-            }
-            2 => {
-                let (affix, rem) = FromBytes::from_bytes(rem)?;
-                let (pointer, rem) = Pointer::from_bytes(rem)?;
-                Ok((Trie::Extension { affix, pointer }, rem))
-            }
-            _ => Err(bytesrepr::Error::Formatting),
+        match FromPrimitive::from_u8(tag) {
```
goral09, line 505: In other places you have `match TryFrom::try_from(tag)?`, would that not work here?

SamuelSchlesinger: Yep, good catch. I believe `KeyTag` is used somewhat inconsistetly from the rest as well. The `TryFrom::try_from(tag)?` is what I intend them all to use before the PR is marked ready for merge, because it removes the explicit tag error case entirely from the deserializer.

SamuelSchlesinger:
Closing due to work on consensus taking precedent.

PR #2633, opened 2022-01-31
Don't assert on the number of leftover events when switching from `joiner` to `participating` in `MultiStageTestReactor`

rafal-ch:
This PR removes the assert on the number of leftover events to be handled by the joiner reactor. The assert is replaced with a warning, as the leftover events do not prevent the node from functioning correctly.

The assert could have triggered spuriously, because at the time of asserting there could still be other components putting the events into the queue.

Neither assert nor warning would be needed if we switch to a single reactor model in the future.

The idea to dump the remaining events into the logs was dropped, because it would require substantial effort and hacky code to get access to the internal data of the queues from within the non-async `dispatch_event()` function.


marc-casperlabs (approved):
LGTM!

Fraser999 (approved):

PR #2637, opened 2022-02-01
Add more detailed logs for fast sync task

rafal-ch:
This PR adds more detailed logging to the fast-sync task and it also changes the place where we log info about peer timeout.

The new log messages are paired like so:
```
"start - description"

"finish - description"
```

The most detailed messages are added on `debug` level, others are `info`.

It can potentially flood the logs, but I prefer to have more logs that I need than have too less. Additionally, some of these logs can be removed after the issues with the fast sync performance is diagnosed.

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -537,16 +537,16 @@ async fn fast_sync_to_most_recent(
     chainspec: &Chainspec,
     node_config: NodeConfig,
 ) -> Result<(KeyBlockInfo, BlockHeader), Error> {
-    info!("start - get_trusted_key_block_info - fast sync");
+    info!("start - fetch the last switch block before the trusted block - fast sync");
     let mut trusted_key_block_info =
         get_trusted_key_block_info(effect_builder, chainspec, &trusted_block_header).await?;
-    info!("finish - get_trusted_key_block_info - fast sync");
+    debug!("finish - fetch the last switch block before the trusted block - fast sync");
 
     // Get the most recent header which has the same version as ours
     // We keep fetching by height until none of our peers have a block at that height and we are in
     // the current era.
     let mut most_recent_block_header = trusted_block_header;
-    info!("start - most recent header - fast sync");
+    info!("start - fetch block headers up to the most recent one - fast sync");
     loop {
         let maybe_fetched_block = fetch_and_store_next::<BlockHeaderWithMetadata>(
             effect_builder,
@@ -578,11 +578,11 @@ async fn fast_sync_to_most_recent(
             break;
         }
     }
-    info!("finish - most recent header - fast sync");
+    debug!("finish - fetch block headers up to the most recent one - fast sync");
 
     // Fetch and store all blocks that can contain not-yet-expired deploys. These are needed for
     // replay detection.
-    info!("start - fetch and store all blocks - fast sync");
+    info!("start - fetch blocks for deploy replay protection - fast sync");
     {
         let mut current_header = most_recent_block_header.clone();
         while trusted_key_block_info
@@ -597,11 +597,11 @@ async fn fast_sync_to_most_recent(
                     .take_header();
         }
     }
-    info!("finish - fetch and store all blocks - fast sync");
+    debug!("finish - fetch blocks for deploy replay protection - fast sync");
 
     // The era supervisor requires enough switch blocks to be stored in the database to be able to
     // initialize the most recent eras.
-    info!("start - for era supervisor - fast sync");
+    info!("start - fetch block headers needed for era supervisor initialization - fast sync");
     {
         let earliest_open_era = chainspec.earliest_open_era(most_recent_block_header.era_id());
         let earliest_era_needed_by_era_supervisor =
@@ -615,17 +615,17 @@ async fn fast_sync_to_most_recent(
             .await?;
         }
     }
-    info!("finish - for era supervisor - fast sync");
+    debug!("finish - fetch block headers needed for era supervisor initialization - fast sync");
 
     // Synchronize the trie store for the most recent block header.
-    info!("start - sync_trie_store - fast sync");
+    info!("start - fetch the latest block's global state - fast sync");
     sync_trie_store(
         effect_builder,
         *most_recent_block_header.state_root_hash(),
         node_config.max_parallel_trie_fetches as usize,
     )
     .await?;
-    info!("finish - sync_trie_store - fast sync");
+    debug!("finish - fetch the latest block's global state - fast sync");
 
     Ok((trusted_key_block_info, most_recent_block_header))
 }
@@ -692,18 +692,18 @@ async fn archival_sync(
 ) -> Result<(KeyBlockInfo, BlockHeader), Error> {
     // Get the trusted block info. This will fail if we are trying to join with a trusted hash in
     // era 0.
-    info!("start - get_trusted_key_block_info - archival sync");
+    info!("start - fetch the latest key block before the trusted block - archival sync");
     let mut trusted_key_block_info =
         get_trusted_key_block_info(effect_builder, chainspec, &trusted_block_header).await?;
-    info!("finish - get_trusted_key_block_info - archival sync");
+    debug!("finish - fetch the latest key block before the trusted block - archival sync");
 
-    info!("start - fetch_and_store_block_by_hash - archival sync");
+    info!("start - fetch the trusted block - archival sync");
     let trusted_block = *fetch_and_store_block_by_hash(
         effect_builder,
         trusted_block_header.hash(chainspec.protocol_config.verifiable_chunked_hash_activation),
     )
     .await?;
-    info!("finish - fetch_and_store_block_by_hash - archival sync");
+    debug!("finish - fetch the trusted block - archival sync");
 
     // Sync to genesis
     let mut walkback_block = trusted_block.clone();
@@ -720,7 +720,7 @@ async fn archival_sync(
             .await?;
         }
     }
-    info!("finish - sync to genesis - archival sync");
+    debug!("finish - sync to genesis - archival sync");
 
     // Sync forward until we are at the current version.
     info!("start - sync forward - archival sync");
@@ -749,7 +749,7 @@ async fn archival_sync(
             trusted_key_block_info = key_block_info;
         }
     }
-    info!("finish - sync forward - archival sync");
+    debug!("finish - sync forward - archival sync");
 
     Ok((trusted_key_block_info, most_recent_block.take_header()))
 }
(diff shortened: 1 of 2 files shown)


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -537,13 +537,16 @@ async fn fast_sync_to_most_recent(
     chainspec: &Chainspec,
     node_config: NodeConfig,
 ) -> Result<(KeyBlockInfo, BlockHeader), Error> {
+    info!("start - get_trusted_key_block_info - fast sync");
```
```suggestion
    info!("start - fetch the last switch block before the trusted block  - fast sync");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -537,13 +537,16 @@ async fn fast_sync_to_most_recent(
...
     let mut trusted_key_block_info =
         get_trusted_key_block_info(effect_builder, chainspec, &trusted_block_header).await?;
+    info!("finish - get_trusted_key_block_info - fast sync");
 
     // Get the most recent header which has the same version as ours
     // We keep fetching by height until none of our peers have a block at that height and we are in
     // the current era.
     let mut most_recent_block_header = trusted_block_header;
+    info!("start - most recent header - fast sync");
```
```suggestion
    info!("start - fetch block headers up to the most recent one - fast sync");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -575,9 +578,11 @@ async fn fast_sync_to_most_recent(
             break;
         }
     }
+    info!("finish - most recent header - fast sync");
 
     // Fetch and store all blocks that can contain not-yet-expired deploys. These are needed for
     // replay detection.
+    info!("start - fetch and store all blocks - fast sync");
```
```suggestion
    info!("start - fetch blocks for deploy replay protection - fast sync");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -592,9 +597,11 @@ async fn fast_sync_to_most_recent(
                     .take_header();
         }
     }
+    info!("finish - fetch and store all blocks - fast sync");
 
     // The era supervisor requires enough switch blocks to be stored in the database to be able to
     // initialize the most recent eras.
+    info!("start - for era supervisor - fast sync");
```
```suggestion
    info!("start - fetch block headers needed for era supervisor initialization - fast sync");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -608,14 +615,17 @@ async fn fast_sync_to_most_recent(
             .await?;
         }
     }
+    info!("finish - for era supervisor - fast sync");
 
     // Synchronize the trie store for the most recent block header.
+    info!("start - sync_trie_store - fast sync");
```
```suggestion
    info!("start - fetch the latest block's global state - fast sync");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -592,9 +597,11 @@ async fn fast_sync_to_most_recent(
                     .take_header();
         }
     }
+    info!("finish - fetch and store all blocks - fast sync");
```
afck: I'd prefer making the INFO-level messages less noisy. We could omit all the "finish" ones since they are right before the next "start" ones anyway?
The last one could just say "finished syncing" or something?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -666,17 +692,22 @@ async fn archival_sync(
 ) -> Result<(KeyBlockInfo, BlockHeader), Error> {
     // Get the trusted block info. This will fail if we are trying to join with a trusted hash in
     // era 0.
+    info!("start - get_trusted_key_block_info - archival sync");
```
```suggestion
    info!("start - fetch the latest key block before the trusted block - archival sync");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -666,17 +692,22 @@ async fn archival_sync(
...
     let mut trusted_key_block_info =
         get_trusted_key_block_info(effect_builder, chainspec, &trusted_block_header).await?;
+    info!("finish - get_trusted_key_block_info - archival sync");
 
+    info!("start - fetch_and_store_block_by_hash - archival sync");
```
```suggestion
    info!("start - fetch the trusted block - archival sync");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -775,12 +811,14 @@ pub(super) async fn run_chain_sync_task(
         if trusted_block_header.is_switch_block()
             && trusted_block_header.next_block_era_id() == last_emergency_restart_era
         {
+            info!("start - sync_trie_store - emergency restart");
```
```suggestion
            info!("start - fetch the global state from before the last emergency restart - emergency restart");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -794,43 +832,56 @@ pub(super) async fn run_chain_sync_task(
         && trusted_block_header.next_block_era_id()
             == chainspec.protocol_config.activation_point.era_id()
     {
+        info!("start - get_trusted_key_block_info - upgrade");
```
```suggestion
        info!("start - fetch the last switch block before the trusted block - upgrade");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -794,43 +832,56 @@ pub(super) async fn run_chain_sync_task(
...
         let trusted_key_block_info =
             get_trusted_key_block_info(effect_builder, &*chainspec, &trusted_block_header).await?;
-        if fetch_and_store_next::<BlockHeaderWithMetadata>(
+        info!("finish - get_trusted_key_block_info - upgrade");
+
+        info!("start - fetch_and_store_next::<BlockHeaderWithMetadata> - upgrade");
```
```suggestion
        info!("start - try fetching the first block after the upgrade - upgrade");
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -794,43 +832,56 @@ pub(super) async fn run_chain_sync_task(
...
             &*chainspec,
         )
-        .await?
-        .is_none()
-        {
+        .await?;
+        info!("finish - fetch_and_store_next::<BlockHeaderWithMetadata> - upgrade");
+
+        if fetch_and_store_next_result.is_none() {
+            info!("start - sync_trie_store - upgrade");
```
```suggestion
            info!("start - fetch the global state from before the upgrade - upgrade");
```

afck (approved):
Looks good to me. A few suggestions, but those aren't blockers.

node/src/components/chain_synchronizer/operations.rs
rafal-ch: I'd keep them for now as I'd really like to avoid that we miss some important debug trace in the overnight test.

After all, they all should be removed, being replaced by proper measurement tools.

Let me just change the "finish" messages to `debug`-level.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1010,6 +1011,28 @@ pub(super) async fn run_chain_sync_task(
     Ok(most_recent_block_header)
 }
 
+async fn fetch_and_store_deploys(
```
afck, line 1014: That could also be used in `sync_deploys_and_transfers_and_state` now.

afck (approved):

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Certainly. I'll deduplicate the code in the PRs coming tomorrow.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -537,13 +537,16 @@ async fn fast_sync_to_most_recent(
     chainspec: &Chainspec,
     node_config: NodeConfig,
 ) -> Result<(KeyBlockInfo, BlockHeader), Error> {
+    info!("start - fetch the last switch block before the trusted block - fast sync");
     let mut trusted_key_block_info =
         get_trusted_key_block_info(effect_builder, chainspec, &trusted_block_header).await?;
+    debug!("finish - fetch the last switch block before the trusted block - fast sync");
```
goral09, line 543: Why not log at `INFO` just like the start message?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -633,18 +643,34 @@ async fn sync_deploys_and_transfers_and_state(
         .cloned()
         .collect();
     let mut stream = futures::stream::iter(hash_iter)
-        .map(|hash| fetch_and_store_deploy(effect_builder, hash))
+        .map(|hash| {
+            debug!("start - fetch_and_store_deploy - archival sync - {}", hash);
```
goral09, line 647: This is reversed, isn't it? Other logs log start at INFO and finish at DEBUG and here it's the opposite.

node/src/components/chain_synchronizer/operations.rs
afck: I thought it was unnecessarily noisy (https://github.com/casper-network/casper-node/pull/2637#discussion_r796833292). Happy to change it back if you disagree.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -866,14 +929,19 @@ pub(super) async fn run_chain_sync_task(
...
+            effect_builder,
+        )
+        .await?;
+
+        let transfers = fetch_and_store_deploys(
+            block.transfer_hashes().to_vec(),
+            max_parallel_deploy_fetches,
+            effect_builder,
+        )
+        .await?;
```
goral09, line 944: ```let (deploys, transfers) = tokio::join!(..., ...)```
would probably improve things even further.

goral09: Or ```futures::future::join```

afck: Or just pass the combined list into a single call of `fetch_and_store_deploys` instead of making two separate calls. That's what we do in `sync_deploys_and_transfers_and_state`.

node/src/components/chain_synchronizer/operations.rs
goral09: No, that's fine. I figured out (too late) that reading timestamps of `start ...` logs is enough to figure out the timeline.

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Unlike in `sync_deploys_and_transfers_and_state`, here we need a separate lists of `deploys` and `transfers` as an output (they are used later). We could probably read them all at once and the partition (assuming a deploy can identify itself as transfer or not).

As mentioned in the PR description, there are better ways. The time was short, though, so was the scope of the PR different :)

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Here both "start" and "finish" was intended as `debug` from the start, as this is expected to be called very often. See the PR description.

node/src/components/chain_synchronizer/operations.rs
rafal-ch: It's being prepared as a part of the bigger PR: https://github.com/rafal-ch/casper-node/commit/070f30d682d66f1581c042f8821919964cd5a946

PR #2638, opened 2022-02-02
Merge `dev` to `feat-fast-sync` 02.02.2022

rafal-ch:
This PR integrates the recent changes from the `dev` (as of 2.02.2022 12:00pm CEST) branch into the `feat-fast-sync` branch.


types/src/era_id.rs
```diff
@@ -59,6 +59,8 @@ impl EraId {
     ///
     /// For `u64::MAX`, this returns `u64::MAX` again: We want to make sure this doesn't panic, and
     /// that era number will never be reached in practice.
+    #[must_use]
+    #[allow(clippy::integer_arithmetic)] // The caller must make sure this doesn't overflow.
```
afck: On this branch we're using `saturating_add` inside, so this is unnecessary.

afck (approved):

types/src/era_id.rs
rafal-ch: Good catch! Fixed in https://github.com/casper-network/casper-node/pull/2638/commits/bf2d72e1a418a0f0016754553ab21bbe9ba40e6f

Fraser999 (approved):

PR #2639, opened 2022-02-02
Don't discard deploys without announcing expiry.

### LINKED ISSUE #2629: Don't drop expired deploys silently.
If a deploy expires we should emit an event.
BlockProposer::add_deploy should not silently drop an incoming deploy.

afck:
Don't silently drop deploys in `BlockProposer::add_deploy`.
Expired deploys are filtered by the deploy acceptor anyway, and are periodically purged, causing an announcement and event.

Also update the log message in `add_deploy` so it distinguishes between transfers and other deploys, like the other log messages.

Closes #2629.

### DIFF
--- a/node/src/components/block_proposer.rs
+++ b/node/src/components/block_proposer.rs
@@ -357,11 +357,6 @@ impl BlockProposerReady {
         hash: DeployOrTransferHash,
         deploy_info: DeployInfo,
     ) {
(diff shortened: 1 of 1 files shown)


goral09 (approved):

rafal-ch (approved):
LGTM :+1:

afck:
bors merge

PR #2640, opened 2022-02-02
Add metrics to `chain_synchronizer`

rafal-ch:
This PR adds the following metrics to the chain synchronization process:
* `chain_sync_total_duration_seconds`
* `chain_sync_emergency_restart_duration_seconds`
* `chain_sync_upgrade_duration_seconds`
* `chain_sync_archival_sync_total_duration_seconds`
* `chain_sync_get_trusted_key_block_info_duration_seconds`
* `chain_sync_sync_to_genesis_duration_seconds`
* `chain_sync_sync_forward_duration_seconds`
* `chain_sync_fast_sync_total_duration_seconds`
* `chain_sync_fetch_block_headers_duration_seconds`
* `chain_sync_replay_protection_duration_seconds`
* `chain_sync_era_supervisor_init_duration_seconds`
* `chain_sync_execute_blocks_duration_seconds`
* `chain_sync_fetch_and_store_initial_trusted_block_header_duration_seconds`
* `chain_sync_sync_trie_store_duration_seconds`

This PR also removes the temporary `start / finish` markers added to the logs.

This is an excerpt of the output from the `/metrics` endpoint:
```
...
# HELP chain_sync_archival_sync_total_duration_seconds total time in seconds of performing the archival sync
# TYPE chain_sync_archival_sync_total_duration_seconds gauge
chain_sync_archival_sync_total_duration_seconds 14
# HELP chain_sync_emergency_restart_duration_seconds time in seconds of handling the emergency restart
# TYPE chain_sync_emergency_restart_duration_seconds gauge
chain_sync_emergency_restart_duration_seconds 0
...

```
The image shows which parts of the chain sync process are being measured.
image

Closes https://github.com/casper-network/casper-node/issues/2676


afck (approved):

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -35,25 +36,49 @@ use crate::{
     utils::work_queue::WorkQueue,
 };
 
+struct ScopeTimer<'a> {
```
marc-casperlabs, line 44: It would be nice to have docstrings on this, explaining what it does.

marc-casperlabs (approved):
Nicely done factoring things cleanly into the `ScopeTimer` 👏 . I am approving this, but please have a look at the comments still.

I think I need to point again that I think this should be a temporary measure. The `ScoperTimer` is essentially re-implementing what is actually an `instrument()`ed tracing `Span`. So as long as we're aware that this is not supposed to last as metrics and be converted into proper event based tracing (as soon as SRE tooling catches up), I am fine with this :).

rafal-ch:
> So as long as we're aware that this is not supposed to last as metrics and be converted into proper event based tracing (as soon as SRE tooling catches up), I am fine with this :).

Thanks, appreciate that. I have, however, a slightly different opinion, i.e. we shouldn't add to the codebase something we already know will be removed. Especially that right now we don't have an urgent need to debug the fast sync process.

Anyway, I'd like to see what kind of tooling SRE is working on to cover time tracking.

cc: @Fraser999

xcthulhu:
I wholeheartedly agree on the decision to put this data on the `/metrics` endpoint.

A reasonable aspiration is to have the participating reactor switch back and forth between consensus and the `chain_synchronizer` process.  This is a way we could support thousands of validator nodes.

In this case we would have ongoing aggregation of these metrics for the entire time the node is online.  Having them available in the standard place is the right thing to do IMO.

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Comment added here: f7a7f6206837ed1598134cf5e6318c7bb6585b46

rafal-ch:
As discussed on the standup, I'm borsing this as-is since it may require some time to get the metrics via logs and because it's still to be agreed whether we'll abandon these metrics.

PR #2646, opened 2022-02-03
GH-2604: Fix AssemblyScript ordering bug

### LINKED ISSUE #2604: Fix ordering in assemblyscript types
It should say bytes[i] instead of bytes[0]
https://github.com/casper-network/casper-node/blob/14175aebff6219fe09fb2b77ac6c141a2f81b757/smart_contracts/contract_as/assembly/key.ts#L47-L55

mpapierski:
Closes #2604

Fixes a bug where account hashes were not properly ordered.


smart_contracts/contract_as/assembly/key.ts
```diff
@@ -47,19 +47,25 @@ export class AccountHash {
     @operator(">")
     graterThan(other: AccountHash): bool {
```
goral09: BTW, this should be _greater_ :)

goral09 (approved):

smart_contracts/contract_as/assembly/key.ts
mpapierski: Fixed

PR #2649, opened 2022-02-05
Fix from_bytes for WriteBid and WriteWithdraw variants of Transform

SamuelSchlesinger:
## Fix FromBytes for WriteBid and WriteWithdraw variants of Transform

Currently, if you attempt to decode previously encoded WriteBid and WriteWithdraw variants of Transform, it would fail with a `bytesrepr::Error::Formatting` error. This commit properly decodes them, and introduces a convention for doing so in the future.


types/src/execution_result.rs
```diff
@@ -378,15 +435,31 @@ pub enum OpKind {
     NoOp,
 }
 
-impl ToBytes for OpKind {
-    fn to_bytes(&self) -> Result<Vec<u8>, bytesrepr::Error> {
+// TODO(sam)
+// See if reviewer thinks these types should be combined --
+// might more content be added to OpKind later on? I'd hope
+// not.
```
goral09: We will most likely be getting rid of `Op`-like types and operations so it's fine to not put more work into combining them.

types/src/execution_result.rs
```diff
@@ -7,6 +7,8 @@
 // TODO - remove once schemars stops causing warning.
 #![allow(clippy::field_reassign_with_default)]
 
+use std::convert::TryFrom;
```
Fraser999:
```suggestion
use core::convert::TryFrom;
```

Fraser999 (approved):
Approved assuming the suggested change is applied (this should fix the CI failure I believe).

Nice work :+1:

types/src/execution_result.rs
Fraser999: As Mateusz hints, this todo could safely be removed.

types/src/execution_result.rs
```diff
@@ -482,159 +554,182 @@ pub enum Transform {
...
+            Transform::AddUInt64(value) => value.serialized_length(),
+            Transform::AddUInt128(value) => value.serialized_length(),
+            Transform::AddUInt256(value) => value.serialized_length(),
+            Transform::AddUInt512(value) => value.serialized_length(),
+            Transform::AddKeys(value) => value.serialized_length(),
+            Transform::Failure(value) => value.serialized_length(),
             Transform::Identity
             | Transform::WriteContractWasm
             | Transform::WriteContract
             | Transform::WriteContractPackage => U8_SERIALIZED_LENGTH,
```
Fraser999:
```suggestion
            | Transform::WriteContractPackage => 0,
```

Fraser999 (changes requested):
I just noticed a bug in the PR.  Blocking pending the fix.

SamuelSchlesinger:
> I just noticed a bug in the PR. Blocking pending the fix.

Merged your suggested change -- good catch. The CI stuff seems to be a repeated problem where my local toolchain is different from the one in CI, not sure what I should do to systematically fix it but I've been reading the CI scripts and invoking bits of them locally as my intermediate fix.

types/src/execution_result.rs
SamuelSchlesinger: Thank you!

Fraser999 (approved):

PR #2652, opened 2022-02-07
Console based queue dumps

### LINKED ISSUE #2546: Replace signal-based queue dumps with console based one
As soon as debug console support is merged, the USR1/USR2 way of dumping the event queue should be unified by replacing it with an additional command on the debug console.

marc-casperlabs:
Replaces USR1 dumps with console based ones. Closes #2546


node/src/components/console/tasks.rs
```diff
@@ -1,24 +1,32 @@
 use std::{
     borrow::Cow,
     fmt::{self, Debug, Display, Formatter},
-    fs, io,
+    fs,
+    io::{self, Seek, SeekFrom},
     path::PathBuf,
 };
 
 use crate::{
```
Fraser999: Can we sort out the import ordering pls?

node/src/effect/announcements.rs
```diff
@@ -43,6 +47,26 @@ pub(crate) enum ControlAnnouncement {
...
     },
+    // An external event queue dump has been requested.
+    QueueDump {
+        /// One-shot serializer for queue dumps.
+        #[serde(skip)]
+        serializer: RefCell<Option<TempFileSerializer>>,
+    },
+}
+
+impl std::fmt::Debug for ControlAnnouncement {
```
Fraser999:
```suggestion
impl Debug for ControlAnnouncement {
```

Fraser999 (commented):
This is looking good, but could you look to add a few tests please?  Seems like there's scope to add some unit tests as well as a couple of integration tests.

Also, renaming from "console" would be helpful.

node/src/components/console.rs
```diff
@@ -17,17 +21,18 @@ use tracing::debug;
...
     types::NodeRng,
     utils::umask,
     WithDir,
 };
 use util::ShowUnixAddr;
 
-mod command;
-mod tasks;
-mod util;
+pub use tasks::TempFileSerializer;
```
rafal-ch: That should possibly go between `use tracing::debug;` and `use super::Component;`

node/src/components/console/tasks.rs
```diff
@@ -176,6 +225,73 @@ impl Session {
                             }
                         }
                     }
+                    Action::DumpQueues => {
+                        match tempfile::tempfile() {
+                            Ok(mut tmp) => {
+                                // We create a second handle to the serializer temporary file. It
+                                // will be discard once serialization is finished.
```
rafal-ch:
```suggestion
                                // will be discarded once serialization is finished.
```

node/src/components/console/tasks.rs
```diff
@@ -86,6 +94,29 @@ impl Display for Session {
     }
 }
 
+pub enum TempFileSerializer {
```
rafal-ch: That looks like _maybe_ it could go into the `util.rs` of the `console` module?

rafal-ch (approved):
LGTM with just few minor nits :+1: 

I agree with @Fraser999 that at least some basic tests could come in handy.

node/src/components/console/tasks.rs
marc-casperlabs: I'm assuming you mean the ordering of blocks, which is supposed to be `std`, third party crates, crate local ones?

node/src/components/console/tasks.rs
marc-casperlabs: It's a little it too specific in my mind. `util` to me is things that, if they were more numerous, could be in a separate crate.

node/src/components/console/tasks.rs
Fraser999: Four sections: std, third party, other casper, local

node/src/utils/round_robin.rs
```diff
@@ -115,70 +113,31 @@ struct Slot<K> {
...
 
 impl<I, K> WeightedRoundRobin<I, K>
 where
     I: Debug,
     K: Copy + Clone + Eq + Hash + IntoEnumIterator + Debug,
 {
-    /// Dump the contents of the queues (`Debug` representation) to a given file.
-    pub async fn debug_dump(&self, file: &mut File) -> Result<(), io::Error> {
+    /// Dump the queue contents to the given dumper function.
+    pub async fn dump<F: FnOnce(&QueueDump<K, I>)>(&self, dumper: F) {
```
goral09: So that is a shorthand for a function that returns `Unit`? Didn't know that.

goral09 (dismissed):
Seems a bit complicated but I guess that's inevitable. Nicely done, TIL :)

node/src/utils/round_robin.rs
marc-casperlabs: Yes, the `-> ()` is optional. I'd wager you're writing the shorthand every time you define a function without a return type already :)

marc-casperlabs:
> Seems a bit complicated but I guess that's inevitable.

You will be pleased to know that I found a way to get rid of the RefCell :)

goral09:
> > Seems a bit complicated but I guess that's inevitable.
> 
> You will be pleased to know that I found a way to get rid of the RefCell :)

And it seems like it did not require changing any other code besides simply removing `RefCell`?

marc-casperlabs:
> And it seems like it did not require changing any other code besides simply removing RefCell?

I had to add a way to convert an event into a `ControlAnnouncement` owned, instead of just getting a by-ref conversion.

marc-casperlabs:
Tests have been added!

marc-casperlabs:
> Where are you verifying that the event to dump queues has been handled properly?

Fair enough, let me fix that.

node/src/effect/announcements.rs
```diff
@@ -44,6 +48,49 @@ pub(crate) enum ControlAnnouncement {
         /// Error message.
         msg: String,
     },
+    // An external event queue dump has been requested.
+    QueueDumpRequest {
```
goral09, line 52: Why is the other request for diagonostic console not part of this `ControlAnnouncement` enum?

node/src/testing/network.rs
```diff
@@ -28,9 +30,20 @@ use crate::{
...
-    fn node_id(&self) -> NodeId;
+    ///
+    /// The default implementation generates a pseudo-id base on its memory address.
+    fn node_id(&self) -> NodeId {
+        #[allow(trivial_casts)]
+        let addr = self as *const _ as usize;
+        let mut raw: [u8; KeyFingerprint::LENGTH] = [0; KeyFingerprint::LENGTH];
+        raw[0..(mem::size_of::<usize>())].copy_from_slice(&addr.to_be_bytes());
+        NodeId::from(KeyFingerprint::from(raw))
+    }
```
goral09, line 46: Just for my own understanding - how is this change related to the task? Did you stumble across this bit when testing the queue dump functionality?

node/src/components/diagnostics_port.rs
```diff
@@ -0,0 +1,247 @@
...
+        debug!(socket_path=%socket_path.display(), "found stale socket file, trying to remove");
+        match fs::remove_file(&socket_path) {
+            Ok(_) => {
+                debug!("stale socket file removed");
+            }
+            Err(err) => {
+                // This happens if a background tasks races us for the removal, as it usually
+                // means the file is already gone. We can ignore this, but make note of it in
+                // the log.
+                debug!(%err, "could not remove stale socket file, assuming race with background task");
```
goral09: Should this be `warn` at least?

node/src/components/diagnostics_port.rs
```diff
@@ -0,0 +1,247 @@
...
+            .accept()
+            .await
+            .expect("could not accept connection");
+
+        let mut buffer = Vec::new();
+        stream
+            .read_to_end(&mut buffer)
+            .await
+            .expect("failed to read to end");
+        assert_eq!(b"hello", buffer.as_slice());
```
goral09: Nitpick but `hello` could be extracted to a variable and shared between the writing task and here.

node/src/components/diagnostics_port/tasks.rs
```diff
@@ -0,0 +1,630 @@
...
+    fs::{self, File},
+    io,
+    path::PathBuf,
+};
+
+use bincode::{
+    config::{AllowTrailing, FixintEncoding, WithOtherIntEncoding, WithOtherTrailing},
+    DefaultOptions, Options,
+};
+use erased_serde::Serializer as ErasedSerializer;
```
goral09, line 13: Smart!

node/src/components/diagnostics_port/tasks.rs
```diff
@@ -0,0 +1,630 @@
...
+enum ObtainDumpError {
+    /// Error trying to create a temporary directory.
+    #[error("could not create temporary directory")]
+    CreateTempDir(#[source] io::Error),
+    /// Error trying to create a file in the temporary directory.
+    #[error("could not create file in temporary directory")]
+    CreateTempFile(#[source] io::Error),
+    /// Error trying to reopen the file in the temporary directory after writing.
+    #[error("could not reopen file in temporary directory")]
+    ReopenTempFile(#[source] io::Error),
```
goral09, line 138: Where is this `#[source]` coming from? What crate? Will it be included in the error output?

node/src/components/diagnostics_port/tasks.rs
```diff
@@ -0,0 +1,630 @@
...
+    ///
+    /// The resulting serializer will write to the given file.
+    fn create_queue_dump_format(&self, file: File) -> QueueDumpFormat {
+        match self.output {
+            OutputFormat::Interactive => QueueDumpFormat::debug(file),
+            OutputFormat::Json => {
+                QueueDumpFormat::serde(FileSerializer::Json(serde_json::Serializer::new(file)))
+            }
+            OutputFormat::Bincode => {
+                QueueDumpFormat::serde(FileSerializer::Bincode(bincode::Serializer::new(
```
goral09, line 178: I wonder, whether we would ever want to use `bincode` - it's not very human readable.

node/src/components/diagnostics_port/tasks.rs
```diff
@@ -0,0 +1,630 @@
...
+    async fn process_line<REv>(
+        &mut self,
+        effect_builder: EffectBuilder<REv>,
+        writer: &mut OwnedWriteHalf,
+        line: &str,
+    ) -> io::Result<bool>
+    where
+        REv: From<DumpConsensusStateRequest> + From<ControlAnnouncement> + Send,
+    {
+        debug!(%line, "line received");
```
goral09, line 200: I know I already reviewed this code. Pity that I have to do it again. Instead, you could have made the changes first and then maybe do a rename as a separate PR.

goral09 (approved):

node/src/components/diagnostics_port.rs
marc-casperlabs: It could be external to the node as well. I'll upgrade the warning (and fix the typo in the comment as well :))

node/src/testing/network.rs
marc-casperlabs: The test reactor for queue dumping has no networking component, but to use the `Network` testing structure it expects a `NetworkedReactor` implementation.

This is a bit of a misnomer, it does not need an actual network, but rather a unique ID per node. So instead of cramming a networking implementation into my tiny test-reactor, I instead decided to provide a unique ID based on something other than a random network ID, in this case the memory address of the instantiated node.

As it turns out, this makes a reasonable default implementation.

node/src/components/diagnostics_port/tasks.rs
marc-casperlabs: I learn from Fraser!

node/src/components/diagnostics_port/tasks.rs
marc-casperlabs: Its from `thiserror::Error`. It will generate a [`source()`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) implementation (the default is to return `None`), which is just good practice.

There are basically two "styles" of chaining errors - one is to use the properly typed methods like `source()` (see link above), which will show up cleanly in an error trace, the other is just stuffing it into a string, e.g.

```rust
 #[error("could not reopen file in temporary directory: {0}")]
    ReopenTempFile(io::Error),
```

I prefer the former, the only drawback is that the default `Display` implementation generated by `thiserror` does not automatically add the `: {}`, as it may be used in an explicit backtrace-style printing (see the docs on `source()` above, contents of the main function).

I usually try to get the best of both worlds by using our helper function `display_error`.

node/src/components/diagnostics_port/tasks.rs
marc-casperlabs: For large dumps that we have special clients for.

We could drop it though, that's a good point; after all, we can on-the-fly gzip larger ones. :thinking:

node/src/components/diagnostics_port/tasks.rs
marc-casperlabs: I'm sorry :(

node/src/components/diagnostics_port/tasks.rs
goral09: I like this one more as well. Thanks.

node/src/components/diagnostics_port/tasks.rs
goral09: >For large dumps that we have special clients for.

We have clients for that? Who?

node/src/effect/announcements.rs
marc-casperlabs: Because `ControlAnnouncement` (which should actually be named `ControlEvent` now, tbh) groups all events that are handled outside of the reactor by the runtime directly.

The other request is for data from consensus and requires no direct interventions from the `Runner`, as opposed to this one.

goral09: Ah, of course. Makes a lot of sense. Thanks.

node/src/components/diagnostics_port/tasks.rs
marc-casperlabs: Err, might write, sorry :)

marc-casperlabs:
bors retry

PR #2662, opened 2022-02-10
NCTL: show node status and chain-height before exiting

TomVasile:
CHANGES:
- output node status and chain-height when sync check times out.
- sync delegation amount to new minimum


Fraser999 (approved):

TomVasile:
bors try

PR #2663, opened 2022-02-14
npm audit in CI

bradjohnl:
This PR introduces the npm audit step for production-related libraries in CI.
If any vulnerabilities with severity >= moderate are left, the step will fail.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -34,7 +34,7 @@ steps:
       - make setup-rs
       - make doc
 
-  - name: cargo-audit
+  - name: audit
     image: casperlabs/node-build-u1804
     commands:
       - make setup-audit
--- a/Makefile
+++ b/Makefile
@@ -118,9 +118,17 @@ lint: lint-contracts-rs
 	$(CARGO) clippy --all-targets --all-features -- -D warnings -A renamed_and_removed_lints
 	cd smart_contracts/contract && $(CARGO) clippy --all-targets -- -D warnings -A renamed_and_removed_lints
 
-.PHONY: audit
-audit:
+.PHONY: audit-rs
+audit-rs:
 	$(CARGO) audit --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2020-0159
+
+.PHONY: audit-as
+audit-as:
+	@# Runs a vulnerability scan that fails if there are prod vulnerabilities with a moderate level or above.
+	cd smart_contracts/contract_as && $(NPM) audit --production --audit-level=moderate
+
+.PHONY: audit
+audit: audit-rs audit-as
 
 .PHONY: doc
 doc:
--- a/.gitignore
+++ b/.gitignore
@@ -139,6 +139,7 @@ cmake-build-debug/
 
 # vscode
 .vscode/
+.dccache
 
 # NCTL transient assets
 /utils/nctl/assets


bradjohnl:
@Fraser999 @mpapierski @hoffmannjan @luis-casper Hi, can one of you please check on this pull request pending review before the sprint ends?

Makefile
```diff
@@ -118,10 +118,18 @@ lint: lint-contracts-rs
...
-.PHONY: audit
-audit:
+.PHONY: audit-rs
+audit-rs:
 	$(CARGO) audit --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2020-0159
 
+.PHONY: audit-as
+audit-as:
+	@# Attempts to fix vulnerabilities and run a scan after which fails if there are still prod vulnerabilities with a moderate level or above.
+	cd smart_contracts/contract_as && $(NPM) audit fix --production && $(NPM) audit --production --audit-level=moderate
```
TomVasile: Wouldn't the use of `fix` here make changes to the changeset in CI?

bradjohnl: I have checked and you are right, I have seen that the package step clones the repo one more time, so the changes would be lost.
Because of this I need to find a way to persist the change. 
I could as a quick solution apply the fix again during the packaging phase until a more elegan solution is found.
I will look into it.

TomVasile: I don't think the change would persist regardless because its not committed. Probably easiest just to fail the pipeline if the audit catches anything and let the dev make the appropriate changes.

.drone.yml
```diff
@@ -106,6 +106,7 @@ steps:
 - name: cargo-test
   image: casperlabs/node-build-u1804
   commands:
+  - make audit-as
```
TomVasile: shouldn't be necessary as it is run in pre-checks

.drone.yml
```diff
@@ -135,6 +136,7 @@ steps:
 - name: build-client-contracts
   image: casperlabs/node-build-u1804
   commands:
+    - make audit-as
```
TomVasile: Likewise; shouldn't be necessary as it is run in pre-checks

Makefile
TomVasile: something like this:

```suggestion
	cd smart_contracts/contract_as && $(NPM) audit --production --audit-level=moderate
```

TomVasile (changes requested):

Makefile
bradjohnl: @TomVasile I think we can start with this as you proposed.

I was looking for ways to cache the changes and have them included in the package/release when doing npm audit fix as they are non-breaking.

Repeating the audit step adds too much redundancy even if the cost is relatively low compared to the median CI job run, and it does not carry the expected results (and it's not really following best practices).

In the current drone setup, to do it properly in that way, you should start caching the build results between steps, but it would be an even better way to achieve it if you have something automatically commit and push the easy automated npm audit fixes (Something like bors or GH actions can do it).

bradjohnl: @TomVasile I think we can start with this as you proposed.

I was looking for ways to cache the changes and have them included in the package/release when doing npm audit fix as they are non-breaking.

Repeating the audit step adds too much redundancy even if the cost is relatively low compared to the median CI job run, and it does not carry the expected results (and it's not really following best practices).

In the current drone setup, to do it properly in that way, you should start caching the build results between steps, but it would be an even better way to achieve it if you have something automatically commit and push the easy automated npm audit fixes.

.drone.yml
bradjohnl: I will change back the pipeline as per the discussion above.

.drone.yml
bradjohnl: I will change back the pipeline as per the discussion above.

bradjohnl:
Hi @TomVasile. I have implemented the suggestions and removed the repeated steps.

Makefile
```diff
@@ -118,10 +118,18 @@ lint: lint-contracts-rs
...
 
-.PHONY: audit
-audit:
+.PHONY: audit-rs
+audit-rs:
 	$(CARGO) audit --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2020-0159
 
+.PHONY: audit-as
+audit-as:
+	@# Attempts to fix vulnerabilities and run a scan after which fails if there are still prod vulnerabilities with a moderate level or above.
```
TomVasile: Just a nit pick, update the comment as this does not attempt to fix any more.

TomVasile (approved):
just the one nit pick from me

PR #2664, opened 2022-02-15
Simplify implementation of `run_chain_sync_task`

rafal-ch:
This PR introduces a refactor to some of the functions in `opeartions.rs`, the main goal being the simplification of ~250 lines long `run_chain_sync_task` function. This change will make it easier to add proper metrics to the fast chain sync process (https://github.com/casper-network/casper-node/pull/2640).

The main idea behind the cleanup is to:
* introduce `ChainSyncContext` 
* split the `run_chain_sync_task()` into parts
* wire-up these parts using the context that holds common data needed for processing


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -532,102 +493,209 @@ async fn sync_trie_store(
...
-                    &most_recent_block_header,
-                    chainspec.protocol_config.verifiable_chunked_hash_activation,
-                ) {
-                    trusted_key_block_info = key_block_info;
-                }
+
+        if let Some(more_recent_block_header_with_metadata) = maybe_fetched_block {
+            most_recent_block_header = more_recent_block_header_with_metadata.block_header;
+
+            // TODO: If the new block is a switch block, update the validator weights, etc...
```
afck: We should return that, and use it in `fetch_blocks_for_deploy_replay_protection`, otherwise it has an outdated key block info…

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -532,102 +493,209 @@ async fn sync_trie_store(
...
 
-    // Fetch and store all blocks that can contain not-yet-expired deploys. These are needed for
-    // replay detection.
+/// Fetch and store all blocks that can contain not-yet-expired deploys. These are needed for
+/// replay detection.
+async fn fetch_blocks_for_deploy_replay_protection(
+    effect_builder: EffectBuilder<JoinerEvent>,
+    chainspec: &Chainspec,
+    most_recent_block_header: &BlockHeader,
+    trusted_key_block_info: &KeyBlockInfo,
```
afck: …so we could probably safely use `most_recent_key_block_info` here instead.
(I guess it's not a blocker: We'll just fetch more blocks than necessary this way.)

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -684,23 +738,24 @@ async fn sync_deploys_and_transfers_and_state(
...
     info!("start - fetch the trusted block - archival sync");
     let trusted_block = *fetch_and_store_block_by_hash(
-        effect_builder,
-        trusted_block_header.hash(chainspec.protocol_config.verifiable_chunked_hash_activation),
+        *ctx.effect_builder,
+        ctx.trusted_block_header.hash(
+            ctx.chainspec
+                .protocol_config
+                .verifiable_chunked_hash_activation,
+        ),
```
afck: Actually, don't we have the trusted _hash_ in the node config? Then we wouldn't need to recompute it here.
Otherwise, maybe we should just add it to `ChainSyncContext`?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -725,12 +781,12 @@ async fn archival_sync(
...
     let mut most_recent_block = trusted_block;
-    while most_recent_block.header().protocol_version() < chainspec.protocol_config.version {
+    while most_recent_block.header().protocol_version() < ctx.chainspec.protocol_config.version {
         let maybe_fetched_block_with_metadata = fetch_and_store_next::<BlockWithMetadata>(
-            effect_builder,
+            *ctx.effect_builder,
             most_recent_block.header(),
             &trusted_key_block_info,
-            chainspec,
+            ctx.chainspec,
```
afck: I wonder whether some of these functions should now be `ChainSyncContext` methods instead?
(Doesn't have to be in this PR; just a thought!)

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -740,156 +796,206 @@ async fn archival_sync(
...
         };
-        sync_deploys_and_transfers_and_state(effect_builder, &most_recent_block, &node_config)
-            .await?;
+        sync_deploys_and_transfers_and_state(
+            *ctx.effect_builder,
+            &most_recent_block,
+            ctx.node_config,
+        )
+        .await?;
         if let Some(key_block_info) = KeyBlockInfo::maybe_from_block_header(
```
afck, line 768: (Then this could be `cts.maybe_key_block_info(most_recent_block.header())`.)

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -740,156 +796,206 @@ async fn archival_sync(
...
+    } else {
+        fast_sync_to_most_recent(&chain_sync_context).await?
+    };
+
+    let most_recent_block_header = execute_blocks(
+        &chain_sync_context,
+        &most_recent_block_header,
+        &trusted_key_block_info,
+    )
+    .await?;
```
afck, line 816: (These functions also all look like `ChainSyncContext` methods now.)

afck (approved):
Looks good to me! :+1:

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Resolved in c33e3e886d0af356d770003062fed1a00114292d

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Resolved in c33e3e886d0af356d770003062fed1a00114292d

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Yes, we have it, but it's an `Option`.
There is already a code that will read it from the highest block in case it is missing in config and will pass it to `run_chain_sync_task()`: https://github.com/casper-network/casper-node/blob/feat-fast-sync/node/src/components/chain_synchronizer.rs#L74-L86

So yes, I think we should use it in `archival_sync()`. I'll add a commit if all tests pass.

afck: Ah, you're right, we need to accept it from either config or storage. Happy to leave it as it is.

rafal-ch: Hmm, so we can actually be fetching the same block header twice:
1. in the constructor for `ChainSynchronizer` to read the trusted hash
2. in `run_chain_sync_task()` to read the trusted block header by trusted hash

We could possibly remove the `trusted_hash` parameter from the `start_syncing()` and `run_chain_sync_task()` functions and let the latter read it by itself (from config or from latest block).

rafal-ch: The `trusted_hash` is now part of `ChainSyncContext` and is reused: ceea7dba7af6914a226021035a89916d7f299196

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Yes, that's a good idea. And yes, I'd rather keep it out of the scope of this PR for now.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -740,156 +793,204 @@ async fn archival_sync(
...
     let trusted_block_header = fetch_and_store_block_header(effect_builder, trusted_hash).await?;
     debug!("finish - fetch trusted header");
 
-    if trusted_block_header.protocol_version() > chainspec.protocol_config.version {
+    let chain_sync_context = ChainSyncContext::new(
+        &effect_builder,
+        &chainspec,
+        &node_config,
+        &trusted_block_header,
+        &trusted_hash,
```
goral09: Isn't `trusted_hash == trusted_block_header.hash()`?

afck: I hope so! It should not just be the entry from the config, but possibly also our latest block from storage.

Sorry, this was based on my earlier comment, and now I'm not sure about it. If you don't like the duplicated data, we could also make `trusted_hash` a _method_ on `ChainSyncContext`?

goral09: It's not about duplication but rather the chance of getting it wrong. If they're equal I think it's just safer to use the hash of the `trusted_block_header`.

afck: You're right, I agree. Sorry for the confusion! Let's make it a `trusted_hash` method then.

node/src/components/chain_synchronizer/operations.rs
rafal-ch: See here: https://github.com/casper-network/casper-node/pull/2664#discussion_r807038024

node/src/components/chain_synchronizer/operations.rs
rafal-ch: See here: https://github.com/casper-network/casper-node/pull/2664#discussion_r807038024

node/src/components/chain_synchronizer/operations.rs
rafal-ch: The `trusted_hash()` method has been added here: https://github.com/casper-network/casper-node/pull/2664/commits/a2de4d36785bd2df35a3dda948986a22029facaf

rafal-ch:
Adding the `do not merge` label, because after the merge of https://github.com/casper-network/casper-node/pull/2650 and fixing the conflicts, the `chain_synchronizer` now uses **all three** "value providers" at the same time:
* chainspec
* chain sync context
* config

I will re-request the review after I clean things up.

rafal-ch:
OK, clean-up finished, re-requesting the review.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -138,18 +165,21 @@ async fn fetch_and_store_block_header(
     }
 }
 
+/// Fetches and stores a deploy.
```
afck, line 168:
```suggestion
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -552,141 +507,203 @@ async fn sync_trie_store(
...
-                    &most_recent_block_header,
-                    config.verifiable_chunked_hash_activation(),
-                ) {
-                    trusted_key_block_info = key_block_info;
-                }
+
+        if let Some(more_recent_block_header_with_metadata) = maybe_fetched_block {
+            most_recent_block_header = more_recent_block_header_with_metadata.block_header;
+
+            // TODO: If the new block is a switch block, update the validator weights, etc...
```
afck: That's done, isn't it?

```suggestion
            // If the new block is a switch block, update the validator weights, etc...
```

afck (approved):

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Removed https://github.com/casper-network/casper-node/pull/2664/commits/273201177f38a8ee28c5aa2eabe8c7efb03c56e7

node/src/components/chain_synchronizer/operations.rs
rafal-ch: I see now - if the header contains the "next era validator weights" it means that it's a switch block.
So yes, it's done, indeed.

https://github.com/casper-network/casper-node/pull/2664/commits/1499d9b250d6b4eb5c91f91dd6f30bb4aa6e431e

rafal-ch:
as agreed, merging now with one approval
cc: @Fraser999

PR #2665, opened 2022-02-15
Bradjohnl gh 1803 fix npm vulns

bradjohnl:
Testing.
Branch rebased on top of the 1802 branch so for now it contains the commits in there as well.
When that is merged, this one will be in line with the dev branch


bradjohnl:
Trying with another approach rather than tweaking the package-lock.json....

PR #2666, opened 2022-02-15
Remove Node ID Generic Parameter from Consensus

SamuelSchlesinger:
I tried to remove it merely from the surrounding territory of the consensus protocol, but that did not work, so I think it makes sense to remove it from everywhere. Once we extract highway and our new consensus protocol as separate crates, we can parameterize them there so users aren't forced to depend on our type.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,3 +153,6 @@ cmake-build-debug/
 
 # OS X
 .DS_Store
+
+# sw* files in vim
+.*.sw*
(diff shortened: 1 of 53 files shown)


xcthulhu (approved):

xcthulhu:
Chatting with @afck this is GTG (@marc-casperlabs and I already approved)

It's a bit unfortunate that there was no clean way to keep the generic parameter...

xcthulhu:
bors merge

PR #2668, opened 2022-02-16
Merge `dev` to `feat-fast-sync` 16.02.2022

rafal-ch:
This PR integrates the recent changes from the `dev` (as of 16.02.2022 9:30am CEST) branch into the `feat-fast-sync` branch.


utils/nctl/sh/scenarios-upgrades/upgrade_scenario_04.sh
```diff
@@ -0,0 +1,370 @@
+#!/usr/bin/env bash
+# ----------------------------------------------------------------
+# Synopsis.
+# ----------------------------------------------------------------
+
+# Step 01: Start network from pre-built stage.
+# Step 02: Await era-id >= 1.
```
afck, line 7: In those cases where it's actually important to wait for the duration of a whole "normal" era, we'll need to wait for era 2 with fast sync, because we now have an era 0 consisting of just a single auto-generated switch block.
(But maybe it doesn't matter and "waiting for era 1" is usually just to wait until the network has started making progress?)

afck (approved):

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_04.sh
rafal-ch: @TomVasile can you please have a look at the comment from @afck ?

TomVasile: Correct the waiting for era 1 is solely just waiting for the network to make some progress. We do this in most of the nctl tests.

TomVasile (approved):

PR #2670, opened 2022-02-17
Low priority networking queue

marc-casperlabs:
Adds a special low-priority incoming network message queue. Any incoming request for data will be put into this queue, preventing these kinds of message from stopping things like consensus message to reach the respective component or interfere with internal node operation (an incoming consensus message will essentially bypass almost all `GetRequests` this way).

This should keep consensus nodes live even when under pressure by joining nodes.

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -33,6 +33,7 @@ All notable changes to this project will be documented in this file.  The format
 * Chain automatically creates a switch block immediately after genesis or an upgrade.
 * Connection handshake timeouts can now be configured via the `handshake_timeout` variable (they were hardcoded at 20 seconds before).
 * `Key::SystemContractRegistry` is now readable and can be queried via the RPC.
+* Requests for data from a peer are now de-prioritized over networking message necessary for consensus and chain advancement.
 
 ### Deprecated
 * Deprecate the `starting_state_root_hash` field from the REST and JSON-RPC status endpoints.
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -33,7 +33,7 @@ All notable changes to this project will be documented in this file.  The format
 * Chain automatically creates a switch block immediately after genesis or an upgrade.
 * Connection handshake timeouts can now be configured via the `handshake_timeout` variable (they were hardcoded at 20 seconds before).
 * `Key::SystemContractRegistry` is now readable and can be queried via the RPC.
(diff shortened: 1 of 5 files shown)


node/src/protocol.rs
```diff
@@ -79,6 +79,17 @@ impl Payload for Message {
         }
     }
 
+    fn is_low_priority(&self) -> bool {
+        match self {
+            Message::Consensus(_) => false,
+            Message::DeployGossiper(_) => false,
+            Message::AddressGossiper(_) => false,
+            Message::GetRequest { .. } => true,
```
Fraser999: Don't we want `GetRequest` to be more fine-grained?  We don't want to de-prioritise requests made by joined validators.

node/src/protocol.rs
```diff
@@ -79,6 +79,17 @@ impl Payload for Message {
...
+            Message::GetResponse { .. } => false,
+            Message::FinalitySignature(_) => false,
```
rafal-ch: As I mentioned earlier, I think these values should be configurable, so we can change them between test runs. This can be addressed in a separate PR, if at all.

marc-casperlabs: Let's prove it does something useful first :)

node/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ All notable changes to this project will be documented in this file.  The format
 * Chain automatically creates a switch block immediately after genesis or an upgrade.
 * Connection handshake timeouts can now be configured via the `handshake_timeout` variable (they were hardcoded at 20 seconds before).
 * `Key::SystemContractRegistry` is now readable and can be queried via the RPC.
+* Requests for data from a peer are now de-prioritized over networking message necessary for consensus and chain advancement.
```
rafal-ch:
```suggestion
* Requests for data from a peer are now de-prioritized over networking messages necessary for consensus and chain advancement.
```

rafal-ch (approved):
LGTM :+1:

node/src/protocol.rs
marc-casperlabs: I think we don't want to overcomplicate it for now. An earlier approach intended to take validator state into account. We won't be worse off than we are right now.

Configurability (mentioned by Rafal) and more fine-grained control could be a second step here. But if this doesn't solve currently scalability issues on its own, we may end up reverting this anyway.

marc-casperlabs: To be a little more precise, I changed it so we only deprioritize trie nodes. This is a very safe bet, as validators will never create these kinds of requests.

Issue #2255, opened 2021-10-22
Discover all production code panics


piotr-dziubecki:
Assert, unwrap, panic, unreachable, i.e. not tests, root/utils, benches. The deliverable for this is a set of links against feat-fast-sync catagorized where it's in the code, e.g. in network, EE, consensus, etc...

HansWestman74:
After discussion with Fraser, Matt and Ed we are closing the initial PR providing just the list. And we consider this work to be done. Follow up stories will be added in future sprints.

@piotr-dziubecki  Per Ed create an epic for this tranche of work. More stories will be created within it. Ask Fraser for details.

PR #2680, opened 2022-02-23
Implement rocksdb as global state store

### LINKED ISSUE #2623: Implement RocksDb as a backing store for Global State

dwerner:
Resolves #2623 

This PR brings in rocksdb as our backing store and implements a migration from lmdb's trie data as an async background task. Once the migration has been confirmed to run on a node it would be possible to delete our old `data.lmdb` file and thereby free the related disk space.

Further work to refactor out existing traits that supported/abstracted the use of lmdb will need to come in another PR.

Incidentally this PR also makes the `storage` component _not_ block the reactor thread, as that showed up as a significant performance bottleneck during testing. It also includes refactorings needed to make the `disk_use` bin tool operational, as this was needed to test the work as it progressed. This tool is not "complete" but can hopefully serve as a basis for some automated tests going forward.

@EdHastingsCasperLabs @Fraser999 @goral09 @mpapierski 

I'm particularly looking for input on this with an eye for minimal scope creep.


goral09:
If I'm reading this correctly, this PR could be separated into multiple smaller ones:
1. Adding ability to record on-disk size (and Trie structure) of DB storage backing the execution engine. For which [you already have a PR](https://github.com/casper-network/casper-node/pull/2549/) (I've also reviewed that one recently.
2. Making access to storage not block the reactor and make it async.
3. Changes to `utils/` making it DB backend generic and not fixed for LMDB.
4. Adding RocksDB as storage backend.

Each of the above can be introduced separately. I would suggest breaking up this PR. Otherwise you're making it much more difficult for the reviewers to verify the changes properly. Which leads to more errors slipping through due to mental fatigue. And we have seen that just recently on a different PR. I would like to avoid doing these same mistakes again.

dwerner:
> If I'm reading this correctly, this PR could be separated into multiple smaller ones:
> 
>     1. Adding ability to record on-disk size (and Trie structure) of DB storage backing the execution engine. For which [you already have a PR](https://github.com/casper-network/casper-node/pull/2549/) (I've also reviewed that one recently.

There was little point in that PR because it is obviated by this one. Things have been updated in this PR, so I have closed it. I can break these changes out, but the rocksdb backend work depended on them.

>     2. Making access to storage not block the reactor and make it async.

~Agreed, this could be broken out, however it was during the investigation of rocksdb in test that it was discovered and resolved. It's a dependency of the rocksdb work, as in, is required by.~

Updated, see my more recent comment on this below.

>     3. Changes to `utils/` making it DB backend generic and not fixed for LMDB.

I'm not sure what you mean here. I did some refactoring of the tools there and actually removed a larger portion of the code in retrieve-state.

>     4. Adding RocksDB as storage backend.
> 
> Each of the above can be introduced separately. I would suggest breaking up this PR.

While it is true that it could be broken up, I don't agree that these are inherently unrelated changes and should definitively be split into separate PRs. Some are refactorings, but they are related temporally and I see them as a dependency chain.

goral09:
The point of breaking out into smaller PR is to make the review process easier which yields better results, tighter review and fewer bugs slipping through.

I understand that RocksDB support depends on all of that. I'm suggesting to extract the orthogonal changes, create separate PRs, review and merge them first, rebase RocksDB addition on top of `dev`  and only then open up the PR.

dwerner:
For reference, here is the on-disk growth of rocksdb (blue) vs lmdb (red). This example uses blocks with 5000 but empirically the growth is linear to the number of transfers, regardless of block size.

image

dwerner:
> ...I'm suggesting to extract the orthogonal changes...

My contention is that the changes are not orthogonal. The storage component functions adequately _in our current world where we only use lmdb_. The timeouts of "event took a very long time to process" of type `StorageRequest`, of which we saw a large number (>3500) meant that the slowdown was only encountered while simultaneously using rocksdb as the datastore for the trie.

RocksDb makes use of background threads and performs compaction on those background threads. This means that the potential of many threads flushing to disk and causing latency is much higher.

node/src/reactor/initializer.rs
```diff
@@ -231,13 +238,24 @@ impl Reactor {
...
-        // integrity checks are even a good idea, as they can block the node for one or more hours
-        // on restarts (online checks are an alternative).
+        {
+            let engine_state = Arc::clone(contract_runtime.engine_state());
+            // We need to be synchronous here, ensures migration (of just the highest block) is
+            // complete before proceeding.
+            match storage.read_highest_block_header()? {
+                Some(highest_block_header) => {
+                    engine_state.migrate_state_root_to_rocksdb_if_needed(
+                        *highest_block_header.state_root_hash(),
```
xcthulhu, line 216: Disk throttling needs to be turned off here....

execution_engine/src/lib.rs
```diff
@@ -18,3 +18,52 @@ pub mod core;
...
+
+    db_opts.set_compression_type(ROCKS_DB_COMPRESSION_TYPE);
+    db_opts.set_compression_options(
+        ROCKS_DB_WAL_BITS,
+        ROCKS_DB_ZSTD_COMPRESSION_LEVEL,
+        ROCKS_DB_ZSTD_STRATEGY,
+        ROCKS_DB_ZSTD_MAX_DICT_BYTES,
+    );
+
+    // seems to lead to a sporatic segfault within rocksdb compaction
```
mpapierski: This comment looks scary. Is there any upstream issue, or some documentation on why this may happen?

execution_engine/src/lib.rs
```diff
@@ -18,3 +18,52 @@ pub mod core;
...
+
+const ROCKS_DB_BLOCK_SIZE_BYTES: usize = 256 * 1024;
+const ROCKS_DB_COMPRESSION_TYPE: rocksdb::DBCompressionType = rocksdb::DBCompressionType::Zstd;
+const ROCKS_DB_COMPACTION_STYLE: rocksdb::DBCompactionStyle = rocksdb::DBCompactionStyle::Level;
+const ROCKS_DB_ZSTD_MAX_DICT_BYTES: i32 = 256 * 1024;
+const ROCKS_DB_MAX_LEVEL_FILE_SIZE_BYTES: u64 = 512 * 1024 * 1024;
+const ROCKS_DB_MAX_OPEN_FILES: i32 = 768;
+const ROCKS_DB_ZSTD_COMPRESSION_LEVEL: i32 = 3; // Default compression level
+const ROCKS_DB_ZSTD_STRATEGY: i32 = 4; // 4: Lazy
+const ROCKS_DB_WAL_BITS: i32 = -14;
```
mpapierski: Why -14?

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,309 @@
...
+        {
+            if entry.metadata().unwrap().is_file() {
+                total += entry.path().size_on_disk().unwrap() as usize;
+            }
+        }
+        total
+    }
+}
+
+// TODO: remove this abstraction entirely when we've moved from lmdb to rocksdb for global state.
```
mpapierski, line 111: I'm not sure about this. There are open PRs upstream to add support for transactions (for example this looks promising https://github.com/rust-rocksdb/rust-rocksdb/pull/565); are we sure we don't want to use transactions in future? Or - if transaction support was present in rust-rocksdb would you use it?

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,309 @@
...
+        rocksdb_opts: rocksdb::Options,
+    ) -> Result<RocksDbStore, rocksdb::Error> {
+        let db = Arc::new(rocksdb::DBWithThreadMode::<MultiThreaded>::open(
+            &rocksdb_opts,
+            path.as_ref(),
+        )?);
+
+        let rocksdb = RocksDb {
+            db,
+            // This is just for testing.
```
mpapierski: If `in_mem` is for testing, then why not use a`InMemoryGlobalState`? What does `in_mem == true` adds to existing InMem database. Also - why mix production code and testing (unrelated to impl) in a single RockDb instance, and switch two different implementations based on a bool.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -198,7 +204,7 @@ impl LmdbWasmTestBuilder {
         let environment = Arc::new(
             LmdbEnvironment::new(
                 &global_state_dir,
-                page_size * DEFAULT_LMDB_PAGES,
+                page_size * 1024 * 1024,
```
mpapierski: Was this change intentional?

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -185,11 +190,12 @@ impl InMemoryWasmTestBuilder {
...
 }
 
-impl LmdbWasmTestBuilder {
-    /// Returns an [`LmdbWasmTestBuilder`] with configuration.
+impl DbWasmTestBuilder {
+    /// Returns an [`DbWasmTestBuilder`] with configuration.
     pub fn new_with_config<T: AsRef<OsStr> + ?Sized>(
         data_dir: &T,
         engine_config: EngineConfig,
+        rocksdb_opts: rocksdb::Options,
```
mpapierski, line 200: What about newtype `DbGlobalStateOptions` and `impl Default for DbGlobalStateOptions` to avoid leaking rocksdb types (like we leak lmdb::DatabaseFlags currently)

execution_engine/src/storage/transaction_source/db.rs
mpapierski: `in_mem` doesn't seem to be used anywhere, so can we get rid of this?

execution_engine/src/storage/error/db.rs
```diff
@@ -14,6 +15,19 @@ pub enum Error {
     #[error(transparent)]
     Lmdb(#[from] lmdb_external::Error),
 
+    /// RocksDb error
+    #[error("{0}")]
+    RocksDb(String),
```
mpapierski:
```suggestion
    #[error(transparent)]
    RocksDb(#[from] rocksdb::Error),
```
(and you may need to remove derive Eq)

As historically storage in EE had leaky abstractions I'd appreciate it if we'd plan to have a clean abstraction starting with errors in `storage::Error`. Have a single variant `Database(Box<dyn std::error::Error>)` that contains either lmdb or rocksdb error at a minimum. Or a few generic error categories represented as variants that will be translated from DB-specific errors similar to https://github.com/casper-network/casper-node/blob/d0b3e00a862631b249a125e4bc2227041b132c99/node/src/components/storage/lmdb_ext.rs#L26-L41

execution_engine/src/storage/global_state/mod.rs
```diff
@@ -4,7 +4,7 @@
 pub mod in_memory;
 
 /// Lmdb implementation of global state.
-pub mod lmdb;
+pub mod db;
```
mpapierski, line 7: Outdated rustdoc

execution_engine/src/lib.rs
```diff
@@ -18,3 +18,52 @@ pub mod core;
...
+
+/// Default constructor for rocksdb options.
+pub fn rocksdb_defaults() -> rocksdb::Options {
```
mpapierski: As I noted elsewhere I'd appreciate it if we'd hide this under a `DbGlobalStateOptions` and `impl Default`

execution_engine/src/lib.rs
dwerner: Found through experimentation, examination of a resulting stack dump, and solved by not using this feature for now. I left the comment here to make sure that we don't turn it on again.

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: Didn't intend on keeping this, thanks - was using valgrind.

execution_engine/src/storage/error/db.rs
dwerner: Ok I've decided to break out a separate DbError enum that explicitly handles lmdb::Error and rocksdb::Error as well as our own errors here. Newtyping over String here was done as a stopgap.

execution_engine/src/storage/transaction_source/db.rs
dwerner: Because I'd like to delete InMemoryGlobalState in a future PR, along with quite a bit of generic code that attempts to abstract over *GlobalState. EngineState need not be so generic, nor WasmTestBuilder. I think this change would allow a great deal of simplification of both production and test code in the future.

execution_engine/src/storage/transaction_source/db.rs
dwerner: If we assume we will move forward with RocksDb, then I would like to first factor out transactions as they bind us to what lmdb thinks of as transactions, and in the future, if we decided to bring transaction support back, we can add it back in a way that does not allow the proliferation of overly-generic traits that still manage to bind us to the implementation. I'm leaning very much YAGNI in this case.

execution_engine/src/lib.rs
dwerner: Well, it's named incorrectly, should be WINDOW_BITS, or W_BITS. But why -14? I found the value as the default ctor sets it here https://github.com/facebook/rocksdb/blob/7d7e88c7d1b676274f6e77daff88e4b696ff7e06/include/rocksdb/advanced_options.h#L177

The rust crate wraps the rocksdb c api and lacks these helpers, so I needed a value.

execution_engine/src/storage/transaction_source/db.rs
xcthulhu: So while we shouldn't paint ourselves into a corner where we can't support transactions, we don't have the abstraction right IMO.

Two out of three implementations (ie, RocksDB or HashMaps) either don't currently or will never support transactions.  So another abstraction that doesn't require creating artificial transactions is desirable.

Maybe something like these: `trait ReadDBHandle { fn get(...) -> Result<BobsYourUncle> {...} }` and `trait ReadWriteDBHandle: ReadDBHandle { fn put(...) -> Result<()> {...} }`?

dwerner: I'm open to discuss but it shouldn't block this PR since the work wouldn't happen here.

execution_engine/src/storage/transaction_source/db.rs
dwerner: Removed now, talking with @xcthulhu , I'll come back to this in a future PR.

xcthulhu (approved):

execution_engine/src/storage/global_state/db.rs
```diff
@@ -1,7 +1,17 @@
...
+    time::{Duration, Instant},
+};
 
 use casper_hashing::Digest;
-use casper_types::{Key, StoredValue};
+use casper_types::{
+    bytesrepr::{self, Bytes, ToBytes},
+    Key, StoredValue,
+};
+use tracing::info;
```
rafal-ch: `tracing` should go before `casper_hashing`

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -256,32 +256,40 @@ where
...
                     missing_descendants.push(trie_key);
                     continue;
                 }
             }
         }
+
+        // Optimization: Don't look for descendants of leaves.
+        if let Some(bytes) = maybe_retrieved_trie_bytes.as_ref() {
+            // if this is a leaf
+            if bytes.as_ref()[0] == 0u8 {
```
rafal-ch: Isn't it a risk here that we can panic with `index out of bounds: the len is 0 but the index is 0`?

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -256,32 +256,40 @@ where
...
+                continue;
+            }
+        }
+
+        let maybe_retrieved_trie: Option<Trie<K, V>> = maybe_retrieved_trie_bytes
+            .map(bytesrepr::deserialize_from_slice)
+            .transpose()?;
+
         match maybe_retrieved_trie {
+            Some(Trie::Leaf { .. }) => unreachable!(),
```
rafal-ch: There is [a ticket in progress](https://github.com/casper-network/casper-node/issues/2255) that focuses on removing `unwrap()`, `expect()`, but also other things that can panic, like `unreachable!()`.

Please see if we can avoid introducing this one.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -256,32 +256,41 @@ where
         if !visited.insert(trie_key) {
             continue;
         }
-        let maybe_retrieved_trie: Option<Trie<K, V>> = store.get(txn, &trie_key)?;
+        let maybe_retrieved_trie_bytes: Option<Bytes> = store.get_raw(txn, &trie_key)?;
```
xcthulhu, line 259:
```suggestion
        let trie_value_bytes: Bytes = match store.get_raw(txn, &trie_key) {
           None => { missing_descendants.push(trie_key); continue }
           Some(trie_value_bytes) => trie_value_bytes
        };
```

This simplifies the code downstream...

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -256,32 +256,41 @@ where
...
             }
         }
+
+        // Optimization: Don't look for descendants of leaves.
+        if let Some(bytes) = maybe_retrieved_trie_bytes.as_ref() {
+            // if this is a leaf
+            if !bytes.is_empty() && bytes[0] == 0u8 {
+                continue;
+            }
+        }
```
xcthulhu:
```suggestion
        match maybe_retrieved_trie_bytes.as_ref().and_then(|bytes| bytes.get(0)) {
            // if this is a leaf don't parse (ie, the first byte is 0), just continue
            Some(0) => continue;
            // Either no entry under trie key or data is empty.  If data is empty then the trie is corrupted.
            // Either way treat as missing.
            None => { missing_descendants.push(trie_key); continue }
             _ => ()
        }
```

If you decide to go with [my suggestion above][1] then don't do `add_then` and just have `get` instead.

[1]: https://github.com/casper-network/casper-node/pull/2680/files#r820254715

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -256,32 +256,41 @@ where
...
+
+        let maybe_retrieved_trie: Option<Trie<K, V>> = maybe_retrieved_trie_bytes
+            .map(bytesrepr::deserialize_from_slice)
+            .transpose()?;
```
xcthulhu:
```suggestion
        match retrieved_trie: Option<Trie<K, V>> = maybe_retrieved_trie_bytes
            .map(bytesrepr::deserialize_from_slice)
            .transpose() {
            Ok(Some(retrieved_trie)) => retrieved_trie,
            Ok(None) => { missing_descendants.push(trie_key); continue }
            // Couldn't parse; treat as missing and continue
            Err(err) => {error!(err) ; missing_descendants.push(trie_key); continue }
        }
```

Again, if you decide to go with [my suggestion above][1] then you only have to treat the `Ok(retrieved_trie)` and `Err(error)` match cases...

[1]: https://github.com/casper-network/casper-node/pull/2680/files#r820254715

dwerner:
@marc-casperlabs @Fraser999 Can I please have a quick review, I cherry-picked Marc's locking changes here. https://github.com/dwerner/casper-node/pull/18 can be closed.

dwerner:
bors try

Makefile
```diff
@@ -1,4 +1,7 @@
 # This supports environments where $HOME/.cargo/env has not been sourced (CI, CLion Makefile runner)
+
+RUSTFLAGS_SSE42AVX = RUSTFLAGS="-C target-feature=+sse4.2,+avx"
```
Fraser999: Is this used anywhere?

dwerner:
bors try

Makefile
```diff
@@ -1,6 +1,6 @@
 # This supports environments where $HOME/.cargo/env has not been sourced (CI, CLion Makefile runner)
 
-RUSTFLAGS_SSE42AVX = RUSTFLAGS="-C target-feature=+sse4.2,+avx"
+RUSTFLAGS_SSE42AVX = RUSTFLAGS="--cfg feature="casper-mainnet" -C target-feature=+sse4.2,+avx"
```
goral09, line 4: Can you please leave a comment what these target features are for?

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -150,6 +180,16 @@ impl EngineState<LmdbGlobalState> {
...
+
+    /// Return a handle to the underlying LMDB environment.
+    pub fn lmdb_environment(&self) -> Arc<LmdbEnvironment> {
+        Arc::clone(&self.state.lmdb_environment)
+    }
+
+    /// Return a handle to the underlying LMDB database.
+    pub fn lmdb_handle(&self) -> Database {
+        self.state.lmdb_trie_store.db
+    }
```
Fraser999: Seem like we don't need these if we also remove the (unused) corresponding methods from the `WasmTestBuilder`.

execution_engine/src/storage/error/mod.rs
```diff
@@ -1,9 +1,9 @@
 //! Storage errors.
 
+/// Errors for LMDB storage implementation.
```
Fraser999:
```suggestion
/// Errors for DB storage implementation.
```

execution_engine/src/lib.rs
```diff
@@ -18,3 +18,55 @@ pub mod core;
 pub mod shared;
 /// Storage for the execution engine.
 pub mod storage;
+
+/// Relative location (to storage) where rocksdb data will be stored.
```
Fraser999: I think all the changes in this file belong down in storage - probably in `storage/transaction_source/db.rs`.

execution_engine/src/lib.rs
```diff
@@ -18,3 +18,55 @@ pub mod core;
...
+const ROCKS_DB_COMPACTION_STYLE: rocksdb::DBCompactionStyle = rocksdb::DBCompactionStyle::Level;
+const ROCKS_DB_ZSTD_MAX_DICT_BYTES: i32 = 256 * 1024;
+const ROCKS_DB_MAX_LEVEL_FILE_SIZE_BYTES: u64 = 512 * 1024 * 1024;
+const ROCKS_DB_MAX_OPEN_FILES: i32 = 768;
+const ROCKS_DB_ZSTD_COMPRESSION_LEVEL: i32 = 3; // Default compression level
+const ROCKS_DB_ZSTD_STRATEGY: i32 = 4; // 4: Lazy
+const ROCKS_DB_WINDOW_BITS: i32 = -14;
+
+/// Column family name for the v1 trie data store.
+pub const ROCKS_DB_TRIE_V1_COLUMN_FAMILY: &str = "trie_v1_column";
```
Fraser999: Doesn't need to be `pub` I think.

execution_engine/src/lib.rs
```diff
@@ -18,3 +18,55 @@ pub mod core;
 pub mod shared;
 /// Storage for the execution engine.
 pub mod storage;
+
+/// Relative location (to storage) where rocksdb data will be stored.
+pub const ROCKS_DB_DATA_DIR: &str = "rocksdb-data";
+
+const ROCKS_DB_BLOCK_SIZE_BYTES: usize = 256 * 1024;
+const ROCKS_DB_COMPRESSION_TYPE: rocksdb::DBCompressionType = rocksdb::DBCompressionType::Zstd;
```
Fraser999: We should import the rocksdb types in this file so we don't need to qualify them all.

execution_engine/src/lib.rs
```diff
@@ -18,3 +18,55 @@ pub mod core;
...
+
+    db_opts.set_compression_type(ROCKS_DB_COMPRESSION_TYPE);
+    db_opts.set_compression_options(
+        ROCKS_DB_WINDOW_BITS,
+        ROCKS_DB_ZSTD_COMPRESSION_LEVEL,
+        ROCKS_DB_ZSTD_STRATEGY,
+        ROCKS_DB_ZSTD_MAX_DICT_BYTES,
+    );
+
+    // seems to lead to a sporatic segfault within rocksdb compaction
```
Fraser999:
```suggestion
    // seems to lead to a sporadic segfault within rocksdb compaction
```

execution_engine/src/storage/global_state/db.rs
```diff
@@ -24,41 +39,144 @@ use crate::{
...
-    pub(crate) environment: Arc<LmdbEnvironment>,
+    pub(crate) lmdb_environment: Arc<LmdbEnvironment>,
     /// Trie store held within LMDB.
-    pub(crate) trie_store: Arc<LmdbTrieStore>,
-    // TODO: make this a lazy-static
+    pub(crate) lmdb_trie_store: Arc<LmdbTrieStore>,
     /// Empty root hash used for a new trie.
     pub(crate) empty_root_hash: Digest,
+    /// Handle to rocksdb.
+    pub rocksdb_store: RocksDbStore,
```
Fraser999:
```suggestion
    pub(crate) rocksdb_store: RocksDbStore,
```

execution_engine/src/storage/global_state/db.rs
```diff
@@ -24,41 +39,144 @@ use crate::{
...
-    /// Environment for LMDB.
-    pub(crate) environment: Arc<LmdbEnvironment>,
-    /// Trie store held within LMDB.
-    pub(crate) store: Arc<LmdbTrieStore>,
+#[derive(Clone)]
+pub struct DbGlobalStateView {
     /// Root hash of this "view".
     pub(crate) root_hash: Digest,
+    /// Handle to rocksdb.
+    pub rocksdb_store: RocksDbStore,
```
Fraser999: These member variables can both be private.

execution_engine/src/storage/global_state/db.rs
```diff
@@ -24,41 +39,144 @@ use crate::{
...
 }
 
-impl LmdbGlobalState {
+impl DbGlobalState {
+    /// Migrate data at the given state roots (if they exist) from lmdb to rocksdb.
+    /// This function uses std::thread::sleep, is otherwise intensive and so needs to be called with
+    /// tokio::task::spawn_blocking.
+    ///
+    /// TODO(dwerner): optionally add multithreading here. Measure time of this method first.
+    pub fn migrate_state_root_to_rocksdb(
```
Fraser999:
```suggestion
    pub(crate) fn migrate_state_root_to_rocksdb(
```

execution_engine/src/storage/global_state/db.rs
```diff
@@ -102,25 +218,64 @@ impl LmdbGlobalState {
...
 
-impl StateReader<Key, StoredValue> for LmdbGlobalStateView {
+fn find_missing_descendants(
+    value_bytes: Bytes,
+    handle: RocksDb,
+    rocksdb: RocksDb,
+    missing_trie_keys: &mut Vec<Digest>,
+    time_in_missing_trie_keys: &mut Duration,
+) -> Result<(), error::Error> {
+    if value_bytes[0] == 0u8 {
```
Fraser999: Is `value_bytes` guaranteed to be non-empty?

execution_engine/Cargo.toml
```diff
@@ -36,18 +38,21 @@ proptest = { version = "1.0.0", optional = true }
 pwasm-utils = "0.16.0"
 rand = "0.8.3"
 rand_chacha = "0.3.0"
+rocksdb = { version = "0.18", default-features = false, features = ["zstd"]}
 schemars = { version = "=0.8.5", features = ["preserve_order"] }
 serde = { version = "1", features = ["derive"] }
 serde_bytes = "0.11.5"
 serde_json = "1"
 thiserror = "1.0.18"
+tokio = { version = "1", features = ["macros", "net", "rt-multi-thread", "sync", "time"] }
```
Fraser999: This seems to be unneeded.

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,262 @@
...
+    ROCKS_DB_TRIE_V1_COLUMN_FAMILY,
+};
+use casper_types::bytesrepr::Bytes;
+
+/// Filename for the LMDB database created by the EE.
+const EE_LMDB_FILENAME: &str = "data.lmdb";
+/// newtype over alt db.
+#[derive(Clone)]
+pub struct RocksDb {
+    pub(crate) db: Arc<DBWithThreadMode<MultiThreaded>>,
```
Fraser999:
```suggestion
    db: Arc<DBWithThreadMode<MultiThreaded>>,
```

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+impl RocksDb {
+    /// Check if a state root has been marked as migrated from lmdb to rocksdb.
+    pub fn is_state_root_migrated(&self, state_root: &[u8]) -> Result<bool, error::Error> {
```
Fraser999:
```suggestion
    pub(crate) fn is_state_root_migrated(&self, state_root: &[u8]) -> Result<bool, error::Error> {
```

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,262 @@
...
+pub struct RocksDbStore {
+    pub(crate) rocksdb: RocksDb,
+    pub(crate) path: PathBuf,
+}
+
+impl RocksDbStore {
+    /// Create a new environment for alternative db.
+    pub fn new(
+        path: impl AsRef<Path>,
+        rocksdb_opts: rocksdb::Options,
```
Fraser999: We should import `Options`.

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,262 @@
...
+    ) -> Result<RocksDbStore, rocksdb::Error> {
+        let db = Arc::new(rocksdb::DBWithThreadMode::<MultiThreaded>::open_cf(
```
Fraser999:
```suggestion
        let db = Arc::new(DBWithThreadMode::<MultiThreaded>::open_cf(
```

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,262 @@
...
+        let rocksdb = RocksDb { db };
+
+        Ok(RocksDbStore {
+            rocksdb,
+            path: path.as_ref().to_path_buf(),
+        })
+    }
+
+    /// Returns the file size on disk of rocksdb.
+    pub fn disk_size_in_bytes(&self) -> usize {
```
Fraser999:
```suggestion
    pub(crate) fn disk_size_in_bytes(&self) -> usize {
```

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,262 @@
...
+        let mut total = 0;
+        for entry in walkdir::WalkDir::new(&self.path)
```
Fraser999: We should import `WalkDir`.

execution_engine/src/storage/transaction_source/db.rs
```diff
@@ -0,0 +1,262 @@
...
+            .into_iter()
+            .filter_map(|e| e.ok())
+        {
+            if entry.metadata().unwrap().is_file() {
+                total += entry.path().size_on_disk().unwrap() as usize;
```
Fraser999: We should avoid the two `unwrap()`s here.  Alternatively, since this function is only really consumed (indirectly) by the test support crate, we could just move it to that crate and not provide it here.  We would maybe need to expose the `RocksDbStore::path` via a public getter in that case.

Also, it seems like we could bind the metadata to a variable and then use the `size_on_disk_fast` function.  Not a big deal though.

execution_engine_testing/test_support/Cargo.toml
```diff
@@ -14,12 +14,15 @@ license = "Apache-2.0"
...
 casper-types = { version = "1.4.5", path = "../../types" }
+filesize = "0.2.0"
 lmdb = "0.8.0"
 log = "0.4.14"
 num-rational = "0.4.0"
 num-traits = "0.2.14"
 once_cell = "1.8.0"
 rand = "0.8.4"
+rocksdb = { version = "0.18", default-features = false, features = ["zstd"]}
+tempfile = "3"
```
Fraser999: Seems like we don't need this?

execution_engine_testing/tests/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Ed Hastings <ed@casperlabs.io>, Henry Till <henrytill@gmail.com>"]
 edition = "2018"
 
 [dependencies]
+anyhow = "1"
```
Fraser999: Seems like we don't need this or `lmdb`.

execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -0,0 +1,192 @@
+use std::{
```
Fraser999, line 1: Could we have a brief README.md for this new binary please?  Similar to the one in `profiling`: just some info about what the binary is for, how to run it, how to interpret the results, etc.

node/src/components/contract_runtime.rs
```diff
@@ -168,36 +169,48 @@ where
...
+                        fatal!(
+                            effect_builder,
+                            "error flushing lmdb environment {:?}",
+                            lmdb_error
+                        )
+                        .await;
+                    } else {
+                        trace!(?result, "commit_genesis response");
+                        responder.respond(result).await;
+                    }
```
Fraser999, line 185: I'm not understanding why this change was made - just want to check I'm not missing some important detail.  Same goes for the upgrade request.

node/src/components/storage.rs
```diff
@@ -96,11 +96,9 @@ use lmdb_ext::{LmdbExtError, TransactionExt, WriteTransactionExt};
 /// Filename for the LMDB database created by the Storage component.
 const STORAGE_DB_FILENAME: &str = "storage.lmdb";
 
-/// We can set this very low, as there is only a single reader/writer accessing the component at any
-/// one time.
-const MAX_TRANSACTIONS: u32 = 1;
+const MAX_TRANSACTIONS: u32 = 1024;
 
-/// One Gibibyte.
+/// One Gigabyte.
```
Fraser999: This was correct before :)

node/src/reactor/initializer.rs
```diff
@@ -199,13 +206,26 @@ impl Reactor {
...
 
-        // TODO: This integrity check is misplaced, it should be part of the components
-        // `handle_event` function. Ideally it would be in the constructor, but since a query to
-        // storage needs to be made, this is not possible.
-        //
-        // Refactoring this has been postponed for now, since it is unclear whether time-consuming
-        // integrity checks are even a good idea, as they can block the node for one or more hours
-        // on restarts (online checks are an alternative).
+        {
+            // TODO(dwerner): profiling this on mainnet- how long are we blocking the initializer.
```
Fraser999: This can be removed now I think.

node/src/reactor/participating.rs
```diff
@@ -1245,6 +1260,76 @@ impl reactor::Reactor for Reactor {
...
+
+        match engine_state
+            .migrate_state_root_to_rocksdb_if_needed(*block_header.state_root_hash(), limit_rate)
+        {
+            Ok(true) => {
+                info!(
+                    block_height = %block_header.height(),
+                    state_root_hash = %block_header.state_root_hash(),
+                    time_migration_took_millis = %start.elapsed().as_millis(),
+                    "successfully state root migrated from lmdb to rocksdb",
```
Fraser999:
```suggestion
                    "successfully migrated state root from lmdb to rocksdb",
```

Fraser999 (commented):
We need changelog entries for EE and the engine test support crates.

Otherwise, mostly just nitpicks.

Makefile
dwerner: It is, however in this revision you reviewed it wasn't because I was debugging a compiler/linking issue that only showed up in CI.

execution_engine/src/lib.rs
dwerner: I don't want to do this because consumers of this library may still want to open the database that is created with these options. That includes tools that shouldn't have to know how to deal with a hypothetical DbGlobalStateOptions type.

execution_engine/src/storage/global_state/db.rs
dwerner: Changed this into an if-let for runtime safety.

execution_engine/src/storage/transaction_source/db.rs
dwerner: I went with exposing a getter on the path and moving this to wasm_test_builder.

execution_engine_testing/tests/bin/disk_use.rs
dwerner: Added to the README.md of execution_engine_testing.

node/src/components/contract_runtime.rs
dwerner: I made this change to ensure that these don't block the reactor thread, and the same for the upgrade request.

execution_engine/src/storage/global_state/db.rs
rafal-ch: Please also [see here](https://github.com/casper-network/casper-node/pull/2680#discussion_r819081059) for a similar issue.

Fraser999 (approved):
Approved assuming https://github.com/dwerner/casper-node/pull/19 is merged in first.

marc-casperlabs:
This PR is getting closed in favor of https://github.com/casper-network/casper-node/pull/2739, which is the same branch with some additional commits pulled in in Dan's absence.

execution_engine/src/storage/trie_store/operations/mod.rs
marc-casperlabs: This seems to have been fixed retroactively via force-push.

execution_engine/src/storage/trie_store/operations/mod.rs
marc-casperlabs: I believe this is fixed in a force-push as well, the only `unreachable!` that remains is a 2 years old one 800 lines down.

execution_engine/src/storage/trie_store/operations/mod.rs
marc-casperlabs: This seems to be have been added: https://github.com/dwerner/casper-node/blob/218377c25745e707a9cb939fb0693d3c876c9107/execution_engine/src/storage/trie_store/operations/mod.rs#L276-L278

execution_engine/src/storage/trie_store/operations/mod.rs
marc-casperlabs: Apparently implemented: https://github.com/dwerner/casper-node/blob/218377c25745e707a9cb939fb0693d3c876c9107/execution_engine/src/storage/trie_store/operations/mod.rs#L259

execution_engine/src/storage/trie_store/operations/mod.rs
marc-casperlabs: See https://github.com/dwerner/casper-node/blob/218377c25745e707a9cb939fb0693d3c876c9107/execution_engine/src/storage/trie_store/operations/mod.rs#L291-L307

Makefile
marc-casperlabs: Addressed here, it seems (no history): https://github.com/dwerner/casper-node/blob/218377c25745e707a9cb939fb0693d3c876c9107/Makefile#L3

execution_engine/src/storage/error/mod.rs
marc-casperlabs: Fixed in https://github.com/dwerner/casper-node/blob/218377c25745e707a9cb939fb0693d3c876c9107/execution_engine/src/storage/error/mod.rs#L3

execution_engine/src/storage/global_state/db.rs
marc-casperlabs: Change is here: https://github.com/dwerner/casper-node/blob/218377c25745e707a9cb939fb0693d3c876c9107/execution_engine/src/storage/global_state/db.rs#L228-L230

PR #2684, opened 2022-02-25
Do not attept to fetch tries when not connected

marc-casperlabs:
When we are not fully connected to any peer yet, the "forever fetch" function will attempt to repeatedly call `fetch_trie` with an empty set of peers. This will cause it to write an error to the logs and drop a responder. This commit fixes this by only calling `fetch_trie` when we have peers to fetch from.

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -121,16 +121,22 @@ async fn fetch_trie_retry_forever(
 ) -> FetchedData<Trie<Key, StoredValue>> {
     loop {
         let peers = ctx.effect_builder.get_fully_connected_peers().await;
-        trace!(?id, "attempting to fetch a trie",);
-        match ctx.effect_builder.fetch_trie(id, peers).await {
-            Ok(fetched_data) => {
-                trace!(?id, "got trie successfully",);
-                return fetched_data;
-            }
-            Err(error) => {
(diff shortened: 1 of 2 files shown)


node/src/effect.rs
```diff
@@ -1258,6 +1258,8 @@ impl<REv> EffectBuilder<REv> {
     }
 
     /// Requests a trie node from a peer.
+    ///
+    /// `peers` must be a **non-empty** set of peers to fetch from.
```
Fraser999:
```suggestion
    /// `peers` should be a **non-empty** set of peers to fetch from.
```

Fraser999 (approved):

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -119,18 +124,38 @@ async fn fetch_trie_retry_forever(
...
+                    trace!(?id, "got trie successfully",);
+                    return fetched_data;
+                }
+                Err(error) => {
+                    warn!(?id, %error, "fast sync could not fetch a trie; trying again")
+                }
             }
         }
+
+        // Emit a warning in the logs only occasionally, avoiding spam when when poorly connected.
```
rafal-ch, line 146:
```suggestion
        // Emit a warning in the logs only occasionally, avoiding spam when poorly connected.
```

PR #2697, opened 2022-02-28
Fetching tries by bytes only

xcthulhu:
Parsing tries during fast sync is CPU intensive for the validator nodes. This is because tries often contain public keys, and elliptic curve operations are expensive.

This PR changes how fetching works so that tries are not parsed.

This PR also removes the trie chunk component.  This is because this component shadowed networking state, and did not restrict fan-out.


dwerner (approved):
lgtm

execution_engine/src/storage/store/mod.rs
```diff
@@ -58,9 +58,20 @@ pub trait Store<K, V> {
         K: AsRef<[u8]>,
         V: ToBytes,
         Self::Error: From<T::Error>,
+    {
+        self.put_raw(txn, key, &value.to_bytes()?)
+    }
+
+    /// Puts a raw `value` into the store at `key` within a transaction, potentially returning an
+    /// error of type `Self::Error` if that fails.
+    fn put_raw<T>(&self, txn: &mut T, key: &K, trie_bytes: &[u8]) -> Result<(), Self::Error>
```
goral09, line 67:
```suggestion
    fn put_raw<T>(&self, txn: &mut T, key: &K, bytes: &[u8]) -> Result<(), Self::Error>
```

Because it's not necessarily a trie, correct?

dwerner: No, it will always be a trie's bytes.

goral09: It is part of `Store` interface which is implemented for all kinds of things. Just b/c it's called in the trie contexts right now does not mean we should be naming (otherwise generic) arguments with callers' contexts. 

Also, you can already see that what you're saying isn't true - `put_raw` is called from `put` and that is called for other types as well.

execution_engine/src/storage/global_state/in_memory.rs
```diff
@@ -275,18 +271,16 @@ impl StateProvider for InMemoryGlobalState {
         &self,
         _correlation_id: CorrelationId,
         trie_key: &Digest,
-    ) -> Result<Option<Trie<Key, StoredValue>>, Self::Error> {
+    ) -> Result<Option<Bytes>, Self::Error> {
         let txn = self.environment.create_read_txn()?;
-        let ret: Option<Trie<Key, StoredValue>> = self.trie_store.get(&txn, trie_key)?;
+        let ret: Option<Bytes> =
+            Store::<Digest, Trie<Digest, StoredValue>>::get_raw(&*self.trie_store, &txn, trie_key)?
```
afck, line 277: Why not `self.trie_store.get_raw`?

hashing/src/error.rs
```diff
@@ -1,21 +1,60 @@
...
+
+    /// Empty merkle proof for trie with chunk.
+    #[error("Chunk with proof has empty merkle proof: {chunk_with_proof:?}")]
+    ChunkWithProofHasEmptyMerkleProof {
+        /// Chunk with empty Merkle proof.
+        chunk_with_proof: ChunkWithProof,
+    },
+
+    /// First digest in indexed Merkle proof did not match hash of chunk.
+    #[error("First digest in Merkle proof did not match hash of chunk. First digest in indexed Merkle proof: {first_digest_in_indexed_merkle_proof:?}. Hash of chunk: {hash_of_chunk:?}.")]
```
afck: Can this line be broken with backslashes?

node/src/components/chain_synchronizer/error.rs
```diff
@@ -126,4 +130,27 @@ pub(crate) enum Error {
...
+}
+
+#[derive(Error, Debug)]
+pub(crate) enum FetchTrieError {
+    /// Fetcher error.
+    #[error(transparent)]
+    FetcherError(#[from] FetcherError<TrieOrChunk>),
+
+    /// Trie was being fetched from peers by chunks but was somehow fetch from storage.
+    #[error("Trie was being fetched from peers by chunks but was somehow fetched from storage. Perhaps there are parallel downloads going on?")]
```
afck: Does this work?
```suggestion
    #[error("Trie was being fetched from peers by chunks but was somehow fetched from storage. \
             Perhaps there are parallel downloads going on?")]
```

hashing/src/error.rs
```diff
@@ -1,21 +1,60 @@
...
 }
 
+/// Error validating a chunk with proof.
+#[derive(thiserror::Error, Debug)]
+pub enum ChunkWithProofVerificationError {
+    /// Indexed Merkle proof verification error.
+    #[error(transparent)]
+    MerkleVerificationError(#[from] MerkleVerificationError),
+
+    /// Empty merkle proof for trie with chunk.
```
rafal-ch:
```suggestion
    /// Empty Merkle proof for trie with chunk.
```

rafal-ch:
A typo in PR description: "Parsing tries during fast sync **can is** CPU intensive for the validator nodes."

rafal-ch:
I'd love to see a benchmark that shows that we're actually faster with this new approach, but it's just a "nice-to-have".

execution_engine/src/storage/store/mod.rs
xcthulhu: > It is part of Store interface which is implemented for all kinds of things.

This function is specialized to the trie store, which only stores trie bytes, so the name seems appropriate.

execution_engine/src/storage/global_state/in_memory.rs
xcthulhu: That would work too.

xcthulhu:
> I'd love to see a benchmark that shows that we're actually faster with this new approach, but it's just a "nice-to-have".

Here's some old profiling I did last year when I was trying to make `retrieve-state` performant on my mac.

image

You can see that `curve25519_dalek` is pinning my CPUs.  Similarly `k256` also creates a lot of CPU pressure (although this screenshot doesn't show this).  This is caused by code in those libraries decompressing the wire formats for public keys.

Dan ran into similar problems in making the rocksdb migration work in https://github.com/casper-network/casper-node/pull/2680 .  The solution is to make `missing_trie_keys` avoid parsing leaves.  But while we were investigating it we forgot to snag a screenshot.

@marc-casperlabs has suggested that we change our `PublicKey` datastructure to be:

```
enum PublicKey {
   System,
   Ed25519([u8 ; 32]),
   Secp256k1([u8; 33])
}
```

While I think we should do this; I suspect we'd need to have them deflated in the consensus component.

@afck @fizyk20 WDYT?

execution_engine/src/storage/global_state/in_memory.rs
xcthulhu: Done in https://github.com/casper-network/casper-node/pull/2697/commits/a23f6e6f3b4d9d243f4435b2728ef1830a25fdab

hashing/src/error.rs
xcthulhu: Fixed in https://github.com/casper-network/casper-node/pull/2697/commits/335032c93ac5d418f7574c67bd2f7dc02f61dbcc

node/src/components/chain_synchronizer/error.rs
xcthulhu: Fixed in https://github.com/casper-network/casper-node/pull/2697/commits/f07d88729f65861a2c09df9beff7f72f5ab8147f

execution_engine/src/storage/global_state/in_memory.rs
xcthulhu: Actually, looks like it can't infer the key type when we use just `self.trie_store.get_raw`, so I've had to revert this in 3b4e1c1

xcthulhu:
bors merge

afck:
@xcthulhu: Would that mean that they need to be deflated for every signing/verification operation?
I agree that for keys in the trie that makes sense, and frequently used consensus keys in memory would need to stay deflated.

xcthulhu:
> @xcthulhu: Would that mean that they need to be deflated for every signing/verification operation?
I agree that for keys in the trie that makes sense, and frequently used consensus keys in memory would need to stay deflated.

@afck That's the idea.

PR #2700, opened 2022-03-01
Remove panicking code from initializer

rafal-ch:
This PR removes a code that could cause panic in the initializer reactor. Appropriate messages are written to the log instead.

### DIFF
--- a/node/src/reactor/initializer.rs
+++ b/node/src/reactor/initializer.rs
@@ -8,7 +8,7 @@ use prometheus::Registry;
 use reactor::ReactorEvent;
 use serde::Serialize;
 use thiserror::Error;
-use tracing::{error, info};
+use tracing::{error, info, warn};
 
 use casper_execution_engine::core::engine_state;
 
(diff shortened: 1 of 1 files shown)


node/src/reactor/initializer.rs
```diff
@@ -93,45 +95,17 @@ impl From<StorageRequest> for Event {
...
-impl From<RestRequest> for Event {
-    fn from(_request: RestRequest) -> Self {
-        unreachable!("no rest requests happen during initialization")
+    fn from(announcement: ChainspecLoaderAnnouncement) -> Self {
+        info!(%announcement, "chainspec announcements received by initializer, ignoring");
+        Event::ChainspecLoaderAnnouncement
     }
 }
 
 impl From<ContractRuntimeAnnouncement> for Event {
```
marc-casperlabs: I think we can just derive this using `#[from]` instead, right? I'd upgrade the `info!` message that handles it to a `warn!` and drop the `error!` here.

marc-casperlabs (commented):
Let's auto-derive the `From` impl, and upgrade the `info!` to `warn!`?

node/src/reactor/initializer.rs
rafal-ch: Auto derive will require extending the enum variant, won't it?

Like so:
```
ChainspecLoaderAnnouncement(ChainspecLoaderAnnouncement),
```

I'd like to avoid it, since we don't need the inner `ChainspecLoaderAnnouncement`.

rafal-ch: > I'd upgrade the `info!` message that handles it to a `warn!` and drop the `error!` here.

`info` level was explicitly requested here. I think it's ok since ignoring the `ChainspecLoaderAnnouncement` can do no harm to the system.

Anyway, I'm happy to change it if requested.

rafal-ch: As agreed during the call, the changes are introduced here: https://github.com/casper-network/casper-node/pull/2700/commits/3b8cbd4b058fb3f79d57362fd7567d7d29f9a1b9

marc-casperlabs (approved):
Looks very good to me!

PR #2702, opened 2022-03-01
Fix macos clippy issue

mpapierski:
On macOS `umask::Mode` is u16 and clippy complains about unnecessary `.into()`.


mpapierski:
Closing as @marc-casperlabs has a solution already to this on his branch which I'll copy and paste temporarily

marc-casperlabs:
Here it is: 7d70b2139

Thanks for bringing it up!

PR #2703, opened 2022-03-01
Add chainspec registry and corresponding RPC/REST endpoint

### LINKED ISSUE #2607: Write chainspec hash to global state, and expose chainspec in an endpoint.
Clients will need the chainspec in the future to fully verify data from the network. E.g. if the hash algorithm is changed, they need to know at which era the switch was made, so they can correctly hash blocks, deploys and the trie. The user shouldn't have to provide the chainspec manually, so the client works with different networks out of the box. It should be able to download and verify any network's chainspec based only on the trusted hash. To support that,
add REST and JSON-RPC endpoints to provide the full chainspec, and
write the chainspec hash to the global state in commit_upgrade.

darthsiroftardis:
CHANGELOG:

* Added a new type `ChainspecRegistry` which contains the hashes of the `chainspec.toml` and will optionally contain the hashes for `accounts.toml` and `global_state.toml`.
* Changed both genesis and upgrade functions to write `ChainspecRegistry` under the fixed `Key::ChainspecRegistry`.
*  Added a new RPC and REST endpoint that returns the bytes of the `chainspec.toml` file.
* Added new `Key::ChainspecRegistry` key variant under which the ChainspecRegistry is written.

Closes #2607


execution_engine/src/core/engine_state/error.rs
```diff
@@ -97,6 +97,9 @@ pub enum Error {
     /// Failed to retrieve the current EraId from the auction state.
     #[error("Failed to retrieve the era_id from the auction state")]
     FailedToRetrieveEraId,
+    #[error("Missing chainspec.toml hash")]
+    /// The chainspec registry did not contain the hash for the chainspec.
```
rafal-ch:
```suggestion
    /// The chainspec registry did not contain the hash for the chainspec.
    #[error("Missing chainspec.toml hash")]
```

execution_engine/src/core/engine_state/genesis.rs
```diff
@@ -1366,6 +1367,31 @@ where
         );
         Ok(())
     }
+
+    pub(crate) fn store_chainspec_registry(
```
rafal-ch, line 1373: Please consider adding a simple UT for this function.

execution_engine_testing/tests/src/test/chainspec_registry.rs
```diff
@@ -0,0 +1,243 @@
...
+};
+use casper_execution_engine::core::{
+    engine_state::{run_genesis_request::RunGenesisRequest, EngineConfig, Error},
+    ChainspecRegistry, CHAINSPEC_RAW, GENESIS_ACCOUNTS_RAW, GLOBAL_STATE_RAW,
+};
+use casper_hashing::Digest;
+
+use casper_types::{EraId, Key, ProtocolVersion};
+use once_cell::sync::Lazy;
+use rand::Rng;
```
rafal-ch: Please sort out the uses: std, 3rd party, casper, etc.

execution_engine_testing/tests/src/test/chainspec_registry.rs
```diff
@@ -0,0 +1,243 @@
...
+        .get(GENESIS_ACCOUNTS_RAW)
+        .expect("must have entry for genesis accounts");
+
+    assert_eq!(*queried_accounts_hash, genesis_account_hash);
+}
+
+#[ignore]
+#[test]
+#[should_panic]
+fn should_fail_to_commit_genesis_when_missing_chainspec_hash() {
```
rafal-ch: If I see correctly, this actually tests when missing both chainspec_hash and genesis accounts (i.e. chainspec registry is empty).

Since there are only two values stored in the chainspec registry now, I'd suggest testing this against all combinations (present vs. missing).

If we add more stuff to the chainspec registry in the future, we might want to move the responsibility of the validation from `GenesisInstaller::store_chainspec_registry()` to the specialized chainspec registry struct itself.

execution_engine_testing/tests/src/test/chainspec_registry.rs
```diff
@@ -0,0 +1,243 @@
...
+#[ignore]
+#[test]
+fn should_write_chainspec_registry_during_an_upgrade() {
+    let mut rng = rand::thread_rng();
+
+    let mut builder = InMemoryWasmTestBuilder::default();
+    builder.run_genesis(&DEFAULT_RUN_GENESIS_REQUEST);
+
+    let mut upgraded_chainspec_registry = ChainspecRegistry::new();
+    let chainspec_bytes_hash = Digest::hash(rng.gen::<[u8; 32]>());
```
rafal-ch: Consider renaming to `upgraded_chainspec_bytes_hash` - it'll make it easier to see what's going on down below at the assert.

execution_engine_testing/tests/src/test/chainspec_registry.rs
```diff
@@ -0,0 +1,243 @@
...
+    upgraded_chainspec_registry.insert(CHAINSPEC_RAW.to_string(), chainspec_bytes_hash);
+    let global_state_toml_hash = Digest::hash(rng.gen::<[u8; 32]>());
```
rafal-ch: Here as well, please consider using `upgraded_global_state_toml_hash`.

execution_engine_testing/tests/src/test/chainspec_registry.rs
```diff
@@ -0,0 +1,243 @@
...
+    let queried_global_state_toml_hash = queried_registry
+        .get(GLOBAL_STATE_RAW)
+        .expect("must have entry for genesis accounts");
+
+    assert_eq!(*queried_global_state_toml_hash, global_state_toml_hash);
+}
+
+#[ignore]
+#[test]
+fn should_upgrade_and_write_registry_from_release_1_4_4() {
```
rafal-ch: The code in this function is like 98% copy&pasted from `should_write_chainspec_registry_during_an_upgrade()`. Please consider some dedup.

execution_engine_testing/tests/src/test/chainspec_registry.rs
```diff
@@ -0,0 +1,243 @@
...
+    let engine_config = EngineConfig::default();
+
+    builder.upgrade_with_upgrade_request(engine_config, &mut upgrade_request);
+
+    let upgrade_result = builder
+        .get_upgrade_result(0)
+        .expect("must have upgrade result")
+        .clone();
+
+    assert!(matches!(upgrade_result, Err(Error::MissingChainspecHash)))
```
rafal-ch: Will this work instead?

```
    assert!(matches!(
        upgrade_result,
        Some(Err(Error::MissingChainspecHash))
    ))
```

node/src/components/chainspec_loader.rs
```diff
@@ -412,6 +453,42 @@ where
             Event::Request(ChainspecLoaderRequest::GetCurrentRunInfo(responder)) => {
                 responder.respond(self.get_current_run_info()).ignore()
             }
+            Event::Request(ChainspecLoaderRequest::GetChainspecRawBytes(responder)) => {
```
rafal-ch: `GetChainspecRawBytes` returns both chainspec bytes and genesis accounts. Is this intended?

node/src/components/chainspec_loader.rs
```diff
@@ -412,6 +453,42 @@ where
...
+                let chainspec_file_path = self
+                    .root_dir
+                    .join(dir_name_from_version(&self.chainspec.protocol_version()))
+                    .join(CHAINSPEC_NAME);
+
+                let chainspec_bytes = match fs::read(chainspec_file_path) {
```
rafal-ch: I wonder if we should actually read a file from `fs` here. Is it possible that chainspec of genesis accounts changes when the node is running? If not, consider loading it once and caching, so we don't need to do the roundtrip to the filesystem each time we handle the GetChainspecRawBytes request.

node/src/components/contract_runtime.rs
```diff
@@ -259,22 +263,72 @@ impl ContractRuntime {
...
+                        chainspec_registry.insert(CHAINSPEC_RAW.into(), chainspec_file_hash);
+                        if let Some(global_state_bytes) =
+                            chainspec_raw_bytes.maybe_global_state_bytes()
+                        {
+                            let global_state_toml_hash = Digest::hash(global_state_bytes);
+                            chainspec_registry
+                                .insert(GLOBAL_STATE_RAW.into(), global_state_toml_hash);
+                        }
+                        chainspec_registry
+                    };
```
rafal-ch: There is some duplication of the code between `ContractRuntimeRequest::Upgrade` and `ContractRuntimeRequest::CommitGenesis`.

If you unify names of bindings (eg. `chainspec_raw_bytes` vs. `raw_chainspec_bytes`) there could be even more duplication :)

Please see if there is a room for some dedup.

types/CHANGELOG.md
```diff
@@ -15,12 +15,14 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Added
 * Added new `bytesrepr::Error::NotRepresentable` error variant that represents values that are not representable by the serialization format.
+* Added new `Key::ChainspecRegistry` key variant under which the ChainspecRegistry is written.
+* Added a new type `WithdrawPurses` which are meant to represent `UnbondingPurses` as they exist in current live networks.
 
 ### Changed
 * Extended `UnbondingPurses` to take a new field `new_validator` which represents the validator to whom tokens will be re-delegated.
 
 ### Added
```
rafal-ch: This `### Added` is no longer needed.

types/CHANGELOG.md
```diff
@@ -15,12 +15,14 @@ All notable changes to this project will be documented in this file.  The format
...
-* Added a new type `WithdrawPurses` which are meant to represent `UnbondingPurses` as they exist in current live networks.
+
 
 ### Changed
```
rafal-ch, line 21: Maybe we can also merge the `Changed` parts while we're here :)

rafal-ch:
Last but not least... :)
I'd love to see a corresponding `casper-test` test that checks the new endpoints.

node/src/components/chainspec_loader.rs
darthsiroftardis: No, so the genesis accounts file will only be present for the 1_0_0 PV for a given network. So one could imagine in a 1.2 PV of a given network there is only the chainspec bytes. So depending on the given PV we may or may not need the value

darthsiroftardis:
> Last but not least... :) I'd love to see a corresponding `casper-test` test that checks the new endpoints.

Yes, @Fraser999 and I discussed this, however, we think its best to see those tests as a follow up PR

execution_engine/src/core/engine_state/genesis.rs
darthsiroftardis: We UT this function via the EE tests, thats the pattern we've kept to so far

node/src/components/chainspec_loader.rs
darthsiroftardis: It should be returning the chainspec bytes, and optionally returning the genesis accounts and the global state toml

execution_engine/src/core/engine_state/error.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/chainspec_registry.rs
darthsiroftardis: Sorted out uses, thanks for pointing it out, idk how I missed it

execution_engine_testing/tests/src/test/chainspec_registry.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/chainspec_registry.rs
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/chainspec_registry.rs
darthsiroftardis: Yeah, thanks for pointing this out, I've collated the code into one function now

execution_engine_testing/tests/src/test/chainspec_registry.rs
darthsiroftardis: This could work, but I prefer the match on the concrete error, and unwrapping it earlier. Do you think we must implement this for some reason I don't get?

node/src/components/contract_runtime.rs
darthsiroftardis: There was, deduped as suggested

types/CHANGELOG.md
darthsiroftardis: Removed the unnecessary section

types/CHANGELOG.md
darthsiroftardis: Merged as requested

execution_engine_testing/tests/src/test/chainspec_registry.rs
rafal-ch: That's ok :+1:

node/src/components/chainspec_loader.rs
rafal-ch: In that case I would consider renaming the `ChainspecRawBytes` struct. Maybe `ChainspecContext` or `GenesisData`, etc.

rafal-ch: > Yes, @Fraser999 and I discussed this, however, we think its best to see those tests as a follow up PR

Great!
Are you going to request a ticket?

rafal-ch:
> Yes, @Fraser999 and I discussed this, however, we think its best to see those tests as a follow up PR

Great!
Are you going to request a ticket?

node/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ All notable changes to this project will be documented in this file.  The format
 * In addition to `consensus` and `deploy_requests`, the following values can now be controlled via the `[network.estimator_weights]` section in config: `gossip`, `finality_signatures`, `deploy_responses`, `block_requests`, `block_responses`, `trie_requests` and `trie_responses`.
 * Nodes will now also gossip deploys onwards while joining.
   Added run-mode field to the `/status` endpoint and the `info_get_status` JSON-RPC.
+* Added a new RPC and REST endpoint that returns the bytes of the `chainspec.toml` file.
```
rafal-ch: Hmm, can we actually explicitly mention the endpoints here? I think it's `info_get_chainspec` and `/chainspec`?

node/src/components/chainspec_loader.rs
darthsiroftardis: No, because its not a "context" object, and the data isn't specific to genesis only,

Fraser999 (approved):

execution_engine/src/core/engine_state/chainspec_registry.rs
```diff
@@ -0,0 +1,154 @@
...
+        if let Some(global_state_raw_hash) = self.global_state_raw_hash {
+            map.insert(
+                Self::GLOBAL_STATE_RAW_MAP_KEY.to_string(),
+                global_state_raw_hash,
+            );
+        }
+        map
+    }
+
+    fn from_map(map: BytesreprChainspecRegistry) -> Result<Self, bytesrepr::Error> {
```
rafal-ch: This signature resembles (or is identical to) `TryFrom::try_from()`.

Can we make it like so?
```
impl TryFrom<BytesreprChainspecRegistry> for ChainspecRegistry {
    type Error = bytesrepr::Error;

    fn try_from(map: BytesreprChainspecRegistry) -> Result<Self, Self::Error> {
        ...
    }
}
```

node/src/components/chainspec_loader.rs
```diff
@@ -453,42 +425,9 @@ where
...
-                let raw_chainspec_bytes = ChainspecRawBytes::new(
-                    chainspec_bytes,
-                    maybe_genesis_accounts_bytes,
-                    maybe_global_state_bytes,
-                );
-
-                responder.respond(raw_chainspec_bytes).ignore()
-            }
+            Event::Request(ChainspecLoaderRequest::GetChainspecRawBytes(responder)) => responder
+                .respond(Arc::clone(&self.chainspec_raw_bytes))
```
rafal-ch, line 429: Nice!
I like that we don't touch the filesystem in the event handler :+1:

rafal-ch (approved):
LGTM with one minor nit to consider :+1:

execution_engine/src/core/engine_state/chainspec_registry.rs
darthsiroftardis: Applied suggestion

PR #2712, opened 2022-03-04
GH-2443: Get auction info backwards compatibility

### LINKED ISSUE #2443: The casper-client cannot retrieve get-auction-info for blocks from version 1.3.4 and before.

mpapierski:
Closes #2443

This PR fixes a problem of the get-auction-state query for pre #1814 (introduction of `Key::SystemContractRegistry`) blocks by passing down an instance of `SystemContractRegistry` initialized from a highest block's state root hash at component initialization time. This avoids querying the registry from a specified state which may not exist for older blocks.

### DIFF
--- a/node/src/components/contract_runtime.rs
+++ b/node/src/components/contract_runtime.rs
@@ -19,12 +19,12 @@ use datasize::DataSize;
 use lmdb::DatabaseFlags;
 use prometheus::Registry;
 use serde::Serialize;
-use tracing::{debug, info, trace};
+use tracing::{debug, error, info, trace};
 
 use casper_execution_engine::{
     core::engine_state::{
         self, genesis::GenesisSuccess, EngineConfig, EngineState, GetEraValidatorsError,
-        GetEraValidatorsRequest, UpgradeConfig, UpgradeSuccess,
+        GetEraValidatorsRequest, SystemContractRegistry, UpgradeConfig, UpgradeSuccess,
     },
     shared::{newtypes::CorrelationId, system_config::SystemConfig, wasm_config::WasmConfig},
     storage::{
@@ -139,6 +139,7 @@ pub(crate) struct ContractRuntime {
 
     /// Finalized blocks waiting for their pre-state hash to start executing.
     exec_queue: ExecQueue,
+    system_contract_registry: SystemContractRegistry,
 }
 
 impl Debug for ContractRuntime {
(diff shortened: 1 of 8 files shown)


goral09:
@mpapierski What's your plan to verify that this fixes the problem?

mpapierski:
> @mpapierski What's your plan to verify that this fixes the problem?

An integration network and casper-client quering get-auction-info through RPC for 1.3.x state root hashes seems like easiest approach

goral09 (approved):

goral09:
> > @mpapierski What's your plan to verify that this fixes the problem?
> 
> An integration network and casper-client quering get-auction-info through RPC for 1.3.x state root hashes seems like easiest approach

Is that going to be a separate PR?

darthsiroftardis (commented):
The code seems fine, but I think we should have a test to see that the fix works before we merge this in

node/src/components/contract_runtime.rs
```diff
@@ -139,6 +139,7 @@ pub(crate) struct ContractRuntime {
 
     /// Finalized blocks waiting for their pre-state hash to start executing.
     exec_queue: ExecQueue,
+    system_contract_registry: SystemContractRegistry,
```
Fraser999, line 142: Please add a comment as to why we need this member variable, since it's not obvious.

node/src/components/contract_runtime/operations.rs
```diff
@@ -31,8 +31,10 @@ use casper_execution_engine::{
 };
 
 /// Executes a finalized block.
+#[allow(clippy::too_many_arguments)]
 pub fn execute_finalized_block(
     engine_state: &EngineState<LmdbGlobalState>,
+    system_contract_registry: &SystemContractRegistry,
```
Fraser999, line 37: I don't think we should be passing `system_contract_registry` here.  In the `feat-fast-sync` branch, we only call `contract_runtime.set_initial_state` in the participating reactor, but I believe `execute_finalized_block` can be called from the chain synchronizer in the joiner reactor.  In that case, it would have an empty `SystemContractRegistry`.

We should probably change `get_era_validators` to take an optional `SystemContractRegistry`, and if it's `None` try to retrieve from the tracking copy.

Fraser999 (changes requested):
We need a changelog entry (probably just in node) for the fix.

Blocking the PR for now until we can discuss my second comment.

node/src/components/contract_runtime/operations.rs
goral09: > In the feat-fast-sync branch, we only call `contract_runtime.set_initial_state` in the participating reactor, but I believe` execute_finalized_block` can be called from the chain synchronizer in the joiner reactor. In that case, it would have an empty `SystemContractRegistry`.

This PR is made to `dev`, not `feat-fast-sync` though. Making `SystemContractRegistry` optional in the `ContractRuntime` component will add unnecessary work to the PR, code that won't quite make sense for `dev`. We can make that work when migrating this fix to `feat-fast-sync`. 

Also, you're right that it optional, and reading from the `scratch_global_state`, is a solution.

PR #2714, opened 2022-03-07
Remove `verify_accounts` configuration option

### LINKED ISSUE #2297: Remove verify_accounts config option and VerifyDeployCryptographicValidity event from deploy acceptor
@darthsiroftardis pls add description

darthsiroftardis:
CHANGELOG:

- Removed the `verify_accounts` configuration setting from the node
- The deploy acceptor will always verify accounts by default
- Removed `VerifyDeployCryptographicValidity` event as it was superfluous 

Closes #2297


node/src/reactor/joiner.rs
```diff
@@ -503,11 +505,8 @@ impl reactor::Reactor for Reactor {
         let block_header_by_hash_fetcher: Fetcher<BlockHeader> =
             Fetcher::new("block_header_by_hash", config.fetcher, registry)?;
 
-        let deploy_acceptor = DeployAcceptor::new(
-            config.deploy_acceptor,
-            &*chainspec_loader.chainspec(),
-            registry,
-        )?;
+        let deploy_acceptor =
+            DeployAcceptor::new(VERIFY_ACCOUNTS, &*chainspec_loader.chainspec(), registry)?;
```
goral09, line 509: If it's not supposed to be changed, why even bother maintaining the option to do so via constructor? It adds confusion b/c tests should still be written for both cases (`true` and `false`).

node/CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ All notable changes to this project will be documented in this file.  The format
 * Connection handshake timeouts can now be configured via the `handshake_timeout` variable (they were hardcoded at 20 seconds before).
 * `Key::SystemContractRegistry` is now readable and can be queried via the RPC.
 
+### Removed
+* Removed the `verify_accounts` configuration setting from `config.toml`
```
goral09, line 30: Please add an explanation what observable changes this has on node operators (i.e. explaining that the node will verify account's existance, minimum balance, etc.).

node/src/reactor/joiner.rs
darthsiroftardis: I've done it as the fetcher test uses a macro to instantiate its DA component, and it always goes via the constructor, and introducing the account verification for the gossiper/fetcher tests will cause them to fail, or require a rework of the test suites for those two components

darthsiroftardis: I wanted to remove it from the constructor as well, its just a constraint from a macro usage in the tests

goral09: Can we have a special constructor for tests-only then? As an intermediate step:
```rust
#[cfg(test)]
fn new(account_verification: bool, ...) -> Self { ... }

fn new(...) -> Self { ... }
```
Would that work?

Long term though, I think we should just fix the tests.

rafal-ch (approved):
LGTM, assuming that the comments from @goral09 will be addressed :+1:

Fraser999:
Closed in favour of #2746.

PR #2753, opened 2022-03-18
Add methods to advance eras.

### LINKED ISSUE #2690: Add utility method on WasmTestBuilder to advance eras by configured amount.
When we want to force an effect of (un)bonding requests, we have to run the StepRequest for AUCTION_DELAY times. Almost always we use DEFAULT_AUCTION_DELAY and write:
for _ in 0..=DEFAULT_AUCTION_DELAY {
        let step_request = StepRequestBuilder::new()
            .with_parent_state_hash(builder.get_post_state_hash())
            .with_protocol_version(ProtocolVersion::V1_0_0)
            .with_next_era_id(builder.get_era().successor())
            .with_run_auction(true)
            .build();

        builder
            .step(step_request)
            .expect("must execute third step request post upgrade");
    }
This pattern is so common that we should have an utility method that hides this.

HansWestman74:
# Add methods to advance eras.

* Added `advance_eras_by(num_eras: u64, Option <fn(StepRequestBuilder) -> StepRequestBuilder>) `
* Added `advance_eras_by_default_auction_delay()`
* Replaced boilerplate code with these methods.

closes #2690


goral09 (changes requested):
New code that uses the helper method is not equivalent to the old one. We used to iterate `0..n` and now it's `0..=n`.

goral09 (commented):
There are a couple more places where you can reuse it:
* https://github.com/casper-network/casper-node/blob/93d415207b5cff76766d26d2b7454d6dcc2da1ea/execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs#L4110-L4118
* https://github.com/casper-network/casper-node/blob/93d415207b5cff76766d26d2b7454d6dcc2da1ea/execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs#L4130-L4138
* if you add the `reward_items` to the API of the new method you can use it here https://github.com/casper-network/casper-node/blob/93d415207b5cff76766d26d2b7454d6dcc2da1ea/execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs#L4166-L4177
* https://github.com/casper-network/casper-node/blob/93d415207b5cff76766d26d2b7454d6dcc2da1ea/execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs#L4264-L4271
* https://github.com/casper-network/casper-node/blob/93d415207b5cff76766d26d2b7454d6dcc2da1ea/execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs#L4329-L4337

And this is just one file. Go through the whole repo and replace the boilerplate with new helper.

Fraser999 (commented):
As @goral09 says, we should look for other places we can remove boilerplate.  Probably searching for instances of `StepRequestBuilder::new()` would be a decent approach.

Also, please make sure to update the changelog for the test_support crate.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -1171,21 +1171,29 @@ where
...
 
     /// Advances eras by num_eras
-    pub fn advance_eras_by(&mut self, num_eras: u64) {
-        for _ in 0..=num_eras {
-            let step_request = StepRequestBuilder::new()
+    pub fn advance_eras_by(
+        &mut self,
+        num_eras: u64,
+        configure: Option<fn(StepRequestBuilder) -> StepRequestBuilder>,
+    ) {
```
goral09, line 1166: Personally, I don't like methods that take closure too much. I think extending the `advance_eras_by` (and `advance_eras_by_default_auction_delay`) with `rewards` (which could be of type `Iterator<Item=T>, T: Into<RewardItem>`) and `slashes` is still fine. @Fraser999 , wdyt?

Fraser999: I agree about disliking the optional closure.

I don't think I'd try and handle that via these two helper methods, as it's not obvious which step to add those rewards and slashes to.  I'd either just not try to handle that via these helpers (and leave that code as it is in `dev`), or since it seems like these args are only used when advancing by a single era, maybe provide a new helper which only advances by a single era, but takes the two collections as args.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3453,17 +3430,14 @@ fn should_handle_redelegation_to_inactive_validator() {
...
-        .step(step_request)
-        .expect("must execute step request");
+    builder.advance_eras_by(
+        1,
+        [
+            RewardItem::new(NON_FOUNDER_VALIDATOR_1_PK.clone(), 1),
+            RewardItem::new(NON_FOUNDER_VALIDATOR_2_PK.clone(), 1),
+        ]
+        .into(),
+    );
```
goral09: As @Fraser999 noted, it seems like it's only the case for when we advance by a single era that we want to add the rewards. Let's change the API of that method then (`advance_eras_by_default_auction_delay`) and please, use `vec![...]` rather than `[...].into()`.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -1168,4 +1169,27 @@ where
         self.transforms = Vec::new();
         self
     }
+
+    /// Advances eras by num_eras
+    pub fn advance_eras_by(&mut self, num_eras: u64, mut reward_items: Vec<RewardItem>) {
```
goral09: Why does `reward_items` has to be mutable?

HansWestman74: The Vec has ownership of the RewardItems which does not implement the copy trait. Thus, we need to pop off the rewards items one by one from reward_items Vec (which then must be declared mutable) and hand over ownership of the popped reward_item to the step_request_builder.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
HansWestman74: Looks like we have one counter example in bids.rs line 3890.

Made the change to vec! instead of into().

goral09: https://github.com/HansWestman74/casper-node-1/blob/GH-2690/execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs#L3890 this? I can't see it.

HansWestman74: This one:

https://github.com/HansWestman74/casper-node-1/blob/a1ad558f07f8384ec73a7ecaca56755d3f1e50b1/execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs#L3900

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -1168,4 +1169,27 @@ where
...
+        for _ in 0..num_eras {
+            let mut step_request_builder = StepRequestBuilder::new()
+                .with_parent_state_hash(self.get_post_state_hash())
+                .with_protocol_version(ProtocolVersion::V1_0_0)
+                .with_next_era_id(self.get_era().successor())
+                .with_run_auction(true);
+
+            while let Some(reward_item) = reward_items.pop() {
```
goral09: I think this can't work. You pass in a vector of `reward_items` that you pop from, which means you can only pop so much and then you run out of elements after which you won't add any new reward items. It means that you will add reward items only for a _single era_ and not `num_eras` of iterations. And the fact that you have `while let Some(_) = reward_items.pop` hides the problem.

This means that the new code is not equivalent to the old one but also, much worse, that current tests are not good enough to find it!

goral09 (changes requested):
New release blocker.

goral09:
I've created a [PR against your repo](https://github.com/HansWestman74/casper-node-1/pull/1) that solves the problem mentioned in https://github.com/casper-network/casper-node/pull/2753#discussion_r834639214 . It does NOT solve the problem that none of our tests caught the problem :/

goral09:
@HansWestman74 We need to fix the build and update the EE's `CHANGELOG.md` appropriately.

execution_engine_testing/test_support/CHANGELOG.md
```diff
@@ -15,10 +15,13 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Added
 * Added `WasmTestBuilder::get_execution_journals` method for returning execution journals for all test runs.
+* Added `WasmTestBuilder::advance_era`, `WasmTestBuilder::advance_eras_by`, and `WasmTestBuilder::advance_era`, and `WasmTestBuilder::advance_eras_by_default_auction_delay` to remove boilerplate code for tests.
```
goral09:
```suggestion
* Added `WasmTestBuilder::advance_era`, `WasmTestBuilder::advance_eras_by`, and `WasmTestBuilder::advance_era`, and `WasmTestBuilder::advance_eras_by_default_auction_delay` to advance chain and run auction contract in test environment.
```

execution_engine_testing/test_support/CHANGELOG.md
```diff
@@ -15,10 +15,13 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Added
 * Added `WasmTestBuilder::get_execution_journals` method for returning execution journals for all test runs.
+* Added `WasmTestBuilder::advance_era`, `WasmTestBuilder::advance_eras_by`, and `WasmTestBuilder::advance_eras_by_default_auction_delay` to advance chain and run auction contract in test environment.
 
 ### Changed
 * `WasmTestBuilder::get_transforms` is deprecated in favor of `WasmTestBuilder::get_execution_journals`.
 * `deploy_hash` field is now defaulted to a random value rather than zeros in `DeployItemBuilder`.
+* Replaced boilerplate code in bids.rs and distribute.rs with methods `WasmTestBuilder::advance_era`, `WasmTestBuilder::advance_eras_by`, and `WasmTestBuilder::advance_eras_by_default_auction_delay`
```
goral09: This is not necessary I think. This change is purely internal and it's not affecting users - i.e. it's not changing the API.

goral09 (approved):

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -1168,4 +1169,29 @@ where
         self.transforms = Vec::new();
         self
     }
+
+    /// Advances eras by num_eras
+    pub fn advance_eras_by(&mut self, num_eras: u64, reward_items: Vec<RewardItem>) {
```
mpapierski:
```suggestion
    pub fn advance_eras_by(&mut self, num_eras: u64, reward_items: impl IntoIterator<Item = RewardItem>) {
```

Why not this - in bunch of places I saw invocations like:

```rust
  .advance_era(vec![reward_item1, reward_item2])
```
We can make it flexible for users so they could just call it with array `advance_era([reward_item1, reward_item2])` if they have reward items created already

execution_engine_testing/test_support/src/step_request_builder.rs
```diff
@@ -48,6 +48,13 @@ impl StepRequestBuilder {
         self
     }
 
+    /// Appends the given vector of [`RewardItem`] into `reward_items`.
+    pub fn with_reward_items(mut self, reward_items: Vec<RewardItem>) -> Self {
```
mpapierski:
```suggestion
    pub fn with_reward_items(mut self, reward_items: impl IntoIterator<Item = RewardItem>) -> Self {
```
This way you could call it with any iterable, and users can use arrays instead of vectors etc.

mpapierski (approved):
One suggestion on the API usability (accept iterator not a vec) otherwise LGTM

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -1171,27 +1171,38 @@ where
...
+    ) {
+        let step_request_builder = StepRequestBuilder::new()
+            .with_protocol_version(ProtocolVersion::V1_0_0)
+            .with_reward_items(reward_items)
+            .with_run_auction(true);
+
         for _ in 0..num_eras {
-            let step_request_builder = StepRequestBuilder::new()
+            let step_request = step_request_builder
+                .clone()
```
goral09: This probably doesn't improve the performance and makes it less readable IMO but let's not change it in this PR, it's already been too long.

execution_engine_testing/test_support/src/step_request_builder.rs
goral09: Done in [`bc80b7d` (#2753)](https://github.com/casper-network/casper-node/pull/2753/commits/bc80b7dfcaba9b846c5d7e01aeef6f52e8042867)

execution_engine_testing/test_support/src/wasm_test_builder.rs
goral09: Done in [`bc80b7d` (#2753)](https://github.com/casper-network/casper-node/pull/2753/commits/bc80b7dfcaba9b846c5d7e01aeef6f52e8042867)

HansWestman74:
bors merge

goral09:
@HansWestman74 , can you fix the build and merge the PR?

HansWestman74:
bors merge

PR #2776, opened 2022-03-22
GH-2740: Move system contracts logic into execution engine.

### LINKED ISSUE #2740: Move system contracts from types into execution engine
This change will allow adding logging inside auction contract. We should initially copy the logic to maintain backwards compatibility. As part of separate tickets we should refactor system contract traits to not use traits anymore incl. providers. System contract related structures (i.e. Bid, Delegator etc.) should still live in casper-types.
Trait based system contracts are artifacts of the past where we had to use both Wasm system contracts, and native system contracts and we deduplicated the code this way. We only support native system contracts and dont need this overhead anymore.

As per Ed - "a single 1 point ticket should be sufficient to simply move the folder structure into EE and move on"

mpapierski:
Closes #2740

Moves system contract logic into ./execution_engine from ./types. Whatever needs to be kept in ./types stays in ./types (i.e. Bid structure thats part of StoredValue etc etc).

This PR should allow us to start adding much necessary logging around system contracts.


types/src/system/auction/withdraw_purse.rs
```diff
@@ -0,0 +1,195 @@
...
+        let secret_key = SecretKey::ed25519_from_bytes([44; SecretKey::ED25519_LENGTH]).unwrap();
+        PublicKey::from(&secret_key)
+    }
+
+    fn unbonder_public_key() -> PublicKey {
+        let secret_key = SecretKey::ed25519_from_bytes([45; SecretKey::ED25519_LENGTH]).unwrap();
+        PublicKey::from(&secret_key)
+    }
+
+    fn amount() -> U512 {
```
darthsiroftardis, line 155: Could we rename this function, eg: `one_less_than_max_amount` or something similar, but given its a test only function, this is  nitpick from my part

darthsiroftardis (approved):
LGTM, I am confused on whether this needs some CHANGLOG entries, 

@Fraser999 do we need CHANGLOG entries for this

Fraser999:
> @Fraser999 do we need CHANGLOG entries for this

Given that there's no actual change to the behaviour or the API except for removing the 4 traits which have always been implementation details of the EE, I think it's fine to omit changelog entries.

Fraser999 (approved):

PR #2785, opened 2022-03-23
Sign immediate switch blocks

### LINKED ISSUE #2567: Sign immediate switch blocks
At genesis or after an upgrade, we now create a predetermined switch block proposed by "the system", which constitutes a whole single-block era not associated with any consensus instance.
However, it would still be useful to create finality signatures for that block.

Fraser999:
This PR updates immediate switch blocks (blocks created immediately after genesis or upgrade) to be signed.

The signing is performed in the participating reactor's constructor, which is necessary until the architecture is updated to use a single reactor.

A 10 second delay precedes broadcasting the new signatures, in order to allow peer connections to form.

This has been tested to work at genesis and at an upgrade using nctl.  As well as the immediate switch blocks being able to be retrieved via the JSON-RPC server with all validator signatures applied, all the signatures are also emitted in the event stream.

Closes #2567.


node/src/effect.rs
```diff
@@ -1749,7 +1735,11 @@ impl<REv> EffectBuilder<REv> {
             // we don't support getting the validators from before the last emergency restart
             return None;
         }
-        self.get_key_block_header_for_era_id_from_storage(era_id)
+        // Era 0 contains no blocks other than the genesis immediate switch block which can be used
+        // to get the validators for era 0.  For any other era `n`, we need the switch block from
+        // era `n-1` to get the validators for `n`.
```
afck, line 1740: I wonder whether it would be better to not sign the genesis block at all.
Signing it this way suggests to the user that it would be safe to request the block with height 0, take the validator set it contains, and then check the corresponding signatures.

Unless you compute the genesis block yourself, you always need a trusted hash, so you never need to check the genesis block's signatures.

node/src/components/chain_synchronizer.rs
```diff
@@ -351,26 +358,52 @@ impl ChainSynchronizer {
...
             Err(error) => {
                 error!(%error, "failed to execute block");
-                fatal!(effect_builder, "{}", error).ignore()
+                return fatal!(effect_builder, "{}", error).ignore();
             }
-        }
+        };
+
+        // If the upgrade block is `None`, this is a genesis switch block, so we can use the
+        // validator set from that.
```
afck, line 387: I wouldn't sign the genesis block (see other comment).

node/src/components/chain_synchronizer.rs
```diff
@@ -276,7 +277,7 @@ impl ChainSynchronizer {
                     PublicKey::System,
                 );
 
-                self.execute_block(effect_builder, initial_pre_state, finalized_block)
+                self.execute_block(effect_builder, None, initial_pre_state, finalized_block)
```
afck: We should rename that to `execute_immediate_switch_block`, since the chain synchronizer also executes regular blocks elsewhere.

node/src/effect.rs
Fraser999: I see your point, but I'm a bit torn on it.  I mildly prefer the consistency of signing all blocks, and ultimately the genesis block will always be an anomaly in that it doesn't have a parent (we use a sentinel value for the parent hash rather than omitting it via `None` for example)  We're also already signing the genesis block in current and previous versions of the code, and your point applies equally there I think?

I'll wait for feedback from other reviewers before actioning this thanks :)

node/src/components/chain_synchronizer.rs
Fraser999: Addressed in [3f0bfa0](https://github.com/casper-network/casper-node/pull/2785/commits/3f0bfa034550037b59232d8bef9b78cf2f007f4e).

node/src/effect.rs
goral09: What's the difference between Genesis block (is that a switch block created right after the start? In the past there was no Genesis block) and a switch block right after an emergency upgrade? We should be consistent w.r.t. to security aspect rather than UX in those cases in my opinion.

node/src/effect.rs
```diff
@@ -1749,7 +1735,11 @@ impl<REv> EffectBuilder<REv> {
...
+        let era_before = era_id.saturating_sub(1);
```
rafal-ch, line 1741: Not required, but consider adding `EraId::predecessor()` similarly to the already existing `EraId::successor()`.

node/src/effect.rs
afck: I just find it misleading to take the expected signers out of the block itself. It's fine in this case because we created the block ourselves and the validator set in the genesis block happens to be the actual set of validators for era 0. But what determines that validator set is really the original `accounts.toml` file, not the set specified in the genesis block header. I also think that was how it was originally implemented, when there were no immediate switch blocks. And I think that _if_ we sign the genesis block, we should use the validator set from there, to avoid misleading (even though correct) code.

I agree that the switch block after an emergency restart has the same problem.

In both cases, there are only two ways to trust the block:
* Either you have the `accounts.toml` or the emergency upgrade specification, in which case you might as well compute the block yourself.
* Or you have a trusted hash from after genesis/the upgrade.

I don't see any conceivable case where you could make any use of the signatures.

Anyway, it's not a blocker.

afck (approved):

node/src/effect.rs
Fraser999: Hmm - I think I'll leave as is thanks, since we'd be replacing a one-line call here with a call which is a little more opaque IMO.  There might be cases where we'd want `checked_sub` so we get `None` in the case of era 0.

fizyk20 (approved):

node/src/effect.rs
Fraser999: As discussed elsewhere, we'll merge as is, with a view to discussing further if required.

Fraser999:
bors merge

Fraser999:
bors merge

PR #2786, opened 2022-03-23
Add the highest available block height to the RPC error response

### LINKED ISSUE #2780: Change rpc_server::rpcs::common::ErrorData to also include HighestBlockHeight(u64)
Change the get_lowest_contiguous_block_height_from_storage to get_highest_contiguous_range_from_storage and update as necessary in storage to provide the lowest contiguous block height and the highest block height in a single responde

rafal-ch:
This PR changes the response to the block query. In case of requesting a block that is missing from the node, the response will now contain the highest contiguous height range of fully available blocks, for example:

```json
{
	"jsonrpc": "2.0",
	"id": "1",
	"error": {
		"code": -32001,
		"message": "block at height 200 not stored on this node",
		"data": {
			"highest_contiguous_block_height_range": [3, 93]
		}
	}
}
```

Closes https://github.com/casper-network/casper-node/issues/2780

### DIFF
--- a/node/src/components/rpc_server.rs
+++ b/node/src/components/rpc_server.rs
@@ -294,13 +294,16 @@ where
                 }
                 .ignore()
             }
-            Event::RpcRequest(RpcRequest::GetLowestContiguousBlockHeight { responder }) => {
+            Event::RpcRequest(RpcRequest::GetHighestContiguousBlockHeightRange { responder }) => {
                 effect_builder
-                    .get_lowest_contiguous_block_height_from_storage()
-                    .event(move |height| Event::GetLowestContiguousBlockHeightResult {
-                        height,
-                        main_responder: responder,
-                    })
+                    .get_highest_contiguous_block_range_from_storage()
+                    .event(
+                        move |(low, high)| Event::GetHighestContiguousBlockHeightRange {
+                            low,
+                            high,
+                            main_responder: responder,
+                        },
+                    )
             }
             Event::GetBlockResult {
                 maybe_id: _,
@@ -337,10 +340,11 @@ where
                 peers,
                 main_responder,
             } => main_responder.respond(peers).ignore(),
-            Event::GetLowestContiguousBlockHeightResult {
-                height,
+            Event::GetHighestContiguousBlockHeightRange {
+                low,
+                high,
                 main_responder,
-            } => main_responder.respond(height).ignore(),
+            } => main_responder.respond((low, high)).ignore(),
         }
     }
 }
--- a/node/src/components/rpc_server.rs
+++ b/node/src/components/rpc_server.rs
@@ -295,15 +295,16 @@ where
                 .ignore()
             }
             Event::RpcRequest(RpcRequest::GetHighestContiguousBlockHeightRange { responder }) => {
-                effect_builder
-                    .get_highest_contiguous_block_range_from_storage()
-                    .event(
-                        move |(low, high)| Event::GetHighestContiguousBlockHeightRange {
-                            low,
-                            high,
-                            main_responder: responder,
-                        },
-                    )
+                async move {
+                    responder
+                        .respond(
+                            effect_builder
+                                .get_highest_contiguous_block_range_from_storage()
+                                .await,
+                        )
+                        .await
+                }
+                .ignore()
             }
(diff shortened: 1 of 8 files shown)


node/src/effect/requests.rs
```diff
@@ -729,7 +729,7 @@ impl Display for RpcRequest {
             RpcRequest::GetDeploy { hash, .. } => write!(formatter, "get {}", hash),
             RpcRequest::GetPeers { .. } => write!(formatter, "get peers"),
             RpcRequest::GetStatus { .. } => write!(formatter, "get status"),
-            RpcRequest::GetLowestContiguousBlockHeight { .. } => {
+            RpcRequest::GetHighestContiguousBlockHeightRange { .. } => {
                 write!(formatter, "get lowest contiguous block height")
```
Fraser999: This should be updated.

node/src/components/rpc_server/rpcs/common.rs
```diff
@@ -77,35 +77,34 @@ pub(super) async fn run_query_and_encode<REv: ReactorEventT>(
 #[serde(deny_unknown_fields)]
 #[serde(rename_all = "snake_case")]
 pub enum ErrorData {
-    /// The lowest block height from which this node has contiguous blocks to the tip of the chain.
-    LowestContiguousBlockHeight(u64),
+    /// The highest contiguous height range of fully available blocks.
```
Fraser999: I think we should add that the range is inclusive to the docs here, as that will carry over to the description provided in the JSON-RPC schema.

Fraser999 (approved):
Approved assuming the two nitpicks are addressed.

node/src/components/rpc_server/rpcs/common.rs
rafal-ch: Comment updated in e748d51648d38e1362794622919a322a76537882

node/src/effect/requests.rs
rafal-ch: Updated in e748d51648d38e1362794622919a322a76537882

node/src/components/rpc_server.rs
```diff
@@ -294,13 +294,16 @@ where
...
-                        main_responder: responder,
-                    })
+                    .get_highest_contiguous_block_range_from_storage()
+                    .event(
+                        move |(low, high)| Event::GetHighestContiguousBlockHeightRange {
+                            low,
+                            high,
+                            main_responder: responder,
+                        },
+                    )
```
goral09: Couldn't you simply map over the result and call `responder.respond((low, high))` ? It would require adding `and_then` on the `EffectBuilder` if I understand correctly.

node/src/components/rpc_server/rpcs/common.rs
```diff
@@ -77,35 +77,34 @@ pub(super) async fn run_query_and_encode<REv: ReactorEventT>(
...
+    let (low, high) = effect_builder
         .make_request(
-            |responder| RpcRequest::GetLowestContiguousBlockHeight { responder },
+            |responder| RpcRequest::GetHighestContiguousBlockHeightRange { responder },
             QueueKind::Api,
         )
         .await;
 
-    info!(lowest_contiguous_block_height, "{}", error_message);
+    info!(low, high, "{}", error_message);
```
goral09: I think it's worth adding some context here other than the `error_message` - otherwise, person reading the logs might get worried unnecessarily.

goral09 (approved):

node/src/components/rpc_server/rpcs/common.rs
rafal-ch: Message updated here: https://github.com/casper-network/casper-node/pull/2786/commits/d4ee34713e7034e8b37478e62597f8d5706f9b7d

node/src/components/rpc_server.rs
rafal-ch: Yes, it could actually work as follows: https://github.com/casper-network/casper-node/pull/2786/commits/bcbbb7b05a8cd3c581b65908ce22c4f0cc3ce959

Is it better now, @goral09 ?

goral09: Not the prettiest thing but achieves what I meant :)

PR #2788, opened 2022-03-24
Check block availability when responding to RPC requests

### LINKED ISSUE #2781: Change RPC server to not provide success responses for data outside its contiguous block range
The intent is that a request specifying a block identifier (height or hash), if that block's height is lower than the lowest contiguous block height as known by storage, then the server should respond with an ErrorData rather than success.

RPCs affected are:
* `chain_get_block`
* `chain_get_block_transfers`
* `chain_get_state_root_hash`
* `chain_get_era_info_by_switch_block`
* `state_get_auction_info`
* `state_get_account_info`
* `query_global_state` - only where using `GlobalStateIdentifier::BlockHash`

These RPCs are not affected:
* `state_get_item`
* `state_get_balance`
* `state_get_dictionary_item`

rafal-ch:
This PR changes the behavior of the RPC server in a way that it'll not provide success responses when asked for a data from blocks that are not a part of the highest contiguous block range known to the node.

The following endpoints are affected:
* `chain_get_block`
* `chain_get_block_transfers`
* `chain_get_state_root_hash`
* `chain_get_era_info_by_switch_block`
* `state_get_auction_info`
* `state_get_account_info`
* `query_global_state` - only when queried by using `GlobalStateIdentifier::BlockHash`

Closes https://github.com/casper-network/casper-node/issues/2781

### DIFF
--- a/node/src/components/rpc_server/rpcs/chain.rs
+++ b/node/src/components/rpc_server/rpcs/chain.rs
@@ -158,9 +158,18 @@ impl RpcWithOptionalParamsExt for GetBlock {
         api_version: ProtocolVersion,
     ) -> BoxFuture<'static, Result<Response<Body>, Error>> {
         async move {
+            // This RPC request is restricted by the block availability index.
+            let only_from_highest_contiguous_range = true;
+
             // Get the block.
             let maybe_block_id = maybe_params.map(|params| params.block_identifier);
-            let json_block = match get_block_with_metadata(maybe_block_id, effect_builder).await {
+            let json_block = match get_block_with_metadata(
+                maybe_block_id,
+                only_from_highest_contiguous_range,
+                effect_builder,
+            )
+            .await
+            {
                 Ok(BlockWithMetadata {
                     block,
                     finality_signatures,
@@ -244,9 +253,18 @@ impl RpcWithOptionalParamsExt for GetBlockTransfers {
         api_version: ProtocolVersion,
     ) -> BoxFuture<'static, Result<Response<Body>, Error>> {
         async move {
+            // This RPC request is restricted by the block availability index.
+            let only_from_highest_contiguous_range = true;
+
             // Get the block.
             let maybe_block_id = maybe_params.map(|params| params.block_identifier);
-            let block_hash = match common::get_block(maybe_block_id, effect_builder).await {
+            let block_hash = match common::get_block(
+                maybe_block_id,
+                only_from_highest_contiguous_range,
+                effect_builder,
+            )
+            .await
+            {
                 Ok(block) => *block.hash(),
                 Err(error) => return Ok(response_builder.error(error)?),
             };
@@ -317,9 +335,18 @@ impl RpcWithOptionalParamsExt for GetStateRootHash {
         api_version: ProtocolVersion,
     ) -> BoxFuture<'static, Result<Response<Body>, Error>> {
         async move {
+            // This RPC request is restricted by the block availability index.
+            let only_from_highest_contiguous_range = true;
+
             // Get the block.
             let maybe_block_id = maybe_params.map(|params| params.block_identifier);
-            let block = match common::get_block(maybe_block_id, effect_builder).await {
+            let block = match common::get_block(
+                maybe_block_id,
+                only_from_highest_contiguous_range,
+                effect_builder,
+            )
+            .await
+            {
                 Ok(block) => block,
                 Err(error) => return Ok(response_builder.error(error)?),
             };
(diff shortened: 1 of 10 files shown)


goral09:
>  it'll not provide success responses when asked for a data from blocks that are not a part of the highest contiguous block range known to the node.

Does it mean it will not give a successful response if I ask for a block that the node _knows of_ but isn't part of **the highest contiguous block range**?

For example, if the node has the following ranges : `{100, 120}, {133, 155}` and I ask about block 110, what will happen?

(disclaimer, I haven't read the code yet - the questions is based only on the PR description).

rafal-ch:
> For example, if the node has the following ranges : `{100, 120}, {133, 155}` and I ask about block 110, what will happen?

An `Err` will be reported, as per this comment (see row: 4)
https://github.com/casper-network/casper-node/issues/2781#issuecomment-1076413274

But, this PR might get invalidated by the following one (there will be no multiple ranges):
https://github.com/casper-network/casper-node/pull/2789

goral09:
> 



> > For example, if the node has the following ranges : `{100, 120}, {133, 155}` and I ask about block 110, what will happen?
> 
> An `Err` will be reported, as per this comment (see row: 4) [#2781 (comment)](https://github.com/casper-network/casper-node/issues/2781#issuecomment-1076413274)

That would be a surprising outcome to me. If we have the block, why not return it?

The ticket you linked says:

> if that block's height is lower than the lowest contiguous block height as known by storage, then the server should respond with an ErrorData

_lower than **lowest** contiguous block height known by storage_ - and that makes sense b/c the requested block is unknown.

Fraser999:
> That would be a surprising outcome to me. If we have the block, why not return it?

We had a discussion elsewhere and agreed to only provide data from the trusted block onwards (for fast-synced nodes).  It should avoid clients having to guess whether a certain node will be able to provide older blocks.

rafal-ch:
> > That would be a surprising outcome to me. If we have the block, why not return it?
> 
> We had a discussion elsewhere and agreed to only provide data from the trusted block onwards (for fast-synced nodes). It should avoid clients having to guess whether a certain node will be able to provide older blocks.

Ok, so this means that https://github.com/casper-network/casper-node/pull/2789 will make the lowest available block equal to the trusted block (for fast-sync), and this PR will make sure that even if we happen to have earlier blocks, we'll send an Err.

node/src/components/rpc_server/rpcs/chain.rs
```diff
@@ -244,9 +253,18 @@ impl RpcWithOptionalParamsExt for GetBlockTransfers {
...
             // Get the block.
             let maybe_block_id = maybe_params.map(|params| params.block_identifier);
-            let block_hash = match common::get_block(maybe_block_id, effect_builder).await {
+            let block_hash = match common::get_block(
+                maybe_block_id,
+                only_from_highest_contiguous_range,
+                effect_builder,
+            )
+            .await
+            {
```
goral09, line 267: It should be fairly trivial to collapse these two requests into one.

node/src/components/storage.rs
```diff
@@ -958,6 +964,11 @@ impl Storage {
                     } else {
                         return Ok(responder.respond(None).ignore());
                     };
+
+                if !self.should_return_block(block.height(), only_from_highest_contiguous_range) {
+                    return Ok(responder.respond(None).ignore());
+                }
```
goral09, line 970: We could do this even before we create a txn or call `get_single_block`. In the other place as well.

node/src/components/storage.rs
```diff
@@ -1428,15 +1484,25 @@ impl Storage {
             Some(block_header) => block_header,
             None => return Ok(None),
         };
+        self.validate_block_header_hash(&block_header, block_hash)?;
+        Ok(Some(block_header))
+    }
+
+    /// Validates the block header hash against the expected block hash.
+    fn validate_block_header_hash(
```
goral09, line 1492: I'm not sure we should be doing that at all. Question for another ticket. @Fraser999

goral09:
> > > That would be a surprising outcome to me. If we have the block, why not return it?
> > 
> > 
> > We had a discussion elsewhere and agreed to only provide data from the trusted block onwards (for fast-synced nodes). It should avoid clients having to guess whether a certain node will be able to provide older blocks.
> 
> Ok, so this means that #2789 will make the lowest available block equal to the trusted block (for fast-sync), and this PR will make sure that even if we happen to have earlier blocks, we'll send an Err.

I can see that you return `Ok(None)` if the check `should_return_block` doesn't pass. Where is it being mapped to an `Err`?

rafal-ch:
> I can see that you return `Ok(None)` if the check `should_return_block` doesn't pass. Where is it being mapped to an `Err`?

The mapping happens in the RPC server (inside the `QueryGlobalState::handle_request` for `query_global_state` and inside `get_block_with_metadata()` for others)

node/src/components/rpc_server/rpcs/chain.rs
rafal-ch: Yes, true. We might change the `GetBlockTransfers` to accept `BlockIdentifier` instead of `BlockHash`, but I'll prefer to do it in a separate PR, if really needed.

Or maybe you had some other idea in mind?

Fraser999: I'd leave for 2.0 if we do want to collapse these requests.

node/src/components/storage.rs
Fraser999: Although it's been split into a new method here, it does already exist in `dev`, so I'd be inclined to leave the PR unchanged here.

I'm not sure it's worth a specific new ticket as this behaviour will necessarily be reviewed when we make some of the upcoming planned changes to storage and data types in the near future.

Fraser999 (approved):

goral09 (approved):

Issue #2487, opened 2021-12-15
 Do sync-to-genesis in parallel with normal operation


piotr-dziubecki:
Downloading all blocks and their state back to genesis can take a long time. It should be possible to start a node and make it follow the network, while it downloads historical data in the background.

xcthulhu:
This duplicates  #2083

xcthulhu:
As I stated [previously](https://github.com/casper-network/casper-node/issues/2083#issuecomment-920514587).

One suggestion put forward by @goral09 is to launch a task in the participating reactor.

One way of doing this is to extract the following from the run_fast_sync_task, delete it there and launch it as an effect in the participating reactor:

```rust
    // Sync to genesis
    let mut walkback_block = *fetch_and_store_block_by_hash(effect_builder, trusted_block_header.hash()).await?;;
    loop {
        sync_deploys_and_transfers_and_state(effect_builder, &walkback_block).await?;
        if walkback_block.height() == 0 {
            break;
        } else {
            walkback_block = *fetch_and_store_block_by_hash(
                effect_builder,
                *walkback_block.header().parent_hash(),
            )
            .await?;
        }
    }
```

PR #2789, opened 2022-03-25
Rework lowest block height in storage

### LINKED ISSUE #2779: Change LowestContiguousBlockHeight in storage to represent the lowest block previously used as a trusted hash
Provide a new `StorageRequest::UpdateLowestContiguousBlockHeight(u64)` which is called from inside `run_chain_sync_task` immediately after `verify_trusted_block_header` is called.  In storage, this height is written to the DB if:
* it doesn't previously exist, or
* the new height is lower than the current one, or
* the new height is higher than the highest block stored as node startup.
 
This replaces the disjoint sequences in storage.

Fraser999:
This PR replaces the disjoint sequence tracking in storage with a persisted `lowest_contiguous_block_height` variable.  The value represents the lowest block from which the node has contiguous blocks (and corresponding global state) to the current chain tip.

It is updated after node startup when the chain synchronization has retrieved the trusted block.  In sync-to-genesis mode, the value is set to `0` and in fast-sync mode, it is set to the height of ~the trusted block~ [the most recent block](https://github.com/casper-network/casper-node/pull/2789/commits/e8486b14819a33f282c2fbc0075a42c33af45db9).

The value is only actually changed and written to the DB if:
* it doesn't previously exist, or
* the new height is lower than the current one, or
* the new height is higher than the highest block stored at node startup (before chain sync commenced)

A new `AvailableBlockRange` struct was added to avoid using `(u64, u64)` to represent the range.  It is more explicit in the code and in the JSON-RPC error response:
```json
"error": {
  "code": -32001,
  "message": "block at height 1314 not stored on this node",
  "data": {
    "available_block_range": {
      "low": 75,
      "high": 112
    }
  }
}
```

Closes #2779.


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -904,6 +904,17 @@ pub(super) async fn run_chain_sync_task(
...
+    let lowest_contiguous_block_height = if config.sync_to_genesis() {
+        // If backfilling to genesis is moved to a background task, we should change this to
+        // `trusted_block_header.height()` and update storage as we backfill.
+        0
+    } else {
+        trusted_block_header.height()
+    };
+    effect_builder
+        .update_lowest_contiguous_block_height_in_storage(lowest_contiguous_block_height)
+        .await;
```
rafal-ch: In the case of `sync_to_genesis()` we'll persist `0` here even before we start syncing. Now, if the sync to genesis takes, let's say, a few hours, we'll pretend that we have all blocks, but we're actually still fetching them.

This could be misleading unless `joiner` doesn't expose the RPC API to be queried.

node/src/components/rpc_server/rpcs/common.rs
```diff
@@ -78,13 +78,7 @@ pub(super) async fn run_query_and_encode<REv: ReactorEventT>(
...
 pub enum ErrorData {
     /// The highest contiguous height range (inclusive) of fully available blocks.
-    HighestContiguousBlockHeightRange((u64, u64)),
-}
-
-impl ErrorData {
-    pub(super) fn highest_contiguous_block_height_range(low: u64, high: u64) -> Self {
-        ErrorData::HighestContiguousBlockHeightRange((low, high))
-    }
+    HighestContiguousBlockHeightRange(ContiguousBlockRange),
```
rafal-ch: Since we now have a single range, it no longer needs to be called `Highest` (also in the comment).

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -904,6 +904,17 @@ pub(super) async fn run_chain_sync_task(
 
     verify_trusted_block_header(&chain_sync_context)?;
 
+    let lowest_contiguous_block_height = if config.sync_to_genesis() {
+        // If backfilling to genesis is moved to a background task, we should change this to
+        // `trusted_block_header.height()` and update storage as we backfill.
+        0
+    } else {
+        trusted_block_header.height()
```
rafal-ch: For the actual `fast-sync` we assume that the lowest contiguous block is `trusted_block_header.height()`, but the process downloads earlier blocks (for replay protection and era supervisor initialization).

Probably because of this it happens that the joiner node when asked for block 2000 responds with:
```json
"message": "block at height 2000 not stored on this node",
"data": {
	"highest_contiguous_block_height_range": {
		"low": 114,
		"high": 283
	}
}
```

But if I ask for block `100` I am still able to retrieve it.

node/src/components/chain_synchronizer/operations.rs
Fraser999: Yes, the joiner has no RPC server, so not a problem.

node/src/components/rpc_server/rpcs/common.rs
Fraser999: As discussed, I'll rename to `AvailableBlockRange`.

node/src/components/chain_synchronizer/operations.rs
Fraser999: Yes, your PR will change that behaviour to not return block 100.

node/src/components/rpc_server/rpcs/common.rs
Fraser999: Updated in [b834b44](https://github.com/casper-network/casper-node/pull/2789/commits/b834b4484c1125de754a4aac4315b64aca9024d4).

rafal-ch (approved):
LGTM :+1:

node/CHANGELOG.md
```diff
@@ -36,7 +36,7 @@ All notable changes to this project will be documented in this file.  The format
 * Connection handshake timeouts can now be configured via the `handshake_timeout` variable (they were hardcoded at 20 seconds before).
 * `Key::SystemContractRegistry` is now readable and can be queried via the RPC.
 * Requests for data from a peer are now de-prioritized over networking messages necessary for consensus and chain advancement.
-* JSON-RPC responses which fail to provide requested data will now also include an indication of that node's highest contiguous block height range, i.e. the block heights for which it holds all global state
+* JSON-RPC responses which fail to provide requested data will now also include an indication of that node's available block range, i.e. the block heights for which it holds all global state
```
goral09: This does not mention how the contiguous sequence is calculated.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -904,6 +904,17 @@ pub(super) async fn run_chain_sync_task(
 
     verify_trusted_block_header(&chain_sync_context)?;
 
+    let lowest_available_block_height = if config.sync_to_genesis() {
+        // If backfilling to genesis is moved to a background task, we should change this to
+        // `trusted_block_header.height()` and update storage as we backfill.
+        0
```
goral09: Do we need a ticket for this? Backfilling TestNet took 5.5 days.

node/CHANGELOG.md
Fraser999: Addressed in [e8486b1](https://github.com/casper-network/casper-node/pull/2789/commits/e8486b14819a33f282c2fbc0075a42c33af45db9).

node/CHANGELOG.md
```diff
@@ -36,7 +36,7 @@ All notable changes to this project will be documented in this file.  The format
 * Connection handshake timeouts can now be configured via the `handshake_timeout` variable (they were hardcoded at 20 seconds before).
 * `Key::SystemContractRegistry` is now readable and can be queried via the RPC.
 * Requests for data from a peer are now de-prioritized over networking messages necessary for consensus and chain advancement.
-* JSON-RPC responses which fail to provide requested data will now also include an indication of that node's highest contiguous block height range, i.e. the block heights for which it holds all global state
+* JSON-RPC responses which fail to provide requested data will now also include an indication of that node's available block range, i.e. the block heights for which it holds all global state.  For nodes running with `[node.sync_to_genesis]` set to true, the range will be the full blockchain, otherwise the range will start at a block near the tip of the chain when the node started running.
```
```suggestion
* JSON-RPC responses which fail to provide requested data will now also include an indication of that node's available block range, i.e. the block heights for which it holds all global state. For nodes running with `[node.sync_to_genesis]` set to true, the range will be the full blockchain, otherwise the range will start at a block near the tip of the chain when the node started running.
```

😎

node/src/components/storage.rs
```diff
@@ -1796,6 +1757,71 @@ impl Storage {
...
+        let high = self
+            .block_height_index
+            .keys()
+            .last()
+            .copied()
+            .unwrap_or_default();
+        let low = self.lowest_available_block_height;
+        match AvailableBlockRange::new(low, high) {
+            Ok(range) => range,
+            Err(error) => {
```
goral09, line 1834: The only possible error here is when `high` is actually lower than `low`. Maybe it makes sense here to flip them? It seems to be harmless and maybe the problem would "work itself out" until the next query (if the index and/or `lowest_available_block_height` get updated).

node/src/components/storage.rs
```diff
@@ -1796,6 +1757,71 @@ impl Storage {
...
+    }
+
+    fn update_lowest_available_block_height(
+        &mut self,
+        new_height: u64,
+    ) -> Result<(), FatalStorageError> {
+        // We should update the value if
+        // * it wasn't already stored, or
+        // * the new value is lower than the available range at startup, or
+        // * the new value is 2 or more higher than (after pruning due to hard-reset) the available
```
goral09, line 1848: nitpick: maybe it's worth including an explanation why is that a problem since it may not be obvious to the reader.  I.e. just write that there would be a single-block gap in the sequence where we might not have the Global State.

goral09 (approved):
Minor comments but no blockers.

node/src/components/chain_synchronizer/operations.rs
Fraser999: It's already been created: #2487.

node/CHANGELOG.md
Fraser999: I have conflicts to sort out in the changelog which I plan to do once #2788 also merges.  I'll fix this when I do that :)

node/src/components/storage.rs
Fraser999: I think I'd rather leave as is.  It's really pointing to a bug in our core logic if that actually arises (but not something worth stopping the node over).  So the default range `[MAX, MAX]` being returned to the client is probably a better indication that something is very wrong server side than asking for e.g. block 150 and getting back an error saying the server's available blocks are 100 -> 200.

Fraser999:
bors merge

PR #2739, opened 2022-03-14
RocksDB as global state store

marc-casperlabs:
Closes #2623

This PR supersedes https://github.com/casper-network/casper-node/pull/2680, which it is based upon. Please see the linked PR for a description.

Additional changes on this PR are from https://github.com/dwerner/casper-node/pull/19

PR #2797, opened 2022-03-30
Parallel storage access

### LINKED ISSUE #2757: Port storage changes from rocksdb PR to dev

marc-casperlabs:
Closes #2757 

This implemented as parallel storage access scheme similar to the one found in #2739. Core differences (although it is a from scratch implementation) are:

* Storage is split into two parts, `Storage` and `StorageInner`, the latter wrapped in an `Arc`. This makes cloning storage as small as possible and goes some ways towards proposed 2.0 storage changes.
* Actual work is performed on a thread earmarked for blocking tasks via `spawn_blocking` to prevent accidental blocks of the core event loop.
* To prevent an ~infinite~ large (=512) amount of blocking tasks being spawned, a semaphore guarding access has been added (see configuration option `max_sync_tasks` which defaults to 32).
* Effects returned by storage are now executed (now necessary since fast-sync introduced storage returning actual effects due to its handling of network-based requests).
* Code has been updated to work on the latest `dev` (previous versions did not include fast sync).


node/src/components/storage.rs
```diff
@@ -1103,8 +1080,10 @@ impl Storage {
                 block_height,
                 responder,
             } => {
+                let indices = self.indices.try_read()?;
```
dwerner: I don't think we should be using `try_read` and `try_write` - they mean that if we can't lock the mutex, we fail the request entirely. Once performance starts to become degraded, then most requests to storage will fail as someone is holding the lock at all times. This applies doubly once we start passing `Storage` around to callsites and use it directly, but that's admittedly a future concern.

I think we should either call `read` and `write` which are blocking variants, or make use of `tokio::sync::RwLock` if you are concerned about blocking the threadpool, as it's `read` and `write` methods will apply back-pressure. I do see we are executing in the context of a `spawn_blocking` closure, so I would recommend simply calling `read` and `write` instead.

Same goes for all other instances here in storage where we call `RwLock::try_*`.

dwerner (changes requested):
Mostly looks great :+1: , but I'm drawing an issue with the way we will throw an error if we encounter a locked mutex.

node/src/components/storage.rs
dwerner: Also if I am correct, an error in storage also means that we `fatal!`, correct?

node/src/utils.rs
```diff
@@ -76,6 +76,31 @@ impl Display for ResolveAddressErrorKind {
     }
 }
 
+/// Backport of `Result::flatten`, see https://github.com/rust-lang/rust/issues/70142.
```
dwerner: The link here is failing CI.

resources/local/config.toml
```diff
@@ -310,6 +310,11 @@ enable_mem_deduplication = true
 # For example, setting this value to 5 means that every 5th time something is put in the pool the cache is swept.
 mem_pool_prune_interval = 4096
 
+# Parallel storage request execution.
+#
+# Controls the maximum number of threads that are spawned to access storage in parallel.
+max_sync_tasks = 32
```
dwerner, line 318: Why 32? Did you do some load testing? `spawn_blocking` will already queue work once you hit the configured limit, so do you need this semaphore at all?

https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

> Tokio will spawn more blocking threads when they are requested through this function until the upper limit configured on the [Builder](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html) is reached. After reaching the upper limit, the tasks are put in a queue.

node/src/components/storage.rs
mbr: Oof, that's a big mess up!

You are right, this is of course wrong. I misread the docs and missed the section where it was non-blocking, I assumed it was solely preventing panics when attempting to lock the same lock twice.

mbr: > Also if I am correct, an error in storage also means that we fatal!, correct?

Yes, for now `FatalStorageError` is defined as an error that is fatal (which has been the previous behavior). I think a panic in a background task constitutes a reason that is as good as the other ones for failing.

resources/local/config.toml
mbr: 32 is just a guess, I assume it can be tweaked using load testing. The `spawn_blocking` maximum number of threads is fairly high, and it also shared across all blocking tasks, not just storage IO ones.

The default limit for `spawn_blocking` according to the docs is 512, which seems large for IO. If we were to bring this down to something in the realm of 32, the contention by CPU-bound tasks could be noticeable.

Looking at the (maybe somewhat dated) [LMDB benchmarks](http://www.lmdb.tech/bench/ondisk/#sec12), assuming 2000 bytes (which seems to be a reasonable size) VM reads, it seems that the sweet spot is somewhere between 32 and 64, with returns diminishing fast after 8:

LMDB 2000 bytes VM parallel reads

node/src/components/storage.rs
marc-casperlabs: Oof, that's a big mess up!

You are right, this is of course wrong. I misread the docs and missed the section where it was non-blocking, I assumed it was solely preventing panics when attempting to lock the same lock twice on the same thread. Turns out I read the wrong section =/

marc-casperlabs: > Also if I am correct, an error in storage also means that we fatal!, correct?

Yes, for now `FatalStorageError` is defined as an error that is fatal (which has been the previous behavior). I think a panic in a background task constitutes a reason that is as good as the other ones for failing.

node/src/utils.rs
marc-casperlabs: That one was new to me. Fixed!

resources/local/config.toml
marc-casperlabs: 32 is just a guess, I assume it can be tweaked using load testing. The `spawn_blocking` maximum number of threads is fairly high, and it also shared across all blocking tasks, not just storage IO ones.

The default limit for `spawn_blocking` according to the docs is 512, which seems large for IO. If we were to bring this down to something in the realm of 32, the contention by CPU-bound tasks could be noticeable.

Looking at the (maybe somewhat dated) [LMDB benchmarks](http://www.lmdb.tech/bench/ondisk/#sec12), assuming 2000 bytes (which seems to be a reasonable size) VM reads, it seems that the sweet spot is somewhere between 32 and 64, with returns diminishing fast after 8:

LMDB 2000 bytes VM parallel reads

node/src/components/storage.rs
marc-casperlabs: This now fixed in https://github.com/casper-network/casper-node/pull/2797/commits/b1829c6b86de4ee6564f8b5a4abeed52f20098ae.

Closing this conversation, @dwerner feel free to reopen it if you find the fix insufficient!

marc-casperlabs:
> I'm drawing an issue with the way we will throw an error if we encounter a locked mutex.

That was a really good catch, thanks!

node/src/components/storage.rs
```diff
@@ -317,9 +317,29 @@ enum GetRequestError {
...
+    storage: StorageInner,
+}
+
+impl Storage {
+    pub(crate) fn read_switch_block_header_by_era_id(
+        &self,
+        switch_block_era_id: EraId,
+    ) -> Result<Option<BlockHeader>, FatalStorageError> {
+        self.storage
+            .read_switch_block_header_by_era_id(switch_block_era_id)
```
goral09: Why is this the only method that gets exposed on `Storage`?

node/src/components/storage.rs
```diff
@@ -2042,6 +2013,52 @@ impl Storage {
...
+            })
+            .transpose()
+    }
+
+    /// Retrieves a single block by height by looking it up in the index and returning it.
+    fn get_block_by_height(&self, height: u64) -> Result<Option<Block>, FatalStorageError> {
+        let mut tx = self.storage.env.begin_ro_txn()?;
+
+        self.storage.get_block_by_height(&mut tx, height)
+    }
```
goral09, line 2311: It's a bit weird that these two methods are so simliar yet so different:
1. First one is called _read_ block header, while the second one _get_ block
2. The first one is `pub`, the second is private.
3. The first one has to implement the logic itself, the second one delegates to `self.storage` 
:)

I'm surprised you didn't make it consistent.

node/src/components/storage/tests.rs
```diff
@@ -1859,28 +1878,31 @@ fn should_restrict_returned_blocks() {
...
-    assert!(storage.should_return_block(4, true));
-    assert!(storage.should_return_block(5, true));
-    assert!(!storage.should_return_block(6, true));
+    assert!(!storage.storage.should_return_block(0, true));
+    assert!(!storage.storage.should_return_block(1, true));
+    assert!(!storage.storage.should_return_block(2, true));
+    assert!(!storage.storage.should_return_block(3, true));
+    assert!(storage.storage.should_return_block(4, true));
+    assert!(storage.storage.should_return_block(5, true));
+    assert!(!storage.storage.should_return_block(6, true));
```
goral09: I'm gonna pretend I didn't see `storage.storage` ;) 🤐

node/src/components/storage.rs
```diff
@@ -358,23 +191,18 @@ pub struct Storage {
...
+    /// Various indices used by the component.
+    #[data_size(skip)]
+    indices: RwLock<Indices>,
     /// Whether or not memory deduplication is enabled.
     enable_mem_deduplication: bool,
     /// An in-memory pool of already loaded serialized items.
     ///
     /// Keyed by serialized item ID, contains the serialized item.
-    serialized_item_pool: ObjectPool<Box<[u8]>>,
+    #[data_size(skip)]
```
goral09, line 209: Why are we skipping this now? Isn't it true that it may contribute to majority of the memory this component uses?

node/src/components/storage.rs
```diff
@@ -358,23 +191,18 @@ pub struct Storage {
...
-    /// blocks stored (and the corresponding global state).
-    lowest_available_block_height: u64,
     /// Highest block available when storage is constructed.
     highest_block_at_startup: u64,
-    /// A map of era ID to switch block ID.
-    switch_block_era_id_index: BTreeMap<EraId, BlockHash>,
-    /// A map of deploy hashes to hashes of blocks containing them.
-    deploy_hash_index: BTreeMap<DeployHash, BlockHash>,
+    /// Various indices used by the component.
+    #[data_size(skip)]
```
goral09, line 201: There's a `DataSize` derived on `Indices` but here we're just skipping it, why?

node/src/components/storage.rs
```diff
@@ -470,7 +466,9 @@ impl StorageInner {
                 // Disable read-ahead. Our data is not storead/read in sequence that would benefit from the read-ahead.
                 | EnvironmentFlags::NO_READAHEAD,
             )
-            .set_max_readers(MAX_TRANSACTIONS)
+            // We need at least `max_sync_tasks` readers, add an additional 8 for unforseen external
+            // reads (not likely, but it does not hurt to increase this limit).
+            .set_max_readers(config.max_sync_tasks as u32 + 8)
```
dwerner, line 476: :+1:

dwerner (approved):
Resolving my block on this now, looks good to me.

resources/local/config.toml
dwerner: I'm fine with the separate semaphore, as it does provide finer-grained control over back pressure here.

node/src/components/storage/tests.rs
```diff
@@ -1854,33 +1925,78 @@ fn should_restrict_returned_blocks() {
...
-    let mut storage = storage_fixture(&harness, verifiable_chunked_hash_activation);
+    let storage = storage_fixture(&harness, verifiable_chunked_hash_activation);
 
     // Create the following disjoint sequences: 1-2 4-5
     [1, 2, 4, 5].iter().for_each(|height| {
         let (block, _) = random_block_at_height(&mut harness.rng, *height, Block::random_v1);
-        storage.write_block(&block).unwrap();
+        storage.storage.write_block(&block).unwrap();
     });
     // The available range is 4-5.
```
rafal-ch, line 1946: I know this is not related to this PR, but I believe this comment should now say:
```
The lowest available block is 4.
```

node/src/components/storage/tests.rs
```diff
@@ -1854,33 +1925,78 @@ fn should_restrict_returned_blocks() {
...
+        .storage
+        .should_return_block(5, false)
+        .expect("should return block failed"));
+    assert!(storage
+        .storage
+        .should_return_block(6, false)
+        .expect("should return block failed"));
 
     // With restriction, the node should attempt to return only the blocks that are
     // on the highest disjoint sequence, i.e blocks 4 and 5 only.
```
rafal-ch, line 1984: Also here, the comment should say:
```
// With restriction, the node should attempt to return only the blocks of height
// greater than or equal to the lowest available block and less than or equal to
// the block at the tip, i.e blocks 4 and 5 only.
```

rafal-ch (commented):
No additional major issues found, except the ones pinpointed already.
I noticed a few rotten comments, though - nothing important.

node/src/components/storage.rs
marc-casperlabs: Any method that is exposed on `Storage` is there because external tooling (think `utils`) needs it, or some other functionality access the component directly (e.g. in reactor creation).

node/src/components/storage/tests.rs
marc-casperlabs: Better naming suggestions are appreciated!

node/src/components/storage.rs
marc-casperlabs: ~Good catch.~

Actually we do not implement `DataSize` for locks, as we do not want to lock just for calculating the memory footprint. We'd have to make a decision about whether to do this or not.

node/src/components/storage.rs
marc-casperlabs: It did the minimum of changes to implement the parallel feature and kept all inconsistencies and such "intact". This was to minimize potential issues when/if this has to be ported to other branches.

I expect a more thorough cleanup of storage to come later with the 2.0 push.

node/src/components/storage.rs
marc-casperlabs: `RWLock`, see above.

marc-casperlabs:
> I noticed a few rotten comments, though - nothing important.

I'm not changing these for now, to keep the changeset as small as possible. A different PR is probably a better fit to sneak this in :)

node/src/components/storage.rs
goral09: I thought so as well. Can you add comments then?

PR #2798, opened 2022-03-30
Initial support for chain operation modes

mpapierski:
Closes CasperLabs/product#54

This PR adds initial support for the private chain operation mode.


goral09:
@mpapierski can you provide more context in the PR description? What's the request/goal here? What we're trying to achieve, what kind of feature are we adding? How are we going about it? I.e. what got changes/added/removed to support it?

execution_engine/src/core/engine_state/engine_config.rs
```diff
@@ -72,4 +57,99 @@ impl EngineConfig {
...
+/// Any field that isn't specified will be defaulted.
+#[derive(Default, Debug)]
+pub struct EngineConfigBuilder {
+    max_query_depth: Option<u64>,
+    max_associated_keys: Option<u32>,
+    max_runtime_call_stack_height: Option<u32>,
+    wasm_config: Option<WasmConfig>,
+    system_config: Option<SystemConfig>,
+    chain_kind: Option<ChainKind>,
+}
```
goral09, line 80: Why not add the new field to the `new` constructor? I don't think that using the defaults for struct like this is a good idea, seems like an easy way to make mistakes.

execution_engine/src/shared/chain_kind.rs
```diff
@@ -0,0 +1,72 @@
...
+use casper_types::bytesrepr::{Error, FromBytes, ToBytes, U8_SERIALIZED_LENGTH};
+
+/// Flag representing a mode of operation of a chain.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize, DataSize, FromPrimitive)]
+#[serde(rename_all = "lowercase")]
+pub enum ChainKind {
+    /// Public (default) mode
+    Public,
+    /// Private chain mode with extra capabilities only available if this mode is set.
+    Private,
```
goral09, line 17: Each private chain could choose a different set of "special features". I don't think we should generalize and assume that all of them will want the same thing.

mpapierski:
Closed in favor of #2883

PR #2807, opened 2022-04-01
Update EE-test support to use chainspec values

### LINKED ISSUE #2313: CasperEngine test support crate needs to be updated to support the use of chainspec files

darthsiroftardis:
CHANGELOG:

- Added a public `fn new_from_chainspec` to the WasmTestBuilder that instatiates a builder using a given chainspec.
- Added `PRODUCTION_GENESIS_REQUEST` which uses values present in the production chainspec.toml
- Deprecated `DEFAULT_GENESIS_REQUEST` in favor of `PRODUCTION_GENESIS_REQUEST`

Closes #2313


execution_engine_testing/test_support/CHANGELOG.md
```diff
@@ -15,11 +15,13 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Added
 * Added `WasmTestBuilder::get_execution_journals` method for returning execution journals for all test runs.
+* Added support to load values from a given Chainspec.
+* Added static and constants that represent Casper-mainnet chainspec values
```
goral09: It's worth noting that chainspec might change with new protocol versions. We should include that in this changelog entry and specify which version of the chainspec it adds.

execution_engine_testing/test_support/src/chainspec_config.rs
```diff
@@ -0,0 +1,190 @@
...
+            chainspec_config.system_costs_config,
+            chainspec_config.core_config.validator_slots,
+            chainspec_config.core_config.auction_delay,
+            locked_funds_period_millis,
+            chainspec_config.core_config.round_seigniorage_rate,
+            chainspec_config.core_config.unbonding_delay,
+            DEFAULT_GENESIS_TIMESTAMP_MILLIS,
+        ))
+    }
+}
```
goral09, line 164: These types are part of the `node` crate and have been copied over. I'm wondering whether it wouldn't make sense to move them now to `casper-types`. Both `execution-engine-testing` and `node` depend on that crate (`casper-types`) so the code would be available to both. There's another advantage - since this approach (of allowing people to use a chainspec for their tests) requires this code to be in sync with latest production chainspec, it does make sense to share the type definitions. Otherwise, users will have runtime errors if the code here doesn't get updated with the chainspec. Sharing it with node will make sure that it never happens. And there won't be unnecessary code duplication.

execution_engine_testing/test_support/src/lib.rs
```diff
@@ -138,20 +141,39 @@ pub static DEFAULT_GENESIS_CONFIG: Lazy<GenesisConfig> = Lazy::new(|| {
...
         *DEFAULT_PROTOCOL_VERSION,
+        #[allow(deprecated)]
         DEFAULT_EXEC_CONFIG.clone(),
     )
 });
 /// Default [`ChainspecRegistry`].
 pub static DEFAULT_CHAINSPEC_REGISTRY: Lazy<ChainspecRegistry> =
     Lazy::new(|| ChainspecRegistry::new_with_genesis(&[1, 2, 3], &[4, 5, 6]));
+// This static constant has been deprecated in favor of the Production counterpart
+// which uses costs tables and values which reflect values used by the Casper main-net.
```
goral09: Isn't it _Mainnet_ ?

goral09 (commented):
We need to bump the minor version of `casper-engine-test-support` - minor b/c we're changing public API in a non-backwards compatible way (constructor now takes optional `post_state_hash`).

goral09 (commented):
This PR would have been much much smaller if we changed the implementation of `Default for InMemoryWasmTestBuilder` to use the new constructor that uses production chainspec. Similarly, `DEFAULT_RUN_GENESIS_REQUEST` would become `PRODUCTION_RUN_GENESIS_REQUEST`. 

I think it's OK to change these, even though they're breaking, b/c we're redefining what the _default_ is. And we're breaking the API already (see my previous https://github.com/casper-network/casper-node/pull/2807#pullrequestreview-930201175).

execution_engine_testing/tests/src/test/regression/gh_1902.rs
```diff
@@ -90,7 +84,13 @@ fn should_not_charge_for_create_purse_in_first_time_bond() {
 
     let bond_amount = U512::from(BOND_AMOUNT);
     // This amount should be enough to make first time add_bid call.
-    let add_bid_payment_amount = U512::from(DEFAULT_ADD_BID_COST);
+    let add_bid_cost = builder
+        .get_engine_state()
+        .config()
+        .system_config()
+        .auction_costs()
```
goral09: We could add `builder.get_auction_costs()` since it seems like a useful method to have.

goral09 (commented):
Reading the code more, maybe we should modify the `WasmTestBuilder` to include the whole `Chainspec` in it, since values like `validator_slots`, `auction_delay`, `unbonding_delay`, etc. are necessary for some actions made during testing (like `RUN_GENESIS_REQUEST`) and then make the function `create_genesis_request_from_production_chainspec` a method on the `WasmTestBuilder` - `builder.create_genesis_request(genesis_accounts)`.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -727,13 +728,14 @@ fn should_get_first_seigniorage_recipients() {
     );
 
     let mut era_validators: EraValidators = builder.get_era_validators();
-    let snapshot_size = DEFAULT_AUCTION_DELAY as usize + 1;
+    let auction_delay = builder.get_auction_delay();
```
goral09, line 727: We could redefine what `DEFAULT_AUCTION_DELAY` is by forwarding this to the lazily instantiated production config/chainspec.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -243,46 +244,90 @@ fn should_distribute_delegation_rate_zero() {
...
+    assert_eq!(total_payout, expected_total_reward_integer);
 
     for request in post_genesis_requests {
         builder.exec(request).commit().expect_success();
     }
 
-    for _ in 0..5 {
-        builder.run_auction(timestamp_millis, Vec::new());
-        timestamp_millis += TIMESTAMP_MILLIS_INCREMENT;
+    for _ in 0..=builder.get_auction_delay() {
```
goral09, line 264: This should have changed the outcome of the test. The production `auction_delay` is 1 while the test used 5. Did it?

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -2220,20 +2455,116 @@ fn should_distribute_by_factor() {
...
+    //     era_info.select(VALIDATOR_2.clone()).next(),
+    //     Some(SeigniorageAllocation::Validator { validator_public_key, amount })
+    //     if *validator_public_key == *VALIDATOR_2 && *amount == expected_validator_2_payout
+    // ));
+    //
+    // assert!(matches!(
+    //     era_info.select(VALIDATOR_3.clone()).next(),
+    //     Some(SeigniorageAllocation::Validator { validator_public_key, amount })
+    //     if *validator_public_key == *VALIDATOR_3 && *amount == expected_validator_3_payout
+    // ));
```
goral09: A lot of commented out code here 👀

execution_engine_testing/test_support/src/chainspec_config.rs
Fraser999: Yes, I agree that all the chainspec types should be moved to the `casper-types` crate, but that should be a separate PR.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
darthsiroftardis: It did indeed change the outcome, however, I refactored the test to account for the change.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
darthsiroftardis: We could do that, however, I prefer using this method as it gives the delay currently being used by the EE itself which I think is a better than having a static that was instantiated once and never gets 'refreshed' or updated.

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
darthsiroftardis: Removed the commented code

execution_engine_testing/tests/src/test/regression/gh_1902.rs
darthsiroftardis: Added the suggested method

execution_engine_testing/test_support/src/lib.rs
darthsiroftardis: Applied correction

execution_engine_testing/test_support/CHANGELOG.md
darthsiroftardis: Applied suggestion

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
goral09: It is changing the test though - it used to use the _default_ value and not it uses the one from latest chainspec. Which may influence the result of the test. Do we want that?

darthsiroftardis: Yes, because we want to assert that the functionality of the EE is consistent with values used in production networks

darthsiroftardis:
@goral09 About the concerns about the breaking change, I believe we are okay with that as we would just have these changes be presented as the next major release. @Fraser999 am I speaking incorrectly here

execution_engine_testing/test_support/CHANGELOG.md
```diff
@@ -15,12 +15,14 @@ All notable changes to this project will be documented in this file.  The format
...
 * Added `WasmTestBuilder::get_execution_journals` method for returning execution journals for all test runs.
+* Added support to load values from a given Chainspec.
+* Added static and constants that represent Casper-mainnet chainspec values. These values will change as new ProtocolVersions are added. The current values reflect ones used in the 1.5.0 ProtocolVersion.
 * Added `WasmTestBuilder::advance_era`, `WasmTestBuilder::advance_eras_by`, and `WasmTestBuilder::advance_eras_by_default_auction_delay` to advance chain and run auction contract in test environment.
 
 ### Changed
 * `WasmTestBuilder::get_transforms` is deprecated in favor of `WasmTestBuilder::get_execution_journals`.
 * `deploy_hash` field is now defaulted to a random value rather than zeros in `DeployItemBuilder`.
-
+* Deprecated the `DEFAULT_GENESIS_REQUEST` in favor of `PRODUCTION_GENESIS_REQUEST`.
```
Fraser999: There should be a separate section for deprecated items.

execution_engine_testing/test_support/src/chainspec_config.rs
```diff
@@ -0,0 +1,194 @@
...
+    FailedToLoadChainspec {
+        /// Path that failed to be read.
+        path: PathBuf,
+        /// The underlying OS error.
+        error: io::Error,
+    },
+    FailedToParseChainspec(toml::de::Error),
+    CouldNotCreateExecConfig,
+    CouldNotParseLockedFundsPeriod,
+    CouldNotCreateGenesisRequest,
```
Fraser999: `FailedTo` rather than `CouldNot` for consistency?

execution_engine_testing/test_support/src/chainspec_config.rs
```diff
@@ -0,0 +1,194 @@
...
+}
+
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
+// Disallow unknown fields to ensure config files and command-line overrides contain valid keys.
+#[serde(deny_unknown_fields)]
```
Fraser999: Do we actually want that here?  In fact, couldn't `CoreConfig` already lose the `era_duration` and `minimum_era_height` fields as they're unused I think?

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -174,6 +181,35 @@ impl InMemoryWasmTestBuilder {
...
+        );
+
+        let global_state = InMemoryGlobalState::empty().expect("should create global state");
+
+        Self::new(global_state, engine_config, post_state_hash)
+    }
+
+    /// Returns an [`InMemoryWasmTestBuilder`] instantiated
+    /// using values from the production chainspec.
+    pub fn new_with_production_chainspec() -> Self {
```
Fraser999: I agree with @goral09 point that this method should be removed and implemented as `Default` instead.

darthsiroftardis: Removed, and implemented as `Default`

execution_engine_testing/test_support/src/chainspec_config.rs
darthsiroftardis: Removed the unused fields and the serde tag

execution_engine_testing/test_support/src/chainspec_config.rs
darthsiroftardis: Renamed to maintain consistency, but introduced a clippy exception

execution_engine_testing/test_support/CHANGELOG.md
darthsiroftardis: Moved to a seperate section

Fraser999 (approved):

PR #2813, opened 2022-04-05
Send `Deploy` down the event stream.

goral09:
Instead of sending a DeployHash to the event stream component and then retrieving it for every subscriber, we now send whole Deploy to save reactor's churn and querying storage.

`DeployGetter` and associated code was not used anymore so it was removed.


marc-casperlabs (approved):
That looks pretty clean to me!

marc-casperlabs:
Almost forgot, maybe mention this in the CHANGELOG, so that node operators know the ramifications of the change?

Fraser999 (approved):

goral09:
Superseeded by https://github.com/casper-network/casper-node/pull/2818

PR #2818, opened 2022-04-05
Send `Deploy` down the event stream.

goral09:
Instead of sending a DeployHash to the event stream component and then retrieving it for every subscriber, we now send whole Deploy to save reactor's churn and querying storage.

`DeployGetter` and associated code was not used anymore so it was removed.Please consider the following when creating a PR:


Fraser999 (approved):

goral09:
bors merge

PR #2825, opened 2022-04-06
NCTL: always use casper-client-rs repo for casper-client

TomVasile:
Changes:
- removes pulling of casper-client from nctl bucket
- switches to always use casper-client-rs


TomVasile:
bors try

Fraser999 (approved):

PR #2827, opened 2022-04-06
always use casper-client-rs repo for casper-client

TomVasile:
sync with: https://github.com/casper-network/casper-node/pull/2825


TomVasile:
bors try

sacherjj (approved):

PR #2810, opened 2022-04-04
RocksDb Support -> feat-rocksdb

dwerner:

PR #2844, opened 2022-04-08
Avoid re-syncing locally available data

### LINKED ISSUE #2842: Avoid re-syncing locally available data
Avoid re-syncing locally available data

Fraser999:
This PR modifies the sync-to-genesis process to skip fetching blocks, deploys and global state which are already stored locally.  This allows for rapid re-joining in the case of a node which had previously synced to genesis, and then stopped running.

The determination is made by retrieving the available block range from storage.  All blocks (and corresponding deploys and global state) included in that range can safely be assumed to be held locally, as they were either retrieved previously during chain synchronization or executed locally.

The PR also includes an optimization of the `missing_trie_keys` function, ported from #2810.  It appears there is a slight overall benefit to sync times with this included, but it could use further benchmarking.

Finally, there are two error level log messages which have been downgraded to warnings as they appeared during testing when two peers were joining concurrently.

Closes #2842.

### DIFF
--- a/execution_engine/src/storage/trie_store/operations/mod.rs
+++ b/execution_engine/src/storage/trie_store/operations/mod.rs
@@ -8,6 +8,8 @@ use std::{
     mem,
 };
 
+use tracing::error;
+
 use casper_hashing::Digest;
 use casper_types::bytesrepr::{self, FromBytes, ToBytes};
 
@@ -252,16 +254,45 @@ where
         if !visited.insert(trie_key) {
             continue;
         }
-        let maybe_retrieved_trie: Option<Trie<K, V>> = store.get(txn, &trie_key)?;
-        match maybe_retrieved_trie {
-            // If we can't find the trie_key; it is missing and we'll return it
+        let maybe_retrieved_trie_bytes = store.get_raw(txn, &trie_key)?;
+        // Optimization: Don't look for descendants of leaves.
+        match maybe_retrieved_trie_bytes.and_then(|bytes| bytes.get(0)) {
+            // if this is a leaf don't parse (ie, the first byte is 0), just continue
+            Some(&Trie::<K, V>::LEAF_TAG) => continue,
+            // Either no entry under trie key or data is empty.  If data is empty then the trie is
+            // corrupted. Either way treat as missing.
             None => {
                 missing_descendants.push(trie_key);
+                continue;
             }
-            // If we could retrieve the node and it is a leaf, the search can move on
-            Some(Trie::Leaf { .. }) => (),
+            _ => (),
+        }
+
+        // Parse the trie, handling errors gracefully.
+        let retrieved_trie: Trie<K, V> = match maybe_retrieved_trie_bytes
+            .map(bytesrepr::deserialize_from_slice)
+            .transpose()
+        {
+            Ok(Some(retrieved_trie)) => retrieved_trie,
+            Ok(None) => {
+                missing_descendants.push(trie_key);
+                continue;
+            }
+            // Couldn't parse; treat as missing and continue
+            Err(err) => {
+                error!(?err, "unable to parse trie");
+                missing_descendants.push(trie_key);
+                continue;
+            }
+        };
+
+        match retrieved_trie {
+            // Should be unreachable due to checking the first byte as a shortcut above.
+            Trie::Leaf { .. } => {
+                error!("did not expect to see a trie leaf in `missing_trie_keys` after shortcut");
+            }
             // If we hit a pointer block, queue up all of the nodes it points to
-            Some(Trie::Node { pointer_block }) => {
+            Trie::Node { pointer_block } => {
                 for (_, pointer) in pointer_block.as_indexed_pointers() {
                     match pointer {
                         Pointer::LeafPointer(descendant_leaf_trie_key) => {
@@ -274,7 +305,7 @@ where
                 }
             }
             // If we hit an extension block, add its pointer to the queue
-            Some(Trie::Extension { pointer, .. }) => trie_keys_to_visit.push(pointer.into_hash()),
+            Trie::Extension { pointer, .. } => trie_keys_to_visit.push(pointer.into_hash()),
         }
     }
     Ok(missing_descendants)
--- a/execution_engine/src/storage/trie_store/operations/mod.rs
+++ b/execution_engine/src/storage/trie_store/operations/mod.rs
@@ -254,31 +254,25 @@ where
         if !visited.insert(trie_key) {
             continue;
         }
-        let maybe_retrieved_trie_bytes = store.get_raw(txn, &trie_key)?;
-        // Optimization: Don't look for descendants of leaves.
-        match maybe_retrieved_trie_bytes.and_then(|bytes| bytes.get(0)) {
-            // if this is a leaf don't parse (ie, the first byte is 0), just continue
-            Some(&Trie::<K, V>::LEAF_TAG) => continue,
-            // Either no entry under trie key or data is empty.  If data is empty then the trie is
-            // corrupted. Either way treat as missing.
+
+        let retrieved_trie_bytes = match store.get_raw(txn, &trie_key)? {
+            Some(bytes) => bytes,
             None => {
+                // No entry under this trie key.
                 missing_descendants.push(trie_key);
                 continue;
             }
-            _ => (),
+        };
+
+        // Optimization: Don't deserialize leaves as they have no descendants.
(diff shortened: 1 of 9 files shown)


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -894,8 +898,33 @@ async fn fetch_forward(
...
+                if lowest_available_block.height() == 0 {
+                    break;
+                }
+                walkback_block = *fetch_and_store_block_by_hash(
+                    *lowest_available_block.header().parent_hash(),
+                    ctx,
+                )
+                .await?;
+            }
+        }
```
goral09: Do we need to repeat that at the every step of the loop? Once we get the _lowest available block from storage_ its parent will be the next _lowest available block from storage_.

node/src/types/available_block_range.rs
```diff
@@ -45,6 +45,16 @@ impl AvailableBlockRange {
     pub fn contains(&self, height: u64) -> bool {
```
afck, line 46: (That comment in line 11 should probably say something about "error" and "low > high"…)

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -252,16 +254,45 @@ where
         if !visited.insert(trie_key) {
             continue;
         }
-        let maybe_retrieved_trie: Option<Trie<K, V>> = store.get(txn, &trie_key)?;
-        match maybe_retrieved_trie {
-            // If we can't find the trie_key; it is missing and we'll return it
+        let maybe_retrieved_trie_bytes = store.get_raw(txn, &trie_key)?;
+        // Optimization: Don't look for descendants of leaves.
```
afck:
```suggestion
        // Optimization: Don't deserialize leaves.
```

(At least I was confused by this comment at first…)

dwerner (approved):

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -252,16 +254,45 @@ where
...
+        {
+            Ok(Some(retrieved_trie)) => retrieved_trie,
+            Ok(None) => {
+                missing_descendants.push(trie_key);
+                continue;
+            }
+            // Couldn't parse; treat as missing and continue
+            Err(err) => {
+                error!(?err, "unable to parse trie");
+                missing_descendants.push(trie_key);
```
afck, line 278: I wonder whether we should delete the invalid data from the store now.
(On the other hand, maybe that's dangerous?)

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -252,16 +254,45 @@ where
...
+            _ => (),
+        }
+
+        // Parse the trie, handling errors gracefully.
+        let retrieved_trie: Trie<K, V> = match maybe_retrieved_trie_bytes
+            .map(bytesrepr::deserialize_from_slice)
+            .transpose()
+        {
+            Ok(Some(retrieved_trie)) => retrieved_trie,
+            Ok(None) => {
```
afck: That case is redundant now.
Maybe before `bytes.get(0)` we should unwrap or `continue`?

```rust
let retrieved_trie_bytes = match maybe_retrieved_trie_bytes {
    Some(bytes) => bytes,
    None => {
        missing_descendants.push(trie_key);
        continue;
    }
};
```

afck (approved):
Looks good, no blockers. :+1:

node/src/components/chain_synchronizer/operations.rs
Fraser999: We should only actually go inside the `if` block once, as the available range is taken before we enter the loop and it's not updated on each iteration.  Maybe I should rename the variable to `locally_available_block_range` to `available_block_range_on_startup` or something?

execution_engine/src/storage/trie_store/operations/mod.rs
Fraser999: I'd rather leave the logic as is, since with this approach, the error gets logged and we also try to recover by marking the invalid data as missing and hence we try to fetch it from a peer.

I'm not sure it'd be dangerous as such to delete the corrupted data, but it would be a slightly bigger change as we'd need to change the method to accept a rw transaction, and this is not a pattern we generally use when encountering parsing errors/corrupted data.

In the unoptimised code currently in `dev`, I believe we actually just exit early, returning an error in this case.

node/src/types/available_block_range.rs
Fraser999: Thanks - fixed in [9fcb768](https://github.com/casper-network/casper-node/pull/2844/commits/9fcb76807c46450853f6ca6d04e054031b4426e3).

execution_engine/src/storage/trie_store/operations/mod.rs
Fraser999: Fixed in [9fcb768](https://github.com/casper-network/casper-node/pull/2844/commits/9fcb76807c46450853f6ca6d04e054031b4426e3).

execution_engine/src/storage/trie_store/operations/mod.rs
Fraser999: I tweaked this logic in [9fcb768](https://github.com/casper-network/casper-node/pull/2844/commits/9fcb76807c46450853f6ca6d04e054031b4426e3) based on this suggestion.

We now don't do any special handling for the case where the returned bytes are empty, instead letting it fall through to the `Err` case when deserializing.  This should not change the net result, but will give an appropriate error log message in this case now too.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -254,31 +254,25 @@ where
...
             None => {
+                // No entry under this trie key.
                 missing_descendants.push(trie_key);
                 continue;
             }
-            _ => (),
+        };
+
+        // Optimization: Don't deserialize leaves as they have no descendants.
+        if let Some(&Trie::<K, V>::LEAF_TAG) = retrieved_trie_bytes.first() {
```
afck, line 268: (Now there's no need to match anymore.)

```suggestion
        if Some(&Trie::<K, V>::LEAF_TAG) == retrieved_trie_bytes.first() {
```

afck (approved):

execution_engine/src/storage/trie_store/operations/mod.rs
goral09: Or there the case when `Bytes` are empty could be handled together with non-existence of `trie_key` in the trie. Does it make a difference whether it's not in the trie or the bytes are empty? i.e.
```rust
None | Some(bytes) if bytes.is_empty() => {
               // No entry under this trie key.
                missing_descendants.push(trie_key);
                continue;
}
```
if this case wasn't handled explicitly, it would probably exit early in the `deserialize_from_slice` anyway - calling 
```rust
                error!(?err, "unable to parse trie");
                missing_descendants.push(trie_key);
                continue;
```

node/src/components/chain_synchronizer/operations.rs
goral09: If that's true then why not extract if outside of the loop? At least it will make it clear to the reader that it will be executed only once.

execution_engine/src/storage/trie_store/operations/mod.rs
afck: Not sure. I think I'd rather consider an existing but empty entry in the database to be the same as "can't deserialize", not "doesn't exist".

goral09: But "can't deserialize" and "doesn't exist" are treated the same way (with the exeception of logging an error). Just compare the `None` branch above and the `Err` case below.

afck: Ah, right, then your suggestion is fine.
(Not sure if you can have an `if` in a `match` arm that applies only to one of the options separated by `|`?)

goral09:
bors merge

execution_engine/src/storage/trie_store/operations/mod.rs
goral09: Since Fraser is on holidays this week, I will merge this PR now and put up a following one with the suggested changes.

node/src/components/chain_synchronizer/operations.rs
goral09: Since Fraser is on holidays this week, I will merge this PR now and put up a following one with the suggested changes.

goral09:
bors merge

goral09:
bors merge

TomVasile:
Looking into the build issues

PR #2847, opened 2022-04-09
Fix sporadic storage test failure

Fraser999:
This PR fixes a case where the storage test would fail due to trying to store two randomly-generated blocks which both happened to be switch blocks for the same era.  Storage disallows this, causing tests to fail when that happened.


marc-casperlabs (approved):
LGTM

PR #2856, opened 2022-04-12
Temporary experiment with missing descendants

rafal-ch:
Do not merge, this is just a temporary, hacky code.

### DIFF
--- a/execution_engine/src/storage/global_state/lmdb.rs
+++ b/execution_engine/src/storage/global_state/lmdb.rs
@@ -6,6 +6,7 @@ use std::{
 
 use casper_hashing::{ChunkWithProof, Digest};
 use casper_types::{bytesrepr::Bytes, Key, StoredValue};
+use tracing::info;
 
 use crate::{
     shared::{additive_map::AdditiveMap, newtypes::CorrelationId, transform::Transform},
@@ -294,6 +295,7 @@ impl StateProvider for LmdbGlobalState {
             .collect();
 
         if trie_keys_filtered.is_empty() {
+            info!("no need to call missing_trie_keys");
             Ok(vec![])
         } else {
             let txn = self.environment.create_read_txn()?;
@@ -305,11 +307,11 @@ impl StateProvider for LmdbGlobalState {
                     LmdbTrieStore,
                     Self::Error,
                 >(correlation_id, &txn, self.trie_store.deref(), trie_keys)?;
+            txn.commit()?;
             if missing_descendants.is_empty() {
                 self.digests_without_missing_descendants
                     .extend(&missing_descendants);
             }
-            txn.commit()?;
             Ok(missing_descendants)
         }
     }
--- a/execution_engine/src/storage/global_state/lmdb.rs
+++ b/execution_engine/src/storage/global_state/lmdb.rs
@@ -299,18 +299,21 @@ impl StateProvider for LmdbGlobalState {
             Ok(vec![])
         } else {
             let txn = self.environment.create_read_txn()?;
-            let missing_descendants =
-                missing_trie_keys::<
-                    Key,
-                    StoredValue,
-                    lmdb::RoTransaction,
-                    LmdbTrieStore,
-                    Self::Error,
-                >(correlation_id, &txn, self.trie_store.deref(), trie_keys)?;
+            let missing_descendants = missing_trie_keys::<
+                Key,
+                StoredValue,
+                lmdb::RoTransaction,
+                LmdbTrieStore,
+                Self::Error,
+            >(
+                correlation_id,
+                &txn,
+                self.trie_store.deref(),
+                trie_keys.clone(),
+            )?;
             txn.commit()?;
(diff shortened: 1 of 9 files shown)


execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -276,21 +282,36 @@ impl StateProvider for LmdbGlobalState {
...
+                    Key,
+                    StoredValue,
+                    lmdb::RoTransaction,
+                    LmdbTrieStore,
+                    Self::Error,
+                >(correlation_id, &txn, self.trie_store.deref(), trie_keys)?;
+            if missing_descendants.is_empty() {
+                self.digests_without_missing_descendants
+                    .extend(&missing_descendants);
+            }
```
goral09: nit: I would do it outside of the transaction to not hold it any longer than necessary.

node/src/components/contract_runtime.rs
```diff
@@ -395,39 +402,43 @@ impl ContractRuntime {
...
+                    //         effect_builder,
+                    //         "error flushing lmdb environment {:?}",
+                    //         lmdb_error
+                    //     )
+                    //     .await;
+                    // } else {
+                    //     metrics.put_trie.observe(start.elapsed().as_secs_f64());
+                    //     trace!(?result, "put_trie response");
+                    //     responder.respond(result).await
+                    // }
```
goral09: Why is this commented out?

node/src/components/contract_runtime.rs
```diff
@@ -176,7 +176,7 @@ impl Display for Event {
 #[derive(DataSize)]
 pub(crate) struct ContractRuntime {
     execution_pre_state: Arc<Mutex<ExecutionPreState>>,
-    engine_state: Arc<EngineState<LmdbGlobalState>>,
+    engine_state: Arc<Mutex<EngineState<LmdbGlobalState>>>,
```
goral09, line 179: I'm guessing that without `Mutex` it can't be made to be mutable and shared?

node/src/components/contract_runtime.rs
rafal-ch: For the sake of simplicity, there was an issue moving `engine state` into an async closure.
I though we could live w/o `flush_environment()` and w/o metrics, but now I see there's also the `responder.respond(result)` commented out.
Need to rethink this.

node/src/components/contract_runtime.rs
rafal-ch: Nope.
Maybe with the interior mutability, though.

Anyway, for the sake of a quick checking, I might consider reverting the introduction of `Mutex` around `EngineState` and use the container inside the `lazy_static` - naturally, just temporarily.

execution_engine/src/storage/global_state/lmdb.rs
marc-casperlabs: Hmm, am I a bit dense? "If A is empty, extend B with A" would just result in B (i.e. `self.digests_without_missing_descendants`) not changing?

rafal-ch: Yup, the missing `!`...

rafal-ch: Actually, the superfluous `if` :)

marc-casperlabs:
Am I understanding this correctly that we want to just memoize the `missing_trie_keys` function?

If that's the case, since the signature is essentially `fn (Vec<Digest>) -> Vec<Digest>`, with every element in the first set being independent, could we not just add a something like

```rust
missing_cache: lru::LruCache<Digest, Vec<Digest>>,
```

```rust
for key in &trie_keys {
    if let Some(found) = self.get(key) {
         // found, add to list
    } else {
         // update and insert into cache
    }
}
```

?

We got mutable access to somewhere we can store the cached data, and I think we can take the hit of caching 4096 entries (or more, as an experiment), if we feel like it. Alternatively we could still be selective in what we cache (only empty ones) and crank up the cache size.

rafal-ch:
@marc-casperlabs 
Yes, we can definitely do it in many better ways; let's first prove that it improves the performance and doesn't break things.

marc-casperlabs:
Alternative is to not cache, but keep an entire mapping, of all missing items we ever requested and a reverse mapping. So once we determine that `A` needs `x,y,z`, we store `A -> [x,y,z]` and `x -> [A], y -> [A], z -> [A]`. If `B` needs `x, q, p`, the table'd now look like

```
A -> [x,y,z]
B -> [x, q, p]

x -> [A,B]
y -> [A]
z -> [A]
p -> [B]
q -> [B]
```

Repeated queries can now be answered. If we receive, say `x`, we remove `x` from the second mapping and update the first

```
A -> [y, z]
B -> [q, p]

y -> [A]
z -> [A]
p -> [B]
q -> [B]
```

By limiting the number of requested (or outstanding nodes) we curb memory usage if necessary.

rafal-ch:
Closed in favor of https://github.com/casper-network/casper-node/pull/2857

PR #2857, opened 2022-04-12
Improve speed of initial sync process.

rafal-ch:
1. Adds the cache used to reduce the time spent on the process of finding missing descendants
2. Changes the `chain_sync_trie_store_duration_seconds` metric to histogram
3. Adds `chain_sync_block_height_synced` gauge to record the height of most recently synchronized block. It can go up and down as syncing process goes back to Genesis and then fetches forward when catching up with the tip.
4. Replaces a panic (`unreachable!`) with an ERROR log: [link](https://github.com/casper-network/casper-node/pull/2857/files#diff-df869a6a66919971b2be52e605cbae62315238aea3be83b32e32d6c1d0422117R706).
5. Adds basic peer scoring system in the fast sync process. We will prioritize "good" nodes and once every `sync_peer_redemption_interval` (configurable; defaults to 10_000) queries give a chance to previously "bad" node.
6. Register `chain_sync_fetch_and_store_initial_trusted_block_header_duration_seconds` metric.
7. Ports the parallel storage access.
8. Changes how we calculate the `max_parallel_{trie|deploy}_requests` to use the currently connected peer count.

Closes https://github.com/casper-network/Governance/issues/285


execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -281,16 +288,43 @@ impl StateProvider for LmdbGlobalState {
...
-        let missing_descendants =
-            missing_trie_keys::<Key, StoredValue, lmdb::RoTransaction, LmdbTrieStore, Self::Error>(
+        if trie_keys
+            .iter()
+            .filter(|digest| {
+                !self
+                    .digests_without_missing_descendants
+                    .read()
+                    .unwrap()
+                    .contains(digest)
```
dwerner: Actually, isn't this going to slow this check down substantially, if we are locking for every single digest? I think you should hoist the read() call to above and then call contains.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -281,16 +288,43 @@ impl StateProvider for LmdbGlobalState {
...
+                StoredValue,
+                lmdb::RoTransaction,
+                LmdbTrieStore,
+                Self::Error,
+            >(
                 correlation_id,
                 &txn,
                 self.trie_store.deref(),
-                trie_keys,
+                trie_keys.clone(),
```
dwerner, line 317: Why did you have to clone this?

dwerner (commented):
This might work depending on what number of tries/nodes we have in mainnet and testnet. I've asked @sacherjj to get those. However, if we do want to use some variant of this, we'll want to be sure we can prune this `HashSet` post join with fast-sync.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -281,16 +289,58 @@ impl StateProvider for LmdbGlobalState {
...
     ) -> Result<Vec<Digest>, Self::Error> {
-        let txn = self.environment.create_read_txn()?;
-        let missing_descendants =
-            missing_trie_keys::<Key, StoredValue, lmdb::RoTransaction, LmdbTrieStore, Self::Error>(
+        if trie_keys
+            .iter()
+            .filter(|digest| {
+                !self
+                    .digests_without_missing_descendants
+                    .read()
```
afck: Would it be cheaper to acquire the read lock only once, before filtering, instead of in every iteration?

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -281,16 +289,58 @@ impl StateProvider for LmdbGlobalState {
...
+                    .unwrap()
+                    .contains(digest)
+            })
+            .count()
+            == 0
```
afck: Even if it's not `0` we still want to filter out the trie keys that we know don't have missing descendants, so that we call `missing_trie_keys` below only with the others.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -281,16 +289,58 @@ impl StateProvider for LmdbGlobalState {
...
+                trie_keys.clone(),
+                &self.digests_without_missing_descendants.write().unwrap(),
             )?;
-        txn.commit()?;
-        Ok(missing_descendants)
+            if missing_descendants.is_empty() {
+                // There were no missing descendants on `trie_keys`, let's add them *and all of
+                // their descendents* to the cache.
+
+                let mut all_descendents: HashSet<Digest> =
```
afck:
```suggestion
                let mut all_descendants: HashSet<Digest> =
```

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -255,6 +256,11 @@ where
             continue;
         }
 
+        if known_complete.contains(&trie_key) {
+            // Skip because we know there are no missing descendents.
```
afck:
```suggestion
            // Skip because we know there are no missing descendants.
```

rafal-ch:
Marked as `do not merge` since the code is still under development as we add more optimization.

execution_engine/src/storage/global_state/lmdb.rs
rafal-ch: Good catch!
Fixed in aac1d6b4d9dba612378a1af8429fe70f09415908

execution_engine/src/storage/global_state/lmdb.rs
rafal-ch: Yup, Dan pointed this one too, thanks!
Fixed in aac1d6b4d9dba612378a1af8429fe70f09415908

execution_engine/src/storage/global_state/lmdb.rs
rafal-ch: Fixed in aac1d6b4d9dba612378a1af8429fe70f09415908

execution_engine/src/storage/trie_store/operations/mod.rs
rafal-ch: Fixed in https://github.com/casper-network/casper-node/commit/aac1d6b4d9dba612378a1af8429fe70f09415908

execution_engine/src/storage/global_state/lmdb.rs
rafal-ch: Because `missing_trie_keys()` consumes the `trie_keys` and we still need it in the code below.

In the meantime, yet another `clone()` appeared, so I made a small modification to get rid of this additional cloning. Please take a look here: 9038f572111a7e4d6f788be8ac05577ca69eed5b

execution_engine/src/storage/global_state/lmdb.rs
goral09: I've looked at the call sites and we always call this method (`missing_trie_keys`) with just one `Digest`. This means that it's either zero or single trie key we will call the `missing_trie_keys` with.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -281,16 +289,68 @@ impl StateProvider for LmdbGlobalState {
...
-            missing_trie_keys::<Key, StoredValue, lmdb::RoTransaction, LmdbTrieStore, Self::Error>(
+        let trie_count = {
+            let digests_without_missing_descendants = self
+                .digests_without_missing_descendants
+                .read()
+                .expect("digest cache read lock");
+            trie_keys
+                .iter()
+                .filter(|digest| !digests_without_missing_descendants.contains(digest))
+                .count()
```
afck, line 300: We should remove the `trie_keys` without missing descendants even if the count is not 0, so we don't redundantly run `missing_trie_keys` on them.

(Pointing this out again because https://github.com/casper-network/casper-node/pull/2857#discussion_r849432211 is "outdated".)

goral09: I put my comment on a wrong thread... reposting here:

I've looked at the call sites and we always call this method (`missing_trie_keys`) with just one `Digest`. This means that it's either zero or single trie key we will call the `missing_trie_keys` with.

afck: OK, then I guess it's not a blocker, but still…
We could make the function take just a single one as an argument.

goral09: Didn't want to change an API for a public method as part of a "bug fix". We can ticket that work.

Fraser999 (approved):

goral09 (approved):

PR #2865, opened 2022-04-14
Remove misleading comment in Deploy acceptor

### LINKED ISSUE #2863: Remove comment about cryptographic validity within the Deploy acceptor's `accept()` method
Change requested - original message:

> "The DA's accept method has a comment that is leading some to believe that the DA isn't performing cryptographic checks for incoming deploys which is not the case, the cryptographic checking is done by a method that hangs off the Deploy struct itself. It is prudent to remove the comment entirely as it leads people to make incorrect inferences."

darthsiroftardis:
CHANGELOG:

- Removed misleading comment leading people to incorrectly infer that the deploy acceptor doesn't perform cryptographic checks.

Closes #2863

### DIFF
--- a/node/src/components/deploy_acceptor.rs
+++ b/node/src/components/deploy_acceptor.rs
@@ -184,7 +184,6 @@ impl DeployAcceptor {
             self.max_associated_keys,
         );
         // checks chainspec values
-        // DOES NOT check cryptographic security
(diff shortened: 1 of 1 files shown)


goral09 (approved):

darthsiroftardis:
bors try

PR #2869, opened 2022-04-15
(feat-rocksdb) Implement scratch trie store

dwerner:
Implements a scratch trie store. This change implements a cache to eliminate interstitial roots from being written to the database. In my tests, this means that the commit_step runs in ~3 seconds consistently, where before we'd see spikes as big as 60s. This also substantially reduced write amplification due to interstitial tries being no longer written to disk. This is effectively "bulk update" of the trie.

image

### DIFF
--- a/execution_engine/src/storage/global_state/scratch.rs
+++ b/execution_engine/src/storage/global_state/scratch.rs
@@ -25,26 +25,33 @@ use super::db::{DbGlobalState, DbGlobalStateView};
 type SharedCache = Arc<RwLock<Cache>>;
 
 struct Cache {
-    stored_values: HashMap<Key, StoredValue>,
+    cached_values: HashMap<Key, (bool, StoredValue)>,
 }
 
 impl Cache {
     fn new() -> Self {
         Cache {
-            stored_values: HashMap::new(),
+            cached_values: HashMap::new(),
         }
     }
 
-    fn insert(&mut self, key: Key, value: StoredValue) {
-        self.stored_values.insert(key, value);
+    fn insert_write(&mut self, key: Key, value: StoredValue) {
+        self.cached_values.insert(key, (true, value));
+    }
+
+    fn insert_read(&mut self, key: Key, value: StoredValue) {
+        self.cached_values.entry(key).or_insert((false, value));
     }
 
     fn get(&self, key: &Key) -> Option<&StoredValue> {
-        self.stored_values.get(key)
+        self.cached_values.get(key).map(|(_dirty, value)| value)
     }
 
-    fn into_inner(self) -> HashMap<Key, StoredValue> {
-        self.stored_values
+    fn into_dirty_writes(self) -> HashMap<Key, StoredValue> {
+        self.cached_values
+            .into_iter()
+            .filter_map(|(key, (dirty, value))| if dirty { Some((key, value)) } else { None })
+            .collect()
     }
 }
 
@@ -77,7 +84,7 @@ impl ScratchGlobalState {
     /// Consume self and return inner cache.
     pub fn into_inner(self) -> HashMap<Key, StoredValue> {
         let cache = mem::replace(&mut *self.cache.write().unwrap(), Cache::new());
-        cache.into_inner()
+        cache.into_dirty_writes()
     }
 }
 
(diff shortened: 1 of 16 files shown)


execution_engine_testing/tests/bin/disk_use.rs
```diff
@@ -71,115 +93,148 @@ pub fn multiple_native_transfers(
...
+            Trie::Extension { affix: _, pointer } => match pointer {
+                Pointer::LeafPointer(digest) | Pointer::NodePointer(digest) => queue.push(digest),
+            },
         }
-        writeln!(&mut time_report, " {}", transfer_count).unwrap();
     }
+}
 
-    println!("wrote reports");
+fn main() {
```
xcthulhu, line 5: I suggest we make this a `#[test]` so it gets run in CI (on a small scale).

Perhaps we could make it configurable with an environment parameter so we could run a longer nightly test to check for regressions?

Or maybe we just do the latter...

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -25,26 +25,33 @@ use super::db::{DbGlobalState, DbGlobalStateView};
...
-        self.stored_values.get(key)
+        self.cached_values.get(key).map(|(_dirty, value)| value)
     }
 
-    fn into_inner(self) -> HashMap<Key, StoredValue> {
-        self.stored_values
+    fn into_dirty_writes(self) -> HashMap<Key, StoredValue> {
+        self.cached_values
+            .into_iter()
+            .filter_map(|(key, (dirty, value))| if dirty { Some((key, value)) } else { None })
```
xcthulhu, line 53: 👍 

@Fraser999 we want at least this part in 1.5.0 IMO.  In a drive-by review, @EdHastingsCasperLabs identified that scratch global state is doing unnecessary writes.  This fixes that issue.

execution_engine/src/storage/trie_store/db.rs
```diff
@@ -60,12 +68,151 @@ impl<K, V> TrieStore<K, V> for LmdbTrieStore {}
...
+#[derive(Clone)]
+pub struct ScratchCache {
+    pub(crate) cache: Cache,
+    pub(crate) rocksdb_store: RocksDbStore,
+}
+
+/// Cached version of the trie store.
+#[derive(Clone)]
+pub struct ScratchTrieStore {
+    pub(crate) store: ScratchCache,
```
xcthulhu, line 90: For the scratch trie store, we can get away with `Arc<Mutex<HashMap<Digest, Trie<Key, StoredValue>>>>`...

dwerner: I opted to newtype over the Arc, but we still need this to pass around as a special associated type in all these operations.

execution_engine/src/storage/trie_store/db.rs
```diff
@@ -60,12 +68,151 @@ impl<K, V> TrieStore<K, V> for LmdbTrieStore {}
...
+        }
+    }
+
+    /// Writes all dirty (modified) cached tries to rocksdb.
+    pub fn write_all_tries_to_rocksdb(self, new_state_root: Digest) -> Result<(), error::Error> {
+        let rocksdb_store = self.store.rocksdb_store;
+        let cache = &mut *self.store.cache.lock().map_err(|_| error::Error::Poison)?;
+
+        let mut txn = rocksdb_store.create_read_write_txn()?;
+        let mut missing_trie_keys = vec![new_state_root];
```
dwerner: TODO: rename to new_trie_keys

execution_engine/src/storage/global_state/scratch.rs
dwerner: I've made a separate PR against `release-1.5.0` here: https://github.com/casper-network/casper-node/pull/2873.

execution_engine_testing/tests/bin/disk_use.rs
dwerner: I did this in https://github.com/casper-network/casper-node/pull/2877

dwerner:
Significant changes here already merged to feat-rocksdb

PR #2870, opened 2022-04-19
Store children before parent tries

### LINKED ISSUE #2860: Speed up fast sync: store trie children before parents.
Since find_missing_descendants is slow and does a lot of redundant work, let's instead introduce the invariant that a trie node can never be stored without its children.
If fast sync were sequential, a depth-first traversal of the trie would thus require O(log(N)) nodes in memory at any time.
We need to find a way to parallelize that a bit, so that with t worker tasks, it requires O(t · log(N)) nodes in memory without doing any redundant downloads

marc-casperlabs:
Changes the fast-sync code to store tries in post-order when downloading, thus obviating the need to find missing descendants. Requires that the trie store has stored only complete tries. This is meant as an experimental improvement for fast sync performance.

This commits eliminates the work queue in favor of recursive solution. It is not "true" recursion, as tasks are spawned instead of awaiting, thus hopefully eliminating potential stack overflows should the trie grow arbitrarily deep (small demo: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=1661dae87acc9e4c9913551f91c24ead). Performance should be at least ballpark-comparable (https://casperlabs-team.slack.com/archives/CGPDWN4MQ/p1649866751247089).

This code has seen very little testing yet; it may very well crash or lock on first try.

## Open questions

The actual type (called `NetTrie` in the code) for the `Trie` nodes received over the network was unclear. A small bit of refactoring has made the ancestor-determining functions available on `Trie` itself using only raw bytes, but for this to work, the type must be known. I was not sure if there are one or two types sent serialized across the wire, see the comment for details.

## Missing features

* A self-healing function should be added, to be either run when accidentally discovering missing tries, and/or at start-up or through manual intervention, to fill potential holes in the store.
* The `utils/retrieve-state` program has not been updated, it will probably corrupt the store when run.
* It could be beneficial to use a lock to enforce only one root state hash to be downloaded at the same time, as they are guaranteed to be cycle free. Otherwise, parallel processes could end up downloading the same trie nodes twice. This is not a current concern, as far as I know though.

Closes #2860.

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -1722,34 +1722,16 @@ where
         Ok(self.state.get_trie_full(correlation_id, &trie_key)?)
     }
 
-    /// Puts a trie and finds missing descendant trie keys.
-    pub fn put_trie_and_find_missing_descendant_trie_keys(
-        &self,
-        correlation_id: CorrelationId,
-        trie_bytes: &[u8],
-    ) -> Result<Vec<Digest>, Error>
+    /// Puts a trie.
+    // TODO: Offer to verify the trie has all of its descendents present?
+    // TODO: Do we want to return the calculated inserted trie key?
+    pub fn put_trie(&self, correlation_id: CorrelationId, trie_bytes: &[u8]) -> Result<(), Error>
     where
         Error: From<S::Error>,
     {
-        let inserted_trie_key = self.state.put_trie(correlation_id, trie_bytes)?;
-        let missing_descendant_trie_keys = self
-            .state
-            .missing_trie_keys(correlation_id, vec![inserted_trie_key])?;
-        Ok(missing_descendant_trie_keys)
-    }
+        let _inserted_trie_key = self.state.put_trie(correlation_id, trie_bytes)?;
 
-    /// Performs a lookup for a list of missing root hashes.
-    pub fn missing_trie_keys(
-        &self,
-        correlation_id: CorrelationId,
-        trie_keys: Vec<Digest>,
-    ) -> Result<Vec<Digest>, Error>
-    where
-        Error: From<S::Error>,
-    {
-        self.state
-            .missing_trie_keys(correlation_id, trie_keys)
-            .map_err(Error::from)
+        Ok(())
     }
 
     /// Obtains validator weights for given era.
--- a/execution_engine/src/storage/trie_store/operations/mod.rs
+++ b/execution_engine/src/storage/trie_store/operations/mod.rs
@@ -264,44 +264,16 @@ where
             }
         };
 
-        // Optimization: Don't deserialize leaves as they have no descendants.
-        if let Some(&Trie::<K, V>::LEAF_TAG) = retrieved_trie_bytes.first() {
-            continue;
-        }
-
-        // Parse the trie, handling errors gracefully.
-        let retrieved_trie = match bytesrepr::deserialize_from_slice(retrieved_trie_bytes) {
-            Ok(retrieved_trie) => retrieved_trie,
-            // Couldn't parse; treat as missing and continue.
+        match Trie::<K, V>::descendants_from_bytes(retrieved_trie_bytes) {
+            Ok(iter) => missing_descendants.extend(iter),
             Err(err) => {
                 error!(?err, "unable to parse trie");
                 missing_descendants.push(trie_key);
                 continue;
             }
-        };
-
-        match retrieved_trie {
-            // Should be unreachable due to checking the first byte as a shortcut above.
-            Trie::<K, V>::Leaf { .. } => {
-                error!("did not expect to see a trie leaf in `missing_trie_keys` after shortcut");
-            }
-            // If we hit a pointer block, queue up all of the nodes it points to
-            Trie::Node { pointer_block } => {
-                for (_, pointer) in pointer_block.as_indexed_pointers() {
-                    match pointer {
-                        Pointer::LeafPointer(descendant_leaf_trie_key) => {
-                            trie_keys_to_visit.push(descendant_leaf_trie_key)
-                        }
-                        Pointer::NodePointer(descendant_node_trie_key) => {
-                            trie_keys_to_visit.push(descendant_node_trie_key)
-                        }
-                    }
-                }
-            }
-            // If we hit an extension block, add its pointer to the queue
-            Trie::Extension { pointer, .. } => trie_keys_to_visit.push(pointer.into_hash()),
         }
     }
+
     Ok(missing_descendants)
 }
 
--- a/execution_engine/src/storage/trie_store/operations/mod.rs
+++ b/execution_engine/src/storage/trie_store/operations/mod.rs
@@ -1,14 +1,7 @@
 #[cfg(test)]
 mod tests;
 
-use std::{
-    cmp,
-    collections::{HashSet, VecDeque},
-    convert::TryInto,
-    mem,
-};
-
-use tracing::error;
+use std::{cmp, collections::VecDeque, convert::TryInto, mem};
 
 use casper_hashing::Digest;
 use casper_types::bytesrepr::{self, FromBytes, ToBytes};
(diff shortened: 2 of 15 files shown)


execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -264,44 +264,16 @@ where
...
-                            trie_keys_to_visit.push(descendant_node_trie_key)
-                        }
-                    }
-                }
-            }
-            // If we hit an extension block, add its pointer to the queue
-            Trie::Extension { pointer, .. } => trie_keys_to_visit.push(pointer.into_hash()),
         }
     }
+
```
afck: The method documentation should probably be upgraded: It doesn't return all missing _descendants_, just all missing children or the nodes themselves.

And `while` could be replaced with `for` now.

(I assume in all the remaining call sites — if there are any — the changed behavior is fine?)

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -570,54 +559,73 @@ async fn fetch_and_store_block_by_hash(
...
-    }
+    recursive_trie_download(state_root_hash, ctx.clone()).await?;
 
     Ok(())
 }
 
+/// Downloads a given trie and all of its descendants, then stores them into the trie store.
+///
+/// Ensures that all descendant nodes are downloaded using post-order tree traversal, i.e. child
+/// nodes are recursively downloaded and store first before the parent node is stored.
```
```suggestion
/// nodes are recursively downloaded and stored first before the parent node is stored.
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -570,54 +559,73 @@ async fn fetch_and_store_block_by_hash(
...
+    let fetched_trie = fetch_trie_retry_forever(hash, &ctx).await?;
+
+    match fetched_trie {
+        TrieAlreadyPresentOrDownloaded::AlreadyPresent => {
+            // The trie node was already present, meaning all its descendants are also present.
+            // Nothing to do.
+            Ok(())
+        }
+        TrieAlreadyPresentOrDownloaded::Downloaded(trie_bytes) => {
+            let mut descendent_handles: FuturesUnordered<_> =
```
```suggestion
            let mut descendant_handles: FuturesUnordered<_> =
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -570,54 +559,73 @@ async fn fetch_and_store_block_by_hash(
...
+
+            // At this point, we know all descendants are stored, so we store the parent node.
+            let descendants = ctx
+                .effect_builder
+                .put_trie_and_find_missing_descendant_trie_keys(trie_bytes)
+                .await?;
+
+            // Warn if we got an unexpected missing descendent.
+            if !descendants.is_empty() {
+                error!(trie_hash=%hash, "stored trie had missing descendants, even though they were all downloaded. this is a bug or the trie store is corrupted");
```
```suggestion
                error!(
                    trie_hash=%hash,
                    "stored trie had missing descendants, even though they were all downloaded. \
                    this is a bug or the trie store is corrupted"
                );
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -570,54 +559,73 @@ async fn fetch_and_store_block_by_hash(
...
+
+                // Note: As an alternative, we can try to recover and download the missing items.
+            }
+
+            Ok(())
+        }
+    }
+}
```
afck, line 617: It seems like the order in which this would attempt to download the trie would be mostly breadth-first, and we'd end up having most of the trie in memory before storing anything, because the Tokio semaphore will return permits in the order in which they are requested.

We should somehow prioritize deeper trie nodes.

execution_engine/src/storage/trie_store/operations/mod.rs
xcthulhu: I would prefer if we just got rid of `missing_trie_keys` altogether, I do not intend to support it in the new trie design.

node/src/components/chain_synchronizer/error.rs
```diff
@@ -143,6 +143,16 @@ pub(crate) enum Error {
         #[serde(skip_serializing)]
         FetchTrieError,
     ),
+
+    /// A received trie matched the correct hash, but failed to serialize into a valid trie.
+    #[error("corrupt trie {hash}: {error}")]
+    CorruptTrie {
+        /// The hash of the corrupt trie.
+        hash: Digest,
+        #[serde(skip_serializing)]
```
xcthulhu: I suspect we are serializing our errors with JSON for logging purposes.  If so, why are we skipping the error message string?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -36,9 +31,12 @@ use crate::{
         Deploy, DeployHash, FinalizedApprovals, FinalizedApprovalsWithId, FinalizedBlock, Item,
         NodeId, TimeDiff, Timestamp,
     },
-    utils::work_queue::WorkQueue,
 };
 
+/// The kind of trie actually sent across the network.
+// TODO: Does the root hash differ in that it is a nested `Trie<Digest, Trie<Digest, StoredValue>>`?
+type NetTrie = Trie<Digest, StoredValue>;
```
xcthulhu, line 38: Shouldn't this be `Trie<Key, StoredValue>` ?

execution_engine/src/storage/trie_store/operations/mod.rs
marc-casperlabs: The function is essentially unused already, only present in a test. I have removed it in a commit as a demonstration: cfc777e82

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -1722,34 +1722,16 @@ where
...
     where
         Error: From<S::Error>,
     {
-        let inserted_trie_key = self.state.put_trie(correlation_id, trie_bytes)?;
-        let missing_descendant_trie_keys = self
-            .state
-            .missing_trie_keys(correlation_id, vec![inserted_trie_key])?;
-        Ok(missing_descendant_trie_keys)
-    }
+        let _inserted_trie_key = self.state.put_trie(correlation_id, trie_bytes)?;
```
afck, line 1732: Should we check whether all children (not descendants!) are present, and return an error if not?

On the other hand that would make it slower, of course. We could offer a `_no_check` version of the method if the caller is confident all children were stored first.

In either case the method documentation should mention the new invariant.

node/src/components/chain_synchronizer/operations.rs
marc-casperlabs: This code has been removed :)

utils/retrieve-state/src/lib.rs
```diff
@@ -463,10 +463,7 @@ pub async fn download_trie_channels(
                 }) => {
                     let mut missing_trie_descendants = tokio::task::block_in_place(|| {
```
afck, line 464: This tool now needs to be rewritten to store the trie in the right order, too.
Actually, I don't understand how this still compiles: We're calling `len()` on this below.

node/src/components/chain_synchronizer/error.rs
marc-casperlabs: The serialization is not used in logging, but in queue dumps only (logs don't allow for arbitrary JSON to be stored).

The skip here is a mistake though, leftover from an intermediate state of the code.

Fixed in 635019f43

utils/retrieve-state/src/lib.rs
marc-casperlabs: I've left that tool out for now, as this is still a draft (it was intended as a quick PoC for potential 1.5.0 inclusion).

node/src/components/chain_synchronizer/operations.rs
marc-casperlabs: That's a good point. The issue is, however, that strictly depth-first cannot be parallelized, if I am not mistaken, since we do not know the ultimate depth of the branch.

What's a good trade-off here? Guessing by depth maybe?

Could we maybe use permits to limit breadth, but make depth essentially free?

PR #2873, opened 2022-04-19
Eliminate writes of non-dirty values in ScratchGlobalState

dwerner:


execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -30,26 +30,33 @@ use crate::{
 type SharedCache = Arc<RwLock<Cache>>;
 
 struct Cache {
-    stored_values: HashMap<Key, StoredValue>,
+    cached_values: HashMap<Key, (bool, StoredValue)>,
```
goral09, line 33: Alternatively, we could have separate cache for read-only and never write those back. I think something similar was done for `TrackingCopy`.

goral09 (approved):
So we're basically not writing back keys that were _read-only_ anymore?

Fraser999 (approved):

dwerner:
> writing back keys that were _read-only_?

Yes, we were writing keys that had only been read.

dwerner:
bors merge

dwerner:
bors merge

PR #2877, opened 2022-04-19
Port scratch trie to dev

dwerner:
This ports ScratchTrie and the related unit test for `disk_use` to dev, making it orthogonal to the rocksdb changes it arose from.

### DIFF
--- a/execution_engine/src/storage/global_state/lmdb.rs
+++ b/execution_engine/src/storage/global_state/lmdb.rs
@@ -102,7 +102,7 @@ impl LmdbGlobalState {
             prestate_hash,
             stored_values,
         )?;
-        scratch_trie.write_all_tries_to_db(new_state_root)?;
+        scratch_trie.write_root_to_db(new_state_root)?;
         Ok(new_state_root)
     }
 
--- a/execution_engine/src/storage/global_state/lmdb.rs
+++ b/execution_engine/src/storage/global_state/lmdb.rs
@@ -18,7 +18,7 @@ use crate::{
             TrieOrChunkId,
         },
         trie_store::{
-            lmdb::LmdbTrieStore,
+            lmdb::{LmdbTrieStore, ScratchTrieStore},
             operations::{
                 keys_with_prefix, missing_trie_keys, put_trie, read, read_with_proof, ReadResult,
             },
@@ -94,14 +94,33 @@ impl LmdbGlobalState {
         prestate_hash: Digest,
         stored_values: HashMap<Key, StoredValue>,
     ) -> Result<Digest, error::Error> {
-        put_stored_values::<LmdbEnvironment, LmdbTrieStore, error::Error>(
-            &self.environment,
-            &self.trie_store,
+        let scratch_trie = self.get_scratch_store();
+        let new_state_root = put_stored_values::<_, _, error::Error>(
+            &scratch_trie,
+            &scratch_trie,
             correlation_id,
             prestate_hash,
             stored_values,
-        )
-        .map_err(Into::into)
+        )?;
+        scratch_trie.write_all_tries_to_db(new_state_root)?;
+        Ok(new_state_root)
+    }
+
+    /// Gets a scratch trie store.
+    fn get_scratch_store(&self) -> ScratchTrieStore {
+        ScratchTrieStore::new(Arc::clone(&self.trie_store), Arc::clone(&self.environment))
+    }
+
+    /// Get a reference to the lmdb global state's environment.
+    #[must_use]
+    pub fn environment(&self) -> &LmdbEnvironment {
+        &self.environment
+    }
+
+    /// Get a reference to the lmdb global state's trie store.
+    #[must_use]
+    pub fn trie_store(&self) -> &LmdbTrieStore {
+        &self.trie_store
     }
 }
 
@@ -237,9 +256,9 @@ impl StateProvider for LmdbGlobalState {
             || Ok(None),
             |bytes| {
                 if bytes.len() <= ChunkWithProof::CHUNK_SIZE_BYTES {
-                    Ok(Some(TrieOrChunk::Trie(bytes.to_owned().into())))
+                    Ok(Some(TrieOrChunk::Trie(bytes)))
                 } else {
-                    let chunk_with_proof = ChunkWithProof::new(bytes, trie_index)?;
+                    let chunk_with_proof = ChunkWithProof::new(&bytes, trie_index)?;
                     Ok(Some(TrieOrChunk::ChunkWithProof(chunk_with_proof)))
                 }
             },
@@ -256,8 +275,7 @@ impl StateProvider for LmdbGlobalState {
     ) -> Result<Option<Bytes>, Self::Error> {
         let txn = self.environment.create_read_txn()?;
         let ret: Option<Bytes> =
-            Store::<Digest, Trie<Digest, StoredValue>>::get_raw(&*self.trie_store, &txn, trie_key)?
-                .map(|slice| slice.to_owned().into());
+            Store::<Digest, Trie<Digest, StoredValue>>::get_raw(&*self.trie_store, &txn, trie_key)?;
         txn.commit()?;
         Ok(ret)
     }
--- a/execution_engine/src/storage/global_state/scratch.rs
+++ b/execution_engine/src/storage/global_state/scratch.rs
@@ -52,6 +52,8 @@ impl Cache {
         self.cached_values.get(key).map(|(_dirty, value)| value)
     }
 
+    /// Consumes self and returns only written values as values that were only read must be filtered
+    /// out to prevent unecessary writes.
     fn into_dirty_writes(self) -> HashMap<Key, StoredValue> {
         self.cached_values
             .into_iter()
--- a/execution_engine/src/storage/global_state/scratch.rs
+++ b/execution_engine/src/storage/global_state/scratch.rs
@@ -30,26 +30,33 @@ use crate::{
 type SharedCache = Arc<RwLock<Cache>>;
 
 struct Cache {
-    stored_values: HashMap<Key, StoredValue>,
+    cached_values: HashMap<Key, (bool, StoredValue)>,
 }
 
 impl Cache {
     fn new() -> Self {
         Cache {
-            stored_values: HashMap::new(),
+            cached_values: HashMap::new(),
         }
     }
 
-    fn insert(&mut self, key: Key, value: StoredValue) {
-        self.stored_values.insert(key, value);
+    fn insert_write(&mut self, key: Key, value: StoredValue) {
+        self.cached_values.insert(key, (true, value));
+    }
+
+    fn insert_read(&mut self, key: Key, value: StoredValue) {
+        self.cached_values.entry(key).or_insert((false, value));
     }
 
     fn get(&self, key: &Key) -> Option<&StoredValue> {
-        self.stored_values.get(key)
+        self.cached_values.get(key).map(|(_dirty, value)| value)
     }
 
-    fn into_inner(self) -> HashMap<Key, StoredValue> {
-        self.stored_values
+    fn into_dirty_writes(self) -> HashMap<Key, StoredValue> {
+        self.cached_values
+            .into_iter()
+            .filter_map(|(key, (dirty, value))| if dirty { Some((key, value)) } else { None })
+            .collect()
     }
 }
 
@@ -96,7 +103,7 @@ impl ScratchGlobalState {
     /// Consume self and return inner cache.
     pub fn into_inner(self) -> HashMap<Key, StoredValue> {
         let cache = mem::replace(&mut *self.cache.write().unwrap(), Cache::new());
-        cache.into_inner()
+        cache.into_dirty_writes()
     }
 }
 
@@ -120,7 +127,7 @@ impl StateReader<Key, StoredValue> for ScratchGlobalStateView {
             key,
         )? {
             ReadResult::Found(value) => {
-                self.cache.write().unwrap().insert(*key, value.clone());
+                self.cache.write().unwrap().insert_read(*key, value.clone());
                 Some(value)
             }
             ReadResult::NotFound => None,
@@ -246,7 +253,7 @@ impl CommitProvider for ScratchGlobalState {
                 },
             };
 
-            self.cache.write().unwrap().insert(key, value);
+            self.cache.write().unwrap().insert_write(key, value);
         }
         Ok(state_hash)
     }
--- a/execution_engine/src/storage/global_state/scratch.rs
+++ b/execution_engine/src/storage/global_state/scratch.rs
@@ -298,9 +298,9 @@ impl StateProvider for ScratchGlobalState {
             || Ok(None),
             |bytes| {
                 if bytes.len() <= ChunkWithProof::CHUNK_SIZE_BYTES {
-                    Ok(Some(TrieOrChunk::Trie(bytes.to_owned().into())))
+                    Ok(Some(TrieOrChunk::Trie(bytes)))
                 } else {
-                    let chunk_with_proof = ChunkWithProof::new(bytes, trie_index)?;
+                    let chunk_with_proof = ChunkWithProof::new(&bytes, trie_index)?;
                     Ok(Some(TrieOrChunk::ChunkWithProof(chunk_with_proof)))
                 }
             },
@@ -317,8 +317,7 @@ impl StateProvider for ScratchGlobalState {
     ) -> Result<Option<Bytes>, Self::Error> {
         let txn = self.environment.create_read_txn()?;
         let ret: Option<Bytes> =
-            Store::<Digest, Trie<Digest, StoredValue>>::get_raw(&*self.trie_store, &txn, trie_key)?
-                .map(|slice| slice.to_owned().into());
+            Store::<Digest, Trie<Digest, StoredValue>>::get_raw(&*self.trie_store, &txn, trie_key)?;
         txn.commit()?;
         Ok(ret)
     }
--- a/execution_engine/src/storage/transaction_source/lmdb.rs
+++ b/execution_engine/src/storage/transaction_source/lmdb.rs
@@ -8,23 +8,23 @@ use lmdb::{
 use crate::storage::{
     error,
     transaction_source::{Readable, Transaction, TransactionSource, Writable},
-    trie_store::lmdb::{ScratchCache, ScratchTrieStore},
+    trie_store::lmdb::ScratchTrieStore,
     MAX_DBS,
 };
 
 /// Filename for the LMDB database created by the EE.
 const EE_DB_FILENAME: &str = "data.lmdb";
 
-impl Transaction for ScratchCache {
+impl Transaction for ScratchTrieStore {
     type Error = error::Error;
-    type Handle = ScratchCache;
+    type Handle = ScratchTrieStore;
     fn commit(self) -> Result<(), Self::Error> {
         // NO OP as scratch doesn't use transactions.
         Ok(())
     }
 }
 
-impl Readable for ScratchCache {
+impl Readable for ScratchTrieStore {
     fn read(&self, handle: Self::Handle, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
         let txn = self.env.create_read_txn()?;
         match lmdb::Transaction::get(&txn, handle.store.get_db(), &key) {
@@ -35,7 +35,7 @@ impl Readable for ScratchCache {
     }
 }
 
-impl Writable for ScratchCache {
+impl Writable for ScratchTrieStore {
     fn write(&mut self, handle: Self::Handle, key: &[u8], value: &[u8]) -> Result<(), Self::Error> {
         let mut txn = self.env.create_read_write_txn()?;
         txn.put(handle.store.get_db(), &key, &value, WriteFlags::empty())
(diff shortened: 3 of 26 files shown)


execution_engine/src/storage/transaction_source/lmdb.rs
```diff
@@ -44,9 +89,9 @@ impl<'a> Transaction for RwTransaction<'a> {
 }
 
 impl<'a> Readable for RwTransaction<'a> {
-    fn read(&self, handle: Self::Handle, key: &[u8]) -> Result<Option<&[u8]>, Self::Error> {
+    fn read(&self, handle: Self::Handle, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
         match lmdb::Transaction::get(self, handle, &key) {
-            Ok(bytes) => Ok(Some(bytes)),
+            Ok(bytes) => Ok(Some(Bytes::from(bytes))),
```
goral09, line 94: Is this necessary? I think we made this change with @mpapierski to avoid unnecessary allocations.

execution_engine/src/storage/global_state/scratch.rs
```diff
@@ -30,26 +30,33 @@ use crate::{
...
     }
 
     fn get(&self, key: &Key) -> Option<&StoredValue> {
-        self.stored_values.get(key)
+        self.cached_values.get(key).map(|(_dirty, value)| value)
     }
 
-    fn into_inner(self) -> HashMap<Key, StoredValue> {
-        self.stored_values
+    fn into_dirty_writes(self) -> HashMap<Key, StoredValue> {
```
goral09, line 57: Can you add a comment what we're doing here exactly? The _dirty_ part is not self-descriptive.

execution_engine/src/storage/transaction_source/lmdb.rs
dwerner: Yes, it is. This reverted @mpapierski 's optimization as a part of the rocksdb changes, but I found it was needed here to work around the Readable/Writable abstraction with this new cache. Previously we passed lifetimes bound by lmdb's transaction, but we now no longer have a transaction. Summing it up:
- since we are caching the lifetimes differ
- when we switch to rocksdb this isn't possible in this way
- callsites immediately copy, in order to put values onto the reactor

execution_engine/src/storage/global_state/scratch.rs
dwerner: Done in [8e07396](https://github.com/casper-network/casper-node/pull/2877/commits/8e073968d9fc46929294be4f4c6d58d786b9960b)

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -339,12 +356,29 @@ impl LmdbWasmTestBuilder {
...
+                .engine_state
+                .write_scratch_to_lmdb(prestate_hash, scratch.into_inner())
+                .unwrap();
+            self.post_state_hash = Some(new_state_root);
+        }
+        self
+    }
+
+    /// run step against scratch global state.
+    pub fn step_with_scratch(&mut self, step_request: StepRequest) -> &mut Self {
```
goral09, line 369: Ideally, it shouldn't be required to call this or `write_scratch_to_lmdb` methods manually as they would be called implicitly on every execution. I'm not sure whether that's possible right now but it's also out of scope for this ticket most probably.

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -94,14 +94,33 @@ impl LmdbGlobalState {
         prestate_hash: Digest,
         stored_values: HashMap<Key, StoredValue>,
     ) -> Result<Digest, error::Error> {
-        put_stored_values::<LmdbEnvironment, LmdbTrieStore, error::Error>(
-            &self.environment,
-            &self.trie_store,
+        let scratch_trie = self.get_scratch_store();
+        let new_state_root = put_stored_values::<_, _, error::Error>(
+            &scratch_trie,
+            &scratch_trie,
```
goral09, line 100: Is it intentional that we use the same ScratchStore twice?

execution_engine/src/storage/transaction_source/lmdb.rs
goral09: This makes sense, thanks!

dwerner:
Since, in this PR, the scratch trie is still backed by LMDB, I ran a new set of disk use and performance benchmarks to compare with this turned off vs on.

image

execution_engine_testing/test_support/src/wasm_test_builder.rs
dwerner: It's the same contract used for ScratchGlobalState:
- use the scratch version of global state or the trie in place of the old
- call it's own implementation to flush to disk

WasmTestBuilder's design assumes that there is no such contract, and that makes it needed to work around. I agree that it isn't ideal, but also that it's not in scope. WasmTestBuilder's "builder"-ness and API would have to be rethought a bit here.

execution_engine/src/storage/global_state/lmdb.rs
dwerner: Yep, we need to emulate a transaction to make use of the generic traits `TransactionSource`, `Readable`, `Writable`, and related `write` `scan` and `read` functions. I think this area is ripe for refactoring, in particular once rocksdb lands, but is entirely out of scope for this changeset.

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
 }
 
 impl<K, V> TrieStore<K, V> for LmdbTrieStore {}
+
+pub(crate) type Cache = Arc<Mutex<HashMap<Digest, (bool, Trie<Key, StoredValue>)>>>;
```
Fraser999, line 179: Would probably be helpful to have a comment explaining what the key and value types of the hashmap represent.

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+/// In-memory cached trie store, backed by rocksdb.
```
Fraser999:
```suggestion
/// In-memory cached trie store, backed by lmdb.
```

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+#[derive(Clone)]
+pub struct ScratchCache {
```
Fraser999: We don't really need this type do we?  Couldn't we just have
```rust
pub(crate) struct ScratchTrieStore {
    pub(crate) cache: Cache,
    pub(crate) store: Arc<LmdbTrieStore>,
    pub(crate) env: Arc<LmdbEnvironment>,
}
```

Either way, I think this (if we keep it) and `ScratchTrieStore` should be restricted to `pub(crate)`.

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+        let store = self.inner.store;
+        let cache = &mut *self.inner.cache.lock().map_err(|_| error::Error::Poison)?;
+
+        let mut missing_trie_keys = vec![new_state_root];
+        let mut validated_tries = HashMap::new();
+
+        let mut txn = env.create_read_write_txn()?;
+
+        while let Some(next_trie_key) = missing_trie_keys.pop() {
+            if cache.is_empty() {
```
Fraser999: It's not clear to me why we need this `if` block.  This would return an error, even if `next_trie_key` was available in the underlying DB.  Seems like we could delete this block and just let the `None` case of the subsequent `match cache.remove()` handle it.

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+        Self {
+            inner: ScratchCache {
+                store,
+                env,
+                cache: Default::default(),
+            },
+        }
+    }
+
+    /// Writes all dirty (modified) cached tries to underlying db.
```
Fraser999: I think this comment is a bit misleading - it certainly fooled me for a while! :)  Rather than writing all dirty tries, it only writes those dirty tries which fall under `new_state_root`.

It's maybe also worth mentioning that all children under `new_state_root` are either read or written, and if reading uncovers a missing child, the method returns an error.

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+                ));
+            }
+            match cache.remove(&next_trie_key) {
+                Some((false, _)) => continue,
+                None => {
+                    txn.read(store.get_db(), next_trie_key.as_ref())?.ok_or(
+                        error::Error::CommitError(CommitError::TrieNotFoundDuringCacheValidate(
+                            next_trie_key,
+                        )),
+                    )?;
```
Fraser999: Not sure if this is worthwhile?  Maybe we should just skip the check (since it would only happen due to programmer error or DB corruption I guess?) as the missing trie might not be a fatal problem, but would result in a fatal error here.

execution_engine_testing/tests/bin/README.md
```diff
@@ -0,0 +1,15 @@
+# `disk_use` binary
+
+A binary that will construct global state and profile the disk use of various operations. It's recommended to run this tool in `--release` mode.
+
+The outcome is a report, `disk_use_report.csv`. This contains time-series data witthe following columns, with one line per block:
```
Fraser999:
```suggestion
The outcome is a report, `disk_use_report.csv`. This contains time-series data with the following columns, with one line per block:
```

execution_engine/src/storage/trie_store/lmdb.rs
xcthulhu:
```suggestion
/// Cache used by the scratch trie.  The keys represent the hash of the trie being cached.  The values represent:
///  1) A boolean, where `false` means the was _not_ written and `true` means it was
///  2) A deserialized trie
pub(crate) type Cache = Arc<Mutex<HashMap<Digest, (bool, Trie<Key, StoredValue>)>>>;
```

xcthulhu (approved):

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+        }
+    }
+
+    /// Writes all dirty (modified) cached tries to underlying db.
+    pub fn write_all_tries_to_db(self, new_state_root: Digest) -> Result<(), error::Error> {
+        let env = self.inner.env;
+        let store = self.inner.store;
+        let cache = &mut *self.inner.cache.lock().map_err(|_| error::Error::Poison)?;
+
+        let mut missing_trie_keys = vec![new_state_root];
```
xcthulhu: They aren't really "missing" trie keys, just the keys under the new state root...

execution_engine/src/storage/trie_store/lmdb.rs
xcthulhu: Yeah it doesn't look like this check is necessary I agree

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+                        )),
+                    )?;
+                }
+                Some((true, trie)) => {
+                    if let Some(children) = trie.children() {
+                        missing_trie_keys.extend(children);
+                    }
+                    validated_tries.insert(next_trie_key, trie);
+                }
+            }
```
xcthulhu:
```suggestion
            if let Some((true, trie)) = cache.remove(&next_trie_key) {
                if let Some(children) = trie.children() {
                    missing_trie_keys.extend(children);
                }
                store.put(&mut txn, digest, trie)?;
            }
```

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,150 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+    }
+
+    /// Writes all dirty (modified) cached tries to underlying db.
+    pub fn write_all_tries_to_db(self, new_state_root: Digest) -> Result<(), error::Error> {
+        let env = self.inner.env;
+        let store = self.inner.store;
+        let cache = &mut *self.inner.cache.lock().map_err(|_| error::Error::Poison)?;
+
+        let mut missing_trie_keys = vec![new_state_root];
+        let mut validated_tries = HashMap::new();
```
xcthulhu:
```suggestion
```

I think we should get rid of this and just `put` in the loop (see below)

execution_engine/src/storage/trie_store/lmdb.rs
xcthulhu:
```suggestion
        let mut trie_keys_to_visit_in_cache = vec![new_state_root];
```

execution_engine/src/storage/trie_store/lmdb.rs
xcthulhu: We may have to rewrite this to obey the invariant that the trie must never have a node which has missing descendants...

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: For now, we need an 'inner' type, distinct from `ScratchTrieStore`, as a workaround for dealing with the `TransactionSource` trait, we set this (`ScratchCache`) type as `Handle` associated type there (also, `ReadTransaction` and `ReadWriteTransaction` are emulated in this way). I used a similar pattern in the rocksdb changes and emulated that here.

dwerner: On your point about making this a crate-private type, unfortunately the `Handle` is leaked.

```
error[E0446]: crate-private type `storage::trie_store::lmdb::ScratchCache` in public interface
   --> execution_engine/src/storage/transaction_source/lmdb.rs:49:5
    |
49  |     type Handle = ScratchCache;
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ can't leak crate-private type
```

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: Going to rework this whole block to write children first.

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: Renamed and rewrote the doc-comment here, but we will revisit the implementation.

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: resolved in [4b78471](https://github.com/casper-network/casper-node/pull/2877/commits/4b78471c0fbac1bcaf8999792dcda1d0ded85d70)

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: resolved in [4b78471](https://github.com/casper-network/casper-node/pull/2877/commits/4b78471c0fbac1bcaf8999792dcda1d0ded85d70)

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: resolved in [4b78471](https://github.com/casper-network/casper-node/pull/2877/commits/4b78471c0fbac1bcaf8999792dcda1d0ded85d70)

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: resolved in [4b78471](https://github.com/casper-network/casper-node/pull/2877/commits/4b78471c0fbac1bcaf8999792dcda1d0ded85d70)

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: Implementation revisited and resolved in [4b78471](https://github.com/casper-network/casper-node/pull/2877/commits/4b78471c0fbac1bcaf8999792dcda1d0ded85d70)

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: resolved in [4b78471](https://github.com/casper-network/casper-node/pull/2877/commits/4b78471c0fbac1bcaf8999792dcda1d0ded85d70)

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: done

execution_engine/src/storage/trie_store/lmdb.rs
dwerner: done

dwerner:
@marc-casperlabs and @afck I wanted to note I stole your iterator implementation (`DescendantsIterator`) here https://github.com/casper-network/casper-node/pull/2870

execution_engine/src/storage/trie/mod.rs
```diff
@@ -551,6 +551,74 @@ impl<K, V> Trie<K, V> {
...
+                // We iterate until we hit the end of the block.
+                while *idx < pointer_block.0.len() {
+                    let cur = *idx;
+                    *idx += 1;
+
+                    match pointer_block.0[cur] {
+                        // An empty slot in the pointer block. Just increment and continue.
+                        None => continue,
+                        Some(ptr) => match ptr {
+                            Pointer::LeafPointer(key) | Pointer::NodePointer(key) => {
```
goral09: If we encounter a `NodePointer`, shouldn't we enqueue its children as well? My impression was that we want to be able to iterate over a trie's descendants until its leaves (which sit at the very bottom of a trie and have no children) but the iterator here would iterate only over immediate children of a trie node. Although, I can see that in your `write_root_to_db` you do that - i.e. fetch the trie and enqueue its children.

goral09: I probably misunderstood what this iterator is supposed to iterate over :)

Fraser999 (approved):

dwerner:
bors merge

execution_engine/src/storage/transaction_source/lmdb.rs
```diff
@@ -1,18 +1,63 @@
...
+
+impl Readable for ScratchTrieStore {
+    fn read(&self, handle: Self::Handle, key: &[u8]) -> Result<Option<Bytes>, Self::Error> {
+        let txn = self.env.create_read_txn()?;
+        match lmdb::Transaction::get(&txn, handle.store.get_db(), &key) {
+            Ok(bytes) => Ok(Some(Bytes::from(bytes))),
+            Err(lmdb::Error::NotFound) => Ok(None),
+            Err(e) => Err(error::Error::Lmdb(e)),
+        }
+    }
```
goral09, line 35: Why not check if we have it in the scatch first?

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,135 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
 }
 
 impl<K, V> TrieStore<K, V> for LmdbTrieStore {}
+
+/// Cache used by the scratch trie.  The keys represent the hash of the trie being cached.  The
+/// values represent:  1) A boolean, where `false` means the trie was _not_ written and `true` means
+/// it was 2) A deserialized trie
```
goral09, line 178: Formatting broke a bit.

execution_engine_testing/test_support/src/transfer.rs
```diff
@@ -0,0 +1,287 @@
...
+            .with_empty_payment_bytes(runtime_args! {})
+            .with_transfer_args(runtime_args)
+            .with_authorization_keys(&[source_account])
+            .build();
+        exec_builder = exec_builder.push_deploy(native_transfer);
+        let exec_request = exec_builder.build();
+        exec_requests.push(exec_request);
+    }
+    exec_requests
+}
```
goral09, line 287: I have a vague memory of seeing it somewhere already but I don't see this code being deleted anywhere. Is this a new code or a copied one?

execution_engine_testing/test_support/src/auction.rs
```diff
@@ -0,0 +1,402 @@
...
+            necessary_tries.len(),
+            total_tries,
+        )
+        .unwrap();
+        report_writer.flush().unwrap();
+    }
+}
+
+// find all necessary tries - hoist to FN
+fn find_necessary_tries<S>(
```
goral09, line 213: This method is basically an iterator over all tries under `state_root`, correct? It _could_ be a method in the `trie.rs` itself (or some other helper file), return an actual iterator over the trie which we would collect into the `HashSet`.

Iterator over descendants of a trie (in this case a state root) is also exactly what you have in `ScratchTrie::write_to_db`.

PR #2886, opened 2022-04-21
Fetch blocks and deploys in a single fetch request.

goral09:
Previously, each fast sync's _sync to Genesis_ round would consist of three, sequential steps:

1. fetch block's deploys
2. fetch block's trie
3. fetch block's parent (and use it as a base block in the next loop)

On top of that, each of the deploys and trie nodes would be requested independently, from a different peer. That adds a lot of latency to the process and slows it down considerably. Especially when the size of data we fetch, like a deploy or a trie node, is small - that makes us pay the cost of latency many times.

In this PR, we change the algorithm to the following:
1. fetch whole block and its deploys
2. fetch block's trie

Now, we make a single request to fetch the block and all of its deploys, hopefully speeding up the process.

**Why is this a safe thing to do?**
Our limit for a network message is ~23MB while a single block, with all of its deploys, can be ~12MB maximum.

 In this PR we're also adding a new histogram metric `chain_sync_fetch_block_and_deploys_duration_seconds` measuring time that it takes to download a block and all of its deploys.

### DIFF
--- a/node/src/components/chain_synchronizer/error.rs
+++ b/node/src/components/chain_synchronizer/error.rs
@@ -14,8 +14,8 @@ use casper_types::{EraId, ProtocolVersion};
 use crate::{
     components::{contract_runtime::BlockExecutionError, fetcher::FetcherError},
     types::{
-        Block, BlockHash, BlockHeader, BlockHeaderWithMetadata, BlockWithMetadata, Deploy,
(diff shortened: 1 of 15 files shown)


goral09:
Superseeded by https://github.com/casper-network/casper-node/pull/3010

PR #2888, opened 2022-04-21
Replace msgpack as the main network encoding format

### LINKED ISSUE #2881: Replace usage of msgpack in small_network

marc-casperlabs:
Closes #2881 

This changes the underlying network protocol to use msgpack only for the initial handshake for backwards compatibility. Actual network messages are instead encoded using fixed-settings `bincode` for now.

The tests have been improved, now using the actual encoder instead of a "replication". Furthermore a potential, rare race condition in the handshake has been removed by making the sending of the handshake a background task.

## Benchmarks

I've benchmarked this quickly using casper-test: https://github.com/casper-network/casper-test/blob/06db5ed0127b9e915201b683196515df55abb96a/src/bin/benchmark_network_traffic_gossiping.rs

`release-1.5.0` results (run twice):

```
Total bytes gossiped across the network: 79420
Total bytes gossiped across the network: 81964
```

With the changes in this PR:

```
Total bytes gossiped across the network: 42838
```

That's roughly a 45% reduction in traffic.

## Important

**Note**: If this merges, the resulting node will be **entirely incompatible with all previous versions of 1.5.0 currently used in testnets**. The underlying network protocol has changed in a manner invisible to the nodes, they will just report encoding errors when talking to older nodes.

### DIFF
--- a/node/src/components/small_network/bincode_format.rs
+++ b/node/src/components/small_network/bincode_format.rs
@@ -18,6 +18,7 @@ use tokio_serde::{Deserializer, Serializer};
 use super::Message;
 
 /// bincode encoder/decoder for messages.
+#[allow(clippy::type_complexity)]
 pub struct BincodeFormat(
     // Note: This scary looking type is because `bincode` encodes its options at the type level.
     //       The exact shape is determined by `BincodeFormat::default()`.
--- a/node/src/components/small_network/bincode_format.rs
+++ b/node/src/components/small_network/bincode_format.rs
@@ -20,8 +20,8 @@ use super::Message;
 /// bincode encoder/decoder for messages.
 #[allow(clippy::type_complexity)]
 pub struct BincodeFormat(
-    // Note: This scary looking type is because `bincode` encodes its options at the type level.
-    //       The exact shape is determined by `BincodeFormat::default()`.
+    // Note: `bincode` encodes its options at the type level. The exact shape is determined by
+    // `BincodeFormat::default()`.
     WithOtherTrailing<
         WithOtherIntEncoding<
             WithOtherEndian<
--- a/node/src/components/small_network/bincode_format.rs
+++ b/node/src/components/small_network/bincode_format.rs
@@ -1,0 +1,81 @@
+//! Bincode wire format encoder.
+//!
+//! An encoder for `Bincode` messages with our specific settings pinned.
+
+use std::{fmt::Debug, io, pin::Pin, sync::Arc};
+
+use bincode::{
+    config::{
+        RejectTrailing, VarintEncoding, WithOtherEndian, WithOtherIntEncoding, WithOtherLimit,
+        WithOtherTrailing,
+    },
+    Options,
+};
+use bytes::{Bytes, BytesMut};
+use serde::{Deserialize, Serialize};
+use tokio_serde::{Deserializer, Serializer};
+
+use super::Message;
+
+/// bincode encoder/decoder for messages.
+pub struct BincodeFormat(
+    // Note: This scary looking type is because `bincode` encodes its options at the type level.
+    //       The exact shape is determined by `BincodeFormat::default()`.
+    WithOtherTrailing<
+        WithOtherIntEncoding<
+            WithOtherEndian<
+                WithOtherLimit<bincode::DefaultOptions, bincode::config::Infinite>,
+                bincode::config::LittleEndian,
+            >,
+            VarintEncoding,
+        >,
+        RejectTrailing,
+    >,
+);
+
+impl Debug for BincodeFormat {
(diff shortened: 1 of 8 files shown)


node/src/components/small_network/bincode_format.rs
```diff
@@ -0,0 +1,81 @@
...
+use bytes::{Bytes, BytesMut};
+use serde::{Deserialize, Serialize};
+use tokio_serde::{Deserializer, Serializer};
+
+use super::Message;
+
+/// bincode encoder/decoder for messages.
+pub struct BincodeFormat(
+    // Note: This scary looking type is because `bincode` encodes its options at the type level.
+    //       The exact shape is determined by `BincodeFormat::default()`.
```
Fraser999:
```suggestion
    // Note: `bincode` encodes its options at the type level. The exact shape is determined by
    //       `BincodeFormat::default()`.
```

node/src/components/small_network/error.rs
```diff
@@ -145,16 +145,44 @@ pub enum ConnectionError {
     /// Peer reported an incompatible version.
     #[error("peer is running incompatible version: {0}")]
     IncompatibleVersion(ProtocolVersion),
-    /// Peer sent a non-handshake message as its first message.
+    /// Peer sent no or a non-handshake message as its first message.
```
Fraser999: I can't make sense of "Peer sent no message as its first message." :)

node/src/components/small_network/error.rs
```diff
@@ -145,16 +145,44 @@ pub enum ConnectionError {
...
+        io::Error,
+    ),
     /// The peer sent a consensus certificate, but it was invalid.
     #[error("invalid consensus certificate")]
     InvalidConsensusCertificate(
         #[serde(skip_serializing)]
         #[source]
         crypto::Error,
     ),
+    /// Failed to reunited handshake sink/stream.
```
Fraser999:
```suggestion
    /// Failed to reunite handshake sink/stream.
```

Fraser999 (approved):

node/src/components/small_network/message.rs
```diff
@@ -432,16 +434,22 @@ mod tests {
 
     /// Serialize a message using the standard serialization method for handshakes.
     fn serialize_message<M: Serialize>(msg: &M) -> Vec<u8> {
-        // The actual serialization/deserialization code can be found at
-        // https://github.com/carllerche/tokio-serde/blob/f3c3d69ce049437973468118c9d01b46e0b1ade5/src/lib.rs#L426-L450
+        let mut serializer = MessagePackFormat;
 
-        rmp_serde::to_vec(&msg).expect("handshake serialization failed")
+        Pin::new(&mut serializer)
```
goral09, line 439: Can you explain why we have to pin it?

node/src/components/small_network/message.rs
```diff
@@ -432,16 +434,22 @@ mod tests {
...
+            .serialize(&msg)
+            .expect("handshake serialization failed")
```
goral09, line 441: Shouldn't we return a `Result` here instead? I understand that most likely we can't recover from it but we could try to handle it "gracefully".

node/src/components/small_network/message.rs
```diff
@@ -432,16 +434,22 @@ mod tests {
 
     /// Serialize a message using the standard serialization method for handshakes.
     fn serialize_message<M: Serialize>(msg: &M) -> Vec<u8> {
-        // The actual serialization/deserialization code can be found at
-        // https://github.com/carllerche/tokio-serde/blob/f3c3d69ce049437973468118c9d01b46e0b1ade5/src/lib.rs#L426-L450
+        let mut serializer = MessagePackFormat;
```
goral09, line 437: Maybe it's worth leaving a comment here why we still have `MessagePack` for the handshake while use `bincode` for everything else.

goral09 (approved):

node/src/components/small_network/message.rs
marc-casperlabs: Because the function signature of the `tokio_serde::Serializer` trait demands it, that's the only reason.

node/src/components/small_network/message.rs
marc-casperlabs: That would be true usually, but this is testing code. The actual code has an error for this, `ConnectionError::CouldNotEncodeOurHandshake`.

node/src/components/small_network/message.rs
marc-casperlabs: This is the test module. Also, it says "standard serialization method for handshakes" right at the top!

goral09: OK, I would still be curious why we use different format for handshakes.

node/src/components/small_network/message.rs
goral09: I missed that it's a test module. I looked at the path only.

PR #2895, opened 2022-04-22
Port ScratchTrie and ScratchGlobalState to 1.4.6

dwerner:


execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,135 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+            } else {
+                // We can write this node since it has no children, or they were already written.
+                store.put(&mut txn, &digest, current_trie)?;
+            }
+        }
+
+        txn.commit()?;
+        Ok(())
+    }
+}
```
xcthulhu: @afck just double checking that we are obeying the invariant that no node is inserted before all of it's descendants are inserted.

xcthulhu: In addition, it'd be nice to have another pair of eyes to check `tries_to_visit` is not going to explode...

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,135 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+            .get(&state_root)
+            .ok_or(CommitError::TrieNotFoundInCache(state_root))?;
+
+        // Early exit if there is no work to do.
+        if !is_root_dirty {
+            return Ok(());
+        }
+
+        let mut txn = env.create_read_write_txn()?;
+        let mut tries_to_visit = vec![(state_root, root_trie, root_trie.iter_descendants())];
```
xcthulhu: Chatting with @afck , we suggest we call this `iter_children`...

execution_engine/src/storage/trie_store/lmdb.rs
```diff
@@ -161,3 +172,135 @@ impl<K, V> Store<Digest, Trie<K, V>> for LmdbTrieStore {
...
+
+        while let Some((digest, current_trie, mut descendants_iterator)) = tries_to_visit.pop() {
+            if let Some(descendant) = descendants_iterator.next() {
+                tries_to_visit.push((digest, current_trie, descendants_iterator));
+                // Only if a node is marked as dirty in the cache do we want to visit it's
+                // descendants
```
xcthulhu:
```suggestion
                // Only if a node is marked as dirty in the cache do we want to visit its
                // children
```

execution_engine/src/storage/trie_store/lmdb.rs
afck: No, looks correct and depth-first to me. :+1:

node/src/components/storage.rs
```diff
@@ -361,15 +359,15 @@ pub struct Storage {
...
-    switch_block_era_id_index: BTreeMap<EraId, BlockHash>,
+    switch_block_era_id_index: Arc<RwLock<BTreeMap<EraId, BlockHash>>>,
     /// A map of deploy hashes to hashes of blocks containing them.
-    deploy_hash_index: BTreeMap<DeployHash, BlockHash>,
+    deploy_hash_index: Arc<RwLock<BTreeMap<DeployHash, BlockHash>>>,
     /// Whether or not memory deduplication is enabled.
     enable_mem_deduplication: bool,
     /// Pool of loaded items.
-    deploy_cache: BlobCache<<Deploy as Item>::Id>,
+    deploy_cache: Arc<RwLock<BlobCache<<Deploy as Item>::Id>>>,
```
Fraser999: I'm very nervous about these multiple `RwLock`s.  Didn't we have a commit from @marc-casperlabs which reduced it to a single lock over a new struct holding all the indices?  I'd be keen to have that commit added here.

Alternatively, is it worth considering making no changes to `Storage` in this PR?  From what I recall, the main reason for these changes to storage was driven by the swap to RocksDB.  Maybe we can just avoid any changes here?

dwerner: Yeah, I'll simply revert these.

Fraser999 (approved):

dwerner:
bors merge

dwerner:
bors try

dwerner:
bors merge

dwerner:
bors merge

PR #2912, opened 2022-04-27
(release-1.4.6) trie compaction tool

dwerner:
This PR adds a simple tool that copies only state roots present in storage's blocks to a new data.lmdb file.


dwerner:
This PR is based on and blocked by https://github.com/casper-network/casper-node/pull/2895

goral09:
Do we need `Scratch*` parts in this PR as well? It seems like the compaction tool works with global state directly.

PR #2917, opened 2022-04-28
Measure time throttled in networking

### LINKED ISSUE #2914: Add metric to indicate time spent being throttled
Should add two metrics called `accumulated_outgoing_limiter_delay` and `accumulated_incoming_limiter_delay` indicating how many milliseconds were spent sleeping due to throttling

Reported by @marc-casperlabs

marc-casperlabs:
Adds a new metric that tracks how much accumulated time was taken throttling overzealous peers.

Closes #2914

### DIFF
--- a/node/src/components/small_network/limiter.rs
+++ b/node/src/components/small_network/limiter.rs
@@ -143,7 +143,7 @@ impl ClassBasedLimiter {
     /// Creates a new class based limiter.
     ///
     /// Starts the background worker task as well. Any time the limiter caused a delay,
-    /// `wait_time_sec` will be increment.
+    /// `wait_time_sec` will be incremented.
     pub(super) fn new(resources_per_second: u32, wait_time_sec: Counter) -> Self {
         let (sender, receiver) = mpsc::unbounded_channel();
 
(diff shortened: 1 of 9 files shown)


node/src/components/small_network/limiter.rs
```diff
@@ -141,14 +142,16 @@ struct ConsumerId {
 impl ClassBasedLimiter {
     /// Creates a new class based limiter.
     ///
-    /// Starts the background worker task as well.
-    pub(super) fn new(resources_per_second: u32) -> Self {
+    /// Starts the background worker task as well. Any time the limiter caused a delay,
+    /// `wait_time_sec` will be increment.
```
Fraser999:
```suggestion
    /// `wait_time_sec` will be incremented.
```

node/src/components/small_network/limiter.rs
```diff
@@ -456,6 +469,13 @@ mod tests {
         let diff = end - start;
         assert!(diff >= Duration::from_secs(5));
         assert!(diff <= Duration::from_secs(6));
+
+        // Ensure metrics recorded the correct number of seconds.
+        println!("wait_metric is {}", wait_metric.get());
```
Fraser999: Should remove this `println!` and the one on line 516.

node/src/components/small_network/metrics.rs
```diff
@@ -185,6 +190,12 @@ impl Metrics {
             "number of connections in the loopback state",
         )?;
 
+        let accumulated_outgoing_limiter_delay = Counter::new(
+            "accumulated_outgoing_limiter_delay",
+            "seconds spent delaying outgoing traffic to non-validators due to limiter, in seconds",
+        )?;
+        let accumulated_incoming_limiter_delay = Counter::new("accumulated_incoming_limiter_delay","seconds spent delaying incoming traffic from non-validators due to limiter, in seconds.")?;
```
Fraser999: Formatting seems off here?

marc-casperlabs: rustfmt abandoned me there, I think.

node/src/components/small_network/limiter.rs
marc-casperlabs: I put it in because usually something like `assert_eq!` would print a helpful message, but since I have to check for a range, it just says "failed" without giving the concrete value.

However I just realized that this could be 

```rust
assert!(wait_metric.get() <= 6.0, "metric too large: {}", wait_metric.get());
```

instead. I'll update it.

Fraser999 (approved):

Issue #2908, opened 2022-04-27
Update Storage component to use bytesrepr exclusively


devendran-m:
- all instances of `bincode` (including within error variants) should be removed
- the `lmdb_ext` traits should all have the serde methods removed and the bytesrepr ones should have the `_bytesrepr` suffix removed
- for any types which need to have `ToBytes`/`FromBytes` implemented, ensure we also implement a test-only `pub fn random(rng: &mut TestRng) -> Self` if it doesn't already exist, and use this in a `bytesrepr::test_serialization_roundtrip` test.  See `deploy::Approvals` for example.

Reported by @Fraser999

Issue #2163, opened 2021-09-27
Remove nested bincode serialization in GetRequests


piotr-dziubecki:

piotr-dziubecki:
Due to the priorities change (rocksdb) we remove this from the current sprint scope.

marc-casperlabs:
Testing changes https://github.com/casper-network/casper-test/pull/73

marc-casperlabs:
This will likely slip due to #2860 and #2866 being added.

marc-casperlabs:
As discussed with @Fraser999, we are closing this, as the underlying problem that was exacerbating this issue (msgpack encoding for network traffic) has been solved and the remaining gains will likely become obsolete with a switch to `ToBytes` encoding in 2.0 the latest. In other words, there is far less bang for our buck implementing this now.

PR #2922, opened 2022-04-29
Use bytesrepr (de)serialization in the storage component

georgepisaltu:
Partially fixes: #2908

This PR adds bytesrepr (de)serialization support for:
- `BlockHeaderWithMetadata`
- `BlockSignatures`
- `BlockWithMetadata`
- `FinalizedApprovals`
- `FinalizedApprovalsWithId`
- `DeployMetadata`
- `TrieOrChunkId`
- `GossipedAddress`

Additionally, `bincode` (de)serialization was removed from the storage component and replaced by `bytesrepr` with the exception of `decode_item_id`, which wasn't changed yet because it's not used only by the storage component.


node/src/components/small_network/gossiped_address.rs
```diff
@@ -28,6 +50,31 @@ impl Display for GossipedAddress {
...
+
+    fn write_bytes(&self, writer: &mut Vec<u8>) -> Result<(), BytesreprError> {
+        self.0.to_string().write_bytes(writer)
+    }
+}
+
+impl FromBytes for GossipedAddress {
+    fn from_bytes(bytes: &[u8]) -> Result<(Self, &[u8]), BytesreprError> {
+        let (sock_addr_str, bytes): (String, &[u8]) = FromBytes::from_bytes(bytes)?;
+        let sock_addr = std::str::FromStr::from_str(sock_addr_str.as_str())
```
rafal-ch:
```suggestion
        let sock_addr: SocketAddr = sock_addr_str.parse().map_err(|_| BytesreprError::Formatting)?;
```

node/src/types/deploy.rs
```diff
@@ -1614,6 +1677,69 @@ pub struct DeployMetadata {
...
+        }
+        Ok(())
+    }
+}
+
+// `bytesrepr::FromBytes` is not implemented for the inner `HashMap` because
+// bytesrepr is no-std and `HashMap` is only available in `std`.
+impl FromBytes for DeployMetadata {
+    fn from_bytes(bytes: &[u8]) -> Result<(Self, &[u8]), bytesrepr::Error> {
+        let (num_keys, mut stream) = u32::from_bytes(bytes)?;
```
rafal-ch: Please consider using the `usize` as a type to represent number of keys. `usize` is what's returned from `HashMap::len()`.

rafal-ch (approved):
LGTM with minor nits :+1:

node/src/components/small_network/gossiped_address.rs
georgepisaltu: Done.

node/src/types/deploy.rs
georgepisaltu: It seems `bytesrepr` (de)serialization of collection objects such as maps require the length to fit in an `u32`, hence the error `bytesrepr::Error::NotRepresentable`. As discussed, we'll keep the implementation consistent with the other code.

node/src/components/small_network/gossiped_address.rs
```diff
@@ -28,6 +50,32 @@ impl Display for GossipedAddress {
...
+        self.write_bytes(&mut buf)?;
+        Ok(buf)
+    }
+
+    fn serialized_length(&self) -> usize {
+        ToBytes::serialized_length(&self.0.to_string())
+    }
+
+    fn write_bytes(&self, writer: &mut Vec<u8>) -> Result<(), BytesreprError> {
+        self.0.to_string().write_bytes(writer)
```
Fraser999: Hmm - this is a bit suboptimal.  The string representation is likely to be more expensive in terms of computation and space.

I think it would be better to implement bytesrepr traits for the following std types in the bytesrepr module of casper-types:
* `Ipv4Addr`
* `Ipv6Addr`
* `IpAddr`
* `SocketAddrV4`
* `SocketAddrV6`
* `SocketAddr`

We should use the [`octets()`](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#method.octets) method on `Ipv4Addr` and `Ipv6Addr` to be efficient.

We should probably share a couple of const tags for the implementation of the `IpAddr` and `SocketAddr` enums - something like:
```rust
pub const IP_V4_TAG: u8 = 0;
pub const IP_V6_TAG: u8 = 1;
```

execution_engine/src/storage/trie/mod.rs
```diff
@@ -450,6 +450,31 @@ impl Display for TrieOrChunkId {
...
 
+impl ToBytes for TrieOrChunkId {
+    fn to_bytes(&self) -> Result<Vec<u8>, bytesrepr::Error> {
+        let mut buf = bytesrepr::allocate_buffer(self)?;
+        self.write_bytes(&mut buf)?;
+        Ok(buf)
+    }
+
+    fn serialized_length(&self) -> usize {
+        ToBytes::serialized_length(&self.0) + ToBytes::serialized_length(&self.1)
```
Fraser999: It's a bit more canonical to write `self.0.serialized_length()`.  Same comment applies to a few such instances in this PR.

execution_engine/src/storage/trie/mod.rs
```diff
@@ -450,6 +450,31 @@ impl Display for TrieOrChunkId {
...
+
+    fn write_bytes(&self, writer: &mut Vec<u8>) -> Result<(), bytesrepr::Error> {
+        self.0.write_bytes(writer)?;
+        self.1.write_bytes(writer)
+    }
+}
+
+impl FromBytes for TrieOrChunkId {
+    fn from_bytes(bytes: &[u8]) -> Result<(Self, &[u8]), bytesrepr::Error> {
+        let (idx, bytes) = FromBytes::from_bytes(bytes)?;
```
Fraser999: I tend to favour being explicit about the type in cases like this - mainly for readability: `u64::from_bytes(bytes)`.  Same comment for a few instances in this PR.

node/src/types/deploy.rs
```diff
@@ -1614,6 +1677,69 @@ pub struct DeployMetadata {
...
+        for (key, value) in self.execution_results.iter() {
+            key.write_bytes(writer)?;
+            value.write_bytes(writer)?;
+        }
+        Ok(())
+    }
+}
+
+// `bytesrepr::FromBytes` is not implemented for the inner `HashMap` because
+// bytesrepr is no-std and `HashMap` is only available in `std`.
```
Fraser999: I think we should implement the traits for `HashMap` in casper-types if the `std` feature is enabled in that crate.

node/src/types/deploy.rs
```diff
@@ -1614,6 +1677,69 @@ pub struct DeployMetadata {
...
+    }
+
+    fn write_bytes(&self, writer: &mut Vec<u8>) -> Result<(), bytesrepr::Error> {
+        let num_keys: u32 = self
+            .execution_results
+            .len()
+            .try_into()
+            .map_err(|_| BytesreprError::NotRepresentable)?;
+        num_keys.write_bytes(writer)?;
+        for (key, value) in self.execution_results.iter() {
```
mpapierski: Can we sort the keys first? HashMap is unordered (and in stdlib hashmap specifically order is randomized) and two instances with the same data would generate different bytes. We should sort the keys first (or create a temporary `BTreeMap<&K, &V>` or smth) to make the representation deterministic.

georgepisaltu: > two instances with the same data would generate different bytes

Is this a problem as long as deserializing those bytes yields the same result? Is there a use for the intermediary bytes representation?

dwerner: @mpapierski, are you concerned about data ending up in the trie at some point? In that case, having a different order to the bytes would affect hashing, as well as extra duplication of mutated data over time. Ideally this would be handled in the ToBytes impl for HashMap that @Fraser999 is suggesting.

dwerner (approved):
Looks good to me, approval pending @Fraser999 's suggestions being implemented.

execution_engine/src/storage/trie/mod.rs
georgepisaltu: Done.

execution_engine/src/storage/trie/mod.rs
georgepisaltu: Done.

node/src/components/small_network/gossiped_address.rs
georgepisaltu: Done.

node/src/types/deploy.rs
georgepisaltu: I'll wait for an answer here to push the change.

node/src/types/deploy.rs
georgepisaltu: Done.

.drone.yml
```diff
@@ -138,11 +138,6 @@ steps:
 
 - name: nctl-upgrade-test
   <<: *buildenv
-  environment:
```
marc-casperlabs, line 141: Is this accidental or on purpose? I am out of the loop here :)

Looks like some branching shenanigans going on?

node/src/components/storage.rs
```diff
@@ -1953,6 +1952,10 @@ fn decode_item_id<T>(raw: &[u8]) -> Result<T::Id, GetRequestError>
 where
     T: Item,
 {
+    // Transition to bytesrepr is not complete here as this function is also used in
+    // other parts of the code which we don't want to change right now. In order to
+    // fully migrate from bincode, we would have to use
+    // `bytesrepr::deserialize_from_slice` here.
     bincode::deserialize(raw).map_err(GetRequestError::MalformedIncomingItemId)
```
marc-casperlabs, line 1959: Yes, ItemIDs are bincode and this is shared across networked components, like the trie-store. Unfortunately, the IDs are raw `Vec<u8>` without a type wrapper at the moment, and have no serialization methods. Sorry.

types/src/bytesrepr.rs
```diff
@@ -670,6 +684,56 @@ where
...
+
+#[cfg(feature = "std")]
+impl<K, V> FromBytes for HashMap<K, V>
+where
+    K: FromBytes + Hash + Ord,
+    V: FromBytes,
+{
+    fn from_bytes(bytes: &[u8]) -> Result<(Self, &[u8]), Error> {
+        let (num_keys, mut stream) = u32::from_bytes(bytes)?;
+        let mut result = HashMap::new();
```
marc-casperlabs, line 728: We should use `with_capacity(usize::from(num_keys))` here, I think.

types/src/bytesrepr.rs
```diff
@@ -772,6 +836,213 @@ impl<T: FromBytes, E: FromBytes> FromBytes for Result<T, E> {
...
+        self.write_bytes(&mut buf)?;
+        Ok(buf)
+    }
+
+    fn serialized_length(&self) -> usize {
+        U16_SERIALIZED_LENGTH + IPV4_SERIALIZED_LENGTH
+    }
+
+    fn write_bytes(&self, writer: &mut Vec<u8>) -> Result<(), Error> {
+        self.port().write_bytes(writer)?;
```
marc-casperlabs: Is there a reason we are writing these in what, at least to me, is "reverse" order? The implementation is platform specific of course, but I assume most people expect these to be in "ip, port" order?

types/src/bytesrepr.rs
```diff
@@ -772,6 +836,213 @@ impl<T: FromBytes, E: FromBytes> FromBytes for Result<T, E> {
...
+        U16_SERIALIZED_LENGTH
+            + U32_SERIALIZED_LENGTH
+            + U32_SERIALIZED_LENGTH
+            + IPV6_SERIALIZED_LENGTH
+    }
+
+    fn write_bytes(&self, writer: &mut Vec<u8>) -> Result<(), Error> {
+        self.port().write_bytes(writer)?;
+        self.flowinfo().write_bytes(writer)?;
+        self.scope_id().write_bytes(writer)?;
```
marc-casperlabs: I am unsure whether we want to persist the `scope_id` and `flowinfo` parts. On the one hand, they are part of the serialization and naively I would expect for them to roundtrip correctly, on the other hand their presence on an IPv6 address seems more like an implementation detail of the underlying C struct.

`scope_id` is only valid on the local machine, as it names a local interface. Should we ever support IPv6, we probably would not be sending this across the network as a gossiped address.

`flowinfo` has meaning outside the local machine, but I wonder whether it should be persisted as well. If used, it is typically filled before sending.

My limited intuition would also opt for an "ip, port" serialization. Maybe someone with more comprehensive IPv6 experience can help out here?

marc-casperlabs (commented):
Some open questions, but I would not be overly upset if someone ignored my comments and pressed on with it :)

.drone.yml
georgepisaltu: I did a `git merge` with upstream/dev because we needed the change which undeprecated the `std` feature gate on casper-types. I don't know exactly why it's showing up like this and not a single merge commit.

node/src/components/storage.rs
georgepisaltu: The next step is to replace `bincode` with `bytesrepr` in the small network component. After that is done, I think we could change this as well.

types/src/bytesrepr.rs
georgepisaltu: Done.

`usize` has `TryFrom` instead of `From` for `u32`. Seems like 16-bit systems are still a thing.

types/src/bytesrepr.rs
georgepisaltu: No reason other than my preference of serializing primitives first. This only makes a difference in the intermediary bytes representation, which I'm still not sure it's [relevant](https://github.com/casper-network/casper-node/pull/2922#discussion_r864859094).

types/src/bytesrepr.rs
georgepisaltu: My personal preference here would be to keep `bytesrepr` implementations as strictly (de)serializing structures completely (no losses) and not actually reason about what fields are relevant or not in the implementation. I think that kind of reasoning should be done at a higher level, kind of how `serde` does it with `bincode`. This is to keep the `bytesrepr` functionality consistent and to avoid any weird corner cases in its behavior. In the absence of a higher level implementation, you would need to either:
- Create new structs which serve as a schema containing exactly what you need (de)serialized. In this particular case, we would have a
```
CustomSocketAddrV6 {
    ip: IpV6Addr,
    port: u16,
}
```
- Reason about the fields pre-serialization/post-deserialization.

types/src/bytesrepr.rs
mpapierski: I'd be careful here with `with_capacity` @marc-casperlabs - `num_keys` is potentially untrusted data, and if somehow the host tries to deserialize `u32::MAX` here on the host, it will abort the process. In `ToBytes for Vec<T>,` we use `try_vec_with_capacity` to avoid this problem.

node/src/types/deploy.rs
mpapierski: @dwerner, yes, exactly this is my concern - @georgepisaltu, we should enforce an order of keys in `ToBytes for HashMap`

types/src/bytesrepr.rs
marc-casperlabs: > usize has TryFrom instead of From for u32. Seems like 16-bit systems are still a thing.

I specifically discounted them, I don't see us supporting those any time soon. But I guess it can't hurt _much_ to use `TryFrom` and we'll be happy to have done that once we port everything to the SNES ❤️ .

@mpapierski makes a very good point that I missed though --- I forgot that there _is_ a limit since the incoming chunk of raw serialized data is likely size limited (otherwise, adding infinite items would have had the same effect). So I guess no `with_capacity` (or a compromise using a small upper bound, but at that point, we might as well not bother, since the size of the key type might affect this as well).

tl;dr my bad, my suggestions weren't exactly stellar.

types/src/bytesrepr.rs
marc-casperlabs: I usually view these things with rose colored glasses of us eventually getting to use a derive macro for these things. If that were the case, I would stick to the field order declared in struct itself, which defines at least one possible canonical ordering.

types/src/bytesrepr.rs
marc-casperlabs: That's something I can certainly support. I kinda blame the actual types tacking on these extra fields :/

My reasoning is that should we ever use a serialized representation of this in a hashed way, we'd have to zero some of the fields.

Right now, the only thing I can think of Ipv6 addresses being used for is them being sent to other nodes as part of the address gossiping the network component does, should we implement IPv6 support. Now in the past we once had signed addresses being sent to peers, and it would have been weird for the gossipped address to local-only information, which it then would have to send on unchanged to keep the signature intact.

It's certainly not the end of the world though, we'd just have to implement our own `SocketAddr` type and create a `From`-impl.

So, since I can't give a really good reason to not serialize it, let's go with your solution :)

node/src/components/storage.rs
marc-casperlabs: I think the networking component is not at fault here, but the fetcher responses and their lack of typing. I've been sitting on #2163 for too long, which should in theory fix this.

marc-casperlabs:
LGTM, sans my ill-guided suggestion that @mpapierski pointed out and will need to be fixed.

types/src/bytesrepr.rs
georgepisaltu: Reverted the commit with the change.

node/src/types/deploy.rs
georgepisaltu: @mpapierski done.

types/src/bytesrepr.rs
georgepisaltu: I refactored it.

mpapierski (approved):

PR #2924, opened 2022-04-29
Fix fetch forward

### LINKED ISSUE #2927: Fix chain sync for missed upgrade
On starting an upgraded node after missing an upgrade, the sync process tries to execute the immediate switch block of the new protocol version without running commit_upgrade.  This yields an incorrect result and causes the node to exit with an error.

Fraser999:
This PR includes multiple fixes for chain sync:

* reverts #2896 which was an attempt to fix an issue only able to occur in a misconfigured emergency-upgrade test.

  The change in that PR caused `fetch_forward` in chain sync to exit once it had downloaded the final switch block and associated state of the previous protocol version.  From that point, chain sync would try to execute forwards.  However, the first block of the new era is the immediate switch block created after the upgrade is committed, and chain sync does not attempt to call `commit_upgrade` at that point.  This causes the execution of the first block to yield invalid results and causes a fatal error.
  
  The fix is to allow `fetch_forward` to download the immediate switch block and associated state before exiting.
* ensures all deploys and global state under the trusted block hash are downloaded
* ensures chain sync takes a copy of the available block range on startup, rather than trying to retrieve it from storage once chain-syncing has already started and possibly affected it

The PR also includes some cleanups of overlooked instances of `rmp_serde` in the `small_network` component.

Closes #2927.

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -949,6 +949,7 @@ async fn sync_to_genesis(ctx: &ChainSyncContext<'_>) -> Result<(KeyBlockInfo, Bl
     let mut trusted_key_block_info = get_trusted_key_block_info(ctx).await?;
 
     let trusted_block = *fetch_and_store_block_by_hash(ctx.trusted_hash(), ctx).await?;
+    sync_deploys_and_transfers_and_state(&trusted_block, ctx).await?;
 
     fetch_to_genesis(&trusted_block, ctx).await?;
 
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -16,7 +16,7 @@ use futures::{
 };
 use prometheus::IntGauge;
 use quanta::Instant;
-use tracing::{info, trace, warn};
+use tracing::{error, info, trace, warn};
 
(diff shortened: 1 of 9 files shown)


TomVasile (approved):
Verified via nctl

fizyk20 (approved):

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1005,33 +1006,55 @@ async fn fetch_forward(
...
         // range.
-        if locally_available_block_range.contains(walkback_block.height()) {
+        if ctx
+            .locally_available_block_range_on_start
+            .contains(walkback_block.height())
+        {
             let maybe_lowest_available_block = ctx
                 .effect_builder
-                .get_lowest_available_block_from_storage()
+                .get_block_at_height_with_metadata_from_storage(
```
georgepisaltu, line 1026: I see there is no `get_block_at_height_from_storage` storage request, but we could implement it to avoid unnecessary fetching of metadata, such as this particular case. As far as I can tell, this wouldn't be a breaking change.

Fraser999: Yes, that's true.  I'll ticket it for a separate PR.

Fraser999:
bors merge

PR #2926, opened 2022-05-01
Implement backpressure for trie downloads

marc-casperlabs:
This PR adds backpressure for `TrieRequest`s by ensuring that the networking component stops reading more messages from a peer if it has too many outstanding trie fetch requests. A request is only considered finished when the response has been handed over to the kernel for sending. Since the latter also has a finite buffer for outgoing data (see `/proc/sys/net/core/wmem_default`, usually set to a few kilobytes), this will eventually limit the memory consumption of any single peer.

On the requesting/joining peer's side, this may _currently_ blow up the joiner's memory if set the number of parallel requests is set too high, but since requests are orders of magnitude smaller than the responses, this is unlikely to happen even with aggressive configuration

The value to play with/adjust is `max_in_flight_demands`, it is currently set to what I assume is a fairly large value of `50`, as it is tricky to find a good size, since tries vary in size so much.

## Implementation

The implementation has multiple parts/steps, which can be followed by looking at the commit history. Roughly:

* A new type of request is introduced called a "Demand" (the name was chosen to avoid names like "TrieRequestRequest"). Where "-Incoming" type of events are network messages that are announced, a "-Demand" is the equivalent of a request. The answer to said request is the networking message to send back.
* The necessary abstractions are added to support creating a "Demand" from a network message. They are special in that they are created directly inside the networking component, including having their receiver task spawned, before being put onto the queue.
* Handling of `TrieDemand` is added to the contract runtime (demands are almost no different in handling than the previous "-RequestIncoming" announcements).
* The `send_message` function has been changed and will only return once the message has been confirmed sent. This has no effect on any called of this message that returns it using an effect with `.ignore`. The previous behavior of respond immediately after enqueuing a message is now found in `enqueue_message`. The only call site that had to be updated was in the era supervisor when sending a message to a random peer, as it was the only one `.await`ing a `send_message` call.
* The networking component has been updated to attempt to first create a demand from any incoming network message. If successful, it passes it on to the message queue after obtaining a semaphore permit. It simultaneously spawns a handling future, which takes ownership of the permit and will schedule the resulting response for sending. If too many requests are in-flight, the `message_reader` blocks while awaiting a semaphore ticket.
* The spawned response handling future sends it, awaiting confirmation of sending (or dropping) of the message (see `send_message` changes above), before returning the semaphore permit. It thus both exerts backpressure for new requests, as well as obeying it should the remote peer not process responses quickly enough.
* At the moment, demands bypass incoming request limitations, but not bandwidth limits. Testing with outgoing bandwidth limitations increased or even removed is encouraged, as it will no longer overwhelm the node.

## Future work and remarks

* `TrieRequestIncoming` and all of its creation/handling can be removed, which should lead to a refactoring of the `FromIncoming` interface as well to return either a demand or a non-demand from one function.
* More types of "Demand"s can and should be implemented, i.e. for blocks and deploys.
* We may need to add a global demand limit in addition to the per-peer demand limit to curb sybil attacks.
* If expanded, a more thorough analysis on whether or not nodes can get stuck via mutual blockade is prudent --- right now, there will never be more than one node requesting tries, thus we will always make progress.
* Automatic throttling that takes advantage of now existing backpressure in the joiner should be added to make the entire system self-adjusting.
* **Consider changing `get_from_peer_timeout` to a higher value like 30 seconds** when testing (and potentially shipping). Since fetchers have been responding with negative ACKs for a while now, this artificially constrains the process.


goral09 (approved):
Neat

rafal-ch (approved):
LVGTM :+1:

PR #2948, opened 2022-05-03
release-1.4.6 trie-compact tool

dwerner:


utils/trie-compact/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use retrieve_state::{
+    storage,
+    storage::{create_storage, normalize_path},
+};
+
+use tracing::info;
+use trie_compact::copy_state_root;
+
+#[derive(Debug, StructOpt)]
```
Fraser999, line 14: I think if this is a user-facing tool, we need to significantly enhance the help messages.  I'm not sure whether each arg refers to a file or folder, whether they're expecting to read and/or write at the given paths.

utils/trie-compact/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    let opts = Opts::from_args();
+    env_logger::init();
+
+    let storage_path = normalize_path(&opts.storage_path)?;
+    let source_lmdb_path = normalize_path(&opts.source_lmdb_path)?;
+    let destination_lmdb_path = normalize_path(&opts.destination_lmdb_path)?;
```
Fraser999, line 33: Seems like this fails if the folder doesn't already exist, which is probably not the desired behaviour?

Either way, since we're using `anyhow`, it would be really helpful to tack on some context to usages of `?` throughout this tool, e.g. the path which failed.

utils/trie-compact/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+    let (source_state, _env) = storage::load_execution_engine(
+        source_lmdb_path,
+        retrieve_state::DEFAULT_MAX_DB_SIZE,
+        Digest::default(),
+        true,
+    )?;
+
+    let (destination_state, _env) = storage::create_execution_engine(
+        destination_lmdb_path,
+        retrieve_state::DEFAULT_MAX_DB_SIZE,
```
Fraser999, line 51: Is that going to be large enough?  Maybe this should be a config option.

utils/trie-compact/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+        true,
+    )?;
+
+    // Create a separate lmdb for block/deploy storage at chain_download_path.
+    let storage = create_storage(&storage_path).expect("should create storage");
```
Fraser999, line 56: I guess we should stick with returning anyhow errors.

utils/trie-compact/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+
+    let mut block = storage.read_highest_block().unwrap().unwrap();
```
Fraser999, line 58: Same here - especially useful if we add context to each of the two potential failure points.  And for `read_block_by_height` below.

Fraser999 (commented):
The PR could use a description and a linked ticket.

No blockers in the lib, I just struggled to use the binary.

Also, if this is meant to be used by node operators then I think we should add tests to validate the behaviour, both in terms of checking useless data in a test trie isn't carried over to the destination trie, and in terms of the binary's handling of paths (test with missing paths/existing paths where there shouldn't be, providing same src and dest, etc.).  Mainly want to be sure we don't somehow accidentally corrupt a user's existing DB.

dwerner:
closing in favor of https://github.com/casper-network/casper-db-utils/pull/11

PR #2966, opened 2022-05-06
NCTL: use staged chainspecs for remote

TomVasile:
PR introduces using remote chainspecs for pre-upgrade protocol versions.

Note: Currently only 1.4.6 and 1.3.0 have the files necessary. Going forward these files are uploaded on push to release branches or tags.


TomVasile:
bors try

Fraser999 (approved):

PR #2967, opened 2022-05-06
Fix unbonding purses serialization

### LINKED ISSUE #2964: Fix deserialization of legacy UnbondingPurses in storage
The `UnbondingPurse` type had a new field `new_validator: Option<PublicKey>` added in recent changes.  The old version of this type has already been stored in the storage DB as part of `ExecutionResult`s (deploy metadata).  If we try to parse the old encoded data using the new struct's method, we usually get a failure.

rafal-ch:
This PR makes the `Storage` component aware that the `UnbondingPurses` struct has two different flavors:
1) Legacy one, i.e. without the `new_validator` field
2) Extended one, i.e. with the `new_validator` field

When retrieving the `UnbondingPurses` from DB, the `Storage` will always return the extended `UnbondingPurses` providing the conversion from the legacy struct on the fly, setting the `new_validator` field to `None`.

Closes #2964

### DIFF
--- a/node/src/components/storage/lmdb_ext.rs
+++ b/node/src/components/storage/lmdb_ext.rs
@@ -164,6 +164,10 @@ where
     }
 }
 
+/// Serializes `value` into the buffer.
+/// In case the `value` is of the `UnbondingPurse` type it uses the specialized
+/// function to provide compatibility with the legacy version of the `UnbondingPurse` struct.
+/// See [`serialize_unbonding_purse`] for more details.
 pub(crate) fn serialize_internal<V: 'static + Serialize>(
     value: &V,
 ) -> Result<Vec<u8>, LmdbExtError> {
@@ -175,6 +179,10 @@ pub(crate) fn serialize_internal<V: 'static + Serialize>(
     Ok(buffer)
 }
 
+/// Deerializes an object from the raw bytes.
+/// In case the expected object is of the `UnbondingPurse` type it uses the specialized
+/// function to provide compatibility with the legacy version of the `UnbondingPurse` struct.
+/// See [`deserialize_unbonding_purse`] for more details.
 pub(crate) fn deserialize_internal<V: 'static + DeserializeOwned>(
     raw: &[u8],
 ) -> Result<Option<V>, LmdbExtError> {
@@ -245,6 +253,13 @@ fn is_legacy(raw: &[u8]) -> bool {
 }
 
 /// Deserializes `UnbondingPurse` from a buffer.
+/// To provide backward compatibility with the previous version of the `UnbondingPurse`,
+/// it checks if the raw bytes stream begins with "magic bytes". If yes, the magic bytes are
+/// stripped and the struct is deserialized as a new version. Otherwise, the raw bytes
+/// are treated as bytes representing the legacy `UnbondingPurse` and deserialized accordingly.
+/// In order for the latter scenario to work, the raw bytes stream is extended with
+/// bytes that represent the `None` serialized with `bincode` - these bytes simulate
+/// the existence of the `new_validator` field added to the `UnbondingPurse` struct.
 pub(super) fn deserialize_unbonding_purse<T: DeserializeOwned>(
     raw: &[u8],
 ) -> Result<T, LmdbExtError> {
@@ -263,6 +278,9 @@ pub(super) fn serialize<T: Serialize>(value: &T) -> Result<Vec<u8>, LmdbExtError
 }
 
 /// Serializes `UnbondingPurse` into a buffer.
+/// To provide backward compatibility with the previous version of the `UnbondingPurse`,
+/// the serialized bytes are prefixed with the "magic bytes", which will be used by the
+/// deserialization routine to detect the version of the `UnbondingPurse` struct.
 #[inline(always)]
 pub(super) fn serialize_unbonding_purse<T: Serialize>(value: &T) -> Result<Vec<u8>, LmdbExtError> {
     let mut serialized = UNBONDING_PURSE_V2_MAGIC_BYTES.to_vec();
--- a/node/src/components/storage/lmdb_ext.rs
+++ b/node/src/components/storage/lmdb_ext.rs
@@ -22,7 +22,7 @@ use casper_types::{
     EraId, PublicKey, URef, U512,
 };
 
-pub const UNBONDING_PURSE_V2_MAGIC_BYTES: &[u8] = &[121, 17, 133, 179, 91, 63, 69, 222];
+const UNBONDING_PURSE_V2_MAGIC_BYTES: &[u8] = &[121, 17, 133, 179, 91, 63, 69, 222];
 
 /// Error wrapper for lower-level storage errors.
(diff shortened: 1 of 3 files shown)


node/src/components/storage/lmdb_ext.rs
```diff
@@ -211,12 +240,54 @@ pub(super) fn deserialize<T: DeserializeOwned>(raw: &[u8]) -> Result<T, LmdbExtE
...
+        let upgraded_unbonding_purse = UnbondingPurse::new(
+            deserialized.bonding_purse,
+            deserialized.validator_public_key,
+            deserialized.unbonder_public_key,
+            deserialized.era_of_creation,
+            deserialized.amount,
+            Option::default(),
+        );
+        let raw = bincode::serialize(&upgraded_unbonding_purse).unwrap();
+        bincode::deserialize(&raw).map_err(|err| LmdbExtError::DataCorrupted(Box::new(err)))
```
rafal-ch: Note to self:
This roundtrip should go away.

node/src/components/storage/lmdb_ext.rs
```diff
@@ -211,12 +240,54 @@ pub(super) fn deserialize<T: DeserializeOwned>(raw: &[u8]) -> Result<T, LmdbExtE
...
+    } else {
+        deserialize(&raw[UNBONDING_PURSE_V2_MAGIC_BYTES.len()..])
+    }
```
Fraser999: I had this idea to get round the double deserialize issue:
```suggestion
    const BINCODE_ENCODED_NONE: [u8; 4] = [0; 4];
    if is_legacy(raw) {
        deserialize(&[raw, &BINCODE_ENCODED_NONE].concat())
    } else {
        deserialize(&raw[UNBONDING_PURSE_V2_MAGIC_BYTES.len()..])
    }
```

node/src/components/storage/tests.rs
```diff
@@ -2071,3 +2074,55 @@ fn should_get_block_header_by_height() {
...
+#[test]
+fn should_read_legacy_unbonding_purse() {
+    // These bytes represent the `UnbondingPurse` struct with the `new_validator` field removed
+    // and serialized with `bincode`.
+    // In theory, we can generate these bytes by serializing the `WithdrawPurse`, but at some point,
+    // these two structs may diverge and it's a safe bet to rely on the bytes
+    // that are consistent with what we keep in the current storage.
+    const LEGACY_BYTES: &str = "0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e0e07010000002000000000000000197f6b23e16c8532c6abc838facd5ea789be0c76b2920334039bfa8b3d368d610100000020000000000000004508a07aa941707f3eb2db94c8897a80b2c1197476b6de213ac273df7d86c4ffffffffffffffffff40feffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+
+    let decoded = hex::decode(LEGACY_BYTES).expect("decode");
```
Fraser999: We normally use the `base16` crate for this - that will avoid you importing `uint::hex`.

node/src/components/storage/lmdb_ext.rs
rafal-ch: Nice one! :+1: 
Fixed in fa65259d147e1213322d2c4638e7328a26f305d1

node/src/components/storage/lmdb_ext.rs
rafal-ch: It's gone thanks to https://github.com/casper-network/casper-node/pull/2967#discussion_r866926169

node/src/components/storage/tests.rs
rafal-ch: Fixed in 0c32238bf66d3d14721e45c9c88d31073af8e9c3

node/src/components/storage/lmdb_ext.rs
```diff
@@ -10,11 +10,18 @@
...
 use serde::{de::DeserializeOwned, Serialize};
 use thiserror::Error;
 
-use casper_types::bytesrepr::{self, FromBytes, ToBytes};
+use casper_types::{
+    bytesrepr::{self, FromBytes, ToBytes},
+    system::auction::UnbondingPurse,
+};
+
+const UNBONDING_PURSE_V2_MAGIC_BYTES: &[u8] = &[121, 17, 133, 179, 91, 63, 69, 222];
```
afck, line 24: Is there a reason why a serialized V1 purse can never start with those bytes?

afck (approved):

node/src/components/storage/lmdb_ext.rs
rafal-ch: Unfortunately not.
The `UnbondingPurse` starts with an `URef`, which starts with an `URefAddr`, which is just a `[u8; 32]` that could be anything.
I used 8 magic bytes as an effort to minimize the chances for a collision.

Fraser999 (approved):

node/src/components/storage/lmdb_ext.rs
```diff
@@ -211,12 +247,47 @@ pub(super) fn deserialize<T: DeserializeOwned>(raw: &[u8]) -> Result<T, LmdbExtE
...
+/// bytes that represent the `None` serialized with `bincode` - these bytes simulate
+/// the existence of the `new_validator` field added to the `UnbondingPurse` struct.
+pub(super) fn deserialize_unbonding_purse<T: DeserializeOwned>(
+    raw: &[u8],
+) -> Result<T, LmdbExtError> {
+    const BINCODE_ENCODED_NONE: [u8; 4] = [0; 4];
+    if is_legacy(raw) {
+        deserialize(&[raw, &BINCODE_ENCODED_NONE].concat())
+    } else {
+        deserialize(&raw[UNBONDING_PURSE_V2_MAGIC_BYTES.len()..])
```
mpapierski, line 270: ~~This will panic if `raw.len() < UNBONDING_PURSE_V2_MAGIC_BYTES.len()`~~ if `is_legacy` returns false then `raw.len() >= UNBONDING_PURSE_V2_MAGIC_BYTES.len()`

node/src/components/storage/lmdb_ext.rs
```diff
@@ -157,15 +164,44 @@ where
...
+) -> Result<Vec<u8>, LmdbExtError> {
+    let buffer = if TypeId::of::<UnbondingPurse>() == TypeId::of::<V>() {
+        serialize_unbonding_purse(value)?
+    } else {
+        serialize(value)?
+    };
+    Ok(buffer)
+}
+
+/// Deerializes an object from the raw bytes.
```
mpapierski:
```suggestion
/// Deserializes an object from the raw bytes.
```

rafal-ch: Resolved in 8100ff7148b1f8eb4dd55aef8f878be2995bbdc9

PR #2979, opened 2022-05-12
Clear visited trie cache after transitioning to participating reactor.

goral09:
When joining, we're caching set of visited trie nodes to speed up the `missing_trie_keys` function call. After transitioning to a participator reactor we don't need that cache anymore and it can be safely cleared.

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -152,6 +152,12 @@ impl EngineState<LmdbGlobalState> {
             .put_stored_values(CorrelationId::new(), state_root_hash, stored_values)
             .map_err(Into::into)
     }
+
+    /// Clears the cache of visited descendants.
+    /// NOTE: Should be called only after node transition to participating mode.
+    pub fn clear_visited_tries_cache(&self) -> Result<(), Error> {
+        self.state.clear_visited_tries_cache().map_err(Into::into)
+    }
 }
 
 impl<S> EngineState<S>
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -154,7 +154,6 @@ impl EngineState<LmdbGlobalState> {
     }
 
     /// Clears the cache of visited descendants.
-    /// NOTE: Should be called only after node transition to participating mode.
     pub fn clear_visited_tries_cache(&self) -> Result<(), Error> {
         self.state.clear_visited_tries_cache().map_err(Into::into)
     }
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -154,8 +154,8 @@ impl EngineState<LmdbGlobalState> {
     }
 
     /// Clears the cache of visited descendants.
-    pub fn clear_visited_tries_cache(&self) -> Result<(), Error> {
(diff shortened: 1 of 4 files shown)


execution_engine/src/core/engine_state/mod.rs
```diff
@@ -152,6 +152,12 @@ impl EngineState<LmdbGlobalState> {
             .put_stored_values(CorrelationId::new(), state_root_hash, stored_values)
             .map_err(Into::into)
     }
+
+    /// Clears the cache of visited descendants.
+    /// NOTE: Should be called only after node transition to participating mode.
```
afck: I'd actually remove that line: It describes internals of a completely different crate, and I think it's self-explanatory what clearing the cache does.
Also, it's not _dangerous_ to call this elsewhere, it just makes things slower, right?

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -111,6 +111,16 @@ impl LmdbGlobalState {
         )
         .map_err(Into::into)
     }
+
+    /// Clears the cache of visited descendants.
+    /// NOTE: Should be called only after node transition to participating mode.
```
afck: (See other comment.)

execution_engine/src/storage/global_state/lmdb.rs
```diff
@@ -111,6 +111,16 @@ impl LmdbGlobalState {
...
+    pub fn clear_visited_tries_cache(&self) -> Result<(), error::Error> {
```
afck: Why return a result at all, if it can't fail?

node/src/reactor/participating.rs
```diff
@@ -537,6 +537,9 @@ impl reactor::Reactor for Reactor {
             node_startup_instant,
         } = config;
 
+        // Clear the visited trie cache. It's not needed in the participator reactor anymore.
```
afck:
```suggestion
        // Clear the visited trie cache. It's not needed in the participating reactor anymore.
```

afck (approved):

execution_engine/src/storage/global_state/lmdb.rs
goral09: It can fail though on `write()` call on `RwLock`.

execution_engine/src/core/engine_state/mod.rs
goral09: Makes sense, thanks for the comment.

node/src/reactor/participating.rs
goral09: Done in https://github.com/casper-network/casper-node/pull/2979/commits/a2143c7e469be9f625d92734616a2f355d4588a0.

execution_engine/src/storage/global_state/lmdb.rs
goral09: Done in https://github.com/casper-network/casper-node/pull/2979/commits/a2143c7e469be9f625d92734616a2f355d4588a0.

execution_engine/src/core/engine_state/mod.rs
goral09: Done in https://github.com/casper-network/casper-node/pull/2979/commits/a2143c7e469be9f625d92734616a2f355d4588a0.

afck (approved):

PR #2984, opened 2022-05-12
(feat-rocksdb) Remove Transaction and TransactionSource

dwerner:
This PR addresses tech debt remaining in feat-rocksdb by removing the Transaction and TransactionSource traits, refactoring affected test code.


execution_engine/src/storage/trie_store/db.rs
```diff
@@ -200,16 +145,215 @@ impl Store<Digest, Trie<Key, StoredValue>> for ScratchTrieStore {
     }
 }
 
-impl TrieStore<Key, StoredValue> for ScratchTrieStore {}
+/// Represents the state of a migration of a state root from lmdb to rocksdb.
+pub enum RootMigration {
+    /// Has the migration been left incomplete
+    NotMigrated,
+    /// Has the migration been not yet been started or completed
+    Incomplete,
```
Fraser999: Maybe [`Ongoing` or `Partial`] and `NotStarted` would be clearer than `NotMigrated` and `Incomplete` respectively?

Fraser999 (approved):
One non-blocking suggestion, otherwise LGTM.

execution_engine/src/storage/trie_store/db.rs
dwerner: Applied in https://github.com/casper-network/casper-node/pull/2984/commits/59622e0dc21b45405b6acca64a39900ad6fef28e

execution_engine/src/storage/store/mod.rs
```diff
@@ -5,30 +5,36 @@ pub(crate) mod tests;
...
+/// Base trait for db operations that can raise an error.
+pub trait ErrorSource: Sized {
+    /// Type of error held by the trait.
     type Error: From<bytesrepr::Error>;
+}
 
-    /// Underlying store type.
-    type Handle;
+/// Represents a store that can be read from.
+pub trait Readable: ErrorSource {
```
goral09, line 16: Why did you change from type member to inheritence?

execution_engine/src/storage/trie_store/db.rs
```diff
@@ -2,76 +2,48 @@
...
-    fn handle(&self) -> Self::Handle {
-        self.db
-    }
-}
+/// Column family name for tracking progress of data migration.
+const LMDB_MIGRATED_STATE_ROOTS_COLUMN_FAMILY: &str = "lmdb_migrated_state_roots_column";
 
-impl<K, V> TrieStore<K, V> for LmdbTrieStore {}
+/// Working set column famliy.
+const WORKING_SET_COLUMN_FAMILY: &str = "working_set";
```
goral09: This is used only in a method that isn't used itself. Also, this change isn't required for _removal of `Trasnaction` and `TransactionSource`_

execution_engine/src/storage/trie_store/db.rs
```diff
@@ -2,76 +2,48 @@
 
 use std::{
     collections::HashMap,
+    path::{Path, PathBuf},
     sync::{Arc, Mutex},
 };
 
-use casper_types::{bytesrepr, Key, StoredValue};
-use lmdb::{Database, DatabaseFlags};
+use rocksdb::{
```
goral09, line 9: There's a comment in line 1 that mentions LMDB.

node/src/components/contract_runtime/config.rs
```diff
@@ -1,68 +1,29 @@
...
     /// Defaults to 5.
     max_query_depth: Option<u64>,
-    /// Enable synchronizing to disk only after each block is written.
-    ///
-    /// Defaults to `false`.
-    enable_manual_sync: Option<bool>,
 }
 
 impl Config {
-    pub(crate) fn max_global_state_size(&self) -> usize {
```
goral09, line 37: RocksDB doesn't need this config?

node/src/reactor/participating.rs
```diff
@@ -733,20 +733,24 @@ impl reactor::Reactor for Reactor {
...
                             "unable to retrieve parent block at height {} for migration to rocksdb {:?}",
                             height, err,
                         );
                         continue;
                     }
                 };
+                info!(
+                    "Will migrate state root {:#?} at height {}",
+                    state_roots, height
+                );
```
goral09: if this is using the tracing crate then it should be:
```rust
info!(?state_roots, ?height, "will migrate state root at height")
```

execution_engine/src/storage/store/mod.rs
dwerner: This was needed to handle everywhere that Transaction's Self::Error type was referenced. I tried adding Error as an associated type but it wasn't as compatible with our existing top level functions (`write`, `put_stored_values` etc) trait bounds, so I settled on this approach.

node/src/components/contract_runtime/config.rs
dwerner: Correct.

node/src/reactor/participating.rs
dwerner: Thanks. And now that I review this as well, this is actually logging the entire debug representation of the state_roots vec, so I have changed this to only log the exact state root.

dwerner: Fixed in 213bf27a50f00424def44d1557567f3f3d184e21

execution_engine/src/storage/trie_store/db.rs
dwerner: Good catch. Fixed in 213bf27a50f00424def44d1557567f3f3d184e21

execution_engine/src/storage/trie_store/db.rs
dwerner: Fixed in 213bf27a50f00424def44d1557567f3f3d184e21

goral09 (approved):

PR #2986, opened 2022-05-16
Replace warp-json-rpc with own implementation

### LINKED ISSUE #2911: Remove warp_json_rpc
On the back of #1514, and having looked at alternatives, it seems like it would be better to remove our dependency on `warp_json_rpc` and hand roll our own solution.

This will result in:
* simpler code
* easier maintenance
* a fix for the issue reported in #1514
* being able to selectively enable or disable specific RPC endpoints trivially
* become fully JSON-RPC compliant in all error cases

Fraser999:
This PR removes the dependency on the `warp-json-rpc`, replacing it with a new crate introduced here: `casper-json-rpc`.

While there are minimal external changes to the JSON-RPC server's API, and zero changes to the schema, the PR includes the following:
1. full compliance with [the JSON-RPC specification](https://www.jsonrpc.org/specification).  Previously, due to limitations in `warp-json-rpc`, there was at least one instance of an error condition where the node was incapable of returning the correct error code.  This is no longer the case.  It also required changing the values for the custom error codes provided in the node, as previously these were in the reserved range as per the spec.  These changes should be communicated to downstream consumers.
2. for error cases where no JSON-RPC response should be returned, the HTTP response's body is now JSON-encoded, and in some cases provides a more useful message to help the caller identify the reason for the rejection than is currently the case
3. full test coverage of the json-rpc crate, and also the error paths of the node's RPC traits
4. the ability for specific RPCs to be enabled or disabled via e.g. config options passed during initialization of the server
5. somewhat reduced build times due to the reduction in complexity of the combined warp filters (the biggest single change in that regard is moving dispatching of requests to a map search rather than via multiple warp `or` combinators)
6. removal of the cause of the `request body already taken in previous filter` error message, frequently reported in the mainnet logs
7. many error conditions return more detailed error messages to the client in the JSON-RPC response

Closes #2911.


node/src/components/rpc_server/rpcs.rs
```diff
@@ -6,69 +6,29 @@ pub mod account;
...
-    fn from(error: anyhow::Error) -> Self {
-        Error(error.to_string())
-    }
-}
+pub use error_code::ErrorCode;
 
 /// A JSON-RPC requiring the "params" field to be present.
-pub trait RpcWithParams {
+#[async_trait]
+pub(super) trait RpcWithParams {
```
goral09, line 31: Maybe we can move that to `casper-json-rpc`? The trait seems rather generic and is specific to JSON RPC server.

Fraser999: Yes, I toyed with that too.  It would be nice since that's about the only place users have to worry about returning a `ReservedErrorCode` (other than if they want to return an internal server error), but I decided against that since I think we'd need to split the current trait up a bit as it has the node-specific `EffectBuilder<REv>` in some of the methods.

We could definitely look to do that in the future, but I did want to try and keep the changes to the existing node codebase as limited as possible in this PR.

json_rpc/README.md
```diff
@@ -0,0 +1,107 @@
...
+        async move { Ok::<_, Infallible>(svc.clone()) }
+    });
+
+    hyper::Server::bind(&([127, 0, 0, 1], 3030).into())
+        .serve(make_svc)
+        .await
+        .unwrap();
+}
+```
+
```
rafal-ch, line 64: Consider adding here the exemplary usage, like so:

```
curl -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"id","method":"get"}' http://127.0.0.1:3030/rpc
```

gives

```
{"jsonrpc":"2.0","id":"id","result":"got it"}
```

json_rpc/src/filters/tests/base_filter_with_recovery_tests.rs
```diff
@@ -0,0 +1,220 @@
...
+        .map(reply) // return an empty body on success
+        .with(warp::cors().allow_origin("http://a.com"))
+        .recover(handle_rejection) // or convert a rejection to JSON-encoded `ResponseBody`
+        .boxed()
+}
+
+fn valid_base_filter_request_builder() -> RequestBuilder {
+    warp::test::request()
+        .path(&format!("/{}", PATH))
+        .header("content-type", "application/json")
```
rafal-ch: Can we use this const here?
https://github.com/casper-network/casper-node/pull/2986/files#diff-c75bad1c05e3a86874006a67db59028158405a0a29f97c94b9817bc9c8d028f9R32

json_rpc/src/request.rs
```diff
@@ -0,0 +1,208 @@
...
+            error,
+            Error::new(
+                ReservedErrorCode::InvalidRequest,
+                "'id' should be a string or integer"
+            )
+        );
+    }
+
+    #[test]
+    fn should_fail_to_validate_id_with_fractional_part() {
```
rafal-ch, line 160: Nice! I like you tested all that.

node/src/components/rpc_server/config.rs
```diff
@@ -7,6 +7,8 @@ use serde::{Deserialize, Serialize};
 const DEFAULT_ADDRESS: &str = "0.0.0.0:0";
 /// Default rate limit in qps.
 const DEFAULT_QPS_LIMIT: u64 = 100;
+/// Default max body bytes.
+const DEFAULT_MAX_BODY_BYTES: u32 = 2_621_440;
```
rafal-ch, line 12: Can we consider adding an explanation of where this number came from?

node/src/components/rpc_server/rpcs.rs
goral09: Got it, this makes sense.

json_rpc/src/error.rs
```diff
@@ -0,0 +1,273 @@
...
+///
+/// See [the JSON-RPC Specification](https://www.jsonrpc.org/specification#error_object) for further
+/// details.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Error {
+    /// A number that indicates the error type that occurred.
+    code: i64,
+    /// A short description of the error.
+    message: Cow<'static, str>,
```
goral09, line 103: Can't we use plain `&' static str` here? As far as I can see, we're always using `Cow::Borrowed`.

json_rpc/src/request.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// A request which has been validated as conforming to the JSON-RPC specification.
+pub(crate) struct Request {
+    pub id: Value,
+    pub method: String,
+    pub params: Option<Value>,
+}
+
+/// Returns `Ok` if `id` is a String, Null or a Number with no fractional part.
+fn is_valid(id: &Value) -> Result<(), Error> {
```
goral09, line 43: maybe we can put that in its own `mod id` module? Its API looks like it can be used on any `Value` type. At the same time, it's a local function not exported anywhere else so it's not that dangerous.

json_rpc/src/response.rs
```diff
@@ -0,0 +1,108 @@
...
+
+/// A JSON-RPC response.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
+#[serde(deny_unknown_fields, untagged)]
+pub enum Response {
+    /// A successful RPC execution.
+    Success {
+        /// The JSON-RPC version field.
+        #[serde(deserialize_with = "deserialize_jsonrpc_field")]
+        jsonrpc: Cow<'static, str>,
```
goral09, line 20: It's always going to be the same constant, can't we use `&'static str`?

json_rpc/src/response.rs
```diff
@@ -0,0 +1,108 @@
...
+        match &self {
+            Response::Success { id, .. } | Response::Failure { id, .. } => id,
+        }
+    }
+}
+
+fn deserialize_jsonrpc_field<'de, D: Deserializer<'de>>(
+    _deserializer: D,
+) -> Result<Cow<'static, str>, D::Error> {
+    Ok(Cow::Borrowed(JSON_RPC_VERSION))
```
goral09, line 107: Are we really deserializing if we ignore the input and always return fixed value?

json_rpc/src/filters.rs
```diff
@@ -0,0 +1,211 @@
...
+                Response::new_failure(malformed.id, error)
+            } else if serde_json::from_slice::<Value>(&*body).is_ok() {
+                debug!(%error, "got json, but not a request");
+                let error = Error::new(ReservedErrorCode::InvalidRequest, error.to_string());
+                Response::new_failure(Value::Null, error)
+            } else {
+                debug!(%error, "got bad json");
+                let error = Error::new(ReservedErrorCode::ParseError, error.to_string());
+                Response::new_failure(Value::Null, error)
+            }
```
goral09, line 110: Do we need this granularity ?

json_rpc/src/filters.rs
```diff
@@ -0,0 +1,211 @@
...
+    } else if let Some(rejection) = error.find::<warp::filters::cors::CorsForbidden>() {
+        trace!("{:?}", rejection);
+        message = rejection.to_string();
+        code = StatusCode::FORBIDDEN;
+    } else {
+        // We should handle all rejection types before this.
+        warn!(?error, "unhandled warp rejection in json-rpc server");
+        message = format!("Internal server error: unhandled rejection: {:?}", error);
+        code = StatusCode::INTERNAL_SERVER_ERROR;
+    }
```
goral09, line 205: Looks like a lot of repetition - can't we add a common trait (if we can't use `warp`'s `Reject`) that will include `code` and `message`?

json_rpc/src/filters.rs
```diff
@@ -0,0 +1,211 @@
...
+///
+/// The reply will normally be built from a JSON-RPC response (which could be a success or failure).
+///
+/// However, the reply could be built from a [`Rejection`] if the request is a Notification as per
+/// the JSON-RPC specification, i.e. the request doesn't contain an "id" field.  In this case, no
+/// JSON-RPC response is sent to the client, only an HTTP response.
+pub fn main_filter(handlers: RequestHandlers) -> BoxedFilter<(WithStatus<reply::Json>,)> {
+    body::bytes()
+        .and_then(move |body| {
+            let handlers = handlers.clone();
```
goral09, line 127: This is cloning whole `HashMap` and all of its handlers, isn't it? Or it isn't b/c internality it's an `Arc` 🤔

node/src/components/rpc_server/rpcs.rs
```diff
@@ -209,65 +172,276 @@ pub trait RpcWithOptionalParams {
...
-            .or(with_invalid_params)
-            .unify()
-            .boxed()
+        handlers_builder: &mut RequestHandlersBuilder,
+    ) {
+        let handler = move |maybe_params| async move {
+            let params = Self::try_parse_params(maybe_params)?;
+            Self::do_handle_request(effect_builder, api_version, params).await
+        };
+        handlers_builder.register_handler(Self::METHOD, Arc::new(handler))
```
goral09, line 208: if I'm not mistaken this function is common accross all traits - meaning there _could_ be a trait they all have to implement that has this implementation as a deafult.

node/src/components/rpc_server/rpcs/state.rs
```diff
@@ -1273,14 +1168,9 @@ pub(super) async fn get_state_root_hash_and_optional_header<REv: ReactorEventT>(
...
             {
                 None => {
-                    let error_msg = format!(
-                        "{} failed to retrieve block header at height {}",
-                        method_name, block_height
-                    );
-                    Err(warp_json_rpc::Error::custom(
-                        ErrorCode::NoSuchBlock as i64,
-                        error_msg,
-                    ))
```
goral09, line 1283: I haven't reviewed every single line in this file super-closely but it all seems to be very mechanical work adapting the old API to a new one.

goral09 (approved):
I was not able to find anything wrong with the code :) nice job!

json_rpc/src/error.rs
Fraser999: I did look at that initially, but the `'static` lifetime becomes an issue when we derive `Deserialize`.  I could have worked around this by manually implementing the `Deserialize` trait, but it would have been a _lot_ of boilerplate for little to no gain I think.

json_rpc/src/request.rs
Fraser999: That's actually how `warp_json_rpc` handles it (i.e. it provides an `Id` enum) but the downside is that parsing from an incoming JSON-RPC request where the "id" field is invalid in the warp filter will fail if we use a strict enum.  This causes a warp rejection, which _can_ be handled later in the flow, but the useful info about why parsing failed is lost.

I prefer this more permissive parsing, so we can handle failures by returning useful data to the user in the error of the JSON-RPC response.

json_rpc/src/response.rs
Fraser999: Same reason as for the `Error` type :)

json_rpc/src/filters.rs
Fraser999: I think we do - at least that's my reading of [the error codes' meanings in the spec](https://www.jsonrpc.org/specification#error_object).

json_rpc/src/filters.rs
Fraser999: Unfortunately I think this about our best option of we want to use the `to_string()` method and extract the right error code.  `Rejection` doesn't provide any way to extract that without converting to a specific different rejection type, so we'd end up having to implement the new trait for all the various types above.  Seems like we'd just be moving the repetitive code from one place to another.

node/src/components/rpc_server/rpcs.rs
Fraser999: The `register_as_handler` is slightly different across the three traits, so I'm not sure it's worthwhile?  We'd still have to implement the function in these three different ways.

node/src/components/rpc_server/rpcs/state.rs
Fraser999: Yes, there should be no externally-observable changes to the actual logic inside the specific RPC handlers.

json_rpc/src/error.rs
goral09: Thanks.

json_rpc/src/filters.rs
goral09: I still prefer to put responsibility on the developer that adds a new `Reject` variant. I think it provides better type-safety as, for example here, one could easily omit a new variant and end up in the `else` clause. My rule of thumb is to avoid `if`s if possible 😝

json_rpc/src/response.rs
goral09: Thanks, but I am now wondering - doesn't it effectively ignore the JSON RPC version set by the client?

rafal-ch (approved):
LGTM :+1:

Fraser999:
bors merge

PR #2988, opened 2022-05-16
Don't query local storage unnecessarily.

goral09:
Previous PR introduced an explicit read from a local storage before using `Fetcher`'s API. Unfortunately, this creates unnecessary event requests in the fetcher itself since fetchers will always first query storage and only in case of item absence fallback to querying peers. 

This PR adds a boolean flag to `FetcherRequest` that indicates whether local storage should be queried before requesting the data from a peer.


goral09:
Closing for now - won't do.

PR #2998, opened 2022-05-18
Increase CHUNK_SIZE_BYTES to 8MiB to avoid chunking tries unnecessarily

Fraser999:
This is a change which was overlooked when the final decision on the temp chainspec value of `max_stored_value_size` was made in node 1.4.5.

This PR also reverts bfa3a02 since the recent work on networking backpressure makes the previous global form of `max_parallel_deploy/trie_fetches` more appropriate.


goral09 (approved):

Fraser999:
bors merge

PR #2999, opened 2022-05-18
(feat-rocksdb) Working set prep - refactor RocksDbStore into DbStore and RocksDbTrieStore

dwerner:
This PR is intended to refactor some database access code to be more general and reusable. Other work on working-set will be based on this.


execution_engine/src/storage/db_store.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    /// Working set column family.
+    pub fn working_set_column_family(&self) -> Result<Arc<BoundColumnFamily<'_>>, error::Error> {
+        self.db.cf_handle(WORKING_SET_COLUMN_FAMILY).ok_or_else(|| {
+            error::Error::UnableToOpenColumnFamily(WORKING_SET_COLUMN_FAMILY.to_string())
+        })
+    }
+
+    /// Trie store iterator.
```
goral09: Can you include in the comment what it is iterating over? It's not iterating over elements in the trie store - trie nodes, is it?

execution_engine/src/storage/global_state/db.rs
```diff
@@ -267,7 +266,7 @@ impl DbGlobalState {
 
 fn memoized_find_missing_descendants(
     value_bytes: Bytes,
-    rocksdb: RocksDbStore,
+    rocksdb_store: &RocksDbTrieStore,
```
goral09: I don't think we have to include the type of the DB in the name - maybe `trie_store` is enough.

execution_engine/src/storage/global_state/db.rs
```diff
@@ -99,12 +99,12 @@ pub struct DbGlobalStateView {
     /// Root hash of this "view".
     root_hash: Digest,
     /// Handle to rocksdb.
-    rocksdb_store: RocksDbStore,
+    rocksdb_store: RocksDbTrieStore,
```
goral09: This `DbGlobalStateView` looks like a good candidate for a working set?

execution_engine/src/storage/trie_store/db.rs
```diff
@@ -142,147 +120,103 @@ impl Store<Digest, Trie<Key, StoredValue>> for ScratchTrieStore {
...
-
-        // recommended to match # of cores on host.
-        db_opts.increase_parallelism(num_cpus::get() as i32);
-
-        db_opts.create_missing_column_families(true);
-        db_opts.create_if_missing(true);
-
-        db_opts
+    /// Access the underlying DbStore.
+    pub fn get_db_store(&self) -> &DbStore {
```
goral09, line 209: Should we be exposing the underlying RocksDB for consumption in the public API? Can we hide it behind domain-specific methods?

goral09 (approved):

execution_engine/src/storage/global_state/db.rs
dwerner: The `...GlobalStateView` stuff is code consumed by the EE. I agree we need access to the working set during execution.

execution_engine/src/storage/trie_store/db.rs
dwerner: We can make this `pub(crate)` for now. Done in [6c20f4a](https://github.com/casper-network/casper-node/pull/2999/commits/6c20f4a0f076a2a375dc7d0f332871c085825545)

execution_engine/src/storage/global_state/db.rs
dwerner: Done in [6c20f4a](https://github.com/casper-network/casper-node/pull/2999/commits/6c20f4a0f076a2a375dc7d0f332871c085825545)

execution_engine/src/storage/db_store.rs
dwerner: Yes, it is, however I intended to move this to `RocksDbTrieStore`. Done in [6c20f4a](https://github.com/casper-network/casper-node/pull/2999/commits/6c20f4a0f076a2a375dc7d0f332871c085825545)

PR #3003, opened 2022-05-18
Optimize deploy metadata db initialization

### LINKED ISSUE #2975: Optimize initialize_deploy_metadata_db in Storage
This function should be refactored to take a list of deploy hashes (rather than block hashes) representing deploys for which their corresponding block has just been deleted due to hard-reset.  This list can then be used to just delete those entries from the deploy_metadata_db rather than iterating every entry in that DB.

rafal-ch:
This PR optimizes the implementation of `initialize_deploy_metadata_db()` function. Instead of iterating over all entries, it'll just delete the deploys & transfers related to the block that is deleted due to the hard reset.

It also modifies the related tests so they support the invariants required by the new implementation.

Closes https://github.com/casper-network/casper-node/issues/2975

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -500,9 +500,11 @@ impl StorageInner {
 
         let mut deleted_block_hashes = HashSet::new();
         let mut deleted_block_body_hashes_v1 = HashSet::new();
+        let mut deleted_deploy_hashes = HashSet::<DeployHash>::new();
         // Note: `iter_start` has an undocumented panic if called on an empty database. We rely on
         //       the iterator being at the start when created.
         for (_, raw_val) in cursor.iter() {
+            let mut body_txn = env.begin_ro_txn()?;
             let block_header: BlockHeader = lmdb_ext::deserialize(raw_val)?;
             if let Some(invalid_era) = hard_reset_to_start_of_era {
                 // Remove blocks that are in to-be-upgraded eras, but have obsolete protocol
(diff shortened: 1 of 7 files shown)


node/src/components/storage.rs
```diff
@@ -1718,7 +1706,19 @@ impl StorageInner {
...
+            &block_header,
+            &Databases {
+                block_body_v1_db: self.block_body_v1_db,
+                block_body_v2_db: self.block_body_v2_db,
+                deploy_hashes_db: self.deploy_hashes_db,
+                transfer_hashes_db: self.transfer_hashes_db,
+                proposer_db: self.proposer_db,
+            },
+            self.verifiable_chunked_hash_activation,
+        );
```
afck, line 1720: Since it's used in multiple places, that could still be extracted into a method (even with the same name).

Fraser999 (approved):

afck (approved):

node/src/components/storage.rs
rafal-ch: Updated in c379033e6d7ae67025dbe58e4056552cede926bf

afck: Sorry, that's not what I meant! there's no reason to use a private getter for that field!
I meant keeping the _method_ `self.get_body_for_block_header(tx, &block_header)` (which you removed), that would then call the _function_ `get_body_for_block_header`. Sorry for the confusion!

rafal-ch: I would not be able to use `self.get_body_for_block_header(tx, &block_header)` in the `StorageInner::new`, that's why I changed it to free function. The previous implementation was retrieving block in the constructor "manually" (probably because constructor is lacking a `self`) and I wanted to keep the body retrieval code in one place.

I'm open to discussion, though.

afck: Yes, I agree: That's why I'd keep the free function as well, and make the method just call it.
(Not important, and not a blocker! But I would prefer reverting https://github.com/casper-network/casper-node/pull/3003/commits/c379033e6d7ae67025dbe58e4056552cede926bf.)

rafal-ch: Sure thing, the revert is here: cd9aee38360eef70ac22c6f787c02a1f1630300f

PR #3007, opened 2022-05-19
Simplify fetcher flow

goral09:
Main things changed in this PR:
 * we will register responder only after failed storage query - there's no need to register the fetcher if data will be found locally since that query will return (almost) immediately.
 * API of methods signaling result to the responder have been refactor to reflect the previous change - if we respond with storage result immediately, there's no need to account for storage result in the API of methods that will get the results only from a peer.


Fraser999 (approved):

goral09:
bors merge

PR #3008, opened 2022-05-19
Provide block hash and height in JSON-RPC deploy query

### LINKED ISSUE #2985: The JSON-RPC must provide the block a deploy was included in when queried for a deploy
The JSON-RPC must provide the block a deploy was included in when queried for a deploy.  It is likely that storage will have to create a reverse-index for this.

georgepisaltu:
Fixes #2985 

This PR:
- Adds `BlockHashAndHeight` - describes the hash and height of a block.
- Adds `DeployMetadataExt` - extended deploy metadata used when the actual metadata is not available to provide more information about a deploy.
- Changes the storage component's `GetDeployAndMetadata`, which is used by the RPC server, to return block information about a deploy when execution results are not available.
- Changes the RPC server to use the new `GetDeployAndMetadata` for a `GetDeploy` RPC request.
- Updates the JSON schema with the new optional block information for a `GetDeployResult`

`StorageRequest::GetDeployAndMetadata` first tries to get the associated execution results, just like before. If there aren't any available, instead of returning an empty execution results `Vec`, it now tries to get the block hash and height of the block this deploy belongs to. These are stored in the `deploy_hash_index` where we now also store the block height in addition to the hash. If the deploy hash wasn't indexed but the deploy was found, we return empty metadata.


node/CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ All notable changes to this project will be documented in this file.  The format
 * Add a new identifier `PurseIdentifier` which is a new parameter to identify URefs for balance related queries.
 * Extend `GlobalStateIdentifier` to include `BlockHeight`.
 * Add a new RPC endpoint `query_balance` which queries for balances underneath a URef identified by a given `PurseIdentifier`.
+* Add new `block_hash` and `block_height` optional fields to JSON-RPC deploy query. They will be present in the response
```
rafal-ch: Can we name the exact query? For example:

```suggestion
* Add new `block_hash` and `block_height` optional fields to `info_get_deploy` RPC query. They will be present in the response
* ```

File: node/src/types/deploy.rs [hunk 2]
```diff
@@ -1615,6 +1615,53 @@ pub struct DeployMetadata {
...
+/// Additional information describing a deploy.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub enum DeployMetadataExt {
+    /// Holds the execution results of a deploy.
+    Metadata(DeployMetadata),
+    /// Holds the hash and height of the block this deploy was included in.
+    BlockInfo(BlockHashAndHeight),
+    /// No execution results or block information available.
+    Empty,
+}
```
rafal-ch, line 1644: Couldn't we just get away with adding more fields to `DeployMetadata` struct? Or put `execution_results` directly in `DeployMetadataExt`? What we have now is a nested metadata :)

Another alternative could be to rename the `DeployMetadata` to `ExecutionResults` :thinking:

node/src/components/storage.rs
```diff
@@ -234,7 +234,7 @@ struct Indices {
     /// A map of era ID to switch block ID.
     switch_block_era_id_index: BTreeMap<EraId, BlockHash>,
     /// A map of deploy hashes to hashes of blocks containing them.
-    deploy_hash_index: BTreeMap<DeployHash, BlockHash>,
+    deploy_hash_index: BTreeMap<DeployHash, BlockHashAndHeight>,
```
rafal-ch, line 237: The comment could be updated, like so: 
```
/// A map of deploy hashes to hashes and heights of blocks containing them.
```

node/src/components/rpc_server/rpcs/info.rs
```diff
@@ -102,6 +107,12 @@ pub struct GetDeployResult {
     pub deploy: Deploy,
     /// The map of block hash to execution result.
     pub execution_results: Vec<JsonExecutionResult>,
+    /// The hash of this deploy's block.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub block_hash: Option<BlockHash>,
+    /// The height of this deploy's block.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub block_height: Option<u64>,
```
Fraser999: I think we should use the `BlockHashAndHeight` to ensure we always have both `Some` or both `None`.  It will need updated to derive `JsonSchema` and should also have `#[serde(deny_unknown_fields)]` applied.  Then we can do:
```rust
    /// The hash and height of the block in which this deploy was executed, only provided if the
    /// full execution results are not know on this node.
    #[serde(skip_serializing_if = "Option::is_none", flatten)]
    pub block_hash_and_height: Option<BlockHashAndHeight>,
```

rafal-ch:
Additionally, this change will probably make our nightly tests fail due to the change in JSON responses. Please consider submitting a companion PR against the `casper-test`.

node/CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ All notable changes to this project will be documented in this file.  The format
...
+when the queried deploy has no associated execution results in storage, but has an associated block.
```
Fraser999: One line per bullet point pls.

node/CHANGELOG.md
georgepisaltu: Done.

node/src/components/rpc_server/rpcs/info.rs
georgepisaltu: Done. I the descriptions were missing so I added them with `#[schemars(description = "...")]`.

node/src/components/storage.rs
georgepisaltu: Done.

node/src/types/deploy.rs
georgepisaltu: We wanted the solution here to be backwards compatible. `DeployMetadata` is used in many other places and, with this in mind, we can do no renames in `DeployMetadata`. As `DeployMetadataExt` is a completely new struct, we could just add execution results into it, but the point was that we only care about block hash and height when we don't have the execution results, and vice versa, so it made sense to have it as an `enum`, with either or. This also makes it easier to work with the (unlikely) case where we have neither.

The point about nested metadata is valid, but I think it's because `DeployMetadata` is not the best name for what is basically just execution results.

Considering all this, if you don't have strong opinions here, I'd like to keep it as is. We will have flexibility with `DeployMetadataExt` in the future as it's only used for this RPC query and we fully control how it looks like to the end user.

node/src/components/rpc_server/rpcs/info.rs
Fraser999: Hmm - I thought the descriptions are automatically derived from the doc comments.  Either way - the change LGTM.

Fraser999 (approved):

node/src/components/rpc_server/rpcs/info.rs
georgepisaltu: They are, but I didn't want to comment `The hash of this deploy's block.` on `BlockHashAndHeight` because the struct itself, `BlockHashAndHeight`, is not really related to deploys in any way.

node/src/types/deploy.rs
rafal-ch: Thanks for explanation :+1:

rafal-ch (approved):
LGTM :+1:

PR #3010, opened 2022-05-19
Batch deploys with block and reduce deploy validation in chain sync

### LINKED ISSUE #3006: Avoid full validation of deploys fetched as part of chain-sync
Deploys being fetched by chain-sync should only have their hash validated before being stored and announced

Fraser999:
This PR builds on top of #2886 and further implements a fix for overly strict deploy validation being performed during chain synchronization.

Closes #3006.


node/src/components/storage.rs
```diff
@@ -1196,6 +1219,34 @@ impl StorageInner {
...
+    /// Puts block and its deploys into storage.
+    pub fn put_block_and_deploys(
+        &self,
+        block_and_deploys: &BlockAndDeploys,
+    ) -> Result<bool, FatalStorageError> {
+        let BlockAndDeploys { block, deploys } = block_and_deploys;
+
+        let write_block_result = self.write_block(block)?;
+
+        let write_deploys_result = self.put_deploys(deploys)?;
```
goral09: I'm wondering now if we shouldn't hold a transaction across both of these operations and not commit the data if writing any of the deploys fails 🤔 What do you think? We don't have good recovery for this kind of errors yet.

node/src/effect/requests.rs
```diff
@@ -469,7 +485,16 @@ impl Display for StorageRequest {
     fn fmt(&self, formatter: &mut Formatter<'_>) -> fmt::Result {
         match self {
             StorageRequest::PutBlock { block, .. } => write!(formatter, "put {}", block),
+            StorageRequest::PutBlockAndDeploys {
+                block: block_deploys,
+                ..
+            } => {
+                write!(formatter, "put block and deploys {}", block_deploys)
```
goral09: Maybe we should log just the hash of a block rather than the whole block?

node/src/types/deploy.rs
```diff
@@ -919,6 +919,23 @@ impl Deploy {
         Ok(())
     }
 
+    pub(crate) fn has_valid_hash(&self) -> Result<(), DeployConfigurationFailure> {
```
goral09, line 933: Can you add a comment please?

goral09 (approved):
No blockers but I think we might want to hold a write txn across both block and all deploys being written to avoid a situation where we write partially and fail.

node/src/components/storage.rs
Fraser999: Good point.  Addressed in [a607d3f](https://github.com/casper-network/casper-node/pull/3010/commits/a607d3fe9cc2a32b520faa06551f0b82b78ef701).

node/src/effect/requests.rs
Fraser999: Fixed in [02645db](https://github.com/casper-network/casper-node/pull/3010/commits/02645db01331a35cc2ab337177741ca0ffae3f3f).

node/src/types/deploy.rs
Fraser999: Fixed in [02645db](https://github.com/casper-network/casper-node/pull/3010/commits/02645db01331a35cc2ab337177741ca0ffae3f3f).

goral09 (approved):
Thanks. There are some compilation errors after merging with `dev`.

node/src/types/block.rs
```diff
@@ -1834,6 +1834,73 @@ impl Item for BlockWithMetadata {
...
+            .block
+            .deploy_hashes()
+            .iter()
+            .chain(self.block.transfer_hashes().iter())
+        {
+            match self
+                .deploys
+                .iter()
+                .find(|&deploy| deploy.id() == deploy_hash)
+            {
```
georgepisaltu, line 1887: What is the relationship between `block.deploy_hashes()` + `block.transfer_hashes()` compared to `deploys`? I see here we're checking that all the deploy hashes in block are also in deploys, but are we supposed to check the inverse as well? Also, I see we can receive this from a peer as well, which I assume is not trusted. Do we have a checks for those deploys belonging to the block or is that covered in `block.verify()` and `deploy.has_valid_hash()`?

node/src/types/block.rs
```diff
@@ -1834,6 +1834,73 @@ impl Item for BlockWithMetadata {
...
+/// Wrapper around block and its deploys.
+pub struct BlockAndDeploys {
+    /// Block part.
+    pub block: Block,
+    /// Deploys.
+    pub deploys: Vec<Deploy>,
+}
+
+impl Display for BlockAndDeploys {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
georgepisaltu:
```suggestion
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```

node/src/components/storage.rs
```diff
@@ -1196,19 +1219,67 @@ impl StorageInner {
...
         // Validate the block prior to inserting it into the database
         block.verify(self.verifiable_chunked_hash_activation)?;
         let mut txn = self.env.begin_rw_txn()?;
-        // Write the block body
+        let result = self.write_validated_block(&mut txn, block);
+        match &result {
+            Ok(false) | Err(_) => txn.abort(),
+            Ok(true) => txn.commit()?,
+        }
+        result
```
georgepisaltu, line 1287: nit: could be
```
Ok(self
    .read_deploys(deploys_count, deploy_hashes)?
    .map(|deploys| BlockAndDeploys { block, deploys }))
```
or
```
self.read_deploys(deploys_count, deploy_hashes)
    .map(|maybe_deploys| {
        maybe_deploys.map(|deploys| BlockAndDeploys { block, deploys })
    })
```

georgepisaltu (commented):
Looks good, just a question and a couple of nits.

node/src/types/block.rs
Fraser999: > What is the relationship between `block.deploy_hashes()` + `block.transfer_hashes()` compared to `deploys`? ... but are we supposed to check the inverse as well?

Transfers are just a special case of Deploys.  In most places we don't distinguish between transfers and non-transfer deploys, but we do in the block.  In this specific case, we expect the combined list of deploy and transfer hashes in the block to be the same as the set of deploys provided in `BlockAndDeploys`.  I think you're right - we should be checking that there aren't extra deploys being provided.  Good catch!

> Do we have a checks for those deploys belonging to the block or is that covered in block.verify() and deploy.has_valid_hash()?

Yes, those checks cover the case of proving the deploy belongs in the given block.

Fraser999: Fixed in [ecf4e89](https://github.com/casper-network/casper-node/pull/3010/commits/ecf4e895ee12a1b1c67900fe7d519bc4ed9dc658).

node/src/types/block.rs
Fraser999: Fixed in [ecf4e89](https://github.com/casper-network/casper-node/pull/3010/commits/ecf4e895ee12a1b1c67900fe7d519bc4ed9dc658).

node/src/components/storage.rs
Fraser999: Fixed in [ecf4e89](https://github.com/casper-network/casper-node/pull/3010/commits/ecf4e895ee12a1b1c67900fe7d519bc4ed9dc658).

georgepisaltu (approved):
LGTM

Fraser999:
bors merge

PR #3011, opened 2022-05-20
Log chainspec and key block hashes when starting era.

afck:
The instance ID is computed from the chainspec hash, the key block hash and the era ID. If it doesn't match it's helpful to know which value was wrong.


goral09 (approved):

goral09:
bors merge

PR #3020, opened 2022-05-24
Archival sync in batches & parallel.

### LINKED ISSUE #3015: Speed up sync to Genesis
Speed up sync to Genesis

goral09:
Closes https://github.com/casper-network/casper-node/issues/3015

The fast-sync algorithm consists of three phases:
1. Initial sync - where we download the trusted block and the trie (state) under its state root hash.
2. syncing to Genesis (if `archival_sync=true`) - where we download all data from before the trusted block: all blocks, deploys and all of the state tries under all blocks.
3. syncing forward - where we synchronize chain from the trusted block onwards, to the latest block at the tip.

This PR affects only the second step in that process.

Historically (before this PR), the _sync to Genesis_ part would start at trusted block and then sequentially download its parent block, its deploys and the state trie for that block. That process would be repeated until we hit Genesis.

This turned out to be too slow - the requested/response for block, independent requests for all of its deploys and then multiple, independent requests for the trie took much longer than expected. The process was estimated to finish (for mainnet data) in *years*.

This PR changes the algorithm (of **2. Syncing to Genesis only**) in the following way:
1. Starting from trusted block, fetch batch of headers (currently a batch of 1000) from `trusted_block.height..=trusted_block.height - 1000`. Every batch is validated (for continuity, proper parent-child relationship, cryptographic verification).
2. Continue syncing batches until we hit Genesis. 
3. After we hit Genesis, we start "walking back" - from Genesis to the trusted block. 
4. We spawn up to `max_parallel_block_fetches` workers, each fetching block AND its deploy (another change compared to the original algorithm, this is safe b/c block and deploys' hashes are not allowed to be bigger than network message size limit). After block with deploys is downloaded, we start downloading its state - the trie under its state root hash. We spawn up to `max_parallel_trie_fetches` workers (shared across all "block workers") to synchronize the trie.
5. When "block workers" recognize that the next block they would be downloading is the trusted block (or its child- by height) - they stop and exit.
6. When all block workers finish - we're synchronized the archival state leading up to the trusted block.


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1074,71 +1075,169 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
+    ctx: &ChainSyncContext<'_>,
+) -> Result<(), Error> {
+    while let Some(job) = queue.next_job().await {
+        let block_hash = *job.inner();
+        info!(worker_id, ?block_hash, "syncing block and deploys");
+        match fetch_and_store_block_with_deploys_by_hash(block_hash, ctx).await {
+            Ok(fetched_block) => {
+                trace!(?block_hash, "downloaded block and deploys");
+                // We want to download the trie only when we know we have the block.
+                sync_trie_store(*fetched_block.block.state_root_hash(), ctx).await?;
```
goral09: `sync_trie_store_worker` we will create `max_parallel_trie_fetches` meaning that the overall process of `fetch_blocks_and_state_since_genesis` will end up creating up to `max_parallel_block_fetches * max_parallel_trie_fetches` parallel `sync_trie_store_worker`s. 

One solution to that is to share the trie `WorkQueue` across all `fetch_block_worker`s.

node/src/components/chain_synchronizer/error.rs
```diff
@@ -53,6 +54,9 @@ pub(crate) enum Error {
     #[error("no such block hash: {bogus_block_hash}")]
     NoSuchBlockHash { bogus_block_hash: BlockHash },
 
+    #[error("no such block height: {0} when archival sync")]
```
marc-casperlabs: Maybe something like "no such block height: {0} encountered during archival sync"?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1074,71 +1075,182 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
-                    skip_to_height = %ctx.locally_available_block_range_on_start.low(),
-                    current_walkback_height = %walkback_block.height(),
-                    "skipping fetch for blocks, deploys and tries in available block range"
+        match fetch_block_headers_batch(&lowest_trusted_block_header, ctx).await {
+            Ok(new_lowest) => {
+                info!(?new_lowest, "new lowest trusted block header stored");
+                lowest_trusted_block_header = new_lowest;
+            }
+            Err(err) => {
+                // If we get an error here it means something must have gone really wrong.
```
marc-casperlabs, line 1132: I dig the long explanatory comment :+1:

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1074,71 +1075,182 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
+    let abort = Arc::new(AtomicBool::new(false));
+
+    for block_height in 0..=ctx.config.max_parallel_block_fetches() as u64 {
+        let block_hash = match ctx
+            .effect_builder
+            .get_block_hash_by_height_from_storage(block_height)
+            .await
+        {
+            Some(hash) => hash,
+            None => return Err(Error::NoSuchBlockHeight(block_height)),
```
marc-casperlabs: If this is error is only used here, the error name and message and such could reflect the fact that we expected the block to be there, since its header was fetched.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1074,71 +1075,182 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
+// Fetches blocks, their transactions and tries from Genesis until `trusted_block`.
+async fn fetch_blocks_and_state_since_genesis(ctx: &ChainSyncContext<'_>) -> Result<(), Error> {
+    info!("syncing blocks and deploys and state since Genesis");
+
+    let queue = Arc::new(WorkQueue::default());
+
+    // Flag set by a worker when it encounters an error.
+    let abort = Arc::new(AtomicBool::new(false));
+
+    for block_height in 0..=ctx.config.max_parallel_block_fetches() as u64 {
```
marc-casperlabs: Are you sure this is the correct range? I believe it should fetch up to the latest block, not `ctx.config.max_parallel_block_fetches()`?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1074,71 +1075,182 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
+        queue.push_job(block_hash);
+    }
+
+    let mut workers: FuturesUnordered<_> = (0..ctx.config.max_parallel_block_fetches())
+        .map(|worker_id| fetch_block_worker(worker_id, abort.clone(), queue.clone(), ctx))
+        .collect();
+
+    while let Some(result) = workers.next().await {
+        result?; // Return the error if a download failed.
+        ctx.metrics.chain_sync_blocks_synced.inc();
```
marc-casperlabs: This will record the number of workers that finished, not the number of blocks synced. IIRC workers should exit only when there are no more jobs coming in, and each worker should usually downloading more than a single block?

node/src/components/chain_synchronizer/operations.rs
marc-casperlabs: The issue here is that the `WorkQueue` has a specific notion of "done" - if all jobs are finished and they have not added any new jobs to the queue, the entire task --- fetching all tries for a specific root hash --- is complete. This sort of signaling needs to be preserved somehow, we currently use the end of all workers as the indicator of jobs done.

node/src/components/chain_synchronizer/operations.rs
goral09: Good catch. Thanks.

node/src/components/chain_synchronizer/operations.rs
goral09: This only creates a correct number of workers. The workers themselves enqueue next block to download.

node/src/components/chain_synchronizer/operations.rs
marc-casperlabs: A quick fix would be to use a semaphore, acquiring a ticket before spawning a worker and releasing it once a worker finishes. I'm sure we can build something better by refactoring, but that's a reasonable start.

We may need to think about persistent worker pools in the future, and properly wrap jobs being passed to them. The harder part is just signaling that a specific subset of tasks is done.

node/src/components/fetcher.rs
```diff
@@ -603,6 +603,42 @@ impl ItemFetcher<BlockHeader> for Fetcher<BlockHeader> {
     }
 }
 
+impl ItemFetcher<BlockHeadersBatch> for Fetcher<BlockHeadersBatch> {
+    const SAFE_TO_RESPOND_TO_ALL: bool = false;
```
marc-casperlabs: I don't think it matters, but I wonder if this is actually `true` :).

node/src/types/block.rs
```diff
@@ -1111,6 +1113,209 @@ impl Item for BlockHeaderWithMetadata {
         self.block_header.height()
     }
 }
+#[derive(DataSize, Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+/// ID idenfitifying a request for a batch of block headers.
+pub struct BlockHeadersBatchId {
+    pub highest: u64,
```
marc-casperlabs, line 1119: I think the block header batches should be hardcoded, with the height either being fixed or determined by the chainspec. For example, if that number is 1024, batch `0` is always going to be headers `genesis..1023`, batch `1` blocks `1024-2047` and so on. I think that would obviate a lot of checking.

node/src/types/block.rs
```diff
@@ -1111,6 +1113,209 @@ impl Item for BlockHeaderWithMetadata {
...
+                expected: *latest_known.parent_hash(),
+                got: highest_hash,
+            });
+        }
+
+        if !Self::is_continuous_and_descending(self.inner(), verifiable_chunked_hash_activation) {
+            return Err(BlockHeadersBatchValidationError::BatchNotContinuous);
+        }
+
+        self.0
```
marc-casperlabs, line 1191: This could also include a size check. If we check anything but the latest chunk, we could indicate we expect this to be of a specific length.

node/src/types/block.rs
```diff
@@ -1111,6 +1113,209 @@ impl Item for BlockHeaderWithMetadata {
...
+    pub fn inner(&self) -> &Vec<BlockHeader> {
+        &self.0
+    }
+
+    /// Tests whether the block header batch is continuous and in descending order.
+    pub fn is_continuous_and_descending(
+        batch: &[BlockHeader],
+        verifiable_chunked_hash_activation: EraId,
+    ) -> bool {
+        batch
```
marc-casperlabs, line 1261: While I do appreciate functional iterator magic, I think 

```rust
for i in 1..batch.len() {
    let current = batch[i];
    let pred = batch[i-1];

     if current.parent_hash() != pred.hash(...) { ... }
     if curent.height() != pred.height() +1 { ... }
}
```

is maybe a little easier to read and review? :)

goral09: I actually find the "iterator magic" more declarative:
* create windows of size 2
* filter out any window that doesn't have 2 elements
* validate pairs

node/src/components/fetcher.rs
goral09: You're probably right since there is only one valid batch - there's only one valid block at any block height. I'll change.

node/src/components/chain_synchronizer/operations.rs
goral09: You're right. I was thinking that I accomodate for any future use of the error but I should focus on the task at hand and expect devs coming after me to update the error message if necessary.

node/src/types/block.rs
goral09: Yeah, we can do that.

node/src/types/block.rs
goral09: Maybe, that's a different approach than what I took. I'll stick to what we have since it's complete.

node/src/types/block.rs
goral09: Done in [`9a8a3cc` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/9a8a3ccb0d005f93b230be0a166628b293b7b903)

node/src/components/fetcher.rs
goral09: Done in [`9c7c3e2` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/9c7c3e2c3492f1f6f3d6583055ec6d6c33cfe53b)

node/src/components/chain_synchronizer/operations.rs
goral09: Done in [`8bf9239` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/8bf923978268ac8fca4d75b89066772fa0b65dab)

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+        }
+
+        if !Self::is_continuous_and_descending(self.inner(), verifiable_chunked_hash_activation) {
+            return Err(BlockHeadersBatchValidationError::BatchNotContinuous);
+        }
+
+        self.0
+            .last()
+            .cloned()
+            .ok_or(BlockHeadersBatchValidationError::BatchEmpty)
```
xcthulhu, line 1194: Why is this returning a block header instead of `()`?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1074,71 +1075,189 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
+                    "failed to download block headers batch with infinite retries"
                 );
+                return Err(err.into());
+            }
+        }
 
-                if lowest_available_block.block.height() == 0 {
-                    ctx.effect_builder
-                        .update_lowest_available_block_height_in_storage(0)
+        if lowest_trusted_block_header.height() == 0 {
```
afck, line 1144: Why not just make it a `while` loop?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1074,71 +1075,189 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
+    let abort = Arc::new(AtomicBool::new(false));
+
+    for block_height in 0..=ctx.config.max_parallel_block_fetches() as u64 {
+        let block_hash = match ctx
+            .effect_builder
+            .get_block_hash_by_height_from_storage(block_height)
+            .await
+        {
+            Some(hash) => hash,
+            None => return Err(Error::NoSuchBlockHeight(block_height)),
```
afck: What if the current chain length is less than the number of workers?

Wouldn't it be simpler to just use a `height_counter: AtomicU64` instead of the work queue? The workers would, in a loop, `let height = height_counter.fetch_add(1, SeqCst);`, read the hash from storage and download block, deploys and global state. It would exit the loop if `height >= ctx.trusted_block_header().height()`.

That way the number of workers would occur only in one place, and not inside the worker logic.

node/src/components/fetcher.rs
```diff
@@ -603,6 +603,42 @@ impl ItemFetcher<BlockHeader> for Fetcher<BlockHeader> {
...
+        &mut self.responders
+    }
+
+    fn metrics(&mut self) -> &Metrics {
+        &self.metrics
+    }
+
+    fn peer_timeout(&self) -> Duration {
+        self.get_from_peer_timeout
+    }
```
afck, line 622: (I wonder whether this trait is a bit redundant: Would it be simpler to move these three getters into `impl<T> Fetcher<T>`, and `get_from_storage` to `Item`? Anyway, has nothing to do with this PR.)

node/src/types/error.rs
```diff
@@ -142,3 +142,15 @@ pub(crate) enum BlockWithMetadataValidationError {
...
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub(crate) enum BlockHeadersBatchValidationError {
+    #[error("Batch has incorrect lenght. Expect: {expected}, got: {got}")]
+    IncorrectLength { expected: u64, got: u64 },
+    #[error("Returned batch is not continuous sequence of blocks of parent-child relationship.")]
+    BatchNotContinuous,
+    #[error("Empty batch")]
+    BatchEmpty,
+    #[error("Hash of the highest block from batch doesn't match expected. Expected: {expected}, got: {got}.")]
```
afck: (Line too long.)

node/src/types/error.rs
```diff
@@ -142,3 +142,15 @@ pub(crate) enum BlockWithMetadataValidationError {
     #[error(transparent)]
     BlockHeaderWithMetadataValidationError(#[from] BlockHeaderWithMetadataValidationError),
 }
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub(crate) enum BlockHeadersBatchValidationError {
+    #[error("Batch has incorrect lenght. Expect: {expected}, got: {got}")]
```
afck:
```suggestion
    #[error("Batch has incorrect length. Expected: {expected}, got: {got}")]
```

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
         self.block_header.height()
     }
 }
+#[derive(DataSize, Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+/// ID idenfitifying a request for a batch of block headers.
```
afck:
```suggestion
/// ID identifying a request for a batch of block headers.
```

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+pub struct BlockHeadersBatchId {
+    pub highest: u64,
+    pub lowest: u64,
+}
+
+impl BlockHeadersBatchId {
+    pub fn new(highest: u64, lowest: u64) -> Self {
+        Self { highest, lowest }
+    }
+    pub fn from_known(lowest_known_block_header: &BlockHeader, max_batch_size: u64) -> Self {
```
afck:
```suggestion
    }

    pub fn from_known(lowest_known_block_header: &BlockHeader, max_batch_size: u64) -> Self {
```

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+        Self { highest, lowest }
+    }
+
+    /// Return an iterator over block header heights starting from highest (inclusive) to lowest
+    /// (inclusive).
+    pub fn iter(&self) -> impl Iterator<Item = u64> {
+        (self.lowest..=self.highest).rev()
+    }
+
+    /// Returns a length of the batch.
```
afck:
```suggestion
    /// Returns the length of the batch.
```

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+    pub fn len(&self) -> u64 {
+        self.highest + 1 - self.lowest
```
afck, line 1145: This can overflow unless `highest >= lowest && highest != u64::MAX`.

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+impl BlockHeadersBatch {
+    /// Validates whether received batch is:
+    /// 1) highest block header from the batch has a hash is `latest_known.hash`
+    /// 2) a link of header[n].parent == header[n+1].hash is maintained
+    ///
+    /// Returns lowest block header from the batch or error if batch fails validation.
+    pub(crate) fn validate(
+        &self,
+        batch_id: &BlockHeadersBatchId,
+        latest_known: &BlockHeader,
```
afck: Isn't this rather `earliest_known` or `lowest_known`?

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+    /// Returns `Some(Self)` if data passes validation, otherwise `None`.
+    pub fn from_vec(
+        batch: Vec<BlockHeader>,
+        requested_id: &BlockHeadersBatchId,
+        verifiable_chunked_hash_activation: EraId,
+    ) -> Option<Self> {
+        match batch.first() {
+            Some(highest) => {
+                if highest.height() != requested_id.highest {
+                    error!(expected_highest=?requested_id.highest, got_highest=?highest, "unexpected highest block header");
```
afck: (Line too long. Also below.)

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+    pub fn is_continuous_and_descending(
+        batch: &[BlockHeader],
+        verifiable_chunked_hash_activation: EraId,
+    ) -> bool {
+        batch
+            .windows(2)
+            .filter_map(|window| match &window {
+                &[l, r] => Some((l, r)),
+                _ => None,
+            })
```
afck, line 1266: You can use `Itertools::tuple_windows` to avoid this `filter_map` step.

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+            .all(|(l, r)| {
+                l.height() == r.height() + 1
+                    && l.parent_hash() == &r.hash(verifiable_chunked_hash_activation)
+            })
+    }
+
+    #[cfg(test)]
+    // Test-only constructor allowing creation otherwise invalid data.
```
afck:
```suggestion
    // Test-only constructor allowing creation of otherwise invalid data.
```

node/src/types/block.rs
```diff
@@ -1111,6 +1113,217 @@ impl Item for BlockHeaderWithMetadata {
...
+            error!(
+                ?lower_batch_height,
+                ?upper_batch_height,
+                "received header batch is empty"
+            );
+        }
+        BlockHeadersBatchId::new(
+            upper_batch_height.unwrap_or_default(),
+            lower_batch_height.unwrap_or_default(),
+        )
```
afck, line 1328: I wonder whether the batch's highest block's hash should just be part of the ID? That way, if we're somehow on the wrong chain, the other nodes wouldn't all send us back a whole range of headers that we then throw away due to the mismatch? It would also remove some of the possibly overflowing arithmetic if the ID was just the highest hash and the length.

node/src/components/storage.rs
```diff
@@ -1186,31 +1200,41 @@ impl StorageInner {
...
+                for block_header in block_headers.into_iter() {
+                    let block_header_hash =
+                        block_header.hash(self.verifiable_chunked_hash_activation);
+                    match self.put_block_header(&mut txn, block_header) {
+                        Ok(single_result) => {
+                            result = result && single_result;
+                        }
+                        Err(err) => {
+                            error!(?err, ?block_header_hash, "error when storing block header");
+                            txn.abort();
```
afck: If we abort, we probably should update the block header index and remove the ones that were added during this transaction? (Also in other places where we call `put_block_header`.)

node/src/components/chain_synchronizer/operations.rs
goral09: > What if the current chain length is less than the number of workers?

Good point. 

I also like your suggestion for alternative. I'll see to integrate it.

node/src/components/storage.rs
goral09: Good observation. I will change the logic to update the index only if the whole batch is stored successfuly.

node/src/types/block.rs
goral09: Overflow if block height is `u64::MAX`, right? We would probably have a lot more problems if that happened.

node/src/types/block.rs
goral09: I assumed we would recognize much earlier in the process whether we are on a wrong chain.

node/src/types/block.rs
afck: Right, otherwise it's an underflow.

node/src/types/block.rs
afck: It depends. If some of our peers are trying to fool us, and we also got the wrong trusted hash?

goral09: If they try to fool us I don't think this would change much - i.e. they could ignore the request ID and respond with anything they want. 

We do validate the response against the lowest trusted block anyway so we would detect it early on whether the response if valid or not.

node/src/types/block.rs
goral09: I would say it's all of that - depending how you interpret it. Do you prefer `lowest_known`? I'll change it.

node/src/types/block.rs
afck: What if you're connected to 20 peers and one of them wants to fool you, and also managed (via some other channel) give you a wrong trusted hash: You'll reject all the answers from the good peers, and only accept his. And then you start requesting batches, by height, and receive ~10 full batches from good peers on average before you ask the bad peer who will finally send you the (wrong) batch that you're expecting.

If you made your requests by hash, the good peers wouldn't have to bother to respond, and could save their bandwidth. Not sure how much of a difference that would make. :shrug:

node/src/types/block.rs
afck: In what sense is it the latest known? In reverse order, because that's the order in which we're downloading them?
(I'm fine with it, if you find that more intuitive.)

goral09: since we're fetching and validating headers in descending orther, the _latest_ one we had downloaded is also _earliest_ & _lowest_.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1210,16 +1195,30 @@ async fn fetch_blocks_and_state_since_genesis(ctx: &ChainSyncContext<'_>) -> Res
...
+        }
+
+        let block_hash = match ctx
+            .effect_builder
+            .get_block_hash_by_height_from_storage(next_block_height)
+            .await
+        {
+            Some(hash) => hash,
+            None => return Err(Error::NoSuchBlockHeight(next_block_height)),
+        };
```
afck: (You could use `ok_or_else` or `ok_or`.)

afck (commented):
Thanks, the `AtomicU64` approach looks good!

node/src/components/chain_synchronizer/operations.rs
goral09: Of course, I will. Thanks.

node/src/types/block.rs
goral09: I'm going to resolve the comment for now :)

node/src/components/storage.rs
goral09: Done in [`7a00eda` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/7a00edae7ec6e8fdc9853c7fd8663b94bb1201a3)

node/src/components/chain_synchronizer/operations.rs
goral09: Done in [`fb0e801` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/fb0e801c87c60a207cbb44d59cdc021ad78e2ac3).

node/src/components/chain_synchronizer/operations.rs
goral09: This is done in [`51a4c1e` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/51a4c1efc029fd4adeb52355bba51646026a0cd6)

node/src/types/error.rs
goral09: Seems like I got rid of this enum along the way. Resolving the comment.

node/src/types/error.rs
goral09: You're killing me with this Andreas :) We shouldn't be expected to manually format things that rustfmt can't.

node/src/types/block.rs
goral09: I'll rename to `earliest_known`.

node/src/types/block.rs
goral09: When we sync headers in batches, back to Genesis, we always start with some "known/trusted" header and then calculate the batch we want to get from there. We fetch these batches sequentially, the first batch starts with _trusted header_, the next one should start at header with height `trusted_header.height - BATCH_SIZE`. After fetching the first batch, we want to get the lowest, currently known header. If we didn't return it from this method we would have to read it from storage.

node/src/types/block.rs
goral09: Done in [`60ac7e8` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/60ac7e8598c652d5e8d043c149df8319bdd03027)

node/src/types/block.rs
goral09: Done in [`60ac7e8` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/60ac7e8598c652d5e8d043c149df8319bdd03027)

node/src/types/block.rs
goral09: Done in [`60ac7e8` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/60ac7e8598c652d5e8d043c149df8319bdd03027)

node/src/types/block.rs
goral09: Done in [`60ac7e8` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/60ac7e8598c652d5e8d043c149df8319bdd03027)

node/src/types/block.rs
goral09: Done in [`60ac7e8` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/60ac7e8598c652d5e8d043c149df8319bdd03027)

node/src/components/chain_synchronizer/error.rs
Fraser999: We need to use the term "sync to genesis" rather than "archival sync"

utils/dry-run-deploys/Cargo.toml
```diff
@@ -22,6 +22,7 @@ once_cell = "1.5.2"
 rand = "0.8.3"
 reqwest = { version = "0.11.1", features = ["json"] }
 retrieve-state = { path = "../retrieve-state" }
+prometheus = "0.12.0"
```
Fraser999:
```suggestion
```

utils/retrieve-state/Cargo.toml
```diff
@@ -20,6 +20,7 @@ jsonrpc-lite = "0.5.0"
 lmdb = "0.8.0"
 num-rational = { version = "0.4.0", features = ["serde"] }
 reqwest = { version = "0.11.1", features = ["json", "gzip"] }
+prometheus = "0.12.0"
```
Fraser999:
```suggestion
```

node/src/types/block.rs
```diff
@@ -1111,6 +1113,226 @@ impl Item for BlockHeaderWithMetadata {
...
+                    error!(
+                        expected_highest=?requested_id.highest,
+                        got_highest=?highest,
+                        "unexpected highest block header"
+                    );
+                    return None;
+                }
+            }
+            None => {
+                error!("response cannot be an empty batch");
```
Fraser999: Seems like the error logs in this method should be warnings only, since they could be invoked by a malicious peer sending invalid data.

node/src/types/block.rs
```diff
@@ -1111,6 +1113,226 @@ impl Item for BlockHeaderWithMetadata {
...
+    ) -> bool {
+        batch
+            .windows(2)
+            .filter_map(|window| match &window {
+                &[l, r] => Some((l, r)),
+                _ => None,
+            })
+            .all(|(l, r)| {
+                l.height() == r.height() + 1
+                    && l.parent_hash() == &r.hash(verifiable_chunked_hash_activation)
```
Fraser999, line 1269: `r.hash` is a pretty expensive call, and we seem to be calling `is_continuous_and_descending` in a few places.

Maybe we should look to replicate the lazy-initialized `is_valid` field of `Deploy` with a `is_continuous_and_descending` field?

node/src/types/block.rs
```diff
@@ -1111,6 +1113,226 @@ impl Item for BlockHeaderWithMetadata {
...
+        Ok(())
+    }
+
+    fn id(&self, _verifiable_chunked_hash_activation: EraId) -> Self::Id {
+        let upper_batch_height = self.0.first().map(|h| h.height());
+        let lower_batch_height = self.0.last().map(|h| h.height());
+        if lower_batch_height.is_none() || upper_batch_height.is_none() {
+            // ID should be infallible but it is possible that the `Vec` is empty.
+            // In that case we log an error to indicate something went really wrong and use `(0,0)`.
+            error!(
```
Fraser999: Probably just a `warn!` here again for the same reason as above?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -1084,75 +1098,191 @@ async fn fetch_to_genesis(trusted_block: &Block, ctx: &ChainSyncContext<'_>) ->
...
+            fetch_retry_forever::<BlockHeadersBatch>(ctx, batch_id).await?;
+        match fetched_headers_data {
+            FetchedData::FromStorage { item } => {
+                return BlockHeadersBatch::validate(
+                    &*item,
+                    &batch_id,
+                    lowest_trusted_block_header,
+                    ctx.config.verifiable_chunked_hash_activation(),
+                )
+                .map_err(FetchBlockHeadersBatchError::InvalidBatchFromStorage)
```
Fraser999: Seems like we're calling `validate` here only to get the lowest block header from the batch.  I'd be inclined to change the validation to return `Result<(), ...>` and have a separate getter on the `BlockHeadersBatch` for the lowest header.

Also, nitpick: this would more normally be written `item.validate(...)` rather than `BlockHeadersBatch::validate(&*item, ...)`.

node/src/types/block.rs
```diff
@@ -1111,6 +1113,226 @@ impl Item for BlockHeaderWithMetadata {
         self.block_header.height()
     }
 }
+#[derive(DataSize, Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+/// ID identifying a request for a batch of block headers.
+pub struct BlockHeadersBatchId {
```
Fraser999: I think all the new `pub` structs in this file and their fields and methods can be `pub(crate)` instead.

Fraser999 (commented):
Mostly nitpicks, but I'd like feedback on the `is_continuous_and_descending` comment before approving.

node/src/components/chain_synchronizer/error.rs
goral09: Will do.

node/src/types/block.rs
goral09: I'm not a fan of the approach taken with `is_valid`. I'll try to rewrite the code so that it's more efficient.

node/src/components/chain_synchronizer/operations.rs
goral09: > Also, nitpick: this would more normally be written item.validate(...) rather than BlockHeadersBatch::validate(&*item, ...).

I agree but we have the same method name (`validate`) on `Item` trait and it was resolving to that one instead of `BlockHeadersBatch`.

goral09: >Seems like we're calling validate here only to get the lowest block header from the batch. 

Not really, I also wanted to validate the batch from the storage. Although, if we agree that we should trust data from the DB I will remove that and simply get the lowest element.

goral09: Done in [`f46bbd2` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/f46bbd27aa8bf8d72a06a0b483d09e29ddc84132)

node/src/types/block.rs
goral09: Done in [`36ded88` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/36ded88067816ec81e9ed1af5e997cf372fb127c)

utils/retrieve-state/Cargo.toml
goral09: Done in [`c9cb619` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/c9cb6196ec98dcc8f347231080172be11d18479f)

node/src/types/block.rs
goral09: Done in [`36ded88` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/36ded88067816ec81e9ed1af5e997cf372fb127c)

node/src/types/block.rs
goral09: Done in [`36ded88` (#3020)](https://github.com/casper-network/casper-node/pull/3020/commits/36ded88067816ec81e9ed1af5e997cf372fb127c)

node/src/types/block.rs
goral09: I removed calls to `is_continuous_and_descending ` from all the places except for `Item for BlockHeadersBatch` which is used in the fetcher.

Fraser999 (approved):

TomVasile:
bors try

TomVasile:
bors try-

PR #438, opened 2020-10-27
EE-1079: enrich balance with merkle proofs, validate

henrytill:
Ref: https://casperlabs.atlassian.net/browse/EE-1079

Followed a similar approach to what we did with `query-state` here (using the purse lookup scheme as opposed to the query recursion), so nothing should be terribly surprising.

PR #3035, opened 2022-05-26
Avoid logging an error on dropping network-level responders

Fraser999:
This PR adds an auto-closing responder which sends a `None` response on being dropped.

It is used in the context of incoming and outgoing network messages, where the enqueued message might be dropped at an arbitrary point, for example when an outgoing network connection is dropped.

Without the auto-close responder, such a case would cause the normal responder type to log an error.  Logging an error is the correct course of action in all other contexts right now, as it could indicate a critical problem in the wireup between components.

However, with the recent backpressure changes to the network, it's now expected that outgoing messages could be dropped under certain conditions, and in that case the corresponding responders being dropped should not cause an error to get logged.

Closes [#438](https://github.com/casper-network/sre/issues/438).

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -99,7 +99,7 @@ use crate::{
     effect::{
         announcements::{BlocklistAnnouncement, ContractRuntimeAnnouncement},
         requests::{BeginGossipRequest, NetworkInfoRequest, NetworkRequest, StorageRequest},
-        EffectBuilder, EffectExt, Effects, Responder,
+        AutoClosingResponder, EffectBuilder, EffectExt, Effects,
     },
     reactor::{EventQueueHandle, Finalize, ReactorEvent},
     tls::{self, TlsCert, ValidationError},
@@ -416,7 +416,7 @@ where
         &self,
         dest: NodeId,
         msg: Arc<Message<P>>,
-        opt_responder: Option<Responder<()>>,
+        opt_responder: Option<AutoClosingResponder<()>>,
     ) {
         // Try to send the message.
         if let Some(connection) = self.outgoing_manager.get_route(dest) {
@@ -899,7 +899,7 @@ where
 
                         if respond_after_queueing {
                             self.send_message(*dest, Arc::new(Message::Payload(*payload)), None);
-                            responder.respond(()).ignore()
+                            responder.into_inner().respond(Some(())).ignore()
                         } else {
                             self.send_message(
                                 *dest,
@@ -913,7 +913,7 @@ where
                         // We're given a message to broadcast.
                         self.net_metrics.broadcast_requests.inc();
                         self.broadcast_message(Arc::new(Message::Payload(*payload)));
-                        responder.respond(()).ignore()
+                        responder.into_inner().respond(Some(())).ignore()
                     }
                     NetworkRequest::Gossip {
                         payload,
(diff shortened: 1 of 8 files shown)


Fraser999:
bors try

node/src/effect.rs
```diff
@@ -196,6 +192,48 @@ pub(crate) struct Responder<T> {
     is_shutting_down: SharedFlag,
 }
 
+/// A responder that will automatically send a `None` on drop.
+#[must_use]
+#[derive(DataSize, Debug)]
+pub(crate) struct AutoClosingResponder<T>(Responder<Option<T>>);
```
goral09, line 198: How about using an `enum`? 
```rust
enum ResponderResult<T> {
  Dropped,
  Ok(T),
}
```
? 

The `None` case of the `Option` has some well-defined meaning - absence of an item. Current encoding also makes it bit weird if `T = Option<M>`, we would get `Some(None)` or `Some(Some(_))` which in turn could encourage users to call `.flatten(_)` but in the process they're loosing information.

node/src/components/small_network.rs
```diff
@@ -928,7 +928,7 @@ where
                             count,
                             exclude,
                         );
-                        responder.respond(sent_to).ignore()
+                        responder.into_inner().respond(Some(sent_to)).ignore()
```
goral09: If we had `respond` defined on `AutoclosingResponder` we wouldn't need `into_inner()` or `Some(_)` and just pass in `responder.respond(_)`.

georgepisaltu: I agree with this. Additionally, we could also have a `respond_none`, `respond_empty` or any better name for when we want to respond with none and this way we could keep the same trait bounds and `fn respond()` signature, if that makes sense.

georgepisaltu (commented):
PR looks good, I left a comment above.

node/src/components/small_network.rs
marc-casperlabs: I kinda prefer the `into_inner` variant, for its explicitness. Just adding `respond` makes it a faux-interface like the containers in the rust stdlib, which I do not think is ideal. But I can be convinced otherwise :)

The `respond_none`/`respond_empty` is a bit tempting, but ultimately it makes it a little harder to understand, in my mind.

What do you guys think of renaming the newstruct, so it is not called "Responder"? Maybe `Autoclosing`?

goral09: I just don't understand why regular `Responder` can have `respond` but this one can't?

node/src/effect.rs
Fraser999: I tend to agree, but turns out we're using the `None` variant in a couple of places already where a function returns an error, and we don't have valid data to send via `respond()`.  So it would be inappropriate to replace `None` at such callsites with `ResponderResult::Dropped`.

I don't really like that, but I tried changing it to work "properly" by swapping the `Option<T>` for a `Result<T, AutoClosingResponderError<E>>`, and this affects way too much code to be included in this PR.

For now, I think it's best to leave this as is.

node/src/components/small_network.rs
Fraser999: I updated as per the majority opinion in [d397288](https://github.com/casper-network/casper-node/pull/3035/commits/d3972884433d962c74480ea4b470587f880e5c19).

rafal-ch (approved):
LGTM :+1:

goral09 (approved):

georgepisaltu (approved):
LGTM

Fraser999:
bors merge

PR #3037, opened 2022-05-26
Check no missing descendants before early exit from initial trie sync

Fraser999:
This PR fixes an issue where, if the node had partially synced the trie store and then stopped, the subsequent runs would not complete the initial trie store sync.

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -357,7 +357,7 @@ async fn fetch_and_store_block_header(
         });
     }
 
-    // We're querying storage directly and short-circuting here (before using the fetcher)
+    // We're querying storage directly and short-circuiting here (before using the fetcher)
     // as joiners don't talk to joiners and in a network comprised only of joining nodes
     // we would never move past the initial sync since we would wait on fetcher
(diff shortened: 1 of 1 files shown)


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -755,11 +755,21 @@ async fn sync_trie_store_worker(
...
+    // We're querying storage directly and short-circuiting here (before using the fetcher)
     // as joiners don't talk to joiners and in a network comprised only of joining nodes
     // we would never move past the initial sync since we would wait on fetcher
     // trying to query a peer for a trie but have no peers to query for the data.
-    if let Ok(Some(_trie)) = ctx.effect_builder.get_trie_full(state_root_hash).await {
+    if ctx
+        .effect_builder
+        .get_trie_full(state_root_hash)
+        .await?
+        .is_some()
```
goral09, line 766: Is this necessary? I think the second check should be enough.

goral09: Ah, it is. We could have all of the descendants but not the root itself.

goral09 (approved):

goral09:
bors merge

PR #3075, opened 2022-06-01
Work queue cleanup

marc-casperlabs:
Cleans up the work queue, attempting to make more invariants verifiable by the compiler.


goral09 (approved):

rafal-ch (approved):
LGTM :+1:

afck (approved):
(Not sure about that test, which really just belongs in Tokio. We could link to it — or do you think we need it to make sure the behavior doesn't change?)

marc-casperlabs:
> (Not sure about that test, which really just belongs in Tokio. We could link to it — or do you think we need it to make sure the behavior doesn't change?)

Most likely that the assumption is correct, so I'd keep it here. I guess it won't hurt anyone :)

PR #3084, opened 2022-06-06
Update storage to persist indices on disk

### LINKED ISSUE #2751: Eliminate Long Storage Indexing Time at Node Startup
Description: Every time the node starts, it creates indices of `block height -> block hash` and `era number -> block hash`.  These are expensive and take ~1.5 minutes on mainnet.

**Acceptance Criterion:** Either eliminate the indices altogether or eliminate the long wait time to create them for good.  We might need a one-time indexing to get rid of them though.

4.11.23 Ed's PR for blockchain_type removes this behavior, and renders this not necessary (enables lazy indexing). The feature that this supports is likely to move to sidecar, not node.

Fraser999:
This PR changes the storage component's indices to be held on disk (in the LMDB database) rather than in memory.

The initial migration should be a one-time cost at the next upgrade.  It has written to be resumable, with progress committed after completing the indices for every 10,000th block.

A test has been provided to allow for checking this against a pre-existing storage LMDB file.  The test is named `should_populate_indices_from_db_file` and reads the env var `CL_TEST_DB_DIR_PATH` to get the path to the DB file.

Example usage:
```
cd casper-node/node
export RUST_LOG=info
export CL_TEST_DB_DIR_PATH=<path to folder containing storage.lmdb>
cargo t --release should_populate_indices_from_db_file -- --nocapture
```

Initial testing using a release build against mainnet data indicates that the indexing performance is probably critically slow, but this should be confirmed using a production spec environment.

Closes #2751.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,7 +131,7 @@ lint-smart-contracts:
 
 .PHONY: audit-rs
 audit-rs:
-	$(CARGO) audit --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2020-0159
+	$(CARGO) audit --ignore RUSTSEC-2020-0071
 
 .PHONY: audit-as
 audit-as:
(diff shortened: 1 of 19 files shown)


Fraser999:
Closing for now, pending investigation into different approaches for the short-term.

PR #3085, opened 2022-06-06
Charge for deploys with incorrect WASM

rafal-ch:
This PR adds charging for execution of wasm files that are erroneous, due to:
* incorrect wasm magic bytes
* empty module (only in session, not payment)
* general corruption (bytes cannot be deserialized)
* no memory section

Closes https://github.com/casper-network/Governance/issues/42


rafal-ch:
~During review, please pay attention to [to this piece of code](https://github.com/casper-network/casper-node/pull/3085/files#diff-08b5ee8febf8ac6b81f2eaf6c72b213f0d6a25c200bb89611998807eb21fb642R178-R183).~

~I was unable to trigger a valid "stack overflow" error. All wasm code that I tried, including `make_n_arg_call_bytes()` from `gov_74.rs` resulted in an `Interpreter::Trap` error instead :thinking:~

The issue above has been clarified.

execution_engine_testing/tests/src/test/regression/gov_42.rs
```diff
@@ -0,0 +1,199 @@
...
+                wasm_bytes_no_memory_section.as_slice(),
+                "Memory section should exist".to_string(),
+                "Memory section should exist".to_string(),
+            ),
+            TestCase::new(
+                wasm_stack_overflow.as_slice(),
+                // TODO: wasm that overflows stack produces this kind of error (?)
+                "Interpreter error: trap: Trap { kind: Unreachable }".to_string(),
+                "Interpreter error: trap: Trap { kind: Unreachable }".to_string(),
+            ),
```
mpapierski: That's what I'm not sure about. Interpreter error returning a Trap { kind: Unreachable } is caused by actually executing the Wasm. It could be simply confirmed by checking the gas counter. Overflow by definition is a runtime error and not a static error happening before we can execute anything. What you could try however is to try to overflow the preprocessing stage here https://github.com/rafal-ch/casper-node//blob/74c77be8dbb73c6783d6c4f7d89e1053ecdad169/execution_engine/src/shared/wasm_prep.rs#L70-L92. Seems like you have test for `OperationForbiddenByGasRules`, what about `StackLimiter` error?

rafal-ch: @mpapierski the code has been updated according to the discussion, re-requesting the review.

execution_engine_testing/tests/src/test/regression/gov_42.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    // TODO:
+    // Introduce proper layers in wasm execution, for example:
+    // Layer 1. Parse (may throw errors related to malformed wasm bytes)
+
+    // Layer 2. Validation (may throw errors for parseable, but unsupported wasm files, for example,
+    // without memory section)
+
+    // Layer 3. Execution (may throw errors strictly related to
+    // execution for example, stack overflow)
```
Fraser999: I'd rather this weren't a todo list in the test.  Let's discuss and create tickets as required.

execution_engine_testing/tests/src/test/regression/gov_42.rs
```diff
@@ -0,0 +1,211 @@
...
+
+fn run_test_case(
+    TestCase {
+        name,
+        input_wasm_bytes,
+        expected_error,
+    }: &TestCase,
+    execution_phase: ExecutionPhase,
+) {
+    println!("Running test case '{}' in {:?}", name, execution_phase);
```
Fraser999: Left in by mistake?

Fraser999 (commented):
I think this is probably worth a mention in the EE changelog.

Other than that and the nitpicks, LGTM.

execution_engine_testing/tests/src/test/regression/gov_42.rs
rafal-ch: No, that's actually intentional. Thanks to this it's easier to see which test scenario failed.
An alternative would be to split the `should_charge_session_with_incorrect_wasm_files()` into dedicated tests, but this will introduce some copy&pasted boilerplate.

rafal-ch:
> I think this is probably worth a mention in the EE changelog.

Changelog updated here: b23a95d1a42d6ec4965e1af933c75dbe12cda1b7

execution_engine_testing/tests/src/test/regression/gov_42.rs
rafal-ch: Comments removed here: 4e18358bba6f4e10e4071536a2ad2d1d262db4c8

execution_engine_testing/tests/src/test/regression/gov_42.rs
darthsiroftardis: I'd prefer the boiler plate over the println

Fraser999: I'm a fan of not duplicating code, but I also like passing tests to not clutter the output.

Let's bring [`scopeguard`](https://github.com/bluss/scopeguard) in as a dev-dependency and do something like this:
```rust
    let _guard = scopeguard::guard_on_unwind((), |_| {
        println!(
            "Failed running test case '{}' in {:?}",
            name, execution_phase
        );
    });
```

execution_engine/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.  The format
 * Change both genesis and upgrade functions to write `ChainspecRegistry` under the fixed `Key::ChainspecRegistry`.
 * Lift the temporary limit of the size of individual values stored in global state.
 * Lift the temporary limit of the global maximum delegator capacity.
+* Providing incorrect WASM for execution will cause the 2.5CSPR to be charged.
```
```suggestion
* Providing incorrect Wasm for execution will cause the default 2.5CSPR to be charged.
```

execution_engine_testing/tests/src/test/regression/gov_42.rs
rafal-ch: > I'd prefer the boiler plate over the println

True, but in this specific case we'll have to create 10 separate tests to cover each "bad" wasm in both `Payment` and `Session`. See [here](https://github.com/rafal-ch/casper-node/compare/gov_42_zero_cost_deploys_with_bad_wasm_bytes...rafal-ch:refactor_tests_1?expand=1).

@Fraser999 The idea with `scopeguard` is neat, but we can also use the `test-case` I mentioned some time ago. Please look [here](https://github.com/rafal-ch/casper-node/compare/gov_42_zero_cost_deploys_with_bad_wasm_bytes...rafal-ch:refactor_tests_2?expand=1) and especially [here](https://github.com/rafal-ch/casper-node/compare/gov_42_zero_cost_deploys_with_bad_wasm_bytes...rafal-ch:refactor_tests_2?expand=1#diff-08b5ee8febf8ac6b81f2eaf6c72b213f0d6a25c200bb89611998807eb21fb642R59-R68). to see how well it allows to organize the repetitive tests.

rafal-ch: As agreed, the solution with splitting the test function manually was selected to keep consistency with how we approach similar cases in other places of the system. Fix [here](https://github.com/casper-network/casper-node/pull/3085/commits/f36a87b7c2a895c1e5251d6eb5a5cc3fabf320f0).

Fraser999 (approved):

mpapierski (approved):

PR #3089, opened 2022-06-07
Track time (and inflight request count) spent waiting on storage sempahore limiter.

goral09:
Also changes the bucket resolution for trie-access methods. Previously started at 10ms and tracked until 10s, now starts at 1ms and trakcs until 1s.


node/src/components/storage.rs
```diff
@@ -316,8 +328,12 @@ where
             // number of synchronous background tasks, especially since tokio does not put
             // meaningful bounds on these by default. For this reason, we limit the maximum number
             // of blocking tasks using a semaphore.
+            let start = Instant::now();
+            sync_task_limiter_inflight_counter.inc();
             let outcome = match sync_task_limiter.acquire().await {
                 Ok(_permit) =>{
+                    sync_task_limiter_inflight_counter.inc_by(-1.0);
```
Fraser999: Shouldn't we decrement in the error branch too?

node/src/components/storage/metrics.rs
```diff
@@ -0,0 +1,59 @@
...
+            "storage_sync_task_limiter_in_flight_count",
+            "number of currently awaiting tasks on the semaphore",
+        )?;
+        registry.register(Box::new(sync_task_limiter_in_flight_counter.clone()))?;
+
+        Ok(Metrics {
+            sync_task_limiter_in_flight_counter,
+            sync_task_limiter_waiting_millis: utils::register_histogram_metric(
+                registry,
+                "storage_sync_task_limiter",
```
Fraser999: Maybe more explicit?
```suggestion
                "storage_sync_task_limiter_waiting_millis",
```

node/src/components/storage.rs
```diff
@@ -360,7 +377,9 @@ impl Storage {
         finality_threshold_fraction: Ratio<u64>,
         last_emergency_restart: Option<EraId>,
         verifiable_chunked_hash_activation: EraId,
+        registry: &Registry,
```
Fraser999: I'd be inclined to make this optional (and hence also the `metrics`), so things like the tools don't need to import and construct a `Registry`.

node/src/components/storage.rs
goral09: Yeah, we might for correctness. It shouldn't matter that much though b/c the error is fatal and we exit the process eventually.

node/src/components/storage.rs
goral09: Sure, didn't think of that :( I'll make the change.

node/src/components/storage.rs
marc-casperlabs: It'd be better, in my mind, to have two counters - one for started and one for finished requests. The in-flight number is then the difference between the two. This shows bugs in the counter earlier and gives us the actual rate for free!

node/src/components/storage/metrics.rs
```diff
@@ -0,0 +1,59 @@
...
+
+/// Bucket count, with the last bucket going to +Inf which will not be included in the results.
+/// - start = 1, factor = 3, count = 10
+/// - start * factor ^ count = 1 * 3.0 ^ 10 = 19683millis
+/// - Values above 19.683s will not fall in a bucket that is kept.
+const EXPONENTIAL_BUCKET_COUNT: usize = 10;
+
+#[derive(Debug)]
+pub(crate) struct Metrics {
+    pub(crate) sync_task_limiter_in_flight_counter: Counter,
```
marc-casperlabs: See comment - above. Let's create two metrics? :)

marc-casperlabs (commented):
See comments about multiple metrics :).

node/src/components/storage.rs
goral09: I'll agree to that but mainly b/c current approach doesn't work with debug releases :)

> This shows bugs in the counter earlier and gives us the actual rate for free!

Grafana is interpreting this data and calculates rates. I don't see a point in reinventing it here.

node/src/components/storage/metrics.rs
goral09: Ah, forgot to update this one.

goral09: Addressed in [`bb58c84` (#3089)](https://github.com/casper-network/casper-node/pull/3089/commits/bb58c848473169841458f236d78596444da3817b)

node/src/components/storage.rs
goral09: Addressed in [`51bf941` (#3089)](https://github.com/casper-network/casper-node/pull/3089/commits/51bf941ccac22112ce0a335fe798413f01ac15ce). Although it required a little bit of work so please review.

node/src/components/storage.rs
goral09: @marc-casperlabs , addressed in [`7ab6c4e` (#3089)](https://github.com/casper-network/casper-node/pull/3089/commits/7ab6c4ef382e8051f56fb3548facb1917f16467d).

node/src/components/storage/metrics.rs
```diff
@@ -0,0 +1,100 @@
...
+        )?;
+
+        let sync_task_limiter_started_counter = Counter::new(
+            "storage_sync_task_limiter_started_count",
+            "number started tasks",
+        )?;
+        registry.register(Box::new(sync_task_limiter_started_counter.clone()))?;
+
+        let sync_task_limiter_completed_counter = Counter::new(
+            "storage_sync_task_limiter_completed_counter",
```
Fraser999:
```suggestion
            "storage_sync_task_limiter_completed_count",
```

Fraser999 (approved):
One nitpick to keep the metrics names' symmetric, but otherwise LGTM.

marc-casperlabs (approved):

PR #3093, opened 2022-06-08
Adds support for the new `speculative_exec` endpoint.

### LINKED ISSUE #2880: Add an endpoint on the node that will enable estimation of the execution cost
We've been asked many times in the past months (or maybe even more than a year) to give people a way to easily estimate gas cost. We have some plans to do that (like leveraging some of the fast-sync logic and extending it a bit) but they're all rather involving and won't be ready earlier than in a couple of months. I've been thinking whether there's some short-term, quick solution to this and I think I found it:

Let's add another endpoint on the node that will skip the gas accounting. It would be off by default, node operators should be able to easily limit the access to it (via iptables or other administratory tools) but anyone willing to sync A state (easy with fast sync) will be able to estimate the execution gas cost against MainNet state.

goral09:
### Adds JSON RPC endpoint executing a deploy and returning effects to the sender.

The new endpoint runs on a separate port from the previous JSON RPC. It can be configured in the `rpc_server` section of the `config.toml`. Setting `speculative_execution_qps_limit` to 0 will disable the endpoint. When sending a request to a node with disabled endpoint we get the following result:
```
* Could not resolve host: POST
* Closing connection 0
curl: (6) Could not resolve host: POST
*   Trying 127.0.0.1:25102...
* TCP_NODELAY set
* connect to 127.0.0.1 port 25102 failed: Connection refused
* Failed to connect to localhost port 25102: Connection refused
* Closing connection 1
curl: (7) Failed to connect to localhost port 25102: Connection refused
```

Endpoint accepts two parameters:
* block_identifier- this can be `BlockHash`, `StateRootHash` or null if deploy is to be executed on top of the latest block
* deploy - a deploy to execute.

(One can think of this as _"deploy dry run"_)

It will execute the deploy on top of the specified block and return the results of the execution to the caller. The effects of the execution won't be committed to the trie (blockchain database/GlobalState). Endpoint can be used for debugging, discovery - for example price estimation. If block_hash can't be found, proper error
will be returned. Any error returned during execution will be forwarded to the caller of the API.

### Example usage
1. Run nctl network - `nctl-compile && nctl-assets-setup && nctl-start`
2. Let it run for a while and then note hash of the recent block: `nctl-view-chain-lfb node=1`.
3. Stop node=1 (`nctl-stop node=1` and edit its config file to enable the new endpoint, its disabled by default. To do that, go to `utils/nctl/assets/net-1/nodes/node-1/config/1_0_0` and edit `speculative_qps_limit` in the `config.toml` to a positive number.
4. Start the node `nctl-start node=1 hash=<hash taken from nctl-view-chain-lfb node=2>`.
5. Create a deploy (in `casper-client-rs` directory), execute the command:
```
./target/release/casper-client make-transfer --chain-name "casper-net-1" --amount 1000 --secret-key ../casper-node/utils/nctl/assets/net-1/nodes/node-1/keys/secret_key.pem --target-account=01bf65587c501685ca5d0cb5293509fa4cecdb4503c7364084e18e14339e46a46f --transfer-id 1 --payment-amount 25000000 | jq -c
```
6. Send deploy to the node
```
curl -v POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","method":"speculative_exec","params":{"deploy": <previously store deploy formatted as JSON>, "block_identifier": <block_identifier> },"id":1}' 'http://localhost:25101/rpc'
```

where `block_identifier` can be a `{"Hash" : <block_hash> }`, `{"Height" : <block_height>}` or absent entirely (i.e. no `block_identifier` value), in which case it will execute deploy on top of the latest block.

7. Collect the results. If one wants to extract the cost, and the execution was a success, pipe the results into `jq` and extract the cost:
```
<previous command> | jq -c .result.execution_result.Success.cost
```

8. Sample response:
```json
{
  "jsonrpc":"2.0",
  "id":1,
  "result":{
    "api_version":"1.0.0",
    "block_hash":"9b97dab12815eeb84f244e011e1c998c3c4baa950be840256328cafae65d58c4",
    "execution_result":{
      "Success":{
        "cost":"100000000",
        "effect":{
          "operations":[],
          "transforms":[
            {"key":"hash-e256cf4b9f230655c2d1d35f1a36f2e2e0b26fd7cac7210aee941a46a13a1622","transform":"Identity"},
            {"key":"hash-0b87ad80a9a7cd6c9b8e1fc5c0eb61d1f9b9ae0229c34025a40061825abc8acb","transform":"Identity"},
            {"key":"hash-0b87ad80a9a7cd6c9b8e1fc5c0eb61d1f9b9ae0229c34025a40061825abc8acb","transform":"Identity"},
            // more transforms
          ],
         "transfers":["transfer-39cd3c67cc35bfd9faa831ad74f73d4be4b7eb2e8145ea4efb753c07946cecc5"]
}}}}
```

Closes https://github.com/casper-network/casper-node/issues/2880

### DIFF
--- a/node/src/components/rpc_server/http_server.rs
+++ b/node/src/components/rpc_server/http_server.rs
@@ -22,9 +22,12 @@ use super::{
         },
         RpcWithOptionalParams, RpcWithParams, RpcWithoutParams,
     },
-    ReactorEventT, RPC_API_PATH,
+    ReactorEventT,
 };
 use crate::effect::EffectBuilder;
+
+/// The URL path for all JSON-RPC requests.
+pub const RPC_API_PATH: &str = "rpc";
 
 /// Run the JSON-RPC server.
 pub(super) async fn run<REv: ReactorEventT>(
--- a/node/src/components/rpc_server/http_server.rs
+++ b/node/src/components/rpc_server/http_server.rs
@@ -1,11 +1,4 @@
-use std::{convert::Infallible, time::Duration};
-
-use http::header::ACCEPT_ENCODING;
 use hyper::server::{conn::AddrIncoming, Builder};
-use tokio::sync::oneshot;
-use tower::builder::ServiceBuilder;
-use tracing::{info, trace};
-use warp::Filter;
 
 use casper_json_rpc::RequestHandlersBuilder;
 use casper_types::ProtocolVersion;
@@ -28,6 +21,8 @@ use crate::effect::EffectBuilder;
 
 /// The URL path for all JSON-RPC requests.
 pub const RPC_API_PATH: &str = "rpc";
+
+pub const RPC_API_SERVER_NAME: &str = "JSON RPC";
 
 /// Run the JSON-RPC server.
 pub(super) async fn run<REv: ReactorEventT>(
@@ -59,32 +54,13 @@ pub(super) async fn run<REv: ReactorEventT>(
     QueryBalance::register_as_handler(effect_builder, api_version, &mut handlers);
     let handlers = handlers.build();
 
-    let make_svc = hyper::service::make_service_fn(move |_| {
-        let service_routes = casper_json_rpc::route(RPC_API_PATH, max_body_bytes, handlers.clone());
-
-        // Supports content negotiation for gzip responses. This is an interim fix until
-        // https://github.com/seanmonstar/warp/pull/513 moves forward.
-        let service_routes_gzip = warp::header::exact(ACCEPT_ENCODING.as_str(), "gzip")
-            .and(service_routes.clone())
-            .with(warp::compression::gzip());
-
-        let service = warp::service(service_routes_gzip.or(service_routes));
-        async move { Ok::<_, Infallible>(service.clone()) }
-    });
-
-    let make_svc = ServiceBuilder::new()
-        .rate_limit(qps_limit, Duration::from_secs(1))
-        .service(make_svc);
-
-    let server = builder.serve(make_svc);
-    info!(address = %server.local_addr(), "started JSON-RPC server");
-
-    let (shutdown_sender, shutdown_receiver) = oneshot::channel::<()>();
-    let server_with_shutdown = server.with_graceful_shutdown(async {
-        shutdown_receiver.await.ok();
-    });
-
-    let _ = tokio::spawn(server_with_shutdown).await;
-    let _ = shutdown_sender.send(());
-    trace!("JSON-RPC server stopped");
+    super::rpcs::run(
+        builder,
+        handlers,
+        qps_limit,
+        max_body_bytes,
+        RPC_API_PATH,
+        RPC_API_SERVER_NAME,
+    )
+    .await;
 }
--- a/node/src/components/rpc_server/rpcs/speculative_exec.rs
+++ b/node/src/components/rpc_server/rpcs/speculative_exec.rs
@@ -1,0 +1,151 @@
+//! RPC related to speculative execution.
+
+// TODO - remove once schemars stops causing warning.
+#![allow(clippy::field_reassign_with_default)]
+
+use std::str;
+
+use async_trait::async_trait;
+use casper_execution_engine::core::engine_state;
+use casper_json_rpc::ReservedErrorCode;
+use once_cell::sync::Lazy;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use casper_types::{ExecutionResult, ProtocolVersion};
+
+use super::{
+    docs::{DocExample, DOCS_EXAMPLE_PROTOCOL_VERSION},
+    Error, ErrorCode, ReactorEventT, RpcWithParams,
+};
+use crate::{
+    effect::{requests::RpcRequest, EffectBuilder},
+    reactor::QueueKind,
+    types::{Block, BlockHash, Deploy},
+};
+
+static SPECULATIVE_EXEC_PARAMS: Lazy<SpeculativeExecParams> = Lazy::new(|| SpeculativeExecParams {
+    block_hash: *Block::doc_example().hash(),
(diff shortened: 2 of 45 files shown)


goral09:
@Fraser999 @EdHastingsCasperLabs @darthsiroftardis , I'm open to naming suggestions as I'm not 100% happy with the current ones :)

node/src/components/rpc_server/rpcs/speculative_exec.rs
```diff
@@ -0,0 +1,151 @@
...
+    api_version: DOCS_EXAMPLE_PROTOCOL_VERSION,
+    execution_result: ExecutionResult::example().clone(),
+});
+
+/// Params for "speculative_exec" RPC request.
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct SpeculativeExecParams {
+    /// Block hash on top of which to execute the deploy.
+    pub block_hash: BlockHash,
```
Fraser999: I'd be inclined to make this an `Option<BlockIdentifier>` where `None` means to use the highest block.  That's in keeping with most of our other RPCs which require a block or state root hash (and those that differ right now should be corrected for 2.0.0).

I guess if we did that, we should include further info in the `SpeculativeExecResult`, i.e. the block hash, height and state root hash actually used.

resources/local/config.toml
```diff
@@ -233,6 +233,14 @@ qps_limit = 100
 # Maximum number of bytes to accept in a single request body.
 max_body_bytes = 2_621_440
 
+# Address to which the speculative execution endpoint binds to.
+speculative_execution_address = "0.0.0.0:7778"
+
+# The global max rate of requests (per second) before they are limited.
+# Request will be delayed to the next 1 second bucket once limited.
+# Setting to 0 disabled the endpoint.
```
Fraser999:
```suggestion
# Setting to 0 disables the endpoint.
```
Same applies to all the config.tomls

node/src/components/rpc_server/speculative_exec_server.rs
```diff
@@ -0,0 +1,62 @@
...
+pub(super) async fn run<REv: ReactorEventT>(
+    builder: Builder<AddrIncoming>,
+    effect_builder: EffectBuilder<REv>,
+    api_version: ProtocolVersion,
+    qps_limit: u64,
+    max_body_bytes: u32,
+) {
+    let mut handlers = RequestHandlersBuilder::new();
+    SpeculativeExec::register_as_handler(effect_builder, api_version, &mut handlers);
+    let handlers = handlers.build();
```
Fraser999, line 27: Looks like the only difference between this `run` and `http_server::run` is the set of handlers.  I think it'd be good to deduplicate the shared code.

Fraser999 (commented):
I think we should look to include the new endpoint in the `ListRpcs::do_handle_request`, but only if the qps is non-zero.

Currently `ListRpcs` uses the static `OPEN_RPC_SCHEMA` to form its response as it's not dynamic.  Possibly the easiest way to handle this is to have two statics, one including the new endpoint and the other without.  And likewise, two separate `ListRpcs` - say the current one and a new one `ListRpcsWithSpeculativeExec`.  We choose between which to use when we're registering them as handlers in `http_server::run`.

node/src/components/rpc_server/rpcs/speculative_exec.rs
goral09: Agree and will do.

node/src/components/rpc_server/speculative_exec_server.rs
goral09: Agree and will do.

resources/local/config.toml
goral09: Done in [`4fd28e3` (#3093)](https://github.com/casper-network/casper-node/pull/3093/commits/4fd28e3f0d172c325745c7fa2cbe98f35fa08a75)

node/src/components/rpc_server/speculative_exec_server.rs
goral09: Done in [`6f7b4d5` (#3093)](https://github.com/casper-network/casper-node/pull/3093/commits/6f7b4d5dfcde77cc311432f4219cb51ff3e72190)

node/src/components/rpc_server/rpcs/speculative_exec.rs
goral09: Ah, I miscommitted and this is part of https://github.com/casper-network/casper-node/pull/3093/commits/4fd28e3f0d172c325745c7fa2cbe98f35fa08a75.

Fraser999 (approved):

goral09:
bors try

goral09:
bors cancel

node/src/components/rpc_server/http_server.rs
```diff
@@ -56,32 +54,13 @@ pub(super) async fn run<REv: ReactorEventT>(
...
-
-    let (shutdown_sender, shutdown_receiver) = oneshot::channel::<()>();
-    let server_with_shutdown = server.with_graceful_shutdown(async {
-        shutdown_receiver.await.ok();
-    });
-
-    let _ = tokio::spawn(server_with_shutdown).await;
-    let _ = shutdown_sender.send(());
-    trace!("JSON-RPC server stopped");
+    super::rpcs::run(
```
EdHastingsCasperAssociation, line 57: Out of curiosity, why do we want to extract loc 59-86 to a new method vs leave it inline here?

EdHastingsCasperAssociation: never mind, i see it getting reused below.

EdHastingsCasperAssociation (approved):
lgtm, great job

goral09:
bors try

goral09:
bors try

PR #3098, opened 2022-06-10
Move compute_rewards from highway to core.

afck:
Whether rewards should be computed is not Highway-specific, so the setting is moved to the `[core]` section.
The era supervisor also enforces this now, by zeroing the rewards in every finalized block.

Part of https://github.com/CasperLabs/product/issues/56.


Fraser999 (approved):

afck:
bors merge

mpapierski (approved):

afck:
bors cancel

PR #3100, opened 2022-06-13
Bradjohnl gh sre 486 new nctl test w custom override

bradjohnl:
Adding tests with the custom overrides to this PR directly so that Michal can test using bors.
This change introduce some duplication in the test overrides, but it is making use of the current approach implemented in the nightly-test.sh logic.
Doing it in another way would have required to rework the CI script logic and it would have risked going out of scope and would have required more time.


mpapierski:
bors try

bradjohnl:
Changes have been cherry-picked in the feat-private-chain branch

PR #3106, opened 2022-06-14
Initial support for private chain

mpapierski:
This is backport of original #3029 on top of v1.4.6 with least amount of extra backported changes to make the original code from PR unchanged.

This PR brings preliminary support for private chains by introducing multiple chainspec configuration flags that change the behavior of a chain.

These settings are orthogonal to each other; any particular combination (and all other pre-existing settings) is valid for a given network, for a wide variety of possible permutations. 

- Administrator accounts: by specifying accounts in a `[[administrators]]` section of `accounts.toml` accounts with special privileges are created and later referred to as 'administrative accounts'
- `core.allow_auction_bids` disables add_bid, delegate entrypoints in auction contract if set to true
- `core.compute_rewards` rewards 0 tokens after each switch block. This means no new tokens are minted and validators aren't gaining more weight.
- `core.allow_unrestricted_transfers` if set to false, the normal user can't send tokens to other normal users; the normal user can transfer only to the administrator; either source or target has to be an admin.
- `core.refund_handling` defines how refunds are computed. 
  - `refund_handling = { type = "refund", refund_ratio = [1, 1] }` means 100% of `payment_amount - gas_spent` is refunded. 
  - `refund_handling = { type = "refund", refund_ratio = [0, 100 ] }` doesn't refund anything
  - `Refund` mode requires that the fraction specified is proper fraction (that is `refund_ratio <= 1`)
- `core.fee_handling` defines how gas spent is processed after deploy is executed.
  - `fee_handling = { type = "accumulate" }` creates special rewards purse in the mint system contract where fees are accumulated into
  - `fee_handling = { type = "pay_to_proposer" }` is a default setting for a public chain - rewards are transferred to block's proposer.
  - `fee_handling = { type = "burn" }` fees are burned.


Fraser999 (approved):

mpapierski:
bors try

PR #3110, opened 2022-06-15
NCTL: Add timeout to waiting for genesis in upgrade scenarios

TomVasile:
Changes:
- `do_await_genesis_era_to_complete`
    - added timeout to command 
        - prevents hanging forever in CI
        - will allow CI to dump the logs
    - maintained backwards compatibility for non-upgrade tests

Adds:
- `do_wait_until_era`
    - utilized by `upgrade_scenario_09.sh` which doesn't use `do_await_genesis_era_to_complete`

Ticket: https://app.zenhub.com/workspaces/cn-sre-60d363546afb85000e491ff6/issues/casper-network/sre/494

### DIFF
--- a/utils/nctl/sh/scenarios/common/itst.sh
+++ b/utils/nctl/sh/scenarios/common/itst.sh
@@ -46,9 +46,43 @@ function log_step() {
 }
 
 function do_await_genesis_era_to_complete() {
-    log_step "awaiting genesis era to complete"
+    local LOG_STEP=${1:-'true'}
+    local TIMEOUT=${2:-'120'}
+
+    if [ "$LOG_STEP" = "true" ]; then
+        log_step "awaiting genesis era to complete: timeout=$TIMEOUT"
+    fi
+
     while [ "$(get_chain_era)" -lt "1" ]; do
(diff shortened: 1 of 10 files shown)


sacherjj (approved):

utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -46,9 +46,43 @@ function log_step() {
 }
 
 function do_await_genesis_era_to_complete() {
-    log_step "awaiting genesis era to complete"
+    local LOG_STEP=${1:-'true'}
+    local TIMEOUT=${2:-'120'}
```
goral09, line 50: That's 2 minutes, correct? How long are eras in these tests usually by default?

TomVasile: `2mins` yes. Locally it was consistently completing in about `60s`. Gave it a little extra buffer.

`era_duration` is `41s` for most. ~~But thanks for the question, reminds me of two more tests I need to extend the timeout on!~~

Turns out not necessary in the other tests.

PR #3113, opened 2022-06-16
Provide remote address in SSE log messages

Fraser999:
This PR adds logging of the socket address of the client if known in the cases of:
1. SSE server rejecting a new client as it has exceeded the `max_concurrent_subscribers` limit
2. SSE server dropping a client connection due to the client lagging

Example log lines are:
```json
{"timestamp":"Jun 16 00:38:31.020","level":"INFO","fields":{"message":"event stream server has max subscribers: rejecting new one","remote_address":"127.0.0.1:51408","max_concurrent_subscribers":"100"},"target":"casper_node::components::event_stream_server::sse_server"}
{"timestamp":"Jun 16 00:48:34.992","level":"INFO","fields":{"message":"client lagged: dropping event stream connection to client","remote_address":"127.0.0.1:39112","lagged_count":"24"},"target":"casper_node::components::event_stream_server::sse_server"}
```

### DIFF
--- a/node/src/components/event_stream_server/sse_server.rs
+++ b/node/src/components/event_stream_server/sse_server.rs
@@ -2,6 +2,7 @@
 
 use std::{
     collections::{HashMap, HashSet},
+    net::SocketAddr,
     sync::{Arc, RwLock},
 };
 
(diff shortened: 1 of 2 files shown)


sacherjj (approved):

Fraser999:
bors merge

PR #3114, opened 2022-06-16
Revert breaking changes in `casper-test-support` and tidy up `casper-execution-engine`

### LINKED ISSUE #3072: Revert breaking API changes in casper-ee

georgepisaltu:
Fixes #3072 #3074 

This PR reverts the breaking changes in `casper-test-support` since the v1.4.6 release.
The changes in `casper-execution-engine` cannot be reverted in a backwards compatible way, so this PR adds `non-exhaustive` to enums and restores `write_scratch_to_db` in a deprecated state.

Changes were found using the `cargo-public-api` tool.

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -147,7 +147,7 @@ impl EngineState<LmdbGlobalState> {
     }
 
     /// Writes state cached in an EngineState<ScratchEngineState> to LMDB.
-    pub fn write_scratch_to_lmdb(
+    pub fn write_scratch_to_db(
         &self,
         state_root_hash: Digest,
         scratch_global_state: ScratchGlobalState,
@@ -156,16 +156,6 @@ impl EngineState<LmdbGlobalState> {
         self.state
             .put_stored_values(CorrelationId::new(), state_root_hash, stored_values)
             .map_err(Into::into)
-    }
-
-    /// Writes state cached in an EngineState<ScratchEngineState> to LMDB.
-    #[deprecated(since = "2.0.0", note = "renamed to `write_scratch_to_lmdb`")]
-    pub fn write_scratch_to_db(
-        &self,
-        state_root_hash: Digest,
-        scratch_global_state: ScratchGlobalState,
-    ) -> Result<Digest, Error> {
-        self.write_scratch_to_lmdb(state_root_hash, scratch_global_state)
     }
 }
 
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -157,6 +157,16 @@ impl EngineState<LmdbGlobalState> {
             .put_stored_values(CorrelationId::new(), state_root_hash, stored_values)
             .map_err(Into::into)
     }
+
+    /// Writes state cached in an EngineState<ScratchEngineState> to LMDB.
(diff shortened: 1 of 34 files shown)


execution_engine/src/core/engine_state/mod.rs
```diff
@@ -157,6 +157,16 @@ impl EngineState<LmdbGlobalState> {
             .put_stored_values(CorrelationId::new(), state_root_hash, stored_values)
             .map_err(Into::into)
     }
+
+    /// Writes state cached in an EngineState<ScratchEngineState> to LMDB.
+    #[deprecated(since = "2.0.0", note = "renamed to `write_scratch_to_lmdb`")]
+    pub fn write_scratch_to_db(
```
Fraser999: I think rather than re-adding this as a deprecated function, we should just rename `write_scratch_to_lmdb` to `write_scratch_to_db` since that is how we'll be naming things as we go forward, moving away from LMDB.

execution_engine_testing/test_support/src/lib.rs
```diff
@@ -65,6 +65,12 @@ pub const DEFAULT_CHAIN_NAME: &str = "casper-execution-engine-testing";
 pub const DEFAULT_GENESIS_TIMESTAMP_MILLIS: u64 = 0;
 /// Default maximum number of associated keys.
 pub const DEFAULT_MAX_ASSOCIATED_KEYS: u32 = 100;
+/// Default max serialized size of `StoredValue`s.
+#[deprecated(
+    since = "2.2.0",
```
Fraser999: Should probably be since 2.3.0 as that's going to be the next published version.  Same comment applies to all deprecation notices added to this crate in this PR.

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -412,6 +412,12 @@ impl LmdbWasmTestBuilder {
         self
     }
 
+    /// Commit scratch to global state, and reset the scratch cache.
+    #[deprecated(since = "2.2.0", note = "renamed to `write_scratch_to_lmdb`")]
```
Fraser999: As above, this function should not be deprecated - it should just replace `write_scratch_to_lmdb`.

execution_engine/src/core/engine_state/mod.rs
goral09: Wasn't the whole point to not "remove" the public method? Renaming it is equivalent in a sense that devs have to change their code.

georgepisaltu: The 1.4.6 name was `write_scratch_to_db`. Currently on `dev`, that function is now called `write_scratch_to_lmdb`, so from the users' point of view, `write_scratch_to_db` doesn't exist anymore. The proposed change reinstated `write_scratch_to_db` as deprecated, but now with @Fraser999 's suggestion, we just reverted the rename and avoided the breaking change.

execution_engine_testing/test_support/src/lib.rs
georgepisaltu: Done.

execution_engine_testing/test_support/src/wasm_test_builder.rs
georgepisaltu: Done.

georgepisaltu (commented):
New pushed changes have been tested to not have any breaking changes in `test_support`.

Fraser999 (approved):

Fraser999:
bors merge

PR #3116, opened 2022-06-17
Make sync_trie_store a noop if already synced

### LINKED ISSUE #3118: sync_trie_store takes excessive time with testnet snapshot

Fraser999:
This PR addresses the issue of `sync_trie_store` taking excessively long.

It uses the reported block range at startup as an indication of which tries it already has locally and skips syncing any in that range.  Further work needs to be done to ensure that the available block range is always accurate.

The PR also adds an info level log indicating if it skips syncing the trie store under a given block.  This log line could be printed excessively frequently, but if that is the case, it points to further places where we can probably optimize by skipping syncing a whole range of blocks.

Closes #3118.

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -768,7 +768,27 @@ async fn sync_trie_store_worker(
 }
 
 /// Synchronizes the trie store under a given state root hash.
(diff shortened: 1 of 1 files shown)


goral09 (approved):

Fraser999:
bors merge

PR #3127, opened 2022-06-21
NCTL: timeout nctl upgrade steps

TomVasile:
Changes:
- adds additional args to `await_n_eras.sh` 
- updates nctl binary upgrade tests to use `nctl-await-n-eras` which is an alias of `await_n_eras.sh`
- adds timeout to `await_n_eras` function and maintain backwards compatibility where possible

Ticket: https://app.zenhub.com/workspaces/cn-sre-60d363546afb85000e491ff6/issues/casper-network/sre/506

### DIFF
--- a/utils/nctl/sh/scenarios-upgrades/upgrade_scenario_05.sh
+++ b/utils/nctl/sh/scenarios-upgrades/upgrade_scenario_05.sh
@@ -138,7 +138,7 @@ function _step_03()
     done
 
     log "... awaiting 2 eras + 1 block"
-    await_n_eras '2' 'true' '5.0' '2'
+    nctl-await-n-eras offset='2' sleep_interval='5.0' timeout='180' node_id='2'
     await_n_blocks '1' 'true' '2'
 }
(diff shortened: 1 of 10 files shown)


utils/nctl/sh/scenarios-upgrades/upgrade_scenario_05.sh
```diff
@@ -138,7 +138,7 @@ function _step_03()
     done
 
     log "... awaiting 2 eras + 1 block"
-    await_n_eras '2' 'true' '5.0' '2'
+    nctl-await-n-eras offset='2' sleep_interval='5.0' timeout='180' node_id='2'
```
goral09, line 141: Personally, I'm not sure whether the `nctl` alias makes much sense. All `nctl-*` commands available from the shell are "useful" for CLI users to interact with the network. I can't imagine a case where one would use `nctl-await-n-eras` there. Not a blocker obviously.

goral09 (approved):

PR #3131, opened 2022-06-23
Block completed announcements

### LINKED ISSUE #3128: Available block range does not correctly reflect actually available block range
The current functions used to indicate available blocks do not take completed blocks into account, but consider blocks complete as soon as the headers are written. This is wrong and should instead depend on the completion of global state synchronization.

marc-casperlabs:
Closes #3128.

This restores the `disjoint_sequences` module in `storage` and extends it with some constructors and `ToBytes` serialization support. The disjoint sequences are then used to track completed blocks in storage (a block is complete if all of its deploys are stored and global state contains its root hash with all dependencies), which is used in the API and will soon be used in extensions to the fast sync code as well.

Completed blocks were inaccurately tracked using the `highest_block_at_startup` and `lowest_available_block_height` attributes in storage, which had issues not being accurate, often not taking the state of global state into account and also supporting only a single possible range of blocks.

With fast sync, blocks can be completed out of order, but form larger ranges later, thus the restoration of the disjoint sequences.


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -863,7 +863,7 @@ async fn fast_sync(ctx: &ChainSyncContext<'_>) -> Result<(KeyBlockInfo, BlockHea
     sync_trie_store(&most_recent_block_header, ctx).await?;
 
     ctx.effect_builder
-        .update_lowest_available_block_height_in_storage(most_recent_block_header.height())
+        .mark_block_completed(most_recent_block_header.height())
         .await;
```
Fraser999: I think this should be removed, as we haven't fetched the deploys for this block at this stage.

node/src/effect.rs
```diff
@@ -820,6 +821,25 @@ impl<REv> EffectBuilder<REv> {
             .await;
     }
 
+    /// Request that a block with a specific height be marked completed.
+    ///
+    /// Completion means that the block itself (along with its header) and all of its deploys have
+    /// been persisted to storage and its global state root hash is missing no dependencies in the
+    /// contract runtime.
```
goral09: I would say _global state_ and not _contract runtime_

node/src/components/chain_synchronizer/operations.rs
marc-casperlabs: Hmm, we really don't have any complete block after fast syncing (but before forward-execution), do we?

Fraser999: No, we don't

Fraser999 (approved):

node/src/components/storage.rs
```diff
@@ -490,14 +479,43 @@ impl Storage {
...
+                //    caused the creation of the completed blocks index, thus would not have
+                //    resulted in a `None` value here.
+                //
+                // For this reason, it is safe to assume that, when handling the `None` case, all
+                // blocks in storage are complete blocks.
+                if let Some(&highest_block) = component.block_height_index.keys().last() {
+                    component.completed_blocks =
+                        DisjointSequences::new(Sequence::new(0, highest_block));
+                    component.persist_completed_blocks()?;
+                } // the `else` case here would mean genesis, so no change.
```
goral09, line 514: This is true if we didn't fail to write previous `DisjoinSequences` - i.e. if this software failed to do that but already has the `block_height_index` then it will incorrectly update the `DisjoinSequence` as well.

goral09 (approved):

node/src/components/storage.rs
marc-casperlabs: If a failed write has corrupted the previous info, I'm not sure there's anything we *can* do though. Or is there a case where we should be able to gracefully recover from a failed write?

marc-casperlabs: (reindexing everything was something that we discarded as too expensive, instead we rely on the logic outlined in the length comment that leads us to assuming we are complete in versions pre 1.5.0)

PR #3132, opened 2022-06-23
Warn if messages are unsafe for joiners

marc-casperlabs:
Adds a warning (actually an `error`, because this should never happen) if a `TrieRequest` is ever sent to a joiner. Since there is no guarantee this will lock up the connection, we will still attempt to send and hope for the best.

The fast sync code's peer selection should prevent this error from ever occuring, this is purely a (loud) safety measure.

### DIFF
--- a/node/src/components/small_network/tasks.rs
+++ b/node/src/components/small_network/tasks.rs
@@ -672,17 +672,19 @@ pub(super) async fn message_sender<P>(
     mut sink: SplitSink<FullTransport<P>, Arc<Message<P>>>,
     limiter: Box<dyn LimiterHandle>,
     counter: IntGauge,
-    joiner_id: Option<(SocketAddr, NodeId)>,
+    remote_joiner_id: Option<(SocketAddr, NodeId)>,
 ) where
     P: Payload,
 {
     while let Some((message, opt_responder)) = queue.recv().await {
         counter.dec();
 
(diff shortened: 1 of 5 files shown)


Fraser999 (approved):

node/src/components/small_network/tasks.rs
```diff
@@ -672,12 +672,20 @@ pub(super) async fn message_sender<P>(
...
+    joiner_id: Option<(SocketAddr, NodeId)>,
 ) where
     P: Payload,
 {
     while let Some((message, opt_responder)) = queue.recv().await {
         counter.dec();
 
+        if let Some((ref addr, ref node_id)) = joiner_id {
+            if message.payload_is_unsafe_for_joiners() {
+                // We should never receive this kind of message here.
```
goral09: This may be confusing in my opinion as we use `receive` and `sender` (in the method name) but their meaning is different. Here, _receive_ means that a task _in the node_ wants to send something to a different node.

Can we change the comment here to, for example: _we should never send this kind of message to joiner_ . 

The `joiner_id` argument in the method may also be confusing to a reader - it's not _our_ id but the receiving end.

goral09 (approved):

node/src/components/small_network/tasks.rs
marc-casperlabs: I'll update the comment and change the name!

PR #3140, opened 2022-06-24
GH-3097: Disabled contracts

### LINKED ISSUE #3097: Calling contract with StoredContractByHash after using storage::disable_contract_version should raise error
Contract API storage::disable_contract_version takes contract_package_hash, and contract_hash, and even after successfully disabling a contract by it’s hash it’s still callable by StoredContractByHash/StoredContractByName variant. Calling same disabled contract through StoredVersionedContractByHash does properly fail as expected. 

We need a regression test to prove this bug, and a fix.

mpapierski:
Closes #3097

This PR makes the contract execution flow easier to follow by moving most of the logic shared between call_contract and call_versioned_contract into execute_contract. It decreases duplication and allows for an extra validation step that will reject calls to disabled contract hashes - when a disabled contract is now called, a special error `DisabledContract(contract_hash)` will be raised.

### DIFF
--- a/execution_engine/src/core/execution/error.rs
+++ b/execution_engine/src/core/execution/error.rs
@@ -170,6 +170,9 @@ pub enum Error {
     /// The runtime stack is `None`.
     #[error("Runtime stack missing")]
     MissingRuntimeStack,
+    /// Contract is disabled.
+    #[error("Contract is disabled")]
+    DisabledContract(ContractHash),
 }
 
 impl From<wasm_prep::PreprocessingError> for Error {
--- a/execution_engine/src/core/runtime/mod.rs
+++ b/execution_engine/src/core/runtime/mod.rs
@@ -1143,28 +1143,14 @@ where
     ) -> Result<CLValue, Error> {
         let (contract, contract_hash, contract_package) = match identifier {
             CallContractIdentifier::Contract { contract_hash } => {
-                let key = contract_hash.into();
-                let contract = match self.context.read_gs(&key)? {
-                    Some(StoredValue::Contract(contract)) => contract,
-                    Some(_) => {
-                        return Err(Error::InvalidContract(contract_hash));
-                    }
-                    None => return Err(Error::KeyNotFound(key)),
-                };
-
-                let contract_package_hash = contract.contract_package_hash();
-                let contract_package_key = contract_package_hash.into();
-                let contract_package = match self.context.read_gs(&contract_package_key)? {
-                    Some(StoredValue::ContractPackage(contract_package)) => contract_package,
-                    Some(_) => {
-                        return Err(Error::InvalidContractPackage(contract_package_hash));
-                    }
-                    None => return Err(Error::KeyNotFound(contract_package_key)),
-                };
+                let contract_key = contract_hash.into();
+                let contract: Contract = self.context.read_gs_typed(&contract_key)?;
+                let contract_package_key = Key::from(contract.contract_package_hash());
+                let contract_package: ContractPackage =
+                    self.context.read_gs_typed(&contract_package_key)?;
 
                 // System contract hashes are disabled at upgrade point
-                let is_calling_system_contract =
-                    self.is_system_contract(Key::from(contract_hash))?;
+                let is_calling_system_contract = self.is_system_contract(contract_key)?;
 
                 // Check if provided contract hash is disabled
                 let is_contract_enabled = contract_package.is_contract_enabled(&contract_hash);
@@ -1179,14 +1165,10 @@ where
                 contract_package_hash,
                 version,
             } => {
-                let contract_package_key = contract_package_hash.into();
-                let contract_package = match self.context.read_gs(&contract_package_key)? {
-                    Some(StoredValue::ContractPackage(contract_package)) => contract_package,
-                    Some(_) => {
-                        return Err(Error::InvalidContractPackage(contract_package_hash));
-                    }
-                    None => return Err(Error::KeyNotFound(contract_package_key)),
-                };
+                let contract_package_key = Key::from(contract_package_hash);
+                let contract_package: ContractPackage =
+                    self.context.read_gs_typed(&contract_package_key)?;
+
                 let contract_version_key = match version {
                     Some(version) => ContractVersionKey::new(
                         self.context.protocol_version().value().major,
@@ -1205,13 +1187,7 @@ where
                     .ok_or(Error::InvalidContractVersion(contract_version_key))?;
 
                 let contract_key = contract_hash.into();
-                let contract = match self.context.read_gs(&contract_key)? {
-                    Some(StoredValue::Contract(contract)) => contract,
-                    Some(_) => {
-                        return Err(Error::InvalidContract(contract_hash));
-                    }
-                    None => return Err(Error::KeyNotFound(contract_key)),
-                };
+                let contract: Contract = self.context.read_gs_typed(&contract_key)?;
 
                 (contract, contract_hash, contract_package)
             }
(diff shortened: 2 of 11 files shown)


execution_engine/src/core/execution/error.rs
```diff
@@ -170,6 +170,9 @@ pub enum Error {
     /// The runtime stack is `None`.
     #[error("Runtime stack missing")]
     MissingRuntimeStack,
+    /// Contract is disabled.
+    #[error("Contract is disabled")]
+    DisabledContract(ContractHash),
```
georgepisaltu, line 175: Does it help users debug fails if we show the contract hash in the error message?

georgepisaltu (commented):
Looks good, just one question.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -1196,12 +1137,92 @@ where
...
+                }
+
+                (contract, contract_hash, contract_package)
+            }
+            CallContractIdentifier::ContractPackage {
+                contract_package_hash,
+                version,
+            } => {
+                let contract_package_key = contract_package_hash.into();
+                let contract_package = match self.context.read_gs(&contract_package_key)? {
```
darthsiroftardis: I have a thought here, if I understand correctly, the `lock_status` here would tell us if a contract package is locked, we could then exit early if the `ContractVersion` is higher than 1?

darthsiroftardis (commented):
Overall code looks fine, I have one outstanding question about a possible early exit, will approve once that has been resolved

execution_engine/src/core/runtime/mod.rs
```diff
@@ -1196,12 +1137,92 @@ where
...
+                    Some(StoredValue::Contract(contract)) => contract,
+                    Some(_) => {
+                        return Err(Error::InvalidContract(contract_hash));
+                    }
+                    None => return Err(Error::KeyNotFound(key)),
+                };
+
+                let contract_package_hash = contract.contract_package_hash();
+                let contract_package_key = contract_package_hash.into();
+                let contract_package = match self.context.read_gs(&contract_package_key)? {
```
rafal-ch: `execute_contracts` has stronger validation requirements than the `call_contract` had before. For example, in `call_contract` we didn't validate the `contract_package_key`.

Maybe it's perfectly ok, but I just wanted to bring this to your attention.

execution_engine/src/core/runtime/mod.rs
```diff
@@ -1196,12 +1137,92 @@ where
...
-        contract_package: ContractPackage,
-        contract_hash: ContractHash,
-        contract: Contract,
-        entry_point: EntryPoint,
+        identifier: CallContractIdentifier,
+        entry_point_name: &str,
         args: RuntimeArgs,
     ) -> Result<CLValue, Error> {
+        let (contract, contract_hash, contract_package) = match identifier {
+            CallContractIdentifier::Contract { contract_hash } => {
```
rafal-ch, line 1145: These two match arms (`::Contract` and `::ContractPackage`) still has the following code in common. Maybe consider moving it in front of the match call.

```
                let key = contract_hash.into();
                let contract = match self.context.read_gs(&key)? {
                    Some(StoredValue::Contract(contract)) => contract,
                    Some(_) => {
                        return Err(Error::InvalidContract(contract_hash));
                    }
                    None => return Err(Error::KeyNotFound(key)),
                };

                let contract_package_key = contract_package_hash.into();
                let contract_package = match self.context.read_gs(&contract_package_key)? {
                    Some(StoredValue::ContractPackage(contract_package)) => contract_package,
                    Some(_) => {
                        return Err(Error::InvalidContractPackage(contract_package_hash));
                    }
                    None => return Err(Error::KeyNotFound(contract_package_key)),
                };
```

execution_engine_testing/tests/src/test/regression/gh_3097.rs
```diff
@@ -0,0 +1,354 @@
...
+        .exec(call_by_hash_v1_request)
+        .expect_failure()
+        .commit();
+
+    let error = builder.get_error().expect("should have error");
+    assert!(
+        matches!(
+            error,
+            casper_execution_engine::core::engine_state::Error::Exec(
+                casper_execution_engine::core::execution::Error::DisabledContract(disabled_contract_hash)
```
rafal-ch: The test looks ok, but I had some trouble understanding why we expect this contract to be disabled. This is (probably) because earlier we executed the other contract that called `storage::disable_contract_version()` on `contract_hash_v1`.

Can you maybe consider adding more comments that explain the test flow for people not that well acquainted with the execution engine?

execution_engine/src/core/runtime/mod.rs
mpapierski: I don't think it's necessary. If calling a locked contract with a version larger than 1, we'll never get to `execute_contract` as the validation happens very early here https://github.com/casper-network/casper-node/blob/1d3b388d5b62b8889c6bd01f0798b3cb20455bf1/execution_engine/src/core/engine_state/executable_deploy_item.rs#L863-L867

execution_engine/src/core/runtime/mod.rs
mpapierski: I moved this code from `call_versioned_contract,` so the validation didn't happen here in `execute_contract,` but once we got here, it was already valid. This PR just moves it on different level.

execution_engine/src/core/runtime/mod.rs
mpapierski: I tried different approaches, but I'm struggling to see an easy way to refactor this to avoid duplication nicely. The only thing that came to my mind was to use `self.context.read_gs_typed` rather than much longer `read_gs.` Hope that helps

execution_engine_testing/tests/src/test/regression/gh_3097.rs
mpapierski: Fair point. I renamed `contract_hash_v1`  and _v2 into `enabled_contract_hash` and `disabled_contract_hash` to make this a bit more readable. I hope that helps with readability

execution_engine/src/core/execution/error.rs
mpapierski: I do believe it helps. When you have a smart contract that does series of calls i.e.

```rust
runtime::call_contract(foo_contract_hash, ...);
runtime::call_contract(bar_contract_hash, ...);
runtime::call_contract(baz_contract_hash, ...);
```

It'll be helpful for the developer to see that call to `bar_contract_hash` failed. Then based on that a developer can query `bar_contract_hash` off-chain, confirm it's disabled, and then find a contract hash that's enabled.

execution_engine/src/core/runtime/mod.rs
rafal-ch: Cool, looks much better with `read_gs_typed` :+1:

rafal-ch (approved):
No blockers from my side.
LGTM :+1:

types/src/contracts.rs
```diff
@@ -808,6 +816,16 @@ impl ContractPackage {
...
     }
 
+    fn find_contract_version_key_by_hash(
+        &self,
+        contract_hash: &ContractHash,
+    ) -> Option<&ContractVersionKey> {
+        self.versions
+            .iter()
+            .filter_map(|(k, v)| if v == contract_hash { Some(k) } else { None })
+            .next()
```
Fraser999, line 826:
```suggestion
            .find_map(|(k, v)| if v == contract_hash { Some(k) } else { None })
```

Fraser999 (approved):

darthsiroftardis (approved):

types/src/contracts.rs
mpapierski: Nice one, applied

PR #3147, opened 2022-06-29
Add option to disable HTTP servers

### LINKED ISSUE #3141: Add enable_server config option for each of the 4 http servers
Currently we have three HTTP server sections in the config.toml.  We should split the speculative execution server's config options into a new section entirely titled `speculative_exec_server`.

These four sections should all have a new boolean option `enable_server` at the top of each section.  All should be set to `true` except for `speculative_exec_server` which should be `false`.

The full list of the new section's options should be
```toml
enable_server = false
address = '0.0.0.0:7778'
qps_limit = 1
max_body_bytes = 2_621_440
```

For each server, if the `enable_server` option is `false`, then it should not be run: neither in the joiner reactor nor in the participating one.

georgepisaltu:
Fixes #3141 

This PR adds functionality to disable the HTTP servers in node (RPC, REST, events, speculative execution). Additionally, it separates the configurations of the RPC and speculative execution servers, which were previously under the same section.

This functionality can't be tested in node, so the tests are in `casper-test`. After this PR gets merged, https://github.com/casper-network/casper-test/pull/88 should also get merged.

### DIFF
--- a/node/src/components/rest_server.rs
+++ b/node/src/components/rest_server.rs
@@ -28,7 +28,7 @@ use std::{convert::Infallible, fmt::Debug, time::Instant};
 use datasize::DataSize;
 use futures::{future::BoxFuture, join, FutureExt};
 use tokio::{sync::oneshot, task::JoinHandle};
-use tracing::{debug, error, warn};
+use tracing::{debug, error, info, warn};
 
 use casper_types::ProtocolVersion;
 
@@ -78,6 +78,8 @@ impl<REv> ReactorEventT for REv where
 
 #[derive(DataSize, Debug)]
 pub(crate) struct RestServer {
+    /// Enable.
+    enable: bool,
     /// When the message is sent, it signals the server loop to exit cleanly.
     #[data_size(skip)]
     shutdown_sender: oneshot::Sender<()>,
@@ -103,18 +105,23 @@ impl RestServer {
     {
         let (shutdown_sender, shutdown_receiver) = oneshot::channel::<()>();
 
-        let builder = utils::start_listening(&config.address)?;
-        let server_join_handle = tokio::spawn(http_server::run(
-            builder,
-            effect_builder,
-            api_version,
-            shutdown_receiver,
-            config.qps_limit,
-        ));
+        let server_join_handle = if config.enable_server {
+            let builder = utils::start_listening(&config.address)?;
+            Some(tokio::spawn(http_server::run(
+                builder,
+                effect_builder,
+                api_version,
+                shutdown_receiver,
+                config.qps_limit,
+            )))
+        } else {
+            None
+        };
 
         Ok(RestServer {
+            enable: config.enable_server,
             shutdown_sender,
-            server_join_handle: Some(server_join_handle),
+            server_join_handle,
             node_startup_instant,
             node_state,
         })
@@ -134,6 +141,9 @@ where
         _rng: &mut NodeRng,
         event: Self::Event,
     ) -> Effects<Self::Event> {
+        if !self.enable {
+            return Effects::new();
+        }
         match event {
             Event::RestRequest(RestRequest::Status { responder }) => {
                 let node_uptime = self.node_startup_instant.elapsed();
@@ -178,16 +188,20 @@ where
 impl Finalize for RestServer {
     fn finalize(mut self) -> BoxFuture<'static, ()> {
         async {
-            let _ = self.shutdown_sender.send(());
+            if self.enable {
+                let _ = self.shutdown_sender.send(());
 
-            // Wait for the server to exit cleanly.
-            if let Some(join_handle) = self.server_join_handle.take() {
-                match join_handle.await {
-                    Ok(_) => debug!("rest server exited cleanly"),
-                    Err(error) => error!(%error, "could not join rest server task cleanly"),
+                // Wait for the server to exit cleanly.
+                if let Some(join_handle) = self.server_join_handle.take() {
+                    match join_handle.await {
+                        Ok(_) => debug!("rest server exited cleanly"),
+                        Err(error) => error!(%error, "could not join rest server task cleanly"),
+                    }
+                } else {
+                    warn!("rest server shutdown while already shut down")
                 }
             } else {
-                warn!("rest server shutdown while already shut down")
+                info!("rest server was disabled in config, no shutdown performed")
             }
         }
         .boxed()
--- a/node/src/components/rest_server.rs
+++ b/node/src/components/rest_server.rs
@@ -78,7 +78,7 @@ impl<REv> ReactorEventT for REv where
 
 #[derive(DataSize, Debug)]
 pub(crate) struct RestServer {
-    /// Enable.
+    /// Indicates whether the REST server is enabled or not.
     enable: bool,
     /// When the message is sent, it signals the server loop to exit cleanly.
     #[data_size(skip)]
--- a/node/src/components/rest_server.rs
+++ b/node/src/components/rest_server.rs
@@ -77,9 +77,7 @@ impl<REv> ReactorEventT for REv where
 }
 
 #[derive(DataSize, Debug)]
-pub(crate) struct RestServer {
-    /// Indicates whether the REST server is enabled or not.
-    enable: bool,
+pub(crate) struct InnerRestServer {
     /// When the message is sent, it signals the server loop to exit cleanly.
     #[data_size(skip)]
     shutdown_sender: oneshot::Sender<()>,
@@ -90,6 +88,12 @@ pub(crate) struct RestServer {
     node_startup_instant: Instant,
     /// The current state of operation.
     node_state: NodeState,
+}
+
+#[derive(DataSize, Debug)]
+pub(crate) struct RestServer {
+    /// Inner server is present only when enabled in the config.
+    inner_rest: Option<InnerRestServer>,
 }
 
 impl RestServer {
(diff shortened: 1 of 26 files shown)


resources/local/config.toml
```diff
@@ -236,20 +239,39 @@ qps_limit = 100
 # Maximum number of bytes to accept in a single request body.
 max_body_bytes = 2_621_440
 
-# Address to which the speculative execution endpoint binds to.
-speculative_execution_address = "0.0.0.0:7778"
+
+# ===================================================================
+# Configuration options for the speculative execution RPC HTTP server
```
goral09: It's JSON RPC server. I don't really understand whether we use these terms correctly but we seem to use that to differentiate.

goral09: Here and everywhere else.

node/src/components/rpc_server.rs
```diff
@@ -219,14 +230,16 @@ where
         event: Self::Event,
     ) -> Effects<Self::Event> {
         match event {
```
goral09: Couldn't we put that `self.enable` here? before we even match on the event?

goral09: Ah, b/c we have requests/events for different server types at the same "level" of types. We would have to differentiate between them somehow. We already kind of do - `Event::RpcRequest(RpcRequest)`. If we had different wrapper types/enum variants for other server types we wouldn't have to write all of these `if self.enable` for every individual variant.

node/src/components/rpc_server.rs
```diff
@@ -352,56 +369,79 @@ where
...
+            | Event::GetBlockTransfersResult { .. }
+            | Event::GetDeployResult { .. }
+            | Event::GetPeersResult { .. }
+            | Event::QueryEraValidatorsResult { .. }
+            | Event::QueryGlobalStateResult { .. }
+                if !self.enable =>
+            {
+                Effects::new()
+            }
+            _ => unreachable!(),
```
goral09: This will panic, correct? We should avoid panicking, even if we think this can't happen in bug-free software. I would change this to a `warn!` and log that we didn't expect to receive the request b/c endpoints should have been disabled.

node/src/components/rpc_server.rs
```diff
@@ -352,56 +369,79 @@ where
...
             Event::GetPeersResult {
                 peers,
                 main_responder,
-            } => main_responder.respond(peers).ignore(),
+            } if self.enable => main_responder.respond(peers).ignore(),
+            Event::RpcRequest(RpcRequest::SpeculativeDeployExecute { .. })
+                if self.speculative_exec_enable =>
+            {
+                Effects::new()
+            }
```
goral09: What's happening here? Why do we ignore the request even if the endpoint is enabled.

georgepisaltu: Forgot the `!` on the condition, the enabled case is handled above.

node/src/components/rpc_server.rs
georgepisaltu: The thing is that while the JSON-RPC and speculative execution servers are both "handled" here, they can be enabled/disabled independently from one another. As long as we have the same `enum` for their respective events, when we `match` to handle one type, we also need to consider the other's events and handle the error case. So right now there is no clean solution.

I chose to go with the condition after each `match` arm because of personal preference, but we could do any of these:
- handle the special case (speculative execution request) before the `match` with an `if let`, then have the big `match` inside an `if self.enabled`
- have another match for `Event::RpcRequest` variants where we handle the speculative execution one separately, then we proceed with the rest of the RPC requests, should yield less conditional code
- separate the speculative execution request into its own thing, looks similar to the first solution

WDYT

node/src/components/rpc_server.rs
georgepisaltu: I'd even go with an `error`, it would mean we have an unhandled event type.

node/src/components/rpc_server.rs
goral09: I think I would go for the second option.

resources/local/config.toml
georgepisaltu: Done.

node/src/components/rpc_server.rs
georgepisaltu: Fixed.

node/src/components/rpc_server.rs
georgepisaltu: Done.

goral09 (approved):
No blockers.

node/src/components/rest_server.rs
```diff
@@ -78,6 +78,8 @@ impl<REv> ReactorEventT for REv where
 
 #[derive(DataSize, Debug)]
 pub(crate) struct RestServer {
+    /// Enable.
```
rafal-ch: Maybe: `Indicates whether the server is enabled or not`.
It may seem superfluous, but I like doc comments to be descriptive.

Up to you, though.

rafal-ch (approved):
LGTM with minor nit :+1:

node/src/components/rest_server.rs
georgepisaltu: Good catch, I actually meant to update those but forgot about them.

georgepisaltu: Done.

Fraser999 (changes requested):
We'll need to update the several other config.toml files.  There are quite a few - if you search e.g. for the term `[rpc_server]` you'll get them all.

We also need to update the node's changelog.  The last two entries under "Added" need to be updated, and we need a new entry (also under "Added" I guess) detailing that the three existing servers have the new option `enable_server`.

Other than that, LGTM.

marc-casperlabs (dismissed):
I would prefer to have a unified approach for this. I will write up my thoughts, please give me a minute.

I am blocking this **temporarily** in the meantime.

georgepisaltu:
> We'll need to update the several other config.toml files. There are quite a few - if you search e.g. for the term `[rpc_server]` you'll get them all.
> 
> We also need to update the node's changelog. The last two entries under "Added" need to be updated, and we need a new entry (also under "Added" I guess) detailing that the three existing servers have the new option `enable_server`.
> 
> Other than that, LGTM.

Updated `CHANGELOG.md` and the remaining configs.

marc-casperlabs (commented):
The way `EventStreamServer` is setup is, in my opinion, the "correct" way this is structured. The component is essentially a state machine and here we use `Option` to indicate whether it is in the enabled or disabled state, and most importantly only include the necessary state to run if the feature has been enabled.

We could consider making this explicit by adding a type 

```rust
enum Feature<T> {
    Enabled<T>,
    Disabled
}
```

which is just `Option` with a new name to make it more clear, but that's a matter of taste.

What I would personally prefer is seeing that applied to the RPC server as well. The `enabled` flags makes for really hard to read code there, especially in the big `match` blocks. To me, there's two scenarios: Either we allow the speculative execution feature to be used in isolation (with the rest of the RPC disabled) or not.

The latter would probably result in a struct similar to this:

```rust
#[derive(DataSize, Debug)]
pub(crate) struct InnerRpcServer {
    /// Indicates whether the speculative execution JSON-RPC server is enabled or not.
    speculative_exec_enable: bool,
    /// The instant at which the node has started.
    node_startup_instant: Instant,
    /// The current state of the node.
    node_state: NodeState,
}

#[derive(DataSize, Debug)]
pub(crate) struct RpcServer {
    server: Feature<InnerRpcServer>,
}
```

Dispatching would work as it does for the event stream (checking if `server` is some) and processing accordingly (no need for `self.enabled` checks anymore). Only a single match arm would check for `speculative_exec_enable`, and might return an HTTP unauthorized response.

Note that this makes it impossible to use the speculative execution feature without having RPC enabled - I think this is somewhat reasonable though.

Alternatively, a construction like this

```rust
#[derive(DataSize, Debug)]
pub(crate) struct InnerRpcServer {
    /// Indicates whether the speculative execution JSON-RPC server is enabled or not.
    speculative_exec_enable: bool,
    /// The instant at which the node has started.
    node_startup_instant: Instant,
    /// The current state of the node.
    node_state: NodeState,
}


#[derive(DataSize, Debug)]
pub(crate) struct RpcServer {
    rpc_enabled: bool,
    spec_exec: bool,
    rpc_server: Feature<InnerRpcServer>,  // or just InnerRpcServer.
}

```

When constructing, `rpc_server` would be instantiated if either `rpc_enabled` or `spec_exec` are true. Alternatively split them into two features:

```rust
#[derive(DataSize, Debug)]
pub(crate) struct RpcServer {
    rpc: Feature<InnerRpcServer>,
    spec_exec: Feature<()>,   // or bool, it's a matter of taste :)
}
```

As for the routing itself (and getting rid of the boolean checks), how about something like this instead (this is for variant one):

```rust
 fn handle_event(
        &mut self,
        _rng: &mut NodeRng,
        event: Self::Event,
    ) -> Effects<Self::Event> {

    if matches!(event, Event::RpcRequest(RpcRequest::SpeculativeDeployExecute {
                block_header,
                deploy,
                responder,
            }) {
        if self.speculative_exec_enable {
            return self.handle_execute_deploy(effect_builder, block_header, *deploy, responder)
        });
    }

    if !self.enabled {
        return Effects::new();
    }


    match event {
        Event::RpcRequest(RpcRequest::SpeculativeDeployExecute {.. }) {
            // log an error, this should never be reached.
            Effects::new()
        }
        // Regular match.
    }    
    }
}
```

For the REST server, I think it can be built similarly.

Anyway, I will unblock the PR. I don't think it is a tremendous issue to keep it the way it is right, but it would be nice to clean the state up a little and not have all these `if self.enabled` guards in the `match` blocks.

marc-casperlabs (commented):
Unblocked!

Fraser999 (approved):

Fraser999:
While I don't disagree with @marc-casperlabs points above, I think we're ok to merge this as is since some of his concerns will be addressed when we clean up the RPC server component in upcoming work.

The rough plan is to get rid of the component entirely, since it's probably surplus to requirements.  The server running in the tokio task will be able to interact directly with the other components as required via holding a copy of the effect builder for its lifetime.

Even if we don't go as extreme as removing the component, we can still probably reduce the event variants to just one or two if we change the server task to not route requests via its own component.

goral09:
> As for the routing itself (and getting rid of the boolean checks), how about something like this instead (this is for variant one):
> [...]
>  I don't think it is a tremendous issue to keep it the way it is right, but it would be nice to clean the state up a little and not have all these if self.enabled guards in the match blocks.

That's what I meant here - https://github.com/casper-network/casper-node/pull/3147#discussion_r910731740

georgepisaltu:
I refactored the logic of the servers in https://github.com/casper-network/casper-node/pull/3147/commits/fe2da27e0e87c7ce5cc3b8b420f2e15b58a128c1.

I ended up ditching the `Feature` suggestion because it felt like a worse `Option` (fewer methods, functionality, etc.) and I found no good place to put it, since it had to be reused in 3 different servers.

@marc-casperlabs @Fraser999 @goral09 let me know what you think. If this is not an improvement, we can just revert this commit and merge without it, since the previous work had all approvals.

georgepisaltu:
bors try

Fraser999 (approved):

Fraser999:
bors merge

marc-casperlabs (commented):
LGTM!

PR #3155, opened 2022-06-30
Restrict JSON-RPC request `params` to JSON Array or Object type

### LINKED ISSUE #3149: Fix handling of params field in JSON-RPC server
The params field should only be deemed valid if it is an Object or Array.

 think I found what breaks the SDK in 1.5 in some RPC methods. If I call
casper-client get-state-root-hash --node-address http://localhost:11101 -vv
{
  "jsonrpc": "2.0",
  "method": "chain_get_state_root_hash",
  "params": null,
  "id": 8065587110217483070
}
you can see that the params value is null. And in @open-rpc/client-js it seems that they don’t tolerate params to be null which might be right according to this spec which says:
If present, parameters for the rpc call MUST be provided as a Structured value. Either by-position through an Array or by-name through an Object.
If I miss params argument in the call, @open-rpc/client-js automatically replaces it with empty array (params: []) which gives me following error from the RPC:
Failed to parse ‘params’ field: invalid length 0, expected struct GetStateRootHashParams with 1 element
This is something that started to happen in 1.5, as when I try doing the same call in 1.4.6 it just works.
I updated the @open-rpc/client-js to the newest version, but it doesn’t help. I can raise an issue in their repo, but we probably should take into consideration that this change can affect other SDKs as well.

Fraser999:
This PR corrects the JSON-RPC handling of the `params` field of requests to be compliant with the standard.

Specifically, the server will now reject a request where the `params` field is set to `null`.

Closes #3149.

### DIFF
--- a/json_rpc/src/filters/tests/main_filter_with_recovery_tests.rs
+++ b/json_rpc/src/filters/tests/main_filter_with_recovery_tests.rs
@@ -11,7 +11,7 @@ use warp::{filters::BoxedFilter, Filter, Reply};
 use super::ResponseBodyOnRejection;
 use crate::{
     filters::{handle_rejection, main_filter},
-    Error, RequestHandlersBuilder, ReservedErrorCode, Response,
+    Error, Params, RequestHandlersBuilder, ReservedErrorCode, Response,
 };
 
 const GET_GOOD_THING: &str = "get good thing";
@@ -31,14 +31,16 @@ impl Serialize for BadThing {
     }
 }
 
-async fn get_good_thing(params: Option<Value>) -> Result<GoodThing, Error> {
+async fn get_good_thing(params: Option<Params>) -> Result<GoodThing, Error> {
     match params {
-        Some(Value::String(good_thing)) => Ok(GoodThing { good_thing }),
+        Some(Params::Array(array)) => Ok(GoodThing {
+            good_thing: array[0].as_str().unwrap().to_string(),
+        }),
         _ => Err(Error::new(ReservedErrorCode::InvalidParams, "no params")),
     }
 }
 
-async fn get_bad_thing(_params: Option<Value>) -> Result<BadThing, Error> {
+async fn get_bad_thing(_params: Option<Params>) -> Result<BadThing, Error> {
     Ok(BadThing)
 }
 
@@ -65,7 +67,7 @@ async fn should_handle_valid_request() {
     // This should get handled by `fn get_good_thing` and return `Ok` as "params" is Some, causing a
     // Response::Success to be returned to the client.
     let http_response = warp::test::request()
-        .body(r#"{"jsonrpc":"2.0","id":"a","method":"get good thing","params":"one"}"#)
+        .body(r#"{"jsonrpc":"2.0","id":"a","method":"get good thing","params":["one"]}"#)
         .filter(&filter)
         .await
         .unwrap()
@@ -91,7 +93,7 @@ async fn should_handle_valid_request_where_rpc_returns_error() {
     // This should get handled by `fn get_good_thing` and return `Err` as "params" is None, causing
     // a Response::Failure (invalid params) to be returned to the client.
     let http_response = warp::test::request()
-        .body(r#"{"jsonrpc":"2.0","id":"a","method":"get good thing","params":null}"#)
+        .body(r#"{"jsonrpc":"2.0","id":"a","method":"get good thing"}"#)
         .filter(&filter)
         .await
         .unwrap()
@@ -142,7 +144,7 @@ async fn should_handle_request_for_method_not_registered() {
     // This should get handled by `filters::handle_body` and return Response::Failure (invalid
     // request) to the client as the ID has fractional parts.
     let http_response = warp::test::request()
-        .body(r#"{"jsonrpc":"2.0","id":1,"method":"not registered","params":"one"}"#)
+        .body(r#"{"jsonrpc":"2.0","id":1,"method":"not registered","params":["one"]}"#)
         .filter(&filter)
         .await
         .unwrap()
@@ -169,7 +171,7 @@ async fn should_handle_request_with_invalid_id() {
     // This should get handled by `filters::handle_body` and return Response::Failure (invalid
     // request) to the client as the ID has fractional parts.
     let http_response = warp::test::request()
-        .body(r#"{"jsonrpc":"2.0","id":1.1,"method":"get good thing","params":"one"}"#)
+        .body(r#"{"jsonrpc":"2.0","id":1.1,"method":"get good thing","params":["one"]}"#)
         .filter(&filter)
         .await
         .unwrap()
@@ -196,7 +198,7 @@ async fn should_handle_request_with_no_id() {
     // This should get handled by `filters::handle_body` and return no JSON-RPC response, only an
     // HTTP response (bad request) to the client as no ID was provided.
     let http_response = warp::test::request()
-        .body(r#"{"jsonrpc":"2.0","method":"get good thing","params":"one"}"#)
+        .body(r#"{"jsonrpc":"2.0","method":"get good thing","params":["one"]}"#)
         .filter(&filter)
         .await
         .unwrap()
@@ -219,7 +221,7 @@ async fn should_handle_request_with_extra_field() {
     // This should get handled by `filters::handle_body` and return Response::Failure (invalid
     // request) to the client as the ID has fractional parts.
     let http_response = warp::test::request()
-        .body(r#"{"jsonrpc":"2.0","id":1,"method":"get good thing","params":2,"extra":"field"}"#)
+        .body(r#"{"jsonrpc":"2.0","id":1,"method":"get good thing","params":[2],"extra":"field"}"#)
         .filter(&filter)
         .await
         .unwrap()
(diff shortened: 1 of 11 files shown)


json_rpc/src/request.rs
```diff
@@ -61,20 +55,76 @@ fn is_valid(id: &Value) -> Result<(), Error> {
...
+        // Just copy "id" field for now to return verbatim in any errors before we get to actually
+        // validating the "id" field itself.
+        let id = request.get(ID_FIELD_NAME).cloned().unwrap_or_default();
+
+        match request.remove(JSONRPC_FIELD_NAME) {
+            Some(Value::String(jsonrpc)) => {
+                if jsonrpc != JSON_RPC_VERSION {
+                    let error = Error::new(
+                        ReservedErrorCode::InvalidRequest,
+                        format!("'jsonrpc' must be '2.0', but was '{}'", jsonrpc),
```
georgepisaltu: I'd phrase as
```suggestion
                        format!("expected 'jsonrpc' to be '2.0', got '{}'", jsonrpc),
```
Here and below. Not feeling strongly about this, up to you.

node/src/components/rpc_server/rpcs.rs
```diff
@@ -403,23 +413,29 @@ mod tests {
...
             let filter = main_filter_with_recovery();
 
-            let rpc_response = send_request(GetPeers::METHOD, Some("3"), &filter).await;
+            let rpc_response = send_request(GetPeers::METHOD, Some("[3]"), &filter).await;
             assert_eq!(
                 rpc_response.error().unwrap(),
                 &Error::new(
                     ReservedErrorCode::InvalidParams,
-                    "'params' field should be null or absent"
+                    "'params' field should be an empty Array '[]', an empty Object '{}' or absent"
```
goral09, line 438: You sent `[3]` which is an array though, right?

json_rpc/src/request/params.rs
```diff
@@ -0,0 +1,245 @@
...
+
+impl Params {
+    pub(super) fn try_from(request_id: &Value, params: Value) -> Result<Self, ErrorOrRejection> {
+        let err_invalid_request = |additional_info: &str| {
+            let error = Error::new(ReservedErrorCode::InvalidRequest, additional_info);
+            Err(ErrorOrRejection::Error {
+                id: request_id.clone(),
+                error,
+            })
+        };
```
georgepisaltu, line 31: nit: I'd have this as a private function outside `impl Params`.

json_rpc/src/filters/tests/main_filter_with_recovery_tests.rs
```diff
@@ -284,10 +286,7 @@ async fn should_handle_malformed_request_but_valid_json() {
     assert_eq!(rpc_response.id(), &Value::Null);
     assert_eq!(
         rpc_response.error().unwrap(),
-        &Error::new(
-            ReservedErrorCode::InvalidRequest,
-            "unknown field `a`, expected one of `jsonrpc`, `id`, `method`, `params` at line 1 column 4"
-        )
+        &Error::new(ReservedErrorCode::InvalidRequest, "Missing 'jsonrpc' field")
```
goral09, line 289: This also changed. If I was very strict I would say we're changing the API since errors returned are now different.

json_rpc/src/filters/tests/main_filter_with_recovery_tests.rs
```diff
@@ -219,7 +221,7 @@ async fn should_handle_request_with_extra_field() {
     // This should get handled by `filters::handle_body` and return Response::Failure (invalid
     // request) to the client as the ID has fractional parts.
     let http_response = warp::test::request()
-        .body(r#"{"jsonrpc":"2.0","id":1,"method":"get good thing","params":2,"extra":"field"}"#)
+        .body(r#"{"jsonrpc":"2.0","id":1,"method":"get good thing","params":[2],"extra":"field"}"#)
```
goral09, line 224: The comment says that id has fractional parts but it's not true. This one is incorrect as it has an extra field?

json_rpc/src/request/params.rs
```diff
@@ -0,0 +1,245 @@
...
+        match Params::try_from(&original_id, Value::Bool(false)).unwrap_err() {
+            ErrorOrRejection::Error { id, error } => {
+                assert_eq!(id, original_id);
+                assert_eq!(
+                    serde_json::to_string(&error).unwrap(),
+                    expected_error("'false'")
+                );
+            }
+            other => panic!("unexpected: {:?}", other),
+        }
```
georgepisaltu: Can we move this to a test function and use it in all of the tests here? We can pass all the arguments required in each test. As the tests would be one liners, we can have each of them in a single test, `bad_value_to_params_conversion` or something like that.

json_rpc/src/request.rs
```diff
@@ -61,20 +55,76 @@ fn is_valid(id: &Value) -> Result<(), Error> {
...
+                    );
+                    return Err(ErrorOrRejection::Error { id, error });
+                }
+            }
+            Some(jsonrpc) => {
+                let error = Error::new(
+                    ReservedErrorCode::InvalidRequest,
+                    format!("'jsonrpc' must be '2.0', but was '{}'", jsonrpc),
```
rafal-ch: Is the error message ok here? We don't seem to compare it to `JSON_RPC_VERSION` in this match arm

georgepisaltu (commented):
Looks good, some nits

json_rpc/src/request.rs
```diff
@@ -85,19 +135,19 @@ impl TryFrom<UnvalidatedRequest> for Request {
...
         };
 
-        if request.jsonrpc != JSON_RPC_VERSION {
+        if !request.is_empty() {
             let error = Error::new(
                 ReservedErrorCode::InvalidRequest,
-                format!("'jsonrpc' must be '2.0', but was '{}'", request.jsonrpc),
+                format!(
+                    "Unexpected field{}: {}",
+                    if request.len() > 1 { "s" } else { "" },
```
rafal-ch, line 143: Great attention to details! :)

node/src/components/rpc_server/rpcs.rs
```diff
@@ -403,23 +413,29 @@ mod tests {
...
+            let rpc_response = send_request(GetPeers::METHOD, Some("{}"), &filter).await;
             assert_eq!(
                 rpc_response.result().as_ref(),
                 Some(GetPeersResult::doc_example())
             );
         }
 
         #[tokio::test]
-        async fn should_return_error_if_params_not_null() {
+        async fn should_return_error_if_params_not_empty() {
```
rafal-ch, line 430: Not related to this PR, but please consider changing the tests to use `now_or_never()` instead of `await`. It'll allow to run these tests w/o tokio runtime and gain a slight performance boost.

goral09 (approved):

node/src/components/rpc_server/rpcs.rs
rafal-ch: I think it's correct, because we use `info_get_peers` which is an `RpcWithoutParams` and we fail on the `check_no_params()` validation.

goral09: Ah, I see 👍

json_rpc/src/request.rs
goral09: I stopped on this for a moment as well - the reason for this is that we handle the case when `jsonrpc` field is a string in the match above, so if we get to this one then it must be that `jsonrpc` value is of incorrect type.

json_rpc/src/filters/tests/main_filter_with_recovery_tests.rs
Fraser999: Yup - good catch - thanks!

json_rpc/src/filters/tests/main_filter_with_recovery_tests.rs
Fraser999: Yeah, but that's already the case compared to the 1.4.6 version.  The errors are now (hopefully) more useful, as well as being compliant to the standard.

json_rpc/src/request/params.rs
Fraser999: I think this is fine - it's a tiny function inside another small function, and it avoids having to pass in the `id` at each callsite.  Not a big deal either way, but I'd rather leave as is.

json_rpc/src/filters/tests/main_filter_with_recovery_tests.rs
Fraser999: Fixed in [a11b164](https://github.com/casper-network/casper-node/pull/3155/commits/a11b164bdc53b713db4f181e77bdcef32c33fae2).

json_rpc/src/request.rs
Fraser999: Fixed in [a11b164](https://github.com/casper-network/casper-node/pull/3155/commits/a11b164bdc53b713db4f181e77bdcef32c33fae2).

json_rpc/src/request/params.rs
Fraser999: Fixed in [a11b164](https://github.com/casper-network/casper-node/pull/3155/commits/a11b164bdc53b713db4f181e77bdcef32c33fae2).

rafal-ch (approved):
LGTM, no blockers :+1:

Fraser999:
bors merge

georgepisaltu (approved):

Fraser999:
bors merge

PR #3160, opened 2022-07-01
Fix for new clippy warnings

Fraser999:
This fixes clippy warnings introduced by Rust 1.62.0.  It also adds missing features to the `casper-types` entry of the `casper-hashing` manifest.

### DIFF
--- a/hashing/Cargo.toml
+++ b/hashing/Cargo.toml
@@ -12,7 +12,7 @@ license = "Apache-2.0"
 [dependencies]
 blake2 = "0.9.0"
 base16 = "0.2.1"
-casper-types = { version = "1.5.0", path = "../types" }
+casper-types = { version = "1.5.0", path = "../types", features = ["datasize", "std"] }
 datasize = "0.2.9"
 hex = { version = "0.4.2", default-features = false, features = ["serde"] }
 hex-buffer-serde = "0.3.0"
--- a/node/src/testing/network.rs
+++ b/node/src/testing/network.rs
@@ -74,9 +74,9 @@ where
     ///
     /// Panics if a duplicate node ID is being inserted. This should only happen in case a randomly
     /// generated ID collides.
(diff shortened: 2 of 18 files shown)


hashing/Cargo.toml
```diff
@@ -12,7 +12,7 @@ license = "Apache-2.0"
 [dependencies]
 blake2 = "0.9.0"
 base16 = "0.2.1"
-casper-types = { version = "1.5.0", path = "../types" }
+casper-types = { version = "1.5.0", path = "../types", features = ["datasize", "std"] }
```
goral09, line 15: What changed that now we need to add these features just now? Is it something in our software or in the rustc?

node/src/testing/network.rs
```diff
@@ -74,9 +74,9 @@ where
     ///
     /// Panics if a duplicate node ID is being inserted. This should only happen in case a randomly
     /// generated ID collides.
-    pub(crate) async fn add_node(
-        &mut self,
-        rng: &mut TestRng,
+    pub(crate) async fn add_node<'a, 'b: 'a>(
```
goral09, line 77: Does it mean that `'b` outlives `'a`?

goral09 (approved):
How can I argue with clippy and rustc 🤷

hashing/Cargo.toml
Fraser999: We probably just never test `casper-hashing` in isolation.  When we run `cargo t` from the workspace root, `casper-hashing` will get compiled using a build of `casper-types` which has the required features enabled due to other crates (like `casper-node`) demanding these features to be enabled.

node/src/testing/network.rs
Fraser999: Yes

Fraser999:
bors merge

mpapierski (approved):

PR #3161, opened 2022-07-01
Add `dictionary_read` method to ext-FFI

### LINKED ISSUE #3145: Provide read_key API method to the Rust contract API
Currently, we only allow reads via the contract API for values under a URef, however, the FFI itself allows for reads under any Key as opposed to just Key::URef . We should provide an additional method (so as to not introduce a breaking change) that allows for reads under Key. Additionally, we should also change Key::Dictionary to be readable

darthsiroftardis:
CHANGELOG:

- Added `dictionary_read` method to the ext-FFI that allows to directly read a value under a `Key::Dictionary`.
- Added API methods to both the Rust and Assembly script API libraries to make use of the newly added ext-FFI method.
- Added a test to check the functioning of the newly added methods
- Added utility method to the Assembly script library to create ` Key::Dictionary` from a seed uref and an item key

Closes: #3145


execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -1187,7 +1183,14 @@ where
         }
 
         let dictionary_key = Key::dictionary(uref, dictionary_item_key_bytes);
+        self.dictionary_read(dictionary_key)
+    }
 
+    /// Gets a dictionary value from a dictionary `Key`.
+    pub(crate) fn dictionary_read(
```
mpapierski, line 1196: In `if let Some(stored_value) = maybe_stored_value {` in this function it should use `dictionary::handle_stored_value` like this:

```patch
    if let Some(stored_value) = maybe_stored_value {
-            let cl_value_indirect: CLValue =
-                stored_value.try_into().map_err(Error::TypeMismatch)?;
-            let dictionary_value: DictionaryValue =
-                cl_value_indirect.into_t().map_err(Error::from)?;
-            let cl_value = dictionary_value.into_cl_value();
-            Ok(Some(cl_value))
+    let stored_value = dictionary::handle_stored_value(dictionary_key, stored_value)?;
+    Ok(Some(stored_value));
    }
```

execution_engine/src/core/runtime/externals.rs
```diff
@@ -998,6 +998,19 @@ where
...
+            FunctionIndex::DictionaryReadFuncIndex => {
+                // args(0) = pointer to key in Wasm memory
+                // args(1) = size of key in Wasm memory
+                // args(2) = pointer to output size (output param)
+                let (key_ptr, key_size, output_size_ptr) = Args::parse(args)?;
+                self.charge_host_function_call(
+                    &host_function_costs.read_value,
+                    [key_ptr, key_size, output_size_ptr],
+                )?;
+                let ret = self.dictionary_read(key_ptr, key_size, output_size_ptr)?;
```
mpapierski, line 1010: Instead of accepting a `Key` instance here, we should accept a `DictionaryAddr` (which is a 32bytes array) and construct a `Key::Dictionary` inside of it to make it work only for dictionaries. Then, we can remove the validation later inside: `if !key.is_dictionary_key()`  and replace this if with `let key = Key::Dictionary(dictionary_addr)`. AFAIR other dictionary-related host function doesn't accept `Key`s for seed, just the URefs.

types/src/key.rs
```diff
@@ -511,6 +511,14 @@ impl Key {
         hasher.finalize_variable(|hash| addr.clone_from_slice(hash));
         Key::Dictionary(addr)
     }
+
+    /// Returns true if the key is of type [`Key::Dictionary`].
+    pub fn is_dictionary_key(&self) -> bool {
```
mpapierski, line 517: Shouldn't need this if we'd make new host function accept a DictionaryAddr rather than arbitrary Key.

mpapierski (approved):
Looking good so far. Optimistic approval with few nits

execution_engine/src/core/runtime/externals.rs
darthsiroftardis: So we want to let users who have the actual `Key::Dictionary` just use it directly, instead of having the key, and then de-structuring it to get the inner Addr. So while we have to do a check for it, it makes for easier use on the callers side

execution_engine/src/core/runtime_context/mod.rs
darthsiroftardis: Good point!

execution_engine/src/core/resolvers/v1_function_index.rs
```diff
@@ -58,6 +58,7 @@ pub(crate) enum FunctionIndex {
     LoadCallStack,
     LoadAuthorizationKeys,
     RandomBytes,
+    DictionaryReadFuncIndex,
```
EdHastingsCasperAssociation, line 61: @piotr-dziubecki , please make a 2.0 tech debt ticket to normalize the variant names of this enum

execution_engine/src/core/resolvers/v1_resolver.rs
```diff
@@ -229,6 +229,10 @@ impl ModuleImportResolver for RuntimeModuleImportResolver {
                 Signature::new(&[ValueType::I32; 1][..], Some(ValueType::I32)),
                 FunctionIndex::NewDictionaryFuncIndex.into(),
             ),
+            "casper_dictionary_read" => FuncInstance::alloc_host(
+                Signature::new(&[ValueType::I32; 3][..], Some(ValueType::I32)),
+                FunctionIndex::DictionaryReadFuncIndex.into(),
+            ),
```
EdHastingsCasperAssociation: move this up next to casper_dictionary_get

execution_engine/src/core/runtime/mod.rs
```diff
@@ -2819,6 +2819,41 @@ where
         Ok(Ok(()))
     }
 
+    /// Reads the `value` under a `key` in a dictionary.
+    fn dictionary_read(
```
EdHastingsCasperAssociation, line 2789: move this up next to dictionary_get

execution_engine/src/core/runtime/mod.rs
```diff
@@ -2819,6 +2819,41 @@ where
...
+        key_ptr: u32,
+        key_size: u32,
+        output_size_ptr: u32,
+    ) -> Result<Result<(), ApiError>, Trap> {
+        if !self.can_write_to_host_buffer() {
+            // Exit early if the host buffer is already occupied
+            return Ok(Err(ApiError::HostBufferFull));
+        }
+
+        let dictionary_address = self.key_from_mem(key_ptr, key_size)?;
```
EdHastingsCasperAssociation: isn't this really `dictionary_key` ?

execution_engine/src/core/runtime_context/mod.rs
```diff
@@ -1187,7 +1183,14 @@ where
         }
 
         let dictionary_key = Key::dictionary(uref, dictionary_item_key_bytes);
+        self.dictionary_read(dictionary_key)
```
EdHastingsCasperAssociation, line 1192: +1

smart_contracts/contract_as/assembly/dictionary.ts
```diff
@@ -73,3 +73,26 @@ export function dictionaryPut(uref: URef, key: String, value: CLValue): void {
         valueBytes.length
     );
 }
+
+/**
+ * Reads `value` under `dictionary-key` in a dictionary.
+ * @category Storage
+ */
+export function dictionaryRead(dictionaryKey: Key): Uint8Array | null {
```
EdHastingsCasperAssociation, line 63: move next to dictionaryGet above

modify the comment on line 39 above to be very clear that the key argument Get expects is dictionary_item_key
>* Reads the value under the dictionary item `key`.

smart_contracts/contract/src/ext_ffi.rs
```diff
@@ -767,6 +767,22 @@ extern "C" {
...
+    /// If the key is de-serialized successfully, then the result of the read is serialized
+    /// and buffered in the runtime. This result can be obtained via the [`casper_read_host_buffer`]
+    /// function. Returns standard error code.
+    ///
+    /// # Arguments
+    ///
+    /// * `key_ptr` - pointer (offset in wasm linear memory) to serialized form of the key to read
+    /// * `key_size` - size of the serialized key (in bytes)
+    /// * `output_size` - pointer to a value where host will write size of bytes read from given key
+    pub fn casper_dictionary_read(
```
EdHastingsCasperAssociation: put next to dictionary_get above

since both the dictionary_item_key expected by dictionary_get and the actual key expected by this ffi method both have the same arg names (key_ptr, key_size) you should make very clear in the comments on 778 and 779 that this expects a Key::Dictionary

execution_engine_testing/tests/src/test/contract_api/dictionary.rs
```diff
@@ -649,3 +650,19 @@ fn should_query_dictionary_items_with_test_builder() {
         assert_eq!(value, dictionary::DEFAULT_DICTIONARY_VALUE);
     }
 }
+
+#[ignore]
+#[test]
+fn should_be_able_to_perform_dictionary_read() {
```
EdHastingsCasperAssociation, line 657: there should also be a test for normal read, which should also work now that is_readable is true for Key::Dictionary

EdHastingsCasperAssociation (changes requested):
see comments

types/src/key.rs
EdHastingsCasperAssociation: iirc the only place we expose the implementation detail of the inner payloads of specific key variants in the ffi is `create_contract_package`, and it was a mistake to leak the abstraction there imo...we end up covering over it in the contract_api

execution_engine/src/core/resolvers/v1_resolver.rs
darthsiroftardis: Addressed in https://github.com/casper-network/casper-node/pull/3161/commits/862d2b060f8391957da23d76025c2b8a6fd92c5f

execution_engine/src/core/runtime/mod.rs
darthsiroftardis: Moved and addressed in https://github.com/casper-network/casper-node/pull/3161/commits/862d2b060f8391957da23d76025c2b8a6fd92c5f

execution_engine/src/core/runtime/mod.rs
darthsiroftardis: Renamed to `dictionary_key` https://github.com/casper-network/casper-node/pull/3161/commits/862d2b060f8391957da23d76025c2b8a6fd92c5f

smart_contracts/contract_as/assembly/dictionary.ts
darthsiroftardis: Moved, and the comment for `Get` has been updated in https://github.com/casper-network/casper-node/pull/3161/commits/862d2b060f8391957da23d76025c2b8a6fd92c5f

smart_contracts/contract/src/ext_ffi.rs
darthsiroftardis: Moved to `dictionary_get` and made the comments more clear https://github.com/casper-network/casper-node/pull/3161/commits/862d2b060f8391957da23d76025c2b8a6fd92c5f

execution_engine_testing/tests/src/test/contract_api/dictionary.rs
darthsiroftardis: Added a test for normal read in https://github.com/casper-network/casper-node/pull/3161/commits/862d2b060f8391957da23d76025c2b8a6fd92c5f

EdHastingsCasperAssociation (approved):

PR #3165, opened 2022-07-04
update to minimise API changes to casper-types

### LINKED ISSUE #3070: Revert breaking API changes in casper-types

Fraser999:
This PR mainly adds `#[non_exhaustive]` to error enums in the `casper-types` crate in order to avoid new variants in these causing breaking changes in the future.

Closes #3070

### DIFF
--- a/types/src/system.rs
+++ b/types/src/system.rs
@@ -7,7 +7,7 @@ pub mod mint;
 pub mod standard_payment;
 mod system_contract_type;
 
-pub use call_stack_element::CallStackElement;
+pub use call_stack_element::{CallStackElement, CallStackElementTag};
 pub use error::Error;
 pub use system_contract_type::{
     SystemContractType, AUCTION, HANDLE_PAYMENT, MINT, STANDARD_PAYMENT,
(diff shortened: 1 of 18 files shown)


types/src/system.rs
```diff
@@ -7,7 +7,7 @@ pub mod mint;
 pub mod standard_payment;
 mod system_contract_type;
 
-pub use call_stack_element::CallStackElement;
+pub use call_stack_element::{CallStackElement, CallStackElementTag};
```
georgepisaltu, line 10: Is this an accidental import? If not, what is its purpose?

georgepisaltu (commented):
Overall looks good! Just one question.

types/src/system.rs
Fraser999: It was (accidentally?) removed despite being part of the actual public API in 1.4.6 and in `dev`.  Not sure why this isn't picked up in lints though.

georgepisaltu (approved):

Fraser999:
bors merge

PR #3167, opened 2022-07-05
Fix update of joining set

### LINKED ISSUE #3172: Fix incorrect update of the "joining set"
We should not update the joining set when we receive an incoming connection, because the message_sender which is handling the corresponding outgoing connection will not receive the update of the remote joiner state.
Such desync may cause the node to try to send requests that are not safe for joiners, to the joining node, because the outgoing connection may outlive the incoming connection, i.e. it may take some time to drop "our" outgoing connection after a peer has closed the corresponding incoming connection.
We should also bump the log level if we detect that a peer is trying to query a joiner from warn to error because with the changes introduced in this ticket we expect it to never happen. So when it does, we want to alert loudly.

rafal-ch:
* the "joining peer set" is **not** updated upon establishing the incoming connection (such an update may cause the node to try to send requests that are not safe for joiners, to the joining node_)
* bumps the log message severity if we detect that a peer is trying to query a joiner from `warn` to `error`
* introduces the helper `HandshakeOutcome` struct

Closes https://github.com/casper-network/casper-node/issues/3172

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -518,8 +518,6 @@ where
                     // joiners, to the joining node, because the outgoing connection may outlive the
                     // incoming connection, i.e. it may take some time to drop "our" outgoing
                     // connection after a peer has closed the corresponding incoming connection.
-
-                    // self.update_joining_set(peer_id, is_joiner);
                 }
 
                 // Now we can start the message reader.
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -473,7 +473,6 @@ where
                 peer_id,
                 peer_consensus_public_key,
                 stream,
-                is_joiner,
             } => {
                 if self.cfg.max_incoming_peer_connections != 0 {
                     if let Some(symmetries) = self.connection_symmetries.get(&peer_id) {
@@ -510,7 +509,17 @@ where
                     .add_incoming(peer_addr, Instant::now())
                 {
                     self.connection_completed(peer_id);
-                    self.update_joining_set(peer_id, is_joiner);
+
+                    // We should not update the joining set when we receive an incoming connection,
+                    // because the `message_sender` which is handling the corresponding outgoing
+                    // connection will not receive the update of the remote joiner state.
+                    //
+                    // Such desync may cause the node to try to send requests that are not safe for
+                    // joiners, to the joining node, because the outgoing connection may outlive the
+                    // incoming connection, i.e. it may take some time to drop "our" outgoing
+                    // connection after a peer has closed the corresponding incoming connection.
+
+                    // self.update_joining_set(peer_id, is_joiner);
                 }
 
                 // Now we can start the message reader.
--- a/node/src/components/small_network/tasks.rs
+++ b/node/src/components/small_network/tasks.rs
@@ -63,10 +63,15 @@ use crate::{
 /// successfully handed over to the kernel for sending.
 pub(super) type MessageQueueItem<P> = (Arc<Message<P>>, Option<AutoClosingResponder<()>>);
 
+/// The outcome of the handshake process.
 struct HandshakeOutcome {
+    /// A framed transport for peer.
     framed_transport: FramedTransport,
+    /// Public address advertised by the peer.
     public_addr: SocketAddr,
+    /// The public key the peer is validating with, if any.
     peer_consensus_public_key: Option<PublicKey>,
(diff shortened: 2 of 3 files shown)


node/src/components/small_network.rs
```diff
@@ -510,7 +509,17 @@ where
...
+                    // We should not update the joining set when we receive an incoming connection,
+                    // because the `message_sender` which is handling the corresponding outgoing
+                    // connection will not receive the update of the remote joiner state.
+                    //
+                    // Such desync may cause the node to try to send requests that are not safe for
+                    // joiners, to the joining node, because the outgoing connection may outlive the
+                    // incoming connection, i.e. it may take some time to drop "our" outgoing
+                    // connection after a peer has closed the corresponding incoming connection.
+
+                    // self.update_joining_set(peer_id, is_joiner);
```
goral09: Is this left commented out on purpose?

node/src/components/small_network/tasks.rs
```diff
@@ -63,6 +63,13 @@ use crate::{
 /// successfully handed over to the kernel for sending.
 pub(super) type MessageQueueItem<P> = (Arc<Message<P>>, Option<AutoClosingResponder<()>>);
 
+struct HandshakeOutcome {
+    framed_transport: FramedTransport,
+    public_addr: SocketAddr,
+    peer_consensus_public_key: Option<PublicKey>,
+    is_joiner: bool,
```
goral09:
```suggestion
    is_peer_joiner: bool,
```

I prefer this over `is_joiner` as the latter can be confused with a flag indicating whether _we_ are a joiner node.

goral09 (approved):

node/src/components/small_network.rs
rafal-ch: Yes, to make the way we attempted to fix the bug explicit. I see no problem removing it if you'd like to.

node/src/components/small_network/tasks.rs
rafal-ch: I agree.
Changed back to `is_peer_joiner` in 96d9555915bfc004396edba41cf9bc4d11a98061

node/src/components/small_network.rs
marc-casperlabs: I think our policy is to remove commented out code, so let's take it out!

node/src/components/small_network/tasks.rs
```diff
@@ -63,6 +63,13 @@ use crate::{
 /// successfully handed over to the kernel for sending.
 pub(super) type MessageQueueItem<P> = (Arc<Message<P>>, Option<AutoClosingResponder<()>>);
 
+struct HandshakeOutcome {
```
marc-casperlabs, line 67: Could you add docs on the struct and its members? :sparkles: I try to be diligent about this, even (or especially :)) internal structs.

node/src/components/small_network/tasks.rs
```diff
@@ -134,11 +141,16 @@ where
...
 
     // Negotiate the handshake, concluding the incoming connection process.
-    match negotiate_handshake::<P, _>(&context, framed, connection_id).await {
-        Ok((framed, public_addr, peer_consensus_public_key, is_joiner)) => {
+    match negotiate_handshake::<P, _>(&context, framed_transport, connection_id).await {
+        Ok(HandshakeOutcome {
+            framed_transport,
+            public_addr,
+            peer_consensus_public_key,
+            is_peer_joiner: is_joiner,
```
marc-casperlabs, line 157: We could do without the rename here, for clarity, I think. I know it keeps the diff smaller, but I'd be too tempted to remove the alias later on :)

marc-casperlabs (approved):
Only nitpicks which would be nice to fix. Otherwise looks good to me!

node/src/components/small_network/tasks.rs
rafal-ch: I'll leave it as is, because this field will probably be removed in https://github.com/casper-network/casper-node/pull/3179

node/src/components/small_network.rs
rafal-ch: Removed in fd62aa2c41eed5489c3576b82438fe5d964603f2

node/src/components/small_network/tasks.rs
rafal-ch: Comments added here: a50a50659bb71cc39d0bf6b1a284c968948ca4f6

PR #3179, opened 2022-07-06
Add `SyncFinished` message to the network protocol

### LINKED ISSUE #3173: Extend the networking protocol with SyncStateChanged
The nodes that are participating may still be running the "sync to genesis" process in the background. Until the background syncing process is finished we shouldn't query such node for tries.
Proposed extension to the network protocol:
- rename the current `is_joiner` flag to `sync_state` flag (includes the related code, like `update_joining_set`, etc.)
- introduce an announcement for flipping the `sync_state` flag
  - extend small_network::Message by adding new `SyncStateChanged(bool)` variant
  - consider allowing the one side flipping (from true to false) as node should never switch from "participating" back to "joiner"
- decouple the handshake from setting the syncing state flag
  - remove the `is_joiner` from the `Message::Handshake` variant
  - on handshake default the flag to `true`
  - send the `SyncStateChanged` announcement with correct state afterwards
- the sync_state flag should be set to true only in these two conditions:
  - node is in the joiner reactor stage
  - node is in the participating reactor, but is doing the sync_to_genesis in the background
- ~upon receiving the SyncStateChanged(true), disconnect from all other nodes currently known as being in sync state (formerly, joiners)~ - there will be just a `SyncFinished` flag, meaning we'll never change the sync status back to true, hence no need for disconnection

- ~consider changes to the currently existing NodeState struct to better describe what mode the node is in at the given moment~ separate ticket will be requested because we need to discuss the actual states we now have in the node

rafal-ch:

- Introduces the `ChainSynchronizerAnnouncement` with `SyncFinished` variant. It is announced in two cases

  -  when node is not configured to `sync_to_genesis` it is sent immediately after the participating reactor spins up the chain synchronizer component
  - otherwise, it is sent after the `sync_to_genesis` is completed

- Makes the `small_network` react to the `SyncFinished` event by disconnecting all incoming connections and updating the internal `syncing` flag to `false`, so on reconnection, the `false` value will be used as an `is_syncing` parameter of the handshake

Closes #3173

### DIFF
--- a/node/src/components/chain_synchronizer.rs
+++ b/node/src/components/chain_synchronizer.rs
@@ -92,7 +92,8 @@ pub(crate) struct ChainSynchronizer<REv> {
     config: Config,
     /// This will be populated once the synchronizer has completed all work, indicating the joiner
     /// reactor can stop running.  It is passed to the participating reactor's constructor via its
-    /// config.
+    /// config. The participating reactor may still use the chain synchronizer component to run a
+    /// sync to genesis in the background.
     joining_outcome: Option<JoiningOutcome>,
     /// Metrics for the chain synchronization process.
     metrics: Metrics,
--- a/node/src/components/chain_synchronizer.rs
+++ b/node/src/components/chain_synchronizer.rs
@@ -126,7 +126,7 @@ impl ChainSynchronizer<ParticipatingEvent> {
         if !synchronizer.config.sync_to_genesis() {
             return Ok((
                 synchronizer,
-                effect_builder.announce_finished_syncing().ignore(),
+                effect_builder.announce_finished_chain_syncing().ignore(),
             ));
         }
 
--- a/node/src/components/chain_synchronizer.rs
+++ b/node/src/components/chain_synchronizer.rs
@@ -47,6 +47,12 @@ pub(crate) use metrics::Metrics;
 use operations::FastSyncOutcome;
 pub(crate) use operations::KeyBlockInfo;
 
+// #[derive(Debug, Serialize)]
+// pub(crate) enum LinearChainSyncAnnouncement {
+//     /// Chain sync finished.
+//     SyncStateFinished,
+// }
+
 #[derive(DataSize, Debug)]
 pub(crate) enum JoiningOutcome {
     /// We need to shutdown for upgrade as we downloaded a block from a higher protocol version.
@@ -120,9 +126,13 @@ impl ChainSynchronizer<ParticipatingEvent> {
             _phantom: PhantomData,
         };
 
-        // If we're not configured to sync-to-genesis, return without doing anything.
+        // If we're not configured to sync-to-genesis, return without doing anything but announcing
+        // that the sync process has finished.
         if !synchronizer.config.sync_to_genesis() {
-            return Ok((synchronizer, Effects::new()));
+            return Ok((
+                synchronizer,
+                effect_builder.announce_finished_syncing().ignore(),
+            ));
         }
 
         let effects = operations::run_sync_to_genesis_task(
--- a/node/src/components/chain_synchronizer.rs
+++ b/node/src/components/chain_synchronizer.rs
@@ -75,7 +75,10 @@ pub(crate) enum JoiningOutcome {
 /// A chain synchronizer announcement.
 #[derive(Debug)]
 pub(crate) enum ChainSynchronizerAnnouncement {
-    /// Synchronization has been finished.
+    /// The node has finished the synchronization it was doing (fast-sync or sync-to-genesis,
+    /// depending on config) and may now accept requests that are unsafe for nodes that are
+    /// synchronizing. Once this message is received, the only way for the peer to signal it's in
+    /// the syncing process is to reconnect.
     SyncFinished,
 }
 
--- a/node/src/components/chain_synchronizer.rs
+++ b/node/src/components/chain_synchronizer.rs
@@ -4,14 +4,7 @@ mod event;
 mod metrics;
 mod operations;
 
-use core::fmt;
-use std::{
-    collections::HashSet,
-    convert::Infallible,
-    fmt::{Debug, Display, Formatter},
-    marker::PhantomData,
-    sync::Arc,
-};
+use std::{collections::HashSet, convert::Infallible, fmt::Debug, marker::PhantomData, sync::Arc};
 
 use datasize::DataSize;
 use prometheus::Registry;
@@ -70,24 +63,6 @@ pub(crate) enum JoiningOutcome {
         validators_to_sign_immediate_switch_block: HashSet<PublicKey>,
         highest_block_header: BlockHeader,
     },
-}
-
-/// A chain synchronizer announcement.
-#[derive(Debug)]
-pub(crate) enum ChainSynchronizerAnnouncement {
-    /// The node has finished the synchronization it was doing (fast-sync or sync-to-genesis,
-    /// depending on config) and may now accept requests that are unsafe for nodes that are
-    /// synchronizing. Once this message is received, the only way for the peer to signal it's in
-    /// the syncing process is to reconnect.
-    SyncFinished,
-}
-
-impl Display for ChainSynchronizerAnnouncement {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        match self {
-            ChainSynchronizerAnnouncement::SyncFinished => write!(f, "chain sync finished"),
-        }
-    }
 }
 
 #[derive(DataSize, Debug)]
--- a/node/src/components/chain_synchronizer.rs
+++ b/node/src/components/chain_synchronizer.rs
@@ -4,7 +4,14 @@ mod event;
 mod metrics;
 mod operations;
 
-use std::{collections::HashSet, convert::Infallible, fmt::Debug, marker::PhantomData, sync::Arc};
+use core::fmt;
+use std::{
+    collections::HashSet,
+    convert::Infallible,
+    fmt::{Debug, Display, Formatter},
+    marker::PhantomData,
+    sync::Arc,
+};
 
 use datasize::DataSize;
 use prometheus::Registry;
@@ -47,12 +54,6 @@ pub(crate) use metrics::Metrics;
 use operations::FastSyncOutcome;
 pub(crate) use operations::KeyBlockInfo;
 
-// #[derive(Debug, Serialize)]
-// pub(crate) enum LinearChainSyncAnnouncement {
-//     /// Chain sync finished.
-//     SyncStateFinished,
-// }
-
 #[derive(DataSize, Debug)]
 pub(crate) enum JoiningOutcome {
     /// We need to shutdown for upgrade as we downloaded a block from a higher protocol version.
@@ -69,6 +70,21 @@ pub(crate) enum JoiningOutcome {
         validators_to_sign_immediate_switch_block: HashSet<PublicKey>,
         highest_block_header: BlockHeader,
     },
+}
+
+/// A chain synchronizer announcement.
+#[derive(Debug)]
+pub(crate) enum ChainSynchronizerAnnouncement {
+    /// Synchronization has been finished.
+    SyncFinished,
+}
+
+impl Display for ChainSynchronizerAnnouncement {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            ChainSynchronizerAnnouncement::SyncFinished => write!(f, "chain sync finished"),
+        }
+    }
 }
 
 #[derive(DataSize, Debug)]
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -774,6 +774,7 @@ where
             Message::Payload(payload) => {
                 effect_builder.announce_incoming(peer_id, payload).ignore()
             }
+            Message::SyncStateChanged => todo!(),
         })
     }
 
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -423,6 +423,13 @@ where
     ) {
         // Try to send the message.
         if let Some(connection) = self.outgoing_manager.get_route(dest) {
+            if msg.payload_is_unsafe_for_syncing_nodes() && self.syncing.contains(&dest) {
+                // We should never attempt to send an unsafe message to a peer that we know is still
+                // syncing. Since "unsafe" does usually not mean immediately catastrophic, we
+                // attempt to carry on, but warn loudly.
+                error!(kind=%msg.classify(), node_id=%dest, "sending unsafe message to syncing node");
+            }
+
             if let Err(msg) = connection.sender.send((msg, opt_responder)) {
                 // We lost the connection, but that fact has not reached us yet.
                 warn!(our_id=%self.context.our_id, %dest, ?msg, "dropped outgoing message, lost connection");
@@ -696,10 +703,6 @@ where
                     self.connection_completed(peer_id);
                 }
 
-                // By default, we assume that the peer is syncing. It'll send us the
-                // `SyncFinished` message in case it isn't.
-                let remote_peer_id = Some((peer_addr, peer_id));
-
                 effects.extend(
                     tasks::message_sender(
                         receiver,
@@ -707,7 +710,6 @@ where
                         self.outgoing_limiter
                             .create_handle(peer_id, peer_consensus_public_key),
                         self.net_metrics.queued_messages.clone(),
-                        remote_peer_id,
                     )
                     .instrument(span)
                     .event(move |_| Event::OutgoingDropped {
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -702,6 +702,9 @@ where
                     .mark_outgoing(now)
                 {
                     self.connection_completed(peer_id);
+
+                    // By default, we assume that newly connecting peer is joiner.
+                    self.update_syncing_nodes_set(peer_id, true);
                 }
 
                 effects.extend(
@@ -795,7 +798,7 @@ where
             }
             Message::SyncFinished => {
                 info!(%peer_id, "peer reported sync finished");
-                self.update_syncing_set(peer_id, false);
+                self.update_syncing_nodes_set(peer_id, false);
                 Effects::new()
             }
         })
@@ -809,7 +812,7 @@ where
 
     /// Updates a set of known joining nodes.
     /// If we've just connected to a non-joining node that peer will be removed from the set.
-    fn update_syncing_set(&mut self, peer_id: NodeId, is_syncing: bool) {
+    fn update_syncing_nodes_set(&mut self, peer_id: NodeId, is_syncing: bool) {
         // Update set of syncing peers.
         if is_syncing {
             self.syncing_nodes.insert(peer_id);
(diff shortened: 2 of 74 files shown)


rafal-ch:
Adding the `do not merge` label, because some housekeeping needs to be done after https://github.com/casper-network/casper-node/pull/3167 is merged.

node/src/components/chain_synchronizer.rs
```diff
@@ -65,12 +72,28 @@ pub(crate) enum JoiningOutcome {
     },
 }
 
+/// A chain synchronizer announcement.
+#[derive(Debug)]
+pub(crate) enum ChainSynchronizerAnnouncement {
+    /// Synchronization has been finished.
+    SyncFinished,
```
marc-casperlabs: ~We should be more specific what kind of sync we finished, since we have multiple ones (to genesis or fast sync) and I would like to avoid confusion. Just changing the name is probably sufficient.~

Let's put into the doc comments that it indicates the nodes has finished the kind of sync it was doing (which is not specified) and will cease to request unsafe data. Also mention that we rely on reconnecting to reset the syncing state.

marc-casperlabs: Also, the "one announcement per component" is a bit of a (widespread) anti-mattern, at least in my mind. This could just be a 

```rust
struct SyncFinishedAnnouncement;
```

instead.

node/src/components/small_network.rs
```diff
@@ -152,8 +152,8 @@ where
     /// Tracks whether a connection is symmetric or not.
     connection_symmetries: HashMap<NodeId, ConnectionSymmetry>,
 
-    /// Tracks nodes that have announced themselves as joining nodes.
-    joining_nodes: HashSet<NodeId>,
+    /// Tracks nodes that have announced themselves as nodes that are syncing.
+    syncing: HashSet<NodeId>,
```
marc-casperlabs: How about `syncing_nodes` instead?

node/src/components/small_network.rs
```diff
@@ -184,6 +184,9 @@ where
 
     /// The era that is considered the active era by the small network component.
     active_era: EraId,
+
+    /// The flag specifying if the node is in the chain sync process.
+    syncing_in_progress: bool,
```
marc-casperlabs: `sync_in_progress` or `synchronization_in_progress` is, I believe, a better term here? Maybe I'm being overly pedantic though.

node/src/components/small_network.rs
```diff
@@ -1075,6 +1089,13 @@ where
 
                 effects
             }
+            Event::SyncFinished => {
+                info!("notifying peers that chain sync has finished");
+                self.net_metrics.broadcast_requests.inc();
```
marc-casperlabs: Please remove this and move the other net metrics incrementation above into `fn broadcast_message` instead, this is way too error prone. I believe there should only be a single call to `.inc()` for almost any metric, along with one for `.dec()`.

node/src/components/small_network.rs
```diff
@@ -1075,6 +1089,13 @@ where
...
+                self.broadcast_message(Arc::new(Message::SyncFinished));
+                self.syncing_in_progress = false;
```
marc-casperlabs: It does not make a difference right now, but I'd prefer the order be swapped (set the flag first, then broadcast) just in case.

node/src/reactor/participating.rs
```diff
@@ -1361,6 +1368,13 @@ impl reactor::Reactor for Reactor {
                 );
                 self.dispatch_event(effect_builder, rng, reactor_event)
             }
+            ParticipatingEvent::ChainSynchronizerAnnouncement(
+                ChainSynchronizerAnnouncement::SyncFinished,
+            ) => self.dispatch_event(
+                effect_builder,
+                rng,
+                ParticipatingEvent::SmallNetwork(small_network::Event::SyncFinished),
```
marc-casperlabs: I think we should keep this in line with the rest of the code and implement `From<ChainSynchronizerAnnouncement>` (or even better `From<SyncFinishedAnnouncement>`, see comments above), instead of having the conversion here. Component events are supposed to be opaque to the reactor.

node/src/components/small_network/event.rs
```diff
@@ -95,6 +95,9 @@ pub(crate) enum Event<P> {
     /// Contract runtime announcement.
     #[from]
     ContractRuntimeAnnouncement(ContractRuntimeAnnouncement),
+
+    #[from]
+    SyncFinished,
```
marc-casperlabs: I think turning this into a `ChainSynchronizerAnnouncement(_)` (see comment below) or `SyncFinishedAnnouncement(_)` variant is better - with the proposed changes we would have to implement `From<_>` manually here.

Also I have no idea what `#[from]` does with the code as-is :)

node/src/components/small_network/message.rs
```diff
@@ -68,17 +68,17 @@ impl<P: Payload> Message<P> {
     #[inline]
     pub(super) fn payload_incoming_resource_estimate(&self, weights: &EstimatorWeights) -> u32 {
         match self {
-            Message::Handshake { .. } => 0,
+            Message::Handshake { .. } | Message::SyncFinished => 0,
```
marc-casperlabs: We do not allow a second handshake, thus we can get away with a hardcoded weight of 0 here. However, we should have a weight for `SyncFinished`, or we need to reject/disconnect whenever a client sends a `SyncFinished` message when it already has synced, to close this DOS vector.

node/src/components/small_network/tasks.rs
```diff
@@ -697,19 +687,19 @@ pub(super) async fn message_sender<P>(
     mut sink: SplitSink<FullTransport<P>, Arc<Message<P>>>,
     limiter: Box<dyn LimiterHandle>,
     counter: IntGauge,
-    remote_joiner_id: Option<(SocketAddr, NodeId)>,
+    remote_peer_id: Option<(SocketAddr, NodeId)>,
```
marc-casperlabs: I think we should remove this parameter entirely, as it duplicates state already found in the component. I made a mistake in adding it here a while ago, sorry!

Instead, we should put the check in the component itself, into the `send_message` function just before the call `connection.sender.send((msg, opt_responder))`. That way, information will always be up-to-date and no potentially stale state duplicated in the sender task.

It will be a tiny bit more complex (more hash set membership checks), but that is negligible overhead I think. If it turns out to be a problem, we can add it to the outgoing connection handle instead, but we expect this feature to go away anyway.

node/src/protocol.rs
```diff
@@ -137,12 +137,12 @@ impl Payload for Message {
...
 
-    fn is_unsafe_for_joiners(&self) -> bool {
+    fn is_unsafe_for_syncing_peers(&self) -> bool {
         match self {
             Message::Consensus(_) => false,
             Message::DeployGossiper(_) => false,
             Message::AddressGossiper(_) => false,
-            // Trie requests can deadlock between joiners.
+            // Trie requests can deadlock between syncing nodes.
             Message::GetRequest { tag, .. } if *tag == Tag::TrieOrChunk => true,
```
marc-casperlabs, line 146: We have a trait describing the same information now in the chain synchronizer, which we need to clean up (`CanUseJoiners`).

I think it'd be best to put the `CanUseJoiners` functionality onto the `Tag` for now:

```rust
impl Tag {
    fn can_use_joiners(self) -> bool;
}
```

For any function handling an item `T: Item`, we can then use `(<T as Item>::Tag).is_unsafe_for_joiners()`.

marc-casperlabs (dismissed):
I'd like to see some changes, the bare minimum being the "is syncing" bool being removed from the sender task.

node/src/components/chain_synchronizer.rs
marc-casperlabs: ~~I believe there is technically now a race condition where the following happens:~~

~~* Node finishes syncing, sends "syncing finished message" to peers.~~
~~* Reactor shuts down (reactor switch). A ton of trie requests are queued at the right moment and stuck long enough in their event queues to move past the reconnection (again, this is because we do not refresh node IDs on reactor change anymore).~~
~~* New reactor again syncs because sync to genesis was requested and it immediately sends the trie requests to peers.~~
~~* -> potential deadlock, as we have two peers backpressuring due to large amounts of outstanding requests.~~

~~This could be alleviated by not sending the syncing finished announcement ever, if we know archival sync is coming next.~~

Nvm, it the check is here :) https://github.com/casper-network/casper-node/pull/3179/files#diff-bc770989714a898b4e4f4932bfddacc6787c9d1e9c1cc133e4ea44fbdac7d606R148

node/src/reactor/participating.rs
goral09: They're not opaque just b/c we implement `Into`.

node/src/components/small_network/tasks.rs
rafal-ch: Done in 3568021397b436d142f9ec6a0c6c98db6e92a919

node/src/components/small_network.rs
rafal-ch: Updated in b6b3b3af763e02c445c408fab3506b1dcb62c548.

node/src/components/small_network.rs
rafal-ch: Updated in b6b3b3af763e02c445c408fab3506b1dcb62c548.

node/src/components/small_network.rs
rafal-ch: Right! Updated in 79c78a21cbb81f79e48ef6d09ff118dab281041f.

node/src/components/small_network/message.rs
rafal-ch: Ok, I'll disconnect from peer if it sends duplicated `SyncFinished` messages.

node/src/protocol.rs
rafal-ch: I'd rather postpone this change to another PR to avoid scope creep as neither the trait nor the function was introduced as part of this PR.

node/src/components/chain_synchronizer.rs
rafal-ch: More explanatory comment added in 706e01007dbdf0eb44f9096de0e5159bda54446b, although I prefer to stick to the "one announcement per component" approach for now if you're not feeling strongly about it.

node/src/components/small_network.rs
rafal-ch: As you prefer: eba559e347d5cef0e137c95972e795280bfafecc

node/src/components/small_network/event.rs
rafal-ch: For now I removed the superfluous `#[from]` attribute and added the missing comment here: d10a025b33873ae906d6851c913294354b5651c9

I'll probably consider changing this to `ChainSynchronizerAnnouncement` though. Please don't resolve yet.

node/src/components/small_network/message.rs
rafal-ch: It turned to be somewhat tricky to detect whether the notification from a given peer is first one or subsequent one. Therefore I decided to use the generic throttling mechanism.

Changes in this commit: b39d1fe990180664995dd4b6ec2cbed23c8c7ea3

node/src/components/small_network/event.rs
rafal-ch: Turned into proper announcement in eedcb1856b8ac54dee18404660b6ed061d5e7960.

node/src/reactor/participating.rs
marc-casperlabs: Emphasis on "supposed". If I had to design the system again, any of these announcements would take the form of 

```rust
pub struct FooAnnouncement(FooAnnouncementInner);

enum FooAnnouncementInner {
    // ...
}
```

node/src/effect.rs
```diff
@@ -1662,6 +1663,20 @@ impl<REv> EffectBuilder<REv> {
             .await
     }
 
+    /// Announce that the sync process has finished.
+    pub(crate) async fn announce_finished_syncing(self)
```
Fraser999: We have more than one notion of syncing in the node - I think we should be specific about this:
```suggestion
    pub(crate) async fn announce_finished_chain_syncing(self)
```

node/src/effect.rs
```diff
@@ -1662,6 +1663,20 @@ impl<REv> EffectBuilder<REv> {
...
+    where
+        REv: From<ChainSynchronizerAnnouncement>,
+    {
+        info!("announcing chain sync finished");
```
Fraser999: I'm not sure we need this log line.  We already have `info!("finished chain sync to genesis");` inside the chain sync code, and the other callsite is when the user has disabled sync-to-genesis, so seeing this log message might be a bit confusing.

Fraser999 (commented):
Partial review - I'll complete if you don't get enough approvals before then.

In the meantime, is it possible to have the small_network not advertise itself as syncing when constructed inside the participating reactor where sync-to-genesis is disabled?

node/src/components/chain_synchronizer.rs
```diff
@@ -120,9 +121,13 @@ impl ChainSynchronizer<ParticipatingEvent> {
             _phantom: PhantomData,
         };
 
-        // If we're not configured to sync-to-genesis, return without doing anything.
+        // If we're not configured to sync-to-genesis, return without doing anything but announcing
+        // that the sync process has finished.
         if !synchronizer.config.sync_to_genesis() {
-            return Ok((synchronizer, Effects::new()));
+            return Ok((
```
marc-casperlabs, line 127: ~~If this is called in the joiner with sync to genesis disabled, won't this announce the sync completed immediately, even though we still need to fast sync?~~ nvm, I misssed the specialization.

node/src/components/small_network.rs
```diff
@@ -152,22 +157,30 @@ where
...
-    /// Tracks nodes that have announced themselves as joining nodes.
-    joining_nodes: HashSet<NodeId>,
+    /// Tracks nodes that have announced themselves as nodes that are syncing.
+    syncing_nodes: HashSet<NodeId>,
 
-    /// Channel signaling a shutdown of the small network.
+    /// Channel signaling a shutdown of the server.
     // Note: This channel is closed when `SmallNetwork` is dropped, signalling the receivers that
-    // they should cease operation.
+    // they should cease operation. It is also replaced with a new channel when we finished the
```
marc-casperlabs: This is not correct - we should _not_ be triggering the shutdown sender in any case anymore except when actually shutting down.

node/src/components/small_network.rs
```diff
@@ -740,14 +762,17 @@ where
...
-                DialRequest::Dial { addr, span } => effects.extend(
-                    tasks::connect_outgoing(self.context.clone(), addr)
-                        .instrument(span.clone())
-                        .event(|outgoing| Event::OutgoingConnection {
-                            outgoing: Box::new(outgoing),
-                            span,
-                        }),
-                ),
+                DialRequest::Dial { addr, span } => {
+                    info!(is_syncing = ?self.context.is_syncing, "connecting");
```
marc-casperlabs: Let's not commit these debug messages here, they will spam the logs. There is detailed information about connection states already available in the outgoing connection statemachine handling.

node/src/components/small_network.rs
```diff
@@ -793,12 +818,14 @@ where
...
     /// Updates a set of known joining nodes.
     /// If we've just connected to a non-joining node that peer will be removed from the set.
-    fn update_joining_set(&mut self, peer_id: NodeId, is_joiner: bool) {
-        // Update set of joining peers.
-        if is_joiner {
-            self.joining_nodes.insert(peer_id);
+    fn update_syncing_nodes_set(&mut self, peer_id: NodeId, is_syncing: bool) {
+        // Update set of syncing peers.
+        if is_syncing {
+            info!(%peer_id, "is syncing");
```
marc-casperlabs: Please lower these to debug level, otherwise SRE will hunt me down and make me fix it.

node/src/components/small_network.rs
```diff
@@ -1076,6 +1103,12 @@ where
 
                 effects
             }
+            Event::ChainSynchronizerAnnouncement(ChainSynchronizerAnnouncement::SyncFinished) => {
+                info!("node has finished syncing");
```
marc-casperlabs: This is not a message we should put out - it belongs in fast sync.

I'd support a (potentially debug) style message that states that the networking component was _informed_ about sync completion though.

node/src/components/small_network/message.rs
```diff
@@ -39,9 +39,9 @@ pub(crate) enum Message<P> {
         /// A self-signed certificate indicating validator status.
         #[serde(default)]
         consensus_certificate: Option<ConsensusCertificate>,
-        /// Joiner node flag.
+        /// Holds the information whether the node is syncing.
```
marc-casperlabs: Maybe just "Whether the node is syncing." ?

node/src/effect.rs
marc-casperlabs: Regardless, I'd prefer we don't put log messages into the effect builder if we can avoid it.

marc-casperlabs (commented):
Still some question :)

Overall we need to really watch what log messages we put into networking, as production will see these 2 orders of magnitude more often then devs in testing (we get complaints about spammy logs). Anything `INFO` and above should be used very sparingly!

marc-casperlabs:
> In the meantime, is it possible to have the small_network not advertise itself as syncing when constructed inside the participating reactor where sync-to-genesis is disabled?

It could, it'd just add more logic. The current approach just reused the functionality we need anyway, and the savings is a 1-second reconnection cycle from clients, so we opted for that. As a nice side effect, we exercise the otherwise rarely hit code path :)

node/src/protocol.rs
marc-casperlabs: Fair enough!

node/src/components/small_network/message.rs
marc-casperlabs: Works for me for now!

Fraser999 (approved):
As discussed elsewhere - optimistic approval by @marc-casperlabs and myself given the outstanding comments are addressed.

node/src/effect.rs
rafal-ch: Updated in 5b91f0f74616a4c60822d297c58f08523917ec38

node/src/effect.rs
rafal-ch: Log message removed in 5b91f0f74616a4c60822d297c58f08523917ec38

node/src/components/small_network.rs
rafal-ch: Change to the comment reverted in 5b91f0f74616a4c60822d297c58f08523917ec38

node/src/components/small_network.rs
rafal-ch: Comment removed in 5b91f0f74616a4c60822d297c58f08523917ec38

node/src/components/small_network.rs
rafal-ch: Done in 5b91f0f74616a4c60822d297c58f08523917ec38

node/src/components/small_network.rs
rafal-ch: I removed the message in 5b91f0f74616a4c60822d297c58f08523917ec38 as it indeed doesn't bring much value.

node/src/components/small_network/message.rs
rafal-ch: Comment changed in 5b91f0f74616a4c60822d297c58f08523917ec38

PR #3178, opened 2022-07-06
Bring feat-2.0 up to date with dev since `a3aefaa`

georgepisaltu:
This PR brings all recent commits since https://github.com/casper-network/casper-node/commit/a3aefaa13cb5b5006cf253f259b8abdf4c526996 from the dev branch to the feat-2.0 branch.

Merge had no conflicts.

PR #3180, opened 2022-07-07
Bring feat-2.0 up to date with dev since `a3aefaa`

georgepisaltu:
This PR brings all recent commits since https://github.com/casper-network/casper-node/commit/a3aefaa13cb5b5006cf253f259b8abdf4c526996 from the dev branch to the feat-2.0 branch.

Merge had no conflicts, but some changes in EE testing needed to be addressed. `InMemoryWasmTestBuilder` use was replaced with `LmdbWasmTestBuilder` in https://github.com/casper-network/casper-node/commit/336f65b02882e7d213ac2b3c2197add70b653f71.

This PR also brings in more recent commits from dev than #3178, which also had the same WASM test builder problem.


Fraser999 (approved):
LGTM.  Thanks for calling out the commit you added to fix the compiler errors - very helpful indeed.

PR #3181, opened 2022-07-07
Update json-rpc to avoid some breaking changes

### LINKED ISSUE #3175: Revert breaking changes in JSON-RPC server

Fraser999:
This PR updates the `casper-json-rpc` crate to avoid some breaking changes in the node JSON-RPC server.

Specifically:
* it allows the `params` field to be null, interpreting that as an empty array, or equivalently, a `None` value
* it relaxes the range of error codes for which it supports creating an error response, since we currently use reserved values
* it can be configured to ignore extra fields provided in JSON-RPC requests (our usage in node configures the server to ignore extra fields as per current behaviour)

Closes #3175.

### DIFF
--- a/json_rpc/src/error.rs
+++ b/json_rpc/src/error.rs
@@ -112,7 +112,7 @@ impl Error {
     ///
     /// Other than when providing a [`ReservedErrorCode`], the converted "code" must not fall in the
     /// reserved range as defined in the JSON-RPC specification, i.e. it must not be between -32768
-    /// and -32000 inclusive.
+    /// and -32100 inclusive.
     ///
     /// If the converted code is within the reserved range when it should not be, or if
     /// JSON-encoding `additional_data` fails, the returned `Self` is built from
@@ -121,7 +121,7 @@ impl Error {
     pub fn new<C: ErrorCodeT, T: Serialize>(error_code: C, additional_info: T) -> Self {
         let (code, message): (i64, &'static str) = error_code.into();
 
-        if !C::is_reserved() && code >= -32768 && code <= -32000 {
+        if !C::is_reserved() && code >= -32768 && code <= -32100 {
             warn!(%code, "provided json-rpc error code is reserved; returning internal error");
             let (code, message) = ReservedErrorCode::InternalError.into();
             return Error {
--- a/json_rpc/src/error.rs
+++ b/json_rpc/src/error.rs
@@ -114,6 +114,10 @@ impl Error {
     /// reserved range as defined in the JSON-RPC specification, i.e. it must not be between -32768
     /// and -32100 inclusive.
     ///
+    /// Note that in an upcoming release, the restriction will be tightened to disallow error codes
+    /// in the implementation-defined server-errors range.  I.e. codes in the range -32768 to -32000
+    /// inclusive will be disallowed.
+    ///
     /// If the converted code is within the reserved range when it should not be, or if
     /// JSON-encoding `additional_data` fails, the returned `Self` is built from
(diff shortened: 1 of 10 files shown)


json_rpc/src/error.rs
```diff
@@ -112,7 +112,7 @@ impl Error {
     ///
     /// Other than when providing a [`ReservedErrorCode`], the converted "code" must not fall in the
     /// reserved range as defined in the JSON-RPC specification, i.e. it must not be between -32768
-    /// and -32000 inclusive.
+    /// and -32100 inclusive.
```
georgepisaltu, line 115: The specification states "The error codes from and including -32768 to -32000 are reserved for pre-defined errors", with `-32099` to `-32000` being "reserved for implementation-defined server-errors". I think this doc comment is ambiguous in this sense. Maybe if we added the bit about implementation-defined server-errors, it would be better.

json_rpc/src/request/params.rs
```diff
@@ -31,11 +34,7 @@ impl Params {
         };
 
         match params {
-            Value::Null => err_invalid_request(
-                "If present, 'params' must be an Array or Object, but was 'null'. If not required \
-                for this request, omit the field or provide an empty Array '[]' or empty Object \
-                '{}'",
-            ),
+            Value::Null => Ok(Params::Array(vec![])),
```
georgepisaltu, line 37: I think we should also mark this as deprecated right now so that in the major upcoming release we can safely delete this custom logic. I was thinking about logging this, but it doesn't make sense in an RPC context, we would have to let the users know in the RPC response. Maybe it would just be enough to mention this in the changelog, let me know what you think.

georgepisaltu (commented):
Looks good, a couple of comments.

json_rpc/src/request/params.rs
Fraser999: It's not a change when compared to the currently-published version, but looking at the changelog, I need to update it to cover the changes here.

json_rpc/src/error.rs
Fraser999: Done in [e0ecbc1](https://github.com/casper-network/casper-node/pull/3181/commits/e0ecbc1a351833231595c0bc9ef6e580fee30532).

json_rpc/src/request/params.rs
Fraser999: Done in [e0ecbc1](https://github.com/casper-network/casper-node/pull/3181/commits/e0ecbc1a351833231595c0bc9ef6e580fee30532).

goral09 (approved):

rafal-ch (changes requested):
Blocking temporarily since I have strange readings from `casper-test`. I'll unlock asap.

georgepisaltu (approved):
LGTM

rafal-ch:
> Blocking temporarily since I have strange readings from `casper-test`. I'll unlock asap.

I used `casper-test` version that predates the alignment with RPC server changes.

I checked the `rpc_jsonrpc_error_codes.rs` test on `1.4.6` and it's passing correctly. On your branch, though, it fails with:
```
errors="'non_existing' is not a supported json-rpc method on this server" is not of type "null"
```

I'm unblocking the PR right now as I see no other issues. I'f your ok with the above, I think it's good to go as is.

rafal-ch (approved):
LGTM with [single comment](https://github.com/casper-network/casper-node/pull/3181#issuecomment-1177632523) that might be worth checking.

Fraser999:
bors merge

Fraser999:
bors merge

PR #3187, opened 2022-07-08
Add chainspec hash to network handshake

### LINKED ISSUE #3184: Update small_network handshake to use chainspec hash
The small_network::Message::Handshake should have the network_name and protocol_version fields replaced with a new field chainspec_hash: Digest

georgepisaltu:
Fixes #3184 

This PR adds `chainspec_hash` field to the handshake message in the `small_network` component. When handshaking with current or newer versions, the node now checks that the peer it's connecting with is running on the same chainspec by comparing the hashes.

Deviating from the original requirements in the issue, the `network_name` field cannot be removed as it is part of the `v1.0.0` handshake and therefore it must be present to not break backwards compatibility. Furthermore, we decided not to remove the `protocol_version` field as well.

We should at some point add a new `V1_5_0_Message` entry in `small_network/message.rs` and a unit test for it, similar to other changes in the past.

### DIFF
--- a/node/src/components/small_network/error.rs
+++ b/node/src/components/small_network/error.rs
@@ -189,6 +189,9 @@ pub enum ConnectionError {
     /// This is usually a bug.
     #[error("handshake sink/stream could not be reunited")]
     FailedToReuniteHandshakeSinkAndStream,
+    /// Peer is running a different chainspec.
+    #[error("peer is on different chainspec")]
+    WrongChainspecHash,
 }
 
 /// IO operation that can time out or close.
--- a/node/src/components/small_network/error.rs
+++ b/node/src/components/small_network/error.rs
@@ -1,5 +1,6 @@
 use std::{error, io, net::SocketAddr, result, sync::Arc};
 
+use casper_hashing::Digest;
 use casper_types::{crypto, ProtocolVersion, SecretKey};
 use datasize::DataSize;
 use openssl::{error::ErrorStack, ssl};
@@ -151,6 +152,13 @@ pub enum ConnectionError {
     /// Peer reported an incompatible version.
     #[error("peer is running incompatible version: {0}")]
     IncompatibleVersion(ProtocolVersion),
+    /// Peer is using a different chainspec.
+    #[error("peer is using a different chainspec, hash: {0}")]
+    WrongChainspecHash(Digest),
+    /// Peer should have included the chainspec hash in the handshake message,
+    /// but didn't.
+    #[error("peer did not include mandatory chainspec hash in the handshake")]
+    MissingChainspecHash,
     /// Peer did not send any message, or a non-handshake as its first message.
     #[error("peer did not send handshake")]
     DidNotSendHandshake,
@@ -189,9 +197,6 @@ pub enum ConnectionError {
     /// This is usually a bug.
     #[error("handshake sink/stream could not be reunited")]
     FailedToReuniteHandshakeSinkAndStream,
-    /// Peer is running a different chainspec.
-    #[error("peer is on different chainspec")]
-    WrongChainspecHash,
 }
 
 /// IO operation that can time out or close.
--- a/node/src/components/small_network/error.rs
+++ b/node/src/components/small_network/error.rs
@@ -157,7 +157,7 @@ pub enum ConnectionError {
     WrongChainspecHash(Digest),
     /// Peer should have included the chainspec hash in the handshake message,
     /// but didn't.
-    #[error("peer did not include mandatory chainspec hash in the handshake")]
+    #[error("peer did not include chainspec hash in the handshake when it was required")]
     MissingChainspecHash,
     /// Peer did not send any message, or a non-handshake as its first message.
     #[error("peer did not send handshake")]
--- a/node/src/components/small_network/message.rs
+++ b/node/src/components/small_network/message.rs
@@ -15,7 +15,7 @@ use serde::{
     Deserialize, Deserializer, Serialize, Serializer,
 };
 
-use crate::{effect::EffectBuilder, types::NodeId};
+use crate::{effect::EffectBuilder, types::NodeId, utils::opt_display::OptDisplay};
 
 use super::counting_format::ConnectionId;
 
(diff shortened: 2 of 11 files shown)


node/src/components/small_network/message.rs
```diff
@@ -260,15 +264,22 @@ impl<P: Display> Display for Message<P> {
                 protocol_version,
                 consensus_certificate,
                 is_joiner,
+                chainspec_hash,
             } => {
                 write!(
                     f,
-                    "handshake: {}, public addr: {}, protocol_version: {}, consensus_certificate: , is_joiner: {}",
```
marc-casperlabs, line 266: This was already broken, we need to fix it. Adding the `is_joiner: {}` at the end was wrong, the `write` only writes until the consensus certificate.

So, let's move the `is_joiner` to the left so that the string ends with `"consensus certificate: ".

Further down we need to add another `f.write_str(" chainspec_hash")`

marc-casperlabs: That's the quick fix. My preferred fix would be this though:

```rust
struct OptDisplay<'a, T>(&T);

impl<'a, T> Display for OptDisplay<'a, T> {
    #[inline]
    fn fmt(&self, f: Formatter<'_>) -> fmt::Result {
        match self {
            None => f.write_str("-")
            Some(val) => val.fmt(f)
        }
    }
}
```

and then use 

```rust
write!(f, "handshake: {}, public addr: {}, protocol_version: {}, consensus_certificate: {}, is_joiner: {}, chainspec_hash: {}",
network_name, public_addr, protocol_version, OptDisplay(&consensus_certificate) is_joiner, OptDisplay(&chainspec_hash)
```

node/src/components/small_network/tasks.rs
```diff
@@ -429,6 +430,14 @@ where
             return Err(ConnectionError::WrongNetwork(network_name));
         }
 
+        // We check the chainspec hash to ensure peer is running on the
+        // same chainspec as us.
+        if let Some(hash) = chainspec_hash {
```
marc-casperlabs: We could debate whether we want to return an error if the chainspec hash is missing - @Fraser999 what do you think?

marc-casperlabs (commented):
Looks good to me overall, just some minor corrections.

node/src/components/small_network/message.rs
Fraser999: I like the `OptDisplay` way, but I'd go for writing "none" rather than "-" to be a little more explicit.  Not a big deal though.

node/src/components/small_network/tasks.rs
georgepisaltu: Maybe it should be an error if `context.reject_incompatible_versions` is set? Kind of what is happening below with protocol version.

Fraser999: I think I'd prefer to have it non-optional in the `Handshake` variant.

We can still use `#[serde(default)]` with it.

node/src/components/small_network/message.rs
georgepisaltu: I think `OptDisplay` is a good idea, but it would be a general purpose thing which would belong in `utils` probably. I would go with the quick fix and add a "todo" here to replace with a robust solution, but that would be better suited to an issue where we do this for all displays in the code. WDYT?

node/src/components/small_network/error.rs
```diff
@@ -189,6 +189,9 @@ pub enum ConnectionError {
     /// This is usually a bug.
     #[error("handshake sink/stream could not be reunited")]
     FailedToReuniteHandshakeSinkAndStream,
+    /// Peer is running a different chainspec.
+    #[error("peer is on different chainspec")]
```
Fraser999:
```suggestion
    #[error("peer is using a different chainspec")]
```

node/src/components/small_network/error.rs
```diff
@@ -189,6 +189,9 @@ pub enum ConnectionError {
...
+    WrongChainspecHash,
```
Fraser999: Let's move this variant to immediately after `IncompatibleVersion` to keep it alongside related errors please.

Also, please include the peer's chainspec hash similar to what we do for `WrongNetwork` and `IncompatibleVersion` to help with debugging.

node/src/components/small_network/tasks.rs
Fraser999: Also, we need to move this check to after the protocol version check just below, otherwise the tarpitting will be effectively dead code.

Fraser999 (changes requested):
We should remove the `reject_incompatible_versions` config option since it now has no effect (this PR makes it effectively always true regardless of the value the user sets it to in the config file).  When making that change, please update the comments for `tarpit_version_threshold` which mention this config option.

marc-casperlabs:
Oh and most importantly, the changelog needs to be updated!

node/src/components/small_network/tasks.rs
marc-casperlabs: > I think I'd prefer to have it non-optional in the Handshake variant.

Uh, so what's the `Default::default` value of a non-`Option<T>` `Handshake`?

node/src/components/small_network/message.rs
georgepisaltu: Done.

node/src/components/small_network/tasks.rs
georgepisaltu: We had it be an `Option` because it makes more sense if you want to see if it was included in the handshake message, as a hash itself wouldn't have a default value. It must be included in the message anyway from this version onwards, please see the updated handling of this case.

node/src/components/small_network/error.rs
georgepisaltu: Done.

node/src/components/small_network/error.rs
georgepisaltu: Done.

node/src/components/small_network/tasks.rs
Fraser999: We have a default impl for `Digest` which I'd prefer over an option for this.  Mainly so we don't have the `MissingChainspecHash` error variant.

georgepisaltu: But in the case where you don't get a chainspec hash in the handshake message, and the deserializer fills in the default, `WrongChainspecHash` is not the semantically correct error here because none was actually provided. I think using `Digest::default` actually subtracts value (from a user point of view) and is actually not at all expensive to maintain for us.

marc-casperlabs: I'm not a fan of a default `Digest` - if it's not there, we just don't want to use it? Then there'd also not be a `MissingChainspecHash`.

Fraser999 (approved):

Fraser999:
bors merge

Issue #3164, opened 2022-07-04
Implement Stream for multiplexer


devendran-m:
The `Multiplexer` currently only has a `Sink` implementation, it needs a respective `Stream` implementation for the receiving side as well.

marc-casperlabs:
ddeac1f68b0b6924c6b1639c37c11a38c81dfe2d now commited to `feat-muxink`.

PR #3188, opened 2022-07-08
Demultiplexer for muxink

SamuelSchlesinger:
Resolves https://github.com/casper-network/casper-node/issues/3164

Adds a `Demultiplexer` to go along with the `Multiplexer` in `muxink`. Tested with a `Stream` which is actually unsafe to call after receiving a `Poll::Ready(None)`.

The only outstanding bit is what to do with messages coming in for a channel which isn't being listened on. I've left a comment with three possible ways to handle this. It's not strictly necessary to do anything, which is option 3 and the default if we merge this PR.

Another possible improvement is to store the waker for the task associated with each `DemultiplexerHandle` and wake it when it's message is ready. I'm not sure this is in scope for this ticket.

### DIFF
--- a/muxink/src/demux.rs
+++ b/muxink/src/demux.rs
@@ -1,0 +1,256 @@
+//! Stream demultiplexing
+//!
+//! Demultiplexes a Stream of Bytes into multiple channels. Up to 256 channels are supported, and
+//! if messages are present on a channel but there isn't an associated DemultiplexerHandle for that
+//! channel, then the Stream will never poll as Ready.
+
+use std::{
+    pin::Pin,
+    sync::{Arc, Mutex},
+    task::{Context, Poll},
+};
+
+use bytes::Bytes;
+use futures::{stream::Fuse, Stream, StreamExt};
+
+/// A frame demultiplexer.
+///
+/// A demultiplexer is not used directly, but used to spawn demultiplexing handles.
+///
+/// TODO What if someone sends data to a channel for which there is no handle?
+///      I can think of two reasonable responses:
+///         1. return an error to the handle which saw this message.
+///         2. drop all messages we receive which don't have a corresponding `DemultiplexerHandle`
+///            yet.
+///         3. allow messages to sit forever and block the rest of the handles, preferring whoever
+///            is sending us the messages to filter out ones which aren't for a channel we're
+///            listening on. this is already what happens if a `DemultiplexerHandle` for any
+///            channel which has messages in the stream doesn't ever take them out.
+pub struct Demultiplexer<S> {
+    stream: Fuse<S>,
+    next_frame: Option<(u8, Bytes)>,
+}
+
+impl<S: Stream> Demultiplexer<S> {
+    /// Creates a new demultiplexer with the given underlying stream.
+    pub fn new(stream: S) -> Demultiplexer<S> {
+        Demultiplexer {
+            stream: stream.fuse(),
+            next_frame: None,
+        }
+    }
+}
+
+impl<S> Demultiplexer<S> {
+    /// Creates a handle listening for frames on the given channel.
+    ///
+    /// Any item on this channel sent to the `Stream` underlying the `Demultiplexer` we used to
+    /// create this handle will be read only when all other messages for other channels have been
+    /// read first. If one has handles on the same channel created via the same underlying
+    /// `Demultiplexer`, each message on that channel will only be received by one of the handles.
+    /// Unless this is desired behavior, this should be avoided.
+    pub fn create_handle(demux: Arc<Mutex<Self>>, channel: u8) -> DemultiplexerHandle<S> {
+        DemultiplexerHandle {
+            channel,
+            demux: demux.clone(),
+        }
+    }
+}
+
+/// A handle to a demultiplexer.
+///
+/// A handle is bound to a specific channel, see [`Demultiplexer::create_handle`] for details.
+pub struct DemultiplexerHandle<S> {
+    /// Which channel this handle is listening on
+    channel: u8,
+    /// A reference to the underlying demultiplexer.
+    demux: Arc<Mutex<Demultiplexer<S>>>, // (probably?) make sure this is a stdmutex
+}
+
+impl<S> Stream for DemultiplexerHandle<S>
+where
+    S: Stream<Item = Bytes> + Unpin,
+{
+    // TODO Result<Bytes, Error>
+    type Item = Bytes;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // Lock the demultiplexer
+        let mut demux = match self.demux.as_ref().try_lock() {
+            Err(_err) => panic!("TODO"), // TODO return Err("TODO")
+            Ok(guard) => guard,
+        };
+
+        // If next_frame has a suitable frame for this channel, return it in a Poll::Ready. If it has
+        // an unsuitable frame, return Poll::Pending. Otherwise, we attempt to read from the stream.
+        if let Some((ref channel, ref bytes)) = demux.next_frame {
+            if *channel == self.channel {
+                let bytes = bytes.clone();
+                demux.next_frame = None;
+                return Poll::Ready(Some(bytes));
+            } else {
+                return Poll::Pending;
+            }
+        }
+
+        // Try to read from the stream, placing the frame into next_frame and returning
+        // Poll::Pending if its in the wrong channel, otherwise returning it in a Poll::Ready.
+        match demux.stream.poll_next_unpin(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Some(bytes)) => {
+                let channel: u8 = *&bytes[0..1][0];
+                let frame = bytes.slice(1..).clone();
+                if channel == self.channel {
+                    Poll::Ready(Some(frame))
+                } else {
+                    demux.next_frame = Some((channel, frame));
+                    Poll::Pending
+                }
+            }
+            Poll::Ready(None) => Poll::Ready(None),
+        }
+
+        // TODO: figure out when we are being polled again, does it work correctly (see waker) or
+        //       will it cause inefficient races? do we need to call wake? probably. (possibly
+        //       necessary) can have table of wakers to only wake the right one.
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::marker::Unpin;
+
+    use super::*;
+    use futures::{FutureExt, Stream, StreamExt};
+
+    // This stream is used because it is not safe to call it after it returns
+    // [`Poll::Ready(None)`], whereas many other streams are. The interface for
+    // streams says that in general it is not safe, so it is important to test
+    // using a stream which has this property as well.
+    struct TestStream<T> {
+        // The items which will be returned by the stream in reverse order
+        items: Vec<T>,
+        // Once this is set to true, this `Stream` will panic upon calling [`Stream::poll_next`]
+        finished: bool,
+    }
+
+    impl<T> TestStream<T> {
+        fn new(mut items: Vec<T>) -> Self {
+            // We reverse the items as we use the pop method to remove them one by one,
+            // thus they come out in the order specified by the `Vec`.
+            items.reverse();
+            TestStream {
+                items,
+                finished: false,
+            }
+        }
+    }
+
+    // We implement Unpin because of the constraint in the implementation of the
+    // `DemultiplexerHandle`.
+    impl<T> Unpin for TestStream<T> {}
+
+    impl<T> Stream for TestStream<T> {
+        type Item = T;
+
+        fn poll_next(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+            // Panic if we've already emitted [`Poll::Ready(None)`]
+            if self.finished {
+                panic!("polled a TestStream after completion");
+            }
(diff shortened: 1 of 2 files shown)


muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+use bytes::Bytes;
+use futures::{stream::Fuse, Stream, StreamExt};
+
+/// A frame demultiplexer.
+///
+/// A demultiplexer is not used directly, but used to spawn demultiplexing handles.
+///
+/// TODO What if someone sends data to a channel for which there is no handle?
+///      I can think of two reasonable responses:
+///         1. return an error to the handle which saw this message.
```
marc-casperlabs: I think this is the best option, as it finds bugs. We need to ensure that _all_ handles receive an error when this happens and the stream is essentially shut down.

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
+//! Stream demultiplexing
```
marc-casperlabs, line 1: The demuxing should probably go into the `mux.rs` file, for now we are keeping both directions in a single file.

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+///      I can think of two reasonable responses:
+///         1. return an error to the handle which saw this message.
+///         2. drop all messages we receive which don't have a corresponding `DemultiplexerHandle`
+///            yet.
+///         3. allow messages to sit forever and block the rest of the handles, preferring whoever
+///            is sending us the messages to filter out ones which aren't for a channel we're
+///            listening on. this is already what happens if a `DemultiplexerHandle` for any
+///            channel which has messages in the stream doesn't ever take them out.
+pub struct Demultiplexer<S> {
+    stream: Fuse<S>,
```
marc-casperlabs, line 22: We need some comments on these fields, no matter how boring :)

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+            stream: stream.fuse(),
+            next_frame: None,
+        }
+    }
+}
+
+impl<S> Demultiplexer<S> {
+    /// Creates a handle listening for frames on the given channel.
+    ///
+    /// Any item on this channel sent to the `Stream` underlying the `Demultiplexer` we used to
```
marc-casperlabs, line 71: I know I should be the absolute last person to talk about this but can "we" tone used here?

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+    /// create this handle will be read only when all other messages for other channels have been
```
marc-casperlabs, line 72: Is this slightly wrong? I would put it differently, e.g. "Items received through a given handle may be blocked if other handles on the same `Demultiplexer` are not polled at the same time."

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+    channel: u8,
+    /// A reference to the underlying demultiplexer.
+    demux: Arc<Mutex<Demultiplexer<S>>>, // (probably?) make sure this is a stdmutex
+}
+
+impl<S> Stream for DemultiplexerHandle<S>
+where
+    S: Stream<Item = Bytes> + Unpin,
+{
+    // TODO Result<Bytes, Error>
```
marc-casperlabs, line 123: I think it'd be good to do a merge of `feat-muxink` into this branch before going ahead; in that there's also a hint on how to do this (the new fragmentizer has an error type).

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+            if *channel == self.channel {
+                let bytes = bytes.clone();
+                demux.next_frame = None;
+                return Poll::Ready(Some(bytes));
+            } else {
+                return Poll::Pending;
+            }
+        }
+
+        // Try to read from the stream, placing the frame into next_frame and returning
```
marc-casperlabs, line 145: `next_frame` instead of next_frame maybe? Same for other bits of code in comments, like `Poll::Ready`.

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+        // Poll::Pending if its in the wrong channel, otherwise returning it in a Poll::Ready.
```
marc-casperlabs, line 146: I think it's "it's" here.

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+                    Poll::Ready(Some(frame))
+                } else {
+                    demux.next_frame = Some((channel, frame));
+                    Poll::Pending
+                }
+            }
+            Poll::Ready(None) => Poll::Ready(None),
+        }
+
+        // TODO: figure out when we are being polled again, does it work correctly (see waker) or
```
marc-casperlabs, line 162: We absolutely need to ensure the waker support is there, if needed.

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+                return Poll::Pending;
+            }
+        }
+
+        // Try to read from the stream, placing the frame into next_frame and returning
+        // Poll::Pending if its in the wrong channel, otherwise returning it in a Poll::Ready.
+        match demux.stream.poll_next_unpin(cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Some(bytes)) => {
+                let channel: u8 = bytes[0..1][0];
```
marc-casperlabs: This will panic when reading 0-sized frames, which I believe an adversary could construct. This needs a custom error variant (compare the `Defragmentizer`'s stream implementation).

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+    use super::*;
+    use futures::{FutureExt, Stream, StreamExt};
+
+    // This stream is used because it is not safe to call it after it returns
+    // [`Poll::Ready(None)`], whereas many other streams are. The interface for
+    // streams says that in general it is not safe, so it is important to test
+    // using a stream which has this property as well.
+    struct TestStream<T> {
+        // The items which will be returned by the stream in reverse order
+        items: Vec<T>,
```
marc-casperlabs, line 181: We could use a `VecDeque` instead, that will allow us to add items to it after construction, should that feature be necessary in later tests. We could simulate a temporarily blocked stream that way (either with an explicit `finish()` method or by storing `Option`s of items). Just a thought at this point though.

muxink/src/demux.rs
```diff
@@ -0,0 +1,254 @@
...
+                Poll::Ready(Some(t))
+            } else {
+                // Before we return None, make sure we set finished to true so that calling this
+                // again will result in a panic, as the specification for `Stream` tells us is
+                // possible with an arbitrary implementation.
+                self.finished = true;
+                Poll::Ready(None)
+            }
+        }
+    }
```
marc-casperlabs, line 220: We should add a simple test that ensures (via `#[should_panic]`) that this stream will indeed panic if polled after completion.

marc-casperlabs (changes requested):
There's some things missing still, as noted in the comments.

I fixed some of the clippy warnings and formatting issues (our nightly rust version does some weird formatting in CI currently :( ), but I did not want to merge this yet.

I'd highly recommend merging `feat-muxink` into this branch - most of the merge conflicts should be easy to solve, as they just related to already deleted code.

muxink/src/demux.rs
SamuelSchlesinger: Why do you want it within one module? If it can be separated dependency-wise, its usually more readable to have separate files IMO.

muxink/src/demux.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// Here, we write the logic for accessing and modifying the bit-field representing the active
+// channels.
+impl<S> Demultiplexer<S> {
+    fn activate_channel(&mut self, channel: u8) {
+        self.active_channels[(channel / 8) as usize] |=
+            2u8.checked_pow((channel % 8) as u32).unwrap();
```
georgepisaltu, line 49: We could just use bit operators here. I don't think it's necessarily faster since I expect the compiler to optimize it away anyway, but it's cleaner than having `checked_pow` followed by `unwrap`. We could use bit operators for selecting the channel as well.

Same for usage below.

muxink/src/demux.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// Any item on this channel sent to the `Stream` underlying the `Demultiplexer` we used to
+    /// create this handle will be read only when all other messages for other channels have been
+    /// read first. If one has handles on the same channel created via the same underlying
+    /// `Demultiplexer`, each message on that channel will only be received by one of the handles.
+    /// Unless this is desired behavior, this should be avoided.
+    pub fn create_handle(demux: Arc<Mutex<Self>>, channel: u8) -> DemultiplexerHandle<S> {
+        let mut guard = match demux.as_ref().try_lock() {
+            Err(_err) => panic!("TODO"),
+            Ok(guard) => guard,
+        };
```
georgepisaltu, line 80: Ideally we'd have a wrapper object which holds the `Demultiplexer` in an `Arc<Mutex>` so that we don't do any synchronization logic here in the `create_handle` code (because the main logic of this piece of code is validate channel, create new shared reference). Users of that demux should choose to lock/unlock and otherwise manage their usage as necessary and are responsible for poisoned locks as well even as the code stands right now. If we don't go with the wrapper, consider having `create_handle` as a standalone public function. We could even add a type for `Arc<Mutex<Demultiplexer<S>>>`.

The wrapper solution would look like
```
// Don't name it "Wrapper"
struct Wrapper<S> {
    inner: Arc<Mutex<Demultiplexer<S>>>
}
```
You can then add a method to clone the reference inside and use it in `create_handle` as well.

muxink/src/demux.rs
```diff
@@ -0,0 +1,325 @@
...
+
+        if guard.channel_is_active(channel) {
+            panic!("TODO")
```
georgepisaltu, line 83: Not relevant to this PR, but for future efforts, you can use `todo!("need to implement X and Y")`.

muxink/src/demux.rs
```diff
@@ -0,0 +1,325 @@
...
+
+impl<S> Drop for DemultiplexerHandle<S> {
+    fn drop(&mut self) {
+        let mut demux = match self.demux.as_ref().try_lock() {
+            Err(_err) => {
+                return;
+            } // TODO What do? Perhaps try_lock is wrong here, but still what about poisoning? Not doing anything seems like the
+            // only sane option
+            Ok(guard) => guard,
+        };
```
georgepisaltu, line 113: This goes away once you separate locking from the rest of the logic. This would be a blanket `Drop` impl from the compiler for the `Arc<Mutex>>` part, and then the code below would be the `Drop` impl for the self-contained object, if that makes sense.

georgepisaltu (commented):
I left some suggestions here, let me know if they make sense.

muxink/src/demux.rs
marc-casperlabs: As discussed, you get to pick one - just make the rest of the crate fix that pattern :)

For this PR, it'd be nice to stick with what makes it match the rest of the crate and then do the refactoring afterwards.

muxink/src/demux.rs
marc-casperlabs: I agree that `1 << channel` is a little easier to read that `check_pow` :)

muxink/src/demux.rs
marc-casperlabs: I don't think the `pseudo_self: Arc<Mutex<Self>>` is that bad, to be honest. You'd almost get the same by just adding an alias `type SharedDemultiplexer<S> = Arc<Mutex<Demultiplexer<S>>>`, similar to your newtype-like suggestion.

> If we don't go with the wrapper, consider having create_handle as a standalone public function. 

That's essentially what this is (no `self`), except it's associated with the type!

> Users of that demux should choose to lock/unlock and otherwise manage their usage as necessary and are responsible for poisoned locks as well even as the code stands right now.

At the end of the day, this method needs a clone of the `Arc` to construct the handle. Passing it in while also having performed the locking seems more dangerous to me.

Overall I think this is between creating a wrapper struct (`Demultiplexer`, ensuring the inner one is renamed to `InnerDemultiplexer`) or not. Frankly, I could see this going either way.

muxink/src/demux.rs
marc-casperlabs: I disagree. If you hold a `DemultiplexerHandle`, you will need a reference to the "original" demultiplexer in some shape or form. Said reference must be protected through a lock.

The code here is responsible for deregistering the channel from said demultiplexer. If another thread has poisoned the mutex, we have an encountered error and there is nothing we can to except cross our fingers.

`try_lock`, however, *is* wrong here, as it just makes the locking non-blocking (coincidentally, I fell into that same trap a few months ago...). We *do* want to wait to deregister in normal operation.

georgepisaltu: I left this comment a while back, I don't remember exactly what I was thinking about back then but I think it would have been something like move the decrementing of the ref counter to the higher level object and not have it be in a `Drop` impl at all, have it done manually so you can decide what to do in case the lock is poisoned when you try to free the resource. This was all to avoid the locking in `drop`, but after reading your comment, I'm not sure it would add that much value so let's go with the original.

> try_lock, however, is wrong here, as it just makes the locking non-blocking (coincidentally, I fell into that same trap a few months ago...). We do want to wait to deregister in normal operation.

Yes, I missed it as well :sweat_smile:. I agree.

marc-casperlabs:
This has been merged and/or built-upon, in a different form, in the `feat-muxink` feature branch already.

PR #3191, opened 2022-07-09
Fix sync-to-genesis

### LINKED ISSUE #3186: Sync-to-genesis should sync as far as the highest block at the end of fast-syncin
sync-to-genesis can finish before "filling in" all the missing blocks and the related data, depending upon the trusted hash used if any.  We should sync forwards as far as the highest block at the end of the fast-sync process, as that block and all subsequent will be guaranteed to be complete, since that was the point at which consensus began execution.

Fraser999:
This PR changes sync-to-genesis to ensure it syncs forward far enough so that once it's complete the node has a full sequence of complete blocks and corresponding global state.

The main thrust of the change is that sync-from-genesis now uses the block returned by the fast-sync process as its trusted block.  Sync-to-genesis synchronises as far as that block, and all subsequent ones are guaranteed to be "complete" as they will have been executed while the node is in participating mode.

It also provides a fix for an issue discovered while testing the PR: namely that the `completed_blocks` in storage was not being adjusted after a hard reset to account for the fact that some blocks could have been removed.

Finally, it fixes an issue whereby `execute_blocks` was being passed the key block info for the trusted block rather than the most recent block.

Closes #3186, #3192 and #3193.

### DIFF
--- a/node/src/components/storage/disjoint_sequences.rs
+++ b/node/src/components/storage/disjoint_sequences.rs
@@ -65,12 +65,6 @@ impl Sequence {
         }
     }
 
-    /// Returns `true` if a sequence contains the value.
-    #[cfg(test)]
-    pub(super) fn contains(&self, value: u64) -> bool {
-        value >= self.low && value <= self.high
-    }
-
     /// Returns the inclusive high end of the sequence.
     pub(crate) fn high(&self) -> u64 {
         self.high
@@ -79,6 +73,12 @@ impl Sequence {
     /// Returns the inclusive low end of the sequence.
     pub(crate) fn low(&self) -> u64 {
         self.low
+    }
+
+    /// Returns `true` if a sequence contains the value.
+    #[cfg(test)]
+    pub(super) fn contains(&self, value: u64) -> bool {
+        value >= self.low && value <= self.high
     }
 }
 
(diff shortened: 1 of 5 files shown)


node/src/components/storage/disjoint_sequences.rs
```diff
@@ -437,6 +457,94 @@ mod tests {
...
+    #[test]
+    fn should_truncate() {
+        const SEQ_HIGH: Sequence = Sequence { high: 11, low: 9 };
+        const SEQ_MID: Sequence = Sequence { high: 6, low: 6 };
+        const SEQ_LOW: Sequence = Sequence { high: 3, low: 1 };
+        fn new_sequences() -> DisjointSequences {
+            DisjointSequences {
+                sequences: vec![SEQ_HIGH, SEQ_MID, SEQ_LOW],
+            }
+        }
```
goral09: Can't you initialize it once and then only clone it?

Fraser999: `DisjointSequences` isn't clonable.  I can add that behind a `test` feature-gate if you have a strong preference, but I'm not sure it's worth it?

goral09: It's not a blocker but it _could_ be clonable since it's just a couple tuples.

goral09 (approved):

node/src/components/storage/disjoint_sequences.rs
Fraser999: Fair enough - done in [c13fba6](https://github.com/casper-network/casper-node/pull/3191/commits/c13fba604943a73ec0642a72f712e3a433d75525).

node/src/components/storage/disjoint_sequences.rs
```diff
@@ -96,6 +96,7 @@ impl Sequence {
 /// For example, if `sequences` contains `[9,9], [7,3]` and `8` is inserted, then `sequences` will
 /// be reduced to `[9,3]`.
 #[derive(Default, Debug, DataSize)]
+#[cfg_attr(test, derive(Clone))]
```
goral09, line 99: It could have been just `Clone` without the test attribute.

fizyk20 (approved):

Fraser999:
bors merge

PR #3219, opened 2022-07-26
GH-3208: Configurable vesting schedule

### LINKED ISSUE #3208: Make vesting unlock schedule chainspec configurable
Currently it’s hardcoded here to 91 days total:
https://github.com/casper-network/casper-node/blob/c04fd3306404cb1cad4ead252315edc5ef770a8c/types/src/system/auction/bid/vesting.rs#L21-L28
We need to have this configurable in the chainspec. Gotcha: ensure existing genesis validator entries aren’t affected by possible changes to (de)serialization.

mpapierski:
Closes #3208

This PR introduces the `vesting_schedule_period` chain spec option with a default of `13 weeks` which was previously hardcoded:

```toml
[core]
# The period in which genesis validator's bid is released over time after it's unlocked.
vesting_schedule_period = '13 weeks'
```

This PR also ensures a `vesting_schedule_period = '0 days'` is a valid option, and additionally a combo of `vesting_schedule_period = '0 days'` and `locked_funds_period = '0 days'` allows an immediate unbond of a genesis validator.


execution_engine_testing/tests/src/test/regression/gh_3208.rs
```diff
@@ -0,0 +1,444 @@
...
+};
+use casper_types::{
+    runtime_args,
+    system::{
+        auction::{self, DelegationRate},
+        standard_payment,
+    },
+    ApiError, Key, Motes, RuntimeArgs, StoredValue, U512,
+};
+use once_cell::sync::Lazy;
```
darthsiroftardis: Import can be moved up I think

types/src/system/auction/bid.rs
```diff
@@ -129,7 +143,7 @@ impl Bid {
         &self.staked_amount
     }
 
-    /// Gets the staked amount of the provided bid
+    /// Gets the staked amount of the provided bidshould_release_vfta_holder_stake
```
darthsiroftardis:
```suggestion
    /// Gets the staked amount of the provided bid should_release_vfta_holder_stake
```

execution_engine/src/core/engine_state/engine_config.rs
```diff
@@ -12,6 +12,13 @@ pub const DEFAULT_MAX_RUNTIME_CALL_STACK_HEIGHT: u32 = 12;
 pub const DEFAULT_MINIMUM_DELEGATION_AMOUNT: u64 = 500 * 1_000_000_000;
 /// Default value for strict argument checking.
 pub const DEFAULT_STRICT_ARGUMENT_CHECKING: bool = false;
+/// 91 days / 7 days in a week = 13 weeks
+/// Length of total vesting schedule in days.
+const VESTING_SCHEDULE_LENGTH_DAYS: usize = 91;
+const DAY_MILLIS: usize = 24 * 60 * 60 * 1000;
+/// Default length of total vesting schedule period expressed in days.
+pub const DEFAULT_VESTING_SCHEDULE_LENGTH_MILLIS: u64 =
```
rafal-ch, line 20: Wouldn't `usize` suffice? I think it should, no?
That could be a way to avoid `as u64` below.

execution_engine/src/core/engine_state/engine_config.rs
```diff
@@ -46,12 +56,14 @@ impl Default for EngineConfig {
 
 impl EngineConfig {
     /// Creates a new engine configuration with provided parameters.
+    #[allow(clippy::too_many_arguments)]
```
rafal-ch, line 59: Looks like we slowly breed yet another "god" object :)
It's good for now, I think, but maybe we can pull it into our EE refactor plans.

types/src/system/auction/bid.rs
```diff
@@ -221,10 +235,23 @@ impl Bid {
...
     pub fn process(&mut self, timestamp_millis: u64) -> bool {
+        self.process_with_vesting_schedule(timestamp_millis, VESTING_SCHEDULE_LENGTH_MILLIS)
+    }
+
+    /// Initializes the vesting schedule of provided bid if the provided timestamp is greater than
+    /// or equal to the bid's initial release timestamp and the bid is owned by a genesis
+    /// validator.
+    ///
+    /// Returns `true` if the provided bid's vesting schedule was initialized.
+    pub fn process_with_vesting_schedule(
```
rafal-ch, line 250: There is some non-trivial processing of the timestamps and some early returns. Do you think it may be worth to add a unit test for this function?

types/src/system/auction/bid/vesting.rs
```diff
@@ -360,16 +442,74 @@ mod tests {
...
         assert!(vested_amounts_match_initial_stake(
             stake,
-            DEFAULT_LOCKED_FUNDS_PERIOD_MILLIS
+            DEFAULT_LOCKED_FUNDS_PERIOD_MILLIS,
+            DEFAULT_VESTING_SCHEDULE_PERIOD_MILLIS,
         ))
     }
 
+    #[test]
+    fn is_vesting_with_default_schedule() {
```
rafal-ch, line 449: This test takes nearly 2 minutes to execute, can we inject a smaller `DEFAULT_VESTING_SCHEDULE_PERIOD_MILLIS` or `LOCKED_AMOUNTS_LENGTH` for tests, possibly using the `[cfg(test)]` attribute?

resources/test/valid/0_9_0/chainspec.toml
```diff
@@ -14,6 +14,7 @@ minimum_era_height = 9
 validator_slots = 5
 auction_delay = 3
 locked_funds_period = '90days'
+vesting_schedule_period = '13 weeks'
```
rafal-ch, line 17: Would the algorithm work if we set this to, let's say `10ms`?
If not, it may be better to use: `vesting_schedule_period_weeks = 13`

execution_engine/src/core/engine_state/engine_config.rs
EdHastingsCasperAssociation: This is not a god object, it's just a DTO. What we see here is legitimately configured behavior / settings being passed in rather than hardcoded or assumed.

execution_engine_testing/test_support/src/lib.rs
```diff
@@ -41,12 +41,17 @@ pub use step_request_builder::StepRequestBuilder;
 pub use upgrade_request_builder::UpgradeRequestBuilder;
 pub use wasm_test_builder::{InMemoryWasmTestBuilder, LmdbWasmTestBuilder, WasmTestBuilder};
 
+const DAY_MILLIS: u64 = 24 * 60 * 60 * 1000;
```
EdHastingsCasperAssociation, line 44: +1

execution_engine_testing/test_support/src/lib.rs
```diff
@@ -41,12 +41,17 @@ pub use step_request_builder::StepRequestBuilder;
...
 /// Default number of validator slots.
 pub const DEFAULT_VALIDATOR_SLOTS: u32 = 5;
 /// Default auction delay.
 pub const DEFAULT_AUCTION_DELAY: u64 = 3;
 /// Default lock-in period of 90 days
-pub const DEFAULT_LOCKED_FUNDS_PERIOD_MILLIS: u64 = 90 * 24 * 60 * 60 * 1000;
+pub const DEFAULT_LOCKED_FUNDS_PERIOD_MILLIS: u64 = 90 * DAY_MILLIS;
+/// Default length of total vesting schedule of 91 days.
+pub const DEFAULT_VESTING_SCHEDULE_PERIOD_MILLIS: u64 = 91 * DAY_MILLIS;
+
```
EdHastingsCasperAssociation, line 54: maybe I'm overthinking this, but for purposes of the test builder which is a general purpose testing tool for everyone to use, I think the DEFAULT for lock-in and vesting should each be 0 days. 

our very specialized subset of tests proving that a post-genesis lock-in delay and a vesting schedule (which was a peculiarity of our mainnet launch which may or may not be relevant to other chains using the codebase) work correctly would then impute specific non-zero values in situ instead of using the defaults.

resources/test/valid/0_9_0/chainspec.toml
EdHastingsCasperAssociation: technically, the period could be any arbitrary non-negative unit of time

types/src/bytesrepr.rs
```diff
@@ -1230,6 +1230,19 @@ where
...
+{
+    fn to_bytes(&self) -> Result<Vec<u8>, Error> {
+        (*self).to_bytes()
+    }
+
+    fn serialized_length(&self) -> usize {
+        (*self).serialized_length()
+    }
+}
+
```
EdHastingsCasperAssociation: um...why not just de-reference from the callsite?

types/src/system/auction/bid.rs
EdHastingsCasperAssociation: /// Gets the staked amount of the provided bid

types/src/system/auction/bid.rs
```diff
@@ -237,15 +264,20 @@ impl Bid {
 
         let mut initialized = false;
 
-        if vesting_schedule.initialize(staked_amount) {
+        if vesting_schedule
+            .initialize_with_weekly_schedule(staked_amount, vesting_schedule_period_millis)
```
EdHastingsCasperAssociation: i don't think you should be renaming this to weekly_schedule; the mechanism should not make assumptions about the period

types/src/system/auction/bid/vesting.rs
```diff
@@ -45,36 +59,56 @@ impl VestingSchedule {
         }
     }
 
+    /// Initializes vesting schedule with a configured amount of weekly releases.
+    ///
     /// Returns `false` if already initialized.
-    pub fn initialize(&mut self, staked_amount: U512) -> bool {
+    pub fn initialize_with_weekly_schedule(
```
EdHastingsCasperAssociation: initialize_with_schedule

types/src/system/auction/bid/vesting.rs
EdHastingsCasperAssociation: agreed; a smaller non-zero lock in and vesting period is sufficient to prove function without undue time cost

EdHastingsCasperAssociation (approved):
optimistic assuming nitpicks addressed

types/src/system/auction/bid.rs
mpapierski: There's rather a long unit test exercising different timestamps named `should_initialize_delegators_different_timestamps` - I modified it to use a non-default vesting schedule and added a separate test for immediate unlock.

types/src/system/auction/bid/vesting.rs
mpapierski: There was an off-by-one error causing a run time to blow up - my machine after fixing it:

```
running 1 test
test system::auction::bid::vesting::tests::is_vesting_with_default_schedule ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 452 filtered out; finished in 0.00s
```

resources/test/valid/0_9_0/chainspec.toml
mpapierski: Correct; technically, we would round it down to full weeks because of how the unlock amounts are stored.

types/src/bytesrepr.rs
mpapierski: Removed; unnecessary in this PR although this makes life much easier as you don't need to clone or dereference T.

execution_engine_testing/test_support/src/lib.rs
Fraser999: Ticketed as [#3231](https://github.com/casper-network/casper-node/issues/3231).

rafal-ch (commented):
LGTM :+1:

Issue #3198, opened 2022-07-11
Add a `bytesrepr` serialization module to `muxink`


devendran-m:
Similar to the `bincode` `Transcoder` and potentially `FrameDecoder` implementation we need a similar one for `bytesrepr`.

Care should be taken to leverage the `bytes::Bytes` compatible methods of `ToBytes` where possible to reduce the amount of memory copied.

PR #3220, opened 2022-07-27
Add `bytesrepr` codec to `muxink`

rafal-ch:
This PR adds the `bytesrepr` codec to `muxink` crate.
It also adds unit tests to the `bincode` codec and fixes some minor issues in it.

Closes https://github.com/casper-network/casper-node/issues/3198

### DIFF
--- a/muxink/Cargo.toml
+++ b/muxink/Cargo.toml
@@ -18,5 +18,5 @@ casper-types = { path = "../types", optional = true }
 tokio-stream = "0.1.8"
 
 [features]
-bincode = [ "dep:serde", "dep:bincode" ]
-bytesrepr = [ "dep:casper-types" ] 
+muxink_bincode_codec = [ "serde", "bincode" ]
+muxink_bytesrepr_codec = [ "casper-types" ] 
--- a/muxink/Cargo.toml
+++ b/muxink/Cargo.toml
@@ -12,9 +12,11 @@ serde = { version = "1.0.138", optional = true }
 thiserror = "1.0.31"
 tokio = { version = "1.18.1", features = [ "full" ] }
 tokio-util = "0.7.2"
+casper-types = { path = "../types", optional = true }
 
 [dev-dependencies]
 tokio-stream = "0.1.8"
 
 [features]
 bincode = [ "dep:serde", "dep:bincode" ]
+bytesrepr = [ "dep:casper-types" ] 
--- a/muxink/src/backpressured.rs
+++ b/muxink/src/backpressured.rs
@@ -80,7 +80,9 @@ impl<S, A, Item> BackpressuredSink<S, A, Item> {
 
 impl<Item, A, S> Sink<Item> for BackpressuredSink<S, A, Item>
 where
-    // TODO: `Unpin` trait bounds can be removed by using `map_unchecked` if necessary.
+    // TODO: `Unpin` trait bounds can be
+    // removed by using `map_unchecked` if
+    // necessary.
     S: Sink<Item> + Unpin,
     Self: Unpin,
     A: Stream<Item = u64> + Unpin,
--- a/muxink/src/codec.rs
+++ b/muxink/src/codec.rs
@@ -24,6 +24,8 @@
 
 #[cfg(feature = "bincode")]
 pub mod bincode;
+#[cfg(feature = "bytesrepr")]
+pub mod bytesrepr;
 pub mod length_delimited;
 
 use std::{
--- a/muxink/src/codec.rs
+++ b/muxink/src/codec.rs
@@ -22,9 +22,9 @@
 //! * [`length_delimited`]: Transforms byte-like values into self-contained frames with a
 //!   length-prefix.
 
-#[cfg(feature = "bincode")]
+#[cfg(feature = "muxink_bincode_codec")]
 pub mod bincode;
-#[cfg(feature = "bytesrepr")]
+#[cfg(feature = "muxink_bytesrepr_codec")]
 pub mod bytesrepr;
 pub mod length_delimited;
 
--- a/muxink/src/codec/bincode.rs
+++ b/muxink/src/codec/bincode.rs
@@ -17,10 +17,10 @@ use serde::{de::DeserializeOwned, Serialize};
 
 use super::{DecodeResult, FrameDecoder, Transcoder};
 
-/// A bincode encoder.
+/// Bincode encoder.
 ///
 /// Every value is encoded with the default settings of `bincode`.
-#[derive(Default)]
+#[derive(Debug, Default)]
 pub struct BincodeEncoder<T> {
     /// Item type processed by this encoder.
     ///
@@ -57,7 +57,7 @@ where
 /// Like [`BincodeEncoder`], uses default settings for decoding. Can be used on bytestreams (via
 /// [`FrameDecoder`]) as well as frames (through [`Transcoder`]). See module documentation for
 /// caveats.
-#[derive(Default)]
+#[derive(Debug, Default)]
 pub struct BincodeDecoder<T> {
     item_type: PhantomData<T>,
 }
--- a/muxink/src/codec/bincode.rs
+++ b/muxink/src/codec/bincode.rs
@@ -46,8 +46,7 @@ where
     type Output = Bytes;
 
     fn transcode(&mut self, input: T) -> Result<Self::Output, Self::Error> {
-        DefaultOptions::new()
-            .reject_trailing_bytes()
+        bincode_transcode_options()
             .serialize(&input)
             .map(Bytes::from)
     }
@@ -82,9 +81,7 @@ where
     type Output = T;
 
     fn transcode(&mut self, input: R) -> Result<Self::Output, Self::Error> {
-        DefaultOptions::new()
-            .reject_trailing_bytes()
-            .deserialize(input.as_ref())
+        bincode_transcode_options().deserialize(input.as_ref())
     }
 }
 
@@ -131,6 +128,12 @@ where
             },
         }
     }
+}
+
+fn bincode_transcode_options() -> impl bincode::config::Options {
+    DefaultOptions::new()
+        .reject_trailing_bytes()
+        .with_varint_encoding()
 }
 
 #[cfg(test)]
--- a/muxink/src/codec/bincode.rs
+++ b/muxink/src/codec/bincode.rs
@@ -97,7 +97,7 @@ where
             let slice: &[u8] = buffer.as_ref();
             let mut cursor = Cursor::new(slice);
             let outcome = DefaultOptions::new()
-                .with_fixint_encoding()
+                .with_varint_encoding()
                 .allow_trailing_bytes()
                 .deserialize_from(&mut cursor);
             (outcome, cursor.position() as usize)
@@ -160,7 +160,7 @@ mod tests {
 
     #[test]
     fn decodes_frame() {
-        let data = b"\x03\0\0\0\0\0\0\0abc\x04\0\0\0\0\0\0\0defg";
+        let data = b"\x03abc\x04defg";
 
         let mut bytes: BytesMut = BytesMut::new();
         bytes.extend(data);
@@ -169,6 +169,20 @@ mod tests {
 
         assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == "abc"));
         assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == "defg"));
+    }
+
+    #[test]
+    fn decodes_frame_of_raw_integers() {
+        // 40000u16 followed by 7u16
+        let data = b"\xfb\x40\x9c\x07";
+
+        let mut bytes: BytesMut = BytesMut::new();
+        bytes.extend(data);
+
+        let mut decoder = BincodeDecoder::<u16>::new();
+
+        assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == 40000));
+        assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == 7));
     }
 
     #[test]
@@ -181,7 +195,7 @@ mod tests {
 
     #[test]
     fn error_when_buffer_not_exhausted() {
-        let data = b"\x03\0\0\0\0\0\0\0abc\x04\0\0\0\0\0\0\0defg";
+        let data = b"\x03abc\x04defg";
 
         let mut decoder = BincodeDecoder::<String>::new();
         let actual_error = *decoder.transcode(data).unwrap_err();
@@ -193,7 +207,7 @@ mod tests {
 
     #[test]
     fn error_when_data_incomplete() {
-        let data = b"\x03\0\0\0\0\0\0\0ab";
+        let data = b"\x03ab";
 
         let mut bytes: BytesMut = BytesMut::new();
         bytes.extend(data);
--- a/muxink/src/codec/bincode.rs
+++ b/muxink/src/codec/bincode.rs
@@ -119,7 +119,15 @@ where
                 bincode::ErrorKind::Io(io_err) if io_err.kind() == io::ErrorKind::UnexpectedEof => {
                     DecodeResult::Incomplete
                 }
-                _ => DecodeResult::Failed(err),
+                bincode::ErrorKind::SizeLimit
+                | bincode::ErrorKind::SequenceMustHaveLength
+                | bincode::ErrorKind::Custom(_)
+                | bincode::ErrorKind::InvalidCharEncoding
+                | bincode::ErrorKind::InvalidTagEncoding(_)
+                | bincode::ErrorKind::DeserializeAnyNotSupported
+                | bincode::ErrorKind::Io(_)
+                | bincode::ErrorKind::InvalidUtf8Encoding(_)
+                | bincode::ErrorKind::InvalidBoolEncoding(_) => DecodeResult::Failed(err),
             },
         }
     }
--- a/muxink/src/codec/bincode.rs
+++ b/muxink/src/codec/bincode.rs
@@ -11,6 +11,7 @@ use std::{
     marker::PhantomData,
 };
 
+use bincode::{DefaultOptions, Options};
 use bytes::{Buf, Bytes, BytesMut};
 use serde::{de::DeserializeOwned, Serialize};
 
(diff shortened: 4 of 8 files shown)


muxink/Cargo.toml
```diff
@@ -12,9 +12,11 @@ serde = { version = "1.0.138", optional = true }
...
 tokio-util = "0.7.2"
+casper-types = { path = "../types", optional = true }
 
 [dev-dependencies]
 tokio-stream = "0.1.8"
 
 [features]
-bincode = [ "dep:serde", "dep:bincode" ]
+muxink_bincode_codec = [ "serde", "bincode" ]
+muxink_bytesrepr_codec = [ "casper-types" ]
```
goral09, line 22: These two features should be excluding one another, right? I.e. it doesn't make sense to have both bincode and bytesrepr at the same time?

muxink/src/codec/bincode.rs
```diff
@@ -110,8 +119,87 @@ where
...
+        bytes.extend(data);
+
+        let mut decoder = BincodeDecoder::<String>::new();
+
+        assert!(matches!(
+            decoder.decode_frame(&mut bytes),
+            DecodeResult::Incomplete
+        ));
+    }
+}
```
goral09, line 222: These tests seem to be generic enough to apply for `bytesrepr` case as well?

muxink/Cargo.toml
marc-casperlabs: It's conceivable to actually use both, they're just encodings. We may even start with `bincode` only, then transfer ACK streams to `ToBytes` first, keeping others using `bincode` and finally do the switch.

muxink/src/backpressured.rs
```diff
@@ -80,7 +80,9 @@ impl<S, A, Item> BackpressuredSink<S, A, Item> {
 
 impl<Item, A, S> Sink<Item> for BackpressuredSink<S, A, Item>
 where
-    // TODO: `Unpin` trait bounds can be removed by using `map_unchecked` if necessary.
+    // TODO: `Unpin` trait bounds can be
```
marc-casperlabs, line 83: Is this nightly `rustfmt` running amok? :(

muxink/Cargo.toml
```diff
@@ -12,9 +12,11 @@ serde = { version = "1.0.138", optional = true }
...
 tokio = { version = "1.18.1", features = [ "full" ] }
 tokio-util = "0.7.2"
+casper-types = { path = "../types", optional = true }
 
 [dev-dependencies]
 tokio-stream = "0.1.8"
 
 [features]
-bincode = [ "dep:serde", "dep:bincode" ]
+muxink_bincode_codec = [ "serde", "bincode" ]
```
marc-casperlabs, line 21: Correct me if I am wrong, but would it not be possible to simply axe the `bincode` features, just marking `bincode` as `optional`? That'll define the correct feature named `bincode`.

Then rename the `muxink_bytesrepr_codec` to `bytesrepr`, keeping the `casper-types` dependency.

muxink/src/codec.rs
```diff
@@ -22,8 +22,10 @@
 //! * [`length_delimited`]: Transforms byte-like values into self-contained frames with a
 //!   length-prefix.
 
-#[cfg(feature = "bincode")]
+#[cfg(feature = "muxink_bincode_codec")]
```
marc-casperlabs, line 25: The features can be renamed, I maintain they should be named `bincode` and `bytesrepr`, see above :)

muxink/src/codec/bincode.rs
```diff
@@ -45,7 +46,10 @@ where
     type Output = Bytes;
 
     fn transcode(&mut self, input: T) -> Result<Self::Output, Self::Error> {
-        bincode::serialize(&input).map(Bytes::from)
+        DefaultOptions::new()
```
marc-casperlabs: On second thought, I think it'd be good to simply hardcore all available options to save us grief later on, should they change.

muxink/src/codec/bincode.rs
```diff
@@ -78,7 +82,9 @@ where
     type Output = T;
 
     fn transcode(&mut self, input: R) -> Result<Self::Output, Self::Error> {
-        bincode::deserialize(input.as_ref())
+        DefaultOptions::new()
```
marc-casperlabs: I'd factor out the `DefaultOptions` into 

```rust
fn mk_bincode() -> impl bincode::config::Options {
    DefaultOptions::new().reject_trailing_bytes()
}
```

as this prevents encoding and decoding options from accidentally drifting apart.

muxink/src/codec/bincode.rs
```diff
@@ -93,7 +99,10 @@ where
         let (outcome, consumed) = {
             let slice: &[u8] = buffer.as_ref();
             let mut cursor = Cursor::new(slice);
-            let outcome = bincode::deserialize_from(&mut cursor);
+            let outcome = DefaultOptions::new()
```
marc-casperlabs, line 99: We could standardize on one set of `bincode` encoding options here even. Jury's still out on whether fixint or varint, I hope the latter will work.

muxink/src/codec/bincode.rs
marc-casperlabs: Hmm, there may be something to it. We can generalize most of these tests over any transcoder `T`, although this makes it quite a bit more complex:

```rust
fn test_transcoder_roundtrip_string<T, E, D>(encoder: E, decoder: D)
    where
        E: Transcoder<String>,
        D: Transcoder<Bytes, Output=String>
 {
    let data = "hello, world";
    let encoded = encoder.transcode(data.to_owned()).expect("encoding failed");
    let decoded = decoder.transcode(encoded).expect("decoding failed");
    assert_eq!(decoded.as_str(), data);
}

fn test_transcoder_error_when_buffer_not_exhausted(encoder: E, decoder: D) 
    where
        E: Transcoder<String>,
        D: Transcoder<Bytes, Output=String>
{
        let mut data_vec: Vec<u8> = encoder.transcode(data.to_owned()).expect("encoding failed").to_vec();
        data_vec.extend(b"extra data");
        let data = Bytes::from(data_vec);

        decoder.transcode(data).expect_err("deserialization did not return an err when trailing data was present");

        // Note: We lost the validation of the returned error kind here. We can
        //        pass in a closure to verify, but this kind of thing gets ugly fast....
}
// ...

// in bincode::tests
#[test]
fn roundtrip() {
    test_transcoder_roundtrip(BincodeEncoder::new(), BincodeDecoder::new()); 
}

#[test]
fn error_when_buffer_not_exhausted() {
     test_transcoder_error_when_buffer_not_exhausted(BincodeEncoder::new(), BincodeDecoder::new()); 
}
```

An alternative would be to either have the test-suite in a macro (not pretty, but saves a lot of where clauses), or a have specific type with a trait that knows how to instantiate the necessary encoders/decoders.

```rust
trait TranscoderTestSuite {
    type StringEncoder : impl Transcoder<String>;
    fn create_string_encoder() -> Self::StringEncoder;
    // ...
}

struct BincodeTestSuite;

impl TranscoderTestSuite for BincodeTestSuite {
    type StringEncoder = BincodeEncoder<String>;
    fn create_string_encoder() -> Self::StringEncoder {
        BincodeEncoder::new()
    }
}

fn run_test_suite<S: TranscoderTestSuite>() {
    test_transcoder_roundtrip_string(S::create_string_encoder(), S::create_string_decoder());
}

```
We can shorten it a bit now:

```rust

// in `bincode::tests` ...
 
#[test]
fn run_test_suide() {
    run_test_suite<BincodeTestSuite>::();
}
```

This is quite a bit more complex than the simple code above though.

muxink/src/codec/bincode.rs
```diff
@@ -110,8 +119,87 @@ where
...
+
+        let mut bytes: BytesMut = BytesMut::new();
+        bytes.extend(data);
+
+        let mut decoder = BincodeDecoder::<String>::new();
+
+        assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == "abc"));
+        assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == "defg"));
+    }
+
```
marc-casperlabs, line 173: It would be nice to add a test that reads multiple integers from a stream that has had integers directly written to it. This would be useful when/if we switch to varint encoding.

muxink/src/codec/bytesrepr.rs
```diff
@@ -0,0 +1,172 @@
+//! Bytesrepr encoding/decoding
```
marc-casperlabs, line 1: Can we add a little more info on this module, potential pitfalls and such, if there are any?

... and a blank line? :)

muxink/src/codec/bytesrepr.rs
```diff
@@ -0,0 +1,172 @@
...
+
+use bytes::{Buf, Bytes, BytesMut};
+use casper_types::bytesrepr::{self, FromBytes, ToBytes};
+use thiserror::Error;
+
+use super::{DecodeResult, FrameDecoder, Transcoder};
+use crate::codec::DecodeResult::Failed;
+
+#[derive(Debug, Error)]
+pub enum Error {
```
marc-casperlabs: I don't think we need to reflexively use an error `enum` every time.

Does `bytesrepr::Error` implement all the necessary traits? If so, we could just use it directly without an intermediate error type. If not, we should at least make it a newtype `struct`.

Compare the `bincode` module.

muxink/src/codec/bytesrepr.rs
```diff
@@ -0,0 +1,172 @@
...
+
+    type Output = Bytes;
+
+    fn transcode(&mut self, input: T) -> Result<Self::Output, Self::Error> {
+        Ok(input.to_bytes().map_err(Error::BytesreprError)?.into())
+    }
+}
+
+/// Bytesrepr decoder.
+#[derive(Default)]
```
marc-casperlabs: Maybe add `Debug`?

muxink/src/codec/bytesrepr.rs
```diff
@@ -0,0 +1,172 @@
...
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("bytesrepr error")]
+    BytesreprError(bytesrepr::Error),
+}
+
+/// A bytesrepr encoder.
+#[derive(Default)]
+pub struct BytesreprEncoder<T> {
```
marc-casperlabs, line 20: `Debug` here as well?

muxink/src/codec/bytesrepr.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        let mut bytes: BytesMut = BytesMut::new();
+        bytes.extend(data);
+
+        let mut decoder = BytesreprDecoder::<String>::new();
+
+        assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == "abc"));
+        assert!(matches!(decoder.decode_frame(&mut bytes), DecodeResult::Item(i) if i == "defg"));
+    }
+
```
marc-casperlabs, line 139: See comments on `bincode` tests about generalizing this.

Also, this also should have the mentioned test added that is a "naked stream" of integers.

muxink/src/fragmented.rs
```diff
@@ -1,8 +1,8 @@
 //! Splits frames into fragments.
 //!
 //! The wire format for fragments is `NCCC...` where `CCC...` is the data fragment and `N` is the
-//! continuation byte, which is `0x00` if more fragments are following, `0xFF` if this is the frame's
-//! last fragment.
+//! continuation byte, which is `0x00` if more fragments are following, `0xFF` if this is the
+//! frame's last fragment.
```
marc-casperlabs, line 5: Does the formatter complain if this is not altered?

muxink/src/lib.rs
```diff
@@ -220,8 +222,10 @@ pub(crate) mod tests {
 
     // #[test]
     // fn from_bytestream_to_multiple_frames() {
-    //     let input = &b"\x06\x00\x00ABCDE\x06\x00\x00FGHIJ\x03\x00\xffKL\x10\x00\xffSINGLE_FRAGMENT\x02\x00\x00C\x02\x00\x00R\x02\x00\x00U\x02\x00\x00M\x02\x00\x00B\x02\x00\xffS"[..];
-    //     let expected: &[&[u8]] = &[b"ABCDEFGHIJKL", b"SINGLE_FRAGMENT", b"CRUMBS"];
+    //     let input =
```
marc-casperlabs: I don't think we need this change here, it'll just produce merge conflicts down the road.

marc-casperlabs (commented):
Some feedback and questions still, but I like where it is heading.

muxink/src/codec/bytesrepr.rs
```diff
@@ -0,0 +1,172 @@
...
+impl<T> Transcoder<T> for BytesreprEncoder<T>
+where
+    T: ToBytes,
+{
+    type Error = Error;
+
+    type Output = Bytes;
+
+    fn transcode(&mut self, input: T) -> Result<Self::Output, Self::Error> {
+        Ok(input.to_bytes().map_err(Error::BytesreprError)?.into())
```
marc-casperlabs: I think this should use `into_bytes` for efficiency.

muxink/src/codec/bincode.rs
goral09: I would spend some time whether there exists an encoding where these tests are easier to write then. Although, they don't look bad right now either.

muxink/src/backpressured.rs
rafal-ch: Yup.

muxink/src/codec/bincode.rs
rafal-ch: I'm all in, let's have a ticket for this to not clog this PR. I already stretched it a bit by adding tests to the `bincode` codec.

muxink/Cargo.toml
rafal-ch: Wouldn't cargo complain that optional `bincode` is not a part of any feature then?

muxink/src/codec/bincode.rs
rafal-ch: I made the function name more explicit, because we actually need a different options for the `decode_frame`.

Surprisingly, the `allow_trailing_bytes()` in `decode_frame` seems redundant, it works correctly even with the `reject_trailing_bytes()` set (in case of which I'd expect the `codec::bincode::tests::decodes_frame` to fail since there **are** more bytes in the buffer.). I left the `allow_trailing_bytes()` intact for clarity of the intent (similarly to the redundant `with_varint_encoding()`).

Updated in 49a507c1d4cdcfc759fdffda2ab6f66d8147067d

muxink/src/codec/bytesrepr.rs
rafal-ch: Here you are: 15a8bdef50fd39e62e3ca905437318e161f3689a

muxink/src/codec/bytesrepr.rs
rafal-ch: Here you are as well :)
15a8bdef50fd39e62e3ca905437318e161f3689a

muxink/src/fragmented.rs
rafal-ch: The formatter itself did the change.

muxink/src/lib.rs
rafal-ch: I reverted the change and added the `#[rustfmt::skip]` which should be removed after the tests are re-enable.

Updated in: a8056e4faf8b3846b6f9d3a2ba60b41e8682f9a9

muxink/src/codec/bytesrepr.rs
rafal-ch: Updated in 44b1fa3b0b4c2f02fb801923920fae3d4c805ad7

muxink/src/codec/bytesrepr.rs
rafal-ch: `bytesrepr::Error` doesn't implement `std::error::Error` which is required by `Transcoder::Error`

muxink/src/codec/bincode.rs
rafal-ch: Test added 04b601a8b98e343e30591a0a0ce258d583550c84 along with the change to varint encoding everywhere.

muxink/src/codec/bytesrepr.rs
rafal-ch: Regarding generalization, see [this comment](https://github.com/casper-network/casper-node/pull/3220#discussion_r932975507).

And the mentioned test is added in 04b601a8b98e343e30591a0a0ce258d583550c84

muxink/src/codec/bincode.rs
rafal-ch: > We can generalize most of these tests over any transcoder T, although this makes it quite a bit more complex

Should be much simpler with macro.

muxink/src/codec/bytesrepr.rs
rafal-ch: Comments updated in 0448111c27126081d80f60cff3b8a7bb9b456483. The comment is rather dull (that's why I decided to not include it in the first place).

Blank line has been added as well.

muxink/src/codec/bincode.rs
rafal-ch: Currently, the varint is used.

muxink/Cargo.toml
marc-casperlabs: Got me, I genuinely don't know. Will have to try that later.

muxink/src/codec/bincode.rs
marc-casperlabs: Superb, thanks!

muxink/src/backpressured.rs
marc-casperlabs: :cry:

muxink/src/codec/bytesrepr.rs
marc-casperlabs: That makes sense then. Still, how about `struct Error(bytesrepr::Error)` instead?

muxink/src/codec/bincode.rs
marc-casperlabs: Sounds good to me. If you add a ticket, I'll vouch for that :)

muxink/Cargo.toml
rafal-ch: I believe I tried such an approach already, but no dice.

muxink/src/codec/bytesrepr.rs
rafal-ch: Sure, changed here: e22f7be6b7200a7a083999dd68db5462a842c9ec

muxink/Cargo.toml
marc-casperlabs: :cry:

marc-casperlabs (approved):
Everything addressed, looks good to me!

Issue #3209, opened 2022-07-13
Ensure unbonding_delay is greater or equal to auction_delay


cspramit:
We should make sure this is explicitly validated in the genesis

PR #3229, opened 2022-08-02
GH-3209: Validate unbonding delay

mpapierski:
Closing #3209

Validates chainspec's `unbonding_delay` to be greater or equal to `auction_delay`

### DIFF
--- a/node/src/types/chainspec.rs
+++ b/node/src/types/chainspec.rs
@@ -82,9 +82,9 @@ impl Chainspec {
 
         let min_era_ms = 1u64 << self.highway_config.minimum_round_exponent;
 
-        if self.core_config.unbonding_delay < self.core_config.auction_delay {
+        if self.core_config.unbonding_delay <= self.core_config.auction_delay {
             warn!(
-                "unbonding delay is set to {} but it should be greater or equal to the auction delay (currently set to {})",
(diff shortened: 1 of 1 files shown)


node/src/types/chainspec.rs
```diff
@@ -81,6 +81,14 @@ impl Chainspec {
         }
 
         let min_era_ms = 1u64 << self.highway_config.minimum_round_exponent;
+
+        if self.core_config.unbonding_delay < self.core_config.auction_delay {
```
afck: If they're equal and I request unbonding in era `N`, I might still be a validator in era `N + auction_delay`, and I'd get my money back right at the end of that era, right?
That still seems a bit risky to me; maybe we should print a warning in that case, as well, or even disallow it?

afck (approved):
(Not a blocker, but I'd prefer being even more strict.)

node/src/types/chainspec.rs
Fraser999: I'd favour disallowing it.  Not sure if there's any advantage to allowing them to be equal?

Fraser999 (approved):

Fraser999:
bors merge

Fraser999:
bors merge

PR #3250, opened 2022-08-09
CI: package up global-state-update-gen

TomVasile:
### Adds:
- `.github/workflows/publish-global-state-update-gen.yml`
    - gha workflow that will publish the `global-state-update-gen tool` on any `v*` tag to our `apt` repos
    - built on both `focal` and `bionic` runners
- `ci/publish_deb_to_repo.sh`
    - handles the juggling of the `aptly` commands
- `utils/global-state-update-gen/Cargo.toml`
    - adds fields use by `cargo-deb`
    
### Ticket:
- https://app.zenhub.com/workspaces/cn-sre-60d363546afb85000e491ff6/issues/casper-network/sre/541

### DIFF
--- a/ci/publish_deb_to_repo.sh
+++ b/ci/publish_deb_to_repo.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+set -e
+
+# DEFAULTS
+PLUGIN_OS_CODENAME="${PLUGIN_OS_CODENAME:-bionic}"
+
+# Verify all variables are present
+if [[ -z $PLUGIN_GPG_KEY || -z $PLUGIN_GPG_PASS || -z $PLUGIN_REGION \
+        || -z $PLUGIN_REPO_NAME || -z $PLUGIN_ACL || -z $PLUGIN_PREFIX \
+        || -z $AWS_SECRET_ACCESS_KEY || -z $AWS_ACCESS_KEY_ID \
+        || -z $PLUGIN_DEB_PATH || -z $PLUGIN_OS_CODENAME ]]; then
+    echo "ERROR: Environment Variable Missing!"
+    exit 1
+fi
+
+# Verify if its the first time publishing. Will need to know later.
+# Probably an easier way to do this check :)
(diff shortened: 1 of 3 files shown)


ci/publish_deb_to_repo.sh
```diff
@@ -0,0 +1,62 @@
...
+# Add .debs to the local repo
+aptly repo add -force-replace "release-$PLUGIN_OS_CODENAME" "$PLUGIN_DEB_PATH"/*.deb
+
+# Publish to S3
+if [ ! "$(aptly publish list | grep $PLUGIN_REPO_NAME | grep $PLUGIN_OS_CODENAME)" ]; then
+    # If the repo is new
+    aptly publish repo -batch -force-overwrite -passphrase="$PLUGIN_GPG_PASS" "release-$PLUGIN_OS_CODENAME" s3:"${PLUGIN_REPO_NAME}":
+else
+    # If the repo exists
+    aptly publish update -batch -force-overwrite -passphrase="$PLUGIN_GPG_PASS" "$PLUGIN_OS_CODENAME" s3:"${PLUGIN_REPO_NAME}":
```
Fraser999, line 61: This doesn't need `"release-$PLUGIN_OS_CODENAME"` rather than `"$PLUGIN_OS_CODENAME"`?

TomVasile: I can see why you might think that. The if.. else is using two different subcommands of `aptly publish`.

For the if clause, `aptly publish repo <name>` where name is the local repo name. 
the else is using `aptly publish update <distribution>` hence only `$PLUGIN_OS_CODENAME`.

This isn't a new script btw, it's copied from the client repo that we've been using for a bit now. Planning on centralizing it shortly because I'm also about to use it for the launcher repo. Was just faster to copy it in for now.

sacherjj (approved):

ci/publish_deb_to_repo.sh
Fraser999: Understood - thanks!

Fraser999 (approved):

fizyk20 (approved):
Looks good as far as I could tell, but note that I'm not familiar at all with what is required for packaging and publishing binaries.

sacherjj (approved):

TomVasile:
Manually merging, tested locally with no issue. Reason: https://forum.bors.tech/t/resource-not-accessible-by-integration/408/12

PR #3254, opened 2022-08-10
Ensure synced deploys of a block are held in the correct order

### LINKED ISSUE #3253: Node failed to join to nctl network while new deploys were being sent
see failure in https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/6072/1/4

Fraser999:
This PR ensures that during chain sync, deploys which are fetched in parallel are inserted into the block for execution in the correct order.

It also updates the `Debug` impl for the mirrored `DeployHash` type in casper-types to use hex-encoded notation as per other hash digests.

Finally, it downgrades an `error!` log in the RPC server to `warn!` as it could be triggered by an invalid client request and doesn't represent an actual internal node error.

Closes #3253.

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -36,12 +36,15 @@ use crate::{
         fetcher::{FetchResult, FetchedData, FetcherError},
     },
     effect::{
-        announcements::{BlocklistAnnouncement, ChainSynchronizerAnnouncement},
+        announcements::{
+            BlocklistAnnouncement, ChainSynchronizerAnnouncement, ControlAnnouncement,
+        },
         requests::{
             ContractRuntimeRequest, FetcherRequest, MarkBlockCompletedRequest, NetworkInfoRequest,
         },
         EffectBuilder,
     },
+    fatal,
     storage::StorageRequest,
     types::{
         AvailableBlockRange, Block, BlockAndDeploys, BlockHash, BlockHeader,
@@ -1876,6 +1879,7 @@ where
         + From<FetcherRequest<TrieOrChunk>>
         + From<BlocklistAnnouncement>
         + From<MarkBlockCompletedRequest>
+        + From<ControlAnnouncement>
         + Send,
 {
     info!("fast syncing chain");
@@ -2098,6 +2102,7 @@ where
         + From<BlocklistAnnouncement>
         + From<StorageRequest>
         + From<MarkBlockCompletedRequest>
+        + From<ControlAnnouncement>
         + Send,
 {
     let _metric = ScopeTimer::new(&ctx.metrics.chain_sync_execute_blocks_duration_seconds);
@@ -2169,11 +2174,13 @@ where
         while !blocks_match {
             // Could be wrong approvals - fetch new sets of approvals from a single peer and retry.
             for peer in get_filtered_fully_connected_peers(ctx).await {
+                attempts += 1;
                 warn!(
-                    block_hash=%block.hash(),
+                    fetched_block=%block,
+                    executed_block=%block_and_execution_effects.block(),
+                    attempts,
                     "retrying execution due to deploy approvals mismatch"
                 );
-                attempts += 1;
                 ctx.progress.retry_executing_block(block.height(), attempts);
                 let block_and_execution_effects = retry_execution_with_approvals_from_peer(
                     &mut deploys,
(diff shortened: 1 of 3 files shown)


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -2240,21 +2247,37 @@ async fn fetch_and_store_deploys<REv>(
...
+        + From<NetworkInfoRequest>
+        + From<ControlAnnouncement>,
 {
     let start_instant = Timestamp::now();
 
-    let hashes: Vec<_> = hashes.cloned().collect();
-    let mut deploys: Vec<Deploy> = Vec::with_capacity(hashes.len());
-    let mut stream = futures::stream::iter(hashes)
-        .map(|hash| fetch_and_store_deploy(hash, ctx))
+    let mut ordered_deploys = BTreeMap::new();
```
goral09: We could use the `mut deploys` from before and insert at an index as well, right? Not sure if a `BTreeMap` isn't a small overkill for this.

goral09 (approved):

node/src/components/chain_synchronizer/operations.rs
Fraser999: Not sure it makes _too_ much of a difference.  Using `Vec::insert` would require actually constructing a Vec with the right number of entries, which would then all get overwritten (and `insert` can panic, making it one more thing to consider during safety reviews of the code).

If we see performance problems, we could indeed switch to a `Vec` and maybe leverage some unsafe methods, but I'd benchmark before doing something like that.

marc-casperlabs: Is there anything wrong with `Vec::sort` (but see below for a different approach).

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -2240,21 +2247,37 @@ async fn fetch_and_store_deploys<REv>(
...
+
+    let mut stream = futures::stream::iter(hashes.cloned().enumerate())
+        .map(|(index, hash)| async move { (index, fetch_and_store_deploy(hash, ctx).await) })
         .buffer_unordered(ctx.config.max_parallel_deploy_fetches());
-    while let Some(result) = stream.next().await {
+    while let Some((index, result)) = stream.next().await {
         let deploy = result?;
         trace!("fetched {:?}", deploy);
-        deploys.push(*deploy);
+        if let Some(deploy) = ordered_deploys.insert(index, *deploy) {
```
marc-casperlabs: I am not a fan of this. It's a lot of code that is essentially checking that `enumerate` works correctly, right?

marc-casperlabs: Can we not replace all of this code, starting from `stream`, with something like this? (I'm assuming `hashes` contains what we need in order):

```rust
let jobs: FuturesOrdered<_> = hashes.into_iter().map(|hash| fetch_and_store_deploy(hash, ctx));
let ordered_deploys: Vec<_> = jobs.collect().await;
```

marc-casperlabs (commented):
I _think_ we can simplify this?

node/src/components/chain_synchronizer/operations.rs
goral09: Ordering thousands of deploys? I would be worried about performance of `PartialEq, Eq` for `Deploy`.

marc-casperlabs: Hmm, I would not be concerned about that at all:

1. Ordering a few thousand items should be quick, especially if our ordering is by hash.
2. We actually don't even order anything there :) The `Ordered` part is just keeping the already existing order.

marc-casperlabs: There may be merit to using `buffered` instead of `buffer_unordered` instead as well.

marc-casperlabs:
```rust
let mut stream = futures::stream::iter(hashes.cloned())
  .map(|hash| async move { fetch_and_store_deploy(hash, ctx).await })
  .buffered(ctx.config.max_parallel_deploy_fetches());
        
let deploys = stream.collect().await;
ctx.metrics
    .observe_fetch_deploys_duration_seconds(start_instant);
Ok(deploys)
```

marc-casperlabs:
After discussion we figured, any approach using `FuturesOrdered` will not work, as it will allow one long fetch to block the remainder from making progress (e.g. if 100 are fetched concurrently with a limit of 5, and the first one is taking 9 minutes to complete, both the manual `FuturesOrdered` and `.buffered`, which actually uses `FuturesOrdered` behind the scenes, will not attempt to fetch a 6th item before the first one has completed).

marc-casperlabs:
With that in mind, I think the sorting-the-Vec approach is simplest in terms of code. As for computational complexity, it has an $O(n \log n)$ sort as the upper bound on complexity.

Note that the `BTreeMap`-based approach has $n$ insertions of $O(\log n)$ complexity, which is the same thing. So it is _not_ worse than the map.

And to complete the code golfing round: Let's replace the entire `while` loop with `let results: BTreeMap<usize, _> = stream.collect().await;`, which :crossed_fingers: is possible?

marc-casperlabs (approved):
Well, pretty much everything discussed here seems to lead to a valid outcome, so whatever variant you pick LGTM :)

goral09 (approved):

Fraser999:
bors merge

PR #3256, opened 2022-08-11
Update to fix new lint warnings since Rust 1.63.0

Fraser999:
This PR fixes new lints warnings introduced by Rust 1.63.0.

### DIFF
--- a/json_rpc/src/error.rs
+++ b/json_rpc/src/error.rs
@@ -125,7 +125,7 @@ impl Error {
     pub fn new<C: ErrorCodeT, T: Serialize>(error_code: C, additional_info: T) -> Self {
         let (code, message): (i64, &'static str) = error_code.into();
 
-        if !C::is_reserved() && code >= -32768 && code <= -32100 {
+        if !C::is_reserved() && (-32768..=-32100).contains(&code) {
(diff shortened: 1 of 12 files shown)


json_rpc/src/error.rs
```diff
@@ -125,7 +125,7 @@ impl Error {
     pub fn new<C: ErrorCodeT, T: Serialize>(error_code: C, additional_info: T) -> Self {
         let (code, message): (i64, &'static str) = error_code.into();
 
-        if !C::is_reserved() && code >= -32768 && code <= -32100 {
+        if !C::is_reserved() && (-32768..=-32100).contains(&code) {
```
goral09, line 128: Are these really equivalent? They do in terms of the outcome but the old one is just 2 comparisons where the new one is an iterator over couple hundred elements of a list.

mpapierski: It's equal as `(x..=y)` is a `Range` object where `start` and `end` is known, so `contains` is the same logical expression. For reference, this is `Range::contains` https://doc.rust-lang.org/src/core/ops/range.rs.html#119-122, and it calls https://doc.rust-lang.org/src/core/ops/range.rs.html#810-813. It looks a bit fancier, but it's the same.

mpapierski (approved):

PR #3263, opened 2022-08-16
Move `hash_bytes_into_chunks` from execution-engine to `Digest` type.

### LINKED ISSUE #3265: When calculating hash of block's execution results hash serialized results instead of just 32-byte digests of it.

goral09:
We want to reuse the chunking logic in more places, to chunk other data types that just trie, and they all have to follow the same scheme:
* be chunked before storing
* have chunks' IDs being computed (in the fetcher) in a way that is aligned with how the data is stored

By moving this method from execution engine to `Digest` type we're making it easier for developers to use the correct logic instead of writing a new version of it.

Closes https://github.com/casper-network/casper-node/issues/3265 and https://github.com/casper-network/casper-node/issues/3264

### DIFF
--- a/node/src/components/contract_runtime/operations.rs
+++ b/node/src/components/contract_runtime/operations.rs
@@ -402,16 +402,23 @@ where
 }
 
 /// Computes the root hash for a Merkle tree constructed from the hashes of execution results.
+///
+/// NOTE: We're hashing vector of execution results, instead of just their hashes, b/c when a joiner
+/// node receives the chunks of *full data* it has to be able to verify it against the merkle root.
 fn compute_execution_results_root_hash<'a>(
     results: &mut impl Iterator<Item = &'a ExecutionResult>,
 ) -> Result<Digest, BlockCreationError> {
-    let mut execution_results = vec![];
+    let mut execution_results_serialized = vec![];
     for result in results {
-        execution_results.push(Digest::hash(
-            &result.to_bytes().map_err(BlockCreationError::BytesRepr)?,
-        ));
+        execution_results_serialized
+            .push(result.to_bytes().map_err(BlockCreationError::BytesRepr)?);
     }
-    Ok(Digest::hash_merkle_tree(execution_results))
+    let execution_results_bytes = execution_results_serialized
+        .to_bytes()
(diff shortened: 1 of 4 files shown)


node/src/components/contract_runtime/operations.rs
```diff
@@ -402,16 +402,23 @@ where
...
 
 /// Computes the root hash for a Merkle tree constructed from the hashes of execution results.
+///
+/// NOTE: We're hashing vector of execution results, instead of just their hashes, b/c when a joiner
+/// node receives the chunks of *full data* it has to be able to verify it against the merkle root.
 fn compute_execution_results_root_hash<'a>(
     results: &mut impl Iterator<Item = &'a ExecutionResult>,
 ) -> Result<Digest, BlockCreationError> {
-    let mut execution_results = vec![];
+    let mut execution_results_serialized = vec![];
```
Fraser999: I think we need to do something more like changing `results` to a `Vec<ExecutionResult>` and just call `let execution_results_bytes = results.to_bytes()...` - that way we get the length of the vector written to the output bytes first, which is what I think we'll need when serializing the actual execution results to send them.

goral09: We should get it here as well since we're calling `ToBytes` for each `ExecutionResult`. The input to `hash_bytes_into_*` is `execution_results_bytes` which is result of calling `ToBytes` on `Vec<Vec<u8>>` . Each individual element of the outer vector should be deserializable to `ExecutionResult` type (since that's what we serialize) and the outer vector should get its length serialized first as well.

I can change this to:
```rust
let execution_results_bytes = results.clone().to_bytes()
Digest::hash_bytes_into_chunks_if_necessary(&execution_results_bytes )
```
since they should be equivalent.

goral09: Fixed in [`51e7d25` (#3263)](https://github.com/casper-network/casper-node/pull/3263/commits/51e7d2555cbe6230bf85db1b3ce073378eac6cc5).

Fraser999 (approved):

goral09:
bors merge

Issue #3275, opened 2022-08-17
Use has_connected_to_network() in chain sync


cspramit:
This new small_network functionality will be used inside the chain sync fetch operations which use the max_sync_fetch_attempts to exit with failure after hitting the attempt limit.  The attempt counter should remain at 0 while has_connected_to_network returns false, meaning we only start counting attempts from when has_connected_to_network is true.

rafal-ch:
Current state of affairs regarding the fetch operations during the chain sync:

```
| item                     | fetch mechanism                                    | comment                                                      |
|--------------------------|----------------------------------------------------|--------------------------------------------------------------|
| BlockHeader              | has_connected_to_network + max_sync_fetch_attempts | for era supervisor initialization                            |
| BlockHeader              | has_connected_to_network + max_sync_fetch_attempts | to get trusted key block info                                |
| Deploy                   | has_connected_to_network + max_sync_fetch_attempts |                                                              |
| TrieOrChunk              | has_connected_to_network + max_sync_fetch_attempts |                                                              |
| Block                    | has_connected_to_network + max_sync_fetch_attempts | for replay protection                                        |
| BlockAndDeploys          | has_connected_to_network + max_sync_fetch_attempts | fetch_blocks_and_state_and_finality_signatures_since_genesis |
| BlockHeadersBatch        | has_connected_to_network + max_sync_fetch_attempts |                                                              |
| BlockSignatures          | FINALITY_SIGNATURE_FETCH_RETRY_COUNT               | only retries in case of time-out                             |
| FinalizedApprovalsWithId | single fetcher request                             |                                                              |
| Block & BlockHeader      | single fetcher request                             | while syncing headers up to current era                      |
| Block & BlockHeader      | single fetcher request                             | while preparing for upgrade                                  |
| Block & BlockHeader      | single fetcher request                             | while fetching and executing blocks                          |
| BlockHeaderWithMetadata  | single fetcher request                             | syncing up to current era                                    |
```

PR #3276, opened 2022-08-18
Reduce the number of minimum peer count by one

rafal-ch:
This PR reduces the number or peers required to perform fetch to avoid node stalls in case it has its own address specified in the `known_addresses` list in the config.

This is a temporary solution and is going to be replaced after https://github.com/casper-network/casper-node/issues/3275 is implemented.

### DIFF
--- a/node/src/components/chain_synchronizer/config.rs
+++ b/node/src/components/chain_synchronizer/config.rs
@@ -35,7 +35,8 @@ pub(super) struct Config {
     pub(crate) redemption_interval: u32,
     /// Block and block header fetch operations are retried forever until we have enough connected
     /// peers. If the operation fails while there are enough peers, the process gives up. By
-    /// default, this is the number of items on the `known_addresses` list in the node config.
+    /// default, this is the number of items on the `known_addresses` list in the node config
+    /// reduced by one.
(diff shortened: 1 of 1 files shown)


fizyk20 (approved):

Fraser999 (approved):

rafal-ch:
bors retry

node/src/components/chain_synchronizer/config.rs
```diff
@@ -35,7 +35,8 @@ pub(super) struct Config {
     pub(crate) redemption_interval: u32,
     /// Block and block header fetch operations are retried forever until we have enough connected
     /// peers. If the operation fails while there are enough peers, the process gives up. By
-    /// default, this is the number of items on the `known_addresses` list in the node config.
+    /// default, this is the number of items on the `known_addresses` list in the node config
+    /// reduced by one.
```
goral09, line 39: Would be nice to know why we're subtracting that 1.

rafal-ch: I included that in the PR description. I've seen no additional benefit of adding the comment in the code since:
1) the reason is solely connected to the way NCTL is configured, not to the system behavior as such
2) it is a throw-away code anyway

PR #3282, opened 2022-08-23
validate approvals based on merkle root in chain sync

SamuelSchlesinger:
issue: https://github.com/casper-network/casper-node/issues/3240

### DIFF
--- a/node/src/components/chain_synchronizer/operations.rs
+++ b/node/src/components/chain_synchronizer/operations.rs
@@ -1,6 +1,6 @@
 use std::{
     cmp,
-    collections::{BTreeMap, VecDeque},
+    collections::{BTreeMap, BTreeSet, VecDeque},
     mem,
     sync::{
         atomic::{AtomicBool, AtomicI64, AtomicU64, Ordering},
@@ -21,9 +21,15 @@ use serde::Serialize;
 use tokio::sync::Semaphore;
 use tracing::{debug, error, info, trace, warn};
 
-use casper_execution_engine::storage::trie::{TrieOrChunk, TrieOrChunkId};
+use casper_execution_engine::{
+    core::engine_state::{self, QueryRequest, QueryResult},
+    storage::trie::{TrieOrChunk, TrieOrChunkId},
+};
 use casper_hashing::Digest;
-use casper_types::{bytesrepr::Bytes, EraId, PublicKey, TimeDiff, Timestamp, U512};
+use casper_types::{
+    bytesrepr::{self, Bytes, ToBytes},
+    CLValueError, EraId, Key, PublicKey, StoredValue, TimeDiff, Timestamp, U512,
+};
 
 use crate::{
     components::{
@@ -46,7 +52,7 @@ use crate::{
     },
     storage::StorageRequest,
     types::{
-        AvailableBlockRange, Block, BlockAndDeploys, BlockHash, BlockHeader,
+        Approval, AvailableBlockRange, Block, BlockAndDeploys, BlockHash, BlockHeader,
         BlockHeaderWithMetadata, BlockHeadersBatch, BlockHeadersBatchId, BlockSignatures,
         BlockWithMetadata, Deploy, DeployHash, FinalizedApprovals, FinalizedApprovalsWithId,
         FinalizedBlock, Item, NodeId,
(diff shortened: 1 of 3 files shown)


node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,147 @@ where
...
+where
+    REv: From<NetworkInfoRequest>
+        + From<FetcherRequest<BlockAndDeploys>>
+        + From<StorageRequest>
+        + From<ContractRuntimeRequest>,
+{
+    match ctx
+        .effect_builder
+        .query_global_state(QueryRequest::new(
+            *ctx.trusted_block_header().id().inner(),
```
afck: Isn't that the block hash? I believe it should be the state root hash?

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,147 @@ where
...
+                        } else {
+                            error!("no approvals for deploy {:?}", dh);
+                            return Err(FetchAndStoreBlockError::NoApprovalsForDeploy(dh));
+                        }
+                    }
+
+                    if !approval_hashes.is_empty()
+                        && Digest::hash_merkle_tree(approval_hashes) != deploy_approvals_root_hash
+                    {
+                        continue 'a;
```
afck, line 1188: Should we disconnect from the peer?

node/src/components/storage.rs
```diff
@@ -556,7 +556,25 @@ impl Storage {
         match incoming.message {
             NetRequest::Deploy(ref serialized_id) => {
                 let id = decode_item_id::<Deploy>(serialized_id)?;
-                let opt_item = self.get_deploy(id).map_err(FatalStorageError::from)?;
+                let mut opt_item = self.get_deploy(id).map_err(FatalStorageError::from)?;
+
+                match &mut opt_item {
+                    Some(deploy) => {
```
afck: Since the `None` branch is empty:

```suggestion
                if let Some(deploy) = &mut opt_item {
```

node/src/components/chain_synchronizer/operations.rs
SamuelSchlesinger: Thank you, yes :+1:

node/src/components/chain_synchronizer/operations.rs
SamuelSchlesinger: I'm not sure, is that what we do in other places if something is off?

SamuelSchlesinger: Yep, it seems appropriate to disconnect here based on the rest of the code.

node/src/components/storage.rs
```diff
@@ -556,7 +556,25 @@ impl Storage {
...
+                        let mut txn = self
+                            .env
+                            .begin_ro_txn()
+                            .expect("could not create RO transaction");
```
goral09: We don't panic! in the code if I'm not mistaken so this should become a fatal error.

SamuelSchlesinger: I saw this kind of panic invoked by a few functions specifically around opening transactions in this module, but perhaps we should remove them all. Maybe this could be a tech debt ticket and not addressed in this PR?\

Edit: Nope, the `FatalStorageError` is the function to use instead.

node/src/components/storage.rs
```diff
@@ -556,7 +556,25 @@ impl Storage {
...
+                        if let Some(finalized_approvals) = txn
+                            .get_value(self.finalized_approvals_db, &id)
+                            .map_err(|e| {
+                                GetRequestError::Fatal(FatalStorageError::InternalStorage(e))
+                            })?
+                        {
+                            deploy.replace_approvals(finalized_approvals);
```
goral09: So here, we would replace approvals we got from the GlobalState with approvals that the node has finalized? If they're different then:
*  Either our state root hashes are different (different approvals could lead to different execution result) and we will not sync up completely.
* Or it is a bug where the same set of deploy approvals (signatures) lead to the same execution results.


If it's the former then it's fine b/c joining node will not become a part of the same network. If it's the latter then I think we should actually use the signatures from the GlobalState. Otherwise users may find it confusing where the signatures from `finalized_approvals_db` differ from the ones in the GlobalState. The data should be coherent.

node/src/components/storage.rs
```diff
@@ -648,10 +666,30 @@ impl Storage {
...
                     .read_block_and_deploys_by_hash(item_id)
                     .map_err(FatalStorageError::from)?;
 
+                match &mut opt_item {
+                    Some(block_and_deploys) => {
+                        for deploy in block_and_deploys.deploys.iter_mut() {
+                            let mut txn = self
+                                .env
+                                .begin_ro_txn()
+                                .expect("could not create RO transaction");
```
goral09: Same - we should not panic.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
-                .put_block_and_deploys_to_storage(block_and_deploys.clone())
-                .await;
-            Ok(block_and_deploys)
-        }
-    };
-    ctx.metrics
-        .observe_fetch_block_and_deploys_duration_seconds(start);
-    res
+    'a: loop {
+        let fetched_block = fetch_retry_forever::<_, BlockAndDeploys>(ctx, block_hash)
```
goral09: We're working on removing all `*_forever` instances from the code (cc @rafal-ch ). Can you make a note to fix that after that work is completed or follow the new approach? Rafał will know better whether we have the code to support that.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+                        .map(|dh| (*dh, approvals.get(dh)))
+                    {
+                        if let Some(approvals) = maybe_approvals {
+                            approval_hashes.push(Digest::hash(
+                                approvals
+                                    .to_bytes()
+                                    .map_err(FetchAndStoreBlockError::BytesRepr)?,
+                            ));
+                        } else {
+                            error!("no approvals for deploy {:?}", dh);
```
goral09:
```suggestion
                            error!(deploy_hash=?dh, "no approvals for deploy");
```
That way we will get an error log in a JSON format where `deploy_hash` is a field in the JSON object, rather than it being a part of a log message, which can be used for querying.

Current:
```json 
{"message": "no approvals for deploy af1c3..."}
```
suggested:
```json
{"message": "no approvals for deploy", "deploy_hash": "af1c3..."}
```

node/src/components/chain_synchronizer/operations.rs
SamuelSchlesinger: I can probably just remove it here. What is the preferred approach?

node/src/components/storage.rs
SamuelSchlesinger: I'm not sure I understand this -- I thought the approval sets could be different and were not previously stored in global state but in a table meant for them. Now, we're only storing the finalized approvals in GlobalState so no one's execution results should differ cause that must be the same after consensus.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+                        .block
+                        .body()
+                        .deploy_hashes()
+                        .iter()
+                        .chain(block_and_deploys.block.body().transfer_hashes().iter())
+                        .map(|dh| (*dh, approvals.get(dh)))
+                    {
+                        if let Some(approvals) = maybe_approvals {
+                            approval_hashes.push(Digest::hash(
+                                approvals
```
goral09, line 1173: This can also be empty, right? If `d.approvals()` returned an empty collection.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+                                    .to_bytes()
+                                    .map_err(FetchAndStoreBlockError::BytesRepr)?,
+                            ));
+                        } else {
+                            error!(deploy_hash=?dh, "no approvals for deploy");
+                            return Err(FetchAndStoreBlockError::NoApprovalsForDeploy(dh));
+                        }
+                    }
+
+                    if !approval_hashes.is_empty()
```
goral09, line 1183: Do we even need to check for this? We should never have a collection of `deploys` where there are no approvals on them. Maybe that can be its own check that is made early on in the flow.

node/src/components/storage.rs
```diff
@@ -556,7 +556,22 @@ impl Storage {
...
+                        .begin_ro_txn()
+                        .map_err(|e| FatalStorageError::InternalStorage(e.into()))?;
+                    if let Some(finalized_approvals) = txn
+                        .get_value(self.finalized_approvals_db, &id)
+                        .map_err(|e| {
+                            GetRequestError::Fatal(FatalStorageError::InternalStorage(e))
+                        })?
+                    {
+                        deploy.replace_approvals(finalized_approvals);
+                    }
```
goral09, line 567: I can see why `NetRequest::BlockAndDeploys` now adds finalized approvals - b/c it's used in the `chain_synchronizer/operations.rs` when syncing but where in the syncing process do we make use of the fact that also `NetResponse::Deploy` will include the finalized approvals?

node/src/components/storage.rs
goral09: It's my fault. I misunderstood the branch we're here in.

node/src/components/storage.rs
SamuelSchlesinger: We don't, but Fraser suggested I make the change to this `NetRequest` as well.

node/src/components/chain_synchronizer/operations.rs
SamuelSchlesinger: I think checking it where its relevant is the most clear thing, but I'm not too fixed on it.

node/src/components/chain_synchronizer/operations.rs
SamuelSchlesinger: Yes, it can be.

goral09: In that case, shouldn't it also return early with an error like the `else` branch?

SamuelSchlesinger: I guess I wasn't sure if here was the place to reject an empty approval set, but I can do so.

node/src/components/chain_synchronizer/operations.rs
rafal-ch: @SamuelSchlesinger I suggest leaving it as is for now as the preferred approach is still in development ([#3273](https://github.com/casper-network/casper-node/issues/3273), [#3274](https://github.com/casper-network/casper-node/issues/3274), [#3275](https://github.com/casper-network/casper-node/issues/3275)).

Depending on whether tickets above or your PR is done first, we'll then together adapt the code.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+        ))
+        .await
+        .map_err(FetchAndStoreBlockError::Engine)?
+    {
+        QueryResult::RootNotFound => {
+            error!("couldn't find the state root for the trusted header");
+            Err(FetchAndStoreBlockError::TrustedStateRootMissing)
+        }
+        QueryResult::DepthLimit { depth } => {
+            error!("reached depth limit: {:?}", depth);
```
rafal-ch: Nit:
```suggestion
            error!("reached depth limit: {}", depth);
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+                    }
+                    ctx.effect_builder
+                        .put_block_and_deploys_to_storage(block_and_deploys.clone())
+                        .await;
+                    Ok(block_and_deploys)
+                } else {
+                    Ok(block_and_deploys)
+                }
+            }
+        };
```
rafal-ch:
```suggestion
                    ctx.effect_builder
                        .put_block_and_deploys_to_storage(block_and_deploys.clone())
                        .await;
                }
                Ok(block_and_deploys)
            }
        };
```

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+        ctx.metrics
+            .observe_fetch_block_and_deploys_duration_seconds(start);
```
rafal-ch, line 1198: We'll not observe the metric in case we return early. It is possible that we will fetch a block with deploys and even the approvals root hash, but will not log time.

Please see if it is worth using the [ScopeTimer](https://github.com/casper-network/casper-node/blob/dev/node/src/components/chain_synchronizer/operations.rs#L73-L81) for this case (although, you'd need to modify it a bit, since it is currently hardcoded to use `self.gauge.set(...)`.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+    CLValue(CLValueError),
+    Engine(engine_state::Error),
+    BytesRepr(bytesrepr::Error),
+    TrustedStateRootMissing,
+    Impossible(String),
+    NoApprovalsForDeploy(DeployHash),
+}
+
+/// Retrieves the deploy approvals root hash from global state
+async fn fetch_deploy_approvals_root_hash<REv>(
```
rafal-ch, line 1072: I think we should add a metric for this fetch.

node/src/components/chain_synchronizer/operations.rs
```diff
@@ -948,34 +954,149 @@ where
...
+                        } else {
+                            error!(deploy_hash=?dh, "no approvals for deploy");
+                            return Err(FetchAndStoreBlockError::NoApprovalsForDeploy(dh));
+                        }
+                    }
+
+                    if !approval_hashes.is_empty()
+                        && Digest::hash_merkle_tree(approval_hashes) != deploy_approvals_root_hash
+                    {
+                        ctx.effect_builder.announce_disconnect_from_peer(peer).await;
```
rafal-ch, line 1187: Please add a `warn!` on disconnect, for example:
```
warn!(?peer, "fetched deploy approvals hash mismatch");
```

node/src/components/chain_synchronizer/operations.rs
afck: Yes, in general if a peer is sending us clearly wrong data (that _must_ be their fault), we disconnect.

node/src/components/chain_synchronizer/operations.rs
SamuelSchlesinger: I have not added a metric before -- is there an example you can point me to?

rafal-ch: Certainly.
You start in the `node/src/components/chain_synchronizer/metrics.rs`, when you need to add a new metric and register it in `new()`. Please decide on a name consistent with other metrics.

Then, because the `fetch_deploy_approvals_root_hash()` is a recurring operation we need to sum up time spend on all invocations; we can do it via a call to `ctx.metrics.observe_...()`.

Please have a look at `fetch_and_store_deploys()` as an example - it does exactly what we need.

Fraser999: I disagree about adding a metric for this, since it should always return fast, but maybe there was a notion that this function would try to get the item from peers if unavailable locally?  I'd suggest renaming the function to `get_deploy_approvals_root_hash` just to try and make it a little more obvious that it's a call to our own storage only and that (unlike the other fetch functions) we don't try any peers.

Fraser999 (commented):
No blockers from me, but I'll wait to approve once the other reviewers' discussions are resolved :)

node/src/components/storage.rs
Fraser999: We fetch `Deploy`s in chain sync too (`fetch_and_store_deploy`).  I think it's least surprising and also probably more efficient overall to have those provided with the finalized approvals too.  I don't think it ever really makes sense for a peer to provide its subjective (different) set of approvals to another peer.

node/src/components/chain_synchronizer/operations.rs
rafal-ch: Yeah, I was fooled by the `fetch_` part of the name. If it's designed to return fast, we should not bloat the metrics.

PR #3284, opened 2022-08-23
Shorten debug formatting on dropped responders

marc-casperlabs:
This PR shortens the debug representation of some potentially large types, as when dropping JSON RPC requests.

It also adds a little bit of testing and documentation to the `utils` module.

### DIFF
--- a/node/src/types/json_compatibility/account.rs
+++ b/node/src/types/json_compatibility/account.rs
@@ -1,16 +1,12 @@
 // TODO - remove once schemars stops causing warning.
 #![allow(clippy::field_reassign_with_default)]
 
-use std::fmt::Debug;
-
 use datasize::DataSize;
 use once_cell::sync::Lazy;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
-use crate::{
-    rpcs::docs::DocExample, types::json_compatibility::vectorize, utils::fmt_limit::LimitSlice,
-};
+use crate::{rpcs::docs::DocExample, types::json_compatibility::vectorize};
 use casper_types::{
     account::{Account as ExecutionEngineAccount, AccountHash},
     NamedKey, PublicKey, SecretKey, URef,
@@ -56,7 +52,7 @@ struct ActionThresholds {
 }
 
 /// Structure representing a user's account, stored in global state.
-#[derive(PartialEq, Eq, Clone, Serialize, Deserialize, DataSize, JsonSchema)]
+#[derive(PartialEq, Eq, Clone, Debug, Serialize, Deserialize, DataSize, JsonSchema)]
 #[serde(deny_unknown_fields)]
 pub struct Account {
     account_hash: AccountHash,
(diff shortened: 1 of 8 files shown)


node/src/types/json_compatibility/account.rs
```diff
@@ -64,6 +68,21 @@ pub struct Account {
...
+impl Debug for Account {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Account")
+            .field("account_hash", &self.account_hash)
+            .field(
+                "named_keys",
+                &LimitSlice::<_, 10>::new(self.named_keys.as_slice()),
+            )
+            .field("main_purse", &self.main_purse)
+            .field("associated_keys", &self.associated_keys)
```
goral09: As per `max_associated_keys = 100` entry in the `chainspec.toml` this can include up to 100 entries as well.

node/src/utils/fmt_limit.rs
```diff
@@ -0,0 +1,62 @@
...
+    #[test]
+    fn limit_debug_works() {
+        let collections: Vec<_> = (0..5).into_iter().collect();
+
+        // Sanity check.
+        assert_eq!(format!("{:?}", collections), "[0, 1, 2, 3, 4]");
+
+        assert_eq!(
+            format!("{:?}", LimitSlice::<'_, _, 3>::new(collections.as_slice())),
+            "[0, 1, 2 ...]"
```
goral09: Why is there no comma after `2` and before `...` ? `2` would have an `idx` of 2 (`enumerate` starts from 0) so `idx + 1 != count` b/c `2 + 1 != 5`

node/src/types/json_compatibility/account.rs
marc-casperlabs: It's not unbounded though, is it? Is it possible that 100 entries is still "reasonable"?

node/src/utils/fmt_limit.rs
marc-casperlabs: > Why is there no comma after 2 and before ... ?

Because I thought it'd look nicer that way :cry: . I guess `[0, 1, 2, ...]` is still better, now that I think about it...

goral09: Ah, right... This particular example uses different `N` . For some reason I thought it would be `5`.

node/src/types/json_compatibility/account.rs
goral09: It's a `BTreeMap<AccountHash, Weight>`.

node/src/utils/fmt_limit.rs
marc-casperlabs: I changed the format and added more tests.

node/src/types/json_compatibility/account.rs
marc-casperlabs: So that's around 5 KB-ish? Not great, but not as as bad I think.

marc-casperlabs:
Revamped this, as the wrong types were hit anyway. Now we just limit the output specifically on the logging of dropped responders.

goral09 (approved):

Fraser999 (approved):

PR #3285, opened 2022-08-23
Network bootstrap completion indication

### LINKED ISSUE #3274: Add `bootstrap_completed` to small_network

marc-casperlabs:
Closes #3274.

Adds the following features to the networking component:

* A mapping of known addresses to the last seen `NodeId` seen when connecting to said known address is kept (`known_node_ids`).
* A variable `bootstrap_completed` is introduced, which tracks whether the network considers itself "started up". Once set, it can never be cleared.
* The state of `bootstrap_completed` can be queried through a request (`EffectBuilder::get_networking_bootstrap_state`), this is intended to be used with fast sync code.
* Three conditions are introduced, and checked in the networking component (see `update_bootstrapping_state`) to keep the bootstrapping state up to date.

No changelog entry, as this behavior is only visible (sans a log message) with fast sync, which is not present in earlier versions.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -560,6 +560,8 @@ where
                     // connection after a peer has closed the corresponding incoming connection.
                 }
 
+                self.update_bootstrapping_state();
+
                 // Now we can start the message reader.
                 let boxed_span = Box::new(span.clone());
                 effects.extend(
@@ -741,6 +743,8 @@ where
                     self.update_syncing_nodes_set(peer_id, is_syncing);
                 }
 
+                self.update_bootstrapping_state();
+
                 effects.extend(
                     tasks::message_sender(
                         receiver,
@@ -873,6 +877,54 @@ where
         }
 
         ret
+    }
+
+    /// Updates the bootstrapping state of the node according to available information.
+    fn update_bootstrapping_state(&mut self) {
+        // Do not spend any effort if we already completed our bootstrapping process.
+        if self.bootstrap_completed {
+            return;
+        }
+
+        // Check for the regular connection threshold, since counting is quicker.
+        let symmetric_connections: HashSet<NodeId> = self
+            .connection_symmetries
+            .iter()
+            .filter_map(|(node_id, cs)| cs.is_symmetric().then_some(*node_id))
+            .collect();
+
+        if symmetric_connections.len() > self.cfg.bootstrap_threshold.connection_count as usize {
+            self.bootstrap_completed = true;
+            info!(
+                target = self.cfg.bootstrap_threshold.connection_count,
+                actual = symmetric_connections.len(),
+                "networking bootstrap completed after connection threshold was met"
+            );
+            return;
+        }
+
+        // Check if we have reached the critical of known nodes connected.
+        let known_nodes_target = (self.cfg.known_addresses.len() as f32
+            * self
+                .cfg
+                .bootstrap_threshold
+                .known_address_connection_percentage)
+            .round() as usize;
+
+        let known_node_ids: HashSet<_> = self.known_node_ids.values().cloned().collect();
+        let known_nodes_connected = symmetric_connections.intersection(&known_node_ids);
+
+        let known_nodes_connected_count = known_nodes_connected.into_iter().count();
+
+        if known_nodes_connected_count > known_nodes_target {
+            self.bootstrap_completed = true;
+            info!(
+                target = known_nodes_target,
+                actual = known_nodes_connected_count,
+                "networking bootstrap completed after known node connection threshold was met"
+            );
+            return;
+        }
     }
 
     /// Returns the node id of this network node.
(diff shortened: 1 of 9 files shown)


resources/local/config.toml
```diff
@@ -206,6 +206,16 @@ trie_responses = 0
 finalized_approvals_requests = 1
 finalized_approvals_responses = 0
 
+# Threshold for when the networking considers itself sufficiently connected.
+#
+# Some processes will not begin (or fail) unless this threshold is reached.
```
Fraser999:
```suggestion
# Thresholds for considering bootstrapping to the network to be finished.
# 
# When any one of the conditions is met, bootstrapping is considered finished.
# Some processes will not begin (or fail) until bootstrapping is finished.
```

resources/local/config.toml
```diff
@@ -206,6 +206,16 @@ trie_responses = 0
...
+[network.bootstrap_threshold]
```
Fraser999:
```suggestion
[network.bootstrap_thresholds]
```

resources/local/config.toml
```diff
@@ -206,6 +206,16 @@ trie_responses = 0
...
+# A percentage of known addresses has been reached.
+known_address_connection_percentage = 0.5
```
Fraser999: In the config, we already have `saturation_limit_percent` which is an integer, and `acceleration_ftt` which is a percentage defined as a TOML array.  I think we should avoid using a float here, and instead use an integer (i.e. change the `0.5` to `50`.

resources/local/config.toml
```diff
@@ -206,6 +206,16 @@ trie_responses = 0
 finalized_approvals_requests = 1
 finalized_approvals_responses = 0
 
+# Threshold for when the networking considers itself sufficiently connected.
+#
+# Some processes will not begin (or fail) unless this threshold is reached.
+[network.bootstrap_threshold]
+# A percentage of known addresses has been reached.
```
Fraser999:
```suggestion
# Symmetric connections to the given percentage of known addresses have been established.
```

node/src/components/small_network.rs
```diff
@@ -156,6 +156,14 @@ where
     outgoing_manager: OutgoingManager<OutgoingHandle<P>, ConnectionError>,
     /// Tracks whether a connection is symmetric or not.
     connection_symmetries: HashMap<NodeId, ConnectionSymmetry>,
+    /// Mapping of known node addresses to their potentially known `NodeId`.
+    known_node_ids: HashMap<SocketAddr, NodeId>,
+
+    /// An indicator whether or not the networking component considers itself bootstrapped.
+    ///
+    /// Indented to be used by other components (which can request it) that rely on the state of
```
rafal-ch:
```suggestion
    /// Intended to be used by other components (which can request it) that rely on the state of
```

node/src/components/small_network.rs
```diff
@@ -851,6 +886,53 @@ where
...
+            self.bootstrap_completed = true;
+            info!(
+                target = self.cfg.bootstrap_threshold.connection_count,
+                actual = symmetric_connections.len(),
+                "networking bootstrap completed after connection threshold was met"
+            );
+            return;
+        }
+
+        // Check if we have reached the critical of known nodes connected.
```
rafal-ch:
```suggestion
        // Check if we have reached the critical number of known nodes connected.
```

node/src/effect.rs
```diff
@@ -778,6 +778,22 @@ impl<REv> EffectBuilder<REv> {
         .await
     }
 
+    /// Gets the networks current bootstrapping state.
```
rafal-ch:
```suggestion
    /// Gets the network's current bootstrapping state.
```

node/src/components/small_network.rs
```diff
@@ -851,6 +886,53 @@ where
         ret
     }
 
+    /// Updates the bootstrapping state of the node according to available information.
+    fn update_bootstrapping_state(&mut self) {
```
rafal-ch, line 891: What about adding a unit test to prove that the calculations, filter_maps, and intersections as a whole work as expected?

I'd suggest introducing a function taking the following parameters:
- self.cfg.bootstrap_threshold
- self.connection_symmetries
- self.cfg.known_addresses
- possibly more (compiler will tell)

Then in `update_bootstrapping_state` you can return early on `self.bootstrap_completed` (as you do) or invoke the proposed function.

This will allow adding unit test with different inputs exercising various edge cases.

Also, to have a single source of truth, I'd suggest adding a check for timeout threshold also here, by having yet another testable `if` instead of an async-based `set_timeout` mechanism.

marc-casperlabs:
Like the underlying issue, this is no longer relevant.

PR #3289, opened 2022-08-24
Pin stable rust

### LINKED ISSUE #3257: Add pinned version of the Rust toolchain to the node
We should add a `rust-toolchain.toml` file to the root of the repo with the following contents:
```toml
[toolchain]
channel = "1.63"
```
The Makefile `setup-rs` target will also need to be updated to replace both instances of `stable` with the value read from the `rust-toolchain.toml` file using something like
```
PINNED_STABLE  := $(shell grep -Eo '[0-9]+\.[0-9]+\.?[0-9]*' rust-toolchain.toml)
```

darthsiroftardis:
CHANGELOG:

- Added a rust-toolchain.toml pinning rust stable 1.63

Closes #3257

### DIFF
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -38,7 +38,7 @@ jobs:
 
       - name: Get stable from rust-toolchain.toml
         id: stable-toolchain
-        run: echo "::set-output name=version::$(grep -Eo '[0-9]+\.[0-9]+\.?[0-9]*' rust-toolchain.toml)"
+        run: echo "::set-output name=version::$(sed -nr 's/channel\s+=\s+\"(.*)\"/\1/p' rust-toolchain.toml)"
 
       - name: Install Toolchain - Stable
         uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af #tag v1.0.7
--- a/.github/workflows/lints.yml
+++ b/.github/workflows/lints.yml
@@ -36,11 +36,15 @@ jobs:
         id: nightly-toolchain
         run: echo "::set-output name=version::$(cat smart_contracts/rust-toolchain)"
 
+      - name: Get stable from rust-toolchain.toml
+        id: stable-toolchain
+        run: echo "::set-output name=version::$(shell grep -Eo '[0-9]+\.[0-9]+\.?[0-9]*' rust-toolchain.toml)"
+
       - name: Install Toolchain - Stable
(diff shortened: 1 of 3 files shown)


Fraser999 (approved):

rafal-ch:
Can you explain in the description **why** we need that? The connected ticket only says what we should do, not why.

.github/workflows/lints.yml
```diff
@@ -36,11 +36,15 @@ jobs:
         id: nightly-toolchain
         run: echo "::set-output name=version::$(cat smart_contracts/rust-toolchain)"
 
+      - name: Get stable from rust-toolchain.toml
+        id: stable-toolchain
+        run: echo "::set-output name=version::$(grep -Eo '[0-9]+\.[0-9]+\.?[0-9]*' rust-toolchain.toml)"
```
rafal-ch: Could you consider using:
```
sed -nr 's/channel\s+=\s+\"(.*)\"/\1/p' rust-toolchain.toml
```

This way we'll also support toolchain specifiers such as
```
[toolchain]
channel = "stable"
```
and possibly all other combinations.

Ultimately, some toml parser could come in handy (something similar to `jq` for json).

Makefile
```diff
@@ -4,6 +4,7 @@ RUSTUP = $(or $(shell which rustup), $(HOME)/.cargo/bin/rustup)
 NPM    = $(or $(shell which npm),    /usr/bin/npm)
 
 PINNED_NIGHTLY := $(shell cat smart_contracts/rust-toolchain)
+PINNED_STABLE  := $(shell grep -Eo '[0-9]+\.[0-9]+\.?[0-9]*' rust-toolchain.toml)
```
rafal-ch: Please consider leveraging `sed` here as well.

.github/workflows/lints.yml
darthsiroftardis: No worries, I was going with the recommendation in the ticket, but can do for sure yeah!

Makefile
darthsiroftardis: Applied suggestion

.github/workflows/lints.yml
darthsiroftardis: Applied suggestion

Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

goral09:
bors cancel

PR #3291, opened 2022-08-25
GH-3268: Add support for CA signed client certificates

### LINKED ISSUE #3268: Add a switch for known_addresses to act as a list of a nodes only allowed to join

mpapierski:
Closes #3268

This commit introduces an option to load a node's identity from existing certificates signed by a CA authority. This is optional and is helpful for some private chains, where joining new nodes to an existing private network is a restricted process.

# Config

There is a new configuration option in `config.toml.`

```toml
[network.identity]
tls_certificate = "node_cert.pem"
secret_key = "node.pem"
ca_certificate = "ca_cert.pem"
```

The node will load the existing secret key, certificate, and CA cert at startup if specified.

When new peer nodes are connecting, the node process will validate peer client certificates against provided CA certificates and rejects those not signed by provided CA.

# Testing

Script that generates `ca_key.pem` `ca_cert.pem` and bunch of `local_node*.pem` certificates signed by CA

https://gist.github.com/mpapierski/bd80f7519f697daa0337e27f44d81b89

Connecting with a `openssl s_client` with client key signed by a different CA cert to a nctl network using shared ca_cert.pem:

```
2022-08-25T14:39:34.127048Z DEBUG init:incoming{; peer_addr=127.0.0.1:55958}: [casper_node::components::small_network small_network.rs:478] incoming connection failed early; err=TLS validation error of peer certificate: the certificate is not signed by provided certificate authority
```

Otherwise, a network reaches consensus if every node in the network is set up with shared CA with certificates signed by said CA.

### DIFF
--- a/node/src/components/small_network.rs
+++ b/node/src/components/small_network.rs
@@ -56,7 +56,7 @@ use std::{
 
 use datasize::DataSize;
 use futures::{future::BoxFuture, FutureExt};
-use openssl::{error::ErrorStack as OpenSslErrorStack, pkey};
+use openssl::{error::ErrorStack as OpenSslErrorStack, pkey, x509::X509};
 use pkey::{PKey, Private};
 use prometheus::Registry;
 use rand::{prelude::SliceRandom, seq::IteratorRandom};
@@ -86,6 +86,7 @@ pub(crate) use self::{
 };
 use self::{
     chain_info::ChainInfo,
+    config::IdentityConfig,
     counting_format::{ConnectionId, CountingFormat, Role},
     error::{ConnectionError, Result},
     event::{IncomingConnection, OutgoingConnection},
@@ -107,7 +108,7 @@ use crate::{
         AutoClosingResponder, EffectBuilder, EffectExt, Effects,
     },
     reactor::{EventQueueHandle, Finalize, ReactorEvent},
-    tls::{self, TlsCert, ValidationError},
+    tls::{self, LoadCertError, LoadSecretKeyError, TlsCert, ValidationError},
     types::NodeId,
     utils::{self, display_error, Source, WithDir},
     NodeRng,
@@ -310,6 +311,7 @@ where
             event_queue,
             our_id: NodeId::from(&small_network_identity),
             our_cert: small_network_identity.tls_certificate,
+            our_ca: small_network_identity.ca_certificate,
             secret_key: small_network_identity.secret_key,
             net_metrics: Arc::downgrade(&net_metrics),
             chain_info,
@@ -1117,6 +1119,10 @@ pub(crate) enum SmallNetworkIdentityError {
     CouldNotGenerateTlsCertificate(OpenSslErrorStack),
     #[error(transparent)]
     ValidationError(#[from] ValidationError),
+    #[error(transparent)]
+    LoadCertError(#[from] LoadCertError),
+    #[error(transparent)]
+    LoadSecretKeyError(#[from] LoadSecretKeyError),
 }
 
 /// An ephemeral [PKey<Private>] and [TlsCert] that identifies this node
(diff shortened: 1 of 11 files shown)


node/src/components/small_network.rs
```diff
@@ -1117,24 +1119,62 @@ pub(crate) enum SmallNetworkIdentityError {
...
 impl SmallNetworkIdentity {
-    pub(crate) fn new() -> result::Result<Self, SmallNetworkIdentityError> {
-        let (not_yet_validated_x509_cert, secret_key) = tls::generate_node_cert()
-            .map_err(SmallNetworkIdentityError::CouldNotGenerateTlsCertificate)?;
-        let tls_certificate = tls::validate_cert(not_yet_validated_x509_cert)?;
-        Ok(SmallNetworkIdentity {
+    fn new(
+        secret_key: PKey<Private>,
+        tls_certificate: TlsCert,
+        ca_certificate: Option<X509>,
```
marc-casperlabs: I don't think we want to tie the CA certificate into the small network identity here. It should just be known to the networking component, unless I am missing something?

If we're loading them from a file, I don't think we need to validate them upon loading in `SmallNetworkIdentity`.

node/src/components/small_network.rs
```diff
@@ -310,6 +311,7 @@ where
             event_queue,
             our_id: NodeId::from(&small_network_identity),
             our_cert: small_network_identity.tls_certificate,
+            our_ca: small_network_identity.ca_certificate,
```
marc-casperlabs: I think this should be named `network_ca`, since it is shared across all nodes in the network.

node/src/reactor/initializer.rs
```diff
@@ -236,7 +236,8 @@ impl Reactor {
 
         let effects = reactor::wrap_effects(Event::Chainspec, chainspec_effects);
 
-        let small_network_identity = SmallNetworkIdentity::new()?;
+        let network_config = config.map_ref(|cfg| cfg.network.clone());
```
marc-casperlabs: Just :golfing_man: , but could this be

```rust
let small_network_identity = config.map_ref(|cfg| SmallNetworkIdentity::from_config(&cfg.network));
```

?

node/src/tls.rs
```diff
@@ -246,23 +245,51 @@ impl PartialEq for TlsCert {
...
+        #[serde(skip_serializing)]
+        #[source]
+        ErrorStack,
+    ),
+}
+
+pub(crate) fn load_secret_key<P: AsRef<Path>>(src: P) -> Result<PKey<Private>, LoadSecretKeyError> {
+    let pem = read_file(src.as_ref()).map_err(LoadSecretKeyError::ReadFile)?;
+
+    // TODO: It might be that we need to call `PKey::private_key_from_pkcs8` instead.
```
marc-casperlabs: Can you elaborate on this? :)

node/src/tls.rs
```diff
@@ -433,13 +460,95 @@ pub enum ValidationError {
...
+        .public_key()
+        .map_err(ValidationError::CannotReadCAPublicKey)?;
+    if !cert
+        .verify(authority_key.as_ref())
+        .map_err(ValidationError::FailedToValidateSignature)?
+    {
+        return Err(ValidationError::WrongCertificateAuthority);
+    }
+
+    // Check expiration times against current time.
```
marc-casperlabs: If this code copy&pasted from the other validation function, can we factor it out or merge these two, so we DRY here?

node/src/tls.rs
```diff
@@ -817,17 +873,79 @@ mod tests {
...
+    }
 
-        assert_eq!(serialized, serialized_again);
+    fn make_ca_signed_cert(private_key: PKey<Private>, ca_private_key: PKey<Private>) -> X509 {
+        let mut builder = X509Builder::new().unwrap();
+        builder.set_version(2).unwrap();
+        builder
+            .set_serial_number(mknum(1).unwrap().as_ref())
+            .unwrap();
+        let issuer = mkname("US", "Casper Blockchain", "node-0").unwrap();
```
marc-casperlabs: `"node-0"` is probably not a good issuer name here. How about `"network"` or `"network-ca"`?

Minor though, since it is just testing code.

marc-casperlabs (changes requested):
Formally, this is missing `CHANGELOG.md` entries and documentation for this new feature (/cc @Fraser999 -- into the `README.md`?)

Other than that, I think the only blockers are the repeated code in cert validation and the integration of the CA cert into the `SmallNetworkIdentity`.

node/src/tls.rs
mpapierski: I don't have more context on that. It's copy pasted from here. I can remove it if it's outdated?

https://github.com/casper-network/casper-node/blob/f7bc8b64630ec871f3c2219d646bdc7ebc91dbd9/node/src/tls.rs#L546-L551

node/src/reactor/initializer.rs
mpapierski: Applied

node/src/tls.rs
marc-casperlabs: Touché.

IIRC, I was not sure about the storage format of the private key. We can remove the comment.

node/src/reactor/initializer.rs
mpapierski: Actually, not. `SmallNetworkIdentity::from_config` returns `Result` and `config.map_ref` returns `WithDir<T>` so it's a bit cumbersome to work with

node/src/tls.rs
mpapierski: Applied (used `Casper Network`)

node/src/tls.rs
mpapierski: Tried my best to DRY as much as possible

node/src/tls.rs
mpapierski: Removed

node/src/components/small_network.rs
mpapierski: Renamed

node/src/components/small_network.rs
mpapierski: Removed in favor of loading network_ca in a network component

marc-casperlabs (approved):
Looks good to me :)

PR #3311, opened 2022-09-19
query-balance -> get-balance

jonas089:
casper-client query-balance does not exist, changed it to get-balance and tested successful.


Fraser999:
Thanks for the PR, however the `query-balance` exists in the Rust client on the `dev` branch.  Generally the casper-client's `dev` branch is kept aligned with the casper-node's `dev` one and once the node undergoes an upgrade, the client's `dev` branch will be merged to `main` and a new client version published.

Thanks anyway, but I'll close this unmerged.

PR #3313, opened 2022-09-20
nctl debug command line switch

zacshowa:
Added an optional flag to nctl-compile to allow for easier non-release compilation.

### DIFF
--- a/utils/nctl/sh/assets/compile.sh
+++ b/utils/nctl/sh/assets/compile.sh
@@ -6,15 +6,32 @@
 #   NCTL - path to nctl home directory.
 ########################################
 
-pull optional flag to run in debug mode
+unset OPTIND #clean OPTIND envvar, otherwise getopts can break.
+COMPILE_MODE="release" #default compile mode to release.
+
 while getopts 'd' opt; do 
     case $opt in
-        d ) export NCTL_COMPILE_TARGET="debug";;
-        * ) echo "nctl-compile only accepts optional flag -d to compile in debug mode."
+        d )
+            COMPILE_MODE="debug"
+            ;;
+        * )
+            
+            COMPILE_MODE="release"
+            ;; #ignore other cl flags
     esac
 done
 
-source "$NCTL"/sh/assets/compile_node.sh 
-source "$NCTL"/sh/assets/compile_node_launcher.sh
-source "$NCTL"/sh/assets/compile_client.sh
-source "$NCTL"/sh/assets/compile_global_state_update_gen.sh
+if [ "$NCTL_COMPILE_TARGET" = "debug" ] || [ "$COMPILE_MODE" == "debug" ]; then
+    source "$NCTL"/sh/assets/compile_node.sh -d
+    source "$NCTL"/sh/assets/compile_node_launcher.sh -d
+    source "$NCTL"/sh/assets/compile_client.sh -d 
+    source "$NCTL"/sh/assets/compile_global_state_update_gen.sh -d
+else
+    source "$NCTL"/sh/assets/compile_node.sh
+    source "$NCTL"/sh/assets/compile_node_launcher.sh
+    source "$NCTL"/sh/assets/compile_client.sh
+    source "$NCTL"/sh/assets/compile_global_state_update_gen.sh
+fi
+
+unset COMPILE_MODE
+unset OPTIND #clean all envvar garbage we may have produced. 
--- a/utils/nctl/sh/assets/compile.sh
+++ b/utils/nctl/sh/assets/compile.sh
@@ -6,7 +6,15 @@
 #   NCTL - path to nctl home directory.
 ########################################
 
-source "$NCTL"/sh/assets/compile_node.sh
+pull optional flag to run in debug mode
+while getopts 'd' opt; do 
+    case $opt in
+        d ) export NCTL_COMPILE_TARGET="debug";;
+        * ) echo "nctl-compile only accepts optional flag -d to compile in debug mode."
+    esac
+done
+
+source "$NCTL"/sh/assets/compile_node.sh 
 source "$NCTL"/sh/assets/compile_node_launcher.sh
 source "$NCTL"/sh/assets/compile_client.sh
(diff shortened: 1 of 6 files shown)


utils/nctl/sh/assets/compile.sh
```diff
@@ -6,7 +6,15 @@
 #   NCTL - path to nctl home directory.
 ########################################
 
-source "$NCTL"/sh/assets/compile_node.sh
+pull optional flag to run in debug mode
```
rafal-ch: Is this intentional? For me it results in:
```
Command 'pull' not found, did you mean
```

utils/nctl/sh/assets/compile_node_launcher.sh
```diff
@@ -14,6 +14,13 @@ source "$NCTL"/sh/utils/main.sh
 pushd "$NCTL_CASPER_NODE_LAUNCHER_HOME" || \
     { echo "Could not find the casper-node-launcher repo - have you cloned it into your working directory?"; exit; }
 
+while getopts 'd' opt; do 
+    case $opt in
+        d ) export NCTL_COMPILE_TARGET="debug";;
+        * ) echo "nctl-compile only accepts optional flag -d to compile in debug mode."
```
rafal-ch: It's not a `nctl-compile` it this case, right? It's rather `compile_node_launcher`.

I would suggest either of the two approaches:

1. use the generic `this script only accepts...` and then stop the script execution
2. (preferred by me) just silently ignore unknown parameter and continue

However, I'm not familiar with the requirements, so this is just a suggestion.

utils/nctl/sh/assets/compile.sh
```diff
@@ -6,7 +6,15 @@
...
+while getopts 'd' opt; do 
+    case $opt in
+        d ) export NCTL_COMPILE_TARGET="debug";;
```
rafal-ch: Handling the parameter like that will result in the following:
```
magister@zacisze:~/Casper/casper-node$ nctl-compile-node
Finished release [optimized] target(s) in 7m 03s
magister@zacisze:~/Casper/casper-node$ nctl-compile-node -d
Finished dev [unoptimized + debuginfo] target(s) in 0.12s
magister@zacisze:~/Casper/casper-node$ nctl-compile-node
Finished dev [unoptimized + debuginfo] target(s) in 0.12s
```

So the third time I started it **without** `-d` flag, yet it still compiled the `dev`.

utils/nctl/sh/assets/compile_client.sh
```diff
@@ -33,6 +33,13 @@ popd || exit
 pushd "$NCTL_CASPER_CLIENT_HOME" || \
     { echo "Could not find the casper-client-rs repo - have you cloned it into your working directory?"; exit; }
 
+while getopts 'd' opt; do 
+    case $opt in
+        d ) export NCTL_COMPILE_TARGET="debug";;
```
rafal-ch: Hmm, just thinking out loud. Maybe we should consider being consistent with the approach taken by Rust ecosystem and by default compile in `debug` and support the `--release` command line option?

That might be a too much of a lift, though, since `NCTL` is already wired-up in many places that expect it to compile in release by default.

So treat it just as something for consideration.

CC: @TomVasile

utils/nctl/sh/assets/compile.sh
zacshowa: This was not intentional and has been fixed in my local files.

utils/nctl/sh/assets/compile_node_launcher.sh
zacshowa: I agree that that this error handling might be a little ambitious for this part of the tool. I chose to implement option 2 and just ignore unknown parameters.

utils/nctl/sh/assets/compile.sh
rafal-ch: ~Can you also add this fix to the PR?~

Nvm, I was looking at the wrong commit. All good :+1:

utils/nctl/sh/assets/compile.sh
rafal-ch: Works as intended, thanks :+1:

utils/nctl/sh/assets/compile_node.sh
```diff
@@ -12,17 +12,26 @@ source "$NCTL"/sh/utils/main.sh
...
+        d ) 
+            COMPILE_MODE="debug"
+            ;;
+        * ) 
+            ;; #ignore other cl flags
     esac
 done
 
-if [ "$NCTL_COMPILE_TARGET" = "debug" ]; then
+if [ "$NCTL_COMPILE_TARGET" = "debug" ] || [ "$COMPILE_MODE" = "debug" ]; then
```
rafal-ch, line 28: Can you add a short info to [the docs](https://github.com/casper-network/casper-node/blob/dev/utils/nctl/docs/commands-setup.md?plain=1#L11) on how to switch between compile modes? I think it's worth mentioning that `debug` mode takes preference, i.e., passing `-d` will override the `NCTL_COMPILE_TARGET=release` and vice versa.

rafal-ch (approved):
LGTM, no blockers :+1: 

But before merging, please get an additional approval from a person with better bash skills than mine :)

utils/nctl/sh/assets/compile_node_launcher.sh
```diff
@@ -14,10 +14,26 @@ source "$NCTL"/sh/utils/main.sh
...
+
+while getopts 'd' opt; do 
+    case $opt in
+        d ) 
+            COMPILE_MODE="debug" ;;
+        * ) ;; #ignore other cl flags
+
+    esac
+done
+
```
TomVasile, line 28: nit pick: double newline

TomVasile (approved):
Tested locally and both the new and old method works as expected. Thanks for the PR!

utils/nctl/sh/assets/compile_node.sh
TomVasile: second adding to docs

PR #3317, opened 2022-09-22
"sslkeylogfile"-style support

### LINKED ISSUE #3316: Add `SSLKEYLOGFILE`-style TLS logging support
To ease debugging of potential networking issues, add support for logging TLS session keys of OpenSSL to the node, allowing captured traffic of `nctl` and similar to be dissected using tools like [wireshark](https://wireshark.org/).

If time permits, a simple dissector script would be helpful as well.

marc-casperlabs:
Closes #3316.

This PR adds support for logging the ephemeral keys generated during regular TLS operations, allowing network dumps of the node to be read with tools like [wireshark](https://wireshark.org).

Instead of using the `SSLKEYLOGFILE` environment file to set this, we use a configuration setting that is disabled everywhere by default, except when using `nctl` to spin up local networks.

Writes to the keylog use file locking, so that one file can be used for all nodes in a single nctl network.


node/src/components/small_network/config.rs
```diff
@@ -81,6 +82,8 @@ pub struct Config {
     pub public_address: String,
     /// Known address of a node on the network used for joining.
     pub known_addresses: Vec<String>,
+    /// If set, logs all TLS keys to this file.
+    pub keylog_path: Option<String>,
```
dwerner, line 86: Suggest using a `PathBuf`

node/src/components/small_network.rs
```diff
@@ -330,12 +331,28 @@ where
 
         let chain_info = chain_info_source.into();
         let protocol_version = chain_info.protocol_version;
+
+        let keylog = match cfg.keylog_path {
```
dwerner, line 335: Since the binding is `Option<_>` anyway, I'd suggest using a `.map()`

node/src/components/small_network/tasks.rs
```diff
@@ -97,14 +97,18 @@ where
...
-        .map_err(ConnectionError::TlsInitialization)?;
+    let mut transport = tls::create_tls_connector(
+        context.our_cert.as_x509(),
+        &context.secret_key,
+        context.keylog.clone(),
+    )
+    .and_then(|connector| connector.configure())
+    .and_then(|mut config| {
+        config.set_verify_hostname(false);
+        config.into_ssl("this-will-not-be-checked.example.com")
```
dwerner, line 108: Why will this not be checked? Suggest a comment here that describes behavior.

node/src/utils.rs
```diff
@@ -405,6 +407,47 @@ pub(crate) async fn wait_for_arc_drop<T>(
     false
 }
 
+/// A thread-safe wrapper around a file that writes chunks.
+///
+/// A chunk can (but needn't) be a line. The writer guarantees it will be written to the wrapped
+/// file, even if other threads are attempting to write chunks at the same time.
+#[derive(Clone)]
+pub(crate) struct LockedLineWriter(Arc<Mutex<File>>);
```
dwerner, line 415: Did you consider `tokio::fs::File` here? Seems like it already wraps a mutex like you're doing here. https://docs.rs/tokio/latest/src/tokio/fs/file.rs.html#93-96

In addition, you probably wouldn't have to handle lock poisoning as per below.

dwerner (approved):
Minor nits, suggestions

node/src/components/small_network/config.rs
marc-casperlabs: I used a string because that's what we have to pass to the OpenSSL API. Is there a benefit for the potentially fallible conversion from `String` (config file) to `PathBuf` (here) just to go back to `String`?

node/src/components/small_network.rs
marc-casperlabs: My bad, good catch!

marc-casperlabs: Ah, I was a little hasty perhaps - there's a `?` operator returning an error in line 342. Using `map` would then make the code require something like `.map(|path| { ... Ok(writer)}).transpose()?` to thread through the error and require a type annotation. Additionally, `cfg.keylog_path.as_ref()` would have to be used to avoid the move.

I think at that point the explicit version becomes preferable for sheer readability.

node/src/components/small_network/tasks.rs
marc-casperlabs: This will not be checked due to the fact that we do not have domain names. TLS verification is usually certificate against domain and issuer. There's no domain to be checked at all, our "name" is the certificates fingerprint, which is unforgeable.

This code did not change in this PR as well, it's just a formatting change due to the line above getting an extra parameter.

node/src/utils.rs
marc-casperlabs: The latest version of `tokio::fs::File` does not implement synchronous `io::Write`, as far as I can tell. We need a synchronous version though, since all of the writing is done in a sync callback that is passed to the OpenSSL C code, so I think we cannot get away with it here.

marc-casperlabs:
@piotr-dziubecki  Note: Holding off on merging this until the `dev` branch has moved on from 1.5.

marc-casperlabs:
Change of plans: This is going into the `feat-muxink-integration` branch. Currently, all these changes can be found squashed as f7ae86e4b6157894902603163c83d15c57b8bf75, although that is bound to change due to upcoming rebases.

Closing this in the meantime.

PR #3326, opened 2022-10-31
`muxink` crate

marc-casperlabs:
This is the `muxink` crate, the basis for backpressuring networking to be merged into the 1.6 feature.

Not much more to write in this PR, as the crate documentation should cover the details -- if it does not, please comment, so it can be fixed.


muxink/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "muxink"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1.0.57"
```
goral09: Can we get rid of this dependency? It always seemed to me that it's a good one if you write a CLI tools and don't bother about handling errors properly. `muxink` is a library and probably a very-well written so it does (🤞 ) return all results in a typed fashion?

marc-casperlabs: Good catch. It's actually not in use anymore, it was used in earlier versions.

marc-casperlabs: Fixed in 2c98125a0

muxink/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+bincode = { version = "1.3.3", optional = true }
+bytes = "1.1.0"
+futures = "0.3.21"
+serde = { version = "1.0.138", optional = true }
+thiserror = "1.0.31"
+tokio = { version = "1", features = [ "full" ] }      # TODO: Reduce features.
```
goral09: While we're here - do you plan to reduce these features as part of the PR?

marc-casperlabs: Good point, fixed in 8c6374d77.

dwerner (approved):

PR #3351, opened 2022-11-09
Update child map and progress timestamp only with sufficient signatures.

afck:
Don't update the block acceptor's progress timestamp or the block accumulator's block children map based on anything other than sufficiently signed blocks, and additional signatures for them.

Otherwise a malicious peer or validator can cause us to increment the progress timestamp indefinitely, or to add wrong entries to the map.

Closes https://github.com/casper-network/casper-node/issues/3342, https://github.com/casper-network/casper-node/issues/3343.

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -9,6 +9,7 @@ use std::{cmp::Ordering, collections::BTreeMap, iter};
 
 use datasize::DataSize;
 use futures::FutureExt;
+use itertools::Itertools;
 use tracing::{debug, error, warn};
 
 use casper_types::{EraId, TimeDiff, Timestamp};
@@ -282,9 +283,6 @@ impl BlockAccumulator {
             debug!(%block_hash, "ignoring outdated block");
             return Effects::new();
         }
-        if let Some(parent_hash) = block.parent() {
-            self.block_children.insert(*parent_hash, *block_hash);
-        }
 
         let acceptor = self
             .block_acceptors
@@ -293,10 +291,14 @@ impl BlockAccumulator {
 
         match acceptor.register_block(block, sender) {
             Ok(_) => match self.validator_matrix.validator_weights(era_id) {
-                Some(evw) => store_block_and_finality_signatures(
-                    effect_builder,
-                    acceptor.should_store_block(&evw),
-                ),
+                Some(evw) => {
+                    let (should_store, faulty_senders) = acceptor.should_store_block(&evw);
+                    self.store_block_and_finality_signatures(
+                        effect_builder,
+                        should_store,
+                        faulty_senders,
+                    )
+                }
                 None => Effects::new(),
             },
             Err(error) => match error {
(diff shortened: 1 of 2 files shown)


node/src/components/block_accumulator/block_acceptor.rs
```diff
@@ -93,7 +93,6 @@ impl BlockAcceptor {
         }
 
         if self.block.is_none() {
-            self.touch();
```
rafal-ch, line 96: So this `touch()` could be removed because we now touch only [here](https://github.com/casper-network/casper-node/blob/91e05f4ab21bfee79bcd9ce9de6cfa8eba055983/node/src/components/block_accumulator/block_acceptor.rs#L207), right?

rafal-ch (approved):
LGTM :+1:

node/src/components/block_accumulator.rs
```diff
@@ -616,62 +671,26 @@ impl<REv: ReactorEvent> Component<REv> for BlockAccumulator {
 
 impl<REv: ReactorEvent> ValidatorBoundComponent<REv> for BlockAccumulator {
     fn handle_validators(&mut self, effect_builder: EffectBuilder<REv>) -> Effects<Self::Event> {
-        let mut effects = Effects::new();
-        for block_acceptor in self
+        let validator_matrix = &self.validator_matrix; // Closure can't borrow all of self.
+        let should_stores = self
```
georgepisaltu, line 675: nit part 2:
```suggestion
        let should_stores: Vec<_> = self
```

node/src/components/block_accumulator.rs
```diff
@@ -616,62 +671,26 @@ impl<REv: ReactorEvent> Component<REv> for BlockAccumulator {
...
-                }
-            }
-        }
-        effects
+            .filter_map(|acceptor| {
+                let era_id = acceptor.era_id()?;
+                let evw = validator_matrix.validator_weights(era_id)?;
+                Some(acceptor.should_store_block(&evw))
+            })
+            .collect_vec();
```
georgepisaltu, line 684: nit part 3:
```suggestion
            .collect();
```

node/src/components/block_accumulator.rs
```diff
@@ -9,6 +9,7 @@ use std::{cmp::Ordering, collections::BTreeMap, iter};
 
 use datasize::DataSize;
 use futures::FutureExt;
+use itertools::Itertools;
```
georgepisaltu, line 12: nit part 1: this is unnecessary.

georgepisaltu (approved):
Just a nit, otherwise LGTM!

node/src/components/block_accumulator/block_acceptor.rs
afck: Yes, and here we don't know yet whether the block is sufficiently signed.

node/src/components/block_accumulator.rs
afck: I tried it, but then Clippy complains due to https://github.com/rust-lang/rust-clippy/issues/9219.
Since we're already using itertools elsewhere anyway, I'd postpone that change until the Clippy bug is fixed.

PR #3354, opened 2022-11-11
sync to historical functioning

EdHastingsCasperAssociation:
joining nodes with sync_to_genesis: true will first catch up to tip, and then while keeping up also acquire older blocks by going backwards.

### DIFF
--- a/node/src/components/block_synchronizer/deploy_acquisition.rs
+++ b/node/src/components/block_synchronizer/deploy_acquisition.rs
@@ -1,7 +1,6 @@
 use std::{
-    collections::HashMap,
+    cmp::Ord,
     fmt::{Display, Formatter},
-    hash::Hash,
 };
 
 use datasize::DataSize;
@@ -17,7 +16,7 @@ pub(crate) enum Error {
     EncounteredNonVacantDeployState,
 }
 
-#[derive(Clone, DataSize, Debug)]
+#[derive(Clone, PartialEq, Eq, DataSize, Debug)]
 pub(super) enum DeployAcquisition {
     ByHash(Acquisition<DeployHash>),
     ById(Acquisition<DeployId>),
@@ -54,19 +53,19 @@ impl DeployAcquisition {
     ) -> Result<(), Error> {
         let new_acquisition = match self {
             DeployAcquisition::ByHash(acquisition) => {
-                let mut new_deploy_ids = HashMap::new();
+                let mut new_deploy_ids = vec![];
                 for ((deploy_hash, deploy_state), approvals_hash) in acquisition
                     .inner
-                    .drain()
+                    .drain(..)
                     .zip(approvals_hashes.approvals_hashes())
                 {
                     if !matches!(deploy_state, DeployState::Vacant) {
                         return Err(Error::EncounteredNonVacantDeployState);
                     };
-                    new_deploy_ids.insert(
+                    new_deploy_ids.push((
                         DeployId::new(deploy_hash, *approvals_hash),
                         DeployState::Vacant,
-                    );
+                    ));
                 }
 
                 DeployAcquisition::ById(Acquisition {
@@ -92,23 +91,20 @@ impl DeployAcquisition {
     }
 }
 
-#[derive(Clone, Copy, PartialEq, Eq, Hash, DataSize, Debug, Default)]
+#[derive(Clone, Copy, PartialEq, Eq, DataSize, Debug, Default)]
 pub(super) enum DeployState {
     #[default]
     Vacant,
     HaveDeployBody,
 }
 
-#[derive(Clone, DataSize, Debug)]
-pub(super) struct Acquisition<T>
-where
-    T: Hash,
-{
-    inner: HashMap<T, DeployState>,
+#[derive(Clone, PartialEq, Eq, DataSize, Debug)]
+pub(super) struct Acquisition<T> {
+    inner: Vec<(T, DeployState)>,
     need_execution_result: bool,
 }
 
-impl<T: Copy + Eq + Hash> Acquisition<T> {
+impl<T: Copy + Ord> Acquisition<T> {
     fn new(deploy_identifiers: Vec<T>, need_execution_result: bool) -> Self {
         let inner = deploy_identifiers
             .into_iter()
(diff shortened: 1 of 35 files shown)


node/src/components/storage.rs
```diff
@@ -1851,14 +1855,15 @@ impl Storage {
                         return Ok(None);
                     }
                 };
-
             result.push(parent_block_header.clone());
             if parent_block_header.is_switch_block() {
-                break;
+                depth += 1;
+                if depth >= era_depth || parent_block_header.is_genesis() {
```
afck: On mainnet, genesis is not a switch block, so I think we would end up returning `Ok(None)`.
Not sure if we're handling that case correctly yet?

node/src/components/storage.rs
```diff
@@ -2290,14 +2295,16 @@ impl Storage {
             None => return Ok(FetchResponse::NotFound(sync_leap_identifier)),
         };
 
+        let trusted_ancestors_only = sync_leap_identifier.trusted_ancestor_only();
+        let era_depth = if trusted_ancestors_only { 2 } else { 1 };
```
afck: Isn't 1 already enough, even for the sync-to-genesis case?
We should only need the switch block that tells us the validators who are supposed to sign `block_hash`?

node/src/components/small_network/tasks.rs
```diff
@@ -665,135 +665,110 @@ where
...
-                                .await;
-                        }
-                        Err(msg) => {
-                            // We've received a non-demand message. Ensure we have the proper amount
-                            // of resources, then push it to the reactor.
-                            limiter
-                                .request_allowance(
-                                    msg.payload_incoming_resource_estimate(
+    let read_messages =
+        async move {
```
afck: I think `make format` uses a different version and would not change the formatting here.

node/src/components/storage.rs
EdHastingsCasperAssociation: removed era_depth per our convo

node/src/components/block_synchronizer/deploy_acquisition.rs
```diff
@@ -52,19 +54,19 @@ impl DeployAcquisition {
     ) -> Result<(), Error> {
         let new_acquisition = match self {
             DeployAcquisition::ByHash(acquisition) => {
-                let mut new_deploy_ids = vec![];
+                let mut new_deploy_ids = HashMap::new();
                 for ((deploy_hash, deploy_state), approvals_hash) in acquisition
                     .inner
-                    .drain(..)
+                    .drain()
```
afck: Won't that zip them in the wrong order, since `acquisition.inner` is now a `HashMap`?

node/src/components/fetcher/item_fetcher.rs
```diff
@@ -144,21 +144,26 @@ pub(super) trait ItemFetcher<T: FetcherItem + 'static> {
...
-                    )
-                    .ignore();
+                return Effects::new();
+                // todo! lets discuss whether its safe to disconnect here
+                // return effect_builder
+                //     .announce_block_peer_with_justification(
+                //         peer,
+                //         BlocklistJustification::SentBadItem { tag: T::TAG },
+                //     )
+                //     .ignore();
```
afck: I agree, it seems dangerous to ban because of that.

node/src/reactor/main_reactor/control.rs
```diff
@@ -736,135 +722,181 @@ impl MainReactor {
...
+                            "Historical: sync leap awaiting".to_string(),
+                            self.control_logic_default_delay.into(),
+                        ),
                     }
-                }
-                BlockSynchronizerProgress::Syncing(_, maybe_block_height, last_progress) => {
-                    error!(
-                        "XXX attempting historical syncing maybe_block_height: {:?}",
-                        maybe_block_height
+                } else if self.block_synchronizer.register_block_by_hash(
```
afck, line 818: `else`… was this the `self.validator_matrix.has_era(&era_id)` case? :sweat_smile: 
At some point we should extract some of this into smaller methods. (Not necessarily in this PR.)

node/src/reactor/main_reactor/control.rs
```diff
@@ -736,135 +722,181 @@ impl MainReactor {
...
+        if let BlockSynchronizerProgress::Syncing(_, _, last_progress) = block_synchronizer_progress
+        {
+            // do idleness / reattempt checking
+            let sync_progress = *last_progress;
+            if sync_progress > self.last_progress {
+                self.last_progress = sync_progress;
+                // if any progress has been made, reset attempts
+                self.attempts = 0;
+            }
+            if Timestamp::now().saturating_diff(self.last_progress) > self.idle_tolerance {
```
afck: I think we have an `elapsed` method:
```suggestion
            if self.last_progress.elapsed() > self.idle_tolerance {
```

node/src/reactor/main_reactor/control.rs
EdHastingsCasperAssociation: yes; we have era validators here.

node/src/reactor/main_reactor/control.rs
EdHastingsCasperAssociation: applied

Fraser999 (approved):
Approved assuming you fix up my mess with the `HashMap` in `DeployAcquisition` :)

node/src/components/block_synchronizer/deploy_acquisition.rs
EdHastingsCasperAssociation: yes, will fix up

node/src/components/fetcher/item_fetcher.rs
EdHastingsCasperAssociation: removed commented out code

node/src/components/block_synchronizer/deploy_acquisition.rs
```diff
@@ -123,12 +119,13 @@ impl<T: Copy + Eq + Hash> Acquisition<T> {
...
-            .insert(deploy_identifier, DeployState::HaveDeployBody)
-            .is_none()
+            .contains(&(deploy_identifier, DeployState::HaveDeployBody))
         {
-            Acceptance::NeededIt
-        } else {
             Acceptance::HadIt
+        } else {
+            self.inner
+                .push((deploy_identifier, DeployState::HaveDeployBody));
```
afck, line 127: Won't it now contain both an entry `(deploy_identifier, DeployState::Vacant)` _and_ `(deploy_identifier, DeployState::HaveDeployBody)`?
And the `HaveDeployBody` entries will be in the order in which we received the deploys, not the order in the block? (Maybe that's fine, but I find it surprising.)

EdHastingsCasperAssociation: it did indeed. I merged before I saw your comment but luckily I independently noticed the issue and fixed it in https://github.com/casper-network/casper-node/pull/3382/files#diff-a3e41acc9e099ca403e69b23f718eff1db0ea277539af686925f24f971a98966R124-R137

PR #3355, opened 2022-11-11
Minor small network changes/improvements

### LINKED ISSUE #3345: Clean up small network

marc-casperlabs:
Closes #3345.

* Removes the notion of a `validator_id` from the limiter (as the presence of an ID is not enough to conclude a peer is a validator).
* The number of eras to consider when determining if a peer is an active/upcoming validator is now calculated from the auction delay.


afck (approved):

marc-casperlabs:
bors cancel

PR #3356, opened 2022-11-14
Make sure we do not propose deploys that have not been gossiped yet

### LINKED ISSUE #3349: Re-add delay to deploy buffer for deploys being considered in appendable_block

rafal-ch:
Deploy buffer is tracking which deploy hashes have been gossiped and only proposes such deploys to the appendable block.

The `DEFAULT_DEPLOY_DELAY` has been removed.

Fixes #3349

### DIFF
--- a/node/src/components/deploy_buffer.rs
+++ b/node/src/components/deploy_buffer.rs
@@ -58,6 +58,9 @@ pub(crate) struct DeployBuffer {
     // to allow consensus to proceed safely without risking
     // duplicating deploys
     chain_index: BTreeMap<u64, Timestamp>,
+    // Stores information about deploy hashes that have already been gossiped.
+    // Deploy whose hashes have not been gossiped will never be proposed.
+    gossiped_deploy_hashes: BTreeSet<DeployHash>,
 }
 
 impl DeployBuffer {
@@ -71,6 +74,7 @@ impl DeployBuffer {
             hold: BTreeMap::new(),
             dead: HashSet::new(),
             chain_index: BTreeMap::new(),
+            gossiped_deploy_hashes: BTreeSet::new(),
         }
     }
 
@@ -175,6 +179,7 @@ impl DeployBuffer {
                 self.buffer.insert(*hash, (expiry_time, None));
             }
             self.dead.insert(*hash);
+            self.gossiped_deploy_hashes.remove(hash);
         }
         // deploys held for proposed blocks which did not get finalized in time are eligible again
         let (hold, _) = mem::take(&mut self.hold)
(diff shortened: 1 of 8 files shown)


node/src/components/deploy_buffer.rs
```diff
@@ -202,6 +217,9 @@ impl DeployBuffer {
         let mut holds = vec![];
         for (deploy_hash, footprint, approvals) in self.proposable() {
             let with_approvals = DeployHashWithApprovals::new(deploy_hash, approvals);
+            if !self.gossiped_deploy_hashes.remove(&deploy_hash) {
```
afck: Is that necessary? It's guaranteed 14 lines above that this is unreachable.

node/src/components/deploy_buffer.rs
```diff
@@ -58,6 +58,9 @@ pub(crate) struct DeployBuffer {
     // to allow consensus to proceed safely without risking
     // duplicating deploys
     chain_index: BTreeMap<u64, Timestamp>,
+    // Stores information about deploy hashes that have already been gossiped.
+    // Deploy whose hashes have not been gossiped will never be proposed.
+    gossiped_deploy_hashes: BTreeSet<DeployHash>,
```
afck: It looks like this will always be a subset of `buffer`'s keys; would it make sense to add a bool to `buffer` instead of another collection? It gets difficult to ensure that all those collections are pruned correctly.

afck (dismissed):
(No blockers.)

node/src/components/deploy_buffer.rs
rafal-ch: I'd like to keep it as is, especially that the guarantee is in a different function. It may happen that someone removes the filter in `fn proposable()` without updating `fn appendable_block()`

node/src/components/deploy_buffer.rs
rafal-ch: Here's the commit that removes the collection: https://github.com/casper-network/casper-node/pull/3356/commits/a0a9ef53c49cae15737716e7c894ee2159fae054 (and here's minor update: https://github.com/casper-network/casper-node/pull/3356/commits/560407c3de693f2c8ca469b411cb69b63d928dcb)

node/src/components/deploy_buffer.rs
```diff
@@ -32,6 +32,35 @@ pub(crate) use event::Event;
 
 type FootprintAndApprovals = (DeployFootprint, BTreeSet<Approval>);
 
+#[derive(DataSize, Debug)]
+struct DeployDetails {
+    expiry_time: Option<Timestamp>,
```
afck, line 37: Ah, right, we can't know the expiry time in all cases anymore, not even an upper bound. :confused: 
But that's also bad for purging.
Maybe we need two timestamps, the expiry time and the time where we added the entry? Then we can purge those entries that don't have the former if the latter got too old?

It would be really good if the deploy gossiper could provide more information.
Alternatively, when we get the deploy gossiped notification, we could just load the deploy from disk, so we have everything? We should have stored the deploy at that point, after all?

rafal-ch: Summoning @Fraser999 as this may require some discussion and possibly a separate ticket.

rafal-ch:
Closing, other approach will be proposed.

PR #3357, opened 2022-11-14
Rate limit block acceptor creation per peer.

### LINKED ISSUE #3344: Block accumulator: add upper bounds for number of acceptors and children map

afck:
This limits the number of new `BlockAcceptor` instances we create, per peer informing us about the block.

The rate is computed based on the `purge_interval` and the minimum block time.

Closes #3344.

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -112,7 +112,7 @@ pub(crate) struct BlockAccumulator {
     /// created the block acceptor, from oldest to newest.
     peer_block_timestamps: BTreeMap<NodeId, VecDeque<(BlockHash, Timestamp)>>,
     /// The minimum time between a block and its child.
-    block_time: TimeDiff,
+    min_block_time: TimeDiff,
 }
 
 impl BlockAccumulator {
@@ -121,7 +121,7 @@ impl BlockAccumulator {
         validator_matrix: ValidatorMatrix,
         local_tip_height_and_era_id: Option<(u64, EraId)>,
         recent_era_interval: u64,
-        block_time: TimeDiff,
+        min_block_time: TimeDiff,
     ) -> Self {
         Self {
             validator_matrix,
@@ -135,7 +135,7 @@ impl BlockAccumulator {
                 .map(|(height, era_id)| LocalTipIdentifier::new(height, era_id)),
             recent_era_interval,
             peer_block_timestamps: Default::default(),
-            block_time,
+            min_block_time,
         }
     }
 
@@ -289,8 +289,8 @@ impl BlockAccumulator {
             }
 
             // Assume a block time of at least 1 millisecond, so we don't divide by zero.
-            let block_time = self.block_time.max(TimeDiff::from(1));
-            let expected_blocks = (purge_interval / block_time) as usize;
+            let min_block_time = self.min_block_time.max(TimeDiff::from(1));
+            let expected_blocks = (purge_interval / min_block_time) as usize;
             let max_block_count = PEER_RATE_LIMIT_MULTIPLIER.saturating_mul(expected_blocks);
             if block_timestamps.len() >= max_block_count {
                 warn!(
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -4,6 +4,8 @@ mod error;
 mod event;
 mod starting_with;
 mod sync_instruction;
+#[cfg(test)]
+mod tests;
 
 use std::{
     cmp::Ordering,
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -5,7 +5,11 @@ mod event;
 mod starting_with;
 mod sync_instruction;
 
-use std::{cmp::Ordering, collections::BTreeMap, iter};
+use std::{
+    cmp::Ordering,
+    collections::{btree_map, BTreeMap, VecDeque},
+    iter,
+};
 
 use datasize::DataSize;
 use futures::FutureExt;
@@ -35,6 +39,10 @@ pub use error::Error;
 pub(crate) use event::Event;
 pub(crate) use starting_with::StartingWith;
 pub(crate) use sync_instruction::SyncInstruction;
+
+/// If a peer "informs" us about more than the expected number of new blocks times this factor,
+/// they are probably spamming, and we refuse to create new block acceptors for them.
+const PEER_RATE_LIMIT_MULTIPLIER: usize = 2;
 
 #[derive(Clone, Copy, DataSize, Debug, Eq, PartialEq)]
 struct LocalTipIdentifier {
@@ -98,6 +106,11 @@ pub(crate) struct BlockAccumulator {
     recent_era_interval: u64,
     /// Tracks activity and assists with perceived tip determination.
     last_progress: Timestamp,
+    /// For each peer, a list of block hashes we first heard from them, and the timestamp when we
+    /// created the block acceptor, from oldest to newest.
+    peer_block_timestamps: BTreeMap<NodeId, VecDeque<(BlockHash, Timestamp)>>,
+    /// The minimum time between a block and its child.
+    block_time: TimeDiff,
 }
 
 impl BlockAccumulator {
@@ -106,6 +119,7 @@ impl BlockAccumulator {
         validator_matrix: ValidatorMatrix,
         local_tip_height_and_era_id: Option<(u64, EraId)>,
         recent_era_interval: u64,
+        block_time: TimeDiff,
     ) -> Self {
         Self {
             validator_matrix,
@@ -118,6 +132,8 @@ impl BlockAccumulator {
             local_tip: local_tip_height_and_era_id
                 .map(|(height, era_id)| LocalTipIdentifier::new(height, era_id)),
             recent_era_interval,
+            peer_block_timestamps: Default::default(),
+            block_time,
         }
     }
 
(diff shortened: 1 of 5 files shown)


node/src/components/block_accumulator.rs
```diff
@@ -227,37 +243,63 @@ impl BlockAccumulator {
...
+        // Check that the sender isn't telling us about more blocks than expected.
+        if let Some(sender) = maybe_sender {
+            let block_timestamps = self.peer_block_timestamps.entry(sender).or_default();
+
+            // Prune the timestamps, so the count reflects only the most recently added acceptors.
+            let purge_interval = self.purge_interval;
+            let is_expired = |(_, timestamp): &(_, Timestamp)| timestamp.elapsed() > purge_interval;
+            while block_timestamps.front().map_or(false, is_expired) {
+                block_timestamps.pop_front();
+            }
```
rafal-ch: Can we inline the `is_expired`, making the type annotations redundant?

```suggestion
            while block_timestamps
                .front()
                .map_or(false, |(_, timestamp)| timestamp.elapsed() > purge_interval)
            {
                block_timestamps.pop_front();
            }
```

node/src/components/block_accumulator.rs
```diff
@@ -227,37 +243,63 @@ impl BlockAccumulator {
...
+
+            // Assume a block time of at least 1 millisecond, so we don't divide by zero.
+            let block_time = self.block_time.max(TimeDiff::from(1));
+            let expected_blocks = (purge_interval / block_time) as usize;
+            let max_block_count = PEER_RATE_LIMIT_MULTIPLIER.saturating_mul(expected_blocks);
```
rafal-ch, line 294: This adds a magic number to the formula which is otherwise calculated based on config/chainspec values. One could be tempted to also make the multiplier configurable :-)

node/src/components/block_accumulator.rs
```diff
@@ -227,37 +243,63 @@ impl BlockAccumulator {
...
+            if block_timestamps.len() >= max_block_count {
+                warn!(
+                    ?sender, %block_hash,
+                    "rejecting block hash from peer who sent us more than {} within {}",
```
rafal-ch, line 298: Did we consider banning/demoting such peer?

node/src/components/block_accumulator.rs
afck: I thought about it, but I feel like a factor of 2 should be just fine. :shrug: 
Let's see what others think.

node/src/components/block_accumulator.rs
afck: I'm not sure; so far we're only banning based on concrete, cryptographic evidence that is independent of timing assumptions.

node/src/components/block_accumulator.rs
marc-casperlabs: I don't think why one needs to make it configurable? And we can always refactor that later.

node/src/components/block_accumulator.rs
marc-casperlabs: That's not entirely correct, we have a number of things we can for that are not based on cryptographic evidence. See https://github.com/casper-network/casper-node/blob/88101c7385081424db1091ec2ec8a4f9cac84383/node/src/components/small_network/blocklist.rs#L17-L73 for some.

The important thing is to ensure we can never accidentally (i.e. build-up of networking messages) trigger this ban during honest operation.

afck: I don't see any variant that could be triggered just by timing.
I'd feel a bit uncomfortable risking that.

marc-casperlabs: Well, unless you're comfortable proving it, let's not block :)

node/src/components/block_accumulator.rs
```diff
@@ -98,6 +108,11 @@ pub(crate) struct BlockAccumulator {
     recent_era_interval: u64,
     /// Tracks activity and assists with perceived tip determination.
     last_progress: Timestamp,
+    /// For each peer, a list of block hashes we first heard from them, and the timestamp when we
+    /// created the block acceptor, from oldest to newest.
+    peer_block_timestamps: BTreeMap<NodeId, VecDeque<(BlockHash, Timestamp)>>,
+    /// The minimum time between a block and its child.
+    block_time: TimeDiff,
```
marc-casperlabs: Should this be `min_block_time` then? Or something like `max_accepted_blocks_per_purge_from_peer`? `peer_block_limit`?

node/src/components/block_accumulator.rs
```diff
@@ -227,37 +245,65 @@ impl BlockAccumulator {
...
+                }
+                return;
+            }
+            btree_map::Entry::Vacant(entry) => entry,
+        };
+
+        // The acceptor doesn't exist. Don't create it if the item's era is not provided, the local
+        // tip doesn't have an era or the item's era is older than the local tip era by more than
+        // `recent_era_interval`.
+        match (maybe_era_id, self.local_tip) {
```
marc-casperlabs, line 272: Want to bust out the `if let ... else` here? Or too soon? :)

Maybe I am a little eager to try it out, it'd have two returns in this case...

node/src/components/block_accumulator.rs
```diff
@@ -227,37 +245,65 @@ impl BlockAccumulator {
...
-            // local tip doesn't have an era or the item's era is older than
-            // the local tip era by more than `recent_era_interval`), we only
-            // register the peer if there is an acceptor for the item already.
-            _ => {
-                if let Some(acceptor) = self.block_acceptors.get_mut(&block_hash) {
-                    acceptor.register_peer(sender)
-                }
+
+            // Assume a block time of at least 1 millisecond, so we don't divide by zero.
+            let block_time = self.block_time.max(TimeDiff::from(1));
```
marc-casperlabs: Quietly discarding bad values here is not optimal (i.e. if there's a bad block time configured).

We should probably ensure the `TimeDiff` is in range in the component constructor and return an error, or panic, or just use `TimeDiff::from(1)` there as a default, then rely on this property here.

node/src/components/block_accumulator.rs
```diff
@@ -227,37 +245,65 @@ impl BlockAccumulator {
...
+            let max_block_count = PEER_RATE_LIMIT_MULTIPLIER.saturating_mul(expected_blocks);
+            if block_timestamps.len() >= max_block_count {
+                warn!(
+                    ?sender, %block_hash,
+                    "rejecting block hash from peer who sent us more than {} within {}",
+                    max_block_count, self.purge_interval,
+                );
+                return;
             }
+            block_timestamps.push_back((block_hash, Timestamp::now()));
```
marc-casperlabs, line 303: In general, I like to lift determination of the current timestamp up as much as possible, since every single one of these is a syscall and it makes testing a headache. I think it's fine here though, since we never upsert more than one thing at a time?

node/src/components/block_accumulator.rs
```diff
@@ -280,11 +326,12 @@ impl BlockAccumulator {
...
             return Effects::new();
         }
+        self.upsert_acceptor(*block_hash, Some(era_id), sender);
 
-        let acceptor = self
-            .block_acceptors
-            .entry(*block_hash)
-            .or_insert_with(|| BlockAcceptor::new(*block_hash, None));
+        let acceptor = match self.block_acceptors.get_mut(block_hash) {
+            None => return Effects::new(),
```
marc-casperlabs, line 332: Would it be better to have `upsert_acceptor` return an `Option<&mut BlockAcceptor>` (or even the filled entry) to avoid this kind of construct? If not, can we at least but an explanation in a comment here?

node/src/components/block_accumulator.rs
```diff
@@ -357,30 +404,13 @@ impl BlockAccumulator {
...
-            // is an acceptor for it already.
-            _ => match self.block_acceptors.get_mut(&block_hash) {
-                Some(acceptor) => acceptor,
-                // When there is no acceptor for it, this function returns
-                // early, ignoring the signature.
-                None => return Effects::new(),
-            },
+        self.upsert_acceptor(block_hash, Some(era_id), sender);
+
+        let acceptor = match self.block_acceptors.get_mut(&block_hash) {
```
marc-casperlabs, line 409: See above. We'd do `match self.upsert_acceptor(..)`.

node/src/components/block_accumulator.rs
```diff
@@ -98,6 +108,11 @@ pub(crate) struct BlockAccumulator {
     recent_era_interval: u64,
     /// Tracks activity and assists with perceived tip determination.
     last_progress: Timestamp,
+    /// For each peer, a list of block hashes we first heard from them, and the timestamp when we
```
marc-casperlabs, line 111: This comment should state the invariant that they need to be in-order, I think.

Frankly, I think the logic is sufficiently complicated that this could have been spun-off as its own testable struct (newtype wrapping `VecDeque<(BlockHash, Timestamp)`).

marc-casperlabs (approved):
Looks functional to me, I only have a few nitpicks to make it easier for latecomers to join into the code. Feel free to ignore all of them.

node/src/components/block_accumulator.rs
afck: I'd _love_ to, but I don't know if it's okay at the moment to become incompatible with earlier Rust versions?

node/src/components/block_accumulator.rs
afck: But in theory, a minimum delay of 0 is valid, or at least it will be in 2.0!
And then we don't really have an upper limit for the block per purge time. I'm not sure what to do in that case.

node/src/components/block_accumulator.rs
afck: Yes, there are currently tons of `Timestamp::now()` in the new 1.5 code. Maybe needs discussion.

node/src/components/block_accumulator.rs
afck: I tried, but it doesn't work with the borrow checker, because then all of `self` is borrowed mutably.
I'll add a comment.

rafal-ch (approved):
LGTM :+1:

georgepisaltu (approved):

node/src/components/block_accumulator.rs
afck: The comment says so, in the next line. ("from oldest to newest").

Didn't get a chance to try extracting the `VecDeque`-with-timestamps that enforces chronological order today. Since I have three approvals I'll merge this now, instead of asking everyone to re-review. If I have time, I can try it tomorrow in a separate PR.

node/src/components/block_accumulator.rs
marc-casperlabs: Historically, stable has always been fair game. I'd say if it passes CI, it's good?

node/src/components/block_accumulator.rs
marc-casperlabs: I guess we sort of have to disable the feature entirely in that case? This does not affect the fact that we can front-load the calculation though, right?

node/src/components/block_accumulator.rs
marc-casperlabs: No worries, all good. And apologies for not reading this properly.

node/src/components/block_accumulator.rs
marc-casperlabs: Right, we're missing partial borrows here. There's no way to convey (at the moment) that the returned reference only borrows one field of the struct mutably, instead of the entire thing. :(

PR #3365, opened 2022-11-16
Rename `small_network` to `network`

rafal-ch:
This PR renames the `small_network` component to `network`.


node/src/components/network.rs
```diff
@@ -1,6 +1,6 @@
 //! Fully connected overlay network
 //!
-//! The *small network* is an overlay network where each node participating is attempting to
+//! The *network* is an overlay network where each node participating is attempting to
```
marc-casperlabs: should probably say "network component" now or something like that

marc-casperlabs (approved):
Looks good :)

node/src/components/network.rs
rafal-ch: :+1: 
updated in 77ecb61294e06626c6e0f2a17e03b29ae671aab7

Fraser999 (approved):

PR #3368, opened 2022-11-16
Changes for next test run

EdHastingsCasperAssociation:
* refactor shutdown for upgrade instructions
* improve local_tip detection
* adjust log levels for a few useful log items

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -10,7 +10,6 @@ mod tests;
 use std::{
     cmp::Ordering,
     collections::{btree_map, BTreeMap, VecDeque},
-    iter,
 };
 
 use datasize::DataSize;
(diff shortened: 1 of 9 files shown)


georgepisaltu (approved):

node/src/components/block_accumulator.rs
```diff
@@ -195,8 +196,9 @@ impl BlockAccumulator {
             };
         }
 
-        if let StartingWith::LocalTip(_, height, era_id) = starting_with {
-            self.register_local_tip(height, era_id);
+        if let Some(new_local_tip) = self.maybe_new_local_tip(&starting_with) {
+            self.local_tip = Some(new_local_tip);
+            info!(local_tip=?self.local_tip, "new local tip detected");
```
afck, line 201: Isn't that equivalent with calling `register_local_tip`? Then we wouldn't need `maybe_new_local_tip`, which seems to duplicate that logic.

node/src/reactor/main_reactor/control.rs
```diff
@@ -371,31 +243,34 @@ impl MainReactor {
...
-                                if let ActivationPoint::Genesis(timestamp) =
-                                    self.chainspec.protocol_config.activation_point
-                                {
-                                    let diff = timestamp.saturating_diff(Timestamp::now());
-                                    if diff > TimeDiff::default() {
+                                if let Some(timestamp) = self.is_genesis() {
+                                    let is_validator = self.should_commit_genesis();
+                                    if false == is_validator {
+                                        return CatchUpInstruction::Fatal(
+                                            "CatchUp: only validating nodes may participate in genesis; cannot proceed without trusted hash".to_string(),
```
afck, line 250: Why not just create the genesis block even if we aren't a validator? Then we can use it as the trusted block.

PR #3375, opened 2022-11-17
fix casper-types build when compiled with std feature

Fraser999:
This PR fixes an issue where `casper-types` fails to compile if the `std` feature is enabled.


afck (approved):

Fraser999:
bors merge

PR #3382, opened 2022-11-21
Various Fixes

EdHastingsCasperAssociation:
* fixed deploy_buffer bug
* fixed deploy_acquisition bug
* log messages & levels tweaks
* reorg'd some types into their own files
* refined DeployAccepted announcement
* added clarifying comments in various places
* added Source::PeerGossiped to differentiate incoming fetched / gossiped messages

### DIFF
--- a/node/src/components/block_synchronizer.rs
+++ b/node/src/components/block_synchronizer.rs
@@ -1,7 +1,9 @@
 mod block_acquisition;
+mod block_acquisition_action;
 mod block_builder;
 mod config;
 mod deploy_acquisition;
+mod error;
 mod event;
 mod execution_results_acquisition;
 mod global_state_synchronizer;
@@ -21,9 +23,10 @@ use casper_execution_engine::core::engine_state;
 use casper_hashing::Digest;
 use casper_types::{EraId, TimeDiff, Timestamp};
 
+use super::network::blocklist::BlocklistJustification;
 use crate::{
     components::{
-        fetcher::{Error, FetchResult, FetchedData},
+        fetcher::{Error as FetcherError, FetchResult, FetchedData},
         Component, ComponentStatus, InitializedComponent, ValidatorBoundComponent,
     },
     effect::{
@@ -43,8 +46,10 @@ use crate::{
     },
     NodeRng,
 };
+
 use block_builder::BlockBuilder;
 pub(crate) use config::Config;
+pub(crate) use error::BlockAcquisitionError;
 pub(crate) use event::Event;
 use execution_results_acquisition::ExecutionResultsAcquisition;
 pub(crate) use execution_results_acquisition::ExecutionResultsChecksum;
@@ -55,8 +60,6 @@ pub(crate) use global_state_synchronizer::{
 pub(crate) use need_next::NeedNext;
 use trie_accumulator::TrieAccumulator;
 pub(crate) use trie_accumulator::{Error as TrieAccumulatorError, Event as TrieAccumulatorEvent};
-
-use super::network::blocklist::BlocklistJustification;
 
 pub(crate) trait ReactorEvent:
     From<FetcherRequest<ApprovalsHashes>>
@@ -513,7 +516,7 @@ impl BlockSynchronizer {
 
     fn block_header_fetched(
         &mut self,
-        result: Result<FetchedData<BlockHeader>, Error<BlockHeader>>,
+        result: Result<FetchedData<BlockHeader>, FetcherError<BlockHeader>>,
     ) {
         let (block_hash, maybe_block_header, maybe_peer_id): (
             BlockHash,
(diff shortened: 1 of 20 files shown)


node/src/types/block.rs
```diff
@@ -641,7 +641,7 @@ impl BlockHash {
 
 impl Display for BlockHash {
     fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
-        write!(formatter, "block hash {}", self.0)
+        write!(formatter, "{}", self.0)
```
afck: I think I've already used that in some other `Display` impls in this file, where I assumed `"{}", hash` would print "block hash …".

node/src/components/deploy_acceptor/tests.rs
```diff
@@ -866,6 +866,12 @@ async fn run_deploy_acceptor_without_timeout(
                         source: Source::Peer(_),
                         ..
                     })
+                ) || matches!(
+                    event,
+                    Event::DeployAcceptorAnnouncement(DeployAcceptorAnnouncement::InvalidDeploy {
+                        source: Source::PeerGossiped(_),
```
afck: I think you could even do this:

```suggestion
                        source: Source::Peer(_) | Source::PeerGossiped(_),
```

node/src/components/block_synchronizer/block_builder.rs
```diff
@@ -405,14 +413,14 @@ impl BlockBuilder {
...
+        acceptance: Result<Option<Acceptance>, BlockAcquisitionError>,
     ) -> Result<(), Error> {
         match acceptance {
-            Ok(Acceptance::NeededIt) => {
+            Ok(Some(Acceptance::NeededIt)) => {
                 self.touch();
                 self.promote_peer(maybe_peer);
             }
-            Ok(Acceptance::HadIt) => (),
+            Ok(Some(Acceptance::HadIt)) | Ok(None) => (),
```
afck, line 443: Isn't `None` unexpected? Should we print an error?

node/src/components/block_synchronizer.rs
```diff
@@ -471,7 +446,7 @@ impl BlockSynchronizer {
                 }
                 NeedNext::EraValidators(era_id) => {
                     warn!(
-                        "block_synchronizer does not have era_validators for era_id: {}",
+                        "BlockSynchronizer: block_synchronizer does not have era_validators for era_id: {}",
```
afck:
```suggestion
                        "BlockSynchronizer: does not have era_validators for era_id: {}",
```

node/src/components/deploy_buffer.rs
```diff
@@ -290,57 +326,64 @@ where
...
+                    for block in blocks {
+                        self.register_block_finalized(&block.into());
+                    }
+                    self.status = ComponentStatus::Initialized;
+                    // start self-expiry management on initialization
+                    effect_builder
+                        .set_timeout(self.cfg.expiry_check_interval().into())
+                        .event(move |_| Event::Expire)
+                } else {
+                    warn!("should not handle this event when component is uninitialized");
```
afck:
```suggestion
                    warn!(?evt, "should not handle this event when component is uninitialized");
```

(Might be useful in case it happens.)

node/src/components/deploy_buffer.rs
```diff
@@ -290,57 +326,64 @@ where
...
-                warn!("should not handle this event when component is uninitialized");
-                Effects::new()
             }
-            (ComponentStatus::Initialized, Event::Initialize(_)) => {
-                error!("deploy buffer already initialized");
-                Effects::new()
-            }
-            (
-                ComponentStatus::Initialized,
+            (ComponentStatus::Initialized, evt) => match evt {
```
afck: I think the `match (&self.status, event)` above should now just be `match self.status`.

```suggestion
            ComponentStatus::Initialized => match event {
```

node/src/components/block_synchronizer/execution_results_acquisition.rs
```diff
@@ -18,16 +22,16 @@ use crate::types::{
 #[derive(Clone, Copy, PartialEq, Eq, DataSize, Debug, Serialize, Deserialize)]
 pub(crate) enum ExecutionResultsChecksum {
     // due to historical reasons, pre-1.5 chunks do not support Merkle proof checking
-    Uncheckable,
+    ApprovalsUncheckable,
     // can be Merkle proof checked
-    Checkable(Digest),
+    ApprovalsCheckable(Digest),
```
afck: It's a bit confusing to name the enum variants `Approvals…`; the fact that approvals and execution results are checkable for the same blocks is kind of accidental, and not relevant in most places where this type is used.

node/src/components/block_synchronizer/block_builder.rs
EdHastingsCasperAssociation: currently Ok(None) is returned in various cases if applying data in a state that doesn't accept that kind of data

afck (approved):

rafal-ch (approved):
LGTM :+1:

PR #3383, opened 2022-11-21
Update status endpoint

### LINKED ISSUE #3374: Update status endpoint
Update `GetStatusResult` as follows:
* move `pub peers: PeersMap,` to be the first field
* move `pub build_version: String,` to immediately follow `api_version`
* `starting_state_root_hash` should be changed to be set to the state root hash of the lowest block of the available block range (comment needs updated and remove deprecation)
* `node_state` should be replaced with `pub reactor_state: ReactorState,`
* add `pub last_progress: Timestamp,`
* add `pub blocks_sync: Vec<BlockSyncStatus>` (see below)
* add `pub available_block_range: AvailableBlockRange,`

```rust
struct BlockSyncStatus {
    block_hash: BlockHash,
    block_height: Option<u64>,
    acquisition_state: String, // BlockAcquisitionState.to_string()
}
```

georgepisaltu:
Fixes #3374 

This PR brings all the changes mentioned in the issue above to the status endpoint.

> Update `GetStatusResult` as follows:
>     - move `pub peers: PeersMap`, to be the first field
>     - move `pub build_version: String`, to immediately follow `api_version`
>     - `starting_state_root_hash` should be changed to be set to the state root hash of the lowest block of the available block range (comment needs updated and remove deprecation)
>     - `node_state` should be replaced with pub `reactor_state: ReactorState`,
>     - add `pub last_progress: Timestamp`,
>     - add `pub blocks_sync: Vec<BlockSyncStatus>` (see below)
>     - add `pub available_block_range: AvailableBlockRange`,

### DIFF
--- a/node/src/types/status_feed.rs
+++ b/node/src/types/status_feed.rs
@@ -7,7 +7,6 @@ use std::{
     time::Duration,
 };
 
-use datasize::DataSize;
 use once_cell::sync::Lazy;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
@@ -17,11 +16,15 @@ use casper_types::{EraId, ProtocolVersion, PublicKey, TimeDiff, Timestamp};
 
 use crate::{
     components::{
+        block_synchronizer::BlockSyncStatus,
         rpc_server::rpcs::docs::{DocExample, DOCS_EXAMPLE_PROTOCOL_VERSION},
         upgrade_watcher::NextUpgrade,
     },
+    reactor::main_reactor::ReactorState,
     types::{ActivationPoint, Block, BlockHash, NodeId, PeersMap},
 };
+
+use super::AvailableBlockRange;
 
 static CHAINSPEC_INFO: Lazy<ChainspecInfo> = Lazy::new(|| {
     let next_upgrade = NextUpgrade::new(
@@ -47,7 +50,11 @@ static GET_STATUS_RESULT: Lazy<GetStatusResult> = Lazy::new(|| {
         round_length: Some(TimeDiff::from(1 << 16)),
         version: crate::VERSION_STRING.as_str(),
         node_uptime: Duration::from_secs(13),
-        node_state: NodeState::Participating,
+        reactor_state: ReactorState::Initialize,
+        last_progress: Timestamp::from(0),
+        available_block_range: AvailableBlockRange::RANGE_0_0,
+        block_sync: vec![],
+        starting_state_root_hash: None,
     };
     GetStatusResult::new(status_feed, DOCS_EXAMPLE_PROTOCOL_VERSION)
 });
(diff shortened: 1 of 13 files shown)


node/src/types/status_feed.rs
```diff
@@ -167,16 +175,17 @@ impl From<Block> for MinimalBlockInfo {
 #[derive(PartialEq, Eq, Serialize, Deserialize, Debug, JsonSchema)]
 #[serde(deny_unknown_fields)]
 pub struct GetStatusResult {
+    /// The node ID and network address of each connected peer.
```
rafal-ch, line 178: Do such changes really matter? Afaik, order of fields in JSON doesn't matter. If we do it for convenience (human-readibility), are we sure that Serde gives us guarantee about order of the serialized fields?

From what I see, this kinda works in REST, but RPC API still returns fields in alphabetical order:

```
{
  "jsonrpc": "2.0",
  "id": "1",
  "result":{
    "api_version": "1.0.0",
    "available_block_range":{
      "high": 0,
      "low": 0
    },
    "block_sync":[],
    "build_version": "1.4.8-1ea1bdc@DEBUG",
     ...and so on...
```

node/src/types/status_feed.rs
```diff
@@ -185,28 +194,35 @@ pub struct GetStatusResult {
...
-            starting_state_root_hash: Digest::from([0u8; 32]),
-            peers: PeersMap::from(status_feed.peers),
+            starting_state_root_hash: status_feed.starting_state_root_hash,
             last_added_block_info: status_feed.last_added_block.map(Into::into),
             our_public_signing_key: status_feed.our_public_signing_key,
             round_length: status_feed.round_length,
             next_upgrade: status_feed.chainspec_info.next_upgrade,
             uptime: status_feed.node_uptime.into(),
-            node_state: status_feed.node_state,
+            reactor_state: status_feed.reactor_state,
```
rafal-ch, line 222: Changelog still mentions `node_state`, but possibly we gonna have a separate ticket to clean-up the changelog. So I guess we can leave it be for now.

resources/test/rest_schema_status.json
```diff
@@ -5,36 +5,44 @@
   "type": "object",
   "required": [
     "api_version",
+    "available_block_range",
```
rafal-ch, line 8: Please also check and update tests in `casper-test`. These two are probably affected:
* `rpc_info_get_status.rs`
* `rpc_list_rpcs.rs`

node/src/types/status_feed.rs
rafal-ch, line 222: Please keep in mind that `casper-test` relies on `node_state` field [here](https://github.com/casper-network/casper-test/blob/master/src/spec/checks.rs#L483).

node/src/types/status_feed.rs
georgepisaltu: Indeed, the order of the fields don't matter in JSON, but it was a requirement in the original ticket. The new JSON schemas are auto-generated, so if it's in alphabetical order in the reference schema, then it's how `schemars` chooses to output it.

node/src/types/status_feed.rs
georgepisaltu: This will need a changelog update, but we can probably include it in the changelog ticket.

georgepisaltu: Yes, `casper-test` will need a PR as well.

resources/test/rest_schema_status.json
georgepisaltu: Will do.

node/src/types/status_feed.rs
rafal-ch: Imho, the change we made doesn't guarantee how the final JSON is rendered, but that's not a blocker.

rafal-ch (approved):
LGTM :+1:

Issue #3359, opened 2022-11-14
Investigate panic in 1.5 after merging in the enhanced diagnostics port


RitaMAllenCA:
After merging the net-info support in diagnostics port into 1.5 we observe a node crashing when trying to inspect unspent allowance. This works correctly on the feat-1.6 branch, so we expect some interference with the async changes made in 1.5. Node panics with: "Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks."

PR #3389, opened 2022-11-21
Fix `debug_inspect_unspent_allowance()`

rafal-ch:
This PR fixes a panic in `debug_inspect_unspent_allowance()` by requesting the data in a separate thread and separate Tokio runtime.

Closes https://github.com/casper-network/casper-node/issues/3359

### DIFF
--- a/node/src/components/network/limiter.rs
+++ b/node/src/components/network/limiter.rs
@@ -6,12 +6,11 @@
 use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, RwLock},
-    thread,
     time::{Duration, Instant},
 };
 
 use prometheus::Counter;
-use tokio::sync::Mutex;
+use tokio::{runtime::Handle, sync::Mutex, task};
 use tracing::{error, trace};
 
 use casper_types::{EraId, PublicKey};
@@ -113,15 +112,9 @@ impl Limiter {
     }
 
     pub(super) fn debug_inspect_unspent_allowance(&self) -> Option<i64> {
-        // Calling `blocking_lock()` from withing an async runtime will cause Tokio to panic with
-        // "Cannot start a runtime from within a runtime..." message. To avoid that we do the
-        // blocking call in a dedicated thread. It's a workaround that has a slight performance
-        // cost, but it's only invoked when using the `net-info` command of the Diagnostics Port,
-        // hence it's assumed to be rare.
-        let resources = Arc::clone(&self.data.resources);
-        thread::spawn(move || Some(resources.blocking_lock().available))
-            .join()
-            .unwrap_or(None)
+        Some(task::block_in_place(move || {
+            Handle::current().block_on(async move { self.data.resources.lock().await.available })
+        }))
     }
 
     pub(super) fn debug_inspect_validators(
--- a/node/src/components/network/limiter.rs
+++ b/node/src/components/network/limiter.rs
@@ -6,11 +6,12 @@
 use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, RwLock},
+    thread,
     time::{Duration, Instant},
 };
 
 use prometheus::Counter;
-use tokio::sync::Mutex;
+use tokio::{runtime::Handle, sync::Mutex};
 use tracing::{error, trace};
 
 use casper_types::{EraId, PublicKey};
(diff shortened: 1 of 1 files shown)


node/src/components/network/limiter.rs
```diff
@@ -112,15 +113,10 @@ impl Limiter {
...
-
-        // node panicked: Cannot start a runtime from within a runtime. This happens because a
-        // function (like `block_on`) attempted to block the current thread while the thread is
-        // being used to drive asynchronous tasks.
-
-        None
-
-        //Some(self.data.resources.blocking_lock().available)
+        let resources = Arc::clone(&self.data.resources);
+        thread::spawn(move || Some(resources.blocking_lock().available))
```
marc-casperlabs: I'm not a fan of this, as it is unclear what the underlying issue is. In other words, if this is necessary to solve the issue, we should consider rearchitecting the entire thing.

 Let's give this another look before taking this measure. If we can't due to time constraints, we'll merge as is.

rafal-ch: Indeed, that's more of a workaround. Let's see how things go.

rafal-ch:
Not merging a workaround since it's not a blocker for the entire diagnostics port.

node/src/components/network/limiter.rs
rafal-ch: Further investigation showed that the original implementation was also panicking. We didn't observe the crash though, because on the `feat-1.6` branch the `Limiter` was implemented as a trait, hence, while testing with NCTL we're actually never been invoking `blocking_lock()` due to the [default implementation](https://github.com/casper-network/casper-node/blob/feat-1.6/node/src/components/small_network/limiter.rs#L56-L58) (NCTL doesn't use [ClassBaseLimiter](https://github.com/casper-network/casper-node/blob/feat-1.6/node/src/components/small_network/limiter.rs#L105)), which contains the [problematic implementation](https://github.com/casper-network/casper-node/blob/feat-1.6/node/src/components/small_network/limiter.rs#L254-L256).

On the `feat-fast-sync-v2` branch, however, the `Limiter` is a struct, hence we've been always invoking the panicking `blocking_lock()`.

The workaround proposed in this PR is again up for the review, with an explanatory comment added in 2a759d5259dc659573a49c6c6a770ccb190873ef.

node/src/components/network/limiter.rs
```diff
@@ -112,15 +113,15 @@ impl Limiter {
...
-        None
-
-        //Some(self.data.resources.blocking_lock().available)
+        // Calling `blocking_lock()` from withing an async runtime will cause Tokio to panic with
+        // "Cannot start a runtime from within a runtime..." message. To avoid that we do the
+        // blocking call in a dedicated thread. It's a workaround that has a slight performance
+        // cost, but it's only invoked when using the `net-info` command of the Diagnostics Port,
+        // hence it's assumed to be rare.
+        let resources = Arc::clone(&self.data.resources);
+        thread::spawn(move || Some(resources.blocking_lock().available))
```
marc-casperlabs: The tokio devs also warn about this construct and mention that the "correct hack" here is to use `block_in_place`  instead of spawning a thread, which is less prone to accidental deadlocks. Let's change it!

marc-casperlabs: To give a little more detail: A deadlock would occur in the following conditions:

1. The lock is held by an async task that then goes to sleep (`tokio::time::sleep`) to wait for enough resources to pile up. It returns and is scheduled behind the current `handle_event`, which is already running.
2. `handle_event` calls the `debug_inspect_unspent_allowance`, which spawns a thread that wants to acquire the lock.
3. The thread is sleeping, waiting for the lock to be released.
4. Meanwhile, `debug_inspect_unspent_allowance` also cannot make any progress, since it is still running the `join()` call. This call will never finish, since on the same thread, the task that would release it is scheduled (and will never run).

This will then deadlock the entire node, as that is the main event processing thread.

marc-casperlabs: In contrast, `block_in_place` solves this situation. From its documentation:

> Calling this function informs the executor that the currently executing task is about to block the thread, so the executor is able to hand off any other tasks it has to a new worker thread before that happens.

Specifically, the task that will ultimately release the lock (see 4. above) will be moved to a different async worker thread and run eventually.

rafal-ch: TIL :+1: 
Thanks for digging into details!

Change to `block_in_place` is here: https://github.com/casper-network/casper-node/pull/3389/commits/a7bfbc96595526db80f7cc5d0b76958740bdd11c

dwerner (approved):
This looks good, as long as it has been/will be tested, I approve.

marc-casperlabs (approved):
LGTM, if it passes testing now :)

PR #3404, opened 2022-11-23
Fix disjoint_sequences test

alsrdn:
`Sequence::New` was removed so some tests are not compiling. This PR adds a small fix for the issue.

### DIFF
--- a/node/src/components/storage/disjoint_sequences.rs
+++ b/node/src/components/storage/disjoint_sequences.rs
@@ -299,6 +299,12 @@ mod tests {
 
     use super::*;
 
+    fn new_sequence(a: u64, b: u64) -> Sequence {
+        let (low, high) = if a <= b { (a, b) } else { (b, a) };
+        assert!(low <= high);
+        Sequence { low, high }
+    }
+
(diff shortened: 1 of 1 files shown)


node/src/components/storage/disjoint_sequences.rs
```diff
@@ -299,6 +299,12 @@ mod tests {
 
     use super::*;
 
+    fn new_sequence(a: u64, b: u64) -> Sequence {
+        let (low, high) = if a <= b { (a, b) } else { (b, a) };
+        assert!(low <= high);
+        Sequence { low, high }
+    }
+
```
rafal-ch, line 307: Minor nit: personally I'd prefer this to be a constructor defined in the `test` module, like so:
```rust
    use super::*;

    impl DisjointSequences {
        fn new(a: u64, b: u64) -> Self {
            let (low, high) = if a <= b { (a, b) } else { (b, a) };
            assert!(low <= high);
            Self {
                sequences: vec![Sequence { low, high }],
            }
        }
    }
```
That could even let us keep the `new_should_order_elements()` test.

But surely not a blocker.

rafal-ch (approved):
LGTM :+1:

EdHastingsCasperAssociation (approved):

PR #3405, opened 2022-11-24
Fix marking blocks complete in storage

### LINKED ISSUE #3388: RPC put_deploy should work the same regardless of reactor state
put_deploy to RPC of any node in CatchUp KeepUp or Validate should work identically

georgepisaltu:
Fixes #3388 

The logic in the block synchronizer and block accumulator was mistakenly marking blocks as complete before they were known to be executed and have the necessary global state. This led to `put-deploy` calls failing when they would request the global state associated with the highest complete block, which sometimes didn't have it.

Additionally, this PR changes `last_added_block_info` in the status RPC call to use the highest complete block rather than the highest block.

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -322,6 +322,7 @@ impl BlockAccumulator {
         effect_builder: EffectBuilder<REv>,
         block: Block,
         sender: Option<NodeId>,
+        executed: bool,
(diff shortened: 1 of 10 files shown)


fizyk20 (approved):

rafal-ch (approved):
LGTM :+1:

PR #3408, opened 2022-11-25
update contract runtime initial state when switching from KeepUp to Validate

rafal-ch:
WIP


rafal-ch:
Closing thanks to the `refresh_contract_runtime` being introduced.

PR #3417, opened 2022-11-29
Change loglevel at runtime

### LINKED ISSUE #3369: Allow for changing the log level at runtime
The diagnostics port should allow changing the log level at runtime. For this, a new command should be added which should be equivalent to setting the loglevel through an ENVVAR. Example:

`set-loglevel INFO,casper_node::components::network=DEBUG`

marc-casperlabs:
Closes #3369.

Allows changing the log level at runtime through the diagnostics port.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,11 @@ being dispatched will be logged as well. Any event has an id (`ev`) and may have
 event whose effects caused the resulting event to be scheduled. As an example, if an incoming network message gets
 asssigned an ID of `ev=123`, the first round of subsequent events will show `a=123` as their ancestor in the logs.
 
+### Changing the logging filter at runtime
+
+If necessary, the filter of a running node can be changed using the diagnostics port, using the `set-log-filter`
+command. See the "Diagnostics port" section for details on how to access it.
+
 ## Debugging
 
 Some additional debug functionality is available, mainly allowed for inspections of the internal event queue.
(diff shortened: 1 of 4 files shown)


rafal-ch (approved):
LGTM :+1: 

I have one nit, which is not related to this PR. Sending a line that cannot be parsed into `Command` (that includes an empty line) results in an `ERR` message, which is not dangerous, but just looks a bit bad :)
```
magister@zacisze:~$ socat - unix:/home/magister/Casper/casper-node/utils/nctl/assets/net-1/nodes/node-1/config/1_0_0/debug.socket

ERR  casper-node 1.4.8
Action to perform

USAGE:
    casper-diagnostics-port <SUBCOMMAND>
    .....
```

```
magister@zacisze:~$ socat - unix:/home/magister/Casper/casper-node/utils/nctl/assets/net-1/nodes/node-1/config/1_0_0/debug.socket

-h
ERR  casper-node 1.4.8
Action to perform

USAGE:
    casper-diagnostics-port <SUBCOMMAND>
    .....
```

I think we could consider improving `process_line()`.

marc-casperlabs:
> I have one nit, which is not related to this PR. Sending a line that cannot be parsed into `Command` (that includes an empty line) results in an `ERR` message, which is not dangerous, but just looks a bit bad :)

That's actually intentional. When using it non-interactively, you will always get back exactly one response indicating the outcome of your operation for every line you sent. I'd rather keep it this way!

The response can be disabled entirely by setting quiet mode.

marc-casperlabs:
bors retry

EdHastingsCasperAssociation (approved):

PR #3418, opened 2022-11-29
Add `ComponentStatus::Paused`

rafal-ch:
This PR adds the new component status: `Paused`. Currently it is used by `BlockSynchronizer` only. When resuming, the previous component state is correctly restored.

It also makes the `BlockSynchronizer` responsive to `BlockSynchronizerRequest::Status` even when paused

Additionally, this PR changes the format of the `block_sync` part of the `/status` response. It is now possible to distinguish between `forward` and `historical` synchronizers, like so:
```json
{
  "peers": [
    {
      "node_id": "tls:4a77..f652",
      "address": "127.0.0.1:22101"
    },
  ],
  "api_version": "1.0.0",
   ... other fields omitted ...
  "block_sync": {
    "historical": {
      "block_hash": "c390ebc5bc747a962fff6609ce4328ee92b682b4e87a7969fa8d38c513a5c940",
      "block_height": 31,
      "acquisition_state": "have strict finality(31) for: block hash c390..c940"
    },
    "forward": {
      "block_hash": "e3480dd433d24b22043992dde4f3a7e365fc7cd0510935c07c1fe10322a13d1f",
      "block_height": 37,
      "acquisition_state": "have block body(37) for: block hash e348..3d1f"
    }
  }
}

```

### DIFF
--- a/node/src/components/block_synchronizer.rs
+++ b/node/src/components/block_synchronizer.rs
@@ -117,6 +117,7 @@ pub struct BlockSyncStatus {
 #[derive(DataSize, Debug)]
 pub(crate) struct BlockSynchronizer {
     status: ComponentStatus,
+    status_before_pause: ComponentStatus,
     validator_matrix: ValidatorMatrix,
     timeout: TimeDiff,
     peer_refresh_interval: TimeDiff,
@@ -135,6 +136,7 @@ impl BlockSynchronizer {
     pub(crate) fn new(config: Config, validator_matrix: ValidatorMatrix) -> Self {
         BlockSynchronizer {
             status: ComponentStatus::Uninitialized,
+            status_before_pause: ComponentStatus::Uninitialized,
             validator_matrix,
             timeout: config.timeout(),
             peer_refresh_interval: config.peer_refresh_interval(),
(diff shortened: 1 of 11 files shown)


node/src/components/block_synchronizer.rs
```diff
@@ -926,13 +989,16 @@ impl<REv: ReactorEvent> Component<REv> for BlockSynchronizer {
...
-        }
-
         // MISSING EVENT: ANNOUNCEMENT OF BAD PEERS
         match (&self.status, event) {
+            (
+                ComponentStatus::Paused,
+                Event::Request(BlockSynchronizerRequest::Status { responder }),
+            ) => responder.respond(self.status()).ignore(),
+            (ComponentStatus::Paused, _) => {
+                warn!("component paused - ignoring event");
```
EdHastingsCasperAssociation, line 999: this should be downgraded to debug! to avoid putting spurious WARNs into the logs when the status endpoints are queried on validators...it isn't actually problematic behavior for the synchronizer to be paused as that is part of its normal lifecycle.

I had already dialed this down on another open PR that hasn't merged yet.

EdHastingsCasperAssociation (approved):
LGTM, with one small nit

PR #3428, opened 2022-12-05
fast-sync-v2 cleanup & sync to ttl

EdHastingsCasperAssociation:
* corrected recent local tip / no trusted hash edge case
* refactored control logic to break up large functions
* sync to ttl feature


node/src/components/block_accumulator.rs
```diff
@@ -172,40 +168,48 @@ impl BlockAccumulator {
         };
 
         if let Some(new_local_tip) = self.maybe_new_local_tip(&starting_with) {
+            let had_no_local_tip = self.local_tip.is_none();
             self.local_tip = Some(new_local_tip);
             info!(local_tip=?self.local_tip, "new local tip detected");
+            if had_no_local_tip {
+                // force a leap when accumulator is starting cold.
+                return SyncInstruction::Leap { block_hash };
```
afck, line 176: Won't this make us leap twice if we're a joining node with empty storage?

node/src/reactor/main_reactor/validate.rs
```diff
@@ -0,0 +1,122 @@
...
+            Some(weights) => weights,
+        };
+        if !highest_era_weights.contains_key(self.consensus.public_key()) {
+            debug!("highest_era_weights does not contain signing_public_key");
+            return Ok(None);
+        }
+
+        if !self
+            .deploy_buffer
+            .have_full_ttl_of_deploys(highest_switch_block_header.height())
```
afck, line 96: (Not introduced in this PR, but:)
It looks like `DeployBuffer::have_full_ttl_of_deploys` uses `now - ttl` instead of `highest_switch_block_header.timestamp() - ttl`. We should make sure we have one TTL worth of blocks, going back from the _start_ of the current era.

node/src/reactor/main_reactor/keep_up.rs
```diff
@@ -0,0 +1,422 @@
...
+        ) {
+            return Ok(SyncBackInstruction::Syncing);
+        }
+
+        if let Some(block_header) = self.storage.get_highest_orphaned_block_header() {
+            if block_header.is_genesis() {
+                return Ok(SyncBackInstruction::GenesisSynced);
+            }
+            if let Some(diff) = self.sync_back_maybe_ttl() {
+                let cutoff = Timestamp::now().saturating_sub(diff);
```
afck: We shouldn't use the current timestamp, but the start time of the current era.
Of course we don't necessarily know that yet; maybe the timestamp of the most recent switch block we know of? And if we don't know any yet, don't start syncing backwards?

node/src/components/block_accumulator.rs
EdHastingsCasperAssociation: shouldn't...but even if it did prompt a leap, the leap status would not be Idle (and thus it would resolve correctly)

node/src/reactor/main_reactor/validate.rs
EdHastingsCasperAssociation: lets address that separately

node/src/reactor/main_reactor/keep_up.rs
EdHastingsCasperAssociation: lets discuss our overall approach to consistent ttl calculation and address on a separate PR

EdHastingsCasperAssociation: Actually, looks like @rafal-ch addressed this side of it in a PR to my branch, which I've merged.

Fraser999 (approved):

Issue #3422, opened 2022-12-02
Fast Sync Metrics - Add block synchronizer metrics


devendran-m:
Add metrics for new fast sync component `block synchronizer`.

PR #3431, opened 2022-12-05
Block synchronizer metrics

alsrdn:
Add block synchronizer time progress metrics.
Addresses https://github.com/casper-network/casper-node/issues/3422

### DIFF
--- a/node/src/components/block_synchronizer/block_builder.rs
+++ b/node/src/components/block_synchronizer/block_builder.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::HashMap,
     fmt::{Display, Formatter},
+    time::Instant,
 };
 
 use datasize::DataSize;
@@ -52,6 +53,7 @@ pub(super) struct BlockBuilder {
     peer_list: PeerList,
 
     // progress tracking
+    sync_start: Instant,
     last_progress: Timestamp,
     in_flight_latch: Option<Timestamp>,
 
@@ -92,6 +94,7 @@ impl BlockBuilder {
             peer_list: PeerList::new(max_simultaneous_peers, peer_refresh_interval),
             should_fetch_execution_state,
             requires_strict_finality,
+            sync_start: Instant::now(),
             last_progress: Timestamp::now(),
             in_flight_latch: None,
         }
@@ -134,6 +137,7 @@ impl BlockBuilder {
             peer_list,
             should_fetch_execution_state,
             requires_strict_finality,
+            sync_start: Instant::now(),
             last_progress: Timestamp::now(),
             in_flight_latch: None,
         }
@@ -169,6 +173,10 @@ impl BlockBuilder {
 
     pub(super) fn should_fetch_execution_state(&self) -> bool {
         self.should_fetch_execution_state
+    }
+
+    pub(super) fn get_sync_start_time(&self) -> Instant {
+        self.sync_start
     }
 
     pub(super) fn last_progress_time(&self) -> Timestamp {
--- a/node/src/components/block_synchronizer/block_builder.rs
+++ b/node/src/components/block_synchronizer/block_builder.rs
@@ -175,7 +175,7 @@ impl BlockBuilder {
         self.should_fetch_execution_state
     }
 
-    pub(super) fn get_sync_start_time(&self) -> Instant {
+    pub(super) fn sync_start_time(&self) -> Instant {
         self.sync_start
     }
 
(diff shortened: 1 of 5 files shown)


node/src/components/block_synchronizer/block_builder.rs
```diff
@@ -171,6 +175,10 @@ impl BlockBuilder {
         self.should_fetch_execution_state
     }
 
+    pub(super) fn get_sync_start_time(&self) -> Instant {
```
Fraser999: We don't normally prefix getters with the term `get_`, but not a big deal.

node/src/components/block_synchronizer/metrics.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use crate::{unregister_metric, utils};
+
+const HIST_SYNC_TIME_NAME: &str = "historical_block_sync_time";
+const HIST_SYNC_TIME_HELP: &str = "Time duration (in sec) to synchronize a historical block";
+const FWD_SYNC_TIME_NAME: &str = "forward_block_sync_time";
+const FWD_SYNC_TIME_HELP: &str = "Time duration (in sec) to synchronize a forward block";
+
+// We use linear buckets to observe the time it takes to synchonize blocks.
+// Buckets have 50 milisec widths and cover up to 500 milisec durations with this granularity.
```
Fraser999:
```suggestion
// We use linear buckets to observe the time it takes to synchronize blocks.
// Buckets have 50 millisec widths and cover up to 500 millisec durations with this granularity.
```

Fraser999 (commented):
A couple of nitpicks, but mainly I'm not sure if the linear buckets up to 500ms will be appropriate for the real-world case (where much larger data sets will be fetched across a real network).  However, we can always adjust that later.

node/src/components/block_synchronizer/metrics.rs
```diff
@@ -0,0 +1,58 @@
+use prometheus::{Histogram, Registry};
+
+use crate::{unregister_metric, utils};
+
+const HIST_SYNC_TIME_NAME: &str = "historical_block_sync_time";
+const HIST_SYNC_TIME_HELP: &str = "Time duration (in sec) to synchronize a historical block";
+const FWD_SYNC_TIME_NAME: &str = "forward_block_sync_time";
+const FWD_SYNC_TIME_HELP: &str = "Time duration (in sec) to synchronize a forward block";
```
Fraser999:
```suggestion
const HIST_SYNC_TIME_HELP: &str = "duration (in sec) to synchronize a historical block";
const FWD_SYNC_TIME_NAME: &str = "forward_block_sync_time";
const FWD_SYNC_TIME_HELP: &str = "duration (in sec) to synchronize a forward block";
```

alsrdn:
> A couple of nitpicks, but mainly I'm not sure if the linear buckets up to 500ms will be appropriate for the real-world case (where much larger data sets will be fetched across a real network). However, we can always adjust that later.

Yeah, my baseline was a local nctl test network. Changed to exponential buckets (in https://github.com/casper-network/casper-node/pull/3431/commits/697cb40c68e51da13d2c87b3210b8de2e91805b4) so we can get better range and also keep good enough resolution in the low end for testing.

Fraser999 (approved):

node/src/components/block_synchronizer/metrics.rs
```diff
@@ -0,0 +1,58 @@
...
+            )?,
+            registry: registry.clone(),
+        })
+    }
+}
+
+impl Drop for Metrics {
+    fn drop(&mut self) {
+        unregister_metric!(self.registry, self.historical_block_sync_time);
+        unregister_metric!(self.registry, self.forward_block_sync_time);
```
rafal-ch, line 56: Just for future reference, it is advised that names of Prometheus metrics contain the unit suffix, for example: `historical_block_sync_time_seconds`.

More info here: https://prometheus.io/docs/practices/naming/

alsrdn: Yup, that's definitely clearer, but it seems that we generally don't follow this rule in other components as well. Maybe it's worth having a pass on all time related metrics and refactor all names and maybe do it in 2.0 just not to create backward compatibility issues.

rafal-ch: I agree, we don't strictly follow the rules. Because, as you noted, changing the name of the metric after release is a breaking change, I suggest using the ~correct~ suggested by prometheus name already at the moment when we introduce a metric.

Also, it's not a big deal, so I leave it up to you.

alsrdn: Yup, I think it's a very good idea. Added https://github.com/casper-network/casper-node/pull/3456 for this, please take a look.

PR #3441, opened 2022-12-06
Fix reactor tests

### LINKED ISSUE #3387: Fix main_reactor tests
we'll need to set min_peers_for_initialization to 0 for any of these which use just a single node

Fraser999:
This PR fixes the remaining reactor unit tests.

In doing so, it exposed a bug in the flow for fetching deploys.  If a non-validating node first received and stored a deploy with a different set of approvals to that which ended up finalized in a block, that node would never store the finalized approvals.

The `BlockSynchronizer` would correctly fetch the approvals hash, but when the `Deploy` was fetched from the peer, the `DeployAcceptor` would not announce a new deploy as it was already stored.

To fix this, I changed the `DeployAcceptor` to recognise this case (a Deploy where the source was `Peer` which had already been stored) and to then store the approvals from that deploy as the finalized approvals, making an announcement if the finalized approvals were new.  This announcement causes the `DeployFetcher` to be re-engaged, allowing the `BlockSynchronizer` to continue its flow.

Closes #3387.


node/src/components/storage.rs
```diff
@@ -2113,7 +2113,7 @@ impl Storage {
         &self,
         deploy_hash: &DeployHash,
         finalized_approvals: &FinalizedApprovals,
-    ) -> Result<(), FatalStorageError> {
+    ) -> Result<bool, FatalStorageError> {
```
rafal-ch, line 2147: Please consider extending the comment to indicate what's the meaning behind this `bool`.

node/src/reactor/main_reactor/tests.rs
```diff
@@ -309,90 +349,120 @@ async fn run_equivocator_network() {
...
 
     let era_count = 4;
 
     let timeout = Duration::from_secs(90 * era_count);
     info!("Waiting for {} eras to end.", era_count);
-    net.settle_on(&mut rng, is_in_era(EraId::new(era_count)), timeout)
-        .await;
+    net.settle_on(
+        &mut rng,
+        has_completed_era(EraId::new(era_count - 1)),
```
rafal-ch, line 409: Is this `- 1` now required?

node/src/reactor/main_reactor/tests.rs
```diff
@@ -309,90 +349,120 @@ async fn run_equivocator_network() {
...
     // (`auction_delay`) era, she is still included in the next era's validator set.
-    assert_eq!(switch_blocks.equivocators(1), [alice_public_key.clone()]);
-    assert_eq!(switch_blocks.inactive_validators(1), []);
-    assert!(bids[1][&alice_public_key].inactive());
+    assert_eq!(
+        switch_blocks.equivocators(1 + offset),
+        [alice_public_key.clone()]
+    );
+    assert_eq!(switch_blocks.inactive_validators(1 + offset), []);
+    assert!(bids[1 + offset as usize][&alice_public_key].inactive());
```
rafal-ch, line 436: I think the comment (`In era 1...`) should get updated as well to accommodate the offset.

rafal-ch (approved):
No blockers, just minor nits :+1:

node/src/reactor/main_reactor/tests.rs
Fraser999: Yes - previously we'd wait until the nodes' consensus components reported they'd started era 4, but now we wait until the nodes' storage components report they've stored the switch block from era 3.  The change was required as not all nodes are necessarily validators and I needed a way to have a similar condition.

node/src/reactor/main_reactor/tests.rs
Fraser999: Well, we have [a long-standing issue](https://github.com/casper-network/casper-node/issues/1859) to fix the non-determinism in this test.  I think it's better to leave the comments as is, since this is the behaviour we want in the test.  The offset should be temporary (terms and conditions apply! :) )

node/src/components/storage.rs
Fraser999: Done in [bc0e8db](https://github.com/casper-network/casper-node/pull/3441/commits/bc0e8db84799164fad1dbd9995d7bda793ff58b5).

PR #3455, opened 2022-12-08
Preserve the order of fields when encoding to json

Fraser999:
This PR enable the `preserve_order` feature of `serde_json`, which causes structs (such as the `GetStatusResult`) to retain the order of the fields when encoding to JSON.


rafal-ch:
TIL :)
We needed that feature here: https://github.com/casper-network/casper-node/pull/3383#discussion_r1028217999

rafal-ch (approved):
LGTM :+1:

PR #3456, opened 2022-12-08
Add unit suffix for newly added metrics

alsrdn:
Since we're adding new metrics, use the suggested unit suffix as described in https://prometheus.io/docs/practices/naming/


rafal-ch (approved):
LGTM :+1:

PR #3469, opened 2022-12-09
Remove cache of recent switch blocks from reactor

Fraser999:
This removes `recent_switch_block_headers` from the main reactor and instead retrieves the required headers from storage as required.

It also closes an outstanding `todo!` comments whereby storage's `read_blocks_since` has been replaced to be less general, and calculates the timestamp itself.

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1182,14 +1182,27 @@ impl Storage {
         Ok(maybe_block)
     }
 
-    pub(crate) fn read_blocks_since(
+    /// Retrieves the contiguous segment of the block chain starting at the highest known switch
+    /// block such that the blocks' timestamps cover a duration of at least the max TTL for deploys
+    /// (a chainspec setting).
+    ///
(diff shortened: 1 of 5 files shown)


EdHastingsCasperAssociation (approved):

node/src/components/storage.rs
```diff
@@ -1182,14 +1182,27 @@ impl Storage {
...
         let mut txn = self
             .env
             .begin_ro_txn()
             .expect("Could not start read only transaction for lmdb");
+        let timestamp = match self.switch_block_era_id_index.keys().last() {
+            Some(era_id) => self
+                .get_switch_block_header_by_era_id(&mut txn, *era_id)?
+                .map(|switch_block| switch_block.timestamp().saturating_sub(self.max_ttl))
+                .unwrap_or_else(Timestamp::now),
+            None => Timestamp::now(),
```
afck, line 1203: ~We could also just return an empty `Vec` in this case? We wouldn't use the returned `Vec` anyway.~

Ah, I see, this is for initialization only; then a best effort makes sense, of course.

PR #3470, opened 2022-12-12
all nodes commit immediate switch block for protocol upgrades

### LINKED ISSUE #3390: All nodes should create immediate switch blocks
Currently, only validating nodes produce immediate switch block at genesis and upgrade, but we have decided that all nodes should produce them when going forward.

Nodes that join after genesis should not produce immediate switch block for genesis and nodes that join after upgrades should not produce immediate switch blocks after upgrade. They will get the via historical sync, as it's already currently does.

EdHastingsCasperAssociation:
As the subject lines says; closes #3390


afck (commented):
Looks good to me, but we should find out what's going on with `is_in_modified_validator_set`, yes.

rafal-ch (approved):
LGTM :+1:

PR #3471, opened 2022-12-12
metrics/contract_runtime: track number of blocks enqueued for execution

alsrdn:
Added the `execution_queue_size` metrics to the contract runtime to monitor the number of blocks enqueued for execution.


node/src/components/contract_runtime.rs
```diff
@@ -238,7 +239,7 @@ where
 impl ContractRuntime {
     /// How many blocks are backed up in the queue
     pub(crate) fn queue_depth(&self) -> usize {
-        self.exec_queue.lock().unwrap().len()
+        self.exec_queue.lock().expect("mutex poisoned").len()
```
Fraser999: +1 for changing to `expect()`, but to make it even more worthwhile, it'd be nice to have a unique error message for each call.

Fraser999 (approved):
LGTM, but with one nitpick.

EdHastingsCasperAssociation (approved):
lgtm

node/src/components/contract_runtime.rs
alsrdn: added more suggestive error messages here: https://github.com/casper-network/casper-node/pull/3471/commits/e0e7449b61d8f0d78e507b31303941aaa2335ac6

node/src/components/contract_runtime.rs
```diff
@@ -480,8 +486,9 @@ impl ContractRuntime {
                         );
                         exec_queue
                             .lock()
-                            .unwrap()
+                            .expect("components::contract_runtime: couldn't enqueue block for execution; mutex poisoned")
                             .insert(finalized_block_height, (finalized_block, deploys));
+                        self.metrics.exec_queue_size.inc();
```
rafal-ch: Technically, `exec_queue` is a `BTreeMap`, hence `insert()` may not actually insert anything, but replace the previous value. So... should this be?:
```suggestion
                        self.metrics.exec_queue_size.set(self.queue_depth() as i64);
```

alsrdn: I guess it would always insert in this case since it has a height bigger than anything it found so far, but I agree to make it a bit more robust and update based on the true size once. Fixed in https://github.com/casper-network/casper-node/pull/3471/commits/dd63ed6169d291f65297d53fc760f5bf086eabe7

rafal-ch (approved):
LGTM :+1:

PR #3472, opened 2022-12-12
Fix setting disjoint sequences during storage construction

Fraser999:
This PR causes the disjoint sequences in storage to be initialised with a single range from 0 to the highest stored block of the previous protocol version.

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -72,10 +72,6 @@ use casper_types::{
     bytesrepr::{FromBytes, ToBytes},
     EraId, ExecutionResult, ProtocolVersion, PublicKey, TimeDiff, Timestamp, Transfer, Transform,
 };
-use disjoint_sequences::DisjointSequences;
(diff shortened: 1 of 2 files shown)


node/src/components/storage.rs
```diff
@@ -477,18 +475,47 @@ impl Storage {
...
+                }
 
-            component.completed_blocks = sequences;
+                component.completed_blocks = sequences;
+            }
+            None => {
+                // No state so far. We can make the following observations:
+                //
+                // 1. Any block already in storage from versions prior to 1.5 (no fast-sync) MUST
+                //    have the corresponding global state in contract runtime due to the way sync
```
afck, line 494: Edit: Sorry, I misread!

```suggestion
                //    has the corresponding global state in contract runtime due to the way sync
```

(Or "blocks" above.)

EdHastingsCasperAssociation (approved):

afck (approved):

PR #3475, opened 2022-12-12
NCTL: client test tweaks

TomVasile:
Changed:
- switched how we derive the subcommands available to the client
- switched how we derive the available shells for shell completion generation


utils/nctl/sh/scenarios/client.sh
```diff
@@ -1317,11 +1315,12 @@ function get_generate_completion_shells() {
 function get_client_subcommand_list() {
     local OUTPUT
     OUTPUT=$($(get_path_to_client) --help \
-        | grep -A 999 SUBCOMMANDS \
-        | awk -F'    ' '{print $2}' \
+        | awk -v RS= 'NR==3' \
+        | sed -e 's/^[ \t]*//' \
```
Fraser999: The latest client code in `release-2.0.0` breaks the help lines at 100 chars if the terminal width can't be determined.  So it seems like this change below strips such lines from the output, as they all have more than 2 leading spaces:
```suggestion
        | sed -e '/^[ \t]\{3,\}/d' \
```

TomVasile: updated

Fraser999 (approved):

PR #3476, opened 2022-12-13
Fix misleading participation log message.

afck:
If participation is logged after the DAG has been purged ("evidence only" mode) it looks like everyone was inactive, so we should skip the log in that case.

This also changes the testing config so that participation is logged every 15 seconds instead of 1 minute, otherwise we would never see the message in the 21 second testing eras.


rafal-ch (approved):
LGTM :+1:

afck:
bors merge

afck:
bors merge

fizyk20 (approved):

afck:
bors merge

afck:
bors merge

PR #3477, opened 2022-12-13
Fix deploy buffer TTL check.

afck:
To prevent and detect replays in an era we need to know about all deploys up to the era start time, starting one max TTL before that.

### DIFF
--- a/node/src/components/deploy_buffer.rs
+++ b/node/src/components/deploy_buffer.rs
@@ -100,7 +100,7 @@ impl DeployBuffer {
     /// which is the switch block's timestamp. A deploy D in a block B cannot have timed out as of
     /// B's timestamp, so it cannot be older than T - ttl, where ttl is the maximum deploy
     /// time-to-live. So if D were also in an earlier block C, then C's timestamp would be at
-    /// least T - ttl. Thus to prevent replays, we need all blocks with a timetstamp between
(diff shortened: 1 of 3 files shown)


afck:
bors try

node/src/components/deploy_buffer.rs
```diff
@@ -92,10 +92,28 @@ impl DeployBuffer {
...
-        let ttl = self.deploy_config.max_ttl;
+    /// Returns `true` if we have enough information to participate in consensus in the era after
+    /// the `switch block`.
+    ///
+    /// To validate and propose blocks we need to be able to avoid and detect deploy replays.
+    /// Each block in the new era E will have a timestamp greater or equal to E's start time T,
+    /// which is the switch block's timestamp. A deploy D in a block B cannot have timed out as of
+    /// B's timestamp, so it cannot be older than T - ttl, where ttl is the maximum deploy
+    /// time-to-live. So if D were also in an earlier block C, then C's timestamp would be at
+    /// least T - ttl. Thus to prevent replays, we need all blocks with a timetstamp between
```
rafal-ch:
```suggestion
    /// least T - ttl. Thus to prevent replays, we need all blocks with a timestamp between
```

rafal-ch (approved):
LGTM :+1: 
with one typo

Fraser999 (approved):

PR #3478, opened 2022-12-13
Implement force resync

### LINKED ISSUE #3460: Add force-resync behaviour
The node's config file should have a new bool flag added to the `node` section named `force_resync` defaulted to `false` in the config.

When `true`, during construction of the storage component, the `completed_blocks` field should be set to `default()` and persisted via `persist_completed_blocks`.  The intent is that the historical sync process will now cause the node to resync all existing data, (picking up any missing finality signatures on the way).

We also need a mechanism to ensure the process can be stopped and restarted correctly.  This should be in the form of an empty marker file named `force_resync` written to the storage folder as soon as the call to `persist_completed_blocks` has executed in the storage constructor.  If the file exists on startup, storage will behave as if the `force_resync` config option is `false`.

georgepisaltu:
Fixes #3460 

This PR adds the option to enable `force-resync` in the node config, which will reset the value tracking completed blocks in storage, causing the node to start syncing all of the blocks again. This will create a `force-resync.pid` pid file in storage which, while present, will disable this behavior on subsequent node restarts to prevent accidentally forcing a resync on a node crash.


resources/local/config.toml
```diff
@@ -38,6 +38,9 @@ max_attempts = 3
 # Default delay for the control events that have no dedicated delay requirements.
 control_logic_default_delay = '1sec'
 
+# Flag which forces the node to resync all of the blocks.
+force_resync = false
+
```
rafal-ch, line 43: Please also add this to `config-example.toml`.

node/src/components/storage.rs
```diff
@@ -473,8 +478,46 @@ impl Storage {
...
+            // Check if resync is already in progress. Force resync will kick
+            // in only when the pid file was cleanly initialized and acquired
+            // (i.e. it didn't exist before).
+            match PidFile::acquire(component.root_path().join(FORCE_RESYNC_PID_FILE_NAME)) {
+                utils::pid_file::PidFileOutcome::Clean(pid_file) => {
+                    // When no pid file was detected, initialize force resync.
+                    info!("Initializing force resync.");
+                    // Default `storage.completed_blocks`.
+                    component.completed_blocks = Default::default();
+                    component.persist_completed_blocks()?;
```
rafal-ch, line 500: Is `PidFile` able to clean itself up? Or shall we delete the `pid_file` here to allow retry on next node startup?

node/src/components/storage.rs
```diff
@@ -473,8 +478,46 @@ impl Storage {
             fault_tolerance_fraction,
             max_ttl,
             metrics,
+            force_resync_pid_file: None,
         };
 
+        if force_resync {
```
rafal-ch, line 483: Can we avoid running "indexing block store", "initialize_block_body_db", etc. in case `force_resync` is set to `true`? (and if `persist_completed_blocks()` did not fail?)

node/src/components/storage.rs
```diff
@@ -1116,6 +1159,29 @@ impl Storage {
...
+                            .block_height_index
+                            .keys()
+                            .last()
+                            .copied()
+                            .unwrap_or_default()
+            })
+            .unwrap_or_default()
+        {
+            // Notify resync is complete.
+        }
```
rafal-ch: This does nothing, right? Shall we add a code to notify?

However, I'm not sure why we need additional notification flow. My assumption is that our current mechanisms should properly handle the state changes according to the content of a block store and select correct KeepUp/CatchUp/Validate state as the blocks are retrieved.

node/src/utils.rs
```diff
@@ -7,6 +7,7 @@ pub(crate) mod ds;
 mod external;
 pub(crate) mod fmt_limit;
 pub(crate) mod opt_display;
+pub(crate) mod pid_file;
```
rafal-ch: Just a question, because I'm not fan o pid files :)
Would it be possible to keep a state in the local storage?

node/src/components/storage.rs
georgepisaltu: I don't think we can. What this setting is supposed to do is just clear the `completed_blocks` index, which will cause the node to sync again gathering any missing information, but I think we're supposed to keep the stuff we have to have it function properly and efficiently.

node/src/components/storage.rs
georgepisaltu: This `PidFile` is not supposed to clean itself up, it should be persistent. Intended use case is that operators will delete this file if they want to enable the `force-resync` behavior repeatedly.

node/src/components/storage.rs
georgepisaltu: It does nothing indeed. This is a remnant of a previous approach and I forgot to delete it. Right now, there is no point in notifying if a force resync has completed.

node/src/utils.rs
georgepisaltu: It would be possible. However, we went with the pid file approach because it is more straightforward for operators to interact with this switch. If we went with the storage approach, we would have to provide dedicated tooling in order to interact with this switch, which has merit but wasn't something we were willing to do right now.

resources/local/config.toml
georgepisaltu: Done.

node/src/components/storage.rs
```diff
@@ -127,6 +127,8 @@ const DEFAULT_MAX_STATE_STORE_SIZE: usize = 10 * GIB;
 const MAX_DB_COUNT: u32 = 9;
 /// Key under which completed blocks are to be stored.
 const COMPLETED_BLOCKS_STORAGE_KEY: &[u8] = b"completed_blocks_disjoint_sequences";
+/// Key under which completed blocks are to be stored.
```
Fraser999: Copy-paste error

node/src/components/storage.rs
```diff
@@ -208,6 +210,8 @@ pub struct Storage {
     metrics: Option<Metrics>,
     /// The maximum TTL of a deploy.
     max_ttl: TimeDiff,
+    #[data_size(skip)]
+    force_resync_pid_file: Option<PidFile>,
```
Fraser999: I don't think we need this member variable - since we don't need to report it (e.g. via status endpoints) it's only used in the storage constructor.

node/src/components/storage.rs
```diff
@@ -473,8 +478,46 @@ impl Storage {
...
         };
 
+        if force_resync {
+            // Check if resync is already in progress. Force resync will kick
+            // in only when the pid file was cleanly initialized and acquired
+            // (i.e. it didn't exist before).
+            match PidFile::acquire(component.root_path().join(FORCE_RESYNC_PID_FILE_NAME)) {
+                utils::pid_file::PidFileOutcome::Clean(pid_file) => {
+                    // When no pid file was detected, initialize force resync.
+                    info!("Initializing force resync.");
```
```suggestion
                    info!("initializing force resync");
```

node/src/utils/pid_file.rs
```diff
@@ -0,0 +1,246 @@
+//! PidFile utility type.
```
Fraser999: I think we should avoid re-introducing the PidFile as it's overkill for this task.  We only need an empty file named `force_resync` (without the `.pid` suffix).

node/src/components/storage.rs
rafal-ch: So the `force_resync` is not intended to resync from scratch, but just to fill potentially missing parts?

node/src/components/storage.rs
georgepisaltu: Fixed.

node/src/components/storage.rs
georgepisaltu: Would have been useful to keep a lock on the pid file, but since we went with regular files, it now serves no purpose so I removed it.

node/src/components/storage.rs
georgepisaltu: As far as I understand, an operator turns on `force_resync` to get any information about the state of the chain, such as additional finality signatures. It is not intended to wipe anything from the database except the `completed_blocks` index, which is just an indication of what blocks are considered "good enough" to not request additional information about them.

node/src/components/storage.rs
georgepisaltu: Done.

node/src/utils.rs
georgepisaltu: After further discussion, we decided to use a regular file instead of a pid file as it meets our requirements and is simpler to implement and use.

node/src/utils/pid_file.rs
georgepisaltu: Done.

node/src/components/storage.rs
rafal-ch: If that's the case, the code as is makes perfect sense, thanks for explanation.

So it's just the parameter name that could be confusing, because it may suggest that we wipe the data and resync. But that's just a minor nit and we can leave it be.

node/src/components/storage.rs
```diff
@@ -475,6 +480,40 @@ impl Storage {
...
+                .write(true)
+                .open(&force_resync_file_path)
+            {
+                Ok(_file) => {
+                    // When the force resync marker file was not present and
+                    // is now created, initialize force resync.
+                    info!("initializing force resync");
+                    // Default `storage.completed_blocks`.
+                    component.completed_blocks = Default::default();
+                    component.persist_completed_blocks()?;
```
Fraser999, line 500: By creating the file first, we have a race condition where the attempt to persist the defaulted sequences could fail (node exits) and force-resync would hence not be re-attempted on the next run.

If we just check for the file's existence first, and only create the file once we successfully persist the defaulted sequences, we still have a race condition, but it's a little safer I think (another process could create the file after we check, but before we create it).

Probably not a big deal, both scenarios seem unlikely, and can be resolved by the operator if required.

Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

node/src/components/storage.rs
georgepisaltu: At this point, we're in the initialization phase of the storage component and the reactor itself. If something fails here (like `persist_completed_blocks`) and the node crashes, the operators have to manage the storage folder manually anyway. Given that this scenario is already highly unlikely, I'll leave it as is.

georgepisaltu (commented):
bors r+

PR #3483, opened 2022-12-14
Never overwrite DeployBuffer hold entries.

### LINKED ISSUE #3490: The deploy buffer must never overwrite entries in hold.
If another block with the same timestamp has been proposed before we'd remove its deploys from hold here. deploy_buffer.rs#L361

afck:
It's an edge case but it can happen that we propose a block with a timestamp of another block that is already proposed. With Highway this is probably only possible if we run two nodes with the same secret key anyway (so we'd equivocate). But with Zug the previous round leader's clock could be slightly out of sync with ours, and they could have made a proposal with the same timestamp.

Either way, it's safer if the deploy buffer keeps track of all `hold` entries.

Closes #3490.

### DIFF
--- a/node/src/components/deploy_buffer.rs
+++ b/node/src/components/deploy_buffer.rs
@@ -3,7 +3,7 @@ mod event;
 mod metrics;
 
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet},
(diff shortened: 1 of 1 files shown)


alsrdn (approved):
LGTM

afck:
bors merge

rafal-ch (approved):
LGTM :+1:

PR #3484, opened 2022-12-14
Pass correct era id to `ee_upgrade_config()` while committing an upgrade

rafal-ch:
This PR changes the era id which is passed to `ee_upgrade_config()` while committing an upgrade.


fizyk20 (approved):
I'm approving, but someone not involved in creating this fix should probably look at it before we merge it.

afck (approved):

afck:
(I was also in the meeting, so some third reviewer would still be good.)

Fraser999 (approved):

PR #3486, opened 2022-12-14
Don't compute sync leap ancestor limit from chainspec.

afck:
This causes inconsistencies across update boundaries. Instead, just verify that the ancestors are in reverse chronological order and contain exactly one switch block, as the last entry.

### DIFF
--- a/node/src/types/chainspec.rs
+++ b/node/src/types/chainspec.rs
@@ -173,23 +173,6 @@ impl Chainspec {
         ))
     }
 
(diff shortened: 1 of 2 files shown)


fizyk20 (approved):

rafal-ch (approved):
LGTM :+1: 
And nodes are no longer banning each other when sync-leaping after an upgrade

afck:
bors merge

Issue #3458, opened 2022-12-08
Chase down why we make a block complete twice in quick succession


piotr-dziubecki:
should only do it once

alsrdn:
Will be addressed as part of https://github.com/casper-network/casper-node/issues/3520.

PR #3487, opened 2022-12-14
Don't mark block complete twice

alsrdn:
Don't mark the block complete from the Block Accumulator component.
This fixes: https://github.com/casper-network/casper-node/issues/3458

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -404,11 +404,9 @@ impl BlockAccumulator {
         };
 
         match acceptor.register_finality_signature(finality_signature, sender) {
-            Ok(Some(finality_signature)) => self.store_block_and_finality_signatures(
-                effect_builder,
-                ShouldStore::SingleSignature(finality_signature),
-                None,
-            ),
+            Ok(Some(finality_signature)) => {
+                Self::store_finality_signature(effect_builder, finality_signature)
+            }
             Ok(None) => match self.validator_matrix.validator_weights(era_id) {
                 Some(evw) => {
                     let (should_store, faulty_senders) = acceptor.should_store_block(&evw);
@@ -609,6 +607,22 @@ impl BlockAccumulator {
             .set(self.block_children.len().try_into().unwrap_or(i64::MIN));
     }
 
+    fn store_finality_signature<REv>(
+        effect_builder: EffectBuilder<REv>,
+        signature: FinalitySignature,
+    ) -> Effects<Event>
+    where
+        REv: From<StorageRequest> + Send,
+    {
+        debug!(%signature, "storing finality signature");
+        effect_builder
+            .put_finality_signature_to_storage(signature.clone())
+            .event(move |_| Event::Stored {
+                block: None,
+                finality_signatures: vec![signature],
+            })
+    }
+
     fn store_block_and_finality_signatures<REv, I>(
         &mut self,
         effect_builder: EffectBuilder<REv>,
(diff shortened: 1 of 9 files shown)


node/src/reactor/main_reactor.rs
```diff
@@ -976,15 +977,26 @@ impl reactor::Reactor for MainReactor {
...
+                    let block_height = block.header().height();
+                    let block = block.clone();
+                    effects.extend(
+                        effect_builder
+                            .put_executed_block_to_storage(
+                                block.clone(),
+                                approvals_hashes,
+                                execution_results_map,
+                            )
+                            .then(move |_| {
```
afck, line 990: What type is the `_` here? I'm just wondering if the block gets passed back to us anyway, so we don't have to clone it?

node/src/components/storage/tests.rs
```diff
@@ -378,15 +381,25 @@ fn get_highest_complete_block_header(
     response
 }
 
-/// Stores a block in a storage component.
+/// Stores a block in a storage component and mark it as complete.
```
```suggestion
/// Stores a block in a storage component and marks it as complete.
```

node/src/components/storage/tests.rs
```diff
@@ -378,15 +381,25 @@ fn get_highest_complete_block_header(
...
     block: Box<Block>,
 ) -> bool {
+    let block_height = block.height();
     let response = harness.send_request(storage, move |responder| {
-        StorageRequest::PutCompleteBlock { block, responder }.into()
+        StorageRequest::PutBlock { block, responder }.into()
     });
+    if response == true {
+        harness.send_request(storage, move |responder| {
+            BlockCompleteConfirmationRequest {
```
afck, line 396: Is that something specific to these tests?

afck (approved):
Makes sense to me.
Trying to remember why `put_complete_block_to_storage` was added in the first place. Was it just because the synchronizer didn't mark them as complete yet at that time? Maybe another pair of reviewer eyes would be good.

node/src/components/block_accumulator.rs
```diff
@@ -323,7 +322,7 @@ impl BlockAccumulator {
             Some(acceptor) => acceptor,
         };
 
-        match acceptor.register_block(block, sender, executed) {
+        match acceptor.register_block(block, sender) {
```
rafal-ch, line 325: This reverts (at least partially) the changes introduced here: https://github.com/casper-network/casper-node/pull/3405

Are we sure that `put-deploy` still [works as expected](https://github.com/casper-network/casper-node/issues/3388)?

node/src/reactor/main_reactor.rs
rafal-ch: I think it's just `bool` as returned from `put_executed_block_to_storage()`

alsrdn: Yes, unfortunately it only gives us back the bool so we need to clone the block.

node/src/components/storage/tests.rs
alsrdn: Not sure if it really needs to mark them complete to be honest. This was changed in https://github.com/casper-network/casper-node/commit/503c4d00fa019f90347a926c2d77c02035e93e3a so I wanted to preserve the same semantics due to that change.

node/src/components/block_accumulator.rs
alsrdn: Yes, `put-deploy` still works for me (testing with `nctl-transfer` on joining nodes.
There have been a number of changes since that PR. After that PR, joining nodes were not executing blocks at all. This behavior changed since https://github.com/casper-network/casper-node/commit/f387f2389d27c4f32ac6b1ad92cdab951eaded5c and blocks were being marked complete after block execution or after historical sync was completed. Essentially, the block_accumulator was marking block complete because it was told that a block was executed. But it was already marked for completion, making the call in the block_accumulator redundant.

rafal-ch (approved):
LGTM :+1:

rafal-ch:
Applied the `do-not-merge` as we need to run some tests here.

node/src/components/storage/tests.rs
alsrdn: fixed in https://github.com/casper-network/casper-node/pull/3487/commits/9d689b2485d2024b664e12bbca632db8200d36a8

alsrdn:
will be addressed by https://github.com/casper-network/casper-node/issues/3520

PR #3491, opened 2022-12-14
nctl: rework ejection health check

TomVasile:
Changes:
- switched out the way we detect ejected nodes from the logs
    - the log message no longer contains `.fields.finalized_block` so it was switched to derive it from `.fields.participation`


rafal-ch (approved):
LGTM :+1: 

Should work as expected, even though `log_participation()` is based on a timer (`log_participation_interval = '1min'`), whereas `finalized_block` was logged immediately.

Please double-check with @fizyk20 maybe?

TomVasile:
@fizyk20 Do you know of a better way I can check for ejected validators based on the logs? I'm open to reworking this if needed.

Maybe we log this out specifically to make it easier to check?

TomVasile:
labeling as do not merge, in standup we decided to log out the ejections when the switch block hits, will use that once merged

PR #3499, opened 2022-12-16
Controlled shutdowns

### LINKED ISSUE #3361: Support Controlled Shutdowns
We want to give the ability to shutdown the node at a clean stopping point to facilitate things like key swaps, database snapshot and so on. This requires a reactor-global variable similar to how signals are processed currently (or in an appropriate component that checks these things), causing the node to shut down cleanly on a specific boundary.
With this, we add a stop-at command via debug console, with arguments like stop --at=block:3, stop --at=era:12, stop --at=block:next or --stop --at=era:next (omitting the --at should still stop immediately).
To make things a little cleaner, SIGERM should perform the same action as stop --at=block:next the first time it is received. The second time, it should be equivalent to stop without any --at parameter.
There should be a way to cancel a shutdown, e.g. with a stop --cancel command.
Our official releases should also account for this in the shipped systemd unit file (bigger timeouts and/or making a second SIGTERM the kill signal).

marc-casperlabs:
Closes #3361 .

This PR adds shutdown capabilities to the diagnostics port by adding a new component named `ShutdownTrigger`.

The `ShutdownTrigger` component is notified of each new block and keeps track of the highest block it has seen, as it will ignore all older blocks. Should it receive an up-to-date block, it will check if it fulfills the shutdown criteria set through a potential `StopAtSpec`. If so, it instructs the node to shutdown via `ControlAnnouncement`.

To play with this in `nctl`, simply run `nctl dport`, `stop --at era:5`, for example.


node/src/components/diagnostics_port/command.rs
```diff
@@ -88,6 +92,21 @@ pub(super) enum Action {
...
     NetInfo,
+    /// Stop the node at a certain condition.
+    Stop {
+        /// When to stop the node.
+        ///
+        /// Supports `block:12345` for block height, `era:123` for eras, `block:next` / `era:end`
+        /// to stop on an upcoming block or switch block, or `now` to stop immediately. Defaults to
+        /// `block:next`."
+        ///
+        /// Returns the previouslys set stopping point, if any.
```
Fraser999:
```suggestion
        /// Returns the previously set stopping point, if any.
```

node/src/components/shutdown_trigger.rs
```diff
@@ -0,0 +1,156 @@
+//! Shutdown trigger control.
+//!
+//! A component that can be primed with a [`StopAtSpec`] and will monitor to the node, until it
+//! detects a specific spec has been triggered. If so, it instructs the system to shut down through
+//!  a [`ControlAnnouncement`].
```
Fraser999:
```suggestion
//! A component that can be primed with a [`StopAtSpec`] and will monitor the node, until it
//! detects a specific spec has been triggered. If so, it instructs the system to shut down through
//! a [`ControlAnnouncement`].
```

node/src/components/shutdown_trigger.rs
```diff
@@ -0,0 +1,156 @@
...
+                } else {
+                    trace!("received block, but no active stop-at spec, ignoring");
+                    return Effects::new();
+                };
+
+                let should_shutdown = match active_spec {
+                    StopAtSpec::BlockHeight(trigger_height) => block.height() >= trigger_height,
+                    StopAtSpec::EraId(trigger_era_id) => block.header().era_id() >= trigger_era_id,
+                    StopAtSpec::Immediately => {
+                        // Immediate stops are handled when the event is received.
```
Fraser999:
```suggestion
                        // Immediate stops are handled when the request is received.
```

node/src/components/shutdown_trigger.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+
+            Event::SetNodeStopRequest(SetNodeStopRequest {
+                mut stop_at,
+                responder,
+            }) => {
+                mem::swap(&mut self.active_spec, &mut stop_at);
+
+                let mut effects = Effects::new();
+                effects.extend(responder.respond(stop_at).ignore());
```
Fraser999:
```suggestion
                let mut effects = responder.respond(stop_at).ignore();
```

Fraser999 (commented):
LGTM (bar a few nitpicks :) )

Fraser999 (approved):

PR #3511, opened 2022-12-20
Fix or remove remaining todos

Fraser999:
This PR removes the remaining `todo!` comments from the codebase.  Of these, two remain unresolved but have been ticketed for fixing.

It also fixes an issue in the block synchronizer where an error when applying an `ExecutionResultOrChunk` was being silently ignored.


node/src/components/storage.rs
```diff
@@ -1269,8 +1269,6 @@ impl Storage {
                 {
                     continue;
                 }
-                // todo!() - right deploy with incorrect approvals found in DB, download and
```
alsrdn, line 1311: Wouldn't we still want to download the correct approvals in this case?

alsrdn (approved):
LGTM

node/src/components/storage.rs
Fraser999: I think that's what would happen when the node falls back to leaping and gets this block via historical sync.

Fraser999:
bors merge

Fraser999:
bors merge

Fraser999:
bors merge

Fraser999:
bors try

Fraser999:
bors try

Fraser999:
bors merge

PR #3523, opened 2022-12-22
Cleanups

Fraser999:
This PR removes some unused files and types.  It also removes `impl From<u64> for TimeDiff` which is somewhat dangerous as it is not clear that the `u64` represents milliseconds.  `TimeDiff` already provides safer alternatives in the form of `from_millis` and `from_seconds`.


fizyk20 (approved):

Fraser999:
bors merge

PR #3524, opened 2022-12-22
Fix deploy buffer issue to allow blocks to be fully filled

### LINKED ISSUE #3502: Investigate "unused block capacity"
after upgrade blocks are not fully saturated (probably not a bug, but let's track it until we confirm)

Fraser999:
This PR fixes an issue where the deploy buffer would exit early when creating a new `AppendableBlock`, causing blocks to be not filled in terms of the deploy and transfer count.

Closes #3502.


fizyk20 (approved):

Fraser999:
bors merge

PR #3528, opened 2022-12-22
Limit entries in validator matrix

### LINKED ISSUE #3515: Limit entries in validator matrix
The validator matrix should be changed so that when new era weights are added, it retains only the three highest and three lowest eras' weights

georgepisaltu:
Fixes #3515 

This PR now limits the amount of entries in `ValidatorMatrix` to 6, of which 3 slots are reserved for the lowest known eras and the other 3 for the highest known eras. Trying to register an era in between will not change the state of the structure.

### DIFF
--- a/node/src/types/validator_matrix.rs
+++ b/node/src/types/validator_matrix.rs
@@ -15,6 +15,8 @@ use casper_types::{EraId, PublicKey, SecretKey, U512};
 
 use super::{BlockHeader, FinalitySignature};
 
+const MAX_VALIDATOR_MATRIX_ENTRIES: usize = 6;
+
 #[derive(Eq, PartialEq, Debug, Copy, Clone)]
 pub(crate) enum SignatureWeight {
     /// Too few signatures to make any guarantees about the block's finality.
@@ -77,11 +79,22 @@ impl ValidatorMatrix {
         validators: EraValidatorWeights,
     ) -> bool {
         let era_id = validators.era_id;
-        self.inner
+        let mut guard = self
+            .inner
             .write()
-            .unwrap()
-            .insert(era_id, validators)
-            .is_none()
+            .expect("poisoned lock on validator matrix");
+        let was_present = guard.insert(era_id, validators).is_some();
+        if guard.len() > MAX_VALIDATOR_MATRIX_ENTRIES {
+            // Safe to unwrap because we check above that we have sufficient entries.
+            let median_key = guard
+                .keys()
+                .nth(MAX_VALIDATOR_MATRIX_ENTRIES / 2)
+                .copied()
+                .unwrap();
+            guard.remove(&median_key);
+            return median_key != era_id;
+        }
+        !was_present
     }
 
     pub(crate) fn register_validator_weights(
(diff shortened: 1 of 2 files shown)


Fraser999 (approved):

alsrdn (approved):
LGTM

node/src/types/validator_matrix.rs
```diff
@@ -279,3 +292,78 @@ impl EraValidatorWeights {
...
+        // Now that we have 6 entries in the validator matrix, try adding more.
+        // We should have an entry for era 3 (we have eras 0 through 5
+        // inclusive).
+        assert!(validator_matrix.has_era(&(MAX_VALIDATOR_MATRIX_ENTRIES as u64 / 2).into()));
+        // Add era 7.
+        era_validator_weights.push(empty_era_validator_weights(
+            (MAX_VALIDATOR_MATRIX_ENTRIES as u64 + 1).into(),
+        ));
+        // Now the entry for era 3 should be dropped and we should be left with
+        // the 3 lowest eras [0, 1, 2] and 3 highest eras [4, 5, 6].
```
rafal-ch, line 347: I'd expect to have [0, 1, 2, 4, 5, **7**] here.

Or did you mean to add Era 6 [here](https://github.com/casper-network/casper-node/pull/3528/files#diff-fc4d1befdc75c72b75fe10cb728dc49dd320576ab84a56971e007667599c2cacR344).

node/src/types/validator_matrix.rs
```diff
@@ -279,3 +292,78 @@ impl EraValidatorWeights {
...
+        assert!(validator_matrix
+            .register_era_validator_weights(era_validator_weights.last().cloned().unwrap()));
+        assert!(!validator_matrix.has_era(&(MAX_VALIDATOR_MATRIX_ENTRIES as u64 / 2).into()));
+        assert_eq!(
+            validator_matrix.read_inner().len(),
+            MAX_VALIDATOR_MATRIX_ENTRIES
+        );
```
rafal-ch, line 354: I'd much more prefer explicit asserts for improved clarity, for example:
```
        assert_eq!(
            vec![0u64, 1, 2, 4, 5, 7],
            validator_matrix
                .read_inner()
                .keys()
                .copied()
                .map(EraId::value)
                .collect_vec()
        );
```

node/src/types/validator_matrix.rs
georgepisaltu: I meant to add the 7th entry, so that would be era `6` if it was continuous, could be era `7` or any greater number for the purpose of the test. To keep it consistent, I'll make it era `7`, which would be the 7th entry.

node/src/types/validator_matrix.rs
georgepisaltu: I think it's a good idea. I'll add these asserts in addition to what we already have in a future PR, if that's ok with you.

georgepisaltu: Done in https://github.com/casper-network/casper-node/pull/3539

PR #3537, opened 2022-12-27
Moving assemblyscript to end of tag for no-version change fails.

sacherjj:
Put AS failure at the end when we don't want to publish.


TomVasile (approved):

sacherjj:
bors merge

PR #3540, opened 2023-01-03
Network watchdog

### LINKED ISSUE #3530: Add keep-alive watchdog (ping/pong) in networking for resilience
To improve reliability and catch potentially stale connections early (as opposed to when we actually want to send data to them), add a periodic ping/pong message being sent to peers.
When failing to respond to a ping in a timely manner, terminate the connection (but do not block the peer).

marc-casperlabs:
Closes #3530.

This PR adds a network watchdog in the form of a ping/pong functionality:

* Nodes will periodically send a `Ping`  down every outgoing connection.
* Any node receiving a `Ping`  will respond with a `Pong` .
* These pings/pongs contain nonces to prevent false positives on retries or allowing for spamming pongs (after a certain amount of invalid pongs, the peer is banned).
* If a ping times out, it is retried a few times.
* Once a certain amount of ping timeouts is hit, the connection is terminated (but the peer is *not* banned).

The core motivation for adding this to 1.5 is to prevent unlikely but possible connection stalls due to deadlock while interdependent nodes fetch backpressured tries from each other. As a side benefit, really slow connections or stalled are also terminated.

Test coverage for the functionality is extensive for the actual logic (see `health.rs`), which attempts to cover every possible edge case. Proper integration in layers up is also tested, but a certain amount of testing remains manual, as there is currently not a good way to easily write a tests that puts the nodes into the deadlocked state.

As a nice side benefit, the node can now be queries for round-trip times to other nodes through `net-info` on the diagnostics port.

Security aspects:

* Ping floods are prevented through rate limiting.
* Pong floods are prevented through rate limiting; also nodes ban peers that send too many unasked pings.
* A 2:1 cost ratio of ping:pong prevents blowing up a peers memory through pings.

### DIFF
--- a/node/src/components/network/message.rs
+++ b/node/src/components/network/message.rs
@@ -606,25 +606,23 @@ mod tests {
 
         let modern_handshake: Message<protocol::Message> = roundtrip_message(&legacy_handshake);
 
-        match modern_handshake {
-            Message::Handshake {
-                network_name,
-                public_addr,
-                protocol_version,
-                consensus_certificate,
-                is_syncing,
-                chainspec_hash,
-            } => {
-                assert_eq!(network_name, "example-handshake");
-                assert_eq!(public_addr, ([12, 34, 56, 78], 12346).into());
-                assert_eq!(protocol_version, ProtocolVersion::V1_0_0);
-                assert!(consensus_certificate.is_none());
-                assert!(!is_syncing);
-                assert!(chainspec_hash.is_none())
-            }
-            Message::Payload(_) | Message::Ping { .. } | Message::Pong { .. } => {
-                panic!("did not expect modern handshake to deserialize to anything but")
-            }
+        if let Message::Handshake {
+            network_name,
+            public_addr,
+            protocol_version,
+            consensus_certificate,
+            is_syncing,
+            chainspec_hash,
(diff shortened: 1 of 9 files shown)


EdHastingsCasperAssociation (approved):

georgepisaltu (approved):

node/src/components/network/message.rs
```diff
@@ -588,144 +606,136 @@ mod tests {
...
+        } = modern_handshake
+        {
+            assert_eq!(network_name, "example-handshake");
+            assert_eq!(public_addr, ([12, 34, 56, 78], 12346).into());
+            assert_eq!(protocol_version, ProtocolVersion::V1_0_0);
+            assert!(consensus_certificate.is_none());
+            assert!(!is_syncing);
+            assert!(chainspec_hash.is_none())
+        } else {
+            panic!("did not expect modern handshake to deserialize to anything but")
```
rafal-ch, line 625: but...? :)

rafal-ch (approved):
LGTM :+1:

node/src/components/network/message.rs
marc-casperlabs: @Fraser999 please mediate :)

Issue #3377, opened 2022-11-17
Add flow charts to new flows introduced in "new" fast sync implementation


piotr-dziubecki:
Description: Use the mermaid format and then the aquamarine crate to create proper docs
Points: 5

rafal-ch:
* :heavy_check_mark: ReactorState
* Interaction between components (aka the original bubble chart)
* :heavy_check_mark: CatchUp flow (on the high level) - also for other reactor states
* :question: BlockSynchronizer (on the high level)
* HotBlock (lifecycle of a block)
* :heavy_check_mark: DeployAcceptor (could be useful for external consumers, smart contract devs?)

abandoned:

* :heavy_check_mark: ComponentState
* :heavy_check_mark: BlockAcquisitionState
* ReactorState
* CatchUp process (_sync leaps_)
* KeepUp process
* Validate process
* BlockAccumulator
* DeployAcquisition
* ExecutionResultsAcquisition
* GlobalStateSynchronizer
* DeployBuffer (_hold, dead, etc._)
* UpgradeWatcher
* :question: Consensus

Issue #3445, opened 2022-12-06
Test sync_leaper component


piotr-dziubecki:
@piotr-dziubecki commented on [Mon Dec 05 2022](https://github.com/casper-network/Governance/issues/412)

PR #3545, opened 2023-01-04
Add protection against double component initialization

### LINKED ISSUE #3519: Fix potential race condition when initializing deploy buffer
Deploy buffer has custom initialization code that may lead to a race-condition resulting in deploy buffer already initialized log message

rafal-ch:
Closes #3519 by adding the transitional component state (`Initializing`) which means that component initialization has already been initiated and further init requests will be rejected.

In addition, it also changes the following:
1. Implementations of `handle_event()` in all components has been unified in a sense that all components produce consistent log messages using the `tracing` crate capabilities. Also, the components are now consistent on what they log (for example, double initialization attempts, fatal errors, etc.).
2. All components now have a `name()`, not only the initialized ones
3. There are minor renames (mainly for consistency, for example: `ComponentStatus` is now `ComponentState` by reference to `ReactorState`), several code clean-ups (in particular, `DeployBuffer` now has its own `initialize_component` moved from the generic `control.rs`) and bugfixes
4. `control_logic_default_delay` is used instead of `Duration::ZERO` to conduct cranking of the initialization phase
5. It moves the initialization of RPC and REST components to the end, so the Status endpoint is only accessible only when other components that provide status info are up
6. It adds a first flow-chart using `aquamarine` crate - the rest will follow here: https://github.com/casper-network/casper-node/issues/3377

### DIFF
--- a/node/src/components/deploy_buffer.rs
+++ b/node/src/components/deploy_buffer.rs
@@ -10,7 +10,10 @@ use std::{
 };
 
 use datasize::DataSize;
+use futures::FutureExt;
+use itertools::Itertools;
 use prometheus::Registry;
+use smallvec::smallvec;
 use tracing::{debug, error, info, warn};
 
 use casper_types::Timestamp;
@@ -25,6 +28,9 @@ use crate::{
         requests::{DeployBufferRequest, StorageRequest},
         EffectBuilder, EffectExt, Effects,
     },
+    fatal,
+    reactor::main_reactor::MainEvent,
+    storage::Storage,
     types::{
         appendable_block::{AddError, AppendableBlock},
         chainspec::DeployConfig,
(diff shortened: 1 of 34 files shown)


rafal-ch:
bors try

rafal-ch:
bors try

rafal-ch:
bors try

rafal-ch:
bors try

rafal-ch:
Merging [this PR](https://github.com/casper-network/casper-node/pull/3548) will unstuck the NCTL upgrade tests.

rafal-ch:
bors try

node/src/components/deploy_buffer.rs
```diff
@@ -94,6 +102,57 @@ impl DeployBuffer {
...
+            debug!(
+                blocks = ?blocks.iter().map(|b| b.height()).collect_vec(),
+                "DeployBuffer: initialization"
+            );
+            info!("initialized {}", <Self as Component<MainEvent>>::name(self));
+            let event = Event::Initialize(blocks);
+            return Some(smallvec![async {
+                smallvec![MainEvent::DeployBuffer(event)]
+            }
+            .boxed()]);
```
georgepisaltu, line 141: Can't we do `wrap_effects` or something similar that is more consistent with the other `Event`/`Effects` code?

georgepisaltu (commented):
LGTM, will approve once the upgrade tests are passing

rafal-ch:
```
2023-01-08T11:12:40.101162 [INFO] [9] NCTL :: ... node-1: TOTAL ERROR COUNT = 2
      1 "dropped without being responded to outside of shutdown"
      1 "request for channel closed, this may be a bug? check if a component is stuck from now on"
```

This is happening occasionally, when `BlockSynchronizer` is being asked for Status while being uninitialized.

rafal-ch:
> This is happening occasionally, when `BlockSynchronizer` is being asked for Status while being uninitialized.

Ok, the same may happen if Network is uninitialized, and we need to get Peers in order to provide status.

I moved the initialization of the RPC and REST server to the end in this commit: https://github.com/casper-network/casper-node/pull/3545/commits/ee6b89b8399ac566bc9cc762eaa4f68bdbe84b94

node/src/components/deploy_buffer.rs
rafal-ch: This is actually consistent with other places when we initialize components, like [here](https://github.com/casper-network/casper-node/blob/e091dbfd1b0cce7d40e7e0b92935bdb0746acc79/node/src/reactor/main_reactor/utils.rs#L19) or [here](https://github.com/casper-network/casper-node/blob/e091dbfd1b0cce7d40e7e0b92935bdb0746acc79/node/src/components/network/tests.rs#L206).

I'm open to changing it, but not as a part of this PR - it has already grown much beyond initially planned scope.

rafal-ch:
bors try

node/src/components/deploy_buffer.rs
georgepisaltu: I meant consistent with the code base in general, but I see what you mean and I agree, let's not do it in this PR.

georgepisaltu (approved):

node/src/components/storage/tests.rs
```diff
@@ -1674,7 +1674,14 @@ fn check_force_resync_with_marker_file() {
     // Add a couple of blocks into storage.
     let first_block = Block::random(&mut harness.rng);
     put_complete_block(&mut harness, &mut storage, Box::new(first_block.clone()));
-    let second_block = Block::random(&mut harness.rng);
+    let second_block = loop {
```
alsrdn, line 1677: nit: we can use `Block::random_with_specifics()`  to avoid doing a loop here.

alsrdn (approved):
LGTM with one small nit.

Fraser999 (approved):

node/src/components/storage/tests.rs
rafal-ch: I discussed this a bit with @georgepisaltu and we had a common understanding that the `random_with...` Block API is a bit too clunky. In this particular case, I would need to provide more specific parameters than just height.

My plan for [sync leaper testing](https://github.com/casper-network/casper-node/issues/3445) is to prepare a builder:
```
TestBlockBuilder::random().with_height().with_something_else().build()
```
so if you're not strong about this change I'll merge as is.

PR #3546, opened 2023-01-05
Config files housekeeping

### LINKED ISSUE #3495: Housekeeping of the config/chainspec values, remove superfluous files
Review if we don't keep superfluous chainspec/config data files in the repo (joiner chainspec could be such a file). Delete them when spotted. Also, review all default values and see if they are rational and consistent (for example, auction_delay should always be 1).

alsrdn:
Series of commits that aim to do a bit of housekeeping of the node configs:
* align comments between nctl and production chainspec files
* align `auction_delay` to 1 in configs and nctl scripts
* update chainspec values that were out of date or missmatched
* remove unused config files

Fixes: https://github.com/casper-network/casper-node/issues/3495


alsrdn:
bors try

Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

PR #3550, opened 2023-01-09
Allow cURL retries on "connection refused" and disable exponential backoff

rafal-ch:
This PR changes the default parameters used by cURL in NCTL when talking to nodes in the following way:
1. Enable retries even for the "connection refused" error (by default this is a hard error even with `--retry`)
2. Disables the exponential backoff algorithm for retries and uses a fixed 1 sec. delay instead (to get the info from the node as soon as possible)


Fraser999 (approved):

georgepisaltu (approved):

rafal-ch:
> Build failed:
> 
>     * [continuous-integration/drone/push](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/7155)

Build failed due to flakiness of one of the tests, which is fixed in another PR [here](https://github.com/casper-network/casper-node/pull/3545/commits/c69fbd287a7646eefa689b92e58f7f6e5aae6ce8).

Issue #3398, opened 2022-11-22
Failing varying size network connect test in 1.6 branch


piotr-dziubecki:
The varying size networking tests have been inadvertently broken during the recent merge and need to be fixed.

marc-casperlabs:
Remainder to be folded into #3333

marc-casperlabs:
Still not fixed, but PR pending for fix in #3554.

PR #3554, opened 2023-01-09
Demuxer waker bugfix

marc-casperlabs:
Adds a regression test (which required expanding the testing infrastructure, which in turn also needed testing) for the issue where all demultiplexer handles all go into a `Pending` state before the first item arrives on the underlying stream.

Closes (hopefully) #3398


georgepisaltu (approved):
LGTM

PR #3561, opened 2023-01-11
Implement tests for `block_accumulator`

### LINKED ISSUE #3346: Implement pruning of reactor's recent switch block headers

georgepisaltu:
Fixes #3346 

This PR adds tests for the `block_accumulator` component and its auxiliary parts.


node/src/components/block_accumulator.rs
```diff
@@ -808,6 +811,9 @@ impl<REv: ReactorEvent> Component<REv> for BlockAccumulator {
             Event::ExecutedBlock { hot_block } => {
                 let height = hot_block.block.header().height();
                 let era_id = hot_block.block.header().era_id();
+                // Logically it makes sense to purge the acceptors here or in
+                // `register_local_tip`, but only on the path where a new tip
+                // is accepted.
                 self.register_local_tip(height, era_id);
```
EdHastingsCasperAssociation: we do purge here (first line inside `register_local_tip`)

I think you are correct that the purge could be moved to within new block detection (`if new_local_tip`) tho

node/src/components/block_accumulator/block_acceptor.rs
```diff
@@ -389,3 +389,39 @@ impl BlockAcceptor {
...
+    pub(super) fn signatures(&self) -> &BTreeMap<PublicKey, (FinalitySignature, BTreeSet<NodeId>)> {
+        &self.signatures
+    }
+
+    #[allow(unused)]
+    pub(super) fn signatures_mut(
+        &mut self,
+    ) -> &mut BTreeMap<PublicKey, (FinalitySignature, BTreeSet<NodeId>)> {
+        &mut self.signatures
+    }
```
EdHastingsCasperAssociation: did you mean to leave #[allow(unused)] on this?

node/src/components/block_accumulator.rs
georgepisaltu: Yes, that is what I meant there.

node/src/components/block_accumulator/block_acceptor.rs
georgepisaltu: I did not, it's a remnant of a previous iteration and I'll remove it.

EdHastingsCasperAssociation (approved):

node/src/components/block_accumulator/tests.rs
```diff
@@ -1,9 +1,272 @@
...
+            Event::NetworkPeerBehaviorAnnouncement(peer_behavior) => {
+                self.blocked_peers.push(peer_behavior);
+                Effects::new()
+            }
+        }
+    }
+}
 
 #[test]
 fn upsert_acceptor() {
```
rafal-ch, line 269: I know that this test is not added in this PR, but while you're here, can you consider extending this test so that it also covers the block timestamp purging?

I see that when `block_timestamps.pop_front()` line is commented, all tests still pass.

node/src/components/block_accumulator/tests.rs
```diff
@@ -61,3 +324,768 @@ fn upsert_acceptor() {
...
+    // should not store anything.
+    acceptor.set_sufficient_finality(true);
+    let (should_store, _offenders) = acceptor.should_store_block(&era_validator_weights);
+    assert_eq!(should_store, ShouldStore::Nothing);
+    // Reset the flag.
+    acceptor.set_sufficient_finality(false);
+
+    let (should_store, offenders) = acceptor.should_store_block(&era_validator_weights);
+    assert_eq!(should_store, ShouldStore::Nothing);
+    assert!(offenders.is_empty());
```
rafal-ch, line 637: Can we also have a test case when offenders are not expected to be empty?

node/src/components/block_accumulator/tests.rs
```diff
@@ -61,3 +324,768 @@ fn upsert_acceptor() {
...
+                let event = super::Event::ExecutedBlock {
+                    hot_block: hot_block_1.clone(),
+                };
+                effect_builder
+                    .into_inner()
+                    .schedule(event, QueueKind::Validation)
+                    .ignore()
+            })
+            .await;
+        for _ in 0..4 {
```
rafal-ch, line 1400: Is there a specific number of times we need to crank? I see that it's set to `4` here, but `6` in other cases.
If yes, please consider using a named constant.

node/src/components/block_accumulator.rs
```diff
@@ -328,6 +328,9 @@ impl BlockAccumulator {
             .as_ref()
             .map_or(false, |local_tip| block_height < local_tip.height)
         {
+            // If the acceptor exists already (through a previously recorded
+            // signature), we should probably purge it if it's older than
```
rafal-ch: I think we should clarify this "probably" before merging this PR, wdyt?

node/src/components/block_accumulator/tests.rs
```diff
@@ -61,3 +324,768 @@ fn upsert_acceptor() {
...
+    assert_eq!(should_store, ShouldStore::Nothing);
+
+    // Create the third validator's signature.
+    let fin_sig = FinalitySignature::create(
+        *block.hash(),
+        block.header().era_id(),
+        &keys[2].0,
+        keys[2].1.clone(),
+    );
+    // The third signature with weight 10% doesn't change make the block go to
```
Fraser999:
```suggestion
    // The third signature with weight 10% doesn't make the block go to
```

node/src/components/block_accumulator.rs
```diff
@@ -328,6 +328,9 @@ impl BlockAccumulator {
...
+            // the local tip.
```
Fraser999: Won't the next purge do that anyway?

node/src/components/block_accumulator/tests.rs
```diff
@@ -1,9 +1,272 @@
-use casper_types::testing::TestRng;
+use std::{
```
rafal-ch, line 1: I think it might be beneficial to cover some utility functions with simple UTs, for example, I can flip the `>` comparison in `should_leap()` and all tests still pass. Such change will probably be detected in NCTL tests, but these are not that quick to run.

Also, `purge()` has a potential to contain some off-by-one errors.

node/src/components/block_accumulator/tests.rs
```diff
@@ -61,3 +324,768 @@ fn upsert_acceptor() {
...
+        };
+        let effects = block_accumulator.handle_event(effect_builder, &mut rng, event);
+        assert!(effects.is_empty());
+        // This should *probably* have no effect on the accumulator since the
+        // block is older than the local tip, but right now it creates the
+        // acceptor if it's in the same era or newer than the local tip era.
+        // See comment in `BlockAccumulator::register_block`.
+        // assert!(!block_accumulator
+        //     .block_acceptors
+        //     .contains_key(older_block.hash()));
```
Fraser999: This is registering a finality signature though rather than a block.

You could maybe provide a signature from an older era and check no new acceptor is created, and/or uncomment this but assert that the acceptor exists (with a comment about why it does).

Fraser999 (commented):
Generally looking good, but just a couple of comments about the comments :)

node/src/components/block_accumulator.rs
georgepisaltu: I think we should, yes. I put this comment here to get input from multiple people.

node/src/components/block_accumulator.rs
georgepisaltu: It will, but at the same time why not purge it at the exact time you find out it's no longer useful?

node/src/components/block_accumulator/tests.rs
georgepisaltu: The operator flip works like that exactly because `should_leap` doesn't have unit tests. I agree we should have those, but at the same time there are lots of functions that don't have unit tests yet (it's a lot of code). In this PR I wanted to test the most *problematic* things in the accumulator. We can incrementally add more and more tests, and I will add the `should_leap` unit test along with the `purge` in the next iteration of this PR.

node/src/components/block_accumulator/tests.rs
georgepisaltu: The explanation for this is the same as the other comment. I will extend the test.

node/src/components/block_accumulator/tests.rs
georgepisaltu: Will do.

node/src/components/block_accumulator/tests.rs
georgepisaltu: These numbers are not constant, they actually represent the number of events that will be processed by scheduling the above event and they are not constant, not even for the same scheduled event. Unfortunately I don't see a nice way to handle this. I can add comments above the `try_cranks` explaining that this is how many events will be handled, but not much more.

node/src/components/block_accumulator/tests.rs
georgepisaltu: The problem scenario is this:
- we have an old block which, if received by the accumulator, would not have an acceptor created
- if you start by sending the signatures for said block, which creates the acceptor if the era is recent enough
- only after receiving the block we realize that this acceptor is useless
- it will eventually get purged

This is wasteful and arguably introduces unnecessary complexity. We could keep track of purged block hashes until their era stops being recent, but this would probably not be worth the hassle. Alternatively, we could just purge when we realize it's a stale block and live with it.

node/src/components/block_accumulator.rs
Fraser999: I'd lean towards keeping the code simpler.  Also, the purge method looks like it could be expensive and if we're only calling from one place it makes it easier to rationalize about.  So I'd favour just deleting this comment and leaving purge as is.

node/src/components/block_accumulator/tests.rs
georgepisaltu: Done.

node/src/components/block_accumulator/tests.rs
georgepisaltu: Done.

node/src/components/block_accumulator/tests.rs
georgepisaltu: Done.

node/src/components/block_accumulator.rs
georgepisaltu: Talked with @Fraser999 and we'll keep it as is.

node/src/components/block_accumulator/tests.rs
georgepisaltu: Done.

node/src/components/block_accumulator/tests.rs
georgepisaltu: We decided not to do anything about this. We accept this behavior to be the expected one.

Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

PR #3566, opened 2023-01-12
Fix duplicate block execution

### LINKED ISSUE #3432: Metric Issue: contract_runtime_run_execute_count
metric value mismatch seen between a 'validator' and 'non-validator' ex. deploy of 30000 correct on validator was showing 32500 on non-validator

georgepisaltu:
Fixes #3432 

This PR fixes a bug which would ultimately cause a block to be executed twice when in keep up mode at the tip of the chain. The details of the bug can be found in the original issue, linked above.

It introduces a new state for forward block builders in the synchronizer, namely `Executing`. This state is a precursor to the already existing `Synced` and prevents the block builder purging prematurely. It also introduces a new event to the synchronizer which is to be triggered when the hot block added announcement from contract runtime reaches the reactor.

After all the components of the block to be sync'd are acquired by the forward block builder, it immediately enters a state of waiting for the execution of the block, delaying the builder purge until the execution is finished, the contract runtime announced the updated hot block and the accumulator marked it complete, therefore eliminating the race condition.

The state transition in the synchronizer should be "instantaneous" relative to the event loop, as the `MarkBlockExecutionEnqueued` event should be handled immediately after enqueueing the block for execution, but I'm curious if this is actually true and would like a second opinion here. At the time of opening this PR, I don't have the means to test this with an NCTL network.


dwerner (approved):
Looks good to me.

node/src/types/block/hot_block/state.rs
```diff
@@ -40,6 +40,7 @@ pub(crate) struct State {
     pub(super) tried_to_sign: bool,
     pub(super) sent_to_consensus_post_execution: bool,
     pub(super) sent_to_accumulator_post_execution: bool,
+    pub(super) sent_to_synchronizer_post_execution: bool,
```
Fraser999: I don't think we need this new field.  Seems like we always want to send `block_synchronizer::Event::ExecutedBlock` to the synchronizer after the block is marked complete by the accumulator (similar to the event stream server events which are created after the block is marked complete).

Fraser999 (changes requested):
Nice job, but one blocker I think.

node/src/types/block/hot_block/state.rs
georgepisaltu: I don't think this is true because once we sent the event once, the state transitions in the synchronizer would put it into the `Synced` state, and subsequently the builder will be purged on the next `keep_up_instruction` call.

In case we allow duplicates and for some reason they actually happen, they will either change the `executing_progress` state change to `Done` a second time (with no additional effect) or miss the call because there is no builder there. Either way seems like a superfluous event to me.

I understand this looks like it's polluting the hot block state structure, but with its current design it looks like that's what it's there for; to keep relevant state changes across components. Is there another reason you'd want this removed that I didn't mention in this comment?

node/src/components/block_accumulator/sync_identifier.rs
```diff
@@ -13,6 +13,7 @@ pub(crate) enum SyncIdentifier {
     // including sufficient finality; this may be a historical
     // block and / or potentially the new highest block
     SyncedBlockIdentifier(BlockHash, u64, EraId),
+    ExecutingBlockIdentifier(BlockHash, u64, EraId),
```
EdHastingsCasperAssociation, line 19: add a comment

node/src/components/block_synchronizer/event.rs
```diff
@@ -29,6 +29,7 @@ pub(crate) enum Event {
         result: Option<(FinalizedBlock, Vec<Deploy>)>,
     },
     MarkBlockExecutionEnqueued(BlockHash),
+    ExecutedBlock(BlockHash),
```
EdHastingsCasperAssociation: MarkBlockExecuted

EdHastingsCasperAssociation (approved):
prelim approval, see nitpicks

node/src/components/block_accumulator/sync_identifier.rs
georgepisaltu: Done.

node/src/components/block_synchronizer/event.rs
georgepisaltu: Renamed.

node/src/types/block/hot_block/state.rs
georgepisaltu: As discussed offline, we want to have this check to enforce no duplicate events are dispatched, but it makes sense to apply that to all the subsequent events that happen in `handle_hot_block`. As agreed, we now have an `all_actions_done` flag in the state instead of `sent_to_synchronizer_post_execution` which guards the rest of the events as well.

node/src/components/block_synchronizer/block_builder.rs
```diff
@@ -233,7 +263,47 @@ impl BlockBuilder {
...
+                (false, None) => {
+                    let block_hash = self.block_hash();
+                    error!(%block_hash, "invalid attempt to start block execution");
+                    self.abort();
+                }
+                (false, Some(executing_progress)) => {
+                    self.touch();
+                    self.execution_progress = executing_progress;
+                }
+            }
```
alsrdn, line 284: **Nit**: I think the semantics of `start()` are a bit hard to read. Since we can only start from `ExecutionProgress::Idle`, I think we can remove `start()` and replace with something like:
```suggestion
            match (
                self.should_fetch_execution_state,
                self.execution_progress,
            ) {
                (true, _) => {
                    let block_hash = self.block_hash();
                    error!(%block_hash, "invalid attempt to start block execution on historical block");
                    self.abort();
                }
                (false, ExecutionProgress::Idle) => {
                    self.touch();
                    self.execution_progress = ExecutionProgress::Started;
                }
                (false, _) => {
                    let block_hash = self.block_hash();
                    error!(%block_hash, execution_progress=%self.execution_progress "cannot track block execution progress");
                    self.abort();
                }
            }
```

node/src/components/block_synchronizer/block_builder.rs
```diff
@@ -233,7 +263,47 @@ impl BlockBuilder {
...
+        }
+    }
+
+    pub(super) fn register_block_executed(&mut self) {
+        match (
+            self.should_fetch_execution_state,
+            self.execution_progress.finish(),
```
alsrdn, line 291: Same as above.

alsrdn (approved):
Looks good with one nit.

Fraser999 (approved):

alsrdn:
bors try

Fraser999 (approved):

PR #3574, opened 2023-01-16
Use `Arc` instead of `Box` for announcements

### LINKED ISSUE #3509: Use Arcs for block announcements
Like other announcements, the block stored announcement is currently processed by multiple components/handlers, causing the Box<Block> to be needlessly copied. It should be replaced by an Arc<Block> to avoid this.

marc-casperlabs:
Closes #3509.

Replaces all uses of `Box` with `Arc` where announcements are involved. This has a ripple effect of `Arc`'ing a few additional things.


marc-casperlabs:
~More follow up commits to come.~

Closing this for now, as it may have been obsoleted by a refactor in the meantime.

PR #3577, opened 2023-01-16
Swap lmdb for lmdb-rkv latest (0.14)

dwerner:
This PR swaps out `lmdb` for `lmdb-rkv`. All that was needed was handling a `Result` type which cursor iterators now yield.


dwerner:
bors try

Fraser999 (approved):

dwerner:
bors merge

Issue #3578, opened 2023-01-17
Nodes as non-validators could not upgrade


cspramit:
There 3 test runs below with continuous loads for 2 tests and no load for the 3rd test. The first two failed to have non-validating nodes successfully upgrade - all such nodes missed the upgrade activation and started erroring with 'incompatible version' or with "message":"unable to initialize contract runtime with a system contract registry.

**2nd test Run** - Test Run 2: Large Network Testing
Tests:
mixed load continuous (transfers,wasm deploy, bloated deploys, nft mint, delegations)
upgrade 1.5.0 -> 1.5.0
Result: Partial Non-Validating Nodes Not Upgraded + Partial Nodes cannot Initialise after upgrade
Observations:
All non-validating nodes did not upgrade cleanly
deploys dispatched or stored on nodes at upgrade
Some Nodes as previous - failed to upgrade and then fails with peer is running incompatible version: 1.5.0
Some Nodes as yesterday - "message":"unable to initialize contract runtime with a system contract registry"
Dumps: http://genesis.casperlabs.io/rti-dev-big/casper-node-dumps/rti-dev-big/17012023_0436/dump_download_list.txt

**1st Test run :**
Name: rti-dev-big
Network: 100 + 20
Branch: dev
Chain: public
Commit: 925bbcb1a6e61594766a5794daa6587583d64d1b
Config Override: None
Test Run 1: Large Network Testing
Tests:
mixed load continuous (transfers,wasm deploy, bloated deploys, nft mint, delegations)
Era 2: upgrade 1.4.9 -> 1.5.0
Era 4: upgrade 1.5.0 -> 1.5.0
Result: All 20 Non-Validating nodes did not Upgrade
Observations:
deploys were dispatched and stored on non-validating nodes
the validators performed upgrade activation correctly
all non-validating nodes missed the upgrade activation and started erroring with 'incompatible version'
Dumps: http://genesis.casperlabs.io/rti-dev-big/casper-node-dumps/rti-dev-big/17012023_0239/dump_download_list.txt
Nodes of Interest: 172.142.82.92(validator), 172.142.85.207 (non-validator)

**3rd test run : Success** 
Test Run 3: Large Network Testing
Tests:
no load to network - idle
Result: All validating and non-validating nodes upgraded successfully

PR #3584, opened 2023-01-18
Avoid fetching all finality signatures concurrently

Fraser999:
This PR changes the `BlockSynchronizer` to avoid creating all fetch requests for finality signatures in bulk.

The change should have the effect of requesting only `max_simultaneous_peers` (i.e. 5) at a time, and rather than asking 5 peers for the same signature, each peer will be asked for a different signature.

While trying to acquire weak finality, rather than stopping as soon as we have weak finality we continue until we have at least requested every potential signature.

There is also a fix for `PeerList::need_peers()` whereby it would previously return `Sufficient` even if it exclusively contained peers which were marked dishonest, meaning a subsequent call to `PeerList::qualified_peers()` would return an empty vector.

Related to #3578.


node/src/components/block_synchronizer/block_acquisition.rs
```diff
@@ -478,16 +507,22 @@ impl BlockAcquisitionState {
...
-                            acquired_signatures.clone(),
-                        ))
+                        if acquired_signatures.have_no_vacant() {
+                            Some(BlockAcquisitionState::HaveWeakFinalitySignatures(
+                                header.clone(),
+                                acquired_signatures.clone(),
+                            ))
+                        } else {
+                            None
+                        }
```
EdHastingsCasperLabs, line 525: i think requiring no vacancy here is incorrect

node/src/components/block_synchronizer/peer_list.rs
```diff
@@ -125,8 +125,12 @@ impl PeerList {
...
 
     pub(super) fn need_peers(&mut self) -> PeersStatus {
-        if self.peer_list.is_empty() {
-            debug!("PeerList: is empty");
+        if !self
+            .peer_list
+            .iter()
+            .any(|(_, pq)| *pq != PeerQuality::Dishonest)
+        {
+            debug!("PeerList: no honest peers");
```
EdHastingsCasperAssociation, line 133: good catch

EdHastingsCasperAssociation (approved):

Fraser999:
bors merge

Fraser999:
bors merge

node/src/components/block_synchronizer.rs
```diff
@@ -519,18 +519,23 @@ impl BlockSynchronizer {
...
                 NeedNext::FinalitySignatures(block_hash, era_id, validators) => {
                     builder.set_in_flight_latch();
-                    results.extend(peers.into_iter().flat_map(|node_id| {
-                        validators.iter().flat_map(move |public_key| {
-                            let id = FinalitySignatureId {
-                                block_hash,
-                                era_id,
-                                public_key: public_key.clone(),
-                            };
+                    for (validator, peer) in validators
```
georgepisaltu, line 551: Just to double check: we are losing certainty that we will get the signatures at this step since we are no longer asking multiple peers for each signature, and it becomes more likely that some of these fetches will miss. However, that will not be a problem because we will fill the missing slots in the following passes through `NeedNext::FinalitySignatures`. Is this correct?

node/src/components/block_synchronizer/block_acquisition.rs
```diff
@@ -478,16 +507,22 @@ impl BlockAcquisitionState {
...
                 maybe_block_hash = Some(header.block_hash());
                 added = acquired_signatures.apply_signature(signature);
                 match validator_weights.signature_weight(acquired_signatures.have_signatures()) {
                     SignatureWeight::Insufficient => None,
                     SignatureWeight::Weak | SignatureWeight::Strict => {
-                        Some(BlockAcquisitionState::HaveWeakFinalitySignatures(
-                            header.clone(),
-                            acquired_signatures.clone(),
-                        ))
+                        if acquired_signatures.have_no_vacant() {
```
georgepisaltu, line 560: Why do we need to enforce that no vacant signatures exist in the signature acquisition at this point? We could already have weak finality even with vacant and/or pending slots. Is this a check to ensure we *at least* asked for all signatures before moving on?

georgepisaltu (commented):
LGTM overall, just a couple of questions before approving.

node/src/components/block_synchronizer/block_acquisition.rs
rafal-ch: I'd say yes, as per the updated comment above.

rafal-ch (approved):
LGTM :+1:

node/src/components/block_synchronizer.rs
Fraser999: Yes, exactly.

node/src/components/block_synchronizer/block_acquisition.rs
Fraser999: Yes again :)

Fraser999:
bors merge

Fraser999:
bors merge

Fraser999:
bors try

Fraser999:
bors try

georgepisaltu (approved):
LGTM

Fraser999:
bors merge

PR #3588, opened 2023-01-18
GH-3585: Precise control over control flow opcode costs

### LINKED ISSUE #3585: Implement precise control over specific control flow opcodes
In 1.4.9 release we increased all opcode costs related to control flow. This was done to protect the network against slow wasms (i.e. busy loops, etc.) by charging more. Incidentally it also increased costs of valid wasms. In this ticket we need to implement precise control over opcode costs, lower cost of opcodes unrelated to slow wasms, but still make slow wasms more expensive to run.

mpapierski:
Closes #3585 

This PR introduces per-opcode costs for the control flow group of opcodes. This way we can ensure that wasms that execute slowly due to loops are correctly charged for busy loops, but general wasms aren't affected. Additionally, as @georgepisaltu mentioned, we decided that we should charge for `br_table` dynamically depending on the number of labels it has - this should discourage the usage of a large number of labels in the code.

This PR also lowers all control flow opcodes to 440 (as in 1.4.8) except `br_table`, `br_if`, and `br`.

Real-world impact:

- NFT is ~50% cheaper in minting requests compared to 1.4.9
- Transfer is ~50% cheaper
- Installing is ~10% cheaper


resources/production/chainspec.toml
```diff
@@ -148,6 +146,27 @@ grow_memory = 240_000
...
 
+# Control flow operations multiplier.
+[wasm.opcode_costs.control_flow]
+block = 440
+loop = 440
+if = 440
+else = 440
+end = 440
+br = 440000
+br_if = 440000
```
Fraser999:
```suggestion
br = 440_000
br_if = 440_000
```

resources/production/chainspec.toml
```diff
@@ -148,6 +146,27 @@ grow_memory = 240_000
...
+return = 440
+select = 440
+call = 440
+call_indirect = 440
+drop = 440
+
+[wasm.opcode_costs.control_flow.br_table]
+# Fixed cost per `br_table` opcode
+cost = 440000
```
Fraser999:
```suggestion
cost = 440_000
```

resources/test/valid/1_0_0/chainspec.toml
```diff
@@ -63,6 +62,26 @@ current_memory = 27
...
 regular = 29
 
+# Control flow operations multiplier.
+[wasm.opcode_costs.control_flow]
+block = 440
+loop = 440
+if = 440
+else = 440
+end = 440
+br = 440000
```
Fraser999: This is missing `br_if` I think.

execution_engine_testing/test_support/src/lib.rs
```diff
@@ -120,7 +122,13 @@ pub static DEFAULT_ACCOUNTS: Lazy<Vec<GenesisAccount>> = Lazy::new(|| {
 /// Default [`ProtocolVersion`].
 pub static DEFAULT_PROTOCOL_VERSION: Lazy<ProtocolVersion> = Lazy::new(|| ProtocolVersion::V1_0_0);
 /// Default payment.
-pub static DEFAULT_PAYMENT: Lazy<U512> = Lazy::new(|| U512::from(1_500_000_000_000u64));
+pub static DEFAULT_PAYMENT: Lazy<U512> = Lazy::new(|| {
+    let amount = env::var("DEFAULT_PAYMENT")
```
Fraser999: Is this env var set anywhere in our tests?

If we don't need it, I'd suggest we remove this.  Otherwise I think it should be documented (maybe in the README and the top of lib.rs for this crate?)

execution_engine_testing/tests/src/test/contract_api/get_call_stack.rs
```diff
@@ -2620,7 +2620,10 @@ mod payment {
         } else {
             builder.exec(execute_request).commit().expect_failure();
             let error = builder.get_error().expect("must have an error");
-            assert!(matches!(error, CoreError::Exec(ExecError::GasLimit)));
+            assert!(matches!(
+                error,
+                CoreError::Exec(ExecError::InvalidContext) | CoreError::Exec(ExecError::GasLimit)
```
Fraser999: It's not clear to me why we'd sometimes get an `InvalidContext` error now.

Maybe we could extract the whole `if call_depth == 0` block to a separate function and a comment there.  It's repeated 5 times in this mod I think.

execution_engine/src/shared/opcode_costs.rs
```diff
@@ -41,11 +40,287 @@ pub const DEFAULT_NOP_COST: u32 = 200;
...
+/// Definition of a cost table for a Wasm `br_table` opcode.
+///
+/// Charge of a `br_table` opcode is calculated as follows:
+///
+/// ```text
+/// cost + (len(br_table.targets) * size_multiplier)
+/// ```
+///
+/// This is done to punish users for writing code with very long `br_table`s and encourage users to
+/// optimize their generated code.
```
Fraser999:
```suggestion
/// This is done to encourage users to avoid writing code with very long `br_table`s.
```

execution_engine_testing/tests/Cargo.toml
```diff
@@ -28,6 +28,7 @@ wabt = "0.10.0"
 wasmi = "0.8.0"
 regex = "1.5.4"
 wat = "1.0.47"
+walrus = "0.19.0"
```
Fraser999: This could be a dev-dependency I think?

execution_engine/src/shared/opcode_costs.rs
```diff
@@ -41,11 +40,287 @@ pub const DEFAULT_NOP_COST: u32 = 200;
...
+        } = self;
+
+        cost.serialized_length() + size_multiplier.serialized_length()
+    }
+}
+
+impl FromBytes for BrTableCost {
+    fn from_bytes(bytes: &[u8]) -> Result<(Self, &[u8]), bytesrepr::Error> {
+        let (cost, bytes): (_, &[u8]) = FromBytes::from_bytes(bytes)?;
+        let (target_size_multiplier, bytes): (_, &[u8]) = FromBytes::from_bytes(bytes)?;
```
Fraser999:
```suggestion
        let (size_multiplier, bytes): (_, &[u8]) = FromBytes::from_bytes(bytes)?;
```

execution_engine/src/shared/opcode_costs.rs
```diff
@@ -41,11 +40,287 @@ pub const DEFAULT_NOP_COST: u32 = 200;
...
+/// Default cost of the `end` Wasm opcode.
+pub const DEFAULT_CONTROL_FLOW_END_OPCODE: u32 = 440;
+/// Default cost of the `br` Wasm opcode.
+pub const DEFAULT_CONTROL_FLOW_BR_OPCODE: u32 = 440000;
+/// Default cost of the `br_if` Wasm opcode.
+pub const DEFAULT_CONTROL_FLOW_BR_IF_OPCODE: u32 = 440000;
+/// Default fixed cost of the `br_table` Wasm opcode.
+pub const DEFAULT_CONTROL_FLOW_BR_TABLE_OPCODE: u32 = 440000;
+/// Default multiplier for the size of targets in `br_table` Wasm opcode.
+pub const DEFAULT_BR_TABLE_MULTIPLIER: u32 = 100;
```
Fraser999:
```suggestion
pub const DEFAULT_CONTROL_FLOW_BR_TABLE_MULTIPLIER: u32 = 100;
```

execution_engine/src/shared/opcode_costs.rs
```diff
@@ -41,11 +40,287 @@ pub const DEFAULT_NOP_COST: u32 = 200;
...
+                size_multiplier: target_size_multiplier,
+            },
+            bytes,
+        ))
+    }
+}
+
+/// Definition of a cost table for a Wasm control flow opcodes.
+#[derive(Copy, Clone, PartialEq, Eq, Serialize, Deserialize, Debug, DataSize)]
+pub struct ControlFlowCost {
```
Fraser999:
```suggestion
pub struct ControlFlowCosts {
```

execution_engine_testing/tests/Cargo.toml
Fraser999: Actually, seems like we can do the following tidy-up:

```
[dependencies]
base16 = "0.2.1"
casper-engine-test-support = { path = "../test_support", features = ["test-support"] }
casper-execution-engine = { path = "../../execution_engine", features = ["test-support"] }
casper-hashing = { path = "../../hashing" }
casper-types = { path = "../../types", features = ["datasize", "json-schema"] }
clap = "2"
fs_extra = "1.2.0"
log = "0.4.8"
parity-wasm = "0.41.0"
rand = "0.8.3"
serde = "1"
serde_json = "1"
tempfile = "3"
wabt = "0.10.0"

[dev-dependencies]
assert_matches = "1.3.0"
dictionary = { path = "../../smart_contracts/contracts/test/dictionary", default-features = false }
dictionary-call = { path = "../../smart_contracts/contracts/test/dictionary-call", default-features = false }
get-call-stack-recursive-subcall = { path = "../../smart_contracts/contracts/test/get-call-stack-recursive-subcall", default-features = false }
gh-1470-regression = { path = "../../smart_contracts/contracts/test/gh-1470-regression", default-features = false }
gh-1470-regression-call = { path = "../../smart_contracts/contracts/test/gh-1470-regression-call", default-features = false }
num-rational = "0.4.0"
num-traits = "0.2.10"
once_cell = "1.5.2"
regex = "1.5.4"
walrus = "0.19.0"
wat = "1.0.47"
```

Fraser999 (commented):
LGTM barring a couple of nitpicks.

execution_engine_testing/test_support/src/lib.rs
mpapierski: It was useful to measure the time of execution of slow wasms, but it's not strictly necessary to keep it in the code - removed

execution_engine_testing/tests/Cargo.toml
mpapierski: Good catch. With one addition that I just realized by running `make lint` we actually use `criterion` for benchmarks

execution_engine_testing/tests/src/test/contract_api/get_call_stack.rs
mpapierski: Refactored this duplicated assertion,

So before increasing control flow codes it was meant to be failing at InvalidContext, then in 1.4.9 this test was heavily refactored to always assume GasLimit (which was incorrect; but it was passing due to high costs). By lowering all control flow opcodes except br, br_table, br_if I had to modify this again as some tests failed with InvalidContext (Correct) but some with GasLimit (now also correct; but such test didn't reach stored contract -> stored session transition that triggers InvalidContext)

Fraser999 (approved):

georgepisaltu (approved):

PR #3590, opened 2023-01-19
Sync leaper tests

### LINKED ISSUE #3445: Test sync_leaper component

rafal-ch:
This PR adds test for the `sync leaper` component.

Closes https://github.com/casper-network/casper-node/issues/3445


rafal-ch:
bors try

node/src/components/sync_leaper.rs
```diff
@@ -26,203 +25,48 @@ use crate::{
...
-        in_flight: usize,
+#[derive(Debug)]
+enum RegisterLeapAttemptOutcome {
+    DoNothing,
+    FetchSyncLeapFromPeers(Vec<NodeId>),
+}
+
+#[derive(Debug, Error)]
+enum Error {
+    #[error("fetched a sync leap from storage - should never happen - {0}")]
```
georgepisaltu: I don't think we should have the "- should never happen -" in the error message itself, maybe as a code comment.

node/src/components/sync_leaper/tests.rs
```diff
@@ -0,0 +1,374 @@
...
+    components::{
+        fetcher::{self, FetchResult, FetchedData},
+        sync_leaper::{LeapState, PeerState, RegisterLeapAttemptOutcome},
+    },
+    types::{Block, BlockHash, Chainspec, NodeId, SyncLeap, SyncLeapIdentifier},
+};
+
+use super::{Error, SyncLeaper};
+
+pub(crate) fn make_default_sync_leap(rng: &mut TestRng) -> SyncLeap {
```
georgepisaltu: I don't think `make_default_sync_leap` is a suitable name here because it's based on a random block, which has no defaults.

node/src/components/sync_leaper/leap_activity.rs
```diff
@@ -0,0 +1,482 @@
...
+
+        let mut leap_activity = LeapActivity {
+            sync_leap_identifier,
+            peers: [peer_1.clone()].iter().cloned().collect(),
+            leap_start: Instant::now(),
+        };
+
+        let (actual_sync_leap, actual_peers) = leap_activity.best_response().unwrap();
+
+        assert!(!actual_peers.is_empty());
```
georgepisaltu, line 220: nit: if we do `actual_peers.get(0).unwrap()` below, there is no need for this assertion.

node/src/components/sync_leaper/leap_activity.rs
```diff
@@ -0,0 +1,482 @@
...
+                    peer_3_best_node_id,
+                    PeerState::Fetched(Box::new(best_sync_leap.clone())),
+                ),
+            ]
+            .iter()
+            .cloned(),
+        );
+
+        let (actual_sync_leap, mut actual_peers) = leap_activity.best_response().unwrap();
+
```
georgepisaltu, line 303: nit: At this point I'd add a peer with a worse response, just to make sure adding a worse valid response doesn't break things.

node/src/components/sync_leaper/leap_state.rs
```diff
@@ -0,0 +1,156 @@
...
+        match self {
+            LeapState::Idle => 0,
+            LeapState::Awaiting { in_flight, .. }
+            | LeapState::Received { in_flight, .. }
+            | LeapState::Failed { in_flight, .. } => *in_flight,
+        }
+    }
+
+    pub(super) fn active(&self) -> bool {
+        self.in_flight() > 0
```
georgepisaltu, line 86: nit: it would be better to do this with an explicit exhaustive match on `LeapState` so that any further changes to the enum will make us revisit this function.

georgepisaltu (approved):
LGTM, nice work! Only a couple of nits but ready to merge.

node/src/components/sync_leaper.rs
EdHastingsCasperAssociation: agreed

node/src/components/sync_leaper/leap_activity.rs
```diff
@@ -0,0 +1,482 @@
...
+        }
+        match self.best_response() {
+            Ok((best_available, from_peers)) => LeapState::Received {
+                in_flight,
+                best_available: Box::new(best_available),
+                from_peers,
+            },
+            // `Unobtainable` means we couldn't download it from any peer so far - don't treat it
+            // as a failure if there are still requests in flight
+            Err(LeapActivityError::Unobtainable(_, _)) if in_flight > 0 => LeapState::Awaiting {
```
alsrdn, line 60: I'm wondering if maybe we should also check `in_flight` for `LeapActivityError::TooOld` as well? Would there be any situation where some peer may say that the trusted hash is too old but others accept it? Maybe if there's a malicious peer?

alsrdn (approved):
LGTM, just have one question

node/src/components/sync_leaper/leap_activity.rs
rafal-ch: I'd say we should, indeed. Let's discuss and possibly handle it as a separate ticket, though. Is it ok?

node/src/components/sync_leaper/leap_state.rs
rafal-ch: Agreed, but I'm not touching this right now to avoid accidental regressions.

node/src/components/sync_leaper/leap_activity.rs
rafal-ch: Good idea - test has been extended in 2cecd103bad13e3d777c9fa249e38d2d4d766556

node/src/components/sync_leaper/leap_activity.rs
rafal-ch: I'd prefer the explicit check, it is more clear to the reader that we actually expect the peers to not be empty.

node/src/components/sync_leaper/tests.rs
rafal-ch: Renamed to `make_test_sync_leap()` in afe2163ce3d4f7bbbcb62208da4dec65c5da75b8

node/src/components/sync_leaper.rs
rafal-ch: That was just a move of an already existing message, which is now updated in afe2163ce3d4f7bbbcb62208da4dec65c5da75b8

node/src/components/sync_leaper/leap_activity.rs
rafal-ch: As discussed off-line, the flow for `TooOld` is correct.

PR #3592, opened 2023-01-19
Various fast-sync fixes

Fraser999:
This PR contains multiple fixes:
* use the highest COMPLETE block to initialize the contract runtime (this guarantees the global state specified in the block exists as the node must have executed or synced the state)
* ensure finality signatures are gossiped to validators and non-validators to try to avoid a network with mostly validators exiting for upgrade before the non-validators have had a chance to participate in the gossiping of the finality signatures of the last block before upgrade
* ~for switch blocks, don't have the validators broadcast their finality signatures to help keep the non-validators following along more closely synced at era boundaries~
* for historical blocks, ensure the block synchronizer doesn't make multiple `MetaBlockAnnouncement`s for the same block, as this results in duplicate events being sent to the event stream server

Related to #3578.

### DIFF
--- a/node/src/components/block_synchronizer.rs
+++ b/node/src/components/block_synchronizer.rs
@@ -418,6 +418,7 @@ impl BlockSynchronizer {
         &mut self,
         effect_builder: EffectBuilder<REv>,
         block_hash: &BlockHash,
+        was_not_previously_marked_complete: bool,
     ) -> Effects<Event>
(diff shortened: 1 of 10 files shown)


alsrdn (approved):

georgepisaltu (approved):

Fraser999:
bors merge

PR #3594, opened 2023-01-20
Propagate unreliable peers upwards from the trie accumulator to the block synchronizer

### LINKED ISSUE #3563: Long Running Network Joining node in catchup/sync leap for long
Observation:  when network is being moderately loaded continuously, the joining nodes does not make it to KeepUp. But when load is cut off and no deploys being executed AND node restarted, the joiner node gets to KeepUp sooner.
Currently after 22hrs has not caughtup.

Name: rti-dev-small
Network: 5 + 2
Branch: dev
Chain: public
Commit: 01c2af5298a5d36fde5a472344da6f702295aa91
Config Override: None
Mixed Deploy and Long Running Network Stability Testing
Test :
25 x multi delegations periodically
1 x multi NFT contract and minting periodically
5 x multi bloated wasm periodically
12k x transfers and wasm transfers periodically
4 x add joining node at certain intervals
random node restarts
Observations:
network running stable
network running for 48+ hrs
4 x joiner nodes @ network with 100k deploys successful
4 x joiner nodes @ network with 500k deploys - in progress - been 22 hrs so far
total network deploys currently at 1+million

Dumps: http://genesis.casperlabs.io/rti-dev-small/casper-node-dumps/rti-dev-small/10012023_1219/dump_download_list.txt
Nodes of Interest: 172.44.57.234, 172.44.68.115 (joiners)
Grafana: https://grafana.casperlabs.io/goto/1JLwb0hVz?orgId=1

alsrdn:
The trie accumulator uses a snapshot of the peer list provided by the block synchronizer through the global state synchronizer in order to get the required tries.

Because we didn't provide the block synchronizer any feedback on what the peers that dind't have the global state data were, the block synchronizer may re-issue the global state sync request with the same peer set as for the previous failed attempt. This may lead to delays in getting the global state data.

This PR adds:
* the ability to propagate the unreliable peers (that didn't have the requested data) upwards to the global state synchronizer and then to the block synchronizer. This gives feedback that the original peer list provided should be changed if another global sync request is made.
* block synchronizer now demotes the peers unreliable received from the global state synchronizer
* a test to verify that the peer list is changed based on the feedback from the global state sync components.
* slight change into how `PeerQuality` is handled when demoting and promoting peers; now peers with `Reliable` or `Unreliable` quality will not be set as `Unknown` when demoting or promoting respectively.

These changes reduce the risk that the global state synchronizer gets stuck asking the same peers all over again.

Further potential improvements that can be made:
* Promote peers that were successful when downloading global state. We only demote the peers now.
* Use a reputation based peer list (similar to the one used in the block synchronizer) in the trie accumulator rather than going through peers one by one.

Fixes: https://github.com/casper-network/casper-node/issues/3563


alsrdn:
bors try

node/src/components/block_synchronizer.rs
```diff
@@ -887,21 +894,43 @@ impl BlockSynchronizer {
...
+            } else {
+                if let Some(root_hash) = maybe_root_hash {
+                    if let Err(error) = builder.register_global_state(root_hash) {
+                        error!(%block_hash, %error, "BlockSynchronizer: failed to apply global state");
+                    }
+                }
+                // Demote all the peers where we didn't find the required global state tries
+                for peer in unreliable_peers.iter() {
+                    builder.demote_peer(Some(*peer));
+                }
```
rafal-ch, line 934: Shouldn't we demote peers unconditionally (i.e., before the `if let` in [line 921](https://github.com/casper-network/casper-node/pull/3594/files#diff-ad7e9cbef898bd82de2888993921f091b4dbcec0c2445eb40ea244873b296aedR921))?

node/src/components/block_synchronizer/peer_list.rs
```diff
@@ -110,13 +110,10 @@ impl PeerList {
                     // no change
                 }
                 Entry::Occupied(entry) => match entry.get() {
-                    PeerQuality::Dishonest | PeerQuality::Unknown => {
+                    PeerQuality::Dishonest | PeerQuality::Unreliable => {
```
rafal-ch, line 113: Can you consider adding a UT for `demote_peer()`, `promote_peer()`, et al., just to have a clear picture of how the functions behave.

node/src/components/block_synchronizer/peer_list.rs
```diff
@@ -148,34 +145,40 @@ impl PeerList {
         PeersStatus::Sufficient
     }
 
-    pub(super) fn qualified_peers(&self, rng: &mut NodeRng) -> Vec<NodeId> {
-        let up_to = self.max_simultaneous_peers as usize;
-
-        // get most useful up to limit
-        let mut peers: Vec<NodeId> = self
-            .peer_list
+    fn get_peers_by_quality(
```
rafal-ch:
```suggestion
    fn get_random_peers_by_quality(
```

node/src/components/block_synchronizer/peer_list.rs
```diff
@@ -148,34 +145,40 @@ impl PeerList {
...
-
-            peers.extend(better_than_nothing);
+            peers.extend(self.get_peers_by_quality(rng, up_to, PeerQuality::Unknown));
         }
+
+        // if still below limit try unreliable peers again until we have the chance to refresh the
+        // peer list
+        let missing = up_to.saturating_sub(peers.len());
+        if missing > 0 {
+            peers.extend(self.get_peers_by_quality(rng, up_to, PeerQuality::Unreliable));
```
rafal-ch: Should we also try to get peers of `Unknown` quality here?

node/src/components/block_synchronizer/tests.rs
```diff
@@ -0,0 +1,212 @@
...
+    assert_eq!(
+        global_sync_request.state_root_hash,
+        *block.state_root_hash()
+    );
+
+    global_sync_request.peers
+}
+
+#[tokio::test]
+async fn global_state_sync_wont_stall_with_bad_peers() {
```
rafal-ch, line 95: Nice test! :+1:

node/src/components/block_synchronizer.rs
alsrdn: Global state should be synced only on historical blocks and we'll still need to match the block hash in the response to the one from the builder because the builder provides the peers in this case.

node/src/components/block_synchronizer/peer_list.rs
alsrdn: Yes, good idea!

node/src/components/block_synchronizer/peer_list.rs
alsrdn: Yes, the intention was to first get `Reliable` peers, then if those are not enough extend the peers list and get `Unknown` peers and even if then we don't have enough peers, get unreliable ones. But I did a mistake here and used `up_to` in all of the calls instead of using `missing`. Good catch! I'll fix this.

alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3594/commits/1122fd2b19b2adcc12ecedf7c8d00363146ca201

node/src/components/block_synchronizer/peer_list.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/pull/3594/commits/1122fd2b19b2adcc12ecedf7c8d00363146ca201

node/src/components/block_synchronizer/peer_list.rs
alsrdn: Added in https://github.com/casper-network/casper-node/pull/3594/commits/1122fd2b19b2adcc12ecedf7c8d00363146ca201

rafal-ch (approved):
LGTM :+1:

node/src/components/block_synchronizer/global_state_synchronizer.rs
```diff
@@ -32,21 +32,44 @@ const COMPONENT_NAME: &str = "global_state_synchronizer";
 
 #[derive(Debug, Clone, Error)]
 pub(crate) enum Error {
-    #[error(transparent)]
-    TrieAccumulator(TrieAccumulatorError),
-    #[error("ContractRuntime failed to put a trie into global state: {0}")]
-    PutTrie(engine_state::Error),
+    #[error("trie accumulator encountered an error while fetching a trie; unreliable peers {0:?}")]
```
Fraser999: I think we should avoid mixing debug output into display output.  There's `utils::DisplayIter` which could probably help here.

Same comment for the `PutTrie` variant below.

Fraser999 (commented):
One minor change requested, otherwise LGTM.

node/src/components/block_synchronizer/global_state_synchronizer.rs
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3594/commits/18b802deab67b7404bb7570a085b695e73a9b8f1

Fraser999 (approved):

node/src/components/block_synchronizer/global_state_synchronizer.rs
```diff
@@ -25,28 +25,52 @@ use crate::{
...
+
+    pub(crate) fn hash(&self) -> &Digest {
+        &self.hash
+    }
+
+    pub(crate) fn unreliable_peers(self) -> Vec<NodeId> {
+        self.unreliable_peers
+    }
+}
+
```
EdHastingsCasperAssociation, line 66: you could do a two state enum for Success / Failure, and flatten the Error as a field on the failure variant; this would allow places that return Result<Response, Error> to be simplified to just return the enum variant (etc).

node/src/components/block_synchronizer/peer_list/tests.rs
```diff
@@ -0,0 +1,118 @@
...
+    let mut peer_list = PeerList::new(5, TimeDiff::from_seconds(1));
+    let test_peer = NodeId::random(&mut rng);
+
+    peer_list.register_peer(test_peer);
+    assert!(peer_list.is_peer_unknown(&test_peer));
+    peer_list.promote_peer(Some(test_peer));
+    assert!(peer_list.is_peer_reliable(&test_peer));
+}
+
+#[test]
```
EdHastingsCasperAssociation, line 80: might as well add the unreliable_peer_remains_unreliable_if_demoted test case

EdHastingsCasperAssociation (approved):
a nit and a suggestion to consider, but LGTM, gj

EdHastingsCasperAssociation:
bors try

node/src/components/block_synchronizer/global_state_synchronizer.rs
alsrdn: Yes, we could to that though I'm more partial to having separate types that are clearly delimited. What I think we can do is have a trait implemented for both the `Response` and `Error` that can extract the unreliable peers which would make the code looking at the response type potentially more nicer. WDYT?

node/src/components/block_synchronizer/peer_list/tests.rs
alsrdn: Added in https://github.com/casper-network/casper-node/pull/3594/commits/baa3461a4c68e9ec8e04d791e84710485fa18c9e

alsrdn:
bors try

alsrdn:
bors merge

PR #3595, opened 2023-01-22
Add tests ensuring that finality signatures propagate on network

rafal-ch:
This PR adds two tests that check if the finality signatures are indeed propagated through the network.

Scenario in both tests is identical:
1. run a 2 node network
2. schedule an upgrade
3. wait until the nodes exit for upgrade with `ExitCode::Success`

The difference is in the stakes used for setting up a network. In first scenario the stakes are equal and in the other one node has a dominant stake.

The test where the stakes are equal should fail without changes made in [this commit](https://github.com/casper-network/casper-node/pull/3592/commits/33a79b8ace6485aafea11687c06b19ad98d06018?diff=split&w=1), because signatures from both participants are needed for the node to decide it can shutdown for upgrade.

In case we break finality signatures propagation we should be able to detect this with these tests instead on relying on the evanescent flakiness of `dont_upgrade_without_switch_block()` test which uses random stakes.

### DIFF
--- a/node/src/reactor/main_reactor/tests.rs
+++ b/node/src/reactor/main_reactor/tests.rs
@@ -65,15 +65,25 @@ impl TestChain {
     /// Instantiates a new test chain configuration.
     ///
     /// Generates secret keys for `size` validators and creates a matching chainspec.
-    fn new(rng: &mut TestRng, size: usize) -> Self {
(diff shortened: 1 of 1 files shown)


georgepisaltu (approved):

alsrdn (approved):

rafal-ch:
bors try

PR #3596, opened 2023-01-22
Update for vulnerability in `libgit2-sys`

rafal-ch:
As reported by `cargo audit`:

```
Crate:         libgit2-sys
Version:       0.14.0+1.5.0
Title:         git2 does not verify SSH keys by default
Date:          2023-01-20
ID:            RUSTSEC-2023-0003
URL:           https://rustsec.org/advisories/RUSTSEC-2023-0003
Solution:      Upgrade to >=0.13.5, <0.14.0 OR >=0.14.2
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,7 +104,7 @@ checksum = "a941c39708478e8eea39243b5983f1c42d2717b3620ee91f4a52115fd02ac43f"
 dependencies = [
  "itertools 0.9.0",
  "proc-macro-error",
- "proc-macro2 1.0.49",
+ "proc-macro2 1.0.50",
  "quote 1.0.23",
  "syn 1.0.107",
 ]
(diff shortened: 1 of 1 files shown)


marc-casperlabs:
So, as far as I can tell, we only use libgit2 as a transitive dependency of `vergen`, which means no SSH is ever involved.

This PR seems to update many crates on the patch level, not just libgit2 - as a result, we also pull in a new dependency, since `reqwests` uses it (only used in casper-test, I think).

Is this something we want to do? In general, I don't see a super big issue with it, just wanted to point out that this is more than just a single crate upgrade.

Fraser999 (approved):

Fraser999:
@marc-casperlabs 
> Is this something we want to do? In general, I don't see a super big issue with it, just wanted to point out that this is more than just a single crate upgrade.

Well, I think we normally just update all deps when we need to update any.  I don't see any advantage right now of not doing so (although that could change in the future).

rafal-ch:
> So, as far as I can tell, we only use libgit2 as a transitive dependency of `vergen`, which means no SSH is ever involved.
> 
> This PR seems to update many crates on the patch level, not just libgit2 - as a result, we also pull in a new dependency, since `reqwests` uses it (only used in casper-test, I think).
> 
> Is this something we want to do? In general, I don't see a super big issue with it, just wanted to point out that this is more than just a single crate upgrade.

I think we should be good with keeping all deps up-to-date with `cargo update`. However, I got your point that I could be more specific about what actually got updated in the process.

PR #3591, opened 2023-01-19
Add more detailed logs to `register_block_header()`

rafal-ch:
This PR adds some more details to logs created by `register_block_header()` to ease up debugging.

When registering a header, instead of such message:
```
BlockAcquisition: registering header for: block hash 3fbe..6f3d
```

we'll see full block hash and a height:
```
BlockAcquisition: registering header for: BlockHash(3fbe346cc015db6b9081e3ce475346651c1497ac1c02fbe9187fd0bb2bb26f3d), height: 42
```

PR #3604, opened 2023-01-24
Fix gossip target

Fraser999:
This PR fixes an issue where the target of gossip messages was not being respected.

As part of the initial approach, it was necessary to separate the `Item` trait into the `GossipItem` and `FetchItem` traits.  However, providing the `SmallGossipItem` trait as a means for accessing the full data from a gossiped ID proved to be a better approach, and so splitting the `Item` trait wasn't strictly required, but is left in the PR as I feel it's an improvement nonetheless.

I also took the opportunity to remove the boxed fn from the gossiper (used to get data from storage), replacing it with an `ItemProvider` trait which can be specialized for the specific type of data being retrieved.

Finally, the PR also contains a couple of trivial improvements to the logging of versions on node startup.

Related to #3578.


node/src/components/block_synchronizer/block_acquisition.rs
```diff
@@ -389,17 +389,17 @@ impl BlockAcquisitionState {
     ) -> Result<Option<Acceptance>, BlockAcquisitionError> {
         let new_state = match self {
             BlockAcquisitionState::Initialized(block_hash, signatures) => {
-                if header.id() == *block_hash {
+                if header.block_hash() == *block_hash {
                     info!(
-                        "BlockAcquisition: registering header for: {:?}, height: {}",
```
rafal-ch, line 394: Debug output for block hash has been deliberately enabled in [this PR](https://github.com/casper-network/casper-node/pull/3591/files). The reason was, IIRC, that in one of the block acquisition flows the full block hash was nowhere to be found and it turned out to be needed in debugging the sync issues.

Consider leaving it as-is, at least until we stabilize the sync process.

node/src/components/fetcher/fetch_item.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// The type of the metadata provided when validating the item.
+    type ValidationMetadata: Eq + Clone + Serialize + Debug + DataSize + Send;
+
+    /// The tag representing the type of the item.
+    const TAG: Tag;
+
+    /// The ID of the specific item.
+    fn fetch_id(&self) -> Self::Id;
+
+    /// Checks cryptographic validity of the item, and returns an error if invalid.
```
rafal-ch: I believe it's just copy&pasted, but:

```suggestion
    /// Checks validity of the item, and returns an error if invalid.
```

That's because `validate()` no longer focuses solely on cryptographic verification.

node/src/components/gossiper/tests.rs
```diff
@@ -47,7 +47,7 @@ use crate::{
         network::{NetworkedReactor, TestingNetwork},
         ConditionCheckReactor, FakeDeployAcceptor,
     },
-    types::{Chainspec, ChainspecRawBytes, Deploy, FinalitySignature, NodeId},
+    types::{Block, Chainspec, ChainspecRawBytes, Deploy, FinalitySignature, NodeId},
```
rafal-ch, line 50: This comment applies to the gossiper tests in general, not this particular line.

Would it pay off to add a test that explicitly checks that different gossip targets are actually respected?

node/src/components/gossiper/provider_impls/block_provider.rs
```diff
@@ -0,0 +1,24 @@
...
+impl ItemProvider<Block> for Gossiper<{ Block::ID_IS_COMPLETE_ITEM }, Block> {
+    fn get_from_storage<REv: From<StorageRequest> + Send>(
+        effect_builder: EffectBuilder<REv>,
+        item_id: BlockHash,
+        requester: NodeId,
+    ) -> Effects<Event<Block>> {
+        effect_builder
+            .get_block_from_storage(item_id)
+            .event(move |results| {
+                let result = results.ok_or_else(|| String::from("block not found"));
```
rafal-ch, line 16: Nit:

Can we add an `item_id` to the error message (like in the `deploy_provider.rs`)?

node/src/components/gossiper/provider_impls/finality_signature_provider.rs
```diff
@@ -0,0 +1,26 @@
...
+{
+    fn get_from_storage<REv: From<StorageRequest> + Send>(
+        effect_builder: EffectBuilder<REv>,
+        item_id: FinalitySignatureId,
+        requester: NodeId,
+    ) -> Effects<Event<FinalitySignature>> {
+        effect_builder
+            .get_finality_signature_from_storage(item_id.clone())
+            .event(move |results| {
+                let result = results.ok_or_else(|| String::from("finality signature not found"));
```
rafal-ch, line 18: Another nit:

Can we add an item_id to the error message (like in the deploy_provider.rs)?

rafal-ch (approved):
No blockers, just few nits :+1:

node/src/components/gossiper/provider_impls/block_provider.rs
Fraser999: I'll actually remove it from the deploy provider as the item ID is included in an error message when we handle the result in `Gossiper::failed_to_get_from_holder`

node/src/components/gossiper/provider_impls/finality_signature_provider.rs
Fraser999: As above

node/src/components/block_synchronizer/block_acquisition.rs
Fraser999: Fixed in [f0568e1](https://github.com/casper-network/casper-node/pull/3604/commits/f0568e1d2e026d4ddea247f47ee6cf34cdbd7418).

node/src/components/fetcher/fetch_item.rs
Fraser999: Fixed in [f0568e1](https://github.com/casper-network/casper-node/pull/3604/commits/f0568e1d2e026d4ddea247f47ee6cf34cdbd7418).

node/src/components/gossiper/tests.rs
Fraser999: Fixed in [f0568e1](https://github.com/casper-network/casper-node/pull/3604/commits/f0568e1d2e026d4ddea247f47ee6cf34cdbd7418) (I added a couple of checks in the existing test reactor that the expected target type is provided).

node/src/components/gossiper/provider_impls/block_provider.rs
Fraser999: Fixed in [f0568e1](https://github.com/casper-network/casper-node/pull/3604/commits/f0568e1d2e026d4ddea247f47ee6cf34cdbd7418).

Fraser999:
bors merge

Fraser999:
bors try

node/src/components/network.rs
```diff
@@ -437,54 +440,37 @@ where
...
-            // TODO - set this to `warn!` once we are normally testing with networks large enough to
-            //        make it a meaningful and infrequent log message.
-            trace!(
-                our_id=%self.context.our_id(),
-                wanted = count,
-                selected = peer_ids.len(),
-                "could not select enough random nodes for gossiping, not enough non-excluded \
-                outgoing connections"
-            );
+            let connected = self.outgoing_manager.connected_peers().count();
```
rafal-ch, line 457: Nits:

* We can move this inside the `if not_excluded > 0 {`.
* Also, `choose_gossip_peers()` could return the `not_excluded`, since it already filters the connected_peers.

These are just a micro-optimizations, if at all, so feel free to leave as is.

node/src/components/network.rs
```diff
@@ -437,54 +440,37 @@ where
...
+            let not_excluded = self
+                .outgoing_manager
+                .connected_peers()
+                .filter(|peer_id| !exclude.contains(peer_id))
+                .count();
+            if not_excluded > 0 {
```
rafal-ch, line 463: So if everyone is excluded we don't want to see the debug message?

node/src/components/network.rs
```diff
@@ -1346,3 +1369,290 @@ where
...
+
+        fn num_non_validators<'a>(&self, input: impl Iterator<Item = &'a NodeId>) -> usize {
+            input
+                .filter(move |&node_id| self.non_validators.contains(node_id))
+                .count()
+        }
+    }
+
+    #[test]
+    fn should_choose_mixed() {
```
rafal-ch, line 1435: I'd consider adding one test to make sure we grab a correct number of peers from a set (validators or non-validators).

Currently, we may change the parameter in [this line](https://github.com/casper-network/casper-node/pull/3604/commits/d7b97474f66dd168bd10bb1dce14e593df15f405#diff-ced0e709fad62f0c013bb3d691fb1082aface83249abccb628224b31e62bec65R1096) to `count - 1` and the incorrect behavior will get hidden by `interleave()`, which is "saturating".

Here's how such test may look like:

```
        // Choose 3 from all peers, exclude all non-validators, should return 3 validators,
        let exclude: HashSet<_> = fixture.non_validators.iter().copied().collect();
        let chosen = choose_gossip_peers(
            &mut rng,
            TARGET,
            3,
            exclude,
            fixture.all_peers.iter().copied(),
            fixture.is_validator_in_era(),
        );
        assert_eq!(chosen.len(), 3);
        assert_eq!(fixture.num_validators(chosen.iter()), 3);
```

It should fail 50% of the time if we happen to mess-up the `choose_multiple` in the future :)

rafal-ch:
Left a few nits, but still LGTM :+1:

node/src/components/fetcher.rs
```diff
@@ -46,15 +50,15 @@ const COMPONENT_NAME: &str = "fetcher";
 #[derive(DataSize, Debug)]
 pub(crate) struct Fetcher<T>
 where
-    T: FetcherItem,
+    T: FetchItem,
```
EdHastingsCasperLabs, line 53: curious about why the name change?

EdHastingsCasperAssociation (approved):

node/src/components/network.rs
Fraser999: Yes, since this happens very frequently when using nctl defaults.

PR #3614, opened 2023-01-27
Increase cost for `withdraw_bid` and `undelegate`

georgepisaltu:
This PR increases the cost for `withdraw_bid` and `undelegate` to match their counterparts, `add_bid` and `delegate`. It also changes the values we use in testing with updated costs to use these new values, the default ones staying at 10000 motes in regular testing.


Fraser999 (commented):
Can we update the cost table in the resources/local/chainspec.toml.in file too please?

Fraser999 (approved):

PR #3615, opened 2023-01-27
Add more logging for historical sync leap

alsrdn:
Add some extra debug logging to the historical sync leap process to help with debugging https://github.com/casper-network/casper-node/issues/3598

### DIFF
--- a/node/src/types/validator_matrix.rs
+++ b/node/src/types/validator_matrix.rs
@@ -220,8 +220,12 @@ impl ValidatorMatrix {
         None
     }
 
-    pub(crate) fn read_inner(&self) -> RwLockReadGuard<BTreeMap<EraId, EraValidatorWeights>> {
+    fn read_inner(&self) -> RwLockReadGuard<BTreeMap<EraId, EraValidatorWeights>> {
         self.inner.read().unwrap()
+    }
+
+    pub(crate) fn eras(&self) -> Vec<EraId> {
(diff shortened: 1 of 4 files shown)


rafal-ch (approved):
LGTM :+1:

node/src/types/validator_matrix.rs
```diff
@@ -220,7 +220,7 @@ impl ValidatorMatrix {
         None
     }
 
-    fn read_inner(&self) -> RwLockReadGuard<BTreeMap<EraId, EraValidatorWeights>> {
+    pub(crate) fn read_inner(&self) -> RwLockReadGuard<BTreeMap<EraId, EraValidatorWeights>> {
```
Fraser999: I strongly think we should avoid doing this - providing a way to return (and hence hold) a lock makes it more likely to cause a deadlock.

Whatever functionality is required, we should ensure the validator matrix can provide it without returning a lock.

Fraser999 (changes requested):
One blocker, but otherwise LGTM.

node/src/types/validator_matrix.rs
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3615/commits/4bfa012fde7102637b1f53b95605ae265b53774f

Fraser999 (approved):

Issue #3583, opened 2023-01-18
Using Mainnet data joining node not  progressing historical syncing while in keepup


cspramit:
Issue: Using mainnet data when one node is joined to the network it is keepingup but unable to progress on historical sync. We see a checksum warning as well( see below) but that is expected

Name: ajith-mainnet-clone
Network: 5 + 2
Branch: dev
Chain: public
Commit: 17ef14ddd9ac426ffed45b5ace8e7e700ebeeb1c
Config Override: None
Joining Node Testing Mainnet Snapshot ( 1034499 )
Tests:
1 x sync_to_genesis joining
Results: Transitioned to KeepUp but not progressing with historical sync
Observations:

- **joining node transitioned to KeepUp and in-sync with tip

- status - block_sync ( not progressed since 1+hrs)**

- "acquisition_state": "have global state(1034492) for: block hash 01c9..3011"
seeing events as 'checksum uncheckable execution results'
{"timestamp":"2023-01-18T04:10:21.664255Z","level":"INFO","fields":{"message":"BlockSynchronizer: need next for block hash 95bd..31b9: execution results for block hash 95bd..31b9 or chunk #0 (checksum uncheckable execution results)"},"target":"casper_node::components::block_synchronizer"}
"block_sync": {
    "historical": {
      "block_hash": "95bda7601307f5d1d954bae9938f6ee70121bd2ef90077d9daddb33c2e4731b9",
      "block_height": 824808,
      "acquisition_state": "have global state(824808) for: block hash 95bd..31b9"
    },
    "forward": null
  }
Dumps: http://genesis.casperlabs.io/ajith-mainnet-clone/casper-node-dumps/ajith-mainnet-clone/18012023_0405/dump_download_list.txt

cspramit:
@thamps-casp for visibility

rafal-ch:
Keeping the issue as Open since the fix should be confirmed by @thamps-casp

rafal-ch:
The DB related issue was fixed here https://github.com/casper-network/casper-node/pull/3665.

The issue with node being stuck in global state sync is tracked in different ticket: https://github.com/casper-network/casper-node/issues/3659

PR #3630, opened 2023-01-31
Add debug logs around fetching execution results

rafal-ch:
This PR adds more detailed logging in order to analyze the fetching of execution results.

It should be reverted once the root-cause of https://github.com/casper-network/casper-node/issues/3583 is identified and the problem is fixed.


alsrdn (approved):

rafal-ch:
Adding `do-not-merge` since most likely these logs won't be needed.

rafal-ch:
Closing as no longer needed.

PR #3636, opened 2023-02-02
Notify consensus of previously executed blocks in restart scenarios

### LINKED ISSUE #3618: ITST14 test failing

georgepisaltu:
Fixes #3618

This PR adds a new announcement for finalized blocks enqueued for execution, but which contract runtime will not execute because their height is lower than the next expected executable height. In this scenario, contract runtime will raise this announcement which will be handled by the reactor. If there is a complete block in storage with that same height, it means the block can be safely considered executed and the reactor raises an event letting consensus know that the finalized block it had enqueued was handled, the same way it would happen if the block was executed. Consensus is oblivious to the fact that execution didn't take place.

We do this to ensure the smooth operation of the consensus validating process in the case of a restart in the middle of an era, where the restarted node couldn't make any consensus proposals until all the previous blocks in the era along with their vertices were "executed".

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1571,6 +1571,27 @@ impl Storage {
             .transpose()
     }
 
+    /// Retrieves a single complete block header by height by looking it up in
+    /// the index and returning it.
+    pub fn read_complete_block_header_by_height(
+        &self,
+        height: u64,
+    ) -> Result<Option<BlockHeader>, FatalStorageError> {
+        let mut txn = self.env.begin_ro_txn()?;
+        let res = self
+            .block_height_index
+            .get(&height)
+            .and_then(|block_hash| {
+                self.get_single_block_header(&mut txn, block_hash)
+                    .transpose()
+            })
+            .transpose();
+        if !(self.should_return_block(height, true)?) {
+            return Ok(None);
+        }
+        res
+    }
+
     /// Retrieves a single block header by hash.
     pub fn read_block_header(
         &self,
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1603,8 +1603,8 @@ impl Storage {
             .transpose()
     }
 
-    /// Retrieves a single complete block header by height by looking it up in
-    /// the index and returning it.
+    /// Retrieves a single block header by height from the available block
+    /// range by looking it up in the index and returning it.
     pub fn read_complete_block_header_by_height(
         &self,
         height: u64,
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1621,12 +1621,11 @@ impl Storage {
         if !(self.should_return_block(height, true)?) {
             return Ok(None);
         }
-        if self
+        if !self
             .completed_blocks
             .sequences()
             .iter()
-            .find(|sequence| sequence.low() <= height && height <= sequence.high())
-            .is_none()
+            .any(|sequence| sequence.low() <= height && height <= sequence.high())
         {
             return Ok(None);
         }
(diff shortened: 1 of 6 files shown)


node/src/components/storage.rs
```diff
@@ -1571,6 +1571,27 @@ impl Storage {
             .transpose()
     }
 
+    /// Retrieves a single complete block header by height by looking it up in
+    /// the index and returning it.
+    pub fn read_complete_block_header_by_height(
```
rafal-ch, line 1608: Would it be possible to use the already existing [get_block_header_by_height_restricted()](https://github.com/casper-network/casper-node/blob/a54d19ba1e4484d26961a1f48da3ba8bffa38322/node/src/components/storage.rs#L1948)?

node/src/reactor/main_reactor.rs
```diff
@@ -448,6 +448,42 @@ impl reactor::Reactor for MainReactor {
...
+                    // required eras and attempts to retrace its steps in the
+                    // era by enqueuing all finalized blocks starting from the
+                    // first one in that era, blocks which should have already
+                    // been executed and marked complete in storage.
+                    warn!(
+                        "Finalized block with height {} enqueued for execution, \
+                        but a complete block header with the same height is \
+                        not present in storage.",
+                        block_height
+                    );
```
rafal-ch: Nit:

Consider logging using this method:

```suggestion
                    warn!(
                        block_height,
                        "Finalized block enqueued for execution, but a complete \
                        block header with the same height is not present in storage."
                    );
```

node/src/components/storage.rs
georgepisaltu: It's certainly possible, but I didn't want to do that mainly because `get_block_header_by_height_restricted` looks like a function which may change in the future (I'm talking about the restrictions) and I wanted something explicit here. I have no strong opinion on it though, we can safely use `get_block_header_by_height_restricted` if it's better.

node/src/reactor/main_reactor.rs
georgepisaltu: Will do.

node/src/components/storage.rs
georgepisaltu: Forgot to mention in my previous comment, `get_block_header_by_height_restricted` also requires a transaction instance (which is probably why it's not a `pub fn`) and I don't want to create those outside of storage.

rafal-ch: Yes, I agree there are differences. Despite not liking this code duplication much, I'm not strong about it. So let it be :)

Anyway, I also see that the only restriction we now have is `only_from_available_block_range`, which is set to `true` in all the flows. This is probably a remnant of the previous implementation of fast sync and we can as well get rid of it, hitting two birds with one stone. That'd be something to consider as a separate ticket (cc: @Fraser999)

fizyk20 (approved):
Looks good :+1: (although I do like Rafał's suggestion about logging :wink:)

node/src/reactor/main_reactor.rs
Fraser999: I'd actually consider this an case worthy of an error level log.

node/src/components/storage.rs
Fraser999: Happy for that suggestion to get folded into this PR, or if not @georgepisaltu let me know and we can create a ticket request.

Fraser999 (approved):

node/src/reactor/main_reactor.rs
georgepisaltu: Done.

node/src/components/storage.rs
georgepisaltu: We will need to create a separate ticket for this cleanup. I tried to do it but it's not as simple as it looks because "only from available block range" is not the same as "complete". There are multiple usages of these functions in storage and I believe they are due for a cleanup, but it would take more investment than it would make sense for the issue this PR solves.

Fraser999 (approved):

PR #3637, opened 2023-02-02
Harden gossiper against malicious messages

### LINKED ISSUE #3572: Avoid starting to gossip data which is already stored locally

Fraser999:
This PR provides basic protection against wasting resources when receiving unexpected gossiper messages.

For an new incoming `Message::Gossip`, no action will be taken if the given item is not already being handled by the gossiper and if the item is already held in storage other than sending a `Message::GossipResponse`.  Potentially even sending the response could be omitted, but this might be overkill given how cheap that is to send and considering that the sender might not be malicious.

For all other incoming message variants, if the given item is not already being handled by the gossiper then no action is taken.

Additionally, there is now a timeout associated with handling a `Message::Item`.  The expectation in this flow is that after making the item body announcement (i.e. after calling `EffectBuilder::announce_item_body_received_via_gossip()`) then some other component will handle validating and storing the item, and will then re-engage the gossiper by creating an `ItemReceived` event.  However, if validation or storing fails, then the gossiper is not notified.  This new timeout ensures that in such a case the entry in the gossip table is force-finished and will subsequently be purged.

Closes #3572.

### DIFF
--- a/node/src/components/gossiper/config.rs
+++ b/node/src/components/gossiper/config.rs
@@ -18,12 +18,15 @@ pub(super) const MAX_SATURATION_LIMIT_PERCENT: u8 = 99;
 pub(super) const DEFAULT_FINISHED_ENTRY_DURATION: &str = "60sec";
 const DEFAULT_GOSSIP_REQUEST_TIMEOUT: &str = "10sec";
 const DEFAULT_GET_REMAINDER_TIMEOUT: &str = "60sec";
+const DEFAULT_VALIDATE_AND_STORE_TIMEOUT: &str = "60sec";
 #[cfg(test)]
 const SMALL_TIMEOUTS_FINISHED_ENTRY_DURATION: &str = "2sec";
 #[cfg(test)]
 const SMALL_TIMEOUTS_GOSSIP_REQUEST_TIMEOUT: &str = "1sec";
 #[cfg(test)]
 const SMALL_TIMEOUTS_GET_REMAINDER_TIMEOUT: &str = "1sec";
+#[cfg(test)]
+const SMALL_TIMEOUTS_VALIDATE_AND_STORE_TIMEOUT: &str = "1sec";
 
 /// Configuration options for gossiping.
 #[derive(Copy, Clone, DataSize, Debug, Deserialize, Serialize)]
@@ -50,6 +53,9 @@ pub struct Config {
     /// The timeout duration in seconds for retrieving the remaining part(s) of newly-discovered
     /// data from a peer which gossiped information about that data to this node.
     get_remainder_timeout: TimeDiff,
+    /// The timeout duration for a newly-received, gossiped item to be validated and stored by
+    /// another component before the gossiper abandons waiting to gossip the item onwards.
+    validate_and_store_timeout: TimeDiff,
 }
 
 impl Config {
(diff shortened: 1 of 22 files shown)


node/src/components/gossiper/config.rs
```diff
@@ -18,12 +18,15 @@ pub(super) const MAX_SATURATION_LIMIT_PERCENT: u8 = 99;
...
 const DEFAULT_GET_REMAINDER_TIMEOUT: &str = "60sec";
+const DEFAULT_VALIDATE_AND_STORE_TIMEOUT: &str = "60sec";
 #[cfg(test)]
 const SMALL_TIMEOUTS_FINISHED_ENTRY_DURATION: &str = "2sec";
 #[cfg(test)]
 const SMALL_TIMEOUTS_GOSSIP_REQUEST_TIMEOUT: &str = "1sec";
 #[cfg(test)]
 const SMALL_TIMEOUTS_GET_REMAINDER_TIMEOUT: &str = "1sec";
+#[cfg(test)]
+const SMALL_TIMEOUTS_VALIDATE_AND_STORE_TIMEOUT: &str = "1sec";
```
georgepisaltu, line 29: This seems rather optimistic. What if event queues are swamped? Even for addresses, which don't need validation, we might not get there in 1 sec. Then the data is stored eventually but not gossiped onwards, right?

node/src/components/gossiper/provider_impls/block_provider.rs
```diff
@@ -1,24 +1,24 @@
...
 use crate::{
-    components::gossiper::{Event, GossipItem, Gossiper, ItemProvider},
-    effect::{requests::StorageRequest, EffectBuilder, EffectExt, Effects},
-    types::{Block, BlockHash, NodeId},
+    components::gossiper::{GossipItem, Gossiper, ItemProvider},
+    effect::{requests::StorageRequest, EffectBuilder},
+    types::{Block, BlockHash},
 };
 
+#[async_trait]
```
georgepisaltu, line 9: I'm not against it, but I'm curious why the move towards async functions here. It's breaking the pattern of using and returning `Effects`.

georgepisaltu (approved):
LGTM, just a couple of questions.

node/src/components/gossiper/tests.rs
```diff
@@ -532,3 +533,223 @@ async fn should_timeout_gossip_response() {
 
     NetworkController::<NodeMessage>::remove_active();
 }
+
+#[tokio::test]
+async fn should_timeout_new_item_from_peer() {
```
rafal-ch, line 538: Nit-like suggestion for one more test :)

Please consider adding a similar one which ensures that `CheckItemReceivedTimeout` is **not** triggered before `VALIDATE_AND_STORE_TIMEOUT` time passes.

rafal-ch (approved):
No blockers :+1:

node/src/components/gossiper/config.rs
Fraser999: Yes, but this is used in the test-only configuration created via `Config::new_with_small_timeouts`.  I guess if start seeing some of the related tests failing, we can increase it.  But I wouldn't expect tests to exceed one second to validate and store data.

node/src/components/gossiper/provider_impls/block_provider.rs
Fraser999: Initially it was because I thought we'd have a couple of different places where we'd be calling `is_stored` with a different flow for each after the result is yielded.  Having a specific event chained to the end of the trait method made that clunky - I was looking at adding a third trait method which effectively also checked for the data in storage but went to different event on returning.

Turns out that wasn't actually needed, but I didn't change this back since it avoids duplication of the `.event(move |result| { Event::IsStoredResult { ... }})` in each of the provider impls.

node/src/components/gossiper/tests.rs
Fraser999: As discussed offline, the pre-exisiting tests cover this case.

Fraser999:
bors merge

PR #3639, opened 2023-02-02
Fix sync to genesis after upgrade from 1.4.x to 1.5

### LINKED ISSUE #3631: Fix broken historical sync to blocks in Era 0 when upgrading from 1.4.8 to 1.5.0
Currently a node joining the network after an upgrade from 1.4.8 to 1.5.0 cannot fully sync to genesis. This happens because a SyncLeap requested for blocks 0-10 which are in Era 0 will not be returned since block 0 is not a switch block in 1.4.8. Without knowing the validator matrix for era 0, the joining node cannot sync the required blocks.

alsrdn:
* Allow for a sync leap to be returned for era 0 if the it contains the genesis block. Even though the sync leap does not contain a switch block, the block headers will still be included in the response.
* Infer validator weights for era 0 when registering the weights for era 1 since they should be the same given that the weights don't change until the unbonding delay elapsed.
* Change the nctl upgrade test to wait and check that the nodes have synced to genesis.

Fixes: https://github.com/casper-network/casper-node/issues/3631

### DIFF
--- a/node/src/reactor/main_reactor/keep_up.rs
+++ b/node/src/reactor/main_reactor/keep_up.rs
@@ -553,7 +553,10 @@ impl MainReactor {
         for evw in era_validator_weights {
             let era_id = evw.era_id();
             debug!(%era_id, "historical: attempt to register validators for era");
-            if self.validator_matrix.register_era_validator_weights(evw) {
+            if self
+                .validator_matrix
+                .register_era_validator_weights_and_infer_era_0(evw)
+            {
                 info!(%era_id, "historical: got era");
             } else {
                 debug!(%era_id, "historical: era already present or is not relevant");
@@ -667,18 +670,24 @@ impl MainReactor {
                     }
                     Ok(None) => {
                         debug!(%parent_hash, "historical: did not find block header in storage");
-                        match block_header.era_id().predecessor() {
+                        let maybe_sync_era = if block_header.era_id() == EraId::from(0) {
+                            // if the block is in era 0 its parent can only be in era 0
+                            Some(EraId::from(0))
+                        } else {
                             // we do not have the parent header and thus don't know what era
                             // the parent block is in (it could be the same era or the previous
                             // era). we assume the worst case and ask
                             // for the earlier era's proof
-                            Some(previous_era_id) => Ok(Some(SyncBackInstruction::Sync {
+                            block_header.era_id().predecessor()
+                        };
+
+                        maybe_sync_era.map_or(Ok(None), |era_id| {
+                            Ok(Some(SyncBackInstruction::Sync {
                                 parent_hash: *parent_hash,
                                 maybe_parent_metadata: None,
-                                era_id: previous_era_id,
-                            })),
-                            None => Ok(None),
-                        }
+                                era_id,
+                            }))
+                        })
                     }
                     Err(err) => Err(err.to_string()),
                 }
--- a/node/src/reactor/main_reactor/keep_up.rs
+++ b/node/src/reactor/main_reactor/keep_up.rs
@@ -670,24 +670,24 @@ impl MainReactor {
                     }
                     Ok(None) => {
                         debug!(%parent_hash, "historical: did not find block header in storage");
-                        let maybe_sync_era = if block_header.era_id() == EraId::from(0) {
+                        let era_id = if block_header.era_id() == EraId::from(0) {
                             // if the block is in era 0 its parent can only be in era 0
-                            Some(EraId::from(0))
+                            EraId::from(0)
                         } else {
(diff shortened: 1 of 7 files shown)


node/src/types/validator_matrix.rs
```diff
@@ -92,6 +92,27 @@ impl ValidatorMatrix {
...
+    // Currently we only infer the validator weights for era 0 from the set registered for era 1.
+    // This is needed for the case where we want to sync leap to a block in era 0 of a pre 1.5.0
+    // network for which we cant get the validator weights from a switch block.
+    pub(crate) fn register_era_validator_weights_and_infer_era_0(
+        &mut self,
+        validators: EraValidatorWeights,
+    ) -> bool {
+        let was_present = self.register_era_validator_weights(validators.clone());
+        if validators.era_id() == EraId::from(1) {
+            self.register_era_validator_weights(EraValidatorWeights::new(
```
rafal-ch, line 108: Can you add an "info" or "debug" level log here, saying "infering validator weights for era 0"?

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_01.sh
```diff
@@ -301,6 +301,25 @@ function _step_08()
...
+            if [ "$RETRY_COUNT" -gt $TIMEOUT ]; then
+                log "ERROR :: NODE-$NODE_ID RETRY :: Failed to sync to genesis within $TIMEOUT seconds; lowest available block $LOWEST_AVAILABLE_BLOCK"
+                exit 1
+            else
+                LOWEST_AVAILABLE_BLOCK=$(get_node_lowest_available_block "$NODE_ID")
+                sleep 1
+                ((RETRY_COUNT=RETRY_COUNT+1))
+            fi
+        done
+    done
```
rafal-ch: What do you think about extracting [do_await_node_1_historical_sync](https://github.com/casper-network/casper-node/blob/749cab93492a9f59c161c152732560e213511540/utils/nctl/sh/scenarios/upgrade_after_emergency_upgrade_test.sh#L200), change it to `do_await_node_historical_sync(node_num)` and use here?

rafal-ch: Actually, if you choose to do the above, you can then easily add this check to all of the upgrade tests, potentially saving us some trouble :)

alsrdn: That's a good point, didn't see that we had that similar function. Will change.

node/src/types/validator_matrix.rs
alsrdn: Info level seems a bit excessive I think but a debug log is a good ideea.

alsrdn: added in https://github.com/casper-network/casper-node/pull/3639/commits/7a7a7adc38ff918e87fc1265f7bef868149c1b54

alsrdn:
bors try

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_01.sh
alsrdn: added in https://github.com/casper-network/casper-node/pull/3639/commits/e02e09277d030a9e81736eadb34d7e81aeb6aac9

node/src/reactor/main_reactor/keep_up.rs
```diff
@@ -667,18 +670,24 @@ impl MainReactor {
...
+                            // if the block is in era 0 its parent can only be in era 0
+                            Some(EraId::from(0))
+                        } else {
                             // we do not have the parent header and thus don't know what era
                             // the parent block is in (it could be the same era or the previous
                             // era). we assume the worst case and ask
                             // for the earlier era's proof
-                            Some(previous_era_id) => Ok(Some(SyncBackInstruction::Sync {
+                            block_header.era_id().predecessor()
+                        };
```
georgepisaltu: This code block now never returns `None`, so we can ditch the `Option` indirection. We can make this even simpler by doing `block_header.era_id().saturating_sub(1)`.

georgepisaltu (commented):
LGTM overall

rafal-ch (approved):
LGTM :+1:

node/src/reactor/main_reactor/keep_up.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/pull/3639/commits/260398e7cfb0a635682911a10d2106b3d5375ec1

Fraser999 (approved):

alsrdn:
bors merge

PR #3642, opened 2023-02-06
Fix race condition in contract runtime

### LINKED ISSUE #3622: Race condition in block execution
[This test](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/7386/2/4) revealed a race condition when attempting to execute a block.

When handling an `ContractRuntimeRequest::EnqueueBlockForExecution` event, we check the requested block's height against the mutex-protected execution pre-state in `self.execution_pre_state`.  However, we release that lock before reacquiring it in `execute_finalized_block_or_requeue` to take a clone of the pre-state.  While the lock is not held, another execution could complete and update the value in `self.execution_pre_state`, causing the execution of the current block to fail.

In the case of the linked CI run, this happened when the same block was requested to be executed twice in quick succession.

Fraser999:
This PR fixes a race condition which can happen when enqueuing blocks for execution.

Closes #3622.

### DIFF
--- a/node/src/components/contract_runtime.rs
+++ b/node/src/components/contract_runtime.rs
@@ -452,8 +452,9 @@ impl ContractRuntime {
             } => {
                 let mut effects = Effects::new();
                 let exec_queue = Arc::clone(&self.exec_queue);
-                let next_block_height = self.execution_pre_state.lock().unwrap().next_block_height;
                 let finalized_block_height = finalized_block.height();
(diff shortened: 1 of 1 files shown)


georgepisaltu (approved):

fizyk20 (approved):

rafal-ch (approved):
LGTM :+1:

Fraser999:
bors merge

PR #3644, opened 2023-02-06
nctl/nightly tests: fix upgrade scenario 9

### LINKED ISSUE #3620: Upgrade 9 test failing

alsrdn:
We allow a node of the same binary to sync to a trusted hash of an older protocol version only if the trusted hash is within the allowed sync leap range. Adjust the test to check that if the trusted hash is not within the allowed range we don't allow the new node to sync.

Fixes: https://github.com/casper-network/casper-node/issues/3620

### DIFF
--- a/utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
+++ b/utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
@@ -8,9 +8,9 @@
 # Step 03: Stage nodes 1-5 and upgrade.
 # Step 04: Assert upgraded nodes 1-5.
 # Step 05: Assert nodes 1-5 didn't stall.
-# Step 06: Await 1 era and stage node-6 with a recent trusted hash.
-# Step 07: Await 5 eras and stange node-7 with an old trusted hash.
-# Step 08: Verify that node-6 successfully syncs and node 7 did not sync as expected.
+# Step 06: Await 1 era and stage nodes 6&7 with a recent trusted hash (pre & post).
+# Step 07: Await 7 eras and stage nodes 8&9 with an old trusted hash (pre & post).
+# Step 08: Verify that nodes 6&7 successfully syncs and nodes 8&9 did not sync as expected.
 # Step 09: Run Health Checks
 # Step 10: Terminate.
 
@@ -56,8 +56,8 @@ function _main()
     POST_UPGRADE_BLOCK_HASH="$($(get_path_to_client) get-block --node-address "$(get_node_address_rpc '2')" | jq -r '.result.block.hash')"
 
     _step_05
-    _step_06 "$POST_UPGRADE_BLOCK_HASH"
-    _step_07 "$PRE_UPGRADE_BLOCK_HASH"
+    _step_06 "$PRE_UPGRADE_BLOCK_HASH" "$POST_UPGRADE_BLOCK_HASH"
+    _step_07 "$PRE_UPGRADE_BLOCK_HASH" "$POST_UPGRADE_BLOCK_HASH"
     _step_08
     _step_09
     _step_10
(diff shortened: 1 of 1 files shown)


utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
```diff
@@ -56,8 +56,14 @@ function _main()
     POST_UPGRADE_BLOCK_HASH="$($(get_path_to_client) get-block --node-address "$(get_node_address_rpc '2')" | jq -r '.result.block.hash')"
 
     _step_05
-    _step_06
-    _step_07 "$STAGE_ID" "$ACTIVATION_POINT" "$PRE_UPGRADE_BLOCK_HASH" "$POST_UPGRADE_BLOCK_HASH"
+    # step 6 - start node 6 with trusted hash withing the allowed range
+    log_step_upgrades 6 "awaiting 1 eras to start node 6"
```
Fraser999:
```suggestion
    # step 6 - start node 6 with trusted hash within the allowed range
    log_step_upgrades 6 "awaiting 1 era to start node 6"
```

Fraser999 (approved):
Couple of nitpicks, but LGTM otherwise.

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3644/commits/658b5260dd58ab458deb8a60910b7ed0c64a56b6

Fraser999 (approved):

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
```diff
@@ -163,46 +169,33 @@ function _step_05()
     done
 }
 
-# Step 06: Await 1 era.
-function _step_06()
```
TomVasile, line 167: Was there a reason to change the convention that the test is using?

TomVasile (changes requested):
Blocking for now.  Running the test locally, I'm seeing node 6 die still. Will discuss in standup.

alsrdn:
> Blocking for now. Running the test locally, I'm seeing node 6 die still. Will discuss in standup.

Node 6 was failing due to the issue fixed by this PR: https://github.com/casper-network/casper-node/pull/3639
I've merged back `dev` into this branch and it should pass now.

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
alsrdn: Initially I saw no point in having them as separate functions, but it makes sense for convention. Added the step functions back in https://github.com/casper-network/casper-node/pull/3644/commits/e8c82e43f80f8f865ed042d1b0d044761a9651be

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
```diff
@@ -163,46 +163,51 @@ function _step_05()
     done
 }
 
-# Step 06: Await 1 era.
+# Step 06: Await 1 era and then start node 6 with trusted hash within the allowed range
```
TomVasile: I'm wondering if we should also add a node here that starts up with the pre-upgrade hash? Since it would be allowed in the range, we should probably verify that as well.

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
```diff
@@ -163,46 +163,51 @@ function _step_05()
...
-    log_step_upgrades 6 "awaiting next era"
-    nctl-await-n-eras offset='1' sleep_interval='5.0' timeout='180' node_id='2'
+    POST_UPGRADE_BLOCK_HASH=${1}
+    log_step_upgrades 6 "awaiting 1 era to start node 6"
+    nctl-await-n-eras offset='1' sleep_interval='5.0' timeout='300' node_id='2'
+    _stage_node_with_trusted_hash "6" "$STAGE_ID" "$ACTIVATION_POINT" "$POST_UPGRADE_BLOCK_HASH" 6
 }
 
-# Step 07: Stage node 6 with post-upgrade trusted hash and node 7 with pre-upgrade trusted hash
+# Step 07: Await 5 eras and then start node 7 with a trusted hash that is too old
```
TomVasile: Similar to my other comment, would a node started with the post-upgrade hash that was out of the range fail here? If so we should cover that too.

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
```diff
@@ -163,46 +163,51 @@ function _step_05()
...
 function _step_07()
 {
-    local STAGE_ID=${1}
-    local ACTIVATION_POINT=${2}
-    local PRE_UPGRADE_BLOCK_HASH=${3}
-    local POST_UPGRADE_BLOCK_HASH=${4}
+    PRE_UPGRADE_BLOCK_HASH=${1}
```
TomVasile:
```suggestion
    local PRE_UPGRADE_BLOCK_HASH=${1}
```

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
```diff
@@ -163,46 +163,51 @@ function _step_05()
     done
 }
 
-# Step 06: Await 1 era.
+# Step 06: Await 1 era and then start node 6 with trusted hash within the allowed range
 function _step_06()
 {
-    log_step_upgrades 6 "awaiting next era"
-    nctl-await-n-eras offset='1' sleep_interval='5.0' timeout='180' node_id='2'
+    POST_UPGRADE_BLOCK_HASH=${1}
```
TomVasile:
```suggestion
    local POST_UPGRADE_BLOCK_HASH=${1}
```

TomVasile (approved):
Ran locally and the test works as is! Just left some comments I think we might want to cover. Approving tentatively.

cc/ @Fraser999

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
alsrdn: It shouldn't matter if the block is pre or post upgrade if the upgrade was successful. The condition for it to work is to provide a trusted block hash within the allowed range. Actually this is the reason why the test was failing in the past; node 7 was using a hash that was within the allowed range and didn't fail as expected.
So we can start 2 more nodes that have their hash within the allowed range, one pre and one post upgrade to see if it works but not sure it is necessary.

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
alsrdn: added in https://github.com/casper-network/casper-node/pull/3644/commits/7f5aecd791c8b22d3cbd70ade3327c49ff75f862

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_09.sh
alsrdn: added in https://github.com/casper-network/casper-node/pull/3644/commits/7f5aecd791c8b22d3cbd70ade3327c49ff75f862

alsrdn:
bors merge

PR #3646, opened 2023-02-07
Fix `upgrade_after_emergency_upgrade_test`

### LINKED ISSUE #3647: Fix upgrade_after_emergency_upgrade_test.sh

alsrdn:
The acceptance was not reported correctly for a signature that was received in a state where only one signature was missing to achieve strict finality.
This happens because we report it only if we should be currently acquiring signatures. This determination was made however after we applied the new signature, making it appear that we should not have acquired signatures which lead to the acceptance not being reported and the latch not being properly reset. This makes the sinchronizer stall making it impossible to transition to the next state.

Fixed by moving the determination of whether the synchronizer is currently acquiring signatures before the new signature is applied


Fraser999 (approved):

fizyk20 (approved):
Wow, nice catch.

alsrdn:
bors merge

PR #3650, opened 2023-02-08
Have sync leap response composed of only complete blocks' headers

### LINKED ISSUE #3643: 1 validating node out of 5 not progressing with sync AFTER an emergency upgrade
1 validating node out of 5 not progressing with sync AFTER an emergency upgrade. Attaching logs
http://genesis.casperlabs.io/ajith-mainnet-clone/casper-node-dumps/ajith-mainnet-clone/06022023_1302/dump_download_list.html

Fraser999:
This PR changes the way in which a sync leap response is created.

Previously there was an explicit check for sufficient signatures on each block header in the `signed_block_headers`.  This had an issue where validator-changing upgrades were not handled correctly and hence the validator set expected to have signed an immediate switch block might not be the correct one, causing the sync leap response to contain no `signed_block_headers` after or including such an immediate switch block.

Rather than fixing the signature checking, we now use the available block range as an inclusion criterion.  This criterion is now also applied to the `trusted_ancestor_headers` where previously no sufficiency check was done.

Closes #3643.

### DIFF
--- a/node/src/components/block_accumulator/tests.rs
+++ b/node/src/components/block_accumulator/tests.rs
@@ -171,7 +171,6 @@ impl Reactor for MockReactor {
 
         let storage = Storage::new(
             &storage_withdir,
-            Ratio::new(1, 3),
             None,
             ProtocolVersion::from_parts(1, 0, 0),
             "test",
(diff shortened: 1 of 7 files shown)


rafal-ch (approved):
LGTM :+1:

alsrdn (approved):
LGTM 👍

fizyk20 (approved):

Fraser999:
bors merge

PR #3663, opened 2023-02-10
Share info about blocks and finality signatures between BlockAccumulator and fetchers

### LINKED ISSUE #3612: Update block accumulator
The responsibility for storing fin sigs and blocks should be removed from the accumulator and given to the reactor, which will poll the accumulator for sigs or blocks which are ready to be stored.  Fin sigs should be stored as soon as they're validated, without waiting for strict finality on their corresponding block.  Fin sigs should only be cryptographically verified after they have been deemed valid for the given block.

Fraser999:
This PR increases the ability of the `BlockFetcher` and `FinalitySignatureFetcher` to fetch locally by querying the `BlockAccumulator` after a failure to fetch from `Storage` before sending a request to peers.

It also adds an announcement for these two fetchers to emit once they retrieve an item from a peer and store it.  The announcements cause the `BlockAccumulator` to be notified of the newly-fetched block or finality signature, using the existing events for handling incoming gossiped items.

Closes #3612.

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -816,7 +816,7 @@ impl<REv: ReactorEvent> Component<REv> for BlockAccumulator {
                 Effects::new()
             }
             Event::ReceivedBlock { block, sender } => {
-                let meta_block = MetaBlock::new(Arc::new(*block), vec![], MetaBlockState::new());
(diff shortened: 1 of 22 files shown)


rafal-ch (approved):
LGTM :+1:

PR #3664, opened 2023-02-10
Fix upgrade_scenario_10

### LINKED ISSUE #3621: Upgrade 10 test failing

alsrdn:
* Fixes invocation of the `global-state-update-gen` in order to correctly generate the missing system contract registry.
* Fixes an issue where the node tries to cache the system contract registry when initializing the contract runtime component, but since the node was not upgraded yet and the previous version didn't have the system contract registry written the initialization would fail. Now caching the system contract registry on initialization will not yield a fatal error and allows it to be created after the upgrade or when needed.

Fixes: https://github.com/casper-network/casper-node/issues/3621

### DIFF
--- a/node/src/components/contract_runtime.rs
+++ b/node/src/components/contract_runtime.rs
@@ -23,13 +23,12 @@ use once_cell::sync::Lazy;
 use prometheus::Registry;
 use serde::Serialize;
 use thiserror::Error;
-use tracing::{debug, error, info, trace, warn};
+use tracing::{debug, error, info, trace};
 
 use casper_execution_engine::{
     core::engine_state::{
-        self, genesis::GenesisError, ChainspecRegistry, EngineConfig, EngineState,
-        Error as EngineStateError, GenesisSuccess, SystemContractRegistry, UpgradeConfig,
-        UpgradeSuccess,
+        self, genesis::GenesisError, ChainspecRegistry, EngineConfig, EngineState, GenesisSuccess,
+        SystemContractRegistry, UpgradeConfig, UpgradeSuccess,
     },
     shared::{newtypes::CorrelationId, system_config::SystemConfig, wasm_config::WasmConfig},
     storage::{
@@ -377,12 +376,7 @@ impl ContractRuntime {
                 let engine_state = Arc::clone(&self.engine_state);
                 let metrics = Arc::clone(&self.metrics);
 
-                // Try to cache system contract registry if possible.
-                if self.system_contract_registry.is_none() {
-                    self.system_contract_registry = engine_state
-                        .get_system_contract_registry(CorrelationId::new(), request.state_hash())
-                        .ok();
-                };
+                self.try_init_system_contract_registry_cache();
 
                 let system_contract_registry = self.system_contract_registry.clone();
                 // Increment the counter to track the amount of times GetEraValidators was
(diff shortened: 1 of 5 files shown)


node/src/components/contract_runtime.rs
```diff
@@ -721,6 +726,13 @@ impl ContractRuntime {
             Ok(system_contract_registry) => {
                 self.system_contract_registry = Some(system_contract_registry);
             }
+            Err(EngineStateError::MissingSystemContractRegistry) => {
+                // The system contract registry might be missing because the node did not have it
+                // before and may create it through an upgrade. Since this is a
+                // cache, allow it to be populated later.
+                self.system_contract_registry = None;
+                warn!("ContractRuntime: system contract registry cache was not initialized.");
+            }
```
Fraser999: I wonder if we can have this be a hard error (i.e. log an `error!` and return `Err`) if we end up here via the `MainReactor::refresh_contract_runtime` flow.  For cases where we end up here via `MainReactor::commit_(genesis|upgrade)` it can be ignored (i.e. warn or even info only).

I also don't think we need to set `self.system_contract_registry = None;` here do we?  It would already be `None` I assume, but if it wasn't we shouldn't be overwriting it.  Maybe worth a `debug_assert!` that it's `None`?

alsrdn: I don't think we should distinguish here if we should log an error here or not depending on the call site. In essence the `self.system_contract_registry` is an internal cache of the contract runtime and it needs to be managed internally by the component. What we're saying here is that we initialized the contract runtime, it is usable, but we couldn't yet cache the system contract registry. It shouldn't fail if it's not needed at this point and should fail when it is used, if that is truly critical, which I don't think it even is right now.

I would apply the same argument for setting it to `None` here, though this case is a little weird since it means that something went wrong with the global state since the system contract registry is stable and should always be present if it has been in the past. Maybe we should log a different warning in this case stating what happened. WDYT?

alsrdn: As discussed, I left only the lazy initialization of the cache at use time.

Fraser999 (approved):

node/src/components/contract_runtime.rs
```diff
@@ -375,6 +375,9 @@ impl ContractRuntime {
                 trace!(?request, "get era validators request");
                 let engine_state = Arc::clone(&self.engine_state);
                 let metrics = Arc::clone(&self.metrics);
+
+                self.try_init_system_contract_registry_cache();
```
rafal-ch, line 379: A nit, but since this function is infallible, shoud it be just `init_system_contract_registry_cache()`?

node/src/components/contract_runtime.rs
```diff
@@ -921,6 +899,25 @@ impl ContractRuntime {
     pub(crate) fn engine_state(&self) -> &Arc<EngineState<LmdbGlobalState>> {
         &self.engine_state
     }
+
+    #[inline]
```
rafal-ch, line 903: Why we would like this to be inlined?

rafal-ch (approved):
LGTM :+1: 
With minor nits.

node/src/components/contract_runtime.rs
alsrdn: Indeed it does not fail but it also does not guarantee that the system contract registry cache will be set either. I don't have a strong preference for naming it. However I think that if the code changes or something goes wrong a reader might be more inclined to investigate that function if it has a `try_` prefix.

node/src/components/contract_runtime.rs
alsrdn: This code is small enough to go into `handle_contract_runtime_request` actually though I preferred to move it for better readability. It doesn't have to be inline, I just considered to mark it that way since it has a single call site and is small enough to be a good candidate for it.

rafal-ch: Sure, thanks for explanation :+1:

alsrdn:
bors merge

PR #3665, opened 2023-02-10
GH-3672: Revert serialization but put restriction in place.

### LINKED ISSUE #3672: Avoid breaking change in serialization of VestingSchedule
The type of VestingSchedule::locked_amounts was changed from Option<[U512; LOCKED_AMOUNTS_LENGTH]> to Option<Vec<U512>> and this constitutes a breaking change of the bincode-encoding format.

mpapierski:
Closes #3672

This reverts bytesrepr and bincode (and serde in general) serialization formats back to before #3219 was merged. But to satisfy the new requirement of configurable vesting schedules, a restriction is put in place so the length would not exceed fixed-sized slots of 14 weeks.

### DIFF
--- a/execution_engine_testing/test_support/src/chainspec_config.rs
+++ b/execution_engine_testing/test_support/src/chainspec_config.rs
@@ -6,13 +6,13 @@ use std::{
 
 use num_rational::Ratio;
 use once_cell::sync::Lazy;
-use serde::{Deserialize, Serialize};
+use serde::{de, Deserialize, Deserializer, Serialize};
 
 use casper_execution_engine::{
     core::engine_state::{run_genesis_request::RunGenesisRequest, ExecConfig, GenesisAccount},
     shared::{system_config::SystemConfig, wasm_config::WasmConfig},
 };
-use casper_types::ProtocolVersion;
+use casper_types::{system::auction::VESTING_SCHEDULE_LENGTH_MILLIS, ProtocolVersion, TimeDiff};
 
 use crate::{
     DEFAULT_ACCOUNTS, DEFAULT_CHAINSPEC_REGISTRY, DEFAULT_GENESIS_CONFIG_HASH,
@@ -39,9 +39,20 @@ pub enum Error {
         error: io::Error,
     },
     FailedToParseChainspec(toml::de::Error),
-    FailedToCreateExecConfig,
-    FailedToParseLockedFundsPeriod,
-    FailedToCreateGenesisRequest,
+}
+
+fn vesting_schedule_period<'de, D: Deserializer<'de>>(
+    deserializer: D,
+) -> Result<TimeDiff, D::Error> {
+    let timediff = TimeDiff::deserialize(deserializer)?;
+    if timediff > TimeDiff::from_millis(VESTING_SCHEDULE_LENGTH_MILLIS) {
+        Err(de::Error::custom(format!(
+            "vesting schedule period can't exceed {} milliseconds",
+            VESTING_SCHEDULE_LENGTH_MILLIS
+        )))
+    } else {
+        Ok(timediff)
+    }
 }
 
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug)]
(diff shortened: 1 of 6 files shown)


node/src/types/chainspec/core_config.rs
```diff
@@ -14,10 +14,25 @@ use serde::{
 use casper_types::testing::TestRng;
 use casper_types::{
     bytesrepr::{self, FromBytes, ToBytes},
+    system::auction::VESTING_SCHEDULE_LENGTH_MILLIS,
     TimeDiff,
 };
 use tracing::{error, warn};
 
+fn vesting_schedule_period<'de, D: Deserializer<'de>>(
```
Fraser999: Let's stick with validating in the `is_valid` method please.  Something like:
```rust
        if self.vesting_schedule_period.millis() > VESTING_SCHEDULE_LENGTH_MILLIS {
            error!(
                vesting_schedule_millis = self.vesting_schedule_period.millis(),
                max_millis = VESTING_SCHEDULE_LENGTH_MILLIS
                "vesting schedule period too long"
            );
            return false;
        }
```

types/src/system/auction/bid/vesting.rs
```diff
@@ -74,23 +67,31 @@ impl VestingSchedule {
...
             vesting_schedule_period_to_weeks(vesting_schedule_period_millis);
 
+        if locked_amounts_length as usize >= LOCKED_AMOUNTS_MAX_LENGTH {
+            // We will not initialize a vesting schedule if the genesis configuration is incorrect.
+            // There are restrictions put in place in the chainspec config to make sure we will not
+            // hit this case.
+            //
+            // This is done to avoid breaking changes in the serialization format.
+            return false;
+        }
```
Fraser999: I think we should panic here, since returning `false` doesn't really convey the error.
```suggestion
        assert!(
            locked_amounts_length < LOCKED_AMOUNTS_LENGTH as u64,
            "vesting schedule period must be less than {} weeks",
            LOCKED_AMOUNTS_LENGTH
        );
```

Let's also add the following to the doc comment:
```rust
    ///
    /// # Panics
    ///
    /// Panics if `vesting_schedule_period_millis` represents more than 13 weeks.
```

types/src/system/auction/bid/vesting.rs
```diff
@@ -151,104 +153,87 @@ impl VestingSchedule {
...
-            None => {
-                UNINITIALIZED_LOCKED_AMOUNTS_TAG.write_bytes(writer)?;
-            }
-            Some(locked_amounts) => {
-                VARIABLE_LOCKED_AMOUNTS_LENGTH_TAG.write_bytes(writer)?;
-                locked_amounts.write_bytes(writer)?;
-            }
+impl FromBytes for [U512; LOCKED_AMOUNTS_MAX_LENGTH] {
+    fn from_bytes(mut bytes: &[u8]) -> Result<(Self, &[u8]), Error> {
+        let mut result: MaybeUninit<[U512; LOCKED_AMOUNTS_MAX_LENGTH]> = MaybeUninit::uninit();
```
Fraser999: Let's avoid `MaybeUninit`.

execution_engine_testing/test_support/src/chainspec_config.rs
```diff
@@ -39,9 +39,20 @@ pub enum Error {
         error: io::Error,
     },
     FailedToParseChainspec(toml::de::Error),
-    FailedToCreateExecConfig,
-    FailedToParseLockedFundsPeriod,
-    FailedToCreateGenesisRequest,
+}
+
+fn vesting_schedule_period<'de, D: Deserializer<'de>>(
```
Fraser999: Again, let's not do this to validate config settings.  This should be changed to match how we do it in the "real" config, ideally sharing code if possible.

types/src/system/auction/bid/vesting.rs
mpapierski: Done, and moved the serialization logic to `fn write_bytes`

types/src/system/auction/bid/vesting.rs
mpapierski: Addressed

execution_engine_testing/test_support/src/chainspec_config.rs
mpapierski: Ticketed the refactor to share most code (# to be announced). For now, refactored just enough to always call the new method `fn is_valid()` before returning a valid `ChainspecConfig` instance.

node/src/types/chainspec/core_config.rs
mpapierski: Addressed

Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

PR #3673, opened 2023-02-10
nctl: allow override of the NCTL base URL for remote assets

alsrdn:
The remote assets are only compiled for x86-64 which prevents running the tests on other platforms (like ARM aarch64).
Add ability to override the base URL so that users can run the tests using assets from a different location, or locally cached assets by specifying a `file://<local_directory>` URL.


Fraser999 (approved):

alsrdn:
bors merge

alsrdn:
bors merge

PR #3674, opened 2023-02-13
Never purge the local tip's acceptor in the block accumulator

### LINKED ISSUE #3661: Don't purge tip from BlockAccumulator
When purging in the BlockAccumulator, we should ensure we don't purge an acceptor for the local tip, even if it meets the criteria for being expired.  See https://casper-labs-team.slack.com/archives/C01ULKL8G8J/p1675046440956279 for an example where this caused completed block fa68..7e0e to not get marked complete due to a delay between the last progress being recorded on the acceptor and the receipt of the ExecutedBlock event.

georgepisaltu:
Fixes #3661 

This PR forbids purging of the local tip in the block accumulator. The old `block_timestamps` entries are, however, still purged. If there is new information received on the local tip's acceptor, new `block_timestamps` entries will be created.


rafal-ch (approved):
LGTM :+1:

Fraser999 (approved):

PR #3680, opened 2023-02-15
add balance check to unbond migration

dwerner:
This PR adds a balance check to the migration from old unbond purses to the new format. If the purse lacks sufficient balance, we fail the upgrade.

Also adds some logging to `unbond`, to provide rather than eat errors.

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -495,9 +495,16 @@ where
                     }
 
                     for (unbond_purse_uref, unbond_amount) in balances {
+                        let key = match tracking_copy
+                            .borrow_mut()
+                            .get_purse_balance_key(correlation_id, unbond_purse_uref.into())
+                        {
+                            Ok(key) => key,
+                            Err(_) => return Err(Error::Mint("purse balance not found".into())),
+                        };
                         let current_balance = tracking_copy
                             .borrow_mut()
-                            .get_purse_balance(CorrelationId::new(), Key::URef(unbond_purse_uref))?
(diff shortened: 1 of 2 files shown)


dwerner:
bors try

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -472,6 +472,53 @@ where
...
+                        };
+                        let current_balance = tracking_copy
+                            .borrow_mut()
+                            .get_purse_balance(CorrelationId::new(), key)?
+                            .value();
+
+                        if unbond_amount > current_balance {
+                            // If we don't have enough balance to migrate, the only thing we can do
+                            // is to fail the upgrade.
+                            error!("commit_upgrade failed during migration - insufficient funds({current_balance}) in purse({unbond_purse_uref}) to unbond({unbond_amount}) ");
```
rafal-ch: A nit:
Consider using the tracing values format, like in [here](https://github.com/casper-network/casper-node/blob/dev/execution_engine/src/core/engine_state/mod.rs#L1781).

execution_engine/src/core/engine_state/mod.rs
```diff
@@ -472,6 +472,53 @@ where
                     .ok_or(Error::FailedToGetWithdrawPurses)?
                     .to_owned();
 
+                // Ensure that sufficient balance exists for all unbond purses that are to be
+                // migrated.
+                {
```
rafal-ch: Consider extracting what we have in this scope to a separate function. That will make it easier to add a unit test.

Fraser999 (approved):

dwerner:
bors merge

PR #3686, opened 2023-02-16
Avoid unnecessary sync leap

### LINKED ISSUE #3685: Avoid sync-leaping unnecessarily
On receiving a sync-leap response, if the highest block in the response is the block we're currently syncing, don't do a further sync-leap

rafal-ch:
This PR adjusts the sync leap behavior:
1. Upon receiving a successful sync leap response, the block_hash of the highest block is memoized in main reactor
2. When another sync leap is about to be requested, the memoized block hash is compared to the block hash we're about to request a sync leap for
3. If the block hashes match, the sync leap attempt is aborted

Closes #3685

### DIFF
--- a/node/src/reactor/main_reactor/catch_up.rs
+++ b/node/src/reactor/main_reactor/catch_up.rs
@@ -328,9 +328,6 @@ impl MainReactor {
             );
         }
 
-        // latch accumulator progress to allow sync-leap time to do work
-        self.block_accumulator.reset_last_progress();
-
         let sync_leap_identifier = SyncLeapIdentifier::sync_to_tip(block_hash);
         let effects =
             self.request_leap_if_not_redundant(sync_leap_identifier, effect_builder, peers_to_ask);
--- a/node/src/reactor/main_reactor/catch_up.rs
+++ b/node/src/reactor/main_reactor/catch_up.rs
@@ -288,7 +288,7 @@ impl MainReactor {
                 best_available,
                 from_peers,
                 ..
-            } => self.catch_up_leap_received(effect_builder, rng, best_available, from_peers),
+            } => self.catch_up_leap_received(effect_builder, rng, *best_available, from_peers),
             LeapState::Failed { error, .. } => {
                 self.catch_up_leap_failed(effect_builder, rng, block_hash, error)
             }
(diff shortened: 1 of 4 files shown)


rafal-ch:
bors try

georgepisaltu (commented):
The approach presented here is implemented well, but I am a bit concerned that we're introducing a lot of state to the reactor logic which we could maybe avoid. Since the accumulator is the component which tells the reactor to leap or not, if we receive a leap response with the highest block hash equal to the local tip in the accumulator, we could just have a `touch` on the accumulator and purge the builders in the synchronizer, since that response would mean we're at tip and the network hasn't progressed. If the state machine is implemented correctly, from what I can tell right now the accumulator would tell the reactor that it's caught up (because we did a `touch`). All of the required information to determine these cases is accessible in the reactor logic without extra state. In the case where we don't have a straightforward way to do a `touch` on the accumulator, we could add a new event for it to do just that. I didn't write a POC for this so I might be missing something, let me know what you think.

node/src/reactor/main_reactor/catch_up.rs
```diff
@@ -333,37 +332,35 @@ impl MainReactor {
         self.block_accumulator.reset_last_progress();
```
Fraser999: Seems like this should only be done if we do actually leap.  It would be good to move it inside of `request_leap_if_not_redundant`.

node/src/reactor/main_reactor/control.rs
```diff
@@ -73,6 +73,7 @@ impl MainReactor {
                 }
                 UpgradingInstruction::CatchUp => {
                     info!("Upgrading: switch to CatchUp");
+                    self.last_sync_leap_highest_block_hash = None;
```
Fraser999: I don't think we need to ever reset `self.last_sync_leap_highest_block_hash` to `None` (applies to all 4 instances in this file).

Fraser999:
@georgepisaltu re your comment, I'm afraid I don't agree that this PR introduces a lot of state to the reactor.  I feel that the simplicity and explicitness of this approach makes it more robust than an alternative which adds more burden on the already complex relationship between the synchronizer and accumulator.

I could kinda agree if the suggestion was to put the new state inside the sync leaper (so e.g. `LeapState::Idle` was updated to include the optional block hash of the latest attempt's highest block), but I don't see that making much difference.

georgepisaltu:
> @georgepisaltu re your comment, I'm afraid I don't agree that this PR introduces a lot of state to the reactor. I feel that the simplicity and explicitness of this approach makes it more robust than an alternative which adds more burden on the already complex relationship between the synchronizer and accumulator.

As I said previously, I think the approach in this PR is implemented well. If both you and @rafal-ch agree on this, I'm fine with it as well.

> I could kinda agree if the suggestion was to put the new state inside the sync leaper (so e.g. `LeapState::Idle` was updated to include the optional block hash of the latest attempt's highest block), but I don't see that making much difference.

My reasoning was that the accumulator is ultimately the component which tells us to leap or do other things, so it made the most sense to me to put it there. After all, it was the accumulator which gave the leap instruction with that hash, so it should know if a leap comes back with that same hash and do some internal updates.

I'm going to approve now. Feel free to merge once all the other comments are addressed.

georgepisaltu (approved):

node/src/reactor/main_reactor/catch_up.rs
rafal-ch: Good catch!
Fixed in 524bfb7606100451d5b8a8c6228b4827f3a3223e

node/src/reactor/main_reactor/control.rs
rafal-ch: Updated in ff7100391140362d24541001ab78a2fb2ff88751

Fraser999 (approved):

PR #3687, opened 2023-02-16
Create acceptors when local tip is unknown

### LINKED ISSUE #3683: BlockAccumulator to allow upserting if local tip unknown
In `upsert_acceptor`, if `self.local_tip` is `None` the new acceptor is not created.  This behaviour needs to be reversed.

georgepisaltu:
Fixes #3683 

This PR introduces a change that relaxes the requirements for creating new block acceptors in `BlockAccumulator::upsert_acceptor`. Previously acceptors would not be created when the accumulator had no local tip, but we decided that this behavior is too restrictive and should be changed.

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -258,12 +258,13 @@ impl BlockAccumulator {
             btree_map::Entry::Vacant(entry) => entry,
         };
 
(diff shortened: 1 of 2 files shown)


Fraser999 (approved):

alsrdn (approved):
LGTM

rafal-ch (approved):
LGTM :+1:

PR #3689, opened 2023-02-16
Update global_state_update_gen tool to not slash by default

### LINKED ISSUE #3688: Avoid slashing by default when running global_state_update_gen tool
This only applies when running the tool in `change-validators` mode

Fraser999:
This PR changes the behaviour of the `change-validators` subcommand of the global_state_update_gen tool to not slash removed validators.

A side effect is that the subcommand takes longer to complete, a change from ~4.5 seconds to ~110 seconds for me using a mainnet snapshot.

Closes #3688.

### DIFF
--- a/utils/global-state-update-gen/src/validators.rs
+++ b/utils/global-state-update-gen/src/validators.rs
@@ -55,7 +55,7 @@ pub(crate) fn generate_validators_update(matches: &ArgMatches<'_>) {
         accounts,
         transfers: vec![],
         only_listed_validators: true,
-        slash_instead_of_unbonding: true, // for consistency with the old behavior
+        slash_instead_of_unbonding: false,
     };
 
     let builder = LmdbWasmTestBuilder::open_raw(data_dir, Default::default(), state_hash);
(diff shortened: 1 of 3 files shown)


georgepisaltu (approved):
LGTM

utils/global-state-update-gen/src/validators.rs
```diff
@@ -55,7 +55,7 @@ pub(crate) fn generate_validators_update(matches: &ArgMatches<'_>) {
         accounts,
         transfers: vec![],
         only_listed_validators: true,
-        slash_instead_of_unbonding: true, // for consistency with the old behavior
+        slash_instead_of_unbonding: false,
```
alsrdn, line 58: Not sure about changing the default here since it may brake scripts that are already using this command and expect this behaviour.
Can't we add a `--slash` parameter to the `change-validators` subcommand and leave the default to true?

Fraser999: Well, I think the whole point was to make the default `false` since slashing is seen as an extreme measure, so if it does break scripts we'll need to adjust them to suit the new behaviour.  Also, this command should be treated as deprecated since the `generic` subcommand is meant to be the go-forward.

alsrdn: Make sense, thanks! I also saw that we're at v0.3.0 with this tool so breaking past usage shouldn't be a concern.

alsrdn (approved):

Fraser999:
bors merge

PR #3691, opened 2023-02-17
Fix slashing behaviour in `global_state_update_gen` tool

### LINKED ISSUE #3682: Use `withdraw` purses in global state update tool
The tool is reading only unbonding purses to create a set of unbonds which need to be ignored, but on 1.4.x data and lower, this needs to also read withdraw purses.

Fraser999:
This PR fixes the slashing behaviour in the following ways:
* withdraws and unbonds are only modified when slashing is enabled
* when acquiring the set of withdraws, rather than reading only `StoredValue::Unbonding` values from global state, we now also read `StoredValue::Withdraw` values, since pre-1.5.0 versions will only have the latter
* having acquired the set of withdraws, as well as replacing the values with an empty `Vec` for those related to validators to be slashed, we now also slash the corresponding validator and delegator unbonding purses (i.e. their balances are set to zero)

Closes #3682.

### DIFF
--- a/utils/global-state-update-gen/src/generic.rs
+++ b/utils/global-state-update-gen/src/generic.rs
@@ -131,7 +131,9 @@ fn update_auction_state<T: StateReader>(
         slash,
     );
 
-    state.remove_withdraws(&validators_diff.removed);
+    if slash {
+        state.remove_withdraws_and_unbonds(&validators_diff.removed);
+    }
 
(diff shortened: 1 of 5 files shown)


Fraser999:
bors try

fizyk20 (approved):
LGTM :+1: Nice one with the tests!

georgepisaltu (approved):

Fraser999:
bors merge

PR #3694, opened 2023-02-21
Fix the `request_evidence` implementation.

real-felix:
The `request_evidence` has a placeholder implementation: this commit fixes that implementation to use a correct one.


fizyk20 (approved):
LGTM :+1:

node/src/components/consensus/protocols/zug.rs
```diff
@@ -2194,15 +2193,19 @@ where
     }
 
     fn request_evidence(&self, peer: NodeId, vid: &C::ValidatorId) -> ProtocolOutcomes<C> {
```
Fraser999: As discussed elsewhere, I think this should be renamed `send_evidence` as the name is currently misleading.

Fraser999 (approved):
Optimistic approval given the nitpick is addressed.

PR #3707, opened 2023-02-24
Make `starting_state_root_hash` required

### LINKED ISSUE #3705: Make `starting_state_root_hash` non-optional
The `starting_state_root_hash` of `GetStatusResult` should be reverted to be a non-optional `Digest`.  When there is no block available to populate it, use `Digest::default()`.  Also replace the deprecation notice about this field in the node's CHANGELOG to a new entry in the `Changed` section, describing what the field now represents (i.e. the state root hash of the lowest available block)

georgepisaltu:
Fixes #3705 

This PR makes the `starting_state_root_hash` field in the REST and JSON-RPC status endpoint responses required and defaulted in case the available block range is empty. It also updates `CHANGELOG.md` to reflect the current state of the status endpoint.

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -64,6 +64,7 @@ All notable changes to this project will be documented in this file.  The format
 * Connections to unresponsive nodes will be terminated, based on a watchdog feature.
 
 ### Changed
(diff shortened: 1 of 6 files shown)


rafal-ch (approved):
LGTM :+1:

PR #3712, opened 2023-02-27
Message size upper bound check based on chainspec values.

### LINKED ISSUE #3582: Make maximum network message size check part of chainspec validation
The message size validation is mostly based on chainspec values and thus not ideal in a static unit test. 
Furthermore, variations in the chainspec configuration could potentially cause it to become invalid.
By hooking the check into the regular validation process for chainspecs, this process becomes more future proof.

marc-casperlabs:
Closes #3582.

Adds message size validation based on auto-generated maximum size messages from a chainspec. Currently only implemented as a unit test, but we can amend this PR with integration into `Chainspec::is_valid`, if we want to.


marc-casperlabs:
Moving this to draft for a short while until the implementation issue above is resolved.

marc-casperlabs:
This is finally in a state where I am somewhat comfortable with a review.

At the moment, validating a chainspec sadly takes 40 seconds in debug mode (1 second in release). We may need to optimize further to not break nctl testing.

EdHastingsCasperAssociation:
what rel is this intended for?

marc-casperlabs:
> what rel is this intended for?

It was designated a "nice to have" for 1.5, otherwise 1.6.

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+
+impl LargestSpecimen for RuntimeArgs {
+    fn largest_specimen<E: SizeEstimator>(_estimator: &E, _cache: &mut Cache) -> Self {
+        Default::default()
+    }
+}
+
+impl LargestSpecimen for ChunkWithProof {
+    fn largest_specimen<E: SizeEstimator>(_estimator: &E, _cache: &mut Cache) -> Self {
+        ChunkWithProof::new(&[0xFF; 8 * 1024 * 1024], 0).expect("the chunk to be correctly created")
```
rafal-ch:
```suggestion
        ChunkWithProof::new(&[0xFF; CHUNK_SIZE_BYTES], 0).expect("the chunk to be correctly created")
```

node/src/utils/specimen.rs
rafal-ch, line 1: An idea:

Is there an easy way to see what's the largest specimen for each type? We could set a baseline and then monitor changes for each release.
Or, we could even add a check: "hey developer, your PR changed the largest specimen for T`.

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+            LargestSpecimen::largest_specimen(estimator, cache),
+            LargestSpecimen::largest_specimen(estimator, cache),
+            Some(btree_map_distinct_from_prop(
+                estimator,
+                "validator_count",
+                cache,
+            )),
+            LargestSpecimen::largest_specimen(estimator, cache),
+        )
+        .expect("did not expect largest speciment creation of block to fail")
```
rafal-ch:
```suggestion
        .expect("did not expect largest specimen creation of block to fail")
```

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+impl LargestSpecimen for Digest {
+    fn largest_specimen<E: SizeEstimator>(_estimator: &E, _cache: &mut Cache) -> Self {
+        // Hashes are fixed size by definition, so any value will do.
+        Digest::hash("")
+    }
+}
+
+impl LargestSpecimen for BlockPayload {
+    fn largest_specimen<E: SizeEstimator>(estimator: &E, cache: &mut Cache) -> Self {
+        // We cannot just use the standard largest speciment for `DeployHashWithApprovals`, as this
```
rafal-ch:
```suggestion
        // We cannot just use the standard largest specimen for `DeployHashWithApprovals`, as this
```

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+        fn generate_key<E: SizeEstimator>(estimator: &E, seed: usize) -> PublicKey {
+            // Like `Signature`, we do not wish to pollute the types crate here.
+            #[derive(Copy, Clone, Debug, EnumIter)]
+            enum PublicKeyDiscriminants {
+                System,
+                Ed25519,
+                Secp256k1,
+            }
+            largest_variant::<PublicKey, PublicKeyDiscriminants, _, _>(estimator, |variant| {
+                // We take advantange of two things here:
```
rafal-ch:
```suggestion
                // We take advantage of two things here:
```

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+                //
+                // 1. The required seed bytes for Ed25519 and Secp256k1 are both the same length of
+                //    32 bytes.
+                // 2. While Secp256k1 does not allow the most trivial seed bytes of 0x00..0001, a
+                //    a hash function output seems to satisfay it, and our current hashing scheme
```
rafal-ch:
```suggestion
                //    a hash function output seems to satisfy it, and our current hashing scheme
```

rafal-ch (commented):
Looks good, but I still need a deeper dive since this is quite massive.

So far just a few nits and one idea :)

node/src/types/chainspec.rs
```diff
@@ -85,14 +91,20 @@ pub struct Chainspec {
...
+        let serialized = generate_largest_message(&estimator, &mut cache);
+
+        if serialized.len() + CHAINSPEC_NETWORK_MESSAGE_SAFETY_MARGIN
+            > self.network_config.maximum_net_message_size as usize
         {
-            warn!(
-                "config value [network][maximum_net_message_size] should be set to at least
-            CHUNK_SIZE_BYTES * 3 ({})",
-                ChunkWithProof::CHUNK_SIZE_BYTES * 3
+            warn!(calculated_length=serialized.len(), configured_maximum=self.network_config.maximum_net_message_size,
```
Fraser999, line 106: Shouldn't we be returning `false` here now that we're not taking a rough guess at whether the config value is unsuitable?

node/src/components/network/message.rs
```diff
@@ -418,6 +428,157 @@ pub struct EstimatorWeights {
...
+impl<'a> SizeEstimator for NetworkMessageEstimator<'a> {
+    fn estimate<T: Serialize>(&self, val: &T) -> usize {
+        serialize_net_message(&val).len()
+    }
+
+    fn get_parameter(&self, name: &'static str) -> Option<i64> {
+        Some(match name {
+            // The name limit will be larger than the actual name, so it is a safe upper bound.
+            "network_name_limit" => self.chainspec.network_config.name.len() as i64,
+            // These limits are making deploys bigger then they actually are, since many items
```
Fraser999:
```suggestion
            // These limits are making deploys bigger than they actually are, since many items
```

node/src/components/network/message.rs
```diff
@@ -418,6 +428,157 @@ pub struct EstimatorWeights {
...
+            "max_transfers_per_block" => {
+                self.chainspec.deploy_config.block_max_transfer_count as i64
+            }
+            "average_approvals_per_deploy_in_block" => {
+                let max_total_deploys = (self.chainspec.deploy_config.block_max_deploy_count
+                    + self.chainspec.deploy_config.block_max_transfer_count)
+                    as i64;
+
+                // Note: The +1 is to overestimate, as depending on the serialization format chosen,
+                //       spreading out the approvals can be increase or decrease the size. For
```
Fraser999:
```suggestion
                //       spreading out the approvals can increase or decrease the size. For
```

node/src/types/deploy/approval.rs
```diff
@@ -40,6 +40,11 @@ impl Approval {
     pub fn signature(&self) -> &Signature {
         &self.signature
     }
+
+    /// Constructs an `Approval` directly without any checks or additional calculations.
+    pub fn from_parts(signer: PublicKey, signature: Signature) -> Self {
```
Fraser999: I'd prefer if this were `pub(crate)` since we don't currently have an easy way to create an invalid Approval.

Better still, implementing `LargestSpecimen for Approval` here in approvals.rs would allow us to remove this method entirely.

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
+//! Specimen support.
+//!
+//! Structs implementing the specimen trait allow for specific sample instances being created, such
+//! as the biggest possible.
+
+use core::convert::TryInto;
```
Fraser999: Should be `std` rather than `core`?

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+            .get_parameter(name)
+            .unwrap_or_else(|| panic!("missing parameter \"{}\" for specimen estimation", name));
+
+        T::try_from(value).unwrap_or_else(|_| {
+            panic!(
+                "Failed to convert the parameter `{name}` of value `{value}` to the type `{}`",
+                core::any::type_name::<T>()
+            )
+        })
+    }
```
Fraser999: I think I understand why we've ended up with two methods here, but since we'll probably only ever implement `get_parameter` in one place, I'd prefer if this trait were simplified by merging the two.  We can rename to just `parameter` to return `T` and have it panic as required inside the implementation.

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+
+    /// Require a parameter, cast into a boolean.
+    ///
+    /// See `require_parameter` for details. Will return `false` if the stored value is `0`,
+    /// otherwise `true`.
+    ///
+    /// ## Panics
+    ///
+    /// Same as `require_parameter`.
+    fn require_parameter_bool(&self, name: &'static str) -> bool {
```
Fraser999: I'd rename this to just `parameter_as_bool` or `parameter_bool` if we make the change suggested above.

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+    }
+}
+
+impl LargestSpecimen for Deploy {
+    fn largest_specimen<E: SizeEstimator>(estimator: &E, cache: &mut Cache) -> Self {
+        Deploy::new(
+            LargestSpecimen::largest_specimen(estimator, cache),
+            LargestSpecimen::largest_specimen(estimator, cache),
+            LargestSpecimen::largest_specimen(estimator, cache),
+            vec![/* Legacy field, always empty */],
```
Fraser999: Unfortunately although this _is_ a deprecated field, Deploys are still deemed valid if up to `[deploys.max_dependencies]` hashes are specified here.

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+
+impl<E> LargeUniqueSequence<E> for Approval
+where
+    Self: Sized + Ord,
+    E: SizeEstimator,
+{
+    fn large_unique_sequence(estimator: &E, count: usize, cache: &mut Cache) -> BTreeSet<Self> {
+        PublicKey::large_unique_sequence(estimator, count, cache)
+            .into_iter()
+            .map(|pk| Approval::from_parts(pk, LargestSpecimen::largest_specimen(estimator, cache)))
```
Fraser999: You triggered me :) 
```suggestion
            .map(|public_key| Approval::from_parts(public_key, LargestSpecimen::largest_specimen(estimator, cache)))
```

node/src/utils/specimen.rs
```diff
@@ -0,0 +1,951 @@
...
+    fn largest_specimen<E: SizeEstimator>(estimator: &E, cache: &mut Cache) -> Self {
+        ContractPackageHash::new(
+            [LargestSpecimen::largest_specimen(estimator, cache); KEY_HASH_LENGTH],
+        )
+    }
+}
+
+impl LargestSpecimen for RuntimeArgs {
+    fn largest_specimen<E: SizeEstimator>(_estimator: &E, _cache: &mut Cache) -> Self {
+        Default::default()
```
Fraser999: An empty vec doesn't seem right here?

We have `[deploys.session_args_max_length]` and `[deploys.payment_args_max_length]` which determine the maximum sizes (bytesrepr-encoded) of the `RuntimeArgs` in a deploy.

node/src/components/network/message.rs
```diff
@@ -418,6 +428,157 @@ pub struct EstimatorWeights {
...
+                ((self.chainspec.deploy_config.block_max_approval_count as i64 + max_total_deploys
+                    - 1)
+                    / max_total_deploys)
+                    .max(0)
+                    + 1
+            }
+            "max_accusations_per_block" => self.chainspec.core_config.validator_slots as i64,
+            // `RADIX` from EE.
+            "max_pointer_per_node" => 255,
+            "endorsements_disabled" => 1,
```
Fraser999: We're negating this at all callsites, so I'd suggest renaming it to `endorsements_enabled` and setting it to `0`.

Also, it might be worth a code comment explaining why this isn't being derived from a chainspec value?

marc-casperlabs: AFAIK it is hard-disabled in the code as a workaround, thus also the inversion. I am assuming either the code will be removed entirely, or the feature re-enabled, both cases obviating the need for the setting.

node/src/types/chainspec.rs
marc-casperlabs: Yes, that's a good point. I simply missed the code was not returning `false` before.

node/src/utils/specimen.rs
marc-casperlabs: Well, the easy way to see what the largest specimen is is simply called the `largest_specimen` method. It does depend on a chainspec, however.

I don't think it is necessary to do this for each type at first, if changes are made, a developer would know what they changed - so maybe initially we can do this just for `Message`? The `bundled_production_chainspec_is_valid()` check already includes this, except it does not record the size of the generated largest specimen. This would mean refactoring this small bit of data, then adding a test that compares it to a constant.

node/src/utils/specimen.rs
marc-casperlabs: :100:

node/src/utils/specimen.rs
marc-casperlabs: Has been removed, as we don't need it if we are just generating the maximum deploy size as bytecode.

node/src/utils/specimen.rs
marc-casperlabs: Let's just ~~fill it with `max_dependencies` deploy hashes then!~~ rely on the session code blowing it up enough.

node/src/utils/specimen.rs
marc-casperlabs: Let's rename ~~`get_require` to `require`~~ `require_parameter` to `parameter` and combine both.

node/src/types/deploy/approval.rs
marc-casperlabs: Fair enough!

node/src/utils/specimen.rs
rafal-ch: Definitely makes sense. I'll appreciate this additional check of `Message` type maximum size against an expected constant.

node/src/types/deploy/approval.rs
marc-casperlabs: Fixed in 3c583cab.

node/src/utils/specimen.rs
marc-casperlabs: Fixed in 6b51cd9b0

node/src/utils/specimen.rs
marc-casperlabs: Fixed in 6b51cd9b0

node/src/utils/specimen.rs
marc-casperlabs: Fixed in 3393622d0

node/src/utils/specimen.rs
marc-casperlabs: https://github.com/casper-network/casper-node/issues/3818

marc-casperlabs:
bors try

rafal-ch (approved):
LGTM :+1:

Fraser999 (approved):

marc-casperlabs:
bors merge

PR #3713, opened 2023-02-27
NCTL: Update nightly upgrade tests to use common sync check

TomVasile:
Changes:
- reworked how we were checking nodes in sync to use `check_network_sync` from the other nightlies
- adds checking the reactor state is in either keepup or validate

ticket: https://app.zenhub.com/workspaces/cn-sre-60d363546afb85000e491ff6/issues/gh/casper-network/sre/715

### DIFF
--- a/utils/nctl/sh/scenarios-upgrades/upgrade_scenario_06.sh
+++ b/utils/nctl/sh/scenarios-upgrades/upgrade_scenario_06.sh
@@ -13,10 +13,11 @@
 # Step 08: Await 1 era.
 # Step 09: Stage nodes 1&10 and restart.
 # Step 10: Assert all nodes are running
-# Step 11: Assert lfbs are in sync
-# Step 12: Assert chain didn't stall.
-# Step 13: Run Health Checks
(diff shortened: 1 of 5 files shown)


utils/nctl/sh/scenarios-upgrades/upgrade_scenario_06.sh
```diff
@@ -284,46 +286,60 @@ function _step_10()
     fi
 }
 
-# Step 11: Assert lfbs are in sync
+# Step 11: Check reactor state
```
rafal-ch, line 289: Looks like the code used in `function _step_11()` is repeated in other tests. Did you consider extracting it to some utility file?

rafal-ch (approved):
LGTM :+1: 
with one suggestion

utils/nctl/sh/scenarios-upgrades/upgrade_scenario_06.sh
TomVasile: discussed in slack, leaving this as is for now

PR #3725, opened 2023-03-02
tightening up validate / accumulator flow

EdHastingsCasperAssociation:


EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

node/src/reactor/main_reactor/keep_up.rs
```diff
@@ -688,9 +687,22 @@ impl MainReactor {
                     Err(err) => Err(err.to_string()),
                 }
             }
-            None => {
-                debug!("historical: did not find any orphaned block headers");
-                Ok(None)
+            HighestOrphanedBlockResult::MissingFromBlockHeightIndex(block_height) => {
+                if block_height == 0 {
+                    Ok(Some(SyncBackInstruction::GenesisSynced))
```
alsrdn: It's not really clear why we return `GenesisSynced` here if the block is missing from the height index. Wouldn't the block be in the height index if we were truly synced to genesis?

node/src/components/block_accumulator.rs
```diff
@@ -153,70 +133,46 @@ impl BlockAccumulator {
     }
 
     pub(crate) fn sync_instruction(&mut self, sync_identifier: SyncIdentifier) -> SyncInstruction {
+        let unset_local_tip = self.local_tip.is_none(); // first time thru, leap
```
alsrdn: nit: maybe `had_no_local_tip` instead of `unset_local_tip` is more clear. WDYT?

node/src/components/block_accumulator/sync_identifier.rs
```diff
@@ -43,6 +43,22 @@ impl SyncIdentifier {
...
+            SyncIdentifier::SyncedBlockIdentifier(_, _, era_id)
+            | SyncIdentifier::ExecutingBlockIdentifier(_, _, era_id)
+            | SyncIdentifier::LocalTip(_, _, era_id) => Some(*era_id),
+        }
+    }
+
+    pub(crate) fn block_height_and_era(&self) -> Option<(u64, EraId)> {
+        match (self.block_height(), self.era_id()) {
+            (None, _) | (_, None) => None,
+            (Some(block_height), Some(era_id)) => Some((block_height, era_id)),
```
alsrdn: nit: we can move this arm up and then do a default match for all the other cases.

node/src/components/block_accumulator.rs
```diff
@@ -153,70 +133,46 @@ impl BlockAccumulator {
...
+        if let Some((block_height, era_id)) = sync_identifier.block_height_and_era() {
+            self.register_local_tip(block_height, era_id);
+        }
+        if unset_local_tip || self.should_leap(self.maybe_block_height(&sync_identifier)) {
+            return SyncInstruction::Leap {
```
alsrdn, line 144: it would be good to add a debug message here to show why we are leaping.
`if unset_local_tip { debug!(%block_hash, "Leap: Block Accumulator: leaping because didn't have local tip);}`

node/src/reactor/main_reactor/validate.rs
```diff
@@ -25,9 +25,12 @@ impl MainReactor {
         rng: &mut NodeRng,
     ) -> ValidateInstruction {
         let queue_depth = self.contract_runtime.queue_depth();
-        if self.contract_runtime.queue_depth() > 0 {
-            debug!("Validate: should_validate queue_depth {}", queue_depth);
-            return ValidateInstruction::KeepUp;
+        if queue_depth > 0 {
+            warn!("Validate: should_validate queue_depth {}", queue_depth);
+            return ValidateInstruction::CheckLater(
```
alsrdn, line 30: So basically this means that we prefer to continue to execute rather than try to keep-up if the queue is backed uo. 
But if this validator continues to be slow and it keeps lagging behind with more and more blocks, do we need an exit condition from here?

node/src/reactor/main_reactor/keep_up.rs
EdHastingsCasperAssociation: yes, that's true. the real issue is it should not be possible for completed_blocks to have _any_ entry that doesn't have a corresponding entry in the block height index and / or that isn't actually complete. 

thus, this should be unreachable. but if it were reached, for the purpose of this logic (historical state) which is concerned with deciding whether the node should attempt to get the parent of this block (which will fail if attempted if this block is not actually complete), it seemed preferable to allow the node to keep running forward if the low bound is reported as 0 since we know that the node wont attempt to sync its parent (a non existent block at block_height == -1) and thus naturally avoids the pitfall. ie since height 0 is the left bound terminator, the process will not run itself off the edge if allowed to continue.

but, i will remove this special case and have it Err even if height is 0.

node/src/components/block_accumulator.rs
EdHastingsCasperAssociation: ok

node/src/components/block_accumulator/sync_identifier.rs
EdHastingsCasperAssociation: we generally avoid wildcard matches, but I'll do an `if let Some = ` if you prefer...seems less readable to me personally...
```
        if let (Some(block_height), Some(era_id)) = (self.block_height(), self.era_id()) {
            return Some((block_height, era_id));
        }
        None
```

node/src/components/block_accumulator.rs
EdHastingsCasperAssociation: that's not necessarily what it means. the accumulator is purposely not aware of the local tip after a start / restart, even if the node itself does have a local tip in storage. the sync_identifier imputed into this function has a block hash which may be the local tip in a few flows. the accumulator is being asked "have you heard about a higher block than the one i am asking you about"? initially, the block it is being asked about is the first block the accumulator becomes aware of. it thus can't answer the question and says to leap which means ask peers if they are aware of an even higher block.

node/src/reactor/main_reactor/validate.rs
EdHastingsCasperAssociation: it means it is preferable to remain in validate mode and allow time for the q to clear.

rafal-ch (approved):
LGTM :+1:

alsrdn (approved):

PR #3726, opened 2023-03-03
Add tests to `SyncLeap` struct

### LINKED ISSUE #3579: Test SyncLeap record
Test SyncLeap

rafal-ch:
This PR adds unit tests to the `SyncLeap` struct and the `FetchItem::<SyncLeap>::validate()` function.

It also updates the latter to actually [defer the cryptographic verification](https://github.com/casper-network/casper-node/commit/22de34884e63dcdd8be964a686f279fbac70649a#diff-a3a4f45b3188dbcf8a07d338a5d501e9577602c51541139461ce335cbd55e6d9R344-R351) to the end, so we don't waste cycles.

Closes #3579


rafal-ch:
bors try

node/src/types/sync_leap.rs
```diff
@@ -191,10 +196,45 @@ impl Display for SyncLeap {
     }
 }
 
+#[derive(Clone, DataSize, Debug, Eq, PartialEq, Serialize)]
+pub(crate) struct SyncLeapValidationMetaData {
```
EdHastingsCasperAssociation: can you put this into its own file, please?

rafal-ch: Certainly! Updated in 5ecdd81cdfaf9130fcaccac19e423214af429d3f

georgepisaltu (approved):
LGTM

node/src/lib.rs
```diff
@@ -126,3 +126,235 @@ pub(crate) fn new_rng() -> NodeRng {
...
+                self.block.header().protocol_version(),
+            )
+            .unwrap();
+            self.block = next.clone();
+            Some(next)
+        }
+    }
+
+    #[test]
+    fn test_block_iter() {
```
Fraser999: We tend to not name our unit tests with `test_` prefix.  Maybe `should_create_valid_chain` or something?

node/src/lib.rs
```diff
@@ -126,3 +126,235 @@ pub(crate) fn new_rng() -> NodeRng {
...
+                current_block.header().parent_hash(),
+                parent_block.hash(),
+                "block's parent should point at previous block"
+            );
+            parent_block = current_block;
+        }
+    }
+
+    #[test]
+    fn test_block_iter_creates_switch_blocks() {
```
Fraser999: Maybe `should_create_switch_blocks`?

node/src/types/block.rs
```diff
@@ -876,6 +853,18 @@ impl BlockHeader {
...
+    #[cfg(any(feature = "testing", test))]
+    pub fn set_height(&mut self, height: u64) {
+        self.height = height;
+    }
+
+    /// Sets the era of this block.
+    #[cfg(any(feature = "testing", test))]
+    pub fn set_era(&mut self, era: EraId) {
+        self.era_id = era;
+    }
```
Fraser999: I think we should avoid adding this (and the `header_mut` method below) since it allows us to invalidate a block.  Because we memoize the `block_hash` in the block header, it's even less apparent that the block is actually invalid.

node/src/types/block.rs
```diff
@@ -1457,6 +1451,26 @@ impl Block {
...
+
+    /// Generates a random non-switch block.
+    #[cfg(any(feature = "testing", test))]
+    pub fn random_non_switch_block(rng: &mut TestRng) -> Self {
+        let mut block = Block::random(rng);
+        block.header.era_end = None;
+        block
+    }
+
```
Fraser999: As above, I'd prefer if these were either removed, or were implemented by calling though to `Block::random_with_specifics`, since these generate invalid blocks.

node/src/types/block.rs
```diff
@@ -2526,141 +2540,33 @@ mod tests {
...
-    fn test_block_iter_creates_switch_blocks() {
-        let switch_block_indices = vec![0, 10, 76];
-
-        let rng = TestRng::new();
-        let test_block = TestBlockSpec::new(rng, Some(switch_block_indices.clone()));
-        let block_batch: Vec<_> = test_block.into_iter().take(100).collect();
-
-        let base_height = block_batch.first().expect("should have block").height();
+#[cfg(any(feature = "testing", test))]
+pub fn random_era_report(rng: &mut TestRng) -> EraReport {
```
Fraser999: This doesn't need to be `pub` I think, but if we remove `random_switch_block` (or implement it using `Block::random_with_specifics`) then this function can also be moved back to inside `FinalizedBlock::random_with_specifics`.

node/src/types/sync_leap_validation_metadata.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    pub(crate) fn from_chainspec(chainspec: &Chainspec) -> Self {
+        Self {
+            recent_era_count: chainspec.core_config.recent_era_count(),
+            activation_point: chainspec.protocol_config.activation_point,
+            global_state_update: chainspec.protocol_config.global_state_update.clone(),
+            finality_threshold_fraction: chainspec.core_config.finality_threshold_fraction,
+        }
+    }
```
Fraser999: Seems like a good case for `impl From`?

node/src/lib.rs
```diff
@@ -126,3 +126,235 @@ pub(crate) fn new_rng() -> NodeRng {
 pub(crate) fn new_rng() -> NodeRng {
     NodeRng::new()
 }
+
+#[cfg(test)]
+mod tests {
```
Fraser999: Seems like this should maybe just live inside the sync leap test module, since that's the only place these helpers are consumed?

If you feel that they will be used more widely in the future, then we have the `crate::testing` module which might be more appropriate?

Fraser999 (commented):
Some initial feedback, but not a complete review yet.

EdHastingsCasperAssociation:
I generally agree w/ Fraser's comments. Would like to see them addressed, and then getting these tests merged in soonish.

node/src/lib.rs
rafal-ch: Updated in 75044ac57adf91f1f801d7cd29336f1fa6e4a64e

node/src/lib.rs
rafal-ch: Updated in 75044ac57adf91f1f801d7cd29336f1fa6e4a64e

node/src/types/block.rs
rafal-ch: Addressed in f7fc9ddb51f53af302de04901c79bb82b9412923

node/src/types/block.rs
rafal-ch: Addressed in 517f01d82ec1e1a611834595cfe8253984cd8c26
I decided to keep the helper functions, but they now use the `random_with_specifics()` internally.

node/src/types/block.rs
rafal-ch: Updated in 00bf59f5a353591de2ec1c7346075c02d17df4c4

node/src/types/sync_leap_validation_metadata.rs
rafal-ch: Updated in b59e7fe416d6099ea60db62b7ab62f0288f042bb

node/src/lib.rs
rafal-ch: Yes, good idea. Historically, this code was moved around since initially it was used to test the [batch block header fetch](https://github.com/casper-network/casper-node/commit/ee3057a41094d86cbc654463d1fe6ee45ce61aae) in the original fast sync implementation.

Addressed in 2d9e61d1ab95cdf8d85bfbf2f13afe56b632a9bf

rafal-ch:
> I generally agree w/ Fraser's comments. Would like to see them addressed, and then getting these tests merged in soonish.

Comments are addressed, but this was just an [initial review](https://github.com/casper-network/casper-node/pull/3726#pullrequestreview-1345737805) from @Fraser999, so I'll refrain from merging until there are more approvals.

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+        pub(crate) weight: Option<U512>,
+    }
+
+    // Utility struct that can be turned into an iterator that generates
+    // continuous and descending blocks (i.e. blocks that have consecutive height
+    // and parent hashes are correctly set). The height of the first block
+    // in a series is chosen randomly.
+    //
+    // Additionally, this struct allows to generate switch blocks at a specific location in the
+    // chain, for example: Setting `switch_block_indices` to [1; 3] and generating 5 blocks will
```
Fraser999:
```suggestion
    // chain, for example: Setting `switch_block_indices` to [1, 3] and generating 5 blocks will
```

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+            signed_block_headers: Default::default(),
+        };
+        let result = sync_leap.validate(&validation_metadata);
+        assert!(matches!(
+            result,
+            Err(SyncLeapValidationError::MissingTrustedAncestors)
+        ));
+
+        // When trusted block height is 0 and trusted ancestors are empty, validate
+        // should yield a result different than `SyncLeapValidationError::MissingTrustedAncestors`.
```
Fraser999:
```suggestion
        // When trusted block height is 0, validation should not fail due trusted ancestors being
        // empty.
```

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+            trusted_ancestor_only: false,
+            trusted_block_header: block.take_header(),
+            trusted_ancestor_headers: Default::default(),
+            signed_block_headers: Default::default(),
+        };
+        let result = sync_leap.validate(&validation_metadata);
+        assert!(!matches!(
+            result,
+            Err(SyncLeapValidationError::MissingTrustedAncestors)
+        ));
```
Fraser999, line 583: It might be worth either asserting that no error is actually returned?

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+
+        // Max allowed size should NOT trigger the `TooManySwitchBlocks` error.
+        let generated_block_count = max_allowed_size;
+
+        let block = random_block_at_height(&mut rng, 0);
+        let sync_leap = SyncLeap {
+            trusted_ancestor_only: false,
+            trusted_block_header: block.take_header(),
+            trusted_ancestor_headers: Default::default(),
+            signed_block_headers: std::iter::repeat_with(|| {
```
Fraser999:
```suggestion
            signed_block_headers: iter::repeat_with(|| {
```

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+        let sync_leap = make_test_sync_leap(
+            &mut rng,
+            &switch_blocks,
+            query,
+            &trusted_ancestor_headers,
+            &signed_block_headers,
+            add_proofs,
+        );
+
+        // `sync_leap` is a well formed SyncLeap structure for the test chain. We can use the blocks
```
Fraser999: For this comment, I'd maybe mention that while it's well-formed, it's invalid.

Same point for the copies of this comment in the following tests.

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+                .0
+                .block_hash(),
+            highest_block.block_hash()
+        );
+        assert_eq!(sync_leap.highest_block_hash(), highest_block_hash);
+        assert_eq!(sync_leap.highest_block_height(), highest_block_height);
+    }
+
+    #[test]
+    fn should_return_highest_block_header_from_trusted_ancestors() {
```
Fraser999, line 1248: I'm not sure we need this test do we?  It's testing for something which is impossible from a valid `SyncLeap` unless I'm misunderstanding the intent?

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+            .take(2)
+            .cloned()
+            .map(|block_header_with_metadata| block_header_with_metadata.block_header)
+            .collect();
+        let sync_leap = SyncLeap {
+            trusted_ancestor_only: false,
+            trusted_block_header: lowest_blocks.first().unwrap().clone(),
+            trusted_ancestor_headers: middle_blocks,
+            signed_block_headers: vec![highest_block],
+        };
```
Fraser999: The initial `sync_leap` (which has the benefit of being valid) will suffice here won't it?  These lines could be removed I think.

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+            .iter()
+            .take(2)
+            .cloned()
+            .collect();
+        let sync_leap = SyncLeap {
+            trusted_ancestor_only: false,
+            trusted_block_header: highest_block.block_header,
+            trusted_ancestor_headers: lowest_blocks,
+            signed_block_headers: middle_blocks,
+        };
```
Fraser999, line 1441: Similar to above, if we construct the initial `sync_leap` with no signed block headers, it will be valid and would allow us to remove all this code.

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+            &switch_blocks,
+            query,
+            &trusted_ancestor_headers,
+            &signed_block_headers,
+            add_proofs,
+        );
+
+        let fault_tolerance_fraction = Ratio::new_raw(1, 3);
+
+        let mut switch_block_iter = sync_leap.signed_block_headers.iter();
```
Fraser999: Nitpick on the name: these aren't all switch blocks (the third one isn't).

node/src/types/sync_leap.rs
```diff
@@ -290,20 +295,1676 @@ impl FetchItem for SyncLeap {
...
+    fn era_validator_weights_with_genesis_without_switch_block_preceding_immediate_switch_block() {
+        let mut rng = TestRng::new();
+
+        let trusted_block = Block::random_non_switch_block(&mut rng);
+
+        let (
+            signed_block_header_with_metadata_1,
+            signed_block_header_with_metadata_2,
+            signed_block_header_with_metadata_3,
+        ) = make_three_switch_blocks_at_era_and_height(rng, (0, 0), (2, 20), (3, 30));
```
Fraser999: I guess I'd have used `(0, 0), (1, 10), (2, 20)` to make it slightly more realistic, but not a big deal :)

Fraser999 (approved):
Nitpicks only - feel free to ignore.  LTGM!

node/src/types/sync_leap.rs
rafal-ch: It refers to the `sync_leap` created above, which is actually valid.

node/src/types/sync_leap.rs
rafal-ch: As long as `highest_block_header_and_signatures()` references `self.trusted_block_header` I think there's a merit in having this scenario covered.

node/src/types/sync_leap.rs
rafal-ch: I'd rather keep this as is, because if we create a `SyncLeap` without `signed_block_headers` we're not going to cover [this line](https://github.com/casper-network/casper-node/blob/dev/node/src/types/sync_leap.rs#L167) with the test.

Since the iteration on switch blocks will not happen, I can, for example, change `==` to `!=` in the implementation and the test will still pass.

node/src/types/sync_leap.rs
rafal-ch: Yup, good suggestion.
Updated in b92726a41b6b5bbebeb51418b32ff4e858816f64

node/src/types/sync_leap.rs
rafal-ch: Updated in b92726a41b6b5bbebeb51418b32ff4e858816f64

node/src/types/sync_leap.rs
rafal-ch: Updated in b92726a41b6b5bbebeb51418b32ff4e858816f64

node/src/types/sync_leap.rs
rafal-ch: Additional assert added in b92726a41b6b5bbebeb51418b32ff4e858816f64

node/src/types/sync_leap.rs
rafal-ch: Updated in b92726a41b6b5bbebeb51418b32ff4e858816f64

node/src/types/sync_leap.rs
rafal-ch: Nice one :)
Updated in b92726a41b6b5bbebeb51418b32ff4e858816f64

node/src/types/sync_leap.rs
rafal-ch: Updated in b92726a41b6b5bbebeb51418b32ff4e858816f64

node/src/types/sync_leap.rs
rafal-ch: Name and meaning has been clarified in f7c9bb3cb81c057f59108a32b17c4aa3e784d6fb and 61c99ee394f256df34b5c29fb55512b81411369d

node/src/types/sync_leap.rs
rafal-ch: As discussed, we'll keep as is for now.

node/src/types/sync_leap.rs
rafal-ch: As discussed, we'll keep as is for now.

PR #3727, opened 2023-03-06
bugfix: completed blocks write anomaly

EdHastingsCasperAssociation:
bug fixes: due to an interaction between the block & finality signature fetchers, the accumulator, and the block synchronizer, it was possible for a block to get marked complete before it was actually stored, which would result in a later unrecoverable error if anything else attempted to read back the supposedly complete block data from storage prior to it actually being stored. this was an edge condition, but when it did occur a node's completed block record was left in an unrecoverable state.

separately, removed a risky optimization added recently that suppressed sync leaps considered to be redundant, which did not consider that other nodes' responses to a request for the same block_hash can differ over time; also, on a larger network with more peers it is healthy to get a broader response. 

finally, added logic to the block synchronizer to actively acquire era validator weights if needed rather than wait for them to potentially be provided which could be a terminal condition if the needed weights were not acquired by other means.

### DIFF
--- a/node/src/components/block_synchronizer.rs
+++ b/node/src/components/block_synchronizer.rs
@@ -53,8 +53,9 @@ use crate::{
     rpcs::docs::DocExample,
     types::{
         ApprovalsHashes, Block, BlockExecutionResultsOrChunk, BlockHash, BlockHeader,
-        BlockSignatures, Deploy, FinalitySignature, FinalitySignatureId, LegacyDeploy, MetaBlock,
-        MetaBlockState, NodeId, SyncLeap, TrieOrChunk, ValidatorMatrix,
+        BlockSignatures, Chainspec, Deploy, FinalitySignature, FinalitySignatureId, LegacyDeploy,
+        MetaBlock, MetaBlockState, NodeId, SyncLeap, SyncLeapIdentifier, TrieOrChunk,
+        ValidatorMatrix,
     },
     NodeRng,
 };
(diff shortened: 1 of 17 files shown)


node/src/components/block_synchronizer.rs
```diff
@@ -898,6 +911,55 @@ impl BlockSynchronizer {
...
+            for evw in era_validator_weights {
+                self.validator_matrix.register_era_validator_weights(evw);
+            }
+        }
+        match (&mut self.forward, &mut self.historical) {
+            (Some(builder), _) | (_, Some(builder)) if builder.block_hash() == block_hash => {
+                if demote_peer {
+                    builder.demote_peer(maybe_peer_id);
+                } else {
+                    builder.register_era_validator_weights(&self.validator_matrix);
```
Fraser999, line 956: Should we promote the peer in this case?

node/src/components/block_synchronizer.rs
```diff
@@ -898,6 +911,55 @@ impl BlockSynchronizer {
...
+                    "BlockSynchronizer: fetched sync leap {:?} from peer {}",
+                    item.fetch_id().block_hash(),
+                    peer
+                );
+
+                (item.fetch_id().block_hash(), Some(item), Some(peer))
+            }
+            Ok(FetchedData::FromStorage { item }) => {
+                (item.fetch_id().block_hash(), Some(item), None)
+            }
```
georgepisaltu: Sync leaps should never be fetched from storage, right? So this should be an error.

georgepisaltu (commented):
LGTM, just one question.

Fraser999 (approved):

node/src/components/block_synchronizer.rs
EdHastingsCasperAssociation: addressed

node/src/components/block_synchronizer.rs
EdHastingsCasperAssociation: agreed, addressed

rafal-ch (approved):
LGTM :+1:

Issue #3704, opened 2023-02-23
Avoid sync leap just before upgrade


devendran-m:
`BlockAccumulator::should_leap` should not return true if it has an acceptor for the final switch block of the current protocol version, and it has at least weak finality on that block.  This will probably mean providing a new `BlockAccumulator::Event` variant like `NextUpgrade` which will be triggered on an `UpgradeWatcherAnnouncement::UpgradeActivationPointRead` .

PR #3731, opened 2023-03-08
#3704 Avoid sync leap just before upgrade

EdHastingsCasperAssociation:
as the title says, and also fixed an incorrect state change to validate mode for non-validators present at genesis which was revealed by tighter leap instruction tracking and testing


rafal-ch:
bors try

Fraser999 (approved):

node/src/components/block_accumulator/leap_instruction.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub(super) fn from_execution_threshold(
+        attempt_execution_threshold: u64,
+        distance_from_highest_known_block: u64,
+        is_upgrade_boundary: bool,
+    ) -> Self {
+        if distance_from_highest_known_block == 0 {
+            return LeapInstruction::AtHighestKnownBlock;
+        }
+        if is_upgrade_boundary
+            && distance_from_highest_known_block <= attempt_execution_threshold * 2
```
rafal-ch, line 28: May be worthwhile to explain in a comment why we think `* 2` is enough.

rafal-ch (approved):
LGTM :+1: 
No blockers.

PR #3735, opened 2023-03-08
Fix purge of builder before validating

### LINKED ISSUE #3733: Investigate the occasionally failing swap_validator_set.sh test
The test is sporadically failing on the CI ([link](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/7674/1/4)).

georgepisaltu:
Fixes #3733 #3738 

This PR introduces a fix to the bug seen in the `swap_validators` test scenario.

The nodes which should become validators in the upcoming era and were previously keeping up are synchronizing the switch block, but before getting to execute it, the associated builder would get purged by the switch to `Validate` mode, triggered by `keep_up_should_validate` into `create_required_eras`. `ContractRuntime` would therefore miss the switch block and we would have a gap which would prevent the node from executing the next blocks.

Another bug which surfaced in this test scenario was that the nodes which were validating and then were supposed to swap places with other nodes in the upcoming era would stall instead of keeping up. From preliminary observations, it seems they switch from `Validate` to `KeepUp` and then to `CatchUp` with a `Leap` instruction because of stale gossip, trying to leap to the same block they just executed before transitioning from `Validate`. The block builder is never doing any `need_next` action, and the synchronizer is stuck. The node fixes itself at some point in the future through a `Leap` when it falls too much behind the actual tip of the chain.

### DIFF
--- a/node/src/components/block_synchronizer/block_synchronizer_progress.rs
+++ b/node/src/components/block_synchronizer/block_synchronizer_progress.rs
@@ -11,6 +11,16 @@ pub(crate) enum BlockSynchronizerProgress {
     Synced(BlockHash, u64, EraId),
 }
 
+impl BlockSynchronizerProgress {
+    pub(crate) fn is_active(&self) -> bool {
+        match self {
+            BlockSynchronizerProgress::Idle | BlockSynchronizerProgress::Synced(_, _, _) => false,
+            BlockSynchronizerProgress::Syncing(_, _, _)
+            | BlockSynchronizerProgress::Executing(_, _, _) => true,
+        }
+    }
+}
+
 impl Display for BlockSynchronizerProgress {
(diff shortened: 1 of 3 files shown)


georgepisaltu:
bors try

georgepisaltu:
bors try

rafal-ch (approved):
LGTM :+1: 
But please wait for more approvals since I was a part of the mob session during which you fixed this.

real-felix (approved):

alsrdn (approved):

Fraser999 (approved):

rafal-ch:
As discussed, the _"Another bug which surfaced in this test scenario (...)"_ is an expected behavior.
cc: @georgepisaltu @EdHastingsCasperLabs @Fraser999

PR #3737, opened 2023-03-08
3720: Panic if there is an overflow

real-felix:
Computing the sum of stakes panics, if there is an overflow.

### DIFF
--- a/node/src/components/consensus/protocols/common.rs
+++ b/node/src/components/consensus/protocols/common.rs
@@ -17,11 +17,14 @@ pub(crate) fn validators<C: Context>(
     inactive: &HashSet<C::ValidatorId>,
     validator_stakes: BTreeMap<C::ValidatorId, U512>,
 ) -> Validators<C::ValidatorId> {
-    let sum_stakes: U512 = validator_stakes.values().copied().sum();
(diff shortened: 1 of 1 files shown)


alsrdn (approved):

node/src/components/consensus/protocols/common.rs
```diff
@@ -17,11 +17,14 @@ pub(crate) fn validators<C: Context>(
     inactive: &HashSet<C::ValidatorId>,
     validator_stakes: BTreeMap<C::ValidatorId, U512>,
 ) -> Validators<C::ValidatorId> {
-    let sum_stakes: U512 = validator_stakes.values().copied().sum();
+    let sum_stakes = safe_sum(validator_stakes.values().copied()).expect("should not overflow");
```
alsrdn, line 20: It would be also nice to have a unit test for this case also.

real-felix: Where would you put it?

real-felix: Done @alsrdn

alsrdn (approved):

rafal-ch (approved):
LGTM :+1:

Issue #3719, opened 2023-02-28
Create a migration that moves the current era info to a stable key, and delete era info objects in trie tip


piotr-dziubecki:
We need to remove the current Key::EraInfo() objects in the trie. There are currently 1 per era that has passed, and this has caused a significant amount of bloat to the tip of the trie.
To do this we need to expose delete trie (done in my branch of the node at r1.4.13-era-info-bloat). We should do this on upgrade and we'll also need to route requests for the old era info to the old switch blocks' state_roots.

RitaMAllenCA:
Fraser's change - block validator fetch bug (new ticket) #3740

PR #3742, opened 2023-03-08
GH-3719 - introduce migration to prune EraInfo from global state

dwerner:
This PR introduces a new chainspec section for migrations, as well as a basic framework for migrations that affect global state and need to be carried out over several blocks.

Most of the interface should be private to the EE, and be manipulated through the values in the chainspec.

Two new `Key` variants representing stable keys for `Key::EraSummary` (which replaces `Key::EraInfo` going forward) as well as `Key::LastMigration`. `Key::LastMigration` is used to store the (u32) id of the highest run migration.

If we add migrations later, we need to ensure that they take this value into account, as migrations equal or less than the current value will **not** be run again.

Also modifies the auction to `record_era_info` to this new stable key (`Key::EraSummary`) and the RPC to query for it, falling back to `Key::EraId` when it's not found.

**And most importantly, after this migration runs, the reachable tip of the trie goes from ~10gb of data down to approx. 780mb in my tests.**


rafal-ch:
@RitaMAllenCA @dwerner 
Can we close this one in favor of #3811?

dwerner:
> @RitaMAllenCA @dwerner Can we close this one in favor of #3811?

yup

PR #3744, opened 2023-03-09
3702 legacy finality threshold

real-felix:
Modifies the legacy finality thresthold during a fast sync:
- Adds a new chainspec parameter that tells how legacy blocks finality behave;
- Fix the logic to take this parameter into account.

### DIFF
--- a/node/src/components/block_synchronizer/block_acquisition_action.rs
+++ b/node/src/components/block_synchronizer/block_acquisition_action.rs
@@ -11,8 +11,8 @@ use crate::{
         ExecutionResultsAcquisition, ExecutionResultsChecksum,
     },
     types::{
-        Block, BlockExecutionResultsOrChunkId, BlockHash, BlockHeader, DeployHash, DeployId,
-        EraValidatorWeights, NodeId,
+        chainspec::LegacyRequiredFinality, Block, BlockExecutionResultsOrChunkId, BlockHash,
+        BlockHeader, DeployHash, DeployId, EraValidatorWeights, NodeId, SignatureWeight,
     },
     NodeRng,
 };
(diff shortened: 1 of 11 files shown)


node/src/components/block_synchronizer/block_acquisition_action.rs
```diff
@@ -296,7 +300,469 @@ impl BlockAcquisitionAction {
...
     }
 }
+
+fn enough_signatures(
+    signature_weight: SignatureWeight,
+    legacy_required_finality: LegacyRequiredFinality,
+    is_checkable: bool,
+    is_historical: bool,
+) -> bool {
+    use SignatureWeight::*;
```
Fraser999: We don't tend to do this (except in the case of tests where we generally do `use super::*;`)

node/src/components/block_synchronizer/block_acquisition_action.rs
```diff
@@ -296,7 +300,469 @@ impl BlockAcquisitionAction {
...
+
+    if is_checkable {
+        // Normal blocks:
+        signature_weight.is_sufficient(is_historical)
```
Fraser999: I think this was a bug in the existing code, but it should be `!is_historical`

node/src/components/block_synchronizer/block_acquisition_action.rs
real-felix: Done

node/src/components/block_synchronizer/block_acquisition_action.rs
real-felix: Done

Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

PR #3746, opened 2023-03-09
Add network soundness test to nightly

### LINKED ISSUE #3728: Implement test to verify if network is stable under the load
The nature of the distributed network makes it hard to cover every edge case with tests, hence there are situations when all the CI pipelines are green, yet we still introduce a regression to the system. This regression is eventually caught in the sre tests, but these are time-consuming, rather expensive and not able to cover every PR. To catch the regressions asap, we should introduce a test that verifies the network stability under the conditions close to real-life.

Such test should:

- spin a network
- put a load on it (deploys, transfers, constant load and bursts) - shall include huge deploys to exercise chunking mechanism
- kill and restart nodes randomly (at reasonable rate)
- join new node(s) repeatedly
- finish after specific amount of time (~30min?) and assert all nodes are in sync
- be a part of the nightly tests pipeline

Next steps (separate tickets):

- put the historical sync in the mix to assure it does not prevent the keepup/validate processes from progressing
- simulate protocol upgrade
- simulate emergency upgrade with changed validator set

rafal-ch:
This PR adds a new test scenario to nightly run:

1. Start the network (nodes 1-5)
2. Start sending deploys in the background
3. Randomly start and stop one or two nodes
4. Incrementally join the nodes (6-10)
5. At each stage assert if the network is progressing - if not, bail with an error
6. Finish test successfully if network runs correctly for 0.5h

For now this test doesn't send huge deploys, it'll be requested in a separate ticket. However, the appropriate chainspec overrides are added.

As a bonus, this PR also adds a `debug` level log for each deploy (with source) that reaches deploy acceptor.

Closes #3728


node/src/components/deploy_acceptor.rs
```diff
@@ -214,6 +214,7 @@ impl DeployAcceptor {
         source: Source,
         maybe_responder: Option<Responder<Result<(), Error>>>,
     ) -> Effects<Event> {
+        debug!(%source, %deploy,  "checking acceptance");
```
Fraser999:
```suggestion
        debug!(%source, %deploy, "checking acceptance");
```

utils/nctl/sh/scenarios/network_soundness.py
```diff
@@ -0,0 +1,244 @@
...
+
+def compile_node():
+    log("*** building nodes ***")
+    command = "nctl-compile"
+    invoke(command)
+
+
+def start_network():
+    log("*** starting network ***")
+    command = "nctl-assets-teardown && NCTL_NODE_LOG_FORMAT=text nctl-assets-setup && RUST_LOG=debug nctl-start"
```
Fraser999: While I strongly prefer the `text` formatting myself when eyeballing logs, I think some of the nctl commands rely on `json` output.

Fraser999 (approved):
I'm no Python expert, but this looks fine to me.

Couple of nitpicks, but no blockers.

node/src/components/deploy_acceptor.rs
rafal-ch: Fixed in 44a9477793f3a7733993f43464c95a722c0e9909

utils/nctl/sh/scenarios/network_soundness.py
rafal-ch: I'll leave the default formatting then. Updated here: 896fcc904d78bb0994c35c8828fdb9c4943c4c89

fizyk20 (approved):

PR #3750, opened 2023-03-10
Allow fetching multiple deploys from same fallback peer in BlockValidator

### LINKED ISSUE #3740: Fix issue fetching deploys in BlockValidator
The scenario is that we're trying to validate a block with multiple deploys, none of which we have stored locally yet.  The proposer is very likely the first peer we try and fetch the deploys from.  If the proposer doesn't provide the first deploy, we try to fetch it from the next peer which has told us it has the deploys [here](https://github.com/casper-network/casper-node/blob/dev/node/src/components/block_validator.rs#L342-L349).  However, that removes the second peer from the list of sources.  When fetching the second deploy from the proposer also times out, we would ideally ask the second peer for that deploy too, but that peer is no longer in the list of sources.  To fix this, fetch requests for all missing deploys should be made as soon as the BlockValidator is made aware of a new source/peer for them.

Fraser999:
This fixes an issue where the BlockValidator would consume `NodeId`s of alternative sources for deploys rather than allowing each source to be queried for multiple deploys.

Now, whenever a new source is learned (via a new `BlockValidation` request) the `BlockValidator` immediately attempts to fetch all deploys from that peer.  The fetcher will take care of not actually generating network traffic for any deploys which are already stored locally.

Closes #3740.

### DIFF
--- a/node/src/components/block_validator.rs
+++ b/node/src/components/block_validator.rs
@@ -12,10 +12,11 @@ mod keyed_counter;
 mod tests;
 
 use std::{
-    collections::{hash_map::Entry, BTreeMap, BTreeSet, HashMap, VecDeque},
+    collections::{BTreeMap, BTreeSet, HashMap},
     convert::Infallible,
     fmt::Debug,
     hash::Hash,
+    marker::PhantomData,
     sync::Arc,
 };
 
@@ -29,6 +30,7 @@ use crate::{
     components::{
         block_proposer::DeployInfo,
         consensus::{ClContext, ProposedBlock},
+        fetcher::FetchResult,
         Component,
     },
     effect::{
@@ -42,8 +44,6 @@ use crate::{
     NodeRng,
 };
 use keyed_counter::KeyedCounter;
-
-use super::fetcher::FetchResult;
 
 #[derive(DataSize, Debug, Display, Clone, Hash, Eq, PartialEq)]
 pub(crate) enum ValidatingBlock {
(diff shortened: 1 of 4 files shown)


rafal-ch (approved):
LGTM :+1:

node/src/components/block_validator.rs
```diff
@@ -270,43 +251,34 @@ where
...
+
+                if state.missing_deploys.is_empty() {
+                    // Block has already been validated successfully, early return to caller.
+                    return responder.respond(true).ignore();
                 }
+
+                // We register ourselves as someone interested in the ultimate validation result.
+                state.responders.push(responder);
+
+                let in_flight = &mut self.in_flight;
```
georgepisaltu: nit: would have liked to not take this extra reference and just use `self.in_flight` to increment.

node/src/components/block_validator/keyed_counter.rs
```diff
@@ -43,7 +39,7 @@ where
                 1
             }
             Some(value) => {
-                value.add_assign(1);
```
georgepisaltu, line 46: What is the reason for this and the `sub_assign` change? Not an issue, just curious.

georgepisaltu (approved):

node/src/components/block_validator.rs
Fraser999: Fair point - fixed in [62c0dee](https://github.com/casper-network/casper-node/pull/3750/commits/62c0dee97415114f65018766476e468595d49324).

node/src/components/block_validator/keyed_counter.rs
Fraser999: Just to be "normal" and also reduce the imports required :)

Fraser999:
bors merge

PR #3760, opened 2023-03-14
 3496: Vendoring pwasm-utils 0.16 under the name wasm-utils

real-felix:
The `pwasm-utils` crate is not maintained anymore (the repository is archived), so this commit vendors it under the name `wasm-utils`. The code is simplified as well, with the ability to use `no-std` being removed since it is not used in the `execution_engine` crate.


wasm-utils/src/export_globals.rs
```diff
@@ -0,0 +1,131 @@
...
+            ),
+        );
+        export_section(module)
+            .expect("added above if does not exists")
+            .entries_mut()
+            .push(new_entry);
+    }
+}
+
+#[cfg(test)]
```
rafal-ch, line 48: I think we need to update our `Makefile` to include these tests in the `test-rs` job.

Fraser999 (changes requested):
I'd prefer if this weren't added as a new crate, but rather as a module inside EE.

If we do want to stick with making this a crate, we'll need to make a few more changes to support that.

I'll block the PR until we get a decision since it shouldn't merge as is.

RitaMAllenCA:
new name: casper-pwasm-utils

real-felix:
PR no more relevant since it's now forked and lives in its own repo.

PR #3763, opened 2023-03-16
Merge core backpressure feature set into 1.6, including ticket system

marc-casperlabs:
This PR is a part of the core 1.6 feature set, namely adding backpressure. This subset has been rebased onto the latest `feat-1.6` and is kept small to make it amenable to being reviewed in isolation.

At its core, it implements two missing features for `muxink`, namely fixed size frames and little endian encoding of fixed size values. The previous design put the encoding at the "wrong end" of the chain, making it impossible to use with multiplexing.

Backpressure is added through a multiplexed back channel for ACKs, with one per outgoing channel. The `Ticket` system is used, meaning that the networking component will consider a message being in the state of "processing" until the ticket is dropped, causing it to send the `ACK` soon after. It is up to individual components to fine-grained precise logic, right now it errs on the side of being too permissive by dropping the ticket early.


RitaMAllenCA:
@marc-casperlabs is there an issue I can connect this to?

marc-casperlabs:
> is there an issue I can connect this to?

This is just "housekeeping" stemming from having these feature branches.

muxink/src/io.rs
```diff
@@ -362,7 +359,7 @@ mod tests {
                 MAX_READ_BUF_INCREMENT,
                 MAX_READ_BUF_INCREMENT,
                 MAX_READ_BUF_INCREMENT,
-                MAX_READ_BUF_INCREMENT - mem::size_of::<u16>()
+                MAX_READ_BUF_INCREMENT - (<u16>::BITS / 8) as usize
```
rafal-ch, line 362: We can use `LENGTH_MARKER_SIZE` const in this (and similar) places.

PR #3764, opened 2023-03-16
Global state synchronizer re-write

alsrdn:
This PR re-writes the global state synchronization logic for a tighter coupling with the block synchronizer.

TODOs:
* keep global state synchronization progress when replacing the builder after a leap if the state root hash is the same
* mark the global state synchronization as complete when it is done as part of determining era validators from global state when historically syncing an immediate switch block after a switch block.
* Determine if/what metrics to add
* Global state sync logic is tested through the block synchronizer tests but it would be good to have unit tests for each of: `EraValidatorsAcquisition`, `GlobalStateAcquisition` and `TrieAcquisition`


alsrdn:
Closing this for now and will be reconsidered for next releases.

PR #3782, opened 2023-03-22
Fix accumulator purging of future acceptors

### LINKED ISSUE #3748: Node doesn't mark a forward block complete and is stuck when there's a delay caused by I/O
Node fails to progress due to the following chain of events:
high disk load -> even queue clog -> long block execution time -> premature purge of acceptors -> mark complete request ignored (not enough finality due to missing acceptor)
Current flow:
- contract runtime executed the block given by consensus and announces meta block
- meta block handler notifies other components (sends to deploy buffer, updates validator matrix, update meta block gossip state, sign the block, store and broadcast the signature) and kick-start the block accumulator
- block accumulator is getting sigs, but it relies on gossiping and broadcast only (no active fetch of signatures)
- if block accumulator can't get enough sigs (strict), block will never be marked complete (and node is stuck)

Proposed solution:
- consensus produces a finalized block
- we send it to block synchornizer (we could short circuit into HaveBlockHeader state)
- synchronizer is supposed to get all sigs, execute block and mark it complete
- currently this may get stuck (chicken & egg problem with first signature vs execution)
- need to change the flow to recognize if this is "our" block and execute it first and broadcast the sig

Alternative solution:
- contract runtime executed the block given by consensus and announces meta block
- meta block handler notifies other components (sends to deploy buffer, updates validator matrix), update meta block gossip state, sign the block, store and broadcast the signature and kick-start both the block accumulator and the block synchronizer (we can short circuit and go directly into HaveAllDeploys - should be ok even for empty block)
- meta block.executed should be set to true
- synchronizer actively fetches the missing signatures (until it has Strict) and then marks the block complete
- (at the same time) accumulator listens for broadcasted signatures (accumulator will be eventually purged)
- accumulator should no longer mark the blocks complete

Third proposed solution: keeping the whole thing as is with minor tweaks:

- when it's a consensus block, we have a flag in the accumulator for it to make the distinction
- we mark them complete as soon as we sign them (as it's consensus stuff so we have a high degree of confidence there)
- in the accumulator we send it with this special flag
- it sits in the accumulator for as long as it needs to gather the sigs
- when the required amount of sigs is there, we purge the acceptor (block is already complete)
- 

As a first part of this task we should create a test that reproduces the issue and then make it green by implementing the solution.

georgepisaltu:
Fixes #3748

### DIFF
--- a/node/src/components/block_accumulator.rs
+++ b/node/src/components/block_accumulator.rs
@@ -388,6 +388,10 @@ impl BlockAccumulator {
             }
         };
 
+        if sender.is_none() {
+            acceptor.set_our_signature(finality_signature.clone());
+        }
+
         debug!(%finality_signature, "registering finality signature");
         match acceptor.register_finality_signature(finality_signature, sender, self.validator_slots)
         {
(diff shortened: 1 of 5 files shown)


georgepisaltu:
bors try

rafal-ch (approved):
LGTM :+1:

Fraser999 (approved):

PR #3789, opened 2023-03-24
Fix latch issue in BlockSynchronizer

### LINKED ISSUE #3784: Latching issue in BlockSynchronizer
After fetching a sync leap due to needing era validators, the latch is not released causing progress to stall.  The fix is trivial, but we should probably create a test for this scenario.

Fraser999:
Fixes an issue where historical syncing would stall due to not releasing the in-flight latch after needing era validator weights.

Closes #3784.

### DIFF
--- a/node/src/components/block_synchronizer/tests.rs
+++ b/node/src/components/block_synchronizer/tests.rs
@@ -209,3 +209,76 @@ async fn global_state_sync_wont_stall_with_bad_peers() {
             .is_peer_unreliable(peer));
     }
 }
+
+#[tokio::test]
+async fn should_not_stall_after_registering_new_era_validator_weights() {
+    let mut rng = TestRng::new();
+    let mock_reactor = MockReactor::new();
+    let peer_count = 5;
+
+    // Set up an empty validator matrix.
+    let mut validator_matrix = ValidatorMatrix::new_with_validator(ALICE_SECRET_KEY.clone());
+
+    // Create a block synchronizer with a maximum of 5 simultaneous peers.
+    let mut block_synchronizer = BlockSynchronizer::new(
+        Config::default(),
+        Arc::new(Chainspec::random(&mut rng)),
+        peer_count,
+        validator_matrix.clone(),
+        prometheus::default_registry(),
+    )
+    .unwrap();
+
+    let peers: Vec<NodeId> = random_peers(&mut rng, peer_count as usize)
+        .iter()
+        .cloned()
+        .collect();
+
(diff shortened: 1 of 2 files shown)


node/src/components/block_synchronizer/tests.rs
```diff
@@ -209,3 +209,76 @@ async fn global_state_sync_wont_stall_with_bad_peers() {
...
+
+    block_synchronizer
+        .historical
+        .as_mut()
+        .expect("should have historical builder")
+        .register_era_validator_weights(&validator_matrix);
+
+    let mut effects = block_synchronizer.need_next(mock_reactor.effect_builder(), &mut rng);
+    assert_eq!(effects.len(), 1);
+    tokio::spawn(async move { effects.remove(0).await });
```
georgepisaltu, line 281: Interesting way to do the effects. From what I can tell, it's spawned but not necessarily finished. Should we wait for that to finish before doing the rest?

georgepisaltu (approved):

node/src/components/block_synchronizer/tests.rs
alsrdn: This is because the effect_builder creates both the event and also the future that awaits for the response. We're not interested in the actual response here because we use a simple reactor here that doesn't even implement the fetcher, storage etc and the response will never come. But we are interested in the event that the effect builder generates so that we can validate if it's the correct flow.

If we await it the test would block since we don't send the response on the responder.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -209,3 +209,76 @@ async fn global_state_sync_wont_stall_with_bad_peers() {
...
+    let event = mock_reactor.crank().await;
+    match event {
+        MockReactorEvent::FinalitySignatureFetcherRequest(_) => (),
+        _ => panic!("unexpected event: {:?}", event),
+    };
```
alsrdn: minor nit: we can use `assert_matches!()`
```suggestion
    assert_matches!(
        event,
        MockReactorEvent::FinalitySignatureFetcherRequest(FetcherRequest {
            id,
            peer,
            ..
        }) if peers.contains(&peer) && id.block_hash == *block.hash()
    );
```

alsrdn (approved):
LGTM👍

node/src/components/block_synchronizer/tests.rs
Fraser999: Done in [ed2fe10](https://github.com/casper-network/casper-node/pull/3789/commits/ed2fe106afc78723b15ad85cf2e249fc2116c778).

Issue #3791, opened 2023-03-24
Historical sync stalls when joiner nodes takes more than 6 eras to catch up to network


RitaMAllenCA:
If a joiner node takes a long time to catch up, it will sync leap repeatedly when it hears from other nodes that new blocks were added. Every time the node receives a leap response that contains a new era, this era will be registered to the validator matrix. The validator matrix is built to hold the validator weights for at most 6 eras (the oldest 3 and the newest 3 eras that were registered). Since the catch up process takes longer than 6 eras, when the 7th era is added, the median era is removed from the validator matrix. When the node eventually catches up and switches to KeepUp, it will start historical sync for the older blocks. When historical sync reaches a block that is in the era that was removed from the validator matrix, the node will try to fetch the era validators for that era through a SyncLeap. But because we only hold the oldest 3 and highest 3 eras in the validator matrix, this era will not be registered and the historical sync will be stuck doing the same process indefinitely.
The fix required a latch to be added when we remove the median era from the validator matrix when registering weights. We shouldn’t be removing weights from the validator matrix for eras that were not used to perform historical sync.

PR #3794, opened 2023-03-24
#3791 validator matrix retrograde latch

EdHastingsCasperAssociation:
https://github.com/casper-network/casper-node/issues/3791

@georgepisaltu will expand test fixtures.


node/src/types/validator_matrix.rs
```diff
@@ -122,17 +130,21 @@ impl ValidatorMatrix {
...
+        let median_era = guard
+            .keys()
+            .nth(MAX_VALIDATOR_MATRIX_ENTRIES / 2)
+            .copied()
+            .unwrap();
+        if median_era <= latch_era {
+            is_new
+        } else {
+            guard.remove(&median_era);
+            median_era != era_id
```
georgepisaltu: On this case we should probably keep removing entries in a loop until we reach `MAX_VALIDATOR_MATRIX_ENTRIES` entries.

node/src/types/validator_matrix.rs
```diff
@@ -122,17 +130,21 @@ impl ValidatorMatrix {
...
-                .keys()
-                .nth(MAX_VALIDATOR_MATRIX_ENTRIES / 2)
-                .copied()
-                .unwrap();
-            guard.remove(&median_key);
-            return median_key != era_id;
+        if guard.len() <= MAX_VALIDATOR_MATRIX_ENTRIES || self.retrograde_latch.is_none() {
+            return is_new;
+        }
+        let latch_era = self.retrograde_latch.unwrap_or_default(); // none checked above
```
georgepisaltu: Generally we try to avoid any unwraps where possible. To get rid of the `unwrap_or_default`, we can do:
```suggestion
        let latch_era = if let Some(era) = self.retrograde_latch.as_ref() {
            *era
        } else {
            return is_new;
        }
        if guard.len() <= MAX_VALIDATOR_MATRIX_ENTRIES {
            return is_new;
        }
```

Fraser999: While I don't agree that we should avoid `unwrap_or_default()` generally, I do slightly prefer the suggested change as it seems a little more straighforward to me.

Fraser999 (commented):
LGTM other than @georgepisaltu's point about removing median values repeatedly.

node/src/types/validator_matrix.rs
EdHastingsCasperAssociation: while it is true we avoid `unwrap` as much as possible, we use `unwrap_or_default` where safe quite a bit. 

as to the rest of it, I prefer early return and DRY; there's no reason to unpack the era if we're not at `MAX_VALIDATOR_MATRIX_ENTRIES` similarly, if the latch has not been set yet, there's no reason to do any further work...so get out. 

but, i'll make the change to move this along.

PR #3807, opened 2023-03-29
better error for 0 balances

jonas089:
The error "account does not exist at prestate_hash" is misleading to unexperienced Casper developers. This error is usually thrown when there is no transactions for an account at a given state root hash. This can either mean that the node is not synced, the account balance is actually 0 and there is no transaction record, or the srh is outdated.

In each scenario the situation becomes more clear with the new error and it'll help developers identify the most common issue - an unfunded account.

Therefore I propose this literal change.

### DIFF
--- a/node/src/components/deploy_acceptor.rs
+++ b/node/src/components/deploy_acceptor.rs
@@ -79,7 +79,7 @@ pub(crate) enum Error {
 #[derive(Clone, DataSize, Ord, PartialOrd, Eq, PartialEq, Hash, Debug, Error, Serialize)]
 pub(crate) enum DeployParameterFailure {
     /// Account does not exist.
-    #[error("account {account_hash} does not exist")]
+    #[error("main purse for account {account_hash} does not exist - balance is 0")]
     NonexistentAccount { account_hash: AccountHash },
     /// Nonexistent contract at hash.
     #[error("contract at {contract_hash} does not exist")]


RitaMAllenCA:
_In each scenario the situation becomes more clear with the new error and it'll help developers identify the most common issue - an unfunded account._
accounts do not exist UNTIL they are funded - so there is no such thing as an unfunded account.  Internally, existence of an account hash does not imply existence, but rather potential existence if/when it gets funded and the node receiving the query does not see that deploy yet.    See the account model documentation for more information.  @ACStoneCL please review the account documentation for externally meaningful errors.

RitaMAllenCA:
@jonas089 - @Fraser999 is taking on following up on updating the message to be more clear and we are actively looking into dev operator documentation to be sure we provide the best guidance.    Stay tuned and thanks for the input

jonas089:
@RitaMAllenCA thank you. I think it could help external developers to be more specific here, because some might feel discouraged when they see a common error like this that doesn't point them towards a solution. 

I tried to google "account does not exist at prestate hash casper" and the results wouldn't help understand the error so external people could struggle here.

Fraser999:
I opened #3833 to hopefully help a little with the issue you raised @jonas089.

I think that should help users determine what state the node is in (e.g. if it's lagging way behind the rest of the network).

For the case of an account existing, but having a balance below the minimum required (including if it has fallen to 0) there is a different error variant which will be returned: "insufficient balance in account with hash ...".

PR #3813, opened 2023-03-31
Fix detection of sync to ttl

rafal-ch:
In this PR
1. Node uses the lowest available block timestamp when calculating if it reached TTL when syncing backwards.
2. TTL detection does not rely on `MainReactor::switch_block_header` (which could be set to `None` when marking block complete), it uses the `switch_block_era_id_index` instead

That introduces some reads from storage when determining if we're synced to TTL, but the performance impact should be negligible.

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -271,6 +271,28 @@ pub(crate) enum HighestOrphanedBlockResult {
     MissingHeader(BlockHash),
 }
 
+impl Display for HighestOrphanedBlockResult {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            HighestOrphanedBlockResult::MissingHighestSequence => {
+                write!(f, "missing highest sequence")
+            }
+            HighestOrphanedBlockResult::MissingFromBlockHeightIndex(height) => {
+                write!(f, "height not found in block height index: {}", height)
+            }
+            HighestOrphanedBlockResult::Orphan(block_header) => write!(
+                f,
+                "orphan, height={}, hash={}",
+                block_header.height(),
+                block_header.height()
+            ),
+            HighestOrphanedBlockResult::MissingHeader(block_hash) => {
+                write!(f, "missing header for block hash: {}", block_hash)
+            }
+        }
+    }
+}
+
 impl<REv> Component<REv> for Storage
 where
     REv: From<FatalAnnouncement> + From<NetworkRequest<Message>> + Send,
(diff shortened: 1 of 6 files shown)


node/src/components/storage.rs
```diff
@@ -1283,6 +1283,13 @@ impl Storage {
             .map(|sequence| sequence.high())
     }
 
+    /// Retrieves the height of the lowest complete block (if any).
+    pub(crate) fn lowest_complete_block_height(&self) -> Option<u64> {
+        self.completed_blocks
+            .highest_sequence()
+            .map(|sequence| sequence.low())
+    }
```
georgepisaltu: This is misleading. The function actually returns the height of the lowest block in the highest completed blocks sequence. However, it seems that this is the behavior we want in the use case presented in the PR. I suggest we rename this to something else or simply use `storage::get_available_block_range`. Same applies for `read_lowest_complete_block_header`.

rafal-ch: The problem with `get_available_block_range()` is that when no blocks are available it returns `AvailableBlockRange::RANGE_0_0`. This is misleading as well and most likely that's why this function is private.

I think I can leverage the already existing `get_highest_orphaned_block_header()`, though.

rafal-ch: Updated here: https://github.com/casper-network/casper-node/pull/3813/commits/0b3421b2212b5e61d8de5593fa3b4b3dfc5800f3

node/src/reactor/main_reactor/keep_up.rs
```diff
@@ -597,7 +597,7 @@ impl MainReactor {
                 if block_header.is_genesis() {
                     return Ok(Some(SyncBackInstruction::GenesisSynced));
                 }
-                if self.sync_back_is_ttl() {
+                if self.sync_back_is_ttl()? {
```
alsrdn: nit: I think it would have been more readable to have this function named `synced_to_ttl()` or something similar and do the test for the config here. This way the function would have a single purpose and we could get rid of the config test inside the function. But this is a personal preference and not blocking.
```suggestion
                if !self.sync_to_genesis && self.synced_to_ttl()? {
```

alsrdn (approved):

node/src/reactor/main_reactor/keep_up.rs
rafal-ch: That's a good suggestion. Update here: b404e158c59208f64f6b5bec11df01758323f5bb

georgepisaltu (approved):

PR #3810, opened 2023-03-30
Add `BlockBuilder` tests and other minor improvements to `block_synchronizer`

georgepisaltu:
Fixes #3437 

This PR adds unit tests for the `block_synchronizer::block_builder` module. It also adds some minor improvements to the synchronizer code, detailed in the first commits.

PR #3816, opened 2023-04-03
Add tests for the block synchronizer forward sync flow

### LINKED ISSUE #3777: Block synchronizer testing - add tests for forward sync flow
Add tests to validate the forward sync flow for the block synchronizer

alsrdn:
Fixes: https://github.com/casper-network/casper-node/issues/3777


rafal-ch:
I'll wait with the review until the fix for `having_block_body_for_block_without_deploys_requires_only_signatures` lands.

types/src/crypto/asymmetric_key.rs
```diff
@@ -443,6 +443,21 @@ impl SecretKey {
...
+            rng.fill_bytes(&mut bytes[..]);
+            SecretKey::ed25519_from_bytes(bytes).unwrap()
+        } else {
+            let mut bytes = [0u8; SecretKey::SECP256K1_LENGTH];
+            rng.fill_bytes(&mut bytes[..]);
+            SecretKey::secp256k1_from_bytes(bytes).unwrap()
+        }
+    }
+}
+
```
Fraser999: I'd prefer if we didn't add this.  We've had to add some impls like this to support e.g. proptest tests, but I prefer to stick to the convention of providing a `random` function which takes a `TestRng` specifically, to try to avoid usage of other rngs which don't dump a seed on failure, making test failures harder to reproduce.

node/src/testing/test_block_builder.rs
```diff
@@ -0,0 +1,194 @@
...
+use rand::Rng;
+
+use crate::{
+    components::consensus::EraReport,
+    types::{Block, BlockHash, BlockPayload, Deploy, DeployHashWithApprovals, FinalizedBlock},
+};
+
+const MAX_ERA_FOR_RANDOM_BLOCK: u64 = 6;
+
+pub(crate) struct TestBlockBuilder {
```
Fraser999: I'd have probably put this in a `block_builder` module inside the `block` module, just to avoid this `testing` module becoming a dumping ground for all test builder helpers.  Not a big deal though.

node/src/testing/test_block_builder.rs
```diff
@@ -0,0 +1,194 @@
...
+    era: Option<u64>,
+    height: Option<u64>,
+    protocol_version: ProtocolVersion,
+    is_switch: Option<bool>,
+    deploys: Vec<Deploy>,
+    state_root_hash: Option<Digest>,
+    parent_hash: Option<BlockHash>,
+    timestamp: Option<Timestamp>,
```
Fraser999: Would be nice to order these in the same way as inside `Block` (and also for the functions below).

node/src/testing/test_block_builder.rs
```diff
@@ -0,0 +1,194 @@
...
+        self
+    }
+
+    #[allow(unused)]
+    pub(crate) fn timestamp(mut self, timestamp: Timestamp) -> Self {
+        self.timestamp = Some(timestamp);
+        self
+    }
+
+    pub(crate) fn build<R: Rng + ?Sized>(self, rng: &mut R) -> Block {
```
Fraser999: Again, I'd prefer if this took a `TestRng` rather than `R: Rng`.

Fraser999 (changes requested):
Initial review - I'll complete my review asap.

Blocking due to the use of `R:Rng`, but I also dislike having two ways of doing the same thing.  I think we should create a ticket to remove all the existing `Block::random_xxx` functions, using the new `TestBlockBuilder` instead, and to maybe leave only `Block::random`, but have it implemented using a `TestBlockBuilder`.

types/src/crypto/asymmetric_key.rs
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3816/commits/2b0eefcbeb36d4742f2b0fc35b4e36711d4a5f25

node/src/testing/test_block_builder.rs
alsrdn: Moved in https://github.com/casper-network/casper-node/pull/3816/commits/2b0eefcbeb36d4742f2b0fc35b4e36711d4a5f25

node/src/testing/test_block_builder.rs
alsrdn: Reordered in https://github.com/casper-network/casper-node/pull/3816/commits/2b0eefcbeb36d4742f2b0fc35b4e36711d4a5f25

node/src/testing/test_block_builder.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/pull/3816/commits/2b0eefcbeb36d4742f2b0fc35b4e36711d4a5f25

alsrdn:
> Initial review - I'll complete my review asap.
> 
> Blocking due to the use of `R:Rng`, but I also dislike having two ways of doing the same thing. I think we should create a ticket to remove all the existing `Block::random_xxx` functions, using the new `TestBlockBuilder` instead, and to maybe leave only `Block::random`, but have it implemented using a `TestBlockBuilder`.

I also dislike having two ways of doing things but I also dislike the amount of boilerplate and code duplication that we have today with `Block::random_with_*_and_*_*` type functions. It's a pain to extend that functionality when we want to add some other configurable parameter to use for block generation. I agree that we should create a new ticket to replace the other uses in the tests with this builder.

node/src/types/block.rs
```diff
@@ -2,6 +2,9 @@
 #![allow(clippy::field_reassign_with_default)]
 #![allow(clippy::boxed_local)] // We use boxed locals to pass on event data unchanged.
 
+#[cfg(any(feature = "testing", test))]
```
Fraser999: Should this just be `#[cfg(test)]` until we expose it publicly (or we could change the gating around the re-export in types.rs to be consistent with this).

node/src/components/block_synchronizer/tests.rs
```diff
@@ -336,3 +452,1231 @@ async fn should_not_stall_after_registering_new_era_validator_weights() {
...
+        events[1],
+        MockReactorEvent::BlockAccumulatorRequest(BlockAccumulatorRequest::GetPeersForBlock {
+            block_hash,
+            ..
+        }) if block_hash == *block.hash()
+    )
+}
+
+#[tokio::test]
+async fn fwd_sync_gets_peers_only_form_accumulator() {
```
Fraser999:
```suggestion
async fn fwd_sync_gets_peers_only_from_accumulator() {
```

node/src/components/block_synchronizer/tests.rs
```diff
@@ -95,35 +180,77 @@ fn check_sync_global_state_event(event: MockReactorEvent, block: &Block) {
...
+            secret_key.as_ref(),
+            PublicKey::from(secret_key.as_ref()),
+        );
+        assert!(signature.is_verified().is_ok());
+        assert!(builder.register_finality_signature(signature, None).is_ok());
+    }
+}
+
+impl BlockSynchronizer {
+    // Create an initialized block synchronizer with default config and maximum 5 simultaneous peers
```
rafal-ch: A nit (don't hardcode magic number) :)

```suggestion
    // Create an initialized block synchronizer with default config and maximum simultaneous peers
```

node/src/components/block_synchronizer/tests.rs
```diff
@@ -201,7 +330,7 @@ async fn global_state_sync_wont_stall_with_bad_peers() {
     let event = mock_reactor.crank().await;
 
     let second_peer_set = peers.iter().copied().choose_multiple(&mut rng, 4);
-    check_sync_global_state_event(event, &block);
+    check_sync_global_state_event(event, block);
 
     // Wait for the latch to reset
     std::thread::sleep(Duration::from_secs(6));
```
rafal-ch, line 336: I think I'll request a tech-debt ticket to add an option to inject the latch reset time. Sleeping for 6 seconds in a test is such a waste of time... Wdyt?

node/src/components/block_synchronizer/tests.rs
```diff
@@ -95,35 +180,77 @@ fn check_sync_global_state_event(event: MockReactorEvent, block: &Block) {
...
-    )
-    .unwrap();
-
-    // Generate more than 5 peers to see if the peer list changes after a global state sync error
-    let num_peers = rng.gen_range(10..20);
-    let peers: Vec<NodeId> = random_peers(&mut rng, num_peers).iter().cloned().collect();
+    let peers = test_env.peers();
+    let block = test_env.block();
+    let validator_matrix = test_env.gen_validator_matrix();
+    let validators_private_keys = test_env.validator_keys();
```
Fraser999: Can we rename "private keys" to "secret keys" throughout this file please?

(I'm overly nitpicky about this because there is a tendency to name variables `pk` for either public keys or private keys.  If we at least have a convention of favouring "secret key" over "public key" then such an abbreviation is a little less dangerous.)

Fraser999 (approved):
A couple of nitpicks, but otherwise LGTM.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -336,3 +452,1231 @@ async fn should_not_stall_after_registering_new_era_validator_weights() {
...
+#[test]
+fn duplicate_register_block_not_allowed_if_builder_is_not_failed() {
+    let mut rng = TestRng::new();
+    let test_env = TestEnv::random(&mut rng);
+    let block = test_env.block();
+    let validator_matrix = test_env.gen_validator_matrix();
+    let mut block_synchronizer = BlockSynchronizer::new_initialized(&mut rng, validator_matrix);
+
+    // Register block for forward sync
+    assert!(block_synchronizer.register_block_by_hash(*block.hash(), false, true));
```
rafal-ch, line 465: I know it's not a part of this PR, but can you extend [the comment](https://github.com/casper-network/casper-node/blob/cc95de3a80cdada1d67a84a2178f23bcf10e36ae/node/src/components/block_synchronizer.rs#L275) for `register_block_by_hash` and describe what is the meaning of the returned `bool`?

node/src/components/block_synchronizer/tests.rs
```diff
@@ -336,3 +452,1231 @@ async fn should_not_stall_after_registering_new_era_validator_weights() {
...
+        // once the latch is reset, we should get some effects
+        generated_effects.extend(effects);
+    }
+
+    // If the effects are empty at this point, then the synchronizer gets stuck
+    assert!(!generated_effects.is_empty());
+}
+
+#[tokio::test]
+async fn next_action_for_have_weak_finality_is_fetching_block_body() {
```
rafal-ch, line 937: Judging by just the name, this is very similar to [this guy](https://github.com/casper-network/casper-node/pull/3816/files#diff-ecca1f5c3a37ff0c7cb64100807c2e3074f1dfdd68582dff2dd768f7a313806bR726).

The latter one is more sophisticated, but please check if we don't duplicate some parts of the test unnecessarily.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -336,3 +452,1231 @@ async fn should_not_stall_after_registering_new_era_validator_weights() {
...
+        .is_ok());
+    fwd_builder.register_era_validator_weights(&block_synchronizer.validator_matrix);
+
+    // Register finality signatures to reach weak finality
+    register_multiple_signatures(
+        fwd_builder,
+        block,
+        validators_private_keys
+            .iter()
+            .take(validators_private_keys.len() / 3 + 1),
```
rafal-ch: Can you use `let num_sigs_required_for_weak_finality = validators_private_keys.len() / 3` here, as in other tests?

node/src/components/block_synchronizer/tests.rs
```diff
@@ -336,3 +452,1231 @@ async fn should_not_stall_after_registering_new_era_validator_weights() {
...
+    assert_matches!(
+        block_synchronizer.forward_builder().block_acquisition_state(),
+        BlockAcquisitionState::HaveBlock(acquired_block, _, _) if acquired_block.hash() == block.hash()
+    );
+}
+
+// TODO: this test fails sometimes depending on the value of the `LegacyRequiredFinality` chainspec
+// parameter. This happens due to an error in determining if signatures are enough to consider that
+// the block has Strict finality if a block is empty. Remove this ignore when the synchronizer
+// behavior is fixed.
```
rafal-ch, line 1083: Good catch, thanks! :+1:

rafal-ch (approved):
No blockers.

Just few nits, but good-to-go as is :+1:

node/src/components/block_synchronizer/tests.rs
alsrdn: We have [this PR](https://github.com/casper-network/casper-node/pull/3810) which makes the latch configurable. Once that's merged we can change the latch interval here so that we don't waste time.

node/src/components/block_synchronizer/tests.rs
alsrdn: You're right, these are similar but actually their target is different. This one puts the builder in `HaveWeakFinality` and checks if the next action is fetching the block body. The other test wants to check that fetching the block body only happens after weak finality was reached, and it goes through some steps of registering signatures to see if it doesn't transition too early. I'll rename the test that you referenced to better state its purpose.

node/src/components/block_synchronizer/tests.rs
alsrdn: Actually adding one extra sig is correct to account for the lossy integer division. I'll update the test that didn't have this criteria.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -69,14 +77,91 @@ impl MockReactor {
         let ((_ancestor, reactor_event), _) = self.scheduler.pop().await;
         reactor_event
     }
+
+    async fn process_effects(&self, mut effects: Effects<Event>) -> Vec<MockReactorEvent> {
+        let mut events = Vec::new();
+        for effect in effects.drain(0..) {
+            tokio::spawn(async move { effect.await });
```
georgepisaltu, line 84: Should we `await` the `JoinHandle` returned here so that we simulate a sequential execution of the effects in the queue? This would make the process predictable.

node/src/components/block_synchronizer/tests.rs
rafal-ch: Great :tada:

node/src/components/block_synchronizer/tests.rs
```diff
@@ -95,35 +180,77 @@ fn check_sync_global_state_event(event: MockReactorEvent, block: &Block) {
...
-
-    // Set up a validator matrix for the era in which our test block was created
-    let mut validator_matrix = ValidatorMatrix::new_with_validator(ALICE_SECRET_KEY.clone());
-    validator_matrix.register_validator_weights(
-        block.header().era_id(),
-        iter::once((ALICE_PUBLIC_KEY.clone(), 100.into())).collect(),
+    let test_env = TestEnv::random(&mut rng).with_block(
+        TestBlockBuilder::new()
+            .era(1)
+            .deploys([Deploy::random(&mut rng)].iter())
```
georgepisaltu: Would be useful to add a `random_deploys(count)` method on the builder (I recently used something similar too), but it's not really necessary to do it in this PR.

georgepisaltu (approved):

node/src/types/block.rs
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3816/commits/8c830a518622c78e4f57c74fa331707740e90648

node/src/components/block_synchronizer/tests.rs
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3816/commits/8c830a518622c78e4f57c74fa331707740e90648

node/src/components/block_synchronizer/tests.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/pull/3816/commits/8c830a518622c78e4f57c74fa331707740e90648

node/src/components/block_synchronizer/tests.rs
alsrdn: We actually can't await here in this format of the `MockReactor` since we don't mock the other node components (Storage, ContractRuntime etc). Effects created through the effect builder create a responder that listens for the result. We don't want to wait for the result to come (because there is no code to respond). We just want a shortcut to assert which event is generated if the other components would have responded.

node/src/components/block_synchronizer/tests.rs
alsrdn: Added here: https://github.com/casper-network/casper-node/pull/3816/commits/6404ee17f8ffc9e42e42ddf19f905b214791ef02

node/src/components/block_synchronizer/tests.rs
georgepisaltu: Ah, I remember our offline talk about this now. I understand, disregard what I wrote above.

PR #3817, opened 2023-04-03
Remove `unregister_metric!`

### LINKED ISSUE #3774: Remove unregister_metric macro, replace with Drop implementations on metrics wrapper
The unregister_metric requires a lot of manual laber (along with the manual registration). For a small memory cost (e.g. < 1 KB per 128 metrics total), we can leverage a wrapper type to perform registration and deregistration, obviating the need for most manual Drop impls, as well as calls to register, greatly minimizing code for metric implementations across all components.

marc-casperlabs:
Closes #3774.

This removes the `unregister_metric!` macro by creating a new `RegisteredMetric` wrapper type that registers a metric on construction and deregisters it on drop. A `RegistryExt` macro is added for convenience to create these and the wrapper has the used methods of all metrics implemented in specialized `impl` blocks for each type.

This greatly reduces the code needed for setting up metrics and reduces potential errors, at the cost of keeping additional copies of `Arc`s around in the form of `Registry`.

### DIFF
--- a/node/src/utils/registered_metric.rs
+++ b/node/src/utils/registered_metric.rs
@@ -2,7 +2,7 @@
 
 use prometheus::{
     core::{Atomic, Collector, GenericCounter, GenericGauge},
-    Counter, IntCounter, IntGauge, Registry,
+    Counter, Histogram, HistogramOpts, IntCounter, IntGauge, Registry,
 };
 
 /// A metric wrapper that will deregister the metric from a given registry on drop.
@@ -86,6 +86,14 @@ where
     }
 }
 
+impl RegisteredMetric<Histogram> {
+    /// Observes a given value.
+    #[inline]
+    pub(crate) fn observe(&self, v: f64) {
+        self.inner().observe(v)
+    }
+}
+
 impl<T> Drop for RegisteredMetric<T>
 where
     T: Collector + 'static,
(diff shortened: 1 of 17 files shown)


node/src/utils/registered_metric.rs
```diff
@@ -0,0 +1,204 @@
...
+            self.registry.unregister(boxed_metric).unwrap_or_else(|_| {
+                tracing::error!("unregistering {} failed: was not registered", desc)
+            })
+        }
+    }
+}
+
+/// Extension trait for [`Registry`] instances.
+pub(crate) trait RegistryExt {
+    /// Creates a new [`IntCounter`] registered to this registry.
```
rafal-ch:
```suggestion
    /// Creates a new [`Counter`] registered to this registry.
```

node/src/utils/registered_metric.rs
```diff
@@ -0,0 +1,204 @@
...
+        Ok(RegisteredMetric {
+            metric: Some(boxed_metric),
+            registry,
+        })
+    }
+
+    /// Returns a reference to the wrapped metric.
+    #[inline]
+    pub(crate) fn inner(&self) -> &T {
+        self.metric.as_ref().expect("metric disappeared")
```
rafal-ch, line 39: Maybe just log an error and allow the node to continue working with broken metrics, as they are not crucial to participation in the network.

marc-casperlabs: I can't create a `&T` out of thin air though :)

I could clone the box (in `Drop`, since that's the only reason this is an `Option`) and instead ignore deregistration failures, that'd remove the `.expect`, but this is the sort of thing that should never happen...

rafal-ch: Ok, I agree to leave as is :+1: 
However, "that should never happen" is my favorite last words :)

EdHastingsCasperAssociation (approved):

rafal-ch (approved):
LGTM :+1:

PR #3821, opened 2023-04-05
block_synchronizer: add a time threshold for detecting stalled builders

### LINKED ISSUE #3815: Add a stall threshold parameter to detect stalled progress on the block synchronizer
Because there are situations where the block synchronizer can get stuck, we want to add a new config parameter to compare against to see if the synchronizer didn’t make any progress. So if a builder did not make any progress for the configured amount of time, we flag it as stalled and the control logic should purge it and try again.

alsrdn:
Because there are situations where the block synchronizer can get stuck, added a new config parameter to compare against to check if the synchronizer didn’t make any progress.
So if a builder did not receive data successfully and updated its progress for the configured amount of time, we flag it as stalled and the control logic should purge it and try again.

Fixes: https://github.com/casper-network/casper-node/issues/3815

### DIFF
--- a/node/src/components/block_synchronizer/config.rs
+++ b/node/src/components/block_synchronizer/config.rs
@@ -9,6 +9,7 @@ const DEFAULT_MAX_PARALLEL_TRIE_FETCHES: u32 = 5000;
 const DEFAULT_PEER_REFRESH_INTERVAL: &str = "90sec";
 const DEFAULT_NEED_NEXT_INTERVAL: &str = "1sec";
 const DEFAULT_DISCONNECT_DISHONEST_PEERS_INTERVAL: &str = "10sec";
+const DEFAULT_PROGRESS_STALL_THRESHOLD_INTERVAL: &str = "120sec";
 
 /// Configuration options for fetching.
 #[derive(Copy, Clone, DataSize, Debug, Deserialize, Serialize)]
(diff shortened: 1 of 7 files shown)


rafal-ch:
~@alsrdn do we have a test anywhere that would show this solution works as expected?~

The test is here: https://github.com/casper-network/casper-node/pull/3816/files#diff-ecca1f5c3a37ff0c7cb64100807c2e3074f1dfdd68582dff2dd768f7a313806bR582

node/src/components/block_synchronizer/config.rs
```diff
@@ -21,6 +22,9 @@ pub struct Config {
     pub need_next_interval: TimeDiff,
     /// Time interval for recurring disconnection of dishonest peers.
     pub disconnect_dishonest_peers_interval: TimeDiff,
+    /// Time interval after which synchronization is considered stalled if no successful sync
+    /// activity happened.
+    pub progress_stall_threshold_interval: TimeDiff,
```
georgepisaltu, line 27: I would just call this `stall_threshold_interval`.

georgepisaltu (approved):
LGTM apart from one comment, but I don't want to block on it.

node/src/components/block_synchronizer/config.rs
Fraser999: Or even `stall_limit` since it's an upper bound on how long we'll allow a stall.

Fraser999 (approved):
No blockers from me, and I only have a weak opinion on the config variable's name.

rafal-ch (approved):
LGTM :+1:

node/src/components/block_synchronizer/config.rs
rafal-ch: This is covered in a separate PR: https://github.com/casper-network/casper-node/pull/3842

PR #3823, opened 2023-04-05
Add sending of huge deploys in soundness test to verify chunking

### LINKED ISSUE #3800: Extend the nightly test to also spam large deploys
Extend the nightly test added in https://github.com/casper-network/casper-node/issues/3728 to also spam the test network with large deploys (use the same bloat wasm which is used in casper-test and in sre). Start with 2 large deploys per minute. The goal is to exercise the trie chunking mechanism.

Next steps (will be requested as separate tickets):
- put the historical sync in the mix to assure it does not prevent the keepup/validate processes from progressing
- simulate protocol upgrade
- simulate emergency upgrade with changed validator set

rafal-ch:
This PR extends the "network soundness" test to also send huge deploys to the network in order to verify if the trie chunking mechanism kicks in.

It also increases the normal (smaller) deploys count sent periodically to put more pressure on the network.

Closes #3800


utils/nctl/sh/scenarios/network_soundness.py
```diff
@@ -139,10 +171,49 @@ def start_sending_deploys():
...
 def test_timer_thread(secs):
     sleep(secs)
-    log("*** " + str(secs) + " secs passed - finishing test ***")
+    log("*** " + str(secs) + " secs passed - running health checks ***")
+    run_health_checks()
+    log("*** test finished successfully ***")
     os._exit(0)
+
+
+def run_health_checks():
```
alsrdn, line 192: nit: maybe rename this to `check_for_chunk_indicator_in_logs`

utils/nctl/sh/scenarios/network_soundness.py
```diff
@@ -54,12 +60,20 @@ def log(msg):
...
-    result = subprocess.check_output(['/bin/bash', '-i', '-c',
-                                      command]).decode("utf-8").rstrip()
-    invoke_lock.release()
-    return result
+    try:
+        result = subprocess.check_output(['/bin/bash', '-i', '-c',
+                                          command]).decode("utf-8").rstrip()
+        return result
+    except subprocess.CalledProcessError as e:
+        log("command returned non-zero code - that is expected since the node is temporarily down"
```
alsrdn: nit: maybe re-phrase this log message.
```suggestion
        log("command returned non-zero exit code - this can be a transitory error if the node is temporarily down"
```

alsrdn (approved):

georgepisaltu (approved):
I didn't check the binary included in the PR, but approved nonetheless

utils/nctl/sh/scenarios/network_soundness.py
rafal-ch: Updated in feffb3b37f8311fb6a1ead1f586a48632d8322b8

utils/nctl/sh/scenarios/network_soundness.py
rafal-ch: I'd like to leave it as is for the sake of consistency with NCTL tests. Also, at some point, we might decide to run more checks than just this specific one.

rafal-ch:
> I didn't check the binary included in the PR, but approved nonetheless

Source code of this binary is available for inspection in the `casper-test` repo.

PR #3833, opened 2023-04-11
Update error messages of deploy acceptor

Fraser999:
This PR improves the `Display` implementations of many of the `DeployAcceptor`'s error variants, which is important as these are returned to clients as part of the JSON-RPC response in the case that putting a deploy fails.

In particular, the term "prestate hash" has been replaced with "state root hash" for clarity and to be consistent with how this is named elsewhere (e.g. in the block).  Also, for all the `InvalidDeployParameters` variants, the block's hash and height is also reported along with the relevant state root hash, to help users determine whether the node being queried is sufficiently caught up to the tip of the chain.

An example of the change is from:
```
deploy parameter failure: account 229826c53651b4314bcd8e713873b00749654c979c92a0d938d32f406bb9766b does not exist at prestate_hash: 169f..a5a7
```
to
```
account with hash 229826c53651b4314bcd8e713873b00749654c979c92a0d938d32f406bb9766b does not exist at state root hash 169f42d7cfdc1c6910adc57033276a1a8de8fe857144a58497eaaf10f412a5a7 of block d16aded7f240665b7b7c32f76ca64800bc8f04dbe7940fbe55e5159619d33d53 at height 170
```

Other than carrying the block header across various `Event` variants in place of the old `prestate_hash`, there should be no logical changes to the `DeployAcceptor`.

### DIFF
--- a/node/src/components/deploy_acceptor.rs
+++ b/node/src/components/deploy_acceptor.rs
@@ -35,8 +35,8 @@ use crate::{
     },
     fatal,
     types::{
-        chainspec::DeployConfig, BlockHeader, Chainspec, Deploy, DeployConfigurationFailure,
-        FinalizedApprovals,
+        chainspec::DeployConfig, BlockHash, BlockHeader, Chainspec, Deploy,
+        DeployConfigurationFailure, FinalizedApprovals,
     },
     utils::Source,
     NodeRng,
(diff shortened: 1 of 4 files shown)


node/src/components/deploy_acceptor.rs
```diff
@@ -760,12 +798,12 @@ impl DeployAcceptor {
                         }
                         None => {
                             debug!(?contract_version, "invalid contract at version");
```
alsrdn, line 800: Not really required but it would be nice to also update all `debug` logs to print out the error info since it contains much more information.

alsrdn (approved):
LGTM

node/src/components/deploy_acceptor.rs
Fraser999: As discussed offline, the errors don't actually give any additional info over the logs, so I'll leave this as is.

rafal-ch (approved):
LGTM :+1:

PR #3838, opened 2023-04-12
Move testing features into casper-types

darthsiroftardis:
CHANGELOG:

- Moved `Timestamp` and `TimeDiff` from casper-node into casper-types
- Moved the `node::crypto` module and moved the necessary functionality into casper-types
- Moved `testing::TestRng` out of `casper-node` and into `casper-types`


types/Cargo.toml
```diff
@@ -16,40 +16,58 @@ base64 = { version = "0.13.0", default-features = false }
...
+k256 = { version = "0.7.2", default-features = false, features = ["ecdsa", "sha256", "zeroize"] }
 num = { version = "0.4.0", default-features = false, features = ["alloc"] }
 num-derive = { version = "0.3.0", default-features = false }
 num-integer = { version = "0.1.42", default-features = false }
 num-rational = { version = "0.4.0", default-features = false }
 num-traits = { version = "0.2.10", default-features = false }
 once_cell = { version = "1.5.2", optional = true }
 proptest = { version = "1.0.0", optional = true }
+proptest-derive = { version = "0.3.0", optional = true }
+pem = { version = "0.8.1", optional = true }
```
Fraser999: Should be right above `proptest`.

types/Cargo.toml
```diff
@@ -16,40 +16,58 @@ base64 = { version = "0.13.0", default-features = false }
...
 uint = { version = "0.9.0", default-features = false }
+untrusted = { version = "0.7.1", optional = true }
 version-sync = { version = "0.9", optional = true }
 
 [dev-dependencies]
 bincode = "1.3.1"
 criterion = "0.3.5"
-getrandom = { version = "0.2.0", features = ["rdrand"] }
+derp = "0.0.14"
+getrandom = { version = "0.2.0" }
```
Fraser999:
```suggestion
getrandom = "0.2.0"
```

types/src/crypto.rs
```diff
@@ -3,10 +3,14 @@
 mod asymmetric_key;
 mod error;
 
+#[cfg(any(feature = "std", test))]
+pub use asymmetric_key::generate_ed25519_keypair;
 #[cfg(any(feature = "gens", test))]
```
Fraser999:
```suggestion
#[cfg(any(feature = "testing", feature = "gens", test))]
```

I think this change should be applied to every place in the `casper-types` crate where we feature gate behind "gens" - 8 instances I believe, including this one.

types/src/crypto/asymmetric_key.rs
Fraser999, line 1: There are a few changes in this file on `dev` which aren't here and probably should be, including a bugfix and a few places where visibility is `pub (super)` rather than `pub`.

I'd recommend just copying over the file from `dev` and removing the few places where we added `#[non_exhaustive]`

types/src/lib.rs
```diff
@@ -68,6 +77,8 @@ pub use deploy_info::DeployInfo;
 pub use execution_result::{
     ExecutionEffect, ExecutionResult, OpKind, Operation, Transform, TransformEntry,
 };
+#[cfg(any(feature = "std", test))]
+pub mod file_utils;
```
Fraser999: This should be up with the other `mod`s, immediately before `mod gas;`

Fraser999 (changes requested):
A few minor changes required, but it looks like asymmetric_key/tests.rs needs a heavy update.  Probably just copying that file over from `dev` will be enough though.

types/Cargo.toml
darthsiroftardis: Moved as requested

types/Cargo.toml
darthsiroftardis: Applied suggestion

types/src/crypto.rs
darthsiroftardis: Applied the correct feature flag and made the changes in the 7 other places

types/src/crypto/asymmetric_key.rs
darthsiroftardis: Copied the file over removing the non exhaustive tag

types/src/lib.rs
darthsiroftardis: Moved

darthsiroftardis:
> A few minor changes required, but it looks like asymmetric_key/tests.rs needs a heavy update. Probably just copying that file over from `dev` will be enough though.

Copied the file as requested

Fraser999 (approved):
LGTM

rafal-ch (dismissed):
Blocking temporarily,
We have important 1.5 PRs to be merged: [#3832](https://github.com/casper-network/casper-node/pull/3832) and [#3821](https://github.com/casper-network/casper-node/pull/3821)

I want to prioritize them, becuase @alsrdn is on holiday and if this PR creates merge conflicts it'll be tricky to resolve them.

darthsiroftardis:
>

PR #3841, opened 2023-04-13
3818: Test asserting the type and size of the largest message specimen

real-felix:
A test is added which:
- Verify that the largest specimen is of type `Message::GetResponse`,
- Verify that its size is `8_388_736` bytes.

### DIFF
--- a/node/src/components/network/message.rs
+++ b/node/src/components/network/message.rs
@@ -568,12 +568,15 @@ where
 }
 
 /// Creates a serialized specimen of the largest possible networking message.
-pub(crate) fn generate_largest_message<E: SizeEstimator>(
-    estimator: &E,
-    cache: &mut Cache,
-) -> Vec<u8> {
-    let specimen = Message::<protocol::Message>::largest_specimen(estimator, cache);
-    serialize_net_message(&specimen)
+pub(crate) fn generate_largest_message(chainspec: &Chainspec) -> Message<protocol::Message> {
+    let estimator = &NetworkMessageEstimator::new(chainspec);
+    let cache = &mut Cache::default();
+
+    Message::largest_specimen(estimator, cache)
+}
+
+pub(crate) fn generate_largest_serialized_message(chainspec: &Chainspec) -> Vec<u8> {
+    serialize_net_message(&generate_largest_message(chainspec))
 }
 
 impl<'a> SizeEstimator for NetworkMessageEstimator<'a> {
@@ -601,6 +604,7 @@ impl<'a> SizeEstimator for NetworkMessageEstimator<'a> {
 mod tests {
     use std::{net::SocketAddr, pin::Pin};
 
+    use assert_matches::assert_matches;
     use bytes::BytesMut;
     use casper_types::ProtocolVersion;
     use serde::{de::DeserializeOwned, Deserialize, Serialize};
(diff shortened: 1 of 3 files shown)


node/src/components/network/message.rs
```diff
@@ -941,4 +941,21 @@ mod tests {
...
+        let specimen = Message::largest_specimen(&estimator, &mut cache);
+
+        assert_matches::assert_matches!(
+            specimen,
+            Message::Payload(protocol::Message::GetResponse { .. })
+        );
+
+        let serialized = serialize_net_message(&specimen);
+
+        assert_eq!(serialized.len(), 8_388_736);
```
marc-casperlabs: I'd put a string explanation here, so that someone inadvertently changing this is informed, e.g.

```rust
assert_eq!(serialized.len(), 8_388_736, "the size of the largest possible network message based on the production chainspec has changed");
```

node/src/components/network/message.rs
```diff
@@ -941,4 +941,21 @@ mod tests {
...
     }
+
+    #[test]
+    fn assert_the_largest_specimen_type_and_size() {
+        let (chainspec, _) = crate::utils::Loadable::from_resources("production");
+        let estimator = NetworkMessageEstimator::new(&chainspec);
+        let mut cache = Cache::default();
+        let specimen = Message::largest_specimen(&estimator, &mut cache);
+
+        assert_matches::assert_matches!(
```
marc-casperlabs: I think it's safe to `use assert_matches::assert_matches` to make this a little more readable.

node/src/components/network/message.rs
```diff
@@ -941,4 +941,21 @@ mod tests {
...
+            specimen,
+            Message::Payload(protocol::Message::GetResponse { .. })
+        );
```
marc-casperlabs, line 958: Same as below. I think a message explanation would be good, since we expect this to hit us in the far future, when we have forgotten all about this.

node/src/components/network/message.rs
```diff
@@ -941,4 +941,21 @@ mod tests {
     fn bincode_roundtrip_certificate() {
         roundtrip_certificate(false)
     }
+
+    #[test]
+    fn assert_the_largest_specimen_type_and_size() {
+        let (chainspec, _) = crate::utils::Loadable::from_resources("production");
+        let estimator = NetworkMessageEstimator::new(&chainspec);
```
marc-casperlabs: This code and the lines below are almost copies of the code inside the chainspec estimation. I am a little concerned that this might make it too easy to change one, but not the other in the future. What do you think about factoring out this functionality (determining the largest specimen from a chainspec), maybe into a chainspec method or something standalone, and then using it in both places?

marc-casperlabs (commented):
We could merge this as is, but I want to make some comments :)

node/src/components/network/message.rs
rafal-ch: Yes, true. Also it might be worth making a `const` out of `8_388_736` and give it a comment like "this is the size of the largest recorded specimen in release 1.5".

real-felix: Done

node/src/components/network/message.rs
real-felix: Done

node/src/components/network/message.rs
real-felix: Done

node/src/components/network/message.rs
real-felix: I refactored the helper function so that it can be used by both.

EdHastingsCasperAssociation (approved):

rafal-ch:
> test components::network::message::tests::assert_the_largest_specimen_type_and_size ... ok
> ...finished in 23.50s

Didn't expect this call to take so long - that's not optimal for UT, but let's leave it as is for now.

rafal-ch (approved):
LGTM :+1:

marc-casperlabs (approved):
Nice work!

PR #3843, opened 2023-04-14
Add random seed info to `dont_upgrade_without_switch_block` test

rafal-ch:
This PR adds an output of `CL_TEST_SEED` to the `dont_upgrade_without_switch_block` which occasionally gets stuck, as can be seen here: https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/7995/1/3.

The suspicion is that it happens due to the particular randomizer seed.

Unfortunately, the test don't fail per-se, instead it is killed by drone after a timeout, hence the `CL_TEST_SEED` is not displayed through the `TestRng::drop()`.

This PR should be reverted after the problematic seed is detected and the test is fixed.

### DIFF
--- a/types/src/testing.rs
+++ b/types/src/testing.rs
@@ -83,6 +83,11 @@ impl TestRng {
             *flag.borrow_mut() = true;
         });
     }
+
+    /// Returns the current `seed`.
+    pub fn seed(&self) -> Seed {
+        self.seed
+    }
 }
 
 impl Default for TestRng {
--- a/types/src/testing.rs
+++ b/types/src/testing.rs
@@ -83,11 +83,6 @@ impl TestRng {
             *flag.borrow_mut() = true;
         });
     }
-
-    /// Returns the current `seed`.
-    pub fn seed(&self) -> Seed {
-        self.seed
-    }
 }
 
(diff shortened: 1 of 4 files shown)


sacherjj (approved):
The .drone.yml changes look good I believe. Drone will let us know for sure.  :)

types/src/testing.rs
```diff
@@ -83,6 +83,11 @@ impl TestRng {
             *flag.borrow_mut() = true;
         });
     }
+
+    /// Returns the current `seed`.
+    pub fn seed(&self) -> Seed {
+        self.seed
+    }
```
Fraser999: Not a blocker, but I don't think we need this getter - the `Display` and `Debug` impls of `TestRng` already print the seed.

rafal-ch: Yes, good catch, it works indeed:
```
Running 'dont_upgrade_without_switch_block' test with rng=TestRng seed: fab95057ffe9475529432a90bbab216a
```

Fixed in 48e994aa86bd00c201281bc08a19fb3f4c22f642

georgepisaltu (approved):

PR #3855, opened 2023-04-17
Unify sync to ttl check

### LINKED ISSUE #3858: Unify the "is synced to ttl" calculation
Currently, there are at least two places where the node calculates the "is synced to ttl". One is DeployBuffer::have_full_ttl_of_deploys which uses chain_index cache, other one is MainReactor::synced_to_ttl and relies solely on the data in storage.

rafal-ch:
This PR makes the validate flow to reuse the "is synced to ttl?" calculation from keep-up flow, making the `DeployBuffer::chain_index` no longer needed.

Closes https://github.com/casper-network/casper-node/issues/3858

### DIFF
--- a/node/src/reactor/main_reactor/event.rs
+++ b/node/src/reactor/main_reactor/event.rs
@@ -311,7 +311,7 @@ impl ReactorEvent for MainEvent {
             MainEvent::ChainspecRawBytesRequest(_) => "ChainspecRawBytesRequest",
             MainEvent::UpgradeWatcherRequest(_) => "UpgradeWatcherRequest",
             MainEvent::StorageRequest(_) => "StorageRequest",
-            MainEvent::BlockCompleteConfirmationRequest(_) => "MarkBlockCompletedRequest",
+            MainEvent::BlockCompleteConfirmationRequest(_) => "BlockCompleteConfirmationRequest",
             MainEvent::DumpConsensusStateRequest(_) => "DumpConsensusStateRequest",
             MainEvent::ControlAnnouncement(_) => "ControlAnnouncement",
             MainEvent::FatalAnnouncement(_) => "FatalAnnouncement",
--- a/node/src/reactor/main_reactor/keep_up.rs
+++ b/node/src/reactor/main_reactor/keep_up.rs
@@ -21,7 +21,9 @@ use crate::{
         requests::BlockSynchronizerRequest, EffectBuilder, EffectExt, EffectResultExt, Effects,
     },
     reactor::main_reactor::{MainEvent, MainReactor},
-    types::{ActivationPoint, BlockHash, GlobalStatesMetadata, SyncLeap, SyncLeapIdentifier},
+    types::{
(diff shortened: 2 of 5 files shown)


rafal-ch:
bors try

node/src/reactor/main_reactor/event.rs
```diff
@@ -311,7 +311,7 @@ impl ReactorEvent for MainEvent {
             MainEvent::ChainspecRawBytesRequest(_) => "ChainspecRawBytesRequest",
             MainEvent::UpgradeWatcherRequest(_) => "UpgradeWatcherRequest",
             MainEvent::StorageRequest(_) => "StorageRequest",
-            MainEvent::BlockCompleteConfirmationRequest(_) => "MarkBlockCompletedRequest",
+            MainEvent::BlockCompleteConfirmationRequest(_) => "BlockCompleteConfirmationRequest",
```
georgepisaltu, line 314: Although it's not in the scope of this PR, I don't like the name discrepancy here. `BlockCompleteConfirmationRequest` and `MarkBlockCompletedRequest` seem to be used interchangeably in the codebase, and they don't seem to mean the same to me as far as naming goes. I checked and it seems it's the storage request for marking blocks complete, but `BlockCompleteConfirmationRequest` sounds like you'd ask storage if this block is complete. Again, not that relevant to this PR, just putting this out there.

node/src/reactor/main_reactor/keep_up.rs
```diff
@@ -681,34 +692,60 @@ impl MainReactor {
...
+            } else {
+                // No latest switch block header is known, we assume that we're not synced to
+                // TTL.
+                info!(
+                    "KeepUp: unable to get latest switch block header (this problem should be transient), \
+                    assuming not synced to ttl");
+                return Ok(false);
+            }
+        }
+    };
```
georgepisaltu, line 722: I would like to see this logic extracted to either in the outer calling function or in a separate function. Same for the other block we should have. The reason is that I think we shouldn't be passing storage around like that and it would be more modular, but it's ultimately up to you. In case we can't get either the switch block or orphaned block, we can exit early in the calling functions.

georgepisaltu (approved):
LGTM, just a couple of comments, but not blockers.

node/src/reactor/main_reactor/event.rs
Fraser999: I agree

Fraser999 (approved):

node/src/reactor/main_reactor/event.rs
rafal-ch: It'll be addressed here: https://github.com/casper-network/casper-node/pull/3868

node/src/reactor/main_reactor/keep_up.rs
rafal-ch: It'll be addressed here: https://github.com/casper-network/casper-node/pull/3868

Issue #3768, opened 2023-03-17
add a test that ensures CPU requirement does not change


RitaMAllenCA:
once we have an analysis tool selected, add a test to ensure the requirements do not change

PR #3859, opened 2023-04-17
Perform CPU instruction set analysis during nightly tests

### LINKED ISSUE #3767: analyze the binaries we ship using a suitable tool to detect the instruction set
analyze the binaries we ship using a suitable tool to detect the instruction set.  Select a "suitable tool" and run it on our binaries. Should this be run on each cadenced release or just major release?

marc-casperlabs:
Closes #3767, #3768.

Adds a new tool (installed through cargo) to the nightly env, [elfx86exts](https://github.com/pkgw/elfx86exts), which is called through scripts in nightly tests, comparing the used CPU instructions against a fixed set of known instructions from the 1.4.13 release. This is intended to prevent accidental "instruction set feature creep", ensuring (to follow) documentation can be kept accurate.


bradjohnl:
@marc-casperlabs Does this check need to be periodically modified to compare extensions in newer releases? 
Or are we going to add extensions if we notice the check is failing because there is a diff to make sure that we document them?

Edit: or both? :smile: anyway, looks good to me

bradjohnl (approved):

rafal-ch (approved):
LGTM :+1:

marc-casperlabs:
> Does this check need to be periodically modified to compare extensions in newer releases?
Or are we going to add extensions if we notice the check is failing because there is a diff to make sure that we document them?

It should be a rare occurence that we need to require additional extensions. This is mainly to prevent them from being introduced accidentally, and we will document them as a requirement.

bradjohnl:
bors try

bradjohnl:
bors try

marc-casperlabs:
bors merge

PR #3861, opened 2023-04-17
Version bumps

Fraser999:
This PR provides version bumps in preparation for the v1.4.14 node release.

It also sets the `prune_batch_size` to 50 in the two chainspec files and bumps the execution-engine dependency `casper-wasm-utils` to 1.0.0.

Finally, there is a small change to the casper-updater tool to avoid it from calling `cargo generate-lockfile`.  This would cause all dependencies to be updated in the Cargo.lock file, not just the casper crates.  Now the tool calls `cargo update`, specifying each casper crate individually and hence avoiding the mass dependency update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -371,7 +371,7 @@ dependencies = [
 
 [[package]]
 name = "casper-engine-test-support"
-version = "3.0.0"
+version = "3.1.0"
 dependencies = [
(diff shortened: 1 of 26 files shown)


sacherjj (approved):

rafal-ch (approved):
LGTM :+1:

PR #3865, opened 2023-04-19
Reintroduce max_delegators_per_validator

darthsiroftardis:
CHANGELOG:

- Added a new optional chainspec setting called `max_delegators_per_validator` which enforces a delegator cap per validator
- Reworked the redelegation implementation to transfer redelegated tokens back to the Account's main purse
- Added tests to assert the newly modified behavior

### DIFF
--- a/execution_engine/src/system/auction/detail.rs
+++ b/execution_engine/src/system/auction/detail.rs
@@ -503,15 +503,7 @@ fn is_under_max_delegator_cap(
     max_delegators_per_validator: Option<u32>,
     new_validator_delegator_len: usize,
 ) -> bool {
-    if max_delegators_per_validator.is_none() {
-        return true;
-    }
-
-    if let Some(max_delegators) = max_delegators_per_validator {
-        if new_validator_delegator_len >= max_delegators as usize {
-            return false;
-        }
-    }
-
-    true
+    max_delegators_per_validator
+        .map(|limit| new_validator_delegator_len < limit as usize)
+        .unwrap_or(true)
 }
--- a/execution_engine/src/system/auction/detail.rs
+++ b/execution_engine/src/system/auction/detail.rs
@@ -192,6 +192,7 @@ where
 /// This function can be called by the system only.
 pub(crate) fn process_unbond_requests<P: Auction + ?Sized>(
     provider: &mut P,
+    max_delegators_per_validator: Option<u32>,
 ) -> Result<(), ApiError> {
     if provider.get_caller() != PublicKey::System.to_account_hash() {
         return Err(Error::InvalidCaller.into());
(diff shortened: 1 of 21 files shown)


node/src/types/chainspec/core_config.rs
```diff
@@ -84,6 +84,9 @@ pub struct CoreConfig {
 
     /// Which consensus protocol to use.
     pub consensus_protocol: ConsensusProtocolName,
+
+    /// The maximum amount of delegators per validator.
+    pub max_delegators_per_validator: Option<u32>,
```
Fraser999: I don't like having optional fields in the chainspec, since it's too easy to forget to update the chainspec.toml (as has happened in this PR).

I'd suggest making this non-optional here, and treating a value of `0` as `None` when converting to the corresponding engine_config field.

Either way - if we decide against that suggestion, I strongly think we should add entries to `resources/production/chainspec.toml` and `resources/local/chainspec.toml.in` along with a note about the default value if not provided (in the case we stick to leaving this optional) or a note about setting it to `0` meaning it's disabled.

execution_engine/src/system/auction/detail.rs
```diff
@@ -483,3 +498,20 @@ pub(crate) fn era_validators_from_snapshot(
...
+        return true;
+    }
+
+    if let Some(max_delegators) = max_delegators_per_validator {
+        if new_validator_delegator_len >= max_delegators as usize {
+            return false;
+        }
+    }
+
+    true
```
Fraser999: A little more concise:
```suggestion
    max_delegators_per_validator
        .map(|limit| new_validator_delegator_len < limit as usize)
        .unwrap_or(true)
```

Fraser999 (commented):
Mostly LGTM, but I made a couple of suggestions.

We should also update the CHANGELOG files:
* EE should have `Lift the temporary limit of the global maximum delegator capacity.` removed
* node should have an entry about the new chainspec setting - I'd suggest right after the current entry ``Add a `consensus_protocol` option to the chainspec...``

node/src/types/chainspec/core_config.rs
darthsiroftardis: I like the idea of having the zero value corresponding to `None`, I will make that change

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -4129,3 +4130,325 @@ fn should_allow_delegations_with_minimal_floor_amount() {
 
     builder.exec(delegation_request_2).expect_success().commit();
 }
+
+#[ignore]
+#[test]
+fn should_enforce_max_delegators_per_validator_cap() {
```
rafal-ch, line 4136: Can you extend this test a little?

What I think of is to test that if a validator is fully saturated with delegators, we can remove one of the delegators and be able to add another one in it's place.

node/src/types/chainspec/core_config.rs
darthsiroftardis: Added the change as requested

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
darthsiroftardis: Extended test

execution_engine/src/system/auction/detail.rs
darthsiroftardis: Applied suggestion

Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

Issue #3827, opened 2023-04-07
Change padding bytes for Key variants


RitaMAllenCA:
Blocked by: 1.4.14 getting merged to dev

We should remove CHAINSPEC_REGISTRY_KEY_BYTES and CHECKSUM_REGISTRY_KEY_BYTES and replace their usage by the new padding const provided in 1.4.14

PR #3871, opened 2023-04-21
Merge `release-1.4.14` into `dev` and disable pruning

### LINKED ISSUE #3752: Merge 1.4.14 back to dev before 1.5
Once 1.4.14 is activated on mainnet, and there’s already a git tag on the main repo. We should merge back 1.4.14 changes to the dev branch before releasing 1.5. It should be possible to do it by cherry-picking individual PRs that went to 1.4.14 to avoid amount of merge conflicts, but we should be careful to not miss any changes.

Fraser999:
This PR brings the code over from `release-1.4.14`.  There were a large number of nontrivial conflicts, so thorough reviews would be appreciated.

A possible point of confusion is that both branches had previously introduced a new nctl upgrade test, both called `upgrade_scenario_11.sh`.  To resolve this, I kept the one from `release-1.4.14` as scenario 11 and renamed the `dev` one to scenario 12.  Both of these two files' contents are unchanged despite the hefty diff showing for them.

Similarly, the `Key` enum had grown different additional variants on each branch, and this was resolved by inserting the ones from `release-1.4.14` before the ones from `dev`.  I also took the opportunity to just change the padding bytes for new variants with no wrapped value to use `0000...` rather than `1111...` and `2222...`.

Given that pruning exactly as implemented in 1.4.14 is not viable due to fast-sync, it has been disabled in the code and a small amount of supporting code removed (mainly in storage and contract runtime).  This was all done in [a standalone commit](https://github.com/casper-network/casper-node/commit/c84c10add001ed14c0b2b07050cc43d3e480e97f).

I'm opening the PR as a draft since it's currently not yet been decided whether pruning should be disabled in this way, or whether the pruning code should be more aggressively pruned, or whether we should fix it and leave it enabled in the code (but still able to be made a no-op via the chainspec setting).

Closes #3752 and #3827.


node/CHANGELOG.md
```diff
@@ -93,11 +93,10 @@ All notable changes to this project will be documented in this file.  The format
 * Remove legacy synchronization from genesis in favor of fast-sync.
 * Remove config options no longer required due to fast-sync: `[linear_chain_sync]`, `[block_proposer]` and `[consensus.highway.standstill_timeout]`.
 * Remove chainspec setting `[protocol.last_emergency_restart]` as fast sync will use the global state directly for recognizing such restarts instead.
+* Remove chainspec setting `[core.prune_batch_size]` as `Key::EraInfo` pruning has been completed and is no longer supported in this version.
```
mpapierski, line 96: See later comment - Can we keep this chainspec setting, but set it to zero to disable this feature? This way we don't need to reintroduce it, then remove it again in each version. There could also be releases where we'd always prune some entries.

node/src/components/block_validator.rs
```diff
@@ -90,7 +88,7 @@ pub(crate) enum Event {
     #[display(fmt = "{} found", dt_hash)]
     DeployFound {
         dt_hash: DeployOrTransferHash,
-        deploy_info: Box<DeployFootprint>,
+        deploy_footprint: Box<DeployFootprint>,
```
mpapierski, line 91: This change seems suspicious at first, as release-1.4.14 branch, and dev both still name it `deploy_info`

node/src/components/block_validator.rs
```diff
@@ -109,33 +107,14 @@ pub(crate) enum Event {
...
     /// A list of responders that are awaiting an answer.
     responders: SmallVec<[Responder<bool>; 2]>,
-    /// Peers that should have the data.
-    sources: VecDeque<NodeId>,
 }
 
 impl BlockValidationState {
-    /// Adds alternative source of data.
-    /// Returns true if we already know about the peer.
-    fn add_source(&mut self, peer: NodeId) -> bool {
```
mpapierski, line 125: This method is present on dev, but not on release-1.4.14. Was this removal intentional?

node/src/components/block_validator.rs
```diff
@@ -231,47 +210,37 @@ where
                     return responder.respond(false).ignore();
                 }
 
-                match self.validation_states.entry(block) {
```
mpapierski, line 234: Present on dev, not on release-1.4.14 - probably the whole file was accidentally mismerged? Unless that's what we want

node/src/components/block_validator/tests.rs
```diff
@@ -53,38 +57,43 @@ impl MockReactor {
...
-        if let ReactorEvent::Fetcher(FetcherRequest {
-            id,
-            peer,
-            validation_metadata: _,
-            responder,
-        }) = reactor_event
-        {
-            match deploy.into() {
-                None => {
+    async fn expect_fetch_deploys(
```
mpapierski, line 60: This method is present on release-1.4.14, not in dev, is that an intentional change?

node/src/components/contract_runtime/operations.rs
```diff
@@ -191,6 +242,90 @@ pub fn execute_finalized_block(
...
+    //
+    // Currently disabled as `key_block_height_for_activation_point` might not be available on a
+    // freshly-joined node at the time this call to execute block happens.
+    if false {
+        if let Some(previous_block_height) = finalized_block.height().checked_sub(1) {
+            // The following three values should be passed into this function if pruning is
+            // re-enabled. For now, they're set so that `calculate_prune_eras` will panic.
+            let activation_point_era_id = EraId::default();
+            let key_block_height_for_activation_point = u64::MAX;
+            let prune_batch_size = 1;
```
mpapierski, line 256: Do we want to remove this from chainspec? I thought you could set it to zero when we don't need to prune anything in a given protocol version (see also changelog comment)

node/src/components/block_validator.rs
mpapierski: Removal introduced in #3750 and we need to carry this bugfix onto dev

Fraser999:
Closed in favour of #3889.

PR #3878, opened 2023-04-24
block_synchronizer/tests: enable sync stall tests

alsrdn:
Because we now have the ability to determine block synchronizer stalls, re-enable the tests that check for stalls.

Also now that the latch interval is configurable, we can adjust the tests that require to wait for the latch to be reset such that the test run takes a reasonable amount of time.


georgepisaltu (approved):

rafal-ch (approved):
LGTM :+1:

PR #3879, opened 2023-04-24
Match 1.5 to 1.6 metrics

### LINKED ISSUE #3773: Ensure network metrics in 1.6 match 1.5
Description: The networking metrics have changed in 1.6 due to the underlying shift of the networking model in 1.6. For backwards compatibility reasons, we need to ensure that the 1.5 are still present, but at a fixed (i.e. 0) value to not break clients that assume they are there.
These metrics should be marked as deprecated and removed in 2.0

marc-casperlabs:
Closes #3773 

Adds a test that will query the metrics endpoint, extract the name of all metrics and check if they match a snapshot of metrics previously found in 1.5, ensuring the latter is a subset.

The more interesting part is that this a test of an external HTTP interface without the need of any external tool like `nctl` or `casper-test`.

To make this possible, the component state workflow has been cleaned up slightly for the REST server, only setting the ready state once the external port is bound and up. Additionally, a macro allows for easily waiting on component initialization, and port retrieval is added. The final piece of the puzzle is the ability to crank a server instance in unit tests in the background, while querying it with requests in the foreground.


node/src/reactor/main_reactor/tests.rs
```diff
@@ -908,3 +914,77 @@ async fn empty_block_validation_regression() {
         inactive => panic!("unexpected inactive validators: {:?}", inactive),
     }
 }
+
+/// Waits until all node have at least initialized the given component.
+///
+/// Expects the ident of a
+macro_rules! wait_for_component_initialization {
```
Fraser999: I'd prefer a normal function for this.  I expect we can use a `component_name: &str` parameter and match on the component's `name()`.  Might need a helper in `MainReactor` like `pub(crate) fn is_initialized(&self, component_name: &str) -> bool`.

I don't even mind if we only support checking for the rest server for now - further support can be added on a case-by-case basis.

If you do decide to stick with the macro, then I think the doc comments need a little attention :)  Quite the cliffhanger!

marc-casperlabs: All good, removed the macro in 6dfd53572d9dec9be81024bda159b641d9e83d2d!

Fraser999 (approved):

PR #3882, opened 2023-04-25
NCTL: Allow filtering certain error messages when doing health checks

### LINKED ISSUE #3804: Get rid of errors=ignore in upgrade_after_emergency_upgrade_test_pre_1.5.sh
We should be filtering out indeterministic errors instead of ignoring all the errors, see https://github.com/casper-network/casper-node/pull/3779#discussion_r1146260979
In order to catch other errors we could add a parameter to the health check to ignore errors matching a particular pattern. Something like: errors='ignore_pattern:could not send response.*to request down oneshot channel' and then add a grep -v $IGNORE_PATTERN to assert_error_count in the health checks.

alsrdn:
Fixes: https://github.com/casper-network/casper-node/issues/3804

Also fixes a bug in the function that waits for the node to do a full historical sync.

### DIFF
--- a/utils/nctl/sh/scenarios/common/health_checks.sh
+++ b/utils/nctl/sh/scenarios/common/health_checks.sh
@@ -41,12 +41,35 @@ function assert_error_count() {
     local COUNT
     local TOTAL
     local UNIQ_ERRORS
+    local IGNORE_PATTERN
+    local IGNORE_ALL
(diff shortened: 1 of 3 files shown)


alsrdn:
bors try

Fraser999 (approved):

fizyk20 (approved):

PR #3889, opened 2023-04-26
Merge `release-1.4.14` into `dev` and rework pruning slightly

### LINKED ISSUE #3752: Merge 1.4.14 back to dev before 1.5

Fraser999:
This PR brings the code over from `release-1.4.14`.  There were a large number of nontrivial conflicts, so thorough reviews would be appreciated.

A possible point of confusion is that both branches had previously introduced a new nctl upgrade test, both called `upgrade_scenario_11.sh`.  To resolve this, I kept the one from `release-1.4.14` as scenario 11 and renamed the `dev` one to scenario 12.  Both of these two files' contents are unchanged despite the hefty diff showing for them.

Similarly, the `Key` enum had grown different additional variants on each branch, and this was resolved by inserting the ones from `release-1.4.14` before the ones from `dev`.  I also took the opportunity to just change the padding bytes for new variants with no wrapped value to use `0000...` rather than `1111...` and `2222...`.

Pruning has been left enabled, although given that the `ChainspecLoader` has been removed in `dev`, some of the flows had to change slightly to support this.  Specifically, `Storage` is now made aware of the activation point, and hence only needs to cache the single key block height rather than the previous cache of a collection of heights.  Despite the pruning process being broken in the case of a joining node not knowing the height of the last block of the previous version, it's expected that such a node will eventually rectify its state, by repeated failures to execute and hence repeated sync leaps.
This was all done in [a standalone commit](https://github.com/casper-network/casper-node/commit/012230cc0b2c001adf6c3575e014fc099329d95f).

Closes #3752 and #3827.


resources/production/chainspec.toml
```diff
@@ -70,6 +70,8 @@ max_associated_keys = 100
 max_runtime_call_stack_height = 12
 # Minimum allowed delegation amount in motes
 minimum_delegation_amount = 500_000_000_000
+# Global state prune batch size (0 = this feature is off)
+prune_batch_size = 50
```
mpapierski: Don't we want to reset it to zero in dev already? So pruning is effectively doing nothing

mpapierski (approved):

resources/production/chainspec.toml
Fraser999: Good point.  Done in [cd55241](https://github.com/casper-network/casper-node/pull/3889/commits/cd552410b6f0c305ef75bb82cfdc6eacfc0ab3fd).

rafal-ch (approved):
LGTM :+1:

dwerner (approved):

dwerner (approved):
Latest changes LGTM

PR #3895, opened 2023-04-27
Add tests for the historical sync

### LINKED ISSUE #3778: Block synchronizer testing - add tests for historical sync flow
Add tests to validate the historical sync flow for the block synchronizer

real-felix:
Fixes #3778

### DIFF
--- a/node/src/components/block_synchronizer/global_state_synchronizer.rs
+++ b/node/src/components/block_synchronizer/global_state_synchronizer.rs
@@ -38,6 +38,11 @@ const COMPONENT_NAME: &str = "global_state_synchronizer";
 pub(crate) struct RootHash(Digest);
 
 impl RootHash {
+    #[cfg(test)]
+    pub(crate) fn new(digest: Digest) -> Self {
+        Self(digest)
+    }
+
     pub(crate) fn into_inner(self) -> Digest {
         self.0
     }
--- a/node/src/components/block_synchronizer/tests.rs
+++ b/node/src/components/block_synchronizer/tests.rs
@@ -10,7 +10,7 @@ use std::{
 };
 
 use casper_types::{
-    AccessRights, CLValue, EraId, PublicKey, SecretKey, StoredValue, TimeDiff, URef, U512,
+    AccessRights, CLValue, EraId, Key, PublicKey, SecretKey, StoredValue, TimeDiff, URef, U512,
 };
 use derive_more::From;
 use rand::{seq::IteratorRandom, Rng};
@@ -23,9 +23,13 @@ use crate::{
         block_synchronizer::block_acquisition::BlockAcquisitionState,
         consensus::tests::utils::{ALICE_PUBLIC_KEY, ALICE_SECRET_KEY},
     },
+    effect::Effect,
     reactor::{EventQueueHandle, QueueKind, Scheduler},
     tls::KeyFingerprint,
-    types::{DeployId, TestBlockBuilder},
+    types::{
+        chainspec::{CoreConfig, LegacyRequiredFinality},
+        DeployId, TestBlockBuilder,
+    },
     utils,
 };
 
@@ -82,9 +86,12 @@ impl MockReactor {
         reactor_event
     }
 
-    async fn process_effects(&self, mut effects: Effects<Event>) -> Vec<MockReactorEvent> {
+    async fn process_effects(
+        &self,
+        effects: impl IntoIterator<Item = Effect<Event>>,
+    ) -> Vec<MockReactorEvent> {
         let mut events = Vec::new();
-        for effect in effects.drain(0..) {
+        for effect in effects {
             tokio::spawn(async move { effect.await });
             let event = self.crank().await;
             events.push(event);
@@ -150,9 +157,8 @@ impl TestEnv {
 
     fn random(rng: &mut TestRng) -> TestEnv {
         let num_validators: usize = rng.gen_range(3..100);
-        let validator_keys: Vec<Arc<SecretKey>> = iter::repeat(())
+        let validator_keys: Vec<_> = iter::repeat_with(|| Arc::new(SecretKey::random(rng)))
             .take(num_validators)
-            .map(|_| Arc::new(SecretKey::random(rng)))
             .collect();
 
         let num_peers = rng.gen_range(10..20);
@@ -215,6 +221,42 @@ fn register_multiple_signatures<'a, I: IntoIterator<Item = &'a Arc<SecretKey>>>(
     }
 }
 
+fn dummy_merkle_proof() -> TrieMerkleProof<Key, StoredValue> {
+    TrieMerkleProof::new(
+        URef::new([255; 32], AccessRights::NONE).into(),
+        StoredValue::CLValue(CLValue::from_t(()).unwrap()),
+        VecDeque::new(),
+    )
+}
+
+trait OneExt: IntoIterator {
+    fn try_one(self) -> Option<Self::Item>;
+    fn one(self) -> Self::Item;
+}
+
+impl<I: IntoIterator> OneExt for I {
+    fn try_one(self) -> Option<Self::Item> {
+        let mut it = self.into_iter();
+        let first = it.next()?;
+
+        it.next().is_none().then_some(first)
+    }
+
+    #[track_caller]
+    fn one(self) -> Self::Item {
+        let mut it = self.into_iter();
+        let first = it
+            .next()
+            .expect("no element in the iterator, but 1 was expected");
+
+        if it.next().is_some() {
+            panic!("more that 1 element in the iterator, but 1 was expected")
+        }
+
+        first
+    }
+}
+
 impl BlockSynchronizer {
     // Create an initialized block synchronizer
     // with a specified config and MAX_SIMULTANEOUS_PEERS peers
@@ -225,7 +267,13 @@ impl BlockSynchronizer {
     ) -> BlockSynchronizer {
         let mut block_synchronizer = BlockSynchronizer::new(
             config,
-            Arc::new(Chainspec::random(rng)),
+            Arc::new(Chainspec {
+                core_config: CoreConfig {
+                    legacy_required_finality: LegacyRequiredFinality::Strict,
+                    ..CoreConfig::random(rng)
+                },
+                ..Chainspec::random(rng)
+            }),
             MAX_SIMULTANEOUS_PEERS as u32,
             validator_matrix,
             &prometheus::Registry::new(),
@@ -872,8 +920,7 @@ async fn fwd_more_signatures_are_requested_if_weak_finality_is_not_reached() {
             generated_effects
                 .into_iter()
                 .rev()
-                .take(MAX_SIMULTANEOUS_PEERS)
-                .collect(),
+                .take(MAX_SIMULTANEOUS_PEERS),
         )
         .await;
     for event in events {
@@ -1326,11 +1373,7 @@ async fn fwd_registering_approvals_hashes_triggers_fetch_for_deploys() {
             .iter()
             .map(|deploy| deploy.approvals_hash().unwrap())
             .collect(),
-        TrieMerkleProof::new(
-            URef::new([255; 32], AccessRights::NONE).into(),
-            StoredValue::CLValue(CLValue::from_t(()).unwrap()),
-            VecDeque::new(),
-        ),
+        dummy_merkle_proof(),
     );
 
     // Since the block has approvals hashes,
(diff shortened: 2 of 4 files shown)


node/src/components/block_synchronizer/tests.rs
```diff
@@ -215,6 +221,42 @@ fn register_multiple_signatures<'a, I: IntoIterator<Item = &'a Arc<SecretKey>>>(
...
+
+impl<I: IntoIterator> OneExt for I {
+    fn try_one(self) -> Option<Self::Item> {
+        let mut it = self.into_iter();
+        let first = it.next()?;
+
+        it.next().is_none().then_some(first)
+    }
+
+    #[track_caller]
```
rafal-ch, line 245: Please remove, since `Location` is not used.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -225,7 +267,13 @@ impl BlockSynchronizer {
     ) -> BlockSynchronizer {
         let mut block_synchronizer = BlockSynchronizer::new(
             config,
-            Arc::new(Chainspec::random(rng)),
+            Arc::new(Chainspec {
```
rafal-ch: Please remove the comment for `fn new_initialized()`. It's already rotten and will rot even more if we keep extending this constructor. I think it's better to read the code instead of comments :)

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
+fn historical_state(block_synchronizer: &BlockSynchronizer) -> &BlockAcquisitionState {
+    block_synchronizer
+        .historical
+        .as_ref()
+        .unwrap()
+        .block_acquisition_state()
+}
+
+#[tokio::test]
+async fn historical_sync_without_a_deploy_works() {
```
rafal-ch: Can you add a simple comment in the test explaining how the "without_a_deploy" influences the flow? I.e. which states are skipped, etc.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
+        ))
+    );
+
+    // ----- HaveBlock -----
+    assert_matches!(
+        historical_state(&block_synchronizer),
+        BlockAcquisitionState::HaveBlock { .. }
+    );
+
+    // Let's not test the detail of the global syncronization event,
```
rafal-ch: Nit:
```suggestion
    // Let's not test the detail of the global synchronization event,
```

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
```
rafal-ch:
```suggestion
    // Let's not test the detail of the global synchronization event,
```

node/src/components/block_synchronizer/global_state_synchronizer.rs
```diff
@@ -38,6 +38,11 @@ const COMPONENT_NAME: &str = "global_state_synchronizer";
 pub(crate) struct RootHash(Digest);
 
 impl RootHash {
+    #[cfg(test)]
+    pub(crate) fn new(digest: Digest) -> Self {
```
rafal-ch, line 42: Maybe we don't need this new function if we can create required root hash [as easily](https://github.com/casper-network/casper-node/blob/dev/node/src/components/block_synchronizer/global_state_synchronizer.rs#L131) as `RootHash(request.state_root_hash)`.

node/src/types/block.rs
```diff
@@ -1874,6 +1874,19 @@ impl BlockExecutionResultsOrChunk {
     pub fn block_hash(&self) -> &BlockHash {
         &self.block_hash
     }
+
+    #[cfg(test)]
+    pub fn new_mock_value(block_hash: BlockHash) -> Self {
```
rafal-ch: This can probably be a `pub(crate)`.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
+        historical_builder,
+        block,
+        validators_secret_keys
+            .iter()
+            .take(weak_finality_threshold(validators_secret_keys.len())),
+    );
+    assert!(historical_builder.register_block(block, None).is_ok());
+
+    /* Step HaveGlobalState */
+
```
alsrdn: Can you change this comment to give details into what we actually want to check?
```suggestion
    // Check that the block synchronizer starts to acquire global state
```

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
+    let mut block_synchronizer =
+        BlockSynchronizer::new_initialized(rng, validator_matrix, Default::default());
+
+    // Register block for historical sync
+    assert!(block_synchronizer.register_block_by_hash(*block.hash(), true));
+    assert!(block_synchronizer.forward.is_none());
+    block_synchronizer.register_peers(*block.hash(), peers.clone());
+
+    /* Skip steps HaveBlockHeader, HaveWeakFinalitySignature, HaveBlock */
+
```
alsrdn: Can you change this comment to the style of the others (using `//`)? Also remove the newline

node/src/types/value_or_chunk.rs
```diff
@@ -77,6 +77,11 @@ impl<V> ValueOrChunk<V> {
             Ok(ValueOrChunk::ChunkWithProof(chunk_with_proof))
         }
     }
+
+    #[cfg(test)]
+    pub(crate) fn new_value(value: V) -> Self {
```
rafal-ch: Nit: 
I'd just inline this, since it's only called in a single place.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
+    let test_env = TestEnv::random(rng);
+    let peers = test_env.peers();
+    let block = test_env.block();
+    let validator_matrix = test_env.gen_validator_matrix();
+    let validators_secret_keys = test_env.validator_keys();
+    let mut block_synchronizer =
+        BlockSynchronizer::new_initialized(rng, validator_matrix, Default::default());
+
+    // Register block for historical sync
+    assert!(block_synchronizer.register_block_by_hash(*block.hash(), true));
```
rafal-ch: Nit, but I think good for legibility.
Could you please update in this and other tests?
```rust
    let should_fetch_execution_state = true;
    assert!(block_synchronizer.register_block_by_hash(*block.hash(), should_fetch_execution_state));
```

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
+        .await;
+    assert_matches!(
+        events.try_one(),
+        Some(Event::GlobalStateSynchronizer(
+            GlobalStateSynchronizerEvent::GetPeers(_)
+        ))
+    );
+
+    /* Step HaveAllExecutionResults */
+
```
alsrdn: The code doesn't really match the comment here. Since this is a block with no deploys, the synchronizer doesn't attempt to get the execution results or deploys and tries to fetch more signatures directly. Can you adjust this comment?

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1951,3 +1994,526 @@ async fn synchronizer_halts_if_block_cannot_be_made_executable() {
...
+        historical_state(&block_synchronizer),
+        BlockAcquisitionState::HaveGlobalState { .. }
+    );
+    let events = mock_reactor.process_effects(effects).await;
+
+    for event in events {
+        assert_matches!(event, MockReactorEvent::FinalitySignatureFetcherRequest(..));
+    }
+
+    // The next steps are the strict finality ones and are similar to the forward sync
```
alsrdn: Actually the next steps are a bit different for the historical sync flow, but they are tested in the `historical_sync_announces_meta_block` test above. I think you can remove this comment.

node/src/components/block_synchronizer/tests.rs
alsrdn: I would probably rename this into `historical_sync_skips_exec_results_and_deploys_if_block_empty` or something similar.

rafal-ch (commented):
Just few nits.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -225,7 +267,13 @@ impl BlockSynchronizer {
     ) -> BlockSynchronizer {
         let mut block_synchronizer = BlockSynchronizer::new(
             config,
-            Arc::new(Chainspec::random(rng)),
+            Arc::new(Chainspec {
+                core_config: CoreConfig {
+                    legacy_required_finality: LegacyRequiredFinality::Strict,
```
alsrdn: There's quite a variation in the historical sync flow with regards to signature sufficiency determination based on this parameter. There's also `CoreConfig::start_protocol_version_with_strict_finality_signatures_required` to take into account which is tightly related to this parameter.
Can you please add some tests to check transitions to `HaveWeakFinality` and to `HaveStrictFinality` based on variations of these 2 parameters on the historical flow?

alsrdn (changes requested):

node/src/components/block_synchronizer/tests.rs
real-felix: `#[track_caller] allows that the line reported during a panic is the one calling `one()` and not the actual panicking line.

node/src/components/block_synchronizer/global_state_synchronizer.rs
real-felix: It's not in the same module and the inner field is private.

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/types/block.rs
real-felix: Done

node/src/types/value_or_chunk.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: Done

node/src/components/block_synchronizer/tests.rs
real-felix: I added a test that uses `legacy_required_finality: LegacyRequiredFinality::Weak`.

node/src/components/block_synchronizer/tests.rs
```diff
@@ -2518,3 +2521,214 @@ async fn historical_sync_legacy_block() {
...
+    let effects = block_synchronizer.handle_event(
+        mock_reactor.effect_builder(),
+        rng,
+        Event::DeployFetched {
+            block_hash: *block.hash(),
+            result: Either::Left(Ok(FetchedData::from_storage(Box::new(deploy.into())))),
+        },
+    );
+
+    // ----- HaveStrictFinalitySignatures -----
```
alsrdn, line 2920: 👍

alsrdn (commented):
Looks ok to me with the latest changes.

There is however one more code path regarding historical finality signatures (see https://github.com/casper-network/casper-node/blob/0d176b0a41b769214a0f428d844c87f2fc47bddf/node/src/components/block_synchronizer/block_acquisition.rs#L528) which concerns the transition to `HaveWeakFinality`. So I would say you could add a test for a historical builder with `LegacyRequiredFinality::Any`  and register one signature (so that it's not even weak finality) and try to register the block header and body and check that the state transitions correctly.  As I said previously, we'll also need to set `chainspec.core_config.start_protocol_version_with_strict_finality_signatures_required` in order for the test to be reliable.

alsrdn (approved):

rafal-ch (approved):
LGTM :+1: 
Good job!

PR #3901, opened 2023-04-27
Minor cleanups

Fraser999:
This PR adds a `note` to the deprecation attribute of `DEFAULT_RUN_GENESIS_REQUEST` in order to help users better resolve deprecation warnings caused by it.

It also moves deprecated methods to the foot of impl blocks (as per standard recommendations) to leave the main portion of generated docs uncluttered with deprecated methods.

Finally, it removes some `println!` statements from various tests.


rafal-ch (approved):
LGTM :+1:

PR #3906, opened 2023-05-01
Improve several nctl functions

### LINKED ISSUE #3862: Fix the flakiness of the nightly tests

Fraser999:
This PR makes the blocking nctl functions more robust in that they should be successful across a network upgrade.  It resolves the issue of `bash: N/A: division by 0 (error token is "A")` causing upgrade scenario 6 to fail (see [this comment](https://github.com/casper-network/casper-node/issues/3862#issuecomment-1529566944) for details of the issue).

To achieve this, several of the functions in queries.sh were updated to use a new common function which attempts (with retries) to get a specific field from the status REST response of a given node.

The PR also removes a small number of unused functions from queries.rs.

Note that this PR merged with #3907 passed the full nightly test suite locally.

Closes #3862.

### DIFF
--- a/utils/nctl/sh/node/join.sh
+++ b/utils/nctl/sh/node/join.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 
+source "$NCTL"/sh/utils/infra.sh
 source "$NCTL"/sh/utils/main.sh
 source "$NCTL"/sh/node/svc_"$NCTL_DAEMON_TYPE".sh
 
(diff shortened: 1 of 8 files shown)


alsrdn (approved):
LGTM 👍

utils/nctl/sh/node/join.sh
```diff
@@ -28,7 +28,7 @@ BID_DELEGATION_RATE=${BID_DELEGATION_RATE:-2}
 # ----------------------------------------------------------------
 
 # Await genesis era to complete.
-if [ "$(get_chain_era)" -eq 0 ]; then
+if [ "$(get_chain_era)" -lt 1 ]; then
```
rafal-ch: Maybe we can use the common `do_await_genesis_era_to_complete` function here?

Fraser999: Done in [944cf35](https://github.com/casper-network/casper-node/pull/3906/commits/944cf355fd199ad84783c566d047e7ba1272b3b9).

fizyk20 (approved):

rafal-ch (approved):
LGTM :+1:

PR #3907, opened 2023-05-01
Fix network_soundness.py test

### LINKED ISSUE #3905: Fix network_soundness.py test
this test fails in the health check phase, where it expects to see evidence of chunking in the logs, but finds none.

Fraser999:
This PR fixes the issue of chunking not being detected in the network_soundness.py test.

The cause was simply the `block_gas_limit` being too low in the chainspec to support executing the huge deploy.  The test now increases that value before starting the nctl network.

Also, the `run_health_checks` function now accesses all nodes' logs rather than just those of nodes 1..[current running count], which could possibly exclude the single set of logs containing the "chunk" lines.

There are also a couple of changes to avoid generating error messages when invoking `nctl-transfer-native` or `nctl-transfer-wasm` where the specified node is not running.  These messages were polluting the output of this test (at least, the `nctl-transfer-wasm` ones were).

Closes #3905.

### DIFF
--- a/utils/nctl/sh/contracts-transfers/do_dispatch_native.sh
+++ b/utils/nctl/sh/contracts-transfers/do_dispatch_native.sh
@@ -28,8 +28,10 @@ function main()
     local CP1_SECRET_KEY
     local CP1_ACCOUNT_KEY
     local CP2_ACCOUNT_KEY
-    local DISPATCHED
+    local DISPATCH_ATTEMPTS
+    local SUCCESSFUL_DISPATCH_COUNT
     local DISPATCH_NODE_ADDRESS
+    local OUTPUT
 
     CHAIN_NAME=$(get_chain_name)
     GAS_PAYMENT=${GAS_PAYMENT:-$NCTL_DEFAULT_GAS_PAYMENT}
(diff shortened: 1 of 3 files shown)


utils/nctl/sh/contracts-transfers/do_dispatch_native.sh
```diff
@@ -70,19 +73,24 @@ function main()
                 --secret-key "$CP1_SECRET_KEY" \
                 --amount "$AMOUNT" \
                 --target-account "$CP2_ACCOUNT_KEY" \
-                --transfer-id $((DISPATCHED + 1)) \
-                | jq '.result.deploy_hash' \
-                | sed -e 's/^"//' -e 's/"$//'
+                --transfer-id $((DISPATCHED + 1))
```
alsrdn: This should be `DISPATCH_ATTEMPTS` since `DISPATCHED` was removed.

Fraser999: Good catch!  Fixed in [d4da077](https://github.com/casper-network/casper-node/pull/3907/commits/d4da077e2e77bc1c24f30ea0d3578534a45eee7d).

alsrdn (approved):

georgepisaltu (approved):
LGTM

rafal-ch:
Nice one :+1:

PR #3919, opened 2023-05-03
Fix `EraValidatorWeights::signature_weight` in `validator_matrix.rs`

### LINKED ISSUE #3908: Fix BlockSynchronizer test
fwd_more_signatures_are_requested_if_weak_finality_is_not_reached fails if run with env var CL_TEST_SEED=e70314a9c1fb488793cbae505102ed6c

georgepisaltu:
Fixes #3908 

This PR fixes the `EraValidatorWeights` structure allowing exactly `1/3` and `2/3` of the total weight to pass as weak and strict finality respectively. It was also the root cause of some tests failing when the test environment would generate 3 validators. There are new tests for this specific function and also other tests which were failing after the fix were adjusted.

### DIFF
--- a/node/src/components/block_synchronizer/signature_acquisition.rs
+++ b/node/src/components/block_synchronizer/signature_acquisition.rs
@@ -279,7 +279,12 @@ mod tests {
 
     #[test]
     fn should_return_weak_when_weight_1_and_1_10_is_required() {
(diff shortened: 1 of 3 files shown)


Fraser999 (approved):
LGTM - particularly like the new unit tests.

alsrdn (approved):
Nice catch! 👍

rafal-ch (approved):
LGTM :+1:

PR #3927, opened 2023-05-05
GH-3869: Prune empty queues

### LINKED ISSUE #3869: Prune Unbond entries from global state when the queue for a validator is empty.
At the end of processing all unbond requests we should also prune value under Key::Unbond if length of queue reaches zero. Not doing it means we’ll keep a lot of entries with empty queues, and due to the design we’ll always iterate over empty queues when processing unbonding requests. Pruning also decreases time for fast-sync as it doesn’t need to go through effectively empty values. We’re introducing a new keyspace as part of 1.5 and with pruning we won’t end up with bloat if we do pruning. We should also prune Withdraw entries when copying data into Unbond keyspace.

mpapierski:
Closes #3869

This PR adds an ability to purge `Key::Withdraw` entries after upgrading to the `Key::Unbond` keyspace and also purges empty queues in the `Key::Unbond` space to reduce the width of a tip of a Merkle trie.

This is achieved by introducing a `Transform::Delete` operation which is emitted on each auction's step, which is then committed and an entry is purged from the new tip of the trie. New tracking copy operation "delete" is private, and is not exposed to the public, as it's only used within the system contract.


execution_engine/src/shared/transform.rs
```diff
@@ -168,23 +170,42 @@ impl Transform {
...
+            Ok(None) => {
+                // Delete transform can't be handled here as it implies a stored value is present.
+                // Delete transforms should be handled before applying effects on stored values to
+                // avoid an unnecessary global state read.
+                unreachable!("Delete operation can't be applied");
+            }
+            Err(error) => Err(error),
+        }
+    }
+    /// Applies the transformation on a specified stored value instance.
```
Fraser999:
```suggestion
    }

    /// Applies the transformation on a specified stored value instance.
```

execution_engine/src/shared/transform.rs
```diff
@@ -168,23 +170,42 @@ impl Transform {
...
+    ///
+    /// This method produces a new [`StoredValue`] instance based on the [`Transform`] variant. If a
+    /// given transform is a [`Transform::Delete`] then `None` is returned as the [`StoredValue`] is
+    /// consumed but no new value is produced.
+    fn apply_optional(self, stored_value: StoredValue) -> Result<Option<StoredValue>, Error> {
```
Fraser999: I don't think we should have a separate method only to handle `Transform::Delete`s without panicking.

I think it would be much clearer if `apply` were just updated to return the optional stored value and we updated the few callsites to handle this.  Probably in many(all?) cases we could just panic at the callsite if `Ok(None)` is returned, but I feel this is more future-proof/less footgunny than putting the panicking code inside `apply`.

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3548,6 +3552,35 @@ fn should_continue_auction_state_from_release_1_4_x() {
...
+            }
+        })
+        .collect::<Vec<_>>();
+
+    assert!(!delete_keys_after_upgrade.is_empty());
+    assert!(delete_keys_after_upgrade
+        .iter()
+        .all(|key| key.as_withdraw().is_some()));
+
+    // Ensure withdraw keys are purged
```
Fraser999:
```suggestion
    // Ensure withdraw keys are pruned
```

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3717,6 +3750,22 @@ fn should_continue_auction_state_from_release_1_4_x() {
...
+    assert_eq!(withdraw_keys.len(), 0);
+
+    // Unbond keys are deleted
+    let unbond_keys = builder
+        .get_keys(KeyTag::Unbond)
+        .expect("should query withdraw keys");
+    assert_eq!(
+        unbond_keys.len(),
+        0,
+        "auction state continued and empty unbond queue should be purged"
```
Fraser999:
```suggestion
        "auction state continued and empty unbond queue should be pruned"
```

node/src/reactor.rs
```diff
@@ -58,7 +58,6 @@ use tokio::time::{Duration, Instant};
 use tracing::{debug, debug_span, error, info, instrument, trace, warn, Span};
 use tracing_futures::Instrument;
 
-#[cfg(target_os = "linux")]
```
Fraser999, line 63: Removed by mistake?

types/src/execution_result.rs
Fraser999, line 1: We should update `impl Distribution<Transform> for Standard` too to include `Transform::Delete`

Fraser999 (commented):
Mostly nitpicks, but generally looks good to me.  Main concern is the panicking code added to `Transform::apply`.

node/src/reactor.rs
mpapierski: Intentional, actually, as currently build fails on macOS

dwerner (approved):

node/src/reactor.rs
Fraser999: Hmm, seems like `#[cfg(target_os = "linux")]` has been removed from `TARGET_OPEN_FILES_LIMIT` and `fn adjust_open_files_limit()` (compare to [this](https://github.com/casper-network/casper-node/blob/039d438f2c7b15a007e9248154d6f955bf63b740/node/src/reactor.rs#L109) and [this](https://github.com/casper-network/casper-node/blob/039d438f2c7b15a007e9248154d6f955bf63b740/node/src/reactor.rs#L113)).

Could you change these back to include the gating please?  (It was probably me who broke this!)

mpapierski: Tried my best to revert as closely as possible here 6af61de22e0b31f0dd186215e2a07cf30aa55596

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
mpapierski: Fixed

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
mpapierski: Fixed

execution_engine/src/shared/transform.rs
mpapierski: Done as discussed: `Transform::apply` now returns `Option<StoredValue>` where Some indicate newly produced value, and None means no new value was produced and the value should be deleted from the global state.

types/src/execution_result.rs
mpapierski: Done

Fraser999 (approved):

execution_engine_testing/tests/src/test/system_contracts/auction_bidding.rs
```diff
@@ -540,10 +538,189 @@ fn should_run_successful_bond_and_unbond_with_release() {
...
+
+    let bid = bids.get(&default_public_key_arg).expect("should have bid");
+    let bid_purse = *bid.bonding_purse();
+    assert_eq!(
+        builder.get_purse_balance(bid_purse),
+        U512::from(GENESIS_ACCOUNT_STAKE) - unbond_amount, // remaining funds
+    );
+}
+
+#[ignore]
```
real-felix, line 554: Maybe there should be an explanation about why it's ignored?

real-felix (approved):

execution_engine_testing/tests/src/test/system_contracts/auction_bidding.rs
mpapierski: This is our convention for integration tests, so `cargo test` runs unit tests (fast), and running ignored tests runs the integration test suite (could be long)

execution_engine/src/core/runtime/auction_internal.rs
```diff
@@ -98,12 +98,15 @@ where
...
     ) -> Result<(), Error> {
-        self.context
-            .metered_write_gs_unsafe(
-                Key::Unbond(account_hash),
-                StoredValue::Unbonding(unbonding_purses),
-            )
-            .map_err(|exec_error| <Option<Error>>::from(exec_error).unwrap_or(Error::Storage))
+        let unbond_key = Key::Unbond(account_hash);
+        if unbonding_purses.is_empty() {
+            self.context.delete_gs_unsafe(unbond_key);
```
rafal-ch, line 103: Deleting a non-existing key is a noop, not a crash, right?

I noticed the `Use with caution` comment, which indicates that at some point someone will use it without caution :)

rafal-ch (approved):
LGTM :+1: 
with once question

execution_engine/src/core/runtime/auction_internal.rs
mpapierski: Actually deleting nonexisting is not a noop, it's an error. We first need to read the StoredValue under a Key, then we'll apply a transform on it which could result a Some(new_stored_value) (write, add, identity, etc) or None (delete). Then persisting stored values would either write new data to global state (`Some`) or delete it (`None`)

> Unsafe

It's a convention we have, for storage APIs that are called directly from system contracts. It's not unsafe as in rust's unsafe keyword, it's just the fact that there's no validation, and we don't want to expose this API to users yet, so most if not all the methods I introduced are `pub(crate)` at most.

PR #3928, opened 2023-05-05
Fix `itst13.sh` test

### LINKED ISSUE #3862: Fix the flakiness of the nightly tests

rafal-ch:

1. Fixes the hardcoded era number (`4`) used in the assertion of the joining node. Instead it uses the relative one,
2. Adds more logging to the NCTL function that reads LFB mainly to be able to monitor timing issues (in some runs, it was observed that reading all LFBs took ~4min., while normally it should be as quick as 1 second

Additionally it reduces the control logic delay for the initialization stage of the reactor, because it's ok to be more greedy on this phase in an effort to initialize faster

Partially fixes #3862

### DIFF
--- a/node/src/reactor/main_reactor/control.rs
+++ b/node/src/reactor/main_reactor/control.rs
@@ -52,21 +52,27 @@ impl MainReactor {
         rng: &mut NodeRng,
     ) -> (Duration, Effects<MainEvent>) {
         match self.state {
-            ReactorState::Initialize => match self.initialize_next_component(effect_builder) {
-                Some(effects) => (self.control_logic_default_delay.into(), effects),
-                None => {
-                    if false == self.net.has_sufficient_fully_connected_peers() {
-                        info!("Initialize: awaiting sufficient fully-connected peers");
-                        return (self.control_logic_default_delay.into(), Effects::new());
+            ReactorState::Initialize => {
+                // We can be more greedy when cranking through the initialization process as the
+                // progress is expected to happen quickly.
+                let initialization_logic_default_delay = self.control_logic_default_delay / 4;
+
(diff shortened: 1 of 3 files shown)


Fraser999 (approved):

rafal-ch:
bors try

node/src/reactor/main_reactor/control.rs
```diff
@@ -52,21 +52,27 @@ impl MainReactor {
...
-                    if false == self.net.has_sufficient_fully_connected_peers() {
-                        info!("Initialize: awaiting sufficient fully-connected peers");
-                        return (self.control_logic_default_delay.into(), Effects::new());
-                    }
-                    if let Err(msg) = self.refresh_contract_runtime() {
-                        return (Duration::ZERO, fatal!(effect_builder, "{}", msg).ignore());
+            ReactorState::Initialize => {
+                // We can be more greedy when cranking through the initialization process as the
+                // progress is expected to happen quickly.
+                let initialization_logic_default_delay = self.control_logic_default_delay / 4;
```
georgepisaltu, line 58: I'd like to have that `4` as a constant, something like `INITIALIZATION_DELAY_SPEED_UP_FACTOR`.

georgepisaltu (approved):
LGTM, just one comment

node/src/reactor/main_reactor/control.rs
rafal-ch: Sure, good idea.
I'll address this in a separate PR, since this just got merged.

rafal-ch: Addressed [here](https://github.com/casper-network/casper-node/pull/3937).

PR #3934, opened 2023-05-06
Missing imports?

jonas089:
I tried to build the node from source and setup NCTL but the build failed. When I added the above imports, the build finished successfully and the network is now up and running. I'm on MacOS. Please let me know whether this is a bug or I am unaware of something related to this issue:
Screenshot 2023-05-06 at 03 09 52

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -78,7 +78,7 @@ use crate::{
         TrieOrChunk,
     },
     unregister_metric,
-    utils::{self, SharedFlag, WeightedRoundRobin},
+    utils::{self, SharedFlag, WeightedRoundRobin, rlimit::ResourceLimit, rlimit::OpenFiles, rlimit::Limit},
     NodeRng, TERMINATION_REQUESTED,
 };
 pub(crate) use queue_kind::QueueKind;


jonas089:
I have to report more problems here. Building the latest version of casper-client from source works, but cargo-install casper-client doesn't work anymore. Also, when trying to put a deploy using the latest casper client and an nctl instance of the latest node, I get this error:
Screenshot 2023-05-08 at 15 16 27

Fraser999:
Thanks for the reports @jonas089

The initial issue will get fixed on `dev` when [this (unrelated) PR](https://github.com/casper-network/casper-node/pull/3927) gets merged.  If that PR is significantly delayed, I expect we'll cut a small, separate PR just to fix this issue on macOS.

Issues building the client are known to @darthsiroftardis and should be resolved there in the near future.

However, the point you made about the default nctl network's new setting of max TTL for deploys is something I plan to bring up in tomorrow's standup.  We might be able to bump up the default for nctl networks again to 30 mins, but no promises :)  Either way, I'll record the outcome of tomorrow's discussion here for your info.

jonas089:
@Fraser999 Thank you for the detailed answer.

jonas089:
The only MacOS specific issue I observed is that bash can't read the toolchain. All other issues occur on Ubuntu as well.

Fraser999:
The macOS issue I meant was the original one you reported - a failure to build on macOS due to calling a function which was not intended to be called in non-Linux environments.

Since the unrelated PR which included a fix (which I have now linked in my previous comment) is indeed going to be delayed, we've opened a new PR for this issue: #3951

jonas089:
I see, thank you.

Fraser999:
PR #3966 should address the default setting of nctl networks' `max_ttl` value being too small.

I'll close this PR for now as I think all aspects are addressed, but again, thanks for the report!

PR #3935, opened 2023-05-08
Remove the reactor `switch_block_header` field

### LINKED ISSUE #3852: Revisit the MainReactor::switch_block_header field
This field was initially intended to be a cache of the most recent switch block (initially set to None, and once set to Some it should stay Some). At some point, we added code that sets it back to None, which is probably incorrect. We need to revisit this code, possibly removing the switch_block_header field and replacing it with a call to storage. This also requires checking if all places that rely on this field do so correctly

real-felix:
And replace it with storage reads.
Fixes #3852

### DIFF
--- a/node/src/reactor/main_reactor/catch_up.rs
+++ b/node/src/reactor/main_reactor/catch_up.rs
@@ -120,16 +120,19 @@ impl MainReactor {
                 // if too much time has passed, the node will shutdown and require a
                 // trusted block hash to be provided via the config file
                 info!("CatchUp: local tip detected, no trusted hash");
-                if block.header().is_switch_block() {
-                    self.switch_block_header = Some(block.header().clone());
-                }
+
                 Either::Left(SyncIdentifier::LocalTip(
                     *block.hash(),
                     block.height(),
                     block.header().era_id(),
                 ))
             }
-            Ok(None) if self.switch_block_header.is_none() => {
+            Ok(None)
+                if self
+                    .storage
+                    .read_highest_switch_block_headers(1)
+                    .map_or(false, |h| h.is_empty()) =>
+            {
                 // no trusted hash, no local block, might be genesis
                 self.catch_up_check_genesis()
             }
@@ -370,10 +373,6 @@ impl MainReactor {
             %block_hash,
             "CatchUp: leap received"
         );
-
-        if let Err(msg) = self.update_highest_switch_block() {
-            return CatchUpInstruction::Fatal(msg);
-        }
 
         for validator_weights in
             sync_leap.era_validator_weights(self.validator_matrix.fault_tolerance_threshold())
--- a/node/src/reactor/main_reactor/catch_up.rs
+++ b/node/src/reactor/main_reactor/catch_up.rs
@@ -127,22 +127,30 @@ impl MainReactor {
                     block.header().era_id(),
                 ))
             }
-            Ok(None)
-                if self
-                    .storage
-                    .read_highest_switch_block_headers(1)
-                    .map_or(false, |h| h.is_empty()) =>
-            {
-                // no trusted hash, no local block, might be genesis
-                self.catch_up_check_genesis()
-            }
             Ok(None) => {
-                // no trusted hash, no local block, no error, must be waiting for genesis
-                info!("CatchUp: waiting to store genesis immediate switch block");
-                Either::Right(CatchUpInstruction::CheckLater(
-                    "waiting for genesis immediate switch block to be stored".to_string(),
-                    self.control_logic_default_delay.into(),
-                ))
+                let highest_switch_block_headers =
+                    match self.storage.read_highest_switch_block_headers(1) {
+                        Ok(headers) => headers,
+                        Err(err) => {
+                            return Either::Right(CatchUpInstruction::Fatal(format!(
+                                "CatchUp: fatal block store error when attempting to read \
+                                                highest complete block: {}",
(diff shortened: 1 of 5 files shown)


real-felix:
bors try

rafal-ch:
I'll jump into review after we clarify if the failing test is not a regression.

node/src/reactor/main_reactor/catch_up.rs
```diff
@@ -120,16 +120,19 @@ impl MainReactor {
...
                     block.height(),
                     block.header().era_id(),
                 ))
             }
-            Ok(None) if self.switch_block_header.is_none() => {
+            Ok(None)
+                if self
+                    .storage
+                    .read_highest_switch_block_headers(1)
+                    .map_or(false, |h| h.is_empty()) =>
```
Fraser999: We should probably return a `CatchUpInstruction::Fatal` containing the storage error as a string if `read_highest_switch_block_headers` returns `Err`.

node/src/reactor/main_reactor/control.rs
```diff
@@ -398,11 +398,13 @@ impl MainReactor {
...
-                return Err("switch_block should be Some".to_string());
-            }
-            Some(header) => header.clone(),
+        let previous_block_header = match self
+            .storage
+            .read_highest_switch_block_headers(1)
+            .map(|headers| headers.into_iter().next())
+        {
+            Ok(Some(header)) => header,
+            _ => return Err("The highest complete block should be a switch block".to_string()),
```
Fraser999, line 409: Again, I think we should bubble up any storage error.

node/src/reactor/main_reactor/control.rs
```diff
@@ -471,16 +473,13 @@ impl MainReactor {
...
-
-        self.chainspec
-            .protocol_config
-            .is_last_block_before_activation(highest_switch_block_header)
+        match &self.storage.read_highest_switch_block_headers(1).as_deref() {
+            Ok([highest_switch_block_header]) => self
+                .chainspec
+                .protocol_config
+                .is_last_block_before_activation(highest_switch_block_header),
+            _ => false,
```
Fraser999: Here it's maybe fine to just `error!` log if storage returns an error.

node/src/reactor/main_reactor/validate.rs
```diff
@@ -35,7 +35,13 @@ impl MainReactor {
                 self.control_logic_default_delay.into(),
             );
         }
-        if self.switch_block_header.is_none() {
+        if false
+            == self
+                .storage
+                .read_highest_block_header()
+                .unwrap_or(None)
+                .map_or(false, |header| header.is_switch_block())
```
Fraser999: Probably a `ValidateInstruction::Fatal` is appropriate here too.

node/src/reactor/main_reactor/catch_up.rs
real-felix: Done

node/src/reactor/main_reactor/control.rs
real-felix: Done

node/src/reactor/main_reactor/control.rs
real-felix: Done

node/src/reactor/main_reactor/validate.rs
real-felix: Done

real-felix:
bors try

node/src/reactor/main_reactor/control.rs
```diff
@@ -398,11 +398,15 @@ impl MainReactor {
         effect_builder: EffectBuilder<MainEvent>,
     ) -> Result<Effects<MainEvent>, String> {
         info!("{:?}: committing upgrade", self.state);
-        let previous_block_header = match &self.switch_block_header {
-            None => {
-                return Err("switch_block should be Some".to_string());
-            }
-            Some(header) => header.clone(),
+        let previous_block_header = match self
```
alsrdn, line 401: There was a pre-existing issue here. In order to commit the upgrade the previous block header needs to be that of a complete block. `read_highest_switch_block_headers` can return the header of a block that is not complete since it can store a header coming from a sync leap. So I don't think this is entirely correct (and would probably fail down the line when the upgrade is committed with the wrong global state hash taken from this potentially unsynchronized block).

node/src/reactor/main_reactor/control.rs
```diff
@@ -471,16 +475,17 @@ impl MainReactor {
     }
 
     pub(super) fn should_commit_upgrade(&self) -> bool {
-        let highest_switch_block_header = match &self.switch_block_header {
-            None => {
-                return false;
+        match &self.storage.read_highest_switch_block_headers(1).as_deref() {
```
alsrdn, line 478: Same comment as the one from `commit_upgrade` applies here since this check guards the entry into that method. We need to have a complete block that is the last block before activation in order to commit the upgrade.

alsrdn (changes requested):

EdHastingsCasperAssociation:
@Boiethios @rafal-ch has this PR been abandoned, or is there a plan to get it mergeable?

rafal-ch:
@EdHastingsCasperLabs should be abandoned, @alsrdn took care of removing the `switch_block_header` in a [different PR](https://github.com/casper-network/casper-node/pull/3971) as a side-effect.

PR #3936, opened 2023-05-08
Add a faster way to check if nodes 1 to 5 are in sync

### LINKED ISSUE #3862: Fix the flakiness of the nightly tests

rafal-ch:
This test provides a faster way to check if nodes 1 to 5 are in sync. It achieves that by querying nodes in parallel and also skipping the check for RPC port availability, which previously was done by a call to `lsof`.

The previous, slower function has been kept intact, since it is still used in few tests that need to check for a different number of nodes (the new one is hardcoded to 1-5 for simplicity).

Partially fixes #3862

### DIFF
--- a/utils/nctl/sh/scenarios/common/itst.sh
+++ b/utils/nctl/sh/scenarios/common/itst.sh
@@ -129,6 +129,35 @@ function get_reactor_state() {
     echo "$REACTOR_STATE"
 }
 
+function check_nodes_1_to_5_sync_fast() {
+    local ALL_HASHES
+    local UNIQUE_HASH_COUNT
+    local TIMEOUT_SEC=${1-300}
+    local ATTEMPTS=0
+    
+    while [ "$ATTEMPTS" -le "$TIMEOUT_SEC" ]; do
+        ALL_HASHES=$(get_chain_latest_block_hash 1 & \
+                     get_chain_latest_block_hash 2 & \
+                     get_chain_latest_block_hash 3 & \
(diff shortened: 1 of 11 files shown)


rafal-ch:
bors try

utils/nctl/sh/scenarios/common/itst.sh
```diff
@@ -129,6 +129,35 @@ function get_reactor_state() {
     echo "$REACTOR_STATE"
 }
 
+function check_nodes_1_to_5_sync_fast() {
```
georgepisaltu: nit
```suggestion
function parallel_check_nodes_1_to_5_sync() {
```

georgepisaltu (approved):
LGTM with a small nit

utils/nctl/sh/scenarios/common/itst.sh
rafal-ch: Renamed in https://github.com/casper-network/casper-node/pull/3936/commits/d2b58bb7e0e667cfc8ee682ba8bdfe8fadd36e8c

alsrdn (approved):
👍 though I dislike the hardcoding of the range to check.

rafal-ch:
>> +1 though I dislike the hardcoding of the range to check.

>yup, I'm not a fan of it too, but that's the quickest I could do to squeeze the fix for the next nightly run. I'm open to changing this in the future.

the problem has been addressed [here](https://github.com/casper-network/casper-node/pull/3949)

PR #3946, opened 2023-05-09
Disable unneeded casper-types feature in casper-hashing

Fraser999:
This avoids enabling `casper-types/gens` in `casper-hashing`.

A future PR will update casper-types to ensure the `gens` feature can be enabled standalone.


rafal-ch (approved):
LGTM :+1:

sacherjj (approved):

PR #3948, opened 2023-05-09
Version bumps

Fraser999:
This PR bumps the versions of all of the published crates in preparation for the node 1.5.0 release.

In running the `casper-updater` tool, I discovered that a couple of Rust smart contracts had used crates.io as the source of casper dependencies, so this has been rectified in the first commit.

The tool itself also needed to be updated to handle semvers containing a pre-release version, and to include `casper-hashing`'s manifest in the set of files to be updated when `casper-types` is version-bumped.

### DIFF
--- a/smart_contracts/contracts/tutorial/hello-world/Cargo.toml
+++ b/smart_contracts/contracts/tutorial/hello-world/Cargo.toml
@@ -2,11 +2,7 @@
 name = "hello-world"
 version = "0.1.0"
 authors = ["darthsiroftardis <karan@casperlabs.io>", "Michał Papierski <michal@casperlabs.io>"]
-edition = "2021"
-
-[dependencies]
-casper-contract = "1.4.4"
-casper-types = "1.5.0"
+edition = "2018"
 
 [[bin]]
 name = "hello_world"
@@ -14,3 +10,7 @@ path = "src/main.rs"
 bench = false
 doctest = false
 test = false
+
+[dependencies]
+casper-contract = { path = "../../../contract" }
+casper-types = { path = "../../../../types" }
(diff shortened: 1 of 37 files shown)


sacherjj (approved):
Looks like all versions that need to are bumping over the 1.4.15-alt tagging.  Checked that settings from 1.4.15 are in chainspec.toml. Can't find anything wrong.

dwerner (approved):
lgtm

smart_contracts/contracts/tutorial/hello-world/Cargo.toml
```diff
@@ -2,15 +2,15 @@
 name = "hello-world"
 version = "0.1.0"
 authors = ["darthsiroftardis <karan@casperlabs.io>", "Michał Papierski <michal@casperlabs.io>"]
-edition = "2021"
-
-[dependencies]
-casper-contract = "1.4.4"
-casper-types = "1.5.0"
+edition = "2018"
```
rafal-ch, line 5: We're going back with edition?
_(also seen in other place below)_

Fraser999: Well, I think these two smart contracts were maybe copied in from elsewhere and the edition came over with it.

Everywhere else we specify `edition = 2018`, so I think these should be aligned, and when we bump the edition it should be a deliberate decision and applied consistently everywhere in the repo.

rafal-ch: Makes sense, thanks for explanation.

rafal-ch (approved):
LGTM :+1:

PR #3949, opened 2023-05-10
Change `check_network_sync` to `parallel_check_network_sync`

rafal-ch:
This PR implements the `parallel_check_network_sync` NCTL function, which allows asking all nodes at once for an LFB, instead of doing it in serial.

It removes:
* the "hacky" `parallel_check_nodes_1_to_5_sync()` function which was hardcoded to talk to nodes from 1 to 5 only.
* the original `check_network_sync` function


rafal-ch:
bors try

Fraser999 (approved):
LGTM - nice one!

PR #3950, opened 2023-05-11
fast sync: Fix stale sync leaper activity that affects recovery code

alsrdn:
Sometimes the sync leaper leap activity is not correctly reset and causes the node to fail to recover later when it gets out of sync. This happens due to a race condition that appears when we update the validator matrix with the validators for a block after an upgrade.

This happens because the validator matrix update is delayed because the validators need to be read from global state but by doing so the update can interleave within the normal determination of the keep up sync instruction. If the update comes after a leap request is sent, the leap activity will not be reset since the keep up look may not need to check the leap status again if it starts to sync a block and then finishes historical sync.

This patch purges any potential stale sync back leap activities after the node is done syncing back since any new received leap activity is not needed anymore (and it will actually never be checked again until the node switches to CatchUp).

### DIFF
--- a/node/src/components/sync_leaper.rs
+++ b/node/src/components/sync_leaper.rs
@@ -115,6 +115,15 @@ impl SyncLeaper {
         }
     }
 
+    // drop the leap activity for historical sync
+    pub(crate) fn purge_sync_back_activity(&mut self) {
+        if let Some(activity) = &self.leap_activity {
+            if activity.sync_leap_identifier().trusted_ancestor_only() {
+                self.leap_activity = None;
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Fraser999 (approved):

rafal-ch (approved):
LGTM :+1: 

However, please do not merge before we have the test ready. If the test turns out to be too much hassle for this edge case, then we'll reconsider merging.

georgepisaltu (approved):

rafal-ch:
As agreed on the test huddle, we can merge w/o a test for control logic.

alsrdn:
Fixes: https://github.com/casper-network/casper-node/issues/3958

PR #3955, opened 2023-05-15
Remove "no more eras to delete" message.

mpapierski:
This PR removes a log message that happens on each block after pruning is completed. 

This message is not entirely correct as it describes very specific behavior. And logging "do nothing" every block is also too verbose even in debug mode. Not logging anything is the desired behavior now that all known networks were migrated successfully.


rafal-ch (approved):
LGTM :+1:

PR #3966, opened 2023-05-17
Revert change to max_ttl setting of local test chainspec

### LINKED ISSUE #3960: Revert change of max_ttl in resources/local/chainspec.toml.in
The max_ttl was reduced to '5min' for some nctl tests a few months ago.  However, this isn't ideal for users to test using the Rust client as it defaults to 30 mins TTL for new deploys.  We should revert the chainspec change and create override files for those nctl tests which actually require the max_ttl to be very low.

Fraser999:
This PR reverts a recent change to the `max_ttl` setting of the chainspec used for local testing and nctl networks.

After running the full test suite including nightly nctl tests, it appears that no tests depend upon this being set to `'5minutes'`.

Closes #3960.


alsrdn (approved):

Fraser999:
bors merge

PR #3968, opened 2023-05-18
Purge `SyncLeaper` when switching to `CatchUp` mode

### LINKED ISSUE #3965: Failure running network_soundness.py test
The nightlies failed on this run: https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8286  It appears that a joining node started sync leaping using a trusted hash which was too old.

Fraser999:
This PR fixes an issue where a historical sync leap is incorrectly interpreted as current just after transitioning from `KeepUp` to `CatchUp` mode.  Ultimately, this causes the node to shut down as further leap attempts are made using the historical block's hash as the trusted hash, and the peers respond with `FetchResponse::NotProvided`.

The fix is to purge the `SyncLeaper` as the node transitions from `KeepUp` to `CatchUp`.

The PR also removes a small, unneeded block from `catch_up_syncing` where the attempt counter is reset.  This has always already just been performed via `self.update_last_progress` in `catch_up_process`.

Finally, the bash and python scripts used by drone in CI have been updated to invoke bash calls in non-interactive mode (`bash -c ...` rather than `bash -i ...`) as the output in CI is polluted with many instances of 
```
bash: cannot set terminal process group (1): Inappropriate ioctl for device
bash: no job control in this shell
```
(see [this test run](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8286/1/4) for example).

Closes #3965.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -136,7 +136,7 @@ steps:
 - name: nctl-compile
   <<: *buildenv
   commands:
-  - bash -i ./ci/nctl_compile.sh
+  - bash -c ./ci/nctl_compile.sh
 
 - name: nctl-upgrade-test
   <<: *buildenv
@@ -146,12 +146,12 @@ steps:
     AWS_SECRET_ACCESS_KEY:
       from_secret: put-drone-aws-sk
   commands:
-  - bash -i ./ci/nctl_upgrade.sh
+  - bash -c ./ci/nctl_upgrade.sh
 
 - name: check CPU features
   <<: *buildenv
   commands:
-  - bash -i ./ci/check_cpu_features.sh  
+  - bash -i ./ci/check_cpu_features.sh
 
 volumes:
 - name: rustup
@@ -480,7 +480,7 @@ steps:
 - name: nctl-compile
   <<: *buildenv
   commands:
-  - bash -i ./ci/nctl_compile.sh
+  - bash -c ./ci/nctl_compile.sh
 
 - name: nctl-nightly-tests
   <<: *buildenv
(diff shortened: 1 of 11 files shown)


.drone.yml
```diff
@@ -146,12 +146,12 @@ steps:
...
       from_secret: put-drone-aws-sk
   commands:
-  - bash -i ./ci/nctl_upgrade.sh
+  - bash -c ./ci/nctl_upgrade.sh
 
 - name: check CPU features
   <<: *buildenv
   commands:
-  - bash -i ./ci/check_cpu_features.sh  
+  - bash -i ./ci/check_cpu_features.sh
```
alsrdn, line 154: Any reason to not run this as non-interactive too?

node/src/reactor/main_reactor/keep_up.rs
```diff
@@ -296,7 +296,7 @@ impl MainReactor {
                     // be required anymore
                     debug!("KeepUp: synced to TTL or Genesis");
                     self.block_synchronizer.purge_historical();
-                    self.sync_leaper.purge_sync_back_activity();
+                    self.sync_leaper.purge();
```
alsrdn, line 299: This is probably not needed anymore since the purge now happens on state transition and it would catch this case as well. But I guess it won't hurt either since `KeepUp` only issues historical leaps so it won't affect the forward flow either.

alsrdn (approved):
LGTM 👍

.drone.yml
Fraser999: I think it failed if I remember right, so I thought this could get left as is since it will only create minimal pollution of the logs.

Fraser999:
bors merge

PR #3978, opened 2023-05-22
GH-3849: Get keys deserialize fix

### LINKED ISSUE #3849: get_keys should never deserialize a StoredValue
Keys iterator in global state should never deserialize a StoredValue. Currently, the iterator is supposed to yield keys matching a prefix, but it also deserializes a StoredValue just to throw it away.

mpapierski:
Closes #3849

This PR ensures that the `keys_with_prefix` iterator never deserializes `StoredValue`. This is accomplished by adding two zero-overhead types:
- `EnsureNeverDeserializes` wrapper for a `TrieStore` that will panic under debug mode when a value is deserialized, and
- `PanickingFromBytes` that wraps V, but the implementation will always panic when it's deserialized.

### DIFF
--- a/execution_engine/src/storage/trie_store/operations/debug_store.rs
+++ b/execution_engine/src/storage/trie_store/operations/debug_store.rs
@@ -1,0 +1,54 @@
+use std::marker::PhantomData;
+
+use casper_hashing::Digest;
+use casper_types::bytesrepr::{self, FromBytes};
+
+use crate::storage::{store::Store, trie::Trie, trie_store::TrieStore};
+
+/// A [`TrieStore`] wrapper that panics in debug mode whenever an attempt to deserialize [`V`] is
+/// made, otherwise it behaves as a [`TrieStore`].
+///
+/// The debug panic is used to ensure that this wrapper has  To ensure this wrapper has zero
+/// overhead, a debug assertion is used.
+pub(crate) struct EnsureNeverDeserializes<'a, K, V, S>(&'a S, PhantomData<*const (K, V)>)
+where
+    S: TrieStore<K, V>;
+
+impl<'a, K, V, S> EnsureNeverDeserializes<'a, K, V, S>
+where
+    S: TrieStore<K, V>,
+{
+    pub(crate) fn new(store: &'a S) -> Self {
+        Self(store, PhantomData)
+    }
+}
+
+impl<'a, K, V, S> Store<Digest, Trie<K, V>> for EnsureNeverDeserializes<'a, K, V, S>
+where
+    S: TrieStore<K, V>,
+{
+    type Error = S::Error;
+
+    type Handle = S::Handle;
+
+    #[inline]
+    fn handle(&self) -> Self::Handle {
+        self.0.handle()
+    }
+
+    #[inline]
+    fn deserialize_value(&self, bytes: &[u8]) -> Result<Trie<K, V>, bytesrepr::Error>
+    where
+        Trie<K, V>: FromBytes,
+    {
+        #[cfg(debug_assertions)]
+        {
+            let _ = bytes;
+            panic!("Tried to deserialize a value but expected no deserialization to happen.")
+        }
+        #[cfg(not(debug_assertions))]
+        {
+            bytesrepr::deserialize_from_slice(bytes)
+        }
+    }
+}
--- a/execution_engine/src/storage/trie_store/operations/mod.rs
+++ b/execution_engine/src/storage/trie_store/operations/mod.rs
@@ -1,4 +1,4 @@
-pub(crate) mod debug_store;
+pub(crate) mod store_wrappers;
 #[cfg(test)]
 mod tests;
 
@@ -26,7 +26,7 @@ use crate::{
     },
 };
 
-use self::debug_store::EnsureNeverDeserializes;
+use self::store_wrappers::NonDeserializingStore;
 
 #[allow(clippy::enum_variant_names)]
 #[derive(Debug, PartialEq, Eq)]
@@ -1038,7 +1038,7 @@ struct VisitedTrieNode<K, V> {
 pub struct KeysIterator<'a, 'b, K, V, T, S: TrieStore<K, V>> {
     initial_descend: VecDeque<u8>,
     visited: Vec<VisitedTrieNode<K, V>>,
-    store: EnsureNeverDeserializes<'a, K, V, S>, //&'a S,
+    store: NonDeserializingStore<'a, K, V, S>,
     txn: &'b T,
     state: KeysIteratorState<K, V, S>,
 }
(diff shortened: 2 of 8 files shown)


execution_engine/src/storage/trie_store/operations/debug_store.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use casper_hashing::Digest;
+use casper_types::bytesrepr::{self, FromBytes};
+
+use crate::storage::{store::Store, trie::Trie, trie_store::TrieStore};
+
+/// A [`TrieStore`] wrapper that panics in debug mode whenever an attempt to deserialize [`V`] is
+/// made, otherwise it behaves as a [`TrieStore`].
+///
+/// The debug panic is used to ensure that this wrapper has  To ensure this wrapper has zero
```
rafal-ch: Some copy&paste error in the comment probably.

execution_engine/src/storage/trie_store/operations/tests/bytesrepr_utils.rs
```diff
@@ -0,0 +1,37 @@
+use casper_types::bytesrepr::{self, FromBytes, ToBytes};
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub(crate) struct PanickingFromBytes<T>(T);
```
rafal-ch, line 4: Is this used anywhere else than tests?

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -1027,15 +1030,15 @@ enum KeysIteratorState<K, V, S: TrieStore<K, V>> {
...
+    trie: LazyTrieLeaf<K, V>,
     maybe_index: Option<usize>,
     path: Vec<u8>,
 }
 
 pub struct KeysIterator<'a, 'b, K, V, T, S: TrieStore<K, V>> {
     initial_descend: VecDeque<u8>,
     visited: Vec<VisitedTrieNode<K, V>>,
-    store: &'a S,
+    store: EnsureNeverDeserializes<'a, K, V, S>, //&'a S,
```
rafal-ch: Stray comment? If no, please make it more descriptive.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -1067,25 +1070,39 @@ where
...
+                                self.state = KeysIteratorState::Failed;
+                                return Some(Err(error.into()));
+                            }
+                        };
                         return Some(Ok(key));
                     }
                 }
-                Trie::Node { ref pointer_block } => {
+                LazyTrieLeaf::Right(Trie::Leaf { .. }) => {
+                    unreachable!("Lazy trie deserializer ensures that this variant never happens.")
```
rafal-ch, line 1103: I'm wondering if we can encode this in the type system, so that this `unreachable!` arm is no longer needed.

execution_engine/src/storage/trie_store/operations/debug_store.rs
```diff
@@ -0,0 +1,54 @@
...
+/// overhead, a debug assertion is used.
+pub(crate) struct EnsureNeverDeserializes<'a, K, V, S>(&'a S, PhantomData<*const (K, V)>)
```
alsrdn: nit: since it's not a trait I'd rename this to something like `NonDeserializingStore`

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -1,11 +1,11 @@
+pub(crate) mod debug_store;
```
alsrdn: nit: the name of the module is kind of weird since it's not only used for debug. I'd suggest naming it `store_wrappers` or similar.

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -1067,25 +1070,39 @@ where
...
-                        }
-                    };
-                    debug_assert!(key_bytes.starts_with(&path));
+                LazyTrieLeaf::Left(leaf_bytes) => {
+                    if leaf_bytes.is_empty() {
+                        self.state = KeysIteratorState::Failed;
+                        return Some(Err(bytesrepr::Error::Formatting.into()));
+                    }
+
+                    let key_bytes = &leaf_bytes[1..]; // Skip `Trie::Leaf` tag
```
alsrdn, line 1082: nit: maybe it would be good to add a comment to `ToBytes`/`FromBytes` in `Trie` to point out that there are places where we do piecewise deserialization of it. It's good to know if we ever change that implementation that we also need to re-factor this code for it to work correctly.

alsrdn (approved):
A few nits but otherwise LGTM

execution_engine/src/storage/trie_store/operations/tests/bytesrepr_utils.rs
mpapierski: It's only used in tests, and in fact, it's in tests mod which is `#[cfg(test)]` only

execution_engine/src/storage/trie_store/operations/mod.rs
mpapierski: We could, something like `enum LazyTrie<K, V> { Leaf(Bytes, PhantomData<(K, V)>), Node(TrieNode), Extension(TrieExtension) }` etc. Now that we'll do more lazy trie leaf deserialization I added a ticket to deal with this

execution_engine/src/storage/trie_store/operations/mod.rs
mpapierski: Added comment

execution_engine/src/storage/trie_store/operations/mod.rs
mpapierski: Renamed

execution_engine/src/storage/trie_store/operations/debug_store.rs
mpapierski: Renamed

execution_engine/src/storage/trie_store/operations/mod.rs
mpapierski: Removed

execution_engine/src/storage/trie_store/operations/debug_store.rs
mpapierski: Fixed

execution_engine/src/storage/trie_store/operations/mod.rs
mpapierski: Ticket https://github.com/casper-network/casper-node/issues/3994

rafal-ch (approved):
LGTM :+1:

PR #3981, opened 2023-05-24
Fix stuck historical sync on legacy deploys without approvals

### LINKED ISSUE #3976: Node stuck on historical sync when a block contains deploy with no approvals
It may happen that invalid deploy has been included in the chain (old binaries used to [allow that](https://github.com/casper-network/casper-node/blob/release-1.0.0/node/src/types/deploy.rs#L806-L817)). Such deploy will have the following execution results: `"error_message": "Authorization failure: not authorized.".`
When the "new" node is trying to sync this block, it'll fetch such deploy and fail to validate it, which will result in the peer being banned:
```
{"message":"banning peer","offender":"tls:d5ec..bf7c","justification":"sent a legacy deploy which failed validation (the deploy has no approvals)"}

{"message":"PeerList: no honest peers"}

```
We should update the validation so that `Fetcher<LegacyDeploy> `doesn't check approvals (the `has_valid_hash() `check should stay intact).
This behavior should ideally be covered with NCTL test when a deploy w/o approvals is sent to node 1.0.0, then there's an upgrade to 1.5.x and the new node is able to sync correctly.

rafal-ch:
This PR loosens the validation criteria for the `LegacyDeploy` fetcher and provides a NCTL test to prove that 1.5+ nodes are not stuck when syncing block with deploy which has no approvals.

Closes https://github.com/casper-network/casper-node/issues/3976

### DIFF
--- a/node/src/types/deploy/legacy_deploy.rs
+++ b/node/src/types/deploy/legacy_deploy.rs
@@ -11,16 +11,6 @@ use crate::components::fetcher::{EmptyValidationMetadata, FetchItem, Tag};
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, DataSize, Debug)]
 pub(crate) struct LegacyDeploy(Deploy);
 
-impl LegacyDeploy {
-    #[allow(clippy::result_large_err)]
-    fn is_valid(&self) -> Result<(), DeployConfigurationFailure> {
-        self.0
-            .is_valid
-            .get_or_init(|| validate_legacy_deploy(&self.0).map_err(|err| *err))
-            .clone()
-    }
-}
-
 impl FetchItem for LegacyDeploy {
     type Id = DeployHash;
     type ValidationError = DeployConfigurationFailure;
@@ -33,7 +23,7 @@ impl FetchItem for LegacyDeploy {
     }
 
     fn validate(&self, _metadata: &EmptyValidationMetadata) -> Result<(), Self::ValidationError> {
-        self.is_valid()
+        self.0.has_valid_hash().map_err(|err| *err)
     }
 }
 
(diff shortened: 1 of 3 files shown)


rafal-ch:
bors try

node/src/types/deploy/legacy_deploy.rs
```diff
@@ -11,6 +11,16 @@ use crate::components::fetcher::{EmptyValidationMetadata, FetchItem, Tag};
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, DataSize, Debug)]
 pub(crate) struct LegacyDeploy(Deploy);
 
+impl LegacyDeploy {
+    #[allow(clippy::result_large_err)]
+    fn is_valid(&self) -> Result<(), DeployConfigurationFailure> {
+        self.0
+            .is_valid
+            .get_or_init(|| validate_legacy_deploy(&self.0).map_err(|err| *err))
+            .clone()
```
Fraser999: I think we should avoid this, since we could set the inner Deploy's `is_valid` field to `Ok` and then extract the inner Deploy via the `From` impl below, yielding a Deploy which claims to be valid, but might not be.

Instead, we could just call `self.0.has_valid_hash()` inside `fn validate` and remove this `impl LegacyDeploy` altogether.

Fraser999 (changes requested):
Partial review for now.

node/src/types/deploy/legacy_deploy.rs
rafal-ch: Good catch!
I'll update after considering if we might want to have another `is_valid` cell on the `LegacyDeploy` itself.

rafal-ch: The `LegacyDeploy::validate()` is called 3 times per deploy; taking into consideration that `has_valid_hash()` is pretty cheap call, we'll go with the simple solution suggested by @Fraser999.

Updated 1b344ac3b6a399e60ed7593772cc8c51e4600e1d.

Fraser999 (approved):

PR #3983, opened 2023-05-24
Fixes speculative server configuration in NCTL setup scripts

### LINKED ISSUE #3982: Outdated configuration of speculative server variables in 'config.toml' setup
The NCTL configuration scripts (e.g.: setup.sh, and similar) seems to be outdated. Speculative server has its own configuration section and the script should be updated accordingly.

We have updated it for testing the Java SDK for the upcoming v1.5 release and will provide a PR for helping to fix this.

@sacherjj, @asladeofgreen asked me to tag you on this one. I'll send the PR shortly.

oak:
Closes #3982

@sacherjj As stated on the related issue, this updates the speculative server configuration by updating related NCTL scripts.

### DIFF
--- a/utils/nctl/sh/assets/setup.sh
+++ b/utils/nctl/sh/assets/setup.sh
@@ -69,7 +69,7 @@ function _set_nodes()
     do
         PATH_TO_CFG=$(get_path_to_node "$IDX")/config/1_0_0
         PATH_TO_FILE="$PATH_TO_CFG"/config.toml
-    
+
         cp "$PATH_TO_CONFIG_TOML" "$PATH_TO_CFG"
         cp "$(get_path_to_net)"/chainspec/* "$PATH_TO_CFG"
 
--- a/utils/nctl/sh/assets/setup.sh
+++ b/utils/nctl/sh/assets/setup.sh
@@ -63,6 +63,7 @@ function _set_nodes()
 
     local IDX
     local PATH_TO_FILE
+    local SPECULATIVE_EXEC_ADDR
 
     for IDX in $(seq 1 "$(get_count_of_nodes)")
     do
@@ -71,6 +72,8 @@ function _set_nodes()
 
         cp "$PATH_TO_CONFIG_TOML" "$PATH_TO_CFG"
         cp "$(get_path_to_net)"/chainspec/* "$PATH_TO_CFG"
+
+        SPECULATIVE_EXEC_ADDR=$(grep 'speculative_exec_server' $PATH_TO_FILE || true)
 
         local SCRIPT=(
             "import toml;"
@@ -83,7 +86,15 @@ function _set_nodes()
             "cfg['rest_server']['address']='0.0.0.0:$(get_node_port_rest "$IDX")';"
             "cfg['rpc_server']['address']='0.0.0.0:$(get_node_port_rpc "$IDX")';"
             "cfg['event_stream_server']['address']='0.0.0.0:$(get_node_port_sse "$IDX")';"
-            "cfg['speculative_exec_server']['address']='0.0.0.0:$(get_node_port_speculative_exec "$IDX")';"
+        )
+
+        if [ ! -z "$SPECULATIVE_EXEC_ADDR" ]; then
+            SCRIPT+=(
+                "cfg['speculative_exec_server']['address']='0.0.0.0:$(get_node_port_speculative_exec "$IDX")';"
+            )
+        fi
+
+        SCRIPT+=(
             "toml.dump(cfg, open('$PATH_TO_FILE', 'w'));"
         )
 
--- a/utils/nctl/sh/assets/setup.sh
+++ b/utils/nctl/sh/assets/setup.sh
@@ -69,11 +69,11 @@ function _set_nodes()
     do
         PATH_TO_CFG=$(get_path_to_node "$IDX")/config/1_0_0
         PATH_TO_FILE="$PATH_TO_CFG"/config.toml
-
+    
         cp "$PATH_TO_CONFIG_TOML" "$PATH_TO_CFG"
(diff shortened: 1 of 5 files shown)


utils/nctl/sh/assets/setup.sh
```diff
@@ -86,12 +86,11 @@ function _set_nodes()
             "cfg['rest_server']['address']='0.0.0.0:$(get_node_port_rest "$IDX")';"
             "cfg['rpc_server']['address']='0.0.0.0:$(get_node_port_rpc "$IDX")';"
             "cfg['event_stream_server']['address']='0.0.0.0:$(get_node_port_sse "$IDX")';"
-            "toml.dump(cfg, open('$PATH_TO_FILE', 'w'));"
         )
 
         if [ ! -z "$SPECULATIVE_EXEC_ADDR" ]; then
             SCRIPT+=(
-                "cfg['rpc_server']['speculative_execution_address']='0.0.0.0:$(get_node_port_speculative_exec "$IDX")';"
+                "cfg['speculative_exec_server']['address']='0.0.0.0:$(get_node_port_speculative_exec "$IDX")';"
```
Fraser999, line 93: I think we'll need the following in each of the three cases too:
```bash
"cfg['speculative_exec_server']['enable_server']=True;"
```

Fraser999 (commented):
Thanks for the PR!

I think it'd be good to enable the speculative execution servers by default in nctl, hence the comment.

Otherwise, looks good to me.

utils/nctl/sh/assets/setup.sh
oak: Thanks for your response, @Fraser999 :)

I think it could be done this way, but I believe these more 'default' settings values come from the original config.toml, am I mistaken? We could just change the enable_server there and follow the other servers pattern, what do you think?

[casper-node/resources/local/config.toml](https://github.com/casper-network/casper-node/blob/dev/resources/local/config.toml#L253-L297)

Fraser999: That's a good point - probably more overt that way :+1:

Fraser999 (approved):
LGTM - thanks!

Fraser999:
bors merge

utils/nctl/sh/assets/setup.sh
Fraser999: Hmm - seems like [one of the nctl upgrade tests failed in CI](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8335/2/4) for this PR.

Looking at the logs and configs from the dump (which can be retrieved via `curl -O https://s3.us-east-2.amazonaws.com/nctl.casperlabs.io/nightly-logs/8335_nctl_dump.tar.gz`) the nodes all have these lines in their `2_0_0/config.toml` files:
```toml
[speculative_exec_server]
enable_server = true
address = "0.0.0.0:7778"
```

I think we can fix this by either following the approach I suggested above (have `enable_server = false`  in `resources/local/config.toml` and only update that value in the bash script where we also set the correct port), or preferably we could unconditionally set the correct port in the bash scripts and leave `enable_server = true` in `resources/local/config.toml`.

I believe the logic around `if [ ! -z "$SPECULATIVE_EXEC_ADDR" ]; then` in the bash scripts is left over from a previous approach to updating configs in tests, and is no longer needed.  We could unconditionally include `"cfg['speculative_exec_server']['address']='0.0.0.0:$(get_node_port_speculative_exec "$IDX")';"` in the `SCRIPT` array when it's being constructed (anywhere between the `"cfg=toml.load...` and `"toml.dump...` entries).  That would let us remove the `SPECULATIVE_EXEC_ADDR` variable altogether.

oak: Thanks for your analysis, can't check the drone output, think I lack authorization to do so, but have seen the tarball.

I agree with your suggestion of always setting the port regardless, I think is the way to go. I was trying to fiddle the less I could with the code, but I share the opinion that this is indeed desirable.

Furthermore, I did this change and also included the setting code in one place that did not follow the same pattern, so I did not find it the first time around. I searched the code for any similar block, and believe I was thorough this time around.

Fraser999: So it seems I steered you a little wrong - apologies!  Apparently we do still need to use the `if [ ! -z "$SPECULATIVE_EXEC_ADDR" ]; then` in the `setup_shared.sh` file (updated of course to use `SPECULATIVE_EXEC_ADDR=$(grep 'speculative_exec_server' $PATH_TO_CONFIG_FILE || true)`.

CI effectively runs the following command from the root folder:
```bash
bash -i ./ci/nctl_upgrade.sh
```

I'm running this just now locally using your branch with the change to `setup_shared.sh` I mentioned above, but it takes quite a while to execute, so I'll not be able to see the results until tomorrow.  It has at least got through the first upgrade test now, where previously it was failing.

oak: Yes, I think the thing we missed was *upgrade.sh*. After running locally, `bash -i ./ci/nctl_upgrade.sh` I was able to reproduce and hopefully fix this. Putting the same logic into *upgrade.sh* appears to have solved the issue, so I reverted the last commit and did another one changing it. I'll push the changes if you want to give it a try, ok? Cheers

Fraser999: Great!  Thanks for sticking with this.  It's passed locally for me too, so I'll try and merge now.

Fraser999:
bors merge

utils/nctl/sh/assets/setup.sh
oak: Cool! Np, thanks for your assistance! I took the chance to learn a bit more about the whole stack, so it was well worth it. :)

PR #3984, opened 2023-05-24
EE: Dont deserialize a value `V` when writing a trie or when reading a trie if the key was not found

### LINKED ISSUE #3850: Trie read operation should deserialize a StoredValue at most 1 time
We should put tests in place to ensure that reading a value from a trie should deserialize it exactly once if a key is found, or zero times if a key does not exists.

### LINKED ISSUE #3851: Trie write operation should never deserialize a StoredValue
We should put tests in place to ensure that we don’t deserialize a value during a write operation.

alsrdn:
Make sure we don't deserialize a value `V` when writing a trie.
Also put in place tests to verify this doesn't happen and that reading a trie deserializes `V` only once if the key is found or 0 times if the key is not found. The tests were extended by using the existing counter that is used to track the call to `from_bytes` for `V`. The tracker can be used now with any operation, not only `delete`.

Fixes: https://github.com/casper-network/casper-node/issues/3850 and https://github.com/casper-network/casper-node/issues/3851


execution_engine/src/storage/trie_store/operations/tests/mod.rs
```diff
@@ -15,6 +15,8 @@ use std::{
     ops::Not,
 };
 
+use backtrace::Backtrace;
+use libc::c_void;
```
EdHastingsCasperAssociation: can I get some info about the choice to attempt backtracing etc for these tests?

execution_engine/src/storage/trie_store/operations/mod.rs
```diff
@@ -925,52 +890,67 @@ where
...
-                    let (new_node, parents) = reparent_leaf(&path, &existing_leaf_path, parents)?;
-                    let parents = add_node_to_parents(&path, new_node, parents);
-                    rehash(new_leaf, parents)?
+                LazyTrieLeaf::Left(leaf_bytes) => {
+                    let trie_tag = trie::lazy_trie_tag(leaf_bytes.as_slice());
+                    assert_eq!(
+                        trie_tag,
+                        Some(TrieTag::Leaf),
+                        "Unexpected trie variant found instead of a `TrieTag::Leaf`"
+                    );
```
rafal-ch, line 913: Are we sure we want to panic inside `write()` in case of mismatch?

execution_engine/src/storage/trie_store/operations/tests/mod.rs
```diff
@@ -129,13 +132,27 @@ impl ToBytes for TestValue {
     }
 }
 
+// Determine if a there exists a caller in the backtrace that matches any of the specified symbols
```
rafal-ch: Nit:
```suggestion
// Determine if there exists a caller in the backtrace that matches any of the specified symbols
```

execution_engine/src/storage/trie_store/operations/mod.rs
mpapierski: Currently, it's guaranteed by lazy trie deserializer that the panic wouldn't happen - we have ticket #3994 to address this by using proper enum rather than Either

mpapierski (approved):

execution_engine/src/storage/trie_store/operations/tests/mod.rs
alsrdn: We wanted to track the number of times `from_bytes` was called when deserializing a value. One way to do this is to track the callers of `from_bytes` and determine if it's one of the functions that we are trying to restrict. Now we use a different strategy to achieve the same thing (with `NonDeserializingStore`, `OnceDeserializingStore` and `PanickingFromBytes`) so `backtrace` is not needed anymore.

execution_engine/src/storage/trie_store/operations/tests/mod.rs
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/3984/commits/60be67bc699821fc9b2f664f05a8ac6db2b057d2

rafal-ch (approved):
LGTM :+1:

alsrdn:
bors merge

PR #4013, opened 2023-06-02
Ensure node can transition out of validate mode

### LINKED ISSUE #4012: Investigate nightly test failure
network_soundness.py test failed in [this run](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8394/1/4).  Log dump can be fetched via curl -O https://s3.us-east-2.amazonaws.com/nctl.casperlabs.io/nightly-logs/8394_nctl_dump.tar.gz
Assignee: me

Fraser999:
This PR fixes an issue whereby a node in `Validate` mode will not revert to `KeepUp` or `CatchUp` if it falls behind the objective tip of the chain.

After initially transitioning to `Validate` having spent a while doing the initial sync, the node could be well behind the objective tip.  In the likely case that its highest complete block is not a switch block, it would effectively be stuck waiting to produce a switch block before re-evaluating whether it should remain in `Validate` mode.  This switch block will never be created as it's fallen behind its peers and does nothing else to rectify that in `Validate` mode.

The fix is to query the `BlockAccumulator` as it does when in `KeepUp` mode.

The `ValidateInstruction::NonSwitchBlock` is also removed as it was just a special case of `ValidateInstruction::CheckLater`, but with no special handling of that variant.

Closes #4012.

### DIFF
--- a/node/src/reactor/main_reactor/validate.rs
+++ b/node/src/reactor/main_reactor/validate.rs
@@ -2,7 +2,10 @@ use std::time::Duration;
 use tracing::{debug, info, warn};
 
 use crate::{
-    components::consensus::ChainspecConsensusExt,
+    components::{
+        block_accumulator::{SyncIdentifier, SyncInstruction},
+        consensus::ChainspecConsensusExt,
+    },
     effect::{EffectBuilder, Effects},
     reactor::{
         self,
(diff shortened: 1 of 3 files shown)


node/src/reactor/main_reactor/validate.rs
```diff
@@ -38,8 +44,23 @@ impl MainReactor {
 
         match self.storage.read_highest_complete_block() {
             Ok(Some(highest_complete_block)) => {
+                // If we're lagging behind the rest of the network, fall back out of Validate mode.
+                let sync_identifier = SyncIdentifier::LocalTip(
+                    *highest_complete_block.hash(),
+                    highest_complete_block.height(),
+                    highest_complete_block.header().era_id(),
+                );
+                match self.block_accumulator.sync_instruction(sync_identifier) {
```
rafal-ch, line 53: Just to be sure - for a block too old this will create the `LeapInstruction::OutsideAttemptExecutionThreshold`, which - in turn, will create `SyncInstruction::Leap`, right?

rafal-ch (approved):
LGTM :+1:

node/src/reactor/main_reactor/validate.rs
Fraser999: Yes, I believe so.

real-felix (approved):

Fraser999:
bors merge

PR #4022, opened 2023-06-05
GH-3734: Increase default max stack height

### LINKED ISSUE #3734: Increase max_memory and max_stack_height chainspec values.
Currently max_memory is set to 64 wasm pages which represents ~4MB of memory, and max_stack_height is set to 188 which at that time were very safe choices in terms of security. Certain uses such as using non-trivial 3rd party crypto libraries requires more memory to operate, and also larger stack height to work.

max_stack_height increase should be safe (within reasonable values), since we’re charging a lot for control flow operations anyway. For increasing max_memory we may consider properly charging per extra pages allocated if the new limit is much higher than current, or find different method.

mpapierski:
Closes #3734

Increase default max stack height.

This should enable the usage of certain 3rd party crates that are quite expensive in terms of stack height cost.

### DIFF
--- a/execution_engine/CHANGELOG.md
+++ b/execution_engine/CHANGELOG.md
@@ -10,6 +10,13 @@ All notable changes to this project will be documented in this file.  The format
 [comment]: <> (Security:   in case of vulnerabilities)
 
 
+## [Unreleased]
+
+### Changed
+* Default value for `max_stack_height` is increased to 500.
+
+
+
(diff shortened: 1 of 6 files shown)


rafal-ch (approved):
LGTM :+1:

Fraser999 (commented):
We should probably mention this change in the EE changelog.

Fraser999 (approved):

PR #4042, opened 2023-06-10
finer grained block_synchronizer latching

EdHastingsCasperAssociation:
As discussed with @alsrdn , a finer grained latching mechanism for block_synchronization.

### DIFF
--- a/node/src/components/block_synchronizer.rs
+++ b/node/src/components/block_synchronizer.rs
@@ -296,7 +296,6 @@ impl BlockSynchronizer {
             should_fetch_execution_state,
             self.max_simultaneous_peers,
             self.config.peer_refresh_interval,
-            self.config.latch_reset_interval,
             self.chainspec.core_config.legacy_required_finality,
             self.chainspec
                 .core_config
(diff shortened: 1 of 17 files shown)


node/src/components/block_synchronizer/block_builder/latch.rs
```diff
@@ -0,0 +1,65 @@
...
+use datasize::DataSize;
+
+use tracing::error;
+
+use casper_types::{TimeDiff, Timestamp};
+
+#[derive(Debug, Default, DataSize)]
+pub(super) struct Latch {
+    #[data_size(skip)]
+    latch: AtomicU8,
```
alsrdn: I don't think using an atomic here gives us any benefit since we process events one by one. I do think that we should use an u32 or larger type here since `max_simultaneous_peers` is an u32 and we use it often to determine the number of requests to send out.

node/src/components/block_synchronizer/block_builder/latch.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+impl Latch {
+    pub(super) fn increment(&mut self, increment_by: u8) {
+        match self.latch.get_mut().checked_add(increment_by) {
+            Some(val) => {
+                self.latch.store(val, Ordering::SeqCst);
+            }
+            None => {
+                error!("latch increment overflowed.");
```
alsrdn: nit:
```suggestion
                error!("BlockSynchronizer: latch increment overflowed.");
```

alsrdn: Also what do you think about maybe returning an error here so we can either try to recover or bail on this sync later on?
If the event tracking is messed up by an overflowing counter, there is the risk that the synchronizer could get stuck or act in an undefined way. Wouldn't it be better to put it in a `Failed` state of some sorts or purge it?

node/src/components/block_synchronizer/block_builder/latch.rs
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+        self.touch();
```
alsrdn: Should we actually update the timestamp if the latch failed to increment?

node/src/components/block_synchronizer.rs
```diff
@@ -584,15 +575,15 @@ impl BlockSynchronizer {
...
+                    builder.latch_by(peers.len());
                     results.extend(peers.into_iter().flat_map(|node_id| {
                         effect_builder
                             .fetch::<Block>(block_hash, node_id, Box::new(EmptyValidationMetadata))
                             .event(Event::BlockFetched)
                     }))
                 }
                 NeedNext::FinalitySignatures(block_hash, era_id, validators) => {
-                    builder.set_in_flight_latch();
+                    builder.latch_by(validators.len());
```
alsrdn: this should be `builder.latch_by(min(max_simultaneous_peers, validators.len()));`

node/src/components/block_synchronizer/tests.rs
```diff
@@ -1116,22 +1262,19 @@ async fn fwd_sync_is_not_blocked_by_failed_signatures_fetch_within_latch_interva
...
 
     // The effects are empty at this point and the synchronizer is stuck
     assert!(generated_effects.is_empty());
 
     // Wait for the stall detection time to pass
-    tokio::time::sleep(Duration::from_millis(
-        TEST_SYNCHRONIZER_STALL_LIMIT_MILLIS * 2,
-    ))
-    .await;
+    tokio::time::sleep(Duration::from(cfg.need_next_interval * 2)).await;
```
alsrdn: Hmm, actually effects should not be empty at this point. The test needs to be modified the same as `fwd_sync_is_not_blocked_by_failed_header_fetch_within_latch_interval`.

node/src/components/block_synchronizer/block_builder/latch.rs
EdHastingsCasperAssociation: I would say its the other way, `max_simultaneous_peers` as a u32 is absurdly too many peers. Will change the config to u8.

node/src/components/block_synchronizer/block_builder/latch.rs
EdHastingsCasperAssociation: i think it's unreachable, but even if it is reached its not worth killing the node over, and just allowing the latch to clear naturally will result in the synchronizer recovering and moving on.

node/src/components/block_synchronizer/block_builder/latch.rs
EdHastingsCasperAssociation: will change

node/src/components/block_synchronizer.rs
EdHastingsCasperAssociation: will change

alsrdn: Changed here: https://github.com/EdHastingsCasperLabs/casper-node/pull/9/commits/fa8a791a91fc84dc1e400afd405057f71afd425b

node/src/components/block_synchronizer/tests.rs
alsrdn: Changed here: https://github.com/EdHastingsCasperLabs/casper-node/pull/9/commits/fa8a791a91fc84dc1e400afd405057f71afd425b

alsrdn (approved):
LGTM

PR #4045, opened 2023-06-13
GH-3766: Update wasmi to 0.13.2

### LINKED ISSUE #3766: Need to bump wasmi to 0.13.2
We need to bump wasmi to 0.13.2 (aka "battle tested") as it gives the most performance benefit with little work

linked the related ticket: #3575 

There are just a few minor code changes required, and there are some depreciations introduced. We should collectively review the diff from 0.9.1 to 0.13.2 to ensure we understand all the changes involved since 0.9.

mpapierski:
Closes #3766 

Update wasmi to 0.13.2.

This wasmi version deprecates allocating `MemoryInstance::get` that always allocates and copies data back and forth between VM linear memory and the heap on the host.

With this commit host does not allocate data, but is deserializing data straight from a linear memory without allocating. Copies are made only when absolutely necessary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -705,9 +705,9 @@ dependencies = [
 
 [[package]]
 name = "casper-wasm-utils"
-version = "1.1.0"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "13cd18418b19bc2cbd2bc724cc9050055848e734182e861af43e130a0d442291"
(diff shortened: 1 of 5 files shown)


EdHastingsCasperAssociation:
do you have any comparative benchmarks, before & after?

EdHastingsCasperAssociation (approved):
id like to see perf & resource consumption numbers, but I like it.

mpapierski:
As discussed initial results as measured with coremark benchmark with all the processed code has about 70% higher score. There are more stories in the backlog to create more detailed measurements of the execution.

PR #4053, opened 2023-06-14
Move `Deploy` and its associated types to `casper-types` crate

### LINKED ISSUE #2138: Move deploy types to casper-types

Fraser999:
This PR moves the `Deploy` type and its associated types to `casper-types` crate.

Primarily the types involved lived in `casper-node` previously, but `ExecutableDeployItem` was moved from `casper-execution-engine`.

Existing types (e.g. `Digest`) were updated to be no-std compatible as far as possible in order to support no-std for as much of the code being moved into `casper-types` as possible.

There was also a small change made in [the first commit](https://github.com/casper-network/casper-node/commit/a608f99d7fa8491607c7659785138aeba759c881) to allow docs to be built with the [`doc_auto_cfg` feature](https://doc.rust-lang.org/rustdoc/unstable-features.html#doc_auto_cfg-automatically-generate-doccfg) enabled.  This allows more comprehensive docs to be generated, including those created automatically on docs.rs.  To view an example, run
```
RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc -p casper-types --all-features --no-deps --open
```

I also copied over the `DeployBuilder` from casper-client in order to allow the client to make better use of this changeset.

Closes #2138.


types/src/deploy.rs
```diff
@@ -0,0 +1,1780 @@
...
+    /// Adds a signature of this `Deploy`'s hash to its approvals.
+    pub fn sign(&mut self, secret_key: &SecretKey) {
+        let approval = Approval::create(&self.hash, secret_key);
+        self.approvals.insert(approval);
+    }
+
+    /// Returns the `ApprovalsHash` of this `Deploy`'s approvals.
+    pub fn compute_approvals_hash(&self) -> Result<ApprovalsHash, bytesrepr::Error> {
+        ApprovalsHash::compute(&self.approvals)
+    }
```
EdHastingsCasperAssociation, line 268: does it make sense to add a OnceCell non-serialized field similar to is_valid to avoid computing this more than once?

types/src/deploy/approvals_hash.rs
```diff
@@ -1,53 +1,51 @@
...
+
     /// Constructs a new `ApprovalsHash` by bytesrepr-encoding `approvals` and creating a [`Digest`]
     /// of this.
     pub fn compute(approvals: &BTreeSet<Approval>) -> Result<Self, bytesrepr::Error> {
         let digest = Digest::hash(approvals.to_bytes()?);
         Ok(ApprovalsHash(digest))
     }
 
-    /// Returns the wrapped inner hash digest.
+    /// Returns the wrapped inner digest.
```
EdHastingsCasperAssociation, line 37: separate ticket, but I think we should consider making specific newtypes for all of these various naked Digests which are not really interchangeable, like we discussed for ChainspecHash.

types/src/deploy/id.rs
```diff
@@ -1,39 +1,32 @@
...
-    Serialize,
-    Deserialize,
-    Debug,
-    Default,
+    Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Serialize, Deserialize, Debug, Default,
 )]
+#[cfg_attr(feature = "datasize", derive(DataSize))]
 #[serde(deny_unknown_fields)]
-pub struct Id {
+pub struct DeployId {
```
EdHastingsCasperAssociation: I like it. Should we rename the file to match (deploy_id.rs).

EdHastingsCasperAssociation (approved):

types/src/deploy.rs
Fraser999: I'll do a ticket request to investigate.  If this is only called once or twice per deploy, it might not be worthwhile.

types/src/deploy/id.rs
Fraser999: Done in [2fa6f99](https://github.com/casper-network/casper-node/pull/4053/commits/2fa6f99034f43c4d559f9bcc118c18144bb82129) (along with a few similar others).

Fraser999:
bors merge

types/src/deploy.rs
EdHastingsCasperAssociation: That's sensible. :+1:

PR #4058, opened 2023-06-15
Purge block synchronizer every time we transition out of `CatchUp`

### LINKED ISSUE #4043: Investigate failure in emergency_upgrade_test_balances.sh
Nodes are expected to have no errors in logs, yet two of them reported the "historical block should not be executed".
Link to the failed build: https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8466/1/4.

rafal-ch:
The chain of events that lead to the error is:
1. Node restart for the emergency upgrade, chain is rewind to block 22 (switch block of Era 2)
2. Immediate switch block 23 is created, storage marks it complete
3. Node switches to CatchUp - and the sync identifier is SyncedBlockIdentifier(block 22)
4. MetaBlock marks the block 23 complete

Thanks to the added purge (in this case, the purge when switching from `CatchUp` to `Upgrading` is important), it is expected that `catch_up_process()` realizes that sync progress is `Idle` and will set-up sync leap with `BlockIdentifier` or `LocalTip`, instead of `SyncedBlockIdentifier(block 22)` as above.

Closes #4043

### DIFF
--- a/node/src/reactor/main_reactor/control.rs
+++ b/node/src/reactor/main_reactor/control.rs
@@ -96,6 +96,7 @@ impl MainReactor {
                 CatchUpInstruction::CommitGenesis => match self.commit_genesis(effect_builder) {
                     GenesisInstruction::Validator(duration, effects) => {
                         info!("CatchUp: switch to Validate at genesis");
+                        self.block_synchronizer.purge();
                         self.state = ReactorState::Validate;
                         (duration, effects)
                     }
@@ -112,6 +113,7 @@ impl MainReactor {
                 CatchUpInstruction::CommitUpgrade => match self.commit_upgrade(effect_builder) {
                     Ok(effects) => {
                         info!("CatchUp: switch to Upgrading");
+                        self.block_synchronizer.purge();
                         self.state = ReactorState::Upgrading;
                         self.last_progress = Timestamp::now();
                         self.attempts = 0;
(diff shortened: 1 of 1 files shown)


rafal-ch:
bors try

rafal-ch:
bors try

Fraser999 (approved):

fizyk20 (approved):

node/src/reactor/main_reactor/control.rs
```diff
@@ -96,6 +96,7 @@ impl MainReactor {
                 CatchUpInstruction::CommitGenesis => match self.commit_genesis(effect_builder) {
                     GenesisInstruction::Validator(duration, effects) => {
                         info!("CatchUp: switch to Validate at genesis");
+                        self.block_synchronizer.purge();
```
EdHastingsCasperAssociation, line 99: this is harmless, but shouldn't the synchronizer not be in use yet in this flow?

EdHastingsCasperAssociation (approved):

alsrdn (approved):

node/src/reactor/main_reactor/control.rs
rafal-ch: It shoudn't, indeed. I wanted to express the "purge whenever we leave CatchUp" scheme.

alsrdn: This way we ensure the CatchUp state is clean for the time we need to switch back to it.

PR #4068, opened 2023-06-19
Make git hash an optional part of the node build version string

### LINKED ISSUE #4060: Ensure the build script doesn't fail if built outside of a git context
Building a project which depends upon casper-node from crates.io fails to build since the node's build script expects the project to be in a git repo.  We should not use expect in the build script, and in node's fn version_string in lib.rs we should not use env!("VERGEN_GIT_SHA_SHORT"), but rather something like if let Ok(git_sha) = env::var("VERGEN_GIT_SHA_SHORT") to optionally append the git hash to the version string.

Fraser999:
This PR updates the node's build script to not require building from inside a git repo.

In the case of building outside a git repo, no git short hash is appended to the build version string, and a warning is emitted during the build and to the log just after running the node.

Closes #4060.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,6 +58,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "ahash"
+version = "0.8.3"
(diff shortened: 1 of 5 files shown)


rafal-ch (approved):
LGTM :+1:

alsrdn (approved):

Fraser999:
bors merge

iyanmv:
This fix is missing in v1.5.3

PR #4073, opened 2023-06-21
Change the reward scheme to a blocks window instead of a delay

### LINKED ISSUE #4065: Update the past_finality_signatures field in BlockBody in accordance with the modified rewards scheme
The field currently describes finality signatures from a single past block. With the modified scheme, it's supposed to contain signatures from up to `rewards_lag` past blocks. We need to change the structure of the field to accommodate that, and modify the field population logic accordingly.
Also, find better names for `past_finality_signatures` and `rewards_lag`.

real-felix:
Fixes #4065

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -23,7 +23,7 @@ use std::{
 
 use anyhow::Error;
 use datasize::DataSize;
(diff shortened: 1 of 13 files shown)


AlexanderLimonov (commented):
See Slack regarding tests

PR #4075, opened 2023-06-21
consensus: don't error when deactivating era more than once

### LINKED ISSUE #4070: Investigate the failure of `bond_its.sh` test
This [test failed](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8525/1/4) in one of the nightly test round (single observation). Validator was evicted, but it's been proposing even after an "auction delay". It is worth checking if all the assertions in the tests are written correctly.

alsrdn:
Don't return a fatal error when trying to deactivate an era that was already deactivated.
In the case where a node unbonds and drops from the validator list, it will switch to `KeepUp` in order to follow the chain. Switching to `KeepUp` will try to deactivate the current era which would have been already deactivated when the last block of the era was added. Print out a debug message to track when this happens.

Fixes: https://github.com/casper-network/casper-node/issues/4070

### DIFF
--- a/node/src/components/consensus/era_supervisor.rs
+++ b/node/src/components/consensus/era_supervisor.rs
@@ -860,7 +860,8 @@ impl EraSupervisor {
             .ok_or_else(|| "attempt to deactivate an era with no eras instantiated!".to_string())?;
         let era = self.era_mut(which_era);
         if false == era.consensus.is_active() {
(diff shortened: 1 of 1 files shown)


fizyk20 (approved):
I thought there was more that would require changing, but this indeed seems to be everything :smile:

rafal-ch (approved):
LGTM :+1:

alsrdn:
bors try

PR #4079, opened 2023-06-21
Fix serialization issues

### LINKED ISSUE #4074: Fix failing upgrade test on feat-2.0
Since merging https://github.com/casper-network/casper-node/pull/4053, running bash -i ci/nctl_upgrade.sh fails on the first test.  The cause appears to be related to a mismatch between the bincode-encoded forms of execution results between release-1.4.13 and feat-2.0 as the nodes after upgrade are unable to parse execution results from storage.

Fraser999:
This PR fixes two issues caused by the recent move of deploy types to casper-types.

As part of that PR, two custom serializers in the `ExecutableDeployItem` enum were removed which shouldn't have been, meaning that parsing such an item which was stored e.g. by a 1.4.x node would fail.

A further issue was that binary serialization of `casper-node::types::DeployHash` delegated to `Digest`, which delegated to `Vec<u8>` for legacy reasons.  When bincode-encoding such a `DeployHash` it has a length prepended to the serialized data.  This is still how the moved `DeployHash` is handled.  However, prior to the move, there was a `casper_types::DeployHash` which delegated to `[u8; 32]` for binary serialization.  Bincode does not prepend a length for a fixed-length array like this.  As such, to retain the ability to parse historical stored data which held an old `casper_types::DeployHash`, we need to provide helpers to use array-style encoding rather than Vec.

Closes #4074.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3094,9 +3094,9 @@ checksum = "624a8340c38c1b80fd549087862da4ba43e08858af025b236e509b6649fc13d5"
 
 [[package]]
 name = "openssl"
-version = "0.10.54"
+version = "0.10.55"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "69b3f656a17a6cbc115b5c7a40c616947d213ba182135b014d6051b73ab6f019"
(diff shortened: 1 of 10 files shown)


rafal-ch (approved):
LGTM :+1:

Fraser999:
bors merge

PR #4081, opened 2023-06-22
Historical SyncLeap for era 0 will now contain single signed block header

### LINKED ISSUE #4077: Investigate stuck historical sync on large network test
Investigate stuck historical sync issue reported here: https://casper-labs-team.slack.com/archives/C01ULKL8G8J/p1687350784393109

rafal-ch:
Node will get stuck on historical sync when it is syncing to genesis through era 0 of the pre 1.5 chain and at the same time it is restarted for upgrade.

We attempt to fix that by extending the response to the historical SyncLeap request. In case the trusted block is in era 0, despite `trusted_ancestor_only` being set to `true`, the response will contain a single signed block header (switch block for era 0).

Details: https://github.com/casper-network/casper-node/issues/4077#issuecomment-1602475466

Closes https://github.com/casper-network/casper-node/issues/4077

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -1972,6 +1972,7 @@ impl Storage {
         txn: &mut Tx,
         trusted_block_header: &BlockHeader,
         highest_signed_block_header: &BlockHeaderWithMetadata,
+        first_one_only: bool,
     ) -> Result<Option<Vec<BlockHeaderWithMetadata>>, FatalStorageError> {
         if trusted_block_header.block_hash()
             == highest_signed_block_header.block_header.block_hash()
@@ -1993,6 +1994,10 @@ impl Storage {
             match self.get_single_block_header_with_metadata(txn, hash)? {
                 Some(block) => result.push(block),
                 None => return Ok(None),
+            }
+
+            if first_one_only {
+                return Ok(Some(result));
             }
         }
         result.push(highest_signed_block_header.clone());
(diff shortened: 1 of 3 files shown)


rafal-ch:
bors try

node/src/components/storage.rs
```diff
@@ -2321,8 +2326,33 @@ impl Storage {
...
+                        trusted_ancestor_only: true,
+                        trusted_block_header,
+                        trusted_ancestor_headers,
+                        signed_block_headers,
+                    }));
+                } else {
+                    warn!("unable to provide validator weights for era 1 to the requesting node")
+                }
+            }
+
```
alsrdn, line 2355: nit: just to avoid duplicating the return statement we can do:
```suggestion
            if trusted_block_header.era_id().is_genesis() {
                let signed_block_headers = if let Some(signed_block_headers) = self.get_signed_block_headers(
                    &mut txn,
                    &trusted_block_header,
                    &highest_complete_block_header,
                    true,
                )? {
                    debug_assert!(signed_block_headers.len() <= 1);
                    signed_block_headers
                } else {
                    warn!("unable to provide validator weights for era 1 to the requesting node");
                    vec![]
                }
            }
```

alsrdn (approved):
Just one nit but otherwise looks good to me!

node/src/components/storage.rs
```diff
@@ -1972,6 +1972,7 @@ impl Storage {
         txn: &mut Tx,
         trusted_block_header: &BlockHeader,
         highest_signed_block_header: &BlockHeaderWithMetadata,
+        first_one_only: bool,
```
EdHastingsCasperAssociation, line 1975: My hot take is this arg should not be unnecessary; I would think that the fn body could be rewritten to handle this situation insitu without recursion.

PR #4086, opened 2023-06-27
node/rpc-server/query_global_state: make state_identifier optional

### LINKED ISSUE #4054: Make state_identifier optional in QueryGlobalStateParams
The JSON-RPC  query_global_state should be changed to allow no state identifier being provided, in the same manner as for query_balance, so that if it is None then the latest block should be used.  We believe this won't be a breaking change, but that should be confirmed.
Release: whenever ready

alsrdn:
Make the `state_identifier` parameter of the `query_global_state` JSON-RPC method optional. If no `state_identifier` is specified, the highest complete block is used as an identifier.

Fixes: https://github.com/casper-network/casper-node/issues/4054

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.  The format
 ### Fixed
 * Now possible to build outside a git repository context (e.g. from a source tarball). In such cases, the node's build version (as reported vie status endpoints) will not contain a trailing git short hash.
 
+### Changed
+* The `state_identifier` parameter of the `query_global_state` JSON-RPC method is now optional. If no `state_identifier` is specified, the highest block known to the node will be used to fulfil the request.
 
 
 ## 1.5.1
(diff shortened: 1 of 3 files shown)


rafal-ch (approved):
LGTM :+1: 

Also checked with `rpc_query_global_state_get_balance` test from `casper-test` and still passes :)

Also, not a breaking change imho, but maybe worth confirming with @Fraser999 before merge.

node/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ All notable changes to this project will be documented in this file.  The format
 ### Fixed
 * Now possible to build outside a git repository context (e.g. from a source tarball). In such cases, the node's build version (as reported vie status endpoints) will not contain a trailing git short hash.
 
+### Changed
+* The `state_identifier` parameter of the `query_global_state` JSON-RPC method is now optional. If no `state_identifier` is specified, the highest block known to the node will be used to fulfil the request.
```
Fraser999:
```suggestion
* The `state_identifier` parameter of the `query_global_state` JSON-RPC method is now optional. If no `state_identifier` is specified, the highest complete block known to the node will be used to fulfill the request.
```

Fraser999 (approved):
LGTM barring one nitpick.

node/CHANGELOG.md
alsrdn: Fixed in https://github.com/casper-network/casper-node/pull/4086/commits/8e56abc611433cd59134e5374b02cc7d4be77ed2

PR #4092, opened 2023-06-28
Clean up EE types

### LINKED ISSUE #2760: Move `macro_rules! make_array_newtype` into `ee_699.rs`

### LINKED ISSUE #4064: remove the correlation_id concept from the Execution Engine
`correlation_id: CorrelationId
`It was added during the time when the scala node called the ee's api at various points and there was a desire to be able to associate a correlating value from the caller side to keep track of activity across many calls. However, it serves no purpose in our unified architecture. We have three basic choices, as I see it:
leave it alone; its vestigial but basically harmless to our current usage
change the node to actually generate and reuse a specific correlation id across multiple ee calls, ie actually use it for its original purpose
get rid of it, removing the argument from all the ee api methods that have it, the type, and the call sites.
(edited)
at some point, as memory serves, Dan and Matt had actually removed it in a 1.x related branch
but we could not move forward w/ that due to breaking change concerns
but we can now if we still want to

Fraser999:
This PR cleans up some of the EE types, including those which were split to the new `casper-storage` crate.

For the most part, types were removed or their scope reduced, allowing the flattening of a couple of modules.

In EE:
* `Config` removed as unused
* `shared::logging` removed as unused (engine test support crate now uses `env-logger`)
* `shared::test_utils` removed as unused
* `shared::utils` removed (only OS page size functionality retained, but moved to `casper-types`)
* `shared::wasm_prep` moved to `runtime` and its functions' scopes reduced
* `shared::execution_journal` moved to `engine_state`
* `shared` removed
* `core` removed and its modules raised a level

In EE tests:
* `logging` removed as unused
* `profiling` folder containing source for three binaries removed along with a single smart contract associated with them.  These tools are no longer useful and don't currently run successfully.

In storage:
* `global_state::shared::make_array_newtype!` moved to the single test where it is used
* `global_state::shared::correlation_id` removed as previously agreed as unnecessary
* `global_state::storage` removed and its modules raised a level

A future PR will address moving `ExecutionJournal` and its nested types into `casper-types`.  This will allow the removal of `storage::global_state::shared` entirely.

Closes [#153](https://github.com/casper-network/roadmap/issues/153)
Closes #2760.
Closes #4064.


dwerner (approved):
lgtm

Fraser999:
bors merge

PR #4098, opened 2023-06-30
Fix failing gh_3710 test.

### LINKED ISSUE #4096: Investigate gh_3710 EE test failure on feat-1.6
Description: We found out feat-1.6 had another test failure in a test test::regression::gh_3710::gh_3710_commit_prune_should_delete_values . We should fix it.

mpapierski:
Closes #4096 

The issue was that `NonDeserializingStore` did not forward all the methods from base `Store` trait, and in result, ScratchTrie was misbehaving due to that.


alsrdn (approved):
LGTM

rafal-ch (approved):
LGTM :+1:

PR #4102, opened 2023-07-03
Print `current stack height` when `TrapCode::Unreachable` is encountered

### LINKED ISSUE #4003: Print current stack height after receiving Unreachable trap when test-support is enabled

rafal-ch:
When an error is encountered during contract execution detailed debug into will be written to `stderr`, for example:

```
Wasm execution error: Trap: Host(InvalidContext)
current_wasm_instance_stack_height=44
current_runtime_stack_height=2
max_runtime_stack_height=12

--- frame 1 ---
StoredSession
    account_hash=58b891759929bd4ed5a9cce20b9d6e3c96a66c21386bed96040e17dd07b79fa7
    contract_hash=ca887c8abe4c7a9ee6ab4d4e821b5aa6e21ba3f8b1e7ca9035d0f56ffb7385ca
    contract_package_hash=d4c13d8b05b4dab4c7384b35fa5a2cd3514979d60125058ee26bcb74cf2dd210

--- frame 0 ---
Session
    account_hash=58b891759929bd4ed5a9cce20b9d6e3c96a66c21386bed96040e17dd07b79fa7
```

```
Wasm execution error: Trap: Code(Unreachable)
current_wasm_instance_stack_height=4
current_runtime_stack_height=2
max_runtime_stack_height=12

--- frame 1 ---
StoredContract
    contract_hash=707b0a2da47d6da254b952371a1a4b571a9e98b67d5cb1f36f08875d8a96a0e8
    contract_package_hash=406ab318520b3e6cbd2d6eedf233035e70d223c1314b9c05221ba36a106fa6ba

--- frame 0 ---
Session
    account_hash=58b891759929bd4ed5a9cce20b9d6e3c96a66c21386bed96040e17dd07b79fa7
```

```
Wasm execution error: Trap: Host(Revert(ApiError::AuctionError(DelegatorFundsLocked) [64550]))
current_wasm_instance_stack_height=33
current_runtime_stack_height=1
max_runtime_stack_height=12

--- frame 0 ---
Session
    account_hash=a02f3ede65d5e8bd96259b721236b121a9de73caca1cba340faace546869d398
```

```
Wasm execution error: Trap: Host(InvalidContext)
current_wasm_instance_stack_height=44
current_runtime_stack_height=3
max_runtime_stack_height=12

--- frame 2 ---
StoredContract
    contract_hash=623ed827aabb04335570a9487bd7c8d656817f3c0b0ad27e1369019edfb218a3
    contract_package_hash=c0e00ead5383b16af0e988dc03931b71da3f7d7ed2acce4e17a42162b1407bd3

--- frame 1 ---
StoredContract
    contract_hash=623ed827aabb04335570a9487bd7c8d656817f3c0b0ad27e1369019edfb218a3
    contract_package_hash=c0e00ead5383b16af0e988dc03931b71da3f7d7ed2acce4e17a42162b1407bd3

--- frame 0 ---
Session
    account_hash=58b891759929bd4ed5a9cce20b9d6e3c96a66c21386bed96040e17dd07b79fa7
```

Closes #4003

### DIFF
--- a/execution_engine/src/core/runtime/mod.rs
+++ b/execution_engine/src/core/runtime/mod.rs
@@ -1081,13 +1081,7 @@ where
         };
 
         #[cfg(feature = "test-support")]
-        if let wasmi::Error::Trap(Trap::Code(TrapCode::Unreachable)) = &error {
-            if let Some(stack_height) = instance.globals().last() {
-                eprintln!("current stack height = {}", stack_height.get())
-            } else {
-                eprintln!("current stack height unknown");
-            }
-        }
+        self.dump_runtime_debug_info(&error, instance);
 
         if let Some(host_error) = error.as_host_error() {
             // If the "error" was in fact a trap caused by calling `ret` then
@@ -1426,13 +1420,7 @@ where
         };
 
         #[cfg(feature = "test-support")]
-        if let wasmi::Error::Trap(Trap::Code(TrapCode::Unreachable)) = &error {
-            if let Some(stack_height) = instance.globals().last() {
-                eprintln!("current stack height = {}", stack_height.get())
-            } else {
-                eprintln!("current stack height unknown");
-            }
-        }
+        self.dump_runtime_debug_info(&error, instance);
 
         if let Some(host_error) = error.as_host_error() {
             // If the "error" was in fact a trap caused by calling `ret` then this is normal
@@ -2970,5 +2958,20 @@ where
         }
 
         Ok(Ok(()))
+    }
+
+    #[cfg(feature = "test-support")]
+    fn dump_runtime_debug_info(&self, error: &wasmi::Error, instance: wasmi::ModuleRef) {
+        eprintln!("Wasm execution error: {}", error);
+        if let Some(stack_height) = instance.globals().last() {
+            eprintln!("current_wasm_instance_stack_height={}", stack_height.get())
+        } else {
+            eprintln!("current_wasm_instance_stack_height=unknown");
+        }
+        if let Some(stack) = &self.stack {
+            eprintln!("{}", stack);
+        } else {
+            eprintln!("runtime_stack=unknown");
+        }
     }
 }
--- a/execution_engine/src/core/runtime/mod.rs
+++ b/execution_engine/src/core/runtime/mod.rs
@@ -1080,6 +1080,14 @@ where
             }
         };
 
+        if let wasmi::Error::Trap(Trap::Code(TrapCode::Unreachable)) = &error {
+            if let Some(stack_height) = instance.globals().last() {
+                eprintln!("current stack height = {}", stack_height.get())
+            } else {
+                eprintln!("current stack height unknown");
+            }
+        }
+
         if let Some(host_error) = error.as_host_error() {
             // If the "error" was in fact a trap caused by calling `ret` then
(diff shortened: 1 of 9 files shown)


execution_engine/src/core/runtime/mod.rs
```diff
@@ -1080,6 +1080,15 @@ where
             }
         };
 
+        #[cfg(feature = "test-support")]
+        if let wasmi::Error::Trap(Trap::Code(TrapCode::Unreachable)) = &error {
```
mpapierski: Since we're here already how about putting more details from `self.frame` i.e.:

```
Wasm execution error {wasm trap (i.e. Unreachable) or a specific host error (i.e. Revert(1234))}:
  0. Session { account_hash: ... }
  1. StoredSEssion { ... }
  2. ...

Current Wasm instance stack height {value of stack_height global variable} <- print this in case of unreachable trap only
```

rafal-ch: More detailed debug info added in 8ff446d6f66cc94f5f21e613bdeef152f76487b3.
However, for consistency, I print `current_wasm_instance_stack_height` for every error.

rafal-ch: @mpapierski as per discussion, this solution has some drawbacks, in particular:

For recursive calls, as observed in `regression_20211110.rs` test, the **single** `RuntimeStackOverflow` will cause **multiple** error messages to be printed (one for each runtime being unwinded from the stack), like so:
```
Wasm execution error: Trap: Host(RuntimeStackOverflow)
current_wasm_instance_stack_height=29
current_runtime_stack_height=11
max_runtime_stack_height=12

--- frame 10 ---
--- frame 9 ---
...
--- frame 0 ---
[.....more details cut for legibility.....]

Wasm execution error: Trap: Host(RuntimeStackOverflow)
current_wasm_instance_stack_height=29
current_runtime_stack_height=10
max_runtime_stack_height=12

--- frame 9 ---
...
--- frame 0 ---
[.....more details cut for legibility.....]

Wasm execution error: Trap: Host(RuntimeStackOverflow)
current_wasm_instance_stack_height=29
current_runtime_stack_height=9
max_runtime_stack_height=12

--- frame 9 ---
...
--- frame 0 ---
[.....more details cut for legibility.....]
```

Additionally, in the initial implementation which was only referencing the `Trap::Unreachable`, we were able to only analyze the error if the call to `invoke_export` failed, because there was no chance for this error to happen earlier in the flow.

Now, if we're trying to display debug info for other errors, we need to cover more places in `execute_contract()`, for example, we may fail at `stack.push(call_stack_element)?;` even before runtime is created.

It would be best to dump the error info in a single place, probably within `Executor::exec()`, but the problem is that internally the `fn execute_contract()` erases info about `wasmi::Error` and yields it as a String representation `Interpreter(String)`, hence we cannot act on a specific error type upper in the call-stack (unless we want to compare strings, but we don't).

Therefore I will:
* Prepare another PR with just the simple, initial debug info (as described in [the ticket](https://github.com/casper-network/casper-node/issues/4003))
* Close this PR
* Request a ticket to build on top of this closed PR and add the ability to provide users with the more detailed dump as was proposed in the comment [here](https://github.com/casper-network/casper-node/pull/4102#discussion_r1251007300)

PR #4103, opened 2023-07-05
Debug info on contract errors

### LINKED ISSUE #4002: Use runtime::print in panic handler when test-support and debug is enabled
Use runtime::print in panic handler when test-support and debug is enabled
Description: Debugging smart contracts is probably the most difficult when you hit Unreachable trap: could be panic, could be max stack height limit, abort, overflow, unwrap, etc. By adding a runtime::print to a panic handler we can print the actual error message that’s happening inside wasm. This will make people lives easier.

### LINKED ISSUE #4003: Print current stack height after receiving Unreachable trap when test-support is enabled

rafal-ch:
This PR adds some debug info output in case error is encountered during contract execution:

```
running 1 test
panicked at 'Simulating stack height limit', smart_contracts/contracts/test/do-panic/src/main.rs:9:5
Wasm execution error: TrapCode::Unreachable
current instance stack height = 19
test test::panic::panic_in_contract_should_yield_trap_unreachable ... ok
```

Closes #4002
Closes #4003

**Things to be considered**
Change introduced in [this commit](https://github.com/casper-network/casper-node/commit/741dbfda5f0296659e57af8c6e0cce67b53527c8) causes some performance tests to fail:
* `host_function_metrics_has_acceptable_gas_cost` -> `used 399500059000 gas; it should use no more than 364434284571 gas`
* `host_function_metrics_has_acceptable_size` -> `host-function-metrics became 116188 bytes long; up to 102447 bytes long would be acceptable.`
* `faucet_costs` -> `EXPECTED_FAUCET_INSTALL_COST` is `78061867880` instead of `75226865930`

Theoretically, the panic print happens only if `test-support` feature is enabled, so it shouldn't affect any contracts in production - only these under development will be affected. Therefore, we may consider disabling this feature in the above performance tests.

However, we need to be sure that this change will not increase the gas cost in uncontrolled way. If there is a risk, we should remove the print and look for other solution, possibly as a part of ticket [#4104](https://github.com/casper-network/casper-node/issues/4104).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1406,6 +1406,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "do-panic"
+version = "0.1.0"
+dependencies = [
+ "casper-contract",
+]
+
+[[package]]
 name = "downcast-rs"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/execution_engine/CHANGELOG.md
+++ b/execution_engine/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file.  The format
 
 ### Changed
 * Default value for `max_stack_height` is increased to 500.
+* `current stack height` is written to `stderr` in case `Trap(Unreachable)` error is encountered during Wasm execution.
 
 
 
--- a/execution_engine/src/core/runtime/mod.rs
+++ b/execution_engine/src/core/runtime/mod.rs
@@ -1080,6 +1080,14 @@ where
             }
         };
 
+        if let wasmi::Error::Trap(Trap::Code(TrapCode::Unreachable)) = &error {
+            if let Some(stack_height) = instance.globals().last() {
+                eprintln!("current stack height = {}", stack_height.get())
+            } else {
+                eprintln!("current stack height unknown");
+            }
+        }
+
         if let Some(host_error) = error.as_host_error() {
             // If the "error" was in fact a trap caused by calling `ret` then
             // this is normal operation and we should return the value captured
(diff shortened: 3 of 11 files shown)


EdHastingsCasperAssociation (approved):

EdHastingsCasperAssociation:
I would think this could target dev?

rafal-ch:
After discussion with @EdHastingsCasperLabs, the following approach is proposed:
- decouple the performance regression check from tests (**long term**)
-- currently, our tests may fail not because the system is broken, but because the performance changed; these tests will even fail on performance *improvement*
-- our target is for the tests to fail if we broke EE, have some missing FFI, etc. and for the performance to be monitored on CI (there should be a gas cost/file size baseline, and CI should inform us when calculated cost/size of the PR varies too much)
- remove the need to use `print` (especially in panic handler) by improving the error reporting (**long term**)
-- having the conditionally compiled print might be confusing (for example, people may try to chase an error in production, they will enable test-support, which will result in an error in different place, because of wasm structure change)
-- we shall be able to provide more detailed reason of failure (aka, more fine grained error than just Trap::Unreachable) so that users can reason about what happened
- split the `host-function-metric` contract in two (**orthogonal**)
-- one part should just call different FFIs (as in the top of the file)
-- second part should exercise the `extern` with long names

Temporary solution:
- make the tests passing by increasing the allowed limits
- remove the "performance improved" asserts
- schedule for 1.6 (but try to implement the proper, long-term solution also for 1.6, so we don't need to release the temporary one)

cc: @RitaMAllenCA (probably some PM discussion over this is needed)

rafal-ch:
Closed in favor of https://github.com/casper-network/casper-node/pull/4229

PR #4105, opened 2023-07-07
Add cors origin to REST, RPC, ESS and speculative execution servers

### LINKED ISSUE #4085: Make CORS behavior configurable on RPC ports
Currently, all of our RPCs return Access-Control-Allow-Origin: * as a convenience for developers, as they neither protect sensitive data nor know any kind of session data. While we still highly recommend reverse proxying or firewalling these in a production environment, we want to add a quality of life improvement for devs and admins by making the header configurable.
This ticket is about adding one (or more, if necessary) configuration options that allow controlling the node's behavior with regard to the header mentioned above. It should probably be a string value, with an empty string value causing the header to be omitted, with it otherwise being set through this setting. The default should likely be "*" to preserve current behavior.

rafal-ch:
This PR adds the ability of configuring CORS origin for REST, RPC, ESS and speculative execution servers.

It also bumps openssl to 0.10.55 to dodge the vulnerability: https://rustsec.org/advisories/RUSTSEC-2023-0044

Closes #4085.

### DIFF
--- a/node/src/components/rest_server/http_server.rs
+++ b/node/src/components/rest_server/http_server.rs
@@ -1,5 +1,6 @@
 use std::{convert::Infallible, time::Duration};
 
+use casper_json_rpc::CorsOrigin;
 use futures::{future, TryFutureExt};
 use hyper::server::{conn::AddrIncoming, Builder};
 use tokio::sync::oneshot;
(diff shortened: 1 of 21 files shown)


node/src/components/rest_server/http_server.rs
```diff
@@ -30,13 +31,65 @@ pub(super) async fn run<REv: ReactorEventT>(
...
+    let rest_chainspec_filter = filters::create_chainspec_filter(effect_builder, api_version);
+
     let service = warp::service(
         rest_status
             .or(rest_metrics)
             .or(rest_open_rpc)
             .or(rest_validator_changes)
             .or(rest_chainspec_filter)
-            .with(warp::cors().allow_any_origin()),
+            .with(match cors_origin {
```
alsrdn, line 89: We also need to allow the `GET` method here for status and metrics.

alsrdn (changes requested):

node/src/components/rest_server/http_server.rs
alsrdn: Actually I don't think it's needed. I seem to have used the wrong origin when trying out the REST server.

alsrdn (approved):

resources/local/config.toml
```diff
@@ -277,6 +277,13 @@ qps_limit = 100
 # Maximum number of bytes to accept in a single request body.
 max_body_bytes = 2_621_440
 
+# Specifies which origin will be reported as allowed by RPC and REST server.
```
Fraser999: This comment should be specific to each of the 4 servers.

Fraser999 (commented):
We should also update `resources/production/config-example.toml`.

When the config specifies an invalid origin, I see the node just panics without writing an error line to the stdout log.  We could maybe change that behaviour so a fatal error is reported, but perhaps as a follow-up PR?

Otherwise LGTM.

rafal-ch:
> When the config specifies an invalid origin, I see the node just panics without writing an error line to the stdout log. We could maybe change that behaviour so a fatal error is reported, but perhaps as a follow-up PR?

Yes, that's because `allow_origin()` [panics](https://github.com/seanmonstar/warp/blob/master/src/filters/cors.rs#L203-L210) on validation errors. There is an [issue](https://github.com/seanmonstar/warp/issues/564) for that and [an open PR](https://github.com/seanmonstar/warp/pull/769), but inactive for ~2.5 years :(

We can consider contributing a similar thing to `warp` or add our own validation and keep it up to date with how `warp` does validation. The latter may eventually get out of sync, but it should anyway minimize the number of operators being surprised with a not-very explanatory `missing scheme` error.

Edit: alternatively, we can roll-back to the initial (_hacky_) solution, i.e. injecting the headers manually using `.with(warp::reply::with::headers(headers))`

resources/local/config.toml
rafal-ch: Fixed in 27703947ed34b26e1cab818915f02346210e9499

Fraser999 (approved):

Issue #4117, opened 2023-07-11
Amend upgrade logic to No-op if transformation of Withdraw purses in not necessary


RitaMAllenCA:
The current upgrade logic transforms withdraw purses into unbonding purses to support the redelegation feature. However, we should amend the logic to No-Op if the transformation has already occurred. This is the preferred approach as extent networks are on different protocol versions, however once all extent networks undergo the transformation we should remove the logic altogether.

PR #4119, opened 2023-07-11
Amend upgrade logic

### LINKED ISSUE #4118: Amend upgrade logic remove changes for EraSummary
The addition of the new EraSummary key was a one time transformation that has been performed, thus the logic in the upgrade machinery is not needed and can be removed.

darthsiroftardis:
CHANGELOG:

- Tweaked upgrade logic transforming withdraw purses to early exit if possible
- Tweaked upgrade logic to not perform the one-time transformation of the `Key::EraSummary`

Closes #4118 and #4117

### DIFF
--- a/execution_engine/src/core/engine_state/mod.rs
+++ b/execution_engine/src/core/engine_state/mod.rs
@@ -330,6 +330,8 @@ where
             CLValue::from_t(upgrade_config.chainspec_registry().clone())
                 .map_err(|error| Error::Bytesrepr(error.to_string()))?;
 
+        // We write the checksums of the chainspec settings to global state
+        // allowing verification of the reported chainspec data via the RPC.
         tracking_copy.borrow_mut().write(
             Key::ChainspecRegistry,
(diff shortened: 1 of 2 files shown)


execution_engine/src/core/engine_state/mod.rs
```diff
@@ -330,6 +330,8 @@ where
             CLValue::from_t(upgrade_config.chainspec_registry().clone())
                 .map_err(|error| Error::Bytesrepr(error.to_string()))?;
 
+        // We write the checksums of the chainspec settings to global state
+        // allowing verification of the reported chainspec data via the RPC.
```
EdHastingsCasperAssociation: // allowing verification of the chainspec data reported via the RPC.

EdHastingsCasperAssociation (approved):

Fraser999 (approved):

execution_engine/src/core/engine_state/mod.rs
darthsiroftardis: Applied suggestion

PR #4126, opened 2023-07-13
Juliet crate

marc-casperlabs:
This is the first iteration of the `juliet` crate, as described in the [Juliet RFC](https://github.com/marc-casperlabs/juliet-rfc/blob/master/juliet.md).

* Features: Complete in terms of API and implementation for everything that is the minimum for the 1.6 release. One feature is still missing (per-request timeout), to be finished as soon as time permits.
* Examples: One working example, `fizzbuzz`, showing that the core functionality works.
* Bugs: A lot still lurking in edge cases, as the test coverage is bad still.
* Tests: Easy unit tests are complete, but cover none of the actual behavior, just data types at the moment.
* Documentation: Almost complete, should cover everything including usage.

I will amend this PR with additional tests over the next few days, and potential bugfixes arising from them.

I recommend reading the crate documentation (e.g. using `cargo doc --open` locally) to get a good overview first, then refer to the linked RFC.


juliet/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "juliet"
```
Fraser999, line 2: Not sure if we should name this `casper-juliet` for consistency with our other crates in this repo?

juliet/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+version = "0.1.0"
+edition = "2021"
```
Fraser999, line 4: I believe we agreed to stick to `2018` until we move all our crates in this repo over to `2021`.

juliet/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+authors = [ "Marc Brinkmann <marc@casperlabs.io>" ]
+# Ensures we do not pull in all the features of dev dependencies when building.
+# Note: Would have to be moved to workspace root.
+# resolver = "2"
```
Fraser999: Left in by mistake?

Also, would be good to get some more of the usual fields in like `description`, `documentation`, `homepage`, `repository`, `license`.

juliet/README.md
```diff
@@ -0,0 +1,23 @@
+# `juliet` protocol implementation
```
Fraser999:
```suggestion
# Juliet protocol implementation
```

juliet/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This crate implements the juliet multiplexing protocol as laid out in the [Juliet RFC](https://github.com/marc-casperlabs/juliet-rfc/blob/master/juliet.md). It aims to be a secure, simple, easy to verify/review implementation that is still reasonably performant.
```
Fraser999:
```suggestion
This crate implements the Juliet multiplexing protocol as laid out in the [Juliet RFC](https://github.com/marc-casperlabs/juliet-rfc/blob/master/juliet.md). It aims to be a secure, simple, easy to verify/review implementation that is still reasonably performant.
```

juliet/Cargo.toml
Fraser999: Let's also exclude the proptest regressions folder as we do for `casper-node`:

`exclude = ["proptest-regressions"]`

juliet/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This crate's implementation includes benefits such as
+
+* a side-effect free implementation of the Juliet protocol,
+* an `async` IO layer integrated with the [`bytes`](https://docs.rs/bytes) crate to use it, and
+* a type-safe RPC layer built on top.
+
+## Examples
+
+For a quick usage example, see `examples/fizzbuzz.rz`.
```
Fraser999:
```suggestion
For a quick usage example, see [examples/fizzbuzz.rs](examples/fizzbuzz.rs).
```

juliet/examples/fizzbuzz.rs
```diff
@@ -0,0 +1,178 @@
...
+//! ```
+//!
+//! Then, in a second terminal launch the client:
+//!
+//! ```
+//! cargo run --example fizzbuzz --features tracing
+//! ```
+//!
+//! You should [Fizz buzz](https://en.wikipedia.org/wiki/Fizz_buzz) solutions being calculated on
+//! the server side and sent back.
```
Fraser999:
```suggestion
//! You should see [Fizz buzz](https://en.wikipedia.org/wiki/Fizz_buzz) solutions being calculated
//! on the server side and sent back.
```

juliet/README.md
Fraser999, line 1: I think we should update to match all the other published crates in this repo.

juliet/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+* carefully designed with security and DoS resilience as its foremoast goal,
+* customizable frame sizes,
+* up to 256 multiplexed, interleaved channels,
+* backpressure support fully baked in, and
+* low overhead (4 bytes per frame + 1-5 bytes depending on payload length).
+
+This crate's implementation includes benefits such as
+
+* a side-effect free implementation of the Juliet protocol,
```
Fraser999:
```suggestion
* a side-effect-free implementation of the Juliet protocol,
```

juliet/src/lib.rs
```diff
@@ -0,0 +1,250 @@
+#![doc = include_str!("../README.md")]
+
+//!
+//!
+//! ## General usage
+//!
+//! This crate is split into three layers, whose usage depends on an applications specific usecase.
+//! At the very core sits the [`protocol`] module, which is a side-effect free implementation of the
+//! protocol. The caller is responsible for all IO flowing in and out, but it instructed by the
+//! state machine what to do next.
```
Fraser999:
```suggestion
//! This crate is split into three layers, whose usage depends on an application's specific use
//! case. At the very core sits the [`protocol`] module, which is a side-effect-free implementation
//! of the protocol. The caller is responsible for all IO flowing in and out, but it is instructed
//! by the state machine what to do next.
```

juliet/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+//!
+//! If there is no need to roll custom IO, the [`io`] layer provides a complete `tokio`-based
+//! solution that operates on [`tokio::io::AsyncRead`] and [`tokio::io::AsyncWrite`]. It handles
+//! multiplexing input, output and scheduling, as well as buffering messages using a wait and a
+//! ready queue.
+//!
+//! Most users of the library will likely use the highest level layer, [`rpc`] instead. It sits on
+//! top the raw [`io`] layer and wraps all the functionality in safe Rust types, making misuse of
+//! the underlying protocol hard, if not impossible.
+
```
Fraser999, line 27: We should add the standard battery of attributes:
```rust
#![doc(html_root_url = "https://docs.rs/casper-juliet/0.1.0")]
#![doc(
    html_favicon_url = "https://raw.githubusercontent.com/casper-network/casper-node/master/images/CasperLabs_Logo_Favicon_RGB_50px.png",
    html_logo_url = "https://raw.githubusercontent.com/casper-network/casper-node/master/images/CasperLabs_Logo_Symbol_RGB.png",
    test(attr(deny(warnings)))
)]
#![warn(
    missing_docs,
    trivial_casts,
    trivial_numeric_casts,
    unused_qualifications
)]
```

Cargo.toml
Fraser999, line 1: Should be included in the `default-members`.

juliet/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+        Self {
+            request_limit: 1,
+            max_request_payload_size: 0,
+            max_response_payload_size: 0,
+        }
+    }
+}
+
+impl ChannelConfiguration {
+    /// Creates a configuration the given request limit (the default is 1).
```
Fraser999:
```suggestion
    /// Creates a configuration with the given request limit (the default is 1).
```

juliet/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub fn with_request_limit(mut self, request_limit: u16) -> ChannelConfiguration {
+        self.request_limit = request_limit;
+        self
+    }
+
+    /// Creates a configuration the given maximum size for request payloads (the default is 0).
```
Fraser999:
```suggestion
    /// Creates a configuration with the given maximum size for request payloads (the default is 0).
```

Fraser999: Might be nice to add info about the expected behaviour if `0` is used.  I don't know (yet) whether it's equivalent to `u32::MAX`, or if payload size checking is disabled, or if it means messages can't have payloads.

juliet/src/lib.rs
```diff
@@ -0,0 +1,250 @@
...
+    pub fn with_max_request_payload_size(
+        mut self,
+        max_request_payload_size: u32,
+    ) -> ChannelConfiguration {
+        self.max_request_payload_size = max_request_payload_size;
+        self
+    }
+
+    /// Creates a configuration the given maximum size for response payloads (the default is 0).
```
Fraser999: Same as previous comment.

Fraser999 (commented):
Initial review only - I still need to review the bulk of the source code.

We need to add this to the casper-updater tool, but that could probably be done in a separate PR if you'd rather ticket for later.

juliet/Cargo.toml
marc-casperlabs: I'd say the jury is still out whether we break this off as a separate crate - I'd be all for it.

juliet/Cargo.toml
marc-casperlabs: It was the default :) Also, see https://github.com/casper-network/casper-node/pull/4126#discussion_r1266452216 ?

juliet/Cargo.toml
marc-casperlabs: I don't think excluding regressions is better than including it, but I'll defer to your wisdom on this one :)

The `resolver = "2"` was left in on purpose, to strike up this conversation. Without the resolver, all the features enabled by dev-dependencies are also features used when compiling the main crate in release builds. Should we consider moving the `resolver = "2"` configuration to the workspace root to enable it (it is a per-workspace setting)?

marc-casperlabs: 8ceafbb2d fixes the missing regressions exclusion.

juliet/README.md
marc-casperlabs: Truth to be told, I haven't even thought about whether I want the name capitalized or not, but I guess it's best to stick with proper grammar :)

juliet/README.md
marc-casperlabs: I had that originally, but it leads to a dead link in the rustdocs. Are you sure it's preferable?

juliet/README.md
marc-casperlabs: Again referring to https://github.com/casper-network/casper-node/pull/4126#discussion_r1266452216 :)

juliet/src/lib.rs
marc-casperlabs: There's no magic numbers, 0 length payloads means you can have payloads of length 0 (which is technically different from no payloads). I'll update the docs.

juliet/README.md
marc-casperlabs: Fixed in 17bf04ac9

juliet/README.md
marc-casperlabs: Fixed in 17bf04ac9

juliet/examples/fizzbuzz.rs
marc-casperlabs: Fixed in 17bf04ac9.

juliet/README.md
marc-casperlabs: Fixed in 17bf04ac9.

juliet/src/lib.rs
marc-casperlabs: Fixed in 17bf04ac9.

juliet/src/lib.rs
marc-casperlabs: Fixed in 17bf04ac9.

juliet/src/lib.rs
marc-casperlabs: The spelling is fixed in 17bf04ac9.

juliet/src/lib.rs
marc-casperlabs: Spelling fixed in 17bf04ac9.

marc-casperlabs: Extra documentation added in 190b4f5e7.

juliet/src/lib.rs
marc-casperlabs: Extra documentation added in 190b4f5e7.

Cargo.toml
marc-casperlabs: https://github.com/casper-network/casper-node/pull/4126#discussion_r1266452216, yep, we need to have that discussion sooner than I thought :)

juliet/src/lib.rs
marc-casperlabs: I added those except I used the `juliet` crate name for now, and had to remove the `unused_qualifications`, since it does not play nice with the `try_outcome!` macro: b259a2085

juliet/Cargo.toml
Fraser999: You mean repo I guess :)

juliet/Cargo.toml
Fraser999: Sure - happy to wait for that decision.

juliet/Cargo.toml
Fraser999: I'd be fine with adding `resolver = "2"` to the workspace manifest.

juliet/README.md
Fraser999: I don't mind whether it's capitalized or not - I just like consistency :)  I went with what I thought was the more common form in the PR!

juliet/README.md
Fraser999: Ah - because you `include_str!("../README.md")`?  I guess that's a good reason to not do that trick :)  It would be even more difficult to navigate to examples/fizzbuzz.rs from the rustdocs if it's not a hyperlink, since they'd have to know to first go to the repo.

I think it's worth duplicating the text to avoid issues like these, (can happen too if you add code samples to the README for example).

I won't insist though - if you strongly prefer the current approach, I'd just ask that the typo in `.rz` is fixed :)

juliet/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+    ) -> ChannelConfiguration {
+        self.max_request_payload_size = max_request_payload_size;
+        self
+    }
+
+    /// Creates a configuration with the given maximum size for response payloads (default is 0).
+    ///
+    /// There is nothing magical about payload sizes, a size of 0 allows for payloads that are no
+    /// longer than 0 bytes in size. On the protocol level, there is a distinction between a request
+    /// with a zero-sized payload and no payload.
```
Fraser999:
```suggestion
    /// longer than 0 bytes in size. On the protocol level, there is a distinction between a
    /// response with a zero-sized payload and no payload.
```

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
+//! `juliet` header parsing and serialization.
+//!
+//! This module is typically only used by the protocol implementation (see [`crate::protocol`]), but
+//! may be of interested to those writing low level tooling.
```
Fraser999:
```suggestion
//! This module is typically only used by the protocol implementation (see
//! [`protocol`](crate::protocol)), but may be of interested to those writing low level tooling.
```

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+use thiserror::Error;
+
+use crate::{ChannelId, Id};
+
+/// Header structure.
+///
+/// Implements [`AsRef<u8>`], which will return a byte slice with the correct encoding of the header
+/// that can be sent directly to a peer.
+// Note: `[u8; 4]` below should ideally be `[u8; Self::SIZE]`, but this prevents the `Zeroable`
+//       derive from working.
```
Fraser999: It'd be a one-liner to just manually implement `unsafe impl Zeroable for Header {}` wouldn't it?  In fact, it seems like you're not even allowed to override the trait's single method.

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+                self.id()
+            )
+        }
+    }
+}
+
+impl Display for Header {
+    #[inline(always)]
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Debug::fmt(self, f)
```
Fraser999: Since we've derived `Error` for `ErrorKind`, I think we should use that rather than just falling back to `Debug::fmt` here.

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// An invalid header was received.
+    #[error("invalid header")]
+    InvalidHeader = 2,
+    /// A segment was sent with a frame where none was allowed, or a segment was too small or missing.
+    #[error("segment violation")]
+    SegmentViolation = 3,
+    /// A `varint32` could not be decoded.
+    #[error("bad varint")]
+    BadVarInt = 4,
+    /// Invalid channel: A channel number greater or equal the highest channel number was received.
```
Fraser999: Shouldn't it just be "greater than" rather than "greater than or equal to"?

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+    #[error("invalid channel")]
+    InvalidChannel = 5,
+    /// A new request or response was sent without completing the previous one.
+    #[error("multi-frame in progress")]
+    InProgress = 6,
+    /// The indicated size of the response would be exceeded the configured limit.
```
Fraser999:
```suggestion
    /// The indicated size of the response would exceed the configured limit.
```

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+    #[error("response too large")]
+    ResponseTooLarge = 7,
+    /// The indicated size of the request would be exceeded the configured limit.
```
Fraser999:
```suggestion
    /// The indicated size of the request would exceed the configured limit.
```

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+    #[error("request too large")]
+    RequestTooLarge = 8,
+    /// Peer attempted to create two in-flight requests with the same ID on the same channel.
+    #[error("duplicate request")]
+    DuplicateRequest = 9,
+    /// Sent a response for request not in-flight.
+    #[error("response for ficticious request")]
```
Fraser999:
```suggestion
    #[error("response for fictitious request")]
```

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+    FictitiousRequest = 10,
+    /// The dynamic request limit has been exceeded.
+    #[error("request limit exceeded")]
+    RequestLimitExceeded = 11,
+    /// Response cancellation for a request not in-flight.
+    #[error("cancellation for ficticious request")]
```
Fraser999:
```suggestion
    #[error("cancellation for fictitious request")]
```

juliet/README.md
marc-casperlabs: I'm typo-blind, my biggest weakness!

Strongly against duplicating the text, especially since it is bound to change some down the line and I'd hate for that to get out of sync.

The rustdoc links will always be dead, since it can be moved around and docs.rs RS won't have the examples uploaded, I think. The alternative would be linking to an example on the web (i.e. the GitHub repo), which will cause a version mismatch when someone views older docs.

Which one is better?

marc-casperlabs: Typo fixed in b2c6855b3

juliet/Cargo.toml
marc-casperlabs: Fixed in 86b06e0d7

juliet/README.md
Fraser999: > The rustdoc links will always be dead, since it can be moved around and docs.rs RS won't have the examples uploaded, I think.

I normally provide a link to the file in the repo for the rustdocs in a case like this, but I believe it would also appear in the rustdocs source folder (see for example https://docs.rs/crate/casper-types/latest/source which contains the benches and tests folders).

I prefer duplicating (although I agree there's a burden to keep the duplicates synced) because that favours the user over our own convenience.  Providing a link which works in the README and the rustdocs outweighs the pain of having to keep duplicated text in sync IMO.

Either way - I won't press it further.  The main thing is the typo is fixed :D

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// A response with no payload.
+    Response = 1,
+    /// A request that includes a payload.
+    RequestPl = 2,
+    /// A response that includes a payload.
+    ResponsePl = 3,
+    /// Cancellation of a request.
+    CancelReq = 4,
+    /// Cancellation of a response.
+    CancelResp = 5,
```
Fraser999, line 111: Might be worth adding the same hint as for `ErrorKind`?
```rust
    // Note: When adding additional kinds, update the `HIGHEST` associated constant.
```

juliet/src/header.rs
```diff
@@ -0,0 +1,433 @@
...
+            // Would violate validity invariant.
+            _ => unreachable!(),
+        }
+    }
+
+    /// Returns the frame kind.
+    ///
+    /// # Panics
+    ///
+    /// Will panic if `Self::is_error()` is not `false`.
```
Fraser999:
```suggestion
    /// Will panic if `Self::is_error()` is `true`.
```

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
+//! `juliet` IO layer
+//!
+//! The IO layer combines a lower-level transport like a TCP Stream with the
+//! [`JulietProtocol`](crate::protocol::JulietProtocol) protocol implementation and some memory
+//! buffers to provide a working high-level transport for juliet messages. It allows users of this
+//! layer to send messages across over multiple channels, without having to worry about frame
+//! multiplexing or request limits.
```
Fraser999:
```suggestion
//! layer to send messages over multiple channels, without having to worry about frame multiplexing
//! or request limits.
```

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
...
+//!
+//! ## Usage
+//!
+//! Most, if not all functionality is provided by the [`IoCore<N, R, W>`] type, which constructed
```
Fraser999:
```suggestion
//! Most, if not all functionality is provided by the [`IoCore<N, R, W>`] type, which is constructed
```

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
...
+    },
+    /// The remote peer violated the protocol and has been sent an error.
+    #[error("error sent to peer")]
+    RemoteProtocolViolation(OutgoingFrame),
+    #[error("local protocol violation")]
+    /// Local protocol violation - caller violated the crate's API.
+    LocalProtocolViolation(#[from] LocalProtocolViolation),
+    /// Internal error.
+    ///
+    /// An error occured that should be impossible, this is indicative of a bug in this library.
```
Fraser999:
```suggestion
    /// An error occurred that should be impossible, this is indicative of a bug in this library.
```

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
...
+#[derive(Clone, Debug)]
+pub struct RequestHandle<const N: usize> {
+    /// Shared portion of the [`IoCore`], required for backpressuring onto clients.
+    shared: Arc<IoShared<N>>,
+    /// Sender for queue items.
+    sender: UnboundedSender<QueuedItem>,
+    /// The next generation [`IoId`].
+    ///
+    /// IoIDs are just generated sequentially until they run out (which at 1 billion at second takes
+    /// roughly 10^22 years).
```
Fraser999: Do we really need a `u128` here?  Seems like a `u64` taking a piffling 584 years would still be enough :)

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
...
+///
+/// Request layer IO IDs are unique across the program per request that originated from the local
+/// endpoint. They are used to allow for buffering large numbers of items without exhausting the
+/// pool of protocol level request IDs, which are limited to `u16`s.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub struct IoId(u128);
+
+/// IO layer for the juliet protocol.
+///
+/// The central structure for the IO layer built on top the juliet protocol, once instance per
```
Fraser999:
```suggestion
/// The central structure for the IO layer built on top of the juliet protocol, one instance per
```

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
...
+/// connection. It manages incoming (`R`) and outgoing (`W`) transports, as well as a queue for
+/// items to be sent.
+///
+/// Once instantiated, a continuously polling of [`IoCore::next_event`] is expected.
```
Fraser999:
```suggestion
/// Once instantiated, a continuous polling of [`IoCore::next_event`] is expected.
```

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
...
+impl<const N: usize, R, W> IoCore<N, R, W>
+where
+    R: AsyncRead + Unpin,
+    W: AsyncWrite + Unpin,
+{
+    /// Retrieve the next event.
+    ///
+    /// This is the central loop of the IO layer. It polls all underlying transports and reads/write
+    /// if data is available, until enough processing has been done to produce an [`IoEvent`]. Thus
+    /// any application using the IO layer should loop over calling this function.
```
Fraser999:
```suggestion
    /// Retrieves the next event.
    ///
    /// This is the central loop of the IO layer. It polls all underlying transports and
    /// reads/writes if data is available, until enough processing has been done to produce an
    /// [`IoEvent`]. Thus any application using the IO layer should loop over calling this function.
```

Cargo.toml
marc-casperlabs: a3f821553 includes it

juliet/README.md
marc-casperlabs: Will revisit once the decision is made.

juliet/Cargo.toml
marc-casperlabs: Will revisit once the decision is made.

juliet/Cargo.toml
marc-casperlabs: Will revisit once the decision is made.

juliet/src/io.rs
```diff
@@ -0,0 +1,992 @@
...
+        &self,
+        channel: ChannelId,
+        id: Id,
+    ) -> Result<(), EnqueueError> {
+        self.sender
+            .send(QueuedItem::ResponseCancellation { id, channel })
+            .map_err(|send_err| EnqueueError::Closed(send_err.0.into_payload()))
+    }
+
+    /// Enqueus an error.
```
Fraser999:
```suggestion
    /// Enqueues an error.
```

juliet/src/protocol/multiframe.rs
```diff
@@ -0,0 +1,196 @@
+//! Multiframe reading support.
+//!
+//! The juliet protocol supports multi-frame messages, which are subject to addtional rules and
```
Fraser999:
```suggestion
//! The juliet protocol supports multi-frame messages, which are subject to additional rules and
```

juliet/src/protocol/multiframe.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+impl MultiframeReceiver {
+    /// Attempt to process a single multi-frame message frame.
+    ///
+    /// The caller MUST only call this method if it has determined that the frame in `buffer` is one
+    /// that includes a payload. If this is the case, the entire receive `buffer` should be passed
+    /// to this function.
+    ///
+    /// If a message payload matching the given header has been succesfully completed, both header
```
Fraser999:
```suggestion
    /// If a message payload matching the given header has been successfully completed, both header
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+//! parameter `N` denotes the number of valid channels, which must be set at compile time. See the
+//! types documentation for more details.
+//!
+//! ## Efficiency
+//!
+//! In general, all bulky data used in the protocol is as zero-copy as possible, for example large
+//! messages going out in multiple frames will still share the one original payload buffer passed in
+//! at construction. The "exception" to this is the re-assembly of multi-frame messages, which
+//! causes fragments to be copied once to form a continguous byte sequence for the payload to avoid
+//! memory-exhaustion attacks based on the semtantics of the underlying [`bytes::BytesMut`].
```
Fraser999:
```suggestion
//! type's documentation for more details.
//!
//! ## Efficiency
//!
//! In general, all bulky data used in the protocol is as zero-copy as possible, for example large
//! messages going out in multiple frames will still share the one original payload buffer passed in
//! at construction. The "exception" to this is the re-assembly of multi-frame messages, which
//! causes fragments to be copied once to form a contiguous byte sequence for the payload to avoid
//! memory-exhaustion attacks based on the semantics of the underlying [`bytes::BytesMut`].
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+use crate::{
+    header::{self, ErrorKind, Header, Kind},
+    try_outcome,
+    util::Index,
+    varint::{decode_varint32, Varint32},
+    ChannelConfiguration, ChannelId, Id,
+    Outcome::{self, Fatal, Incomplete, Success},
+};
+
+/// A channel ID to fill in when the channel is actually or not relevant unknown.
```
Fraser999:
```suggestion
/// A channel ID to fill in when the channel is actually unknown or not relevant.
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+///
+/// Note that this is not a reserved channel, just a default chosen -- it may clash with an
+/// actually active channel.
+const UNKNOWN_CHANNEL: ChannelId = ChannelId::new(0);
+
+/// An ID to fill in when the ID should not matter.
+///
+/// Note a reserved id, it may clash with existing ones.
```
Fraser999:
```suggestion
/// Not a reserved id, it may clash with existing ones.
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+            channels,
+            max_frame_size: self.max_frame_size,
+        }
+    }
+
+    /// Sets the maximum frame size.
+    ///
+    /// # Panics
+    ///
+    /// Will panic if the maximum size is too small to holder a header, payload length and at least
```
Fraser999:
```suggestion
    /// Will panic if the maximum size is too small to hold a header, payload length and at least
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+    /// A set of request IDs from requests received that have not been answered with a response or
+    /// cancellation yet.
+    incoming_requests: HashSet<Id>,
+    /// A set of request IDs for requests made for which no response or cancellation has been
+    /// received yet.
+    outgoing_requests: HashSet<Id>,
+    /// The multiframe receiver state machine.
+    ///
+    /// Every channel allows for at most one multi-frame message to be in progress at the same time.
+    current_multiframe_receive: MultiframeReceiver,
```
Fraser999:
```suggestion
    current_multiframe_receiver: MultiframeReceiver,
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+            config,
+            prev_request_id: 0,
+        }
+    }
+
+    /// Returns whether or not the peer has exhausted the number of requests allowed.
+    ///
+    /// Depending on the size of the payload an [`OutgoingMessage`] may span multiple frames. On a
+    /// single channel, only one multi-frame message may be in the process of sending at a time,
+    /// thus it is not permissable to begin sending frames of a different multi-frame message before
```
Fraser999:
```suggestion
    /// thus it is not permissible to begin sending frames of a different multi-frame message before
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+    /// the send of a previous one has been completed.
+    ///
+    /// Additional single-frame messages can be interspersed in between at will.
+    ///
+    /// [`JulietProtocol`] does not track whether or not a multi-channel message is in-flight; it is
```
Fraser999:
```suggestion
    /// [`JulietProtocol`] does not track whether or not a multi-frame message is in-flight; it is
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+/// A correct implementation of a client should never encounter this, thus simply unwrapping every
+/// instance of this as part of a `Result<_, LocalProtocolViolation>` is usually a valid choice.
+///
+/// Higher level layers like [`rpc`](crate::rpc) should make it impossible to encounter
+/// [`LocalProtocolViolation`]s.
+#[derive(Copy, Clone, Debug, Error)]
+pub enum LocalProtocolViolation {
+    /// A request was not sent because doing so would exceed the request limit on channel.
+    ///
+    /// Wait for addtional requests to be cancelled or answered. Calling
```
Fraser999:
```suggestion
    /// Wait for additional requests to be cancelled or answered. Calling
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+    pub fn builder(config: ChannelConfiguration) -> ProtocolBuilder<N> {
+        ProtocolBuilder {
+            channel_config: [config; N],
+            max_frame_size: 1024,
+        }
+    }
+
+    /// Looks up a given channel by ID.
+    ///
+    /// Returns a `LocalProtocolViolation` if called with non-existant channel.
```
Fraser999:
```suggestion
    /// Returns a `LocalProtocolViolation` if called with non-existent channel.
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+        if channel.0 as usize >= N {
+            Err(LocalProtocolViolation::InvalidChannel(channel))
+        } else {
+            Ok(&self.channels[channel.0 as usize])
+        }
+    }
+
+    /// Looks up a given channel by ID, mutably.
+    ///
+    /// Returns a `LocalProtocolViolation` if called with non-existant channel.
```
Fraser999:
```suggestion
    /// Returns a `LocalProtocolViolation` if called with non-existent channel.
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+            if payload.len() > chan.config.max_request_payload_size as usize {
+                return Err(LocalProtocolViolation::PayloadExceedsLimit);
+            }
+        }
+
+        if !chan.allowed_to_send_request() {
+            return Err(LocalProtocolViolation::WouldExceedRequestLimit);
+        }
+
+        // The `unwrap_or_default` below should never be triggered, as long as `u16::MAX` or less
```
Fraser999:
```suggestion
        // The `unwrap_or` below should never be triggered, as long as `u16::MAX` or less
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+        id: Id,
+    ) -> Result<Option<OutgoingMessage>, LocalProtocolViolation> {
+        let chan = self.lookup_channel_mut(channel)?;
+
+        if !chan.incoming_requests.remove(&id) {
+            // The request has been cancelled, no need to send a response.
+            return Ok(None);
+        }
+
+        let header = Header::new(header::Kind::CancelReq, channel, id);
```
Fraser999:
```suggestion
        let header = Header::new(header::Kind::CancelResp, channel, id);
```

juliet/src/protocol.rs
```diff
@@ -0,0 +1,856 @@
...
+                        log_frame!(header);
+                        return Success(CompletedRead::ReceivedResponse {
+                            channel: header.channel(),
+                            id: header.id(),
+                            payload: None,
+                        });
+                    }
+                }
+                Kind::RequestPl => {
+                    // Make a note whether or not we are continueing an existing request.
```
Fraser999:
```suggestion
                    // Make a note whether or not we are continuing an existing request.
```

Fraser999 (commented):
Can you update the rust-toolchain.toml to a version for which this compiles please?  (Or else avoid using `std::sync::OnceLock`, etc)

Still not quite completed the review, but will finish tomorrow.

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+/// The server will automatically be shutdown if the last [`JulietRpcClient`] is dropped.
+pub struct JulietRpcServer<const N: usize, R, W> {
+    core: IoCore<N, R, W>,
+    handle: Handle,
+    pending: HashMap<IoId, Arc<RequestGuardInner>>,
+    new_requests_receiver: UnboundedReceiver<NewRequest>,
+}
+
+/// Internal structure representing a new outgoing request.
+struct NewRequest {
```
Fraser999: I'd suggest renaming to `OutgoingRequest` for clarity.

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+            channel,
+            payload: None,
+            timeout: None,
+        }
+    }
+}
+
+/// An error produced by the RPC error.
+#[derive(Debug, Error)]
+
```
Fraser999:
```suggestion
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+    R: AsyncRead + Unpin,
+    W: AsyncWrite + Unpin,
+{
+    /// Produce the next request from the peer.
+    ///
+    /// Runs the underlying IO until another [`IncomingRequest`] has been produced by the remote
+    /// peer. On success, this function should be called again immediately.
+    ///
+    /// On a regular shutdown (`None` returned) or an error ([`RpcServerError`] returned), a caller
+    /// must stop calling [`next_request`](Self::next_request) and shoudl drop the entire
```
Fraser999:
```suggestion
    /// must stop calling [`next_request`](Self::next_request) and should drop the entire
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+    RemoteClosed(Option<Bytes>),
+    /// Sent, but never received a reply.
+    ///
+    /// Request was sent, but we never received anything back before the [`IoCore`] was shut down.
+    #[error("never received reply before remote closed connection")]
+    Shutdown,
+    /// Local timeout.
+    ///
+    /// The request was cancelled on our end due to a timeout.
+    #[error("request timed out ")]
```
Fraser999:
```suggestion
    #[error("request timed out")]
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+    TimedOut,
+    /// Remove responsed with cancellation.
```
Fraser999:
```suggestion
    /// Remote responded with cancellation.
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+    #[error("remote cancelled our request")]
+    RemoteCancelled,
+    /// Cancelled locally.
+    ///
+    /// Request was cancelled on our end.
+    #[error("request cancelled locally")]
+    Cancelled,
+    /// API misuse
+    ///
+    /// Either the API was misued, or a bug in this crate appeared.
```
Fraser999:
```suggestion
    /// API misuse.
    ///
    /// Either the API was misused, or a bug in this crate appeared.
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+    #[error("API misused or other internal error")]
+    Error(LocalProtocolViolation),
+}
+
+/// Handle to an in-flight outgoing request.
+///
+/// The existance of a [`RequestGuard`] indicates that a request has been made or is on-going. It
```
Fraser999:
```suggestion
/// The existence of a [`RequestGuard`] indicates that a request has been made or is ongoing. It
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+    fn do_cancel(&mut self) {
+        // TODO: Implement eager cancellation locally, potentially removing this request from the
+        //       outbound queue.
+        // TODO: Implement actual sending of the cancellation.
+    }
+
+    /// Forgets the request was made.
+    ///
+    /// Similar [`cancel`](Self::cancel), except that it will not cause an actual cancellation, so
+    /// the peer will likely perform all the work. The response will be discarded.
```
Fraser999:
```suggestion
    /// Similar to [`cancel`](Self::cancel), except that it will not cause an actual cancellation,
    /// so the peer will likely perform all the work. The response will be discarded.
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+}
+
+impl IncomingRequest {
+    /// Returns a reference to the payload, if any.
+    #[inline(always)]
+    pub fn payload(&self) -> &Option<Bytes> {
+        &self.payload
+    }
+
+    /// Returns a reference to the payload, if any.
```
Fraser999:
```suggestion
    /// Returns a mutable reference to the payload, if any.
```

juliet/src/rpc.rs
```diff
@@ -0,0 +1,598 @@
...
+            if let Err(err) = handle.enqueue_response_cancellation(self.channel, self.id) {
+                match err {
+                    EnqueueError::Closed(_) => {
+                        // Do nothing, just discard the response.
+                    }
+                    EnqueueError::BufferLimitHit(_) => {
+                        unreachable!("cannot hit request limit when responding")
+                    }
+                    EnqueueError::LocalProtocolViolation(_) => {
+                        todo!("what to do with this?")
```
Fraser999: Presumably not panic, since this is called in `drop` below?  Can we do more than just log an error?

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+use super::payload_is_multi_frame;
+
+/// A message to be sent to the peer.
+///
+/// [`OutgoingMessage`]s are generated when the protocol requires data to be sent to the peer.
+/// Unless the connection is terminated, they should not be dropped, but can be sent in any order.
+///
+/// While *frames* can be sent in any order, a message may span one or more frames, which can be
+/// interspersed with other messages at will. In general, the [`OutgoingMessage::frames()`] iterator
+/// should be used, even for single-frame messages.
```
Fraser999: Don't we mean frames must be sent in order, but despite that, the peer could receive messages in a different order as smaller messages could "overtake" larger ones?  Not sure how best to convey that, but my main point is that `While *frames* can be sent in any order` seems to be wrong to me.

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+#[derive(Debug)]
+pub struct OutgoingMessage {
+    /// The common header for all outgoing messages.
+    header: Header,
+    /// The payload, potentially split across multiple messages.
+    payload: Option<Bytes>,
+}
+
+impl OutgoingMessage {
+    /// Constructs a new outgoing messages.
```
Fraser999:
```suggestion
    /// Constructs a new outgoing message.
```

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// Returns the outgoing message's header.
+    #[inline(always)]
+    pub fn header(&self) -> Header {
+        self.header
+    }
+}
+
+/// Combination of header and potential frame payload length.
+///
+/// A message with a payload always start with an initial frame that has a header and a varint
```
Fraser999:
```suggestion
/// A message with a payload always starts with an initial frame that has a header and a varint
```

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+    header: Header,
+    /// The payload length. If [`Varint32::SENTINEL`], it will always be omitted from output.
+    payload_length: Varint32,
+}
+
+impl Display for Preamble {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        Display::fmt(&self.header, f)?;
+        if !self.payload_length.is_sentinel() {
+            write!(f, " [l={}]", self.payload_length.decode())?;
```
Fraser999: I think this is too opaque.  Could we stretch to two more chars?
```suggestion
            write!(f, " [len={}]", self.payload_length.decode())?;
```

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// The outgoing message in its entirety.
+    msg: OutgoingMessage,
+    /// Number of bytes output using `OutgoingFrame`s so far.
+    bytes_processed: usize,
+}
+
+impl FrameIter {
+    /// Returns the next frame to send.
+    ///
+    /// Will return the next frame, and `Some(self)` is there are additional frames to send to
```
Fraser999:
```suggestion
    /// Will return the next frame, and `Some(self)` if there are additional frames to send to
```

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+    /// complete the message, `None` otherwise.
+    ///
+    /// # Note
+    ///
+    /// While different [`OutgoingMessage`]s can have their send order mixed or interspersed, a
+    /// caller MUST NOT send [`OutgoingFrame`]s of a single messagw in any order but the one
```
Fraser999:
```suggestion
    /// caller MUST NOT send [`OutgoingFrame`]s of a single message in any order but the one
```

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    /// Returns the outgoing message's header.
+    #[inline(always)]
+    pub fn header(&self) -> Header {
+        self.header
+    }
+}
+
+/// Combination of header and potential frame payload length.
```
Fraser999: There is no "frame payload length" is there (unless I'm misinterpreting it).  I'd call that a message payload length, since frame payload length makes me expect that each frame with a payload will have a length included, and that the length is of the frame payload rather than the message payload.

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+                Varint32::encode(payload_remaining as u32)
+            } else {
+                Varint32::SENTINEL
+            };
+
+            let preamble = if self.bytes_processed == 0 {
+                Preamble::new(self.msg.header, length_prefix)
+            } else {
+                Preamble::new(self.msg.header, Varint32::SENTINEL)
+            };
```
Fraser999:
```suggestion
            // If this is the first frame, include the message payload length.
            let length_prefix = if self.bytes_processed == 0 {
                Varint32::encode(payload_remaining as u32)
            } else {
                Varint32::SENTINEL
            };

            let preamble = Preamble::new(self.msg.header, length_prefix);
```

juliet/src/protocol/outgoing_message.rs
```diff
@@ -0,0 +1,431 @@
...
+
+            let frame_capacity = max_frame_size as usize - preamble.len();
+            let frame_payload_len = frame_capacity.min(payload_remaining);
+
+            let range = self.bytes_processed..(self.bytes_processed + frame_payload_len);
+            let frame_payload = payload.slice(range);
+            self.bytes_processed += frame_payload_len;
+
+            // Update payload remaining, now that an additional frame has been produced.
+            payload_remaining = payload.len() - self.bytes_processed;
```
Fraser999, line 251: Worth doing a `checked_sub().expect()` here to not silently fail if `max_frame_size` is configured too low?

If not, a comment explaining how `max_frame_size` is guaranteed to always be >= max possible value of `preamble.len()` would help.

Fraser999 (commented):
Review completed now - this is shaping up really nicely.

juliet/src/header.rs
marc-casperlabs: Guess it's a matter of taste what's better - the `unsafe impl` or the repetition of `Header::SIZE`.

At the time that code was written there was zero `unsafe` in the entire crate, and right now it is restricted to two instantiations of `NonZeroU*::new_unchecked`, which are easy to verify.

marc-casperlabs: Actually, I found a workaround - `Self::SIZE` does not work, but `Header::SIZE` does!

juliet/src/header.rs
marc-casperlabs: It's not derived for `Kind` though, and I really don't know what the format should look like. Suggestions?

juliet/src/header.rs
marc-casperlabs: Yeah, actually I meant to write "number of channels" instead of "highest channel number", it's 0-based. I'll change it to be just "greater".

juliet/src/protocol/outgoing_message.rs
marc-casperlabs: What, and double computational resources spent from 2 bytes to 4 bytes??

marc-casperlabs: Done.

juliet/src/protocol/outgoing_message.rs
marc-casperlabs: Fixed in 64c5d2b5d -- I added a newtype to capture the requirement instead. I think the comment may not be necessary then.

juliet/src/protocol/outgoing_message.rs
marc-casperlabs: Fixed in 0e21f52ce

juliet/src/lib.rs
marc-casperlabs: ~~It holds true for both, requests and responses (both can be multi-frame). The correct term would be "message" here.~~

My bad.

juliet/src/io.rs
marc-casperlabs: I only get 58.4?

```
(2**64)/(10e9)/(365.25*24*3600)
58.45420460906264
```

But I can live with just 100 million per second and 584 years, I think.

marc-casperlabs: This nicely got rid of the `portable-atomic` dependency: ccc324c42.

juliet/src/rpc.rs
marc-casperlabs: I'd be up for `NewOutgoingRequest` -- just `OutgoingRequest` implies something that can be sent already, but this is an intermediate type that needs to be converted into a proper message still.

marc-casperlabs: 1e551c3cf

juliet/src/protocol/outgoing_message.rs
marc-casperlabs: That's 100% accurate. Fixed in 6fd81d599.

juliet/src/protocol/outgoing_message.rs
marc-casperlabs: Clarified in bc709545e

juliet/src/header.rs
marc-casperlabs: Fixed in f156a293e.

juliet/src/rpc.rs
marc-casperlabs: Fixed in e7b2ebe2d.

Fraser999 (approved):

PR #4130, opened 2023-07-14
Introduce new `Transaction` type

Fraser999:
This PR introduces the `Transaction` type a.k.a "Version 2 Deploy".  It is expected that further changes and additions will be needed to many of the new types as they are integrated into the EE and node codebases.

Much of the code is copied from `Deploy`, but there are some significant differences:
* session code has been renamed `body: TransactionKind` which is a much more fine grained breakdown of the various execution modes compared to deploy's `ExecutableDeployItem`
* there is no payment code: standard payment will be assumed unless the Transaction includes custom payment logic in provided body Wasm
* ~the `TransactionHash` is a hash over "signing hash + approvals", where the "signing hash" is a hash over the header and the body.  The signing hash is not recorded anywhere; it is calculated only when required (once for verification, and once per new approval added).  This approach, where the `TransactionHash` includes the approvals, means that there should be no need to record hashes of approvals in global state.~ This change was reverted on instructions of @EdHastingsCasperLabs 
* ~the body hash has been removed from the header, since it served no useful purpose.  Transactions (and deploys) are never provided to clients or other nodes as just headers, they are always provided as full transactions/deploys, hence there is no benefit to recording the body hash in the header~ This change was reverted on instructions of @EdHastingsCasperLabs 
* the `dependencies` field has been removed from the header as it is also unused
* the `gas_price` has been repurposed into a `PricingMode` where `PricingMode::GasPriceMultiplier` represents the current `gas_price` behaviour

Note that the "interesting" changes are contained within the first commit.  The second commit touches a lot of files, but is simply a cleanup of the `runtime_args!` macro.

Closes [#181](https://github.com/casper-network/roadmap/issues/181).
Closes [#157](https://github.com/casper-network/roadmap/issues/157).


resources/production/chainspec.toml
```diff
@@ -94,6 +94,16 @@ maximum_round_length = '132seconds'
 # Expressed as a fraction (1/5 by default).
 reduced_reward_multiplier = [1, 5]
 
+[transactions]
```
AlexanderLimonov, line 97: I assume that we intend to revisit this and "merge" "deploys" and "transactions," but then why break out separate settings for transactions only?

types/src/transaction/userland_transaction.rs
```diff
@@ -0,0 +1,304 @@
...
+        contract_package_id: Option<ContractPackageIdentifier>,
+        /// Raw Wasm module bytes with 'call' exported as an entrypoint.
+        module_bytes: Bytes,
+        /// Runtime arguments.
+        args: RuntimeArgs,
+    },
+    /// A `Transaction` targeting a stored contract.
+    DirectCall(DirectCall),
+    /// A `Transaction` which doesn't modify global state.
+    Noop {
```
AlexanderLimonov, line 58: What is the purpose of such a transaction? Effectively this is just speculative execution

types/src/transaction.rs
```diff
@@ -0,0 +1,785 @@
...
+        *Timestamp::example(),
+        TimeDiff::from_seconds(3_600),
+        PricingMode::GasPriceMultiplier(1),
+        String::from("casper-example"),
+        TransactionKind::from(NativeTransaction::new_mint_transfer(args)),
+        AccountAndSecretKey::SecretKey(secret_key),
+    )
+});
+
+/// A signed smart contract.
```
AlexanderLimonov: I feel like this line "A signed smart contract" is some ancient holdover from devnet

types/src/transaction.rs
```diff
@@ -0,0 +1,785 @@
...
+    is_verified: OnceCell<Result<(), TransactionConfigFailure>>,
+}
+
+impl Transaction {
+    /// Called by the `TransactionBuilder` to construct a new `Transaction`.
+    #[cfg(any(feature = "std", test))]
+    fn build(
+        timestamp: Timestamp,
+        ttl: TimeDiff,
+        pricing_mode: PricingMode,
```
AlexanderLimonov, line 118: I know that this is where this discussion has tended, but this does only make sense if there are multiple lanes potentially accommodating the same TransactionKind with different pricing. I suppose one could have, just for UserlandTransaction type, the "FIFO" and "competitive" lanes at the least, so this is reasonable, but we should consider far in advance what configuration of lanes we actually want, besides "transfers, installers, everything else and _maybe_ reserved_".

types/src/transaction/pricing_mode.rs
```diff
@@ -0,0 +1,122 @@
...
+)]
+#[serde(deny_unknown_fields)]
+#[non_exhaustive]
+pub enum PricingMode {
+    /// Multiplies the gas used by the given amount.
+    ///
+    /// This is the same behaviour as for the `Deploy::gas_price`.
+    GasPriceMultiplier(u64),
+    /// Irrelevant.
+    Noop,
```
AlexanderLimonov: Should this be perhaps "Fixed" for FIFO lanes?

AlexanderLimonov (approved):
No objections from me, but, of course, things will need to be fleshed out as we go

resources/production/chainspec.toml
Fraser999: I think we'll need to keep two separate sections for the lifespan of the v2 node, until we remove support for `Deploy`s in 3.0.  I expected the new `Transaction` type would have a different set of configuration options since many configurable aspects don't have equivalents for `Deploy`.

types/src/transaction/userland_transaction.rs
Fraser999: Hmm - that was skimmed in the planning discussion, but I'd interpreted it to mean a `Transaction` which can read from global state, but not write.  Speculative exec can perform writes during the execution, but nothing is actually committed to global state at the end.

Ultimately, all these new variants will need "proper" documentation over and above my scribbled rustdoc comments here :)

types/src/transaction.rs
Fraser999: True.  I'll maybe borrow [the description of `Deploy` from our official glossary](https://docs.casper.network/concepts/glossary/D/#deploy).

types/src/transaction/pricing_mode.rs
Fraser999: Happy to change it just now.  I think we have plenty of wiggle room to modify it again in the near future if required.

types/src/transaction.rs
Fraser999: Done in [633530d](https://github.com/casper-network/casper-node/pull/4130/commits/633530d1a7ad8d8957b9363ae6935a9318d6694c).

types/src/transaction/pricing_mode.rs
Fraser999: Done in [633530d](https://github.com/casper-network/casper-node/pull/4130/commits/633530d1a7ad8d8957b9363ae6935a9318d6694c).

types/src/deploy/runtime_args.rs
```diff
@@ -218,19 +241,19 @@ impl FromBytes for RuntimeArgs {
...
+/// use casper_types::runtime_args;
 /// let _named_args = runtime_args! {
 ///   "foo" => 42,
 ///   "bar" => "Hello, world!"
 /// };
 /// ```
 #[macro_export]
 macro_rules! runtime_args {
-    () => (RuntimeArgs::new());
+    () => ($crate::RuntimeArgs::new());
```
mpapierski, line 249: Finally, no more pain importing both `runtime_args` and `RuntimeArgs` 👏

cspramit:
@EdHastingsCasperLabs should look at it.

types/src/transaction/auction_transaction.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Calls the `read_era_id` entry point to provide the current `EraId`.
+    ///
+    /// Requires no runtime args.
+    #[cfg_attr(
+        feature = "json-schema",
+        schemars(
+            description = "Calls the `read_era_id` entry point to provide the current era ID."
+        )
+    )]
+    ReadEraId(RuntimeArgs),
```
mpapierski, line 138: So, with this request being a part of `Transaction` now, and `"read_era_id"` entry point returns a value containing the `EraId` do we know how do we want to send this value back to the user? Does it mean we want to let users return values directly from their entrypoints as well?

types/src/transaction/auction_transaction.rs
```diff
@@ -0,0 +1,269 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn bytesrepr_roundtrip() {
+        let rng = &mut TestRng::new();
+        for _ in 0..10 {
+            bytesrepr::test_serialization_roundtrip(&AuctionTransaction::random(rng));
```
mpapierski, line 266: Some types are doing randomization, while others (older types?) are doing proptest - is there possibility we could stick to one convention for 2.0? (personally not a big fan of randomization) If we were to go with proptest I'd consider `proptest_derive` macro

types/src/transaction/native_transaction.rs
```diff
@@ -0,0 +1,176 @@
...
+            description = "Calls the `transfer` entry point of the mint to transfer `Motes` from
+            a source purse to a target purse."
+        )
+    )]
+    MintTransfer(RuntimeArgs),
+
+    /// A transaction targeting the auction.
+    Auction(AuctionTransaction),
+
+    /// A transaction reserving a future execution.
```
mpapierski, line 60: I remember that in the past there were some hard issues to solve before implementing this i.e. how do price something in the future, how do we ensure we wouldn't queue more than we can reasonably execute in a future block, etc. Is there a plan for this?

mpapierski (approved):
Few nonblocking questions. LGTM

types/src/transaction/auction_transaction.rs
Fraser999: I'm not clear on that aspect myself yet.

types/src/transaction/auction_transaction.rs
Fraser999: The older types using proptest predate the `TestRng` being available in `casper-types`.

I strongly prefer the `TestRng` approach and would hope that we could eventually migrate many of the proptests to this style in the future.  It's probably unreasonable to expect that to happen (or exclusively using proptest rather than `TestRng`) for 2.0, but I think we should stick with `TestRng` for all new types being created unless there's a compelling reason to use proptest for a given type.

types/src/transaction/native_transaction.rs
Fraser999: Again, I'm not sure of the details here yet.

Fraser999:
bors merge

PR #4133, opened 2023-07-14
Use execution journal instead of execution effects 

### LINKED ISSUE #2124: Store the actual journal and not the json helper types.
We should look to replace usage of `ExecutionEffect` with `ExecutionJournal`, and further we should store and serve these as `ExecutionJournal`s rather than as their lossy JSON-helper equivalents (the types defined in execution_results.rs of casper-types crate).

not required: migration of legacy deploys.

Fraser999:
This PR changes the node to use `ExecutionJournal` in place of `ExecutionEffects`.

`ExecutionJournal` has been moved to `casper-types` accordingly, and while the old `ExecutionEffects` type in `casper-types` has been retained for syncing and reporting of historical data, the equivalent type in EE has been removed.

The `ExecutionJournal` also dropped the `Operations` field as this was unused and not populated, leaving the journal as simply a `Vec<Transform>`.

This work introduced a need for a new enum `ExecutionResult` holding old and new forms of execution results.  The `casper_execution_engine::engine_state::ExecutionResult` was updated to hold an `ExecutionJournal`, and we only need to support one flavour of this type, so it's not versioned.  However, the `casper_types::execution::ExecutionResult` became `casper_types::execution::ExecutionResultV1` and was otherwise structurally unchanged, as this is the type held in Storage and which is sent to other nodes/clients on demand.  The new `casper_types::execution::ExecutionResultV2` is equivalent to the updated `casper_execution_engine::engine_state::ExecutionResult` (with the difference continuing to be that the error type of the `casper_types` one is a `String` as `casper-types` has no access to EE error types).

Since Storage had to be updated to hold the new `ExecutionResult`s, I also took the opportunity to fix the long-standing issue of multiple execution results being attributed to a single deploy.  There is now a 1-1 mapping of deploy to execution result.

This PR is blocked by [client PR #85](https://github.com/casper-ecosystem/casper-client-rs/pull/85), since CI will likely fail until the client is updated.

Closes #2124.
Closes [#160](https://github.com/casper-network/roadmap/issues/160).


execution_engine/src/tracking_copy/byte_size.rs
```diff
@@ -63,7 +63,8 @@ impl HeapSizeOf for Key {
 // TODO: contract has other fields (re a bunch) that are not repr here...on purpose?
 impl HeapSizeOf for Account {
     fn heap_size(&self) -> usize {
-        self.named_keys().heap_size()
+        // self.named_keys().0.heap_size()
```
mpapierski: Commented (why though, seems legit)

execution_engine_testing/test_support/src/wasm_test_builder.rs
```diff
@@ -86,25 +86,25 @@ pub type LmdbWasmTestBuilder = WasmTestBuilder<DataAccessLayer<LmdbGlobalState>>
...
     genesis_hash: Option<Digest>,
     /// Post state hash.
     post_state_hash: Option<Digest>,
-    /// Cached transform maps after subsequent successful runs i.e. `transforms[0]` is for first
-    /// exec call etc.
-    transforms: Vec<ExecutionJournal>,
-    /// Cached genesis transforms
+    /// Cached effects after successful runs i.e. `effects[0]` is the collection of effects for
+    /// first exec call, etc.
+    effects: Vec<ExecutionJournal>,
```
mpapierski: How about

```suggestion
    journal: Vec<ExecutionJournal>,
```
also the comment mentions effects rather than journal. I see you're naming a vec of ExecJournal "effects"?

cspramit:
@alsrdn to review first pass

mpapierski (approved):

execution_engine/src/tracking_copy/byte_size.rs
Fraser999: Good catch!

Fixed in [b04fa35](https://github.com/casper-network/casper-node/pull/4133/commits/b04fa3536174ea7fbab986b55bae6bf5af1a840c).

execution_engine_testing/test_support/src/wasm_test_builder.rs
Fraser999: I was aiming to consistently stick with calling this `effects` as that's the term people are most familiar with.  I'd actually like to go the whole way and rename `ExecutionJournal` to `Effects`, but I don't know if that would be acceptable to others?

EdHastingsCasperAssociation: im ok with it

node/src/components/storage.rs
```diff
@@ -1464,43 +1500,63 @@ impl Storage {
         &mut self,
         txn: &mut RwTransaction,
         block_hash: &BlockHash,
+        block_height: u64,
         execution_results: HashMap<DeployHash, ExecutionResult>,
     ) -> Result<bool, FatalStorageError> {
+        insert_to_deploy_index(
```
alsrdn, line 1506: I'm not quite sure if this is needed. The block should always be written before the execution results, right? If it comes from a `PutExecutedBlock` request `write_validated_block` is called before and it updates the index. If we write execution results as part of a sync, then the block should have been stored as part of fetching.

node/src/components/storage/tests.rs
```diff
@@ -1018,18 +933,37 @@ fn store_execution_results_twice_for_same_block_deploy_pair() {
...
+        &mut harness,
+        &mut storage,
+        block_hash,
+        block_height,
+        exec_result_1,
+    );
 
     // Storing a second execution result for the same deploy on the same block should panic.
-    put_execution_results(&mut harness, &mut storage, block_hash, exec_result_2);
+    put_execution_results(
```
alsrdn, line 962: Not from this PR but I'm a bit confused about the statement in the comment above. The operation is expected to panic but the test is not marked as `#[should_panic]`.

node/src/types/block/block_execution_results_or_chunk.rs
```diff
@@ -28,29 +33,127 @@ pub struct BlockExecutionResultsOrChunk {
...
+            let value = make_value_or_chunk(execution_results, &block_hash, chunk_index)?;
+            return Some(BlockExecutionResultsOrChunk {
+                block_hash,
+                value,
+                is_valid: OnceCell::new(),
+            });
+        }
+
+        // If it is V1, we need to construct the `ValueOrChunk` from a `Vec<ExecutionResultV1>` if
+        // it's big enough to need chunked, otherwise we need to use the
```
alsrdn: nit: big enough to need chunking

node/src/types/block/block_execution_results_or_chunk.rs
```diff
@@ -28,29 +33,127 @@ pub struct BlockExecutionResultsOrChunk {
...
+                v1_results.push(v1_result);
+            } else {
+                error!(
+                    ?execution_results,
+                    "all execution results should be version 1"
+                );
+                return None;
+            }
+        }
+        if v1_results.serialized_length() <= ChunkWithProof::CHUNK_SIZE_BYTES {
```
alsrdn, line 92: So `v1_results` is a vec of `ExecutionResultV1` values and `execution_results` is a vec of `ExecutionResult`. I think it is possible for `v1_results.serialized_length() <= ChunkWithProof::CHUNK_SIZE_BYTES` to be true but `execution_results.serialized_length()` to be larger than `ChunkWithProof::CHUNK_SIZE_BYTES` because the serialized size of `ExecutionResult` is larger than `ExecutionResultV1` (1 byte extra for the version tag).

In turn I think `make_value_or_chunk(execution_results)` would actually chunk the data and return `ValueOrChunk::ChunkWithProof` instead of the `ValueOrChunk::Value` that we expect.

alsrdn (commented):
Did a first pass of this PR, but didn't manage to finish going through all of it. I'm posting the comments I have so far and will follow up with the full review.

types/src/account/associated_keys.rs
```diff
@@ -5,21 +5,28 @@ use alloc::{
...
 
-/// A mapping that represents the association of a [`Weight`] with an [`AccountHash`].
+/// A collection of weighted public keys (represented as account hashes) associated with an account.
 #[derive(Default, PartialOrd, Ord, PartialEq, Eq, Clone, Debug, Serialize, Deserialize)]
 #[cfg_attr(feature = "datasize", derive(DataSize))]
-pub struct AssociatedKeys(BTreeMap<AccountHash, Weight>);
+#[cfg_attr(feature = "json-schema", derive(JsonSchema))]
+#[serde(deny_unknown_fields)]
+pub struct AssociatedKeys(
+    #[serde(with = "BTreeMapToArray::<AccountHash, Weight, Labels>")] BTreeMap<AccountHash, Weight>,
```
alsrdn: nit: can we add the type on a newline here for cleaner layout?

types/src/block/block_hash.rs
```diff
@@ -26,13 +26,10 @@ use crate::{
...
     derive(JsonSchema),
-    schemars(
-        with = "String",
-        description = "Hex-encoded cryptographic hash of a block."
-    )
+    schemars(description = "Hex-encoded cryptographic hash of a block.")
 )]
 #[serde(deny_unknown_fields)]
-pub struct BlockHash(Digest);
+pub struct BlockHash(#[cfg_attr(feature = "json-schema", schemars(skip, with = "String"))] Digest);
```
alsrdn: nit: same here, can we add the attributes and type on newlines?

resources/test/rest_schema_status.json
```diff
@@ -265,14 +261,49 @@
...
       "description": "The state of the reactor.",
-      "type": "string",
-      "enum": [
-        "Initialize",
-        "CatchUp",
-        "Upgrading",
-        "KeepUp",
-        "Validate",
-        "ShutdownForUpgrade"
+      "oneOf": [
```
alsrdn, line 264: I find this schema to be rather weird for an enum. I was expecting the same schema as before. Not sure we can do anything about it since this is how `schemars` generates it and it would work the same for validation.  But I'm wondering, would this would complicate the clients that generate code based on the schema provided?

smart_contracts/contracts/test/ordered-transforms/src/main.rs
```diff
@@ -29,15 +29,14 @@ pub extern "C" fn call() {
...
 
     let n: u32 = runtime::get_named_arg("n");
-    let named_keys = (0..n)
-        .map(|i| (format!("uref-{}", i), Key::URef(storage::new_uref(0_i32))))
-        .chain(core::iter::once((
-            "n-urefs".to_string(),
-            Key::URef(storage::new_uref(n)),
-        )));
+    let mut named_keys = NamedKeys::new();
+    (0..n).for_each(|i| {
```
alsrdn: nit: since we're not using `chain` anymore I think a plain `for` loop is more appropriate.

types/src/named_keys.rs
```diff
@@ -0,0 +1,155 @@
...
+#[cfg(feature = "json-schema")]
+impl KeyValueJsonSchema for Labels {
+    const JSON_SCHEMA_KV_NAME: Option<&'static str> = Some("NamedKey");
+    const JSON_SCHEMA_KV_DESCRIPTION: Option<&'static str> = Some("A key with a name.");
+    const JSON_SCHEMA_KEY_DESCRIPTION: Option<&'static str> = Some("The name of the entry.");
+    const JSON_SCHEMA_VALUE_DESCRIPTION: Option<&'static str> =
+        Some("The value of the entry: a casper `Key` type.");
+}
+
+// #[cfg(test)]
```
alsrdn: Maybe it would be good to add a TODO rather than this commented block.

execution_engine/src/engine_state/mod.rs
```diff
@@ -381,37 +376,48 @@ where
             // 3.1.2.4 if new total validator slots is provided, update auction contract state
             let auction_contract = tracking_copy.borrow_mut().get_contract(*auction_hash)?;
 
-            let validator_slots_key = auction_contract.named_keys()[VALIDATOR_SLOTS_KEY];
+            let validator_slots_key = auction_contract
+                .named_keys()
+                .get(VALIDATOR_SLOTS_KEY)
+                .expect("validator_slots key must exist in auction contract's named keys");
```
alsrdn, line 389: nit: should we maybe take the opportunity to return a fatal error in these cases instead of panicking?

alsrdn (commented):
Finished going through the PR.

node/src/components/storage.rs
Fraser999: Hmm, I think that's probably true.  On the other hand, I'd rather not bank on it always being true, and it shouldn't hurt to also try inserting here too.

resources/test/rest_schema_status.json
Fraser999: Yeah, our hands are a bit tied here since I don't know of any replacement for `schemars` (although I haven't looked for a while).

types/src/named_keys.rs
Fraser999: Good catch!  I'll just add the test I had in mind when I added this.  (I should have never commented it out!)

execution_engine/src/engine_state/mod.rs
Fraser999: I'll leave this as is for now, as EE has a lot of legacy instances where that is applicable.  I'd rather maintain consistency and add a tech-debt ticket if we decide to bring EE in line with the rest of the node codebase in that regard.

node/src/types/block/block_execution_results_or_chunk.rs
Fraser999: This is catch of the year for my money - fantastic!

I'll fix this and add a regression test.

node/src/components/storage/tests.rs
Fraser999: Fixed in [8822cab](https://github.com/casper-network/casper-node/pull/4133/commits/8822cab81f3c698153361261ffdb73a69ccdcf0a).

node/src/types/block/block_execution_results_or_chunk.rs
Fraser999: Fixed in [8822cab](https://github.com/casper-network/casper-node/pull/4133/commits/8822cab81f3c698153361261ffdb73a69ccdcf0a).

node/src/types/block/block_execution_results_or_chunk.rs
Fraser999: Fixed in [8822cab](https://github.com/casper-network/casper-node/pull/4133/commits/8822cab81f3c698153361261ffdb73a69ccdcf0a).

types/src/account/associated_keys.rs
Fraser999: Needed to add `#[rustfmt::skip]`, but fixed in [8822cab](https://github.com/casper-network/casper-node/pull/4133/commits/8822cab81f3c698153361261ffdb73a69ccdcf0a).

types/src/block/block_hash.rs
Fraser999: Fixed in [8822cab](https://github.com/casper-network/casper-node/pull/4133/commits/8822cab81f3c698153361261ffdb73a69ccdcf0a).

smart_contracts/contracts/test/ordered-transforms/src/main.rs
Fraser999: Fixed in [8822cab](https://github.com/casper-network/casper-node/pull/4133/commits/8822cab81f3c698153361261ffdb73a69ccdcf0a).

types/src/named_keys.rs
Fraser999: Fixed in [8822cab](https://github.com/casper-network/casper-node/pull/4133/commits/8822cab81f3c698153361261ffdb73a69ccdcf0a).

alsrdn (approved):

EdHastingsCasperAssociation (approved):
im mildly surprised how pervasive this turned out to be; gj

Fraser999:
bors merge

Fraser999:
bors merge

Fraser999:
bors merge

Fraser999:
bors merge

PR #4163, opened 2023-07-25
Introduce AddressableEntity

darthsiroftardis:
CHANGELOG:

- Introduce the `AddressableEntity` type which unifies the `Account` and `Contract` structure and behavior
- Renamed `ContractPackage` to `Package` and created the `package.rs` mod which contains the relevant data types and behavior
- Added the `PackageKind` enum which identifies the type of `Package` (formerly the ContractPackage)
- Extended the `ContractPackage` struct to include a `PackageKind` enum which reflects the kind of `Package`
- Added migration capability to move legacy `Accounts`, `Contracts` and `ContractPackages` 
- Amended both `Runtime` and `RuntimeContext` structs to work with the unified `AddressableEntity` type


cspramit:
ideally 3 reviews needed on this one, big ticket item.

execution_engine/src/engine_state/genesis.rs
```diff
@@ -874,37 +872,89 @@ where
...
 
         let contract_wasm = ContractWasm::new(vec![]);
-        let contract = Contract::new(
+        let main_purse = self.create_purse(starting_balance)?;
+        let associated_keys = contract_package_kind.associated_keys();
+        let maybe_account_hash = contract_package_kind.maybe_account_hash();
+        let named_keys = match maybe_named_keys {
+            Some(named_keys) => named_keys,
+            None => NamedKeys::default(),
+        };
```
mpapierski:
```suggestion
        let named_keys = maybe_named_keys.unwrap_or_default();
```

execution_engine/src/engine_state/mod.rs
```diff
@@ -676,30 +680,123 @@ where
...
+            .borrow_mut()
+            .read(&Key::Account(account_hash))
+            .map_err(Into::into)?;
+
+        match maybe_stored_value {
+            Some(StoredValue::Account(account)) => {
+                let mut generator =
+                    AddressGenerator::new(account.main_purse().addr().as_ref(), Phase::System);
+
+                let contract_wasm_hash = ContractWasmHash::new(ACCOUNT_WASM_ADDR);
```
mpapierski: How's `ACCOUNT_WASM_ADDR` differs from `DEFAULT_ADDRESS` constant? Could we use DEFAULT_ADDRESS here? If we want separate constants could we change `ACCOUNT_WASM_ADDR` to `const ACCOUNT_WASM_ADDR: ContractWasmHash = ContractWasmHash::new(DEFAULT_ADDRESS)`

execution_engine/src/engine_state/mod.rs
```diff
@@ -676,30 +680,123 @@ where
...
+                let contract_by_account = match CLValue::from_t(contract_key) {
+                    Ok(cl_value) => cl_value,
+                    Err(_) => {
+                        return Err(Error::Bytesrepr("Failed to convert to CLValue".to_string()))
+                    }
+                };
+
+                tracking_copy.borrow_mut().write(
+                    Key::Account(account_hash),
+                    StoredValue::CLValue(contract_by_account),
```
mpapierski: It's clever to overwrite `StoredValue::Account` with `StoredValue::CLValue` that points to a newly generated `Key,`, but it should be safe to just write an addressable entity under `account_hash`, and also copy over all contract hashes, and assume there won't be any collision practically speaking. Then, this could be simplified to just purge the account from the tip. 

Also, one less indirection in the system, one less expensive I/O read etc.

execution_engine/src/engine_state/mod.rs
```diff
@@ -676,30 +680,123 @@ where
...
+                );
+                Ok(())
+            }
+            Some(StoredValue::CLValue(_)) => Ok(()),
```
mpapierski, line 807: If we'd consider purging, you could assume "None" variant is already migrated account. (at least you could add a comment here about what happened)

execution_engine/src/engine_state/mod.rs
```diff
@@ -891,12 +1029,68 @@ where
...
+                            let contract_wasm_hash = ContractWasmHash::new(ACCOUNT_WASM_ADDR);
+                            let contract_hash = ContractHash::new(generator.new_hash_address());
+                            let contract_package_hash =
+                                ContractPackageHash::new(generator.new_hash_address());
+
+                            let associated_keys = AssociatedKeys::new(public_key, Weight::new(1));
+                            let named_keys = NamedKeys::default();
+                            let entry_points = EntryPoints::new();
+
+                            let entity = AddressableEntity::new(
```
mpapierski: This code to create an `AddressableEntity` seems duplicated to the one around line 700 in `fn migrate_account` - could we have a designated constructor for creating new entities similar to `Account::create` to decrease the amount of boilerplate?

Or, if I'm reading this correctly, this block could as well be replaced with a call to `fn migrate_account`?

execution_engine/src/engine_state/mod.rs
```diff
@@ -1403,16 +1629,17 @@ where
 
         // the proposer of the block this deploy is in receives the gas from this deploy execution
         let proposer_purse = {
-            let proposer_account: Account = match tracking_copy
+            let proposer_contract = match tracking_copy
```
mpapierski: `let proposer_addressable_entity`, no? `AddressableEntity` could point to a contract that's `Account`, right?

execution_engine/src/engine_state/mod.rs
```diff
@@ -1897,11 +2130,9 @@ where
...
         let system_account_addr = PublicKey::System.to_account_hash();
 
-        let virtual_system_account = {
-            let named_keys = NamedKeys::new();
-            let purse = URef::new(Default::default(), AccessRights::READ_ADD_WRITE);
-            Account::create(system_account_addr, named_keys, purse)
-        };
+        let virtual_system_contract_by_account = tracking_copy
+            .borrow_mut()
+            .get_addressable_entity_by_account_hash(protocol_version, system_account_addr)?;
```
mpapierski, line 2095: It could be (and should be, IMO) a virtual system account - one less expensive IO read if anything

execution_engine/src/engine_state/upgrade.rs
```diff
@@ -3,14 +3,24 @@ use std::{cell::RefCell, fmt, rc::Rc};
 
 use thiserror::Error;
 
+use crate::ACCOUNT_WASM_ADDR;
```
mpapierski: Move around line 20th for `crate::` level imports

execution_engine/src/engine_state/upgrade.rs
```diff
@@ -202,4 +195,115 @@ where
...
+        let contract_hash = ContractHash::new(address_generator.new_hash_address());
+        let contract_package_hash = ContractPackageHash::new(address_generator.new_hash_address());
+
+        let contract_wasm = ContractWasm::new(vec![]);
+
+        let account_hash = PublicKey::System.to_account_hash();
+        let associated_keys = AssociatedKeys::new(account_hash, Weight::new(1));
+
+        let main_purse = {
+            let purse_addr = address_generator.new_hash_address();
```
mpapierski, line 239: Hold on, why are you creating a new purse for a system account where you could just take the existing system account's purse and move it to addressable entity? If you want to generate a new main purse for the system's addressable entity, we should burn the tokens in the old main purse if the balance > 0.

execution_engine/src/execution/error.rs
```diff
@@ -177,6 +179,12 @@ pub enum Error {
     /// Transform error.
     #[error(transparent)]
     Transform(transform::Error),
+    /// Invalid key
+    #[error("Invalid key {0}")]
+    InvalidKey(Key),
```
mpapierski: To keep the naming consistent to `UnexpectedStoredValueVariant` it should say `UnexpectedKeyVariant` (arguably then we should remove `KeyIsNotAURef` variant as well which seems to be used only in tests)

execution_engine/src/execution/error.rs
```diff
@@ -177,6 +179,12 @@ pub enum Error {
...
+    /// Invalid Contract package kind.
+    #[error("Invalid contract package kind")]
```
mpapierski:
```suggestion
    #[error("Invalid contract package kind: {0}")]
```
for better error reporting (needs `impl Display`)

execution_engine/src/execution/executor.rs
```diff
@@ -103,7 +109,15 @@ impl Executor {
                 // These args are passed through here as they are required to construct the new
                 // `Runtime` during the contract's execution (i.e. inside
                 // `Runtime::execute_contract`).
-                runtime.call_contract_with_stack(contract_hash, &entry_point_name, args, stack)
+                match runtime.migrate_contract_and_contract_package(contract_hash) {
```
mpapierski, line 112: I'd find it more readable:

```rust
runtime.migrate_contract_and_contract_package(contract_hash)?;
runtime.call_contract_with_stack(...)?;
Ok(())
```

execution_engine/src/runtime/mod.rs
```diff
@@ -1304,17 +1290,36 @@ where
...
+                }
+                SystemContractType::HandlePayment => {
+                    return self.call_host_handle_payment(
+                        entry_point.name(),
+                        &context_args,
+                        access_rights,
+                        stack,
+                    );
+                }
+                SystemContractType::StandardPayment => {}
```
mpapierski, line 1313: `self.call_host_standard_payment` is missing here?

types/src/stored_value.rs
```diff
@@ -48,10 +50,10 @@ pub enum StoredValue {
     Account(Account),
     /// Variant that stores [`ContractWasm`].
     ContractWasm(ContractWasm),
-    /// Variant that stores [`Contract`].
+    /// Variant that stores [`ContractWasm`].
```
mpapierski:
```suggestion
    /// Variant that stores [`Contract`].
```

types/src/system/system_contract_type.rs
```diff
@@ -29,6 +47,44 @@ pub enum SystemContractType {
     Auction,
 }
 
+impl ToBytes for SystemContractType {
+    fn to_bytes(&self) -> Result<Vec<u8>, Error> {
```
mpapierski, line 51: We should implement `fn write_bytes` and `fn to_bytes` should call write_bytes. Like here https://github.com/casper-network/casper-node/blob/f61081b83cd5bf32fefc4090158bdaa7329b7d62/node/src/types/chainspec/global_state_update.rs#L73-L77 (CC @Fraser999)

execution_engine/src/engine_state/mod.rs
darthsiroftardis: No I disagree, we write addressable entities associated with contracts under psuedo randomly generated hashes, so we would introduce an inconsistent primary key generation; i.e account entities under derived keys and contract entities under psuedo random hashes.

execution_engine/src/engine_state/mod.rs
darthsiroftardis: Yeah, I will introduce a `create_addressable_entity_from_account` to help with this, thanks for the comment, it also helped me spot a subtle error

execution_engine/src/engine_state/mod.rs
darthsiroftardis: Correct, yes, will rename

execution_engine/src/runtime/mod.rs
darthsiroftardis: No, in the previous code tree we only checked if it was the other three system contracts, thus I've left this code path as is

execution_engine/src/execution/executor.rs
darthsiroftardis: Tried applying this, but give the return value in this match block the compiler doesn't allow for it

execution_engine/src/runtime/mod.rs
EdHastingsCasperAssociation: standard payment isn't actually a system contract; it's payment logic and is not callable.

execution_engine/src/engine_state/upgrade.rs
EdHastingsCasperAssociation: this is legit. should fetch the existing account

execution_engine/src/runtime_context/mod.rs
```diff
@@ -113,36 +121,39 @@ where
     /// Where we already have a runtime context, consider using `new_from_self()`.
     #[allow(clippy::too_many_arguments)]
     pub fn new(
-        tracking_copy: Rc<RefCell<TrackingCopy<R>>>,
-        entry_point_type: EntryPointType,
         named_keys: &'a mut NamedKeys,
-        access_rights: ContextAccessRights,
-        runtime_args: RuntimeArgs,
+        entity: &'a AddressableEntity,
+        entity_address: Key,
```
EdHastingsCasperAssociation, line 126: i think this is initiating_entity, and the type should be AddressableEntityHash instead of Key

execution_engine/src/runtime_context/mod.rs
```diff
@@ -245,20 +284,21 @@ where
...
-                    let mut account: Account = self.read_gs_typed(&account_hash)?;
-                    account.named_keys_mut().remove(name);
-                    account
+                let (contract, contract_key): (AddressableEntity, Key) = {
+                    let contract_key = self
+                        .read_gs_typed::<CLValue>(&account_hash)?
+                        .into_t::<Key>()?;
+
+                    let contract: AddressableEntity = self.read_gs_typed(&contract_key)?;
+                    (contract, contract_key)
```
EdHastingsCasperAssociation, line 295: since addressableentity is not content addressable and is mutable, there's nothing preventing us from just sticking the addressableentityhash onto the addressableentity struct / making it self describing. this would simplify a lot of places where we juggle the hash or key + the record, as well as off chain consumers who also have to do it.

execution_engine/src/runtime_context/mod.rs
```diff
@@ -402,8 +437,13 @@ where
...
-    pub fn base_key(&self) -> Key {
-        self.base_key
+    pub fn get_entity_address(&self) -> Key {
+        self.entity_address
+    }
+
+    /// Returns the initiater of the call chain.
+    pub fn get_caller(&self) -> AccountHash {
+        self.account_hash
     }
```
EdHastingsCasperAssociation, line 447: maybe (AccountHash, AddressableEntityHash) or a newtype of same will be more ergonomic

execution_engine/src/shared/test_utils.rs
```diff
@@ -0,0 +1,14 @@
...
+    contracts::NamedKeys,
+    AccessRights, Key, StoredValue, URef,
+};
+
+/// Returns an account value paired with its key
+pub fn mocked_account(account_hash: AccountHash) -> Vec<(Key, StoredValue)> {
+    let purse = URef::new([0u8; 32], AccessRights::READ_ADD_WRITE);
+    let account = Account::create(account_hash, NamedKeys::new(), purse);
+    vec![(Key::Account(account_hash), StoredValue::Account(account))]
+}
```
EdHastingsCasperAssociation: shouldn't need this

execution_engine/src/tracking_copy/ext.rs
```diff
@@ -18,11 +23,28 @@ pub trait TrackingCopyExt<R> {
...
+    fn get_addressable_entity_by_account_hash(
+        &mut self,
+        protocol_version: ProtocolVersion,
+        account_hash: AccountHash,
+    ) -> Result<AddressableEntity, Self::Error>;
 
-    /// Reads the account at a given account address.
-    fn read_account(&mut self, account_hash: AccountHash) -> Result<Account, Self::Error>;
+    /// Reads the entity key for the account at a given account address.
+    fn read_account(&mut self, account_hash: AccountHash) -> Result<Key, Self::Error>;
```
EdHastingsCasperAssociation, line 40: return AddressableEntityHash

execution_engine/src/tracking_copy/ext.rs
```diff
@@ -34,33 +56,32 @@ pub trait TrackingCopyExt<R> {
...
         contract_wasm_hash: ContractWasmHash,
     ) -> Result<ContractWasm, Self::Error>;
 
-    /// Gets a contract header by Key.
-    fn get_contract(&mut self, contract_hash: ContractHash) -> Result<Contract, Self::Error>;
+    /// Gets an addressable entity  by Key.
+    fn get_contract(
+        &mut self,
+        contract_hash: ContractHash,
+    ) -> Result<AddressableEntity, Self::Error>;
```
EdHastingsCasperAssociation, line 78: get_addressable_entity

execution_engine/src/tracking_copy/ext.rs
```diff
@@ -34,33 +56,32 @@ pub trait TrackingCopyExt<R> {
...
 
     /// Gets a contract package by Key.
     fn get_contract_package(
```
EdHastingsCasperAssociation, line 81: get_package (and so on...I'll not call out every case where `contract` or `account` etc labeling is still present, but they should all be addressed before the feature branch merges to dev).

execution_engine/src/tracking_copy/ext.rs
```diff
@@ -18,11 +23,28 @@ pub trait TrackingCopyExt<R> {
     /// The type for the returned errors.
     type Error;
 
-    /// Gets the account at a given account address.
-    fn get_account(&mut self, account_hash: AccountHash) -> Result<Account, Self::Error>;
+    /// Gets the contract hash for the account at a given account address.
+    fn get_entity_hash_by_account_hash(
+        &mut self,
+        account_hash: AccountHash,
+    ) -> Result<ContractHash, Self::Error>;
```
EdHastingsCasperAssociation, line 30: redundant w/ read_account

execution_engine/src/tracking_copy/ext.rs
```diff
@@ -18,11 +23,28 @@ pub trait TrackingCopyExt<R> {
     /// The type for the returned errors.
     type Error;
 
-    /// Gets the account at a given account address.
-    fn get_account(&mut self, account_hash: AccountHash) -> Result<Account, Self::Error>;
+    /// Gets the contract hash for the account at a given account address.
```
EdHastingsCasperAssociation, line 26: i think we can take a rework pass on the trait and potentially simplify away some of this

execution_engine_testing/tests/src/test/groups.rs
```diff
@@ -520,7 +498,7 @@ fn should_call_group_unrestricted_contract_caller() {
     let _account = builder
         .query(None, Key::Account(*DEFAULT_ACCOUNT_ADDR), &[])
         .expect("should query account")
-        .as_account()
+        .as_cl_value()
```
EdHastingsCasperAssociation, line 501: can be micro optimized to specific method

execution_engine_testing/tests/src/test/regression/ee_599.rs
```diff
@@ -9,7 +9,7 @@ use casper_types::{account::AccountHash, runtime_args, RuntimeArgs, U512};
 const CONTRACT_EE_599_REGRESSION: &str = "ee_599_regression.wasm";
 const CONTRACT_TRANSFER_TO_ACCOUNT: &str = "transfer_to_account_u512.wasm";
 const DONATION_PURSE_COPY_KEY: &str = "donation_purse_copy";
-const EXPECTED_ERROR: &str = "InvalidContext";
+const EXPECTED_ERROR: &str = "ForgedReference";
```
EdHastingsCasperAssociation, line 12: presumably this is because there is no longer a distinction btwn account and contract contexts; its all just addressable entity contexts...thus the true error is forged uref?

execution_engine_testing/tests/src/test/system_contracts/auction/bids.rs
```diff
@@ -3562,18 +3563,28 @@ fn should_continue_auction_state_from_release_1_4_x() {
...
+    let accounts = vec![*BID_ACCOUNT_1_ADDR, *BID_ACCOUNT_2_ADDR, *DELEGATOR_1_ADDR];
+
+    for legacy_account in accounts {
+        let do_nothing_request =
+            ExecuteRequestBuilder::standard(legacy_account, DO_NOTHING_WASM, runtime_args! {})
+                .build();
+
+        builder.exec(do_nothing_request).expect_success().commit();
+    }
+
```
EdHastingsCasperAssociation, line 3575: put a comment such as:
// this forces on-use migration of accounts

execution_engine_testing/tests/src/test/system_contracts/auction/distribute.rs
```diff
@@ -2994,22 +2994,19 @@ fn should_distribute_delegation_rate_full_after_upgrading() {
...
-        METHOD_DISTRIBUTE,
-        runtime_args! {
-            ARG_ENTRY_POINT => METHOD_DISTRIBUTE,
-            ARG_VALIDATOR => VALIDATOR_1.clone()
-        },
-    )
-    .build();
-
-    builder.exec(distribute_request).commit().expect_success();
+    builder
```
EdHastingsCasperAssociation, line 3001: good tech debt payback

node/src/components/deploy_acceptor.rs
```diff
@@ -343,10 +346,110 @@ impl DeployAcceptor {
...
+                    })
+            }
+            Some(StoredValue::Account(account)) => {
+                let authorization_keys = event_metadata
+                    .deploy
+                    .approvals()
+                    .iter()
+                    .map(|approval| approval.signer().to_account_hash())
+                    .collect();
+                if !account.can_authorize(&authorization_keys) {
```
EdHastingsCasperAssociation: see if you can dedup common logic w/ EntityResult...should flatten out easily

node/src/reactor/main_reactor/event.rs
```diff
@@ -53,7 +53,7 @@ use crate::{
 // 192 is six 256 bit copies, ideally we'd be below, but for now we enforce this as an upper limit.
 // 200 is where the `large_enum_variant` clippy lint draws the line as well.
 const _MAIN_EVENT_SIZE: usize = mem::size_of::<MainEvent>();
-const_assert!(_MAIN_EVENT_SIZE <= 192);
+//const_assert!(_MAIN_EVENT_SIZE <= 192);
```
EdHastingsCasperAssociation, line 56: ?

storage/src/global_state/shared/transform.rs
```diff
@@ -564,14 +562,14 @@ mod tests {
...
-        ));
-        assert_yields_type_mismatch_error(account);
+        // let account = StoredValue::Account(Account::new(
+        //     ZERO_PUBLIC_KEY,
+        //     NamedKeys::new(),
+        //     uref,
+        //     AssociatedKeys::default(),
+        //     ActionThresholds::default(),
+        // ));
+        // assert_yields_type_mismatch_error(account);
```
EdHastingsCasperAssociation: del or fix

types/benches/bytesrepr_bench.rs
```diff
@@ -446,41 +447,41 @@ fn deserialize_u512(b: &mut Bencher) {
...
-    let account = sample_account(10, 10);
-    b.iter(|| ToBytes::to_bytes(black_box(&account)));
-}
-
-fn deserialize_account(b: &mut Bencher) {
-    let account = sample_account(10, 10);
-    let account_bytes = Account::to_bytes(&account).unwrap();
-    b.iter(|| Account::from_bytes(black_box(&account_bytes)).unwrap());
-}
+// fn sample_account(associated_keys_len: u8, named_keys_len: u8) -> Account {
```
EdHastingsCasperAssociation: del or fix

types/benches/bytesrepr_bench.rs
```diff
@@ -851,8 +856,8 @@ fn bytesrepr_bench(c: &mut Criterion) {
     c.bench_function("deserialize_u256", deserialize_u256);
     c.bench_function("serialize_u512", serialize_u512);
     c.bench_function("deserialize_u512", deserialize_u512);
-    c.bench_function("bytesrepr::serialize_account", serialize_account);
-    c.bench_function("bytesrepr::deserialize_account", deserialize_account);
+    // c.bench_function("bytesrepr::serialize_account", serialize_account);
+    // c.bench_function("bytesrepr::deserialize_account", deserialize_account);
```
EdHastingsCasperAssociation, line 824: del

types/src/addressable_entity/account_hash.rs
```diff
@@ -0,0 +1,218 @@
+use alloc::{string::String, vec::Vec};
```
EdHastingsCasperAssociation, line 1: do we need this (the entire file and all the things in it)?

EdHastingsCasperAssociation (approved):
some nits where noted. the main outstanding issue is a renaming pass is badly needed to remove contract_ / Contract mentions in argument, variable, and function names (etc). I'm ok with that happening in a separate PR, so long as it does happen before we merge to dev

execution_engine/src/engine_state/mod.rs
mpapierski: I see your point, although it still seems more practical to address entities by `account_hash`: you can purge old `Key::Account` so each account occupies only a single entity in the tip (instead of two), you don't need to tweak query logic to "hide" the indirect pseudo-random hash pointing from `Key::Account` to `Key::Entity` etc. And if you think about it `account_hash` is also a derived key from public key, technically a public key is also derived from a secret key, and a secret key comes from a pseudo-random generator.

That said I won't hold the PR over this, it's something we can decide to discuss later having enough time.

darthsiroftardis:
bors try

darthsiroftardis:
bors try

Issue #2064, opened 2021-09-13
Support factory pattern in the core


mpapierski:
There's feedback coming from the contract developers that the factory pattern is something they want to use. For example, a common scenario might be that a contract could own an ERC20 token with additional custom logic.

Technically we do support it on the EE, but we never introduced necessary tests, and it was never introduced into the core as a supported feature.

At a minimum, we need to implement tests for a session code that will deploy contract A (aka factory contract) with entrypoints marked as `group(vec![])` (essentially can't be called) and a contract entry point that creates a contract B (new contract) with the same set of entrypoints but different entry point access.

It will be worthwhile to discuss the possibility if the factory contract needs to have symbols in the WASM present, or we can somehow create those reserved entrypoints on the fly without the burden on developers to remember to create all possible symbols in advance.

RitaMAllenCA:
Install_or_update_contract should use a factory pattern.  
First instance - creates a package; subsequent calls update the version.  
Note: previous packages are not disabled by design - contract owner has to explicitly disable previous versions. Insert_contract_version: Key for contract version is manufactured: protocol version major + contract version.  This is because these contracts come from users and they dictate the versioning scheme.   
Want to remove this Current limitation: if protocol version > major version of the contract --> do not allow it to run.  In 2.0, if it CAN run, let it run. 

Requested change: WASM = Binary.   Engine loads it and calls an entry point "call" with a "verb" argument.   Op code based - incur a cost to find specific [byte] address of a cost code.  Includes a host side table of addresses/registers to [assembly] JUMP to and begin executing (emulating a call stack).    

self-replication: a contract can then install a contract (including itself).   Want a factory method to be called to install contracts.   Factory: Allows a single call to INSTALL a contract and another call to INVOKE a method on a contract to be executed.    It will NOT allow another factory contract for the same contract to be installed.   Thus, once a contract is installed, it can be called, but not replicated.    This should be native so gas costs are efficient.  Contracts are installed to disk for later use and active only when called. 

Session code: executes WASM code passed in and called in the context of the account (not a contract).  Stored session code: instead of passing the WASM every time, INSTALL the wasm as a stored session (set up with required entry points). 
Need tickets for: implementation, testing, examples, documentation.   
Host enforced: using metadata, have the host support (and enforce)

PR #4182, opened 2023-07-31
GH-2064: Factory pattern MVP

### LINKED ISSUE #4179: Implement factory pattern MVP

mpapierski:
Closes #4179 

This PR implements MVP of a host-side factory pattern. 

- New `EntryPointType::Install` which marks the entry point as an installer aka the factory method.
- New `EntryPointType::Normal` variant that defines default entry point type that in 2.0 will act as a 1.x -
- Both new variants act as a 1.x `EntryPointType::Contract`
- New `EntryPointAccess::Abstract` variant that marks the entry point export as exists in the wasm bytecode but is not callable. It can be referenced from a contract that is stored as `EntryPointType::Install`.
- It is possible for `EntryPointType::Install` to define the new `EntryPointType::Install` as in the factory of factories.

An example smart contract that utilizes this pattern can be found here: https://github.com/mpapierski/casper-node/blob/gh-2064-factory-pattern/smart_contracts/contracts/test/counter-factory/src/main.rs


cspramit:
@darthsiroftardis to review

execution_engine_testing/tests/src/test/counter_factory.rs
```diff
@@ -0,0 +1,204 @@
...
+        CONTRACT_FACTORY_ENTRY_POINT,
+        CONTRACT_FACTORY_DEFAULT_ENTRY_POINT,
+    ]);
+    assert_eq!(factory_wasm_exports, expected_entrypoints);
+
+    let contract_hash = (*contract_hash_key).into_contract_hash().unwrap();
+    let exec_request_1 = ExecuteRequestBuilder::contract_call_by_hash(
+        *DEFAULT_ACCOUNT_ADDR,
+        contract_hash,
+        "call", // should not be able to call "call" entry point
```
rafal-ch: Can we use `DEFAULT_ENTRY_POINT_NAME` here?

execution_engine_testing/tests/src/test/counter_factory.rs
```diff
@@ -0,0 +1,204 @@
...
+const DECREASE_ENTRY_POINT: &str = "decrement";
+const INCREASE_ENTRY_POINT: &str = "increment";
+const ARG_INITIAL_VALUE: &str = "initial_value";
+const ARG_NAME: &str = "name";
+const NEW_COUNTER_1_NAME: &str = "new-counter-1";
+const NEW_COUNTER_2_NAME: &str = "new-counter-2";
+
+#[ignore]
+#[test]
+fn should_install_factory() {
```
rafal-ch: It looks like this test does more than just install. I think splitting it to more focused tests could be an overkill, because it'll cause a lot of copy&pasted boilerplate, but maybe you can add a comment describing the scenario exercised by this test? Something like:
```
1. Execute factory contract
2. Make sure we can't call "call" (with brief explanation why)
3. Make sure we can't call "increase" endpoint, because...
4. etc.
```

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -841,7 +841,7 @@ fn faucet_costs() {
     // This test will fail if execution costs vary.  The expected costs should not be updated
     // without understanding why the cost has changed.  If the costs do change, it should be
     // reflected in the "Costs by Entry Point" section of the faucet crate's README.md.
-    const EXPECTED_FAUCET_INSTALL_COST: u64 = 75_226_865_930;
+    const EXPECTED_FAUCET_INSTALL_COST: u64 = 75_227_558_260;
```
rafal-ch: I think we also need to update the info in `smart_contracts/contracts/explorer/faucet/README.md`

types/src/contracts/entry_point_type.rs
```diff
@@ -0,0 +1,71 @@
...
+use num_traits::FromPrimitive;
+#[cfg(feature = "json-schema")]
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::bytesrepr::{self, FromBytes, ToBytes};
+
+/// Context of method execution
+///
+/// First bit represents version i.e.
```
rafal-ch: Isn't it the "last" bit, technically? :)
Maybe consider referring to it as "most significant bit"?

types/src/contracts/entry_point_type.rs
```diff
@@ -0,0 +1,71 @@
...
+    Session = 0b00000000,
+    /// Runs within contract's context
+    Contract = 0b00000001,
+    /// Pattern of entry point types introduced in 2.0.
+    ///
+    /// If this bit is missing, that means given entry point type was defined in pre-2.0 world.
+    /// Installer entry point.
+    Install = 0b10000000,
+    /// Normal entry point.
+    Normal = 0b10000010,
```
rafal-ch: Could it be set to `Normal = 0b10000001`?

execution_engine/src/runtime/mod.rs
```diff
@@ -2512,14 +2532,21 @@ where
...
-            self.context.validate_uref(uref).is_ok()
-        })
+        match access {
+            EntryPointAccess::Public => Ok(()),
+            EntryPointAccess::Groups(groups) => {
+                runtime_context::validate_group_membership(package, groups, |uref| {
+                    self.context.validate_uref(uref).is_ok()
+                })
+            }
+            EntryPointAccess::Abstract => Err(Error::NoSuchMethod(name.to_string())),
```
darthsiroftardis: Feels like we should add a bespoke error here, `Error::InvalidEntryPointAccess` , I feel like no such method is sightly misleading

execution_engine_testing/tests/src/test/contract_api/get_call_stack.rs
```diff
@@ -268,6 +269,7 @@ fn assert_each_context_has_correct_call_stack_info_module_bytes(
             EntryPointType::Session => {
                 builder.get_call_stack_from_session_context(&stored_call_stack_key)
             }
+            _ => unreachable!(),
```
darthsiroftardis: Can we handle this differently than an `unreachable`

execution_engine/src/runtime/mod.rs
```diff
@@ -1081,6 +1081,11 @@ where
             }
             (EntryPointType::Session, EntryPointType::Contract)
             | (EntryPointType::Contract, EntryPointType::Contract) => Ok(contract_hash.into()),
+
```
darthsiroftardis: I think we need to add another case here for `Installer` calling into `Session`

types/src/access_rights.rs
```diff
@@ -33,13 +32,24 @@ bitflags! {
...
-        const ADD_WRITE      = Self::ADD.bits  | Self::WRITE.bits;
+        const ADD_WRITE      = Self::ADD.bits()  | Self::WRITE.bits();
         /// Permission to read, add to, or write the value under the associated `URef`.
-        const READ_ADD_WRITE = Self::READ.bits | Self::ADD.bits | Self::WRITE.bits;
+        const READ_ADD_WRITE = Self::READ.bits() | Self::ADD.bits() | Self::WRITE.bits();
+    }
+}
+
+#[cfg(feature = "datasize")]
+impl DataSize for AccessRights {
```
darthsiroftardis: How come we didn't need to implement this before??

darthsiroftardis (approved):
Looks good, however, I think we can get rid of the `EntryPointType::Normal` as it seems to be no different from the `Contract` variant

execution_engine/src/runtime/mod.rs
mpapierski: Added `AbstractMethod` variant (that follows language of `NoSuchMethod`)

execution_engine_testing/tests/src/test/contract_api/get_call_stack.rs
mpapierski: Fixed

execution_engine/src/runtime/mod.rs
mpapierski: Good catch; fixed

types/src/access_rights.rs
mpapierski: I was about to update the bitfield crate which required all derives to be explicit, although I didn't commit to it due to gas costs and dubious value

rafal-ch (approved):
LGTM :+1:

execution_engine/src/engine_state/mod.rs
```diff
@@ -2553,6 +2553,7 @@ fn should_charge_for_errors_in_wasm(execution_result: &ExecutionResult) -> bool
                 | ExecError::NoActiveContractVersions(_)
                 | ExecError::InvalidContractVersion(_)
                 | ExecError::NoSuchMethod(_)
+                | ExecError::AbstractMethod(_)
```
EdHastingsCasperAssociation: I don't want to devolve to bike shedding, but I pretty strongly feel that  AbstractMethod is not a good term to use here; because that exact term has a specific general meaning, I think that it strongly suggests that there is some kind of inheritance model at play. However, there is no such model here, and the overall semantics are entirely different as well. So, I think it's confusing language.

While I can't think of a _perfect_ name off hand that precisely captures what's going on here, I came up with `SourceMethod` or `TemplateMethod` or perhaps `IntegrantMethod` (part of a whole), by reasoning from effect.

EdHastingsCasperAssociation (approved):
Aside from my concern over the term AbstractMethod, LGTM

PR #4184, opened 2023-07-31
Rename Delete to Prune

### LINKED ISSUE #4175: Rename Transform::Delete to Transform::Purge
In order to unify language around purging entries from the global state, we should rename Transform::Delete into Purge, as we don’t delete anything from the global state, it’s just removed from the tip of the trie and historical records aren’t affected.

mpapierski:
Closes #4175 

Renames `Transform::Delete` to `Transform::Purge` and also changes APIs incl. rustdocs to align to the same language (Delete keys -> Purge keys etc.)


execution_engine/src/shared/transform.rs
```diff
@@ -236,7 +236,7 @@ impl Transform {
                     Err(StoredValueTypeMismatch::new(expected, found).into())
                 }
             },
-            Transform::Delete => {
+            Transform::Purge => {
                 // Delete does not produce new values, it just consumes a stored value that it
```
EdHastingsCasperAssociation: nit: comment still says Delete

execution_engine/src/core/engine_state/execution_effect.rs
```diff
@@ -31,7 +31,7 @@ impl From<ExecutionJournal> for ExecutionEffect {
                 | Transform::AddUInt256(_)
                 | Transform::AddUInt512(_)
                 | Transform::AddKeys(_) => ops.insert_add(key, Op::Add),
-                Transform::Delete => ops.insert_add(key, Op::Delete),
+                Transform::Purge => ops.insert_add(key, Op::Purge),
```
EdHastingsCasperAssociation: Should be `Prune` not `Purge` right?

types/src/execution_result.rs
```diff
@@ -586,7 +586,7 @@ impl Transform {
             Transform::AddKeys(_) => TransformTag::AddKeys,
             Transform::Failure(_) => TransformTag::Failure,
             Transform::WriteUnbonding(_) => TransformTag::WriteUnbonding,
-            Transform::Delete => TransformTag::Delete,
+            Transform::Purge => TransformTag::Delete,
```
EdHastingsCasperAssociation: TransformTag::Prune

EdHastingsCasperAssociation (approved):
approved, assuming verb is changed from Purged to Prune

PR #4190, opened 2023-08-02
node/KeepUp: don't check TTL synced if node is not synced back past start of active era

### LINKED ISSUE #4160: Investigate failed sync test
The nctl sync_test.sh failed in this nightly run: https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8714/1/4

alsrdn:
It is possible that if a node just caught up, its highest orphaned block is within the current active era. We need to let the node sync back past the start of the active era and only after that we can determine if it is synced to TTL.

Fixes: https://github.com/casper-network/casper-node/issues/4160

### DIFF
--- a/node/src/types/max_ttl.rs
+++ b/node/src/types/max_ttl.rs
@@ -29,22 +29,14 @@ impl MaxTtl {
     /// Determine if orphaned block header is older than ttl requires.
     pub fn synced_to_ttl(
         &self,
-        latest_switch_block_timestamp: Timestamp,
+        latest_complete_switch_block_timestamp: Timestamp,
         highest_orphaned_block_header: &BlockHeader,
     ) -> Result<bool, String> {
         if highest_orphaned_block_header.is_genesis() {
             Ok(true)
         } else {
-            // It is possible that if a node just caught up, its highest orphaned block
-            // is within the current active era. We need to let the node sync back past the
-            // start of the active era and only after that we can determine if it is synced
-            // to TTL.
-            if latest_switch_block_timestamp < highest_orphaned_block_header.timestamp() {
-                return Ok(false);
-            }
-
             self.ttl_elapsed(
-                latest_switch_block_timestamp,
+                latest_complete_switch_block_timestamp,
                 highest_orphaned_block_header.timestamp(),
             )
         }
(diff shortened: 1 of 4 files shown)


EdHastingsCasperAssociation (approved):

node/src/types/max_ttl.rs
```diff
@@ -35,6 +35,14 @@ impl MaxTtl {
...
             Ok(true)
         } else {
+            // It is possible that if a node just caught up, its highest orphaned block
+            // is within the current active era. We need to let the node sync back past the
+            // start of the active era and only after that we can determine if it is synced
+            // to TTL.
+            if latest_switch_block_timestamp < highest_orphaned_block_header.timestamp() {
+                return Ok(false);
+            }
+
```
EdHastingsCasperAssociation: with era length of ~2hrs and TTL of much greater than that I don't think this is necessary...so, am I misreading it or is this guard condition for a scenario where era length exceeds TTL or sth?

Fraser999 (approved):

Fraser999 (approved):

node/src/types/max_ttl.rs
alsrdn: Actually it was not the complete solve for the issue because we should determine an unbroken sequence of complete blocks from the start of the current era. The latest switch block header that was used may not have been from a complete block.
In the particular case of the sync test failure, the buddha block to which the node was first caught up was towards the middle of the current era. But since we did have the header of the last block of the previous era (got through sync leap or if the global state sync takes a long time to complete) we were comparing the old timestamp with the newer timestamp of the highest orphaned block which is the buddha block we just caught up to.

If we only check the TTL awareness using blocks from the complete range we don't have this issue anymore because if the last switch block is complete, it means that the highest orphaned block is either that switch block or a block that is older than it, so the previous check isn't needed anymore.

Fixed in this commit: https://github.com/casper-network/casper-node/pull/4190/commits/e0e41960028931f9d469c0bb142e4ca19239db68

EdHastingsCasperAssociation: Lets discuss in standup tomorrow. I want to make sure I'm understanding the full picture before proceeding.

alsrdn:
closed in favor of https://github.com/casper-network/casper-node/pull/4195

PR #4196, opened 2023-08-05
Feat 2.0 bid rework

EdHastingsCasperAssociation:
the main thrust of this change is, delegator bids are stored as top-level entries in global state under a composite key instead of embedded in a collection within their applicable validator's bid; this allows each bid to be interacted with atomically and avoids the deep copy issue of embedded collections within a stored value. there is a downside; getting total stake for a validator, reporting, and distributing rewards requires a lot more reads. this can be mitigated by adding a lazy last-era in-mem cache, which I propose we do in a separate optimization PR.

secondarily, empty (0 balance) bids and slashed bids are now pruned from the global state pathing (will not be included under the newly calculated root hash); this reduces width of a slice of global state at the tip of the chain. 

thirdly, the unbond experience is simplified. an attempt to unbond an amount greater than stake previously resulted in an error. the logic has been changed to simply unbond all of the stake when the attempted amount is higher rather than error. adjacent to this, previously an attempt to fully unbond was complicated by the fact that we re-stake rewards. there is an edge case of a fully unbonded bid gaining stake again due to the end of era reward for their last, leading to "dust" in their bonding purse. the logic now instead applies the delayed reward payout to the appropriate unbonding purse avoiding this issue.


EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

utils/global-state-update-gen/src/generic.rs
```diff
@@ -308,110 +333,177 @@ fn find_large_bids<T: StateReader>(
...
-    let check_bid = |bid: &Bid| {
-        let bid_delegators: BTreeMap<_, _> = bid
-            .delegators()
+    let unchanged_bid = |bid: &ValidatorBid, delegators: &Vec<Delegator>| {
+        let bid_delegators: BTreeMap<_, _> = delegators
             .iter()
-            .map(|(key, delegator)| (key.clone(), *delegator.staked_amount()))
+            .map(|delegator| {
+                (
+                    delegator.validator_public_key().clone(),
```
fizyk20: I think we want the `delegator_public_key` as the key here, don't we? Otherwise, if all the delegators are for our validator, we'll only get a single entry instead of the proper map.

utils/global-state-update-gen/src/generic.rs
```diff
@@ -308,110 +333,177 @@ fn find_large_bids<T: StateReader>(
...
 
-    let stake = *recipient.stake();
-    let new_bid = if let Some(old_bid) = state.get_bids().get(pub_key) {
-        let mut bid = Bid::unlocked(
-            pub_key.clone(),
-            *old_bid.bonding_purse(),
-            stake,
-            *recipient.delegation_rate(),
-        );
+    if let Some(existing_bid) = existing_bids.validator_bid(validator_public_key) {
```
fizyk20: This won't be `Some` if there's a `Unified` bid present for this validator, right? I think we have to handle that case, too, and differently than a non-existing bid.

utils/global-state-update-gen/src/generic/state_tracker.rs
```diff
@@ -63,6 +63,23 @@ impl<T: StateReader> StateTracker<T> {
...
+            BidKind::Delegator(delegator) => BidAddr::new_from_public_keys(
+                delegator.validator_public_key(),
+                Some(delegator.delegator_public_key()),
+            ),
+        };
+
+        let _ = self
+            .entries_to_write
+            .insert(bid_addr.into(), bid_kind.into());
+    }
```
fizyk20, line 88: I think this function should also update `self.staking`, so that it we never add an entry to `entries_to_write` without also reflecting the changes in the cache.

I see that you update `self.staking` separately in `set_bid`, but I think it would be safer to do this in this function. Also, if we write a bid for a validator/delegator with an existing bid, we should overwrite the old one in the cache instead of adding a new one, so it might be easier to have `self.staking` as a map instead of as a `Vec`.

utils/global-state-update-gen/src/generic/state_tracker.rs
```diff
@@ -208,134 +225,106 @@ impl<T: StateReader> StateTracker<T> {
...
+            Some(existing_bid) => {
+                //let previous_stake = self.get_purse_balance(existing_bid_kind.bonding_purse());
+                let previous_stake = existing_bid.staked_amount();
+                if existing_bid.bonding_purse() != bid_kind.bonding_purse() {
+                    self.set_purse_balance(existing_bid.bonding_purse(), U512::zero());
+                    self.set_purse_balance(bid_kind.bonding_purse(), previous_stake);
+                }
+                previous_stake
+            }
+        };
```
fizyk20: I think this might become more complex, because if the previously existing bid was a `Unified` one, and the new one will be a `Delegator` one, it will be hard to handle the case properly.

Maybe we should add a preprocessing step that converts all the `Unified` bids into `Validator` + `Delegator` kinds? So that the later steps, like this one, can just handle the new format instead of the old one... Or do we not want that?

fizyk20 (commented):
I've only looked at the `global-state-update-gen` part - I have a few comments there.

utils/global-state-update-gen/src/generic/state_tracker.rs
EdHastingsCasperAssociation: all the unified bids are being turned into Validator / Delegator bids in commit_upgrade

utils/global-state-update-gen/src/generic.rs
EdHastingsCasperAssociation: I think the legacy unified bids should have been already migrated to validator bids by this point in the logic?

utils/global-state-update-gen/src/generic/state_tracker.rs
EdHastingsCasperAssociation: > I think this function should also update self.staking, so that it we never add an entry to entries_to_write without also reflecting the changes in the cache. I see that you update self.staking separately in set_bid, but I think it would be safer to do this in this function. 

Are you positive about that? I believe I'm mirroring what the original code was doing in regards to bids_cache / staking and entries_to_write

utils/global-state-update-gen/src/generic.rs
EdHastingsCasperAssociation: Bart and I did a zoom session and reworked this section of the code together. Will be in next commit.

utils/global-state-update-gen/src/generic/state_tracker.rs
EdHastingsCasperAssociation: After discussion, Bart agreed with me on the main point, and I agreed with him on the secondary point.

EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

EdHastingsCasperAssociation:
bors try

storage/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license = "Apache-2.0"
 
 [dependencies]
 bincode = "1.3.1"
-casper-types = { version = "3.0.0", path = "../types", features = ["datasize", "json-schema", "std"] }
+casper-types = { version = "3.0.0", path = "../types", features = ["datasize", "json-schema", "std", "gens"] }
```
Fraser999: I don't think this change is needed.  Same point about `itertools` below.

storage/src/global_state/state/scratch.rs
```diff
@@ -113,6 +132,9 @@ impl StateReader<Key, StoredValue> for ScratchGlobalStateView {
     type Error = error::Error;
 
     fn read(&self, key: &Key) -> Result<Option<StoredValue>, Self::Error> {
+        if self.cache.read().unwrap().pruned.contains(key) {
+            return Ok(None);
+        }
         if let Some(value) = self.cache.read().unwrap().get(key) {
             return Ok(Some(value.clone()));
         }
```
Fraser999:
```suggestion
        {
            let cache = self.cache.read().unwrap();
            if cache.pruned.contains(key) {
                return Ok(None);
            }
            if let Some(value) = cache.get(key) {
                return Ok(Some(value.clone()));
            }
        }
```

storage/src/global_state/state/scratch.rs
```diff
@@ -166,7 +191,11 @@ impl StateReader<Key, StoredValue> for ScratchGlobalStateView {
         let mut ret = Vec::new();
         for result in keys_iter {
             match result {
-                Ok(key) => ret.push(key),
+                Ok(key) => {
+                    if !self.cache.read().unwrap().pruned.contains(&key) {
```
Fraser999: Might be more efficient to move the read lock outside the for loop?

storage/src/global_state/state/scratch.rs
```diff
@@ -232,7 +261,14 @@ impl CommitProvider for ScratchGlobalState {
                 }
             };
 
-            self.cache.write().unwrap().insert_write(key, value);
+            match instruction {
+                TransformInstruction::Store(value) => {
+                    self.cache.write().unwrap().insert_write(key, value);
```
Fraser999: Same with this maybe (i.e. acquire the write lock outside the for loop) since there shouldn't be contention on the lock.  Although that might change in the future I guess, so this one is more debatable.

storage/src/global_state/trie_store/operations/tests/delete.rs
Fraser999, line 1: I guess this file should be renamed `prune.rs`.

storage/src/global_state/trie_store/operations/tests/delete.rs
```diff
@@ -1,12 +1,12 @@
 use super::*;
-use crate::global_state::{transaction_source::Writable, trie_store::operations::DeleteResult};
+use crate::global_state::{transaction_source::Writable, trie_store::operations::PruneResult};
 
 fn checked_delete<K, V, T, S, E>(
```
Fraser999: `checked_prune`?

storage/src/global_state/trie_store/operations/tests/delete.rs
```diff
@@ -16,19 +16,19 @@ where
...
     }
     Ok(delete_result)
 }
 
 mod partial_tries {
     use super::*;
-    use crate::global_state::trie_store::operations::DeleteResult;
+    use crate::global_state::trie_store::operations::PruneResult;
 
     fn delete_from_partial_trie_had_expected_results<'a, K, V, R, S, E>(
```
Fraser999: `prune_from_partial_trie_had_expected_results`?

types/src/key.rs
```diff
@@ -813,8 +913,16 @@ impl FromBytes for Key {
...
+                match BidAddr::from_bytes(remainder) {
+                    // attempt new style first
+                    Ok((bid_addr, rem)) => Ok((bid_addr.into(), rem)),
+                    Err(_) => {
+                        // attempt legacy before failing
+                        let (account_hash, rem) = AccountHash::from_bytes(remainder)?;
+                        let bid_addr = BidAddr::legacy(account_hash.value());
+                        Ok((bid_addr.into(), rem))
+                    }
+                }
```
Fraser999: I think we need to rethink this approach, as it can give the wrong result.  For example, if we serialized with the previous (legacy) encoding, and the account hash happened to have its first byte as 0, 1, or 2, and there were further bytes in the stream after the serialized key bytes, then `BidAddr::from_bytes` will succeed, giving a bad value for the parsed key and consuming an extra byte of the remainder.

I think we'll need to add a new Key variant altogether (which would allow us to simplify the bytesrepr impls for `Key` and `BidAddr`) or look at a migration.  A simpler option might be to rename `KeyTag::Bid` to `KeyTag::LegacyBid` (keeping its value as `7`, and adding `KeyTag::Bid = 15`.  Then in `write_bytes()` we just have
```rust
Key::Bid(bid_addr) => bid_addr.write_bytes(writer),
```
and in `from_bytes()`:
```rust
tag if tag == KeyTag::LegacyBid as u8 => {
    let (account_hash, rem) = AccountHash::from_bytes(remainder)?;
    let bid_addr = BidAddr::legacy(account_hash.value());
    Ok((bid_addr.into(), rem))
}
tag if tag == KeyTag::Bid as u8 => {
    let (bid_addr, rem) = BidAddr::from_bytes(remainder)?;
    Ok((bid_addr.into(), rem))
}
```

Any of these alternatives seem like they could be used to get rid of the legacy aspect of the `BidAddr` since there'd be no use for it then I think.

types/src/system/auction.rs
```diff
@@ -51,3 +76,683 @@ pub type UnbondingPurses = BTreeMap<AccountHash, Vec<UnbondingPurse>>;
...
+
+        None
+    }
+}
+
+/// Bid Address
+#[derive(Default, PartialOrd, Ord, PartialEq, Eq, Hash, Clone, Copy, Serialize, Deserialize)]
+#[cfg_attr(feature = "datasize", derive(DataSize))]
+#[cfg_attr(feature = "json-schema", derive(JsonSchema))]
+pub struct BidAddr(pub AccountHash, pub Option<AccountHash>, BidAddrTag);
```
Fraser999: I think we should make these fields private since we have public getters for them.

Even better would be to turn this into an enum like
```rust
enum BidAddr {
    Validator(AccountHash),
    Delegator {
        validator: AccountHash,
        delegator: AccountHash,
    },
}
```

(as per comments on the `Key` enum, I don't think we need the legacy support in this type).  The advantage of using an enum is that we're guaranteed to not have a mismatch between the `BidAddrTag` value and the value of the `Option<AccountHash>`.  Just now, we could in theory have a `BidAddr(addr, None, 2)` which would bytesrepr serialize to `2, <32 bytes of addr>` and would subsequently fail to deserialize.  With the enum approach, this isn't possible.

Also, the JSON encoded form might be slightly more obvious (at the cost of being more verbose) with the enum approach.

types/src/system/auction.rs
```diff
@@ -51,3 +76,683 @@ pub type UnbondingPurses = BTreeMap<AccountHash, Vec<UnbondingPurse>>;
...
+        pair: ([u8; ACCOUNT_HASH_LENGTH], [u8; ACCOUNT_HASH_LENGTH]),
+    ) -> Self {
+        BidAddr(
+            AccountHash::new(pair.0),
+            Some(AccountHash::new(pair.1)),
+            BidAddrTag::Delegator,
+        )
+    }
+
+    #[allow(missing_docs)]
```
Fraser999: Should this be `#[doc(hidden)]` instead?

types/src/system/auction.rs
```diff
@@ -51,3 +76,683 @@ pub type UnbondingPurses = BTreeMap<AccountHash, Vec<UnbondingPurse>>;
...
+    }
+
+    /// If true, this instance is the key for a delegator bid record.
+    /// Else, it is the key for a validator bid record.
+    pub fn is_delegator_bid_addr(&self) -> bool {
+        self.1.is_some()
+    }
+
+    /// How long will be the serialized value for this instance.
+    pub fn serialized_length(&self) -> usize {
```
Fraser999: Couldn't this just be implemented in the `impl ToBytes`?

types/src/system/auction.rs
```diff
@@ -51,3 +76,683 @@ pub type UnbondingPurses = BTreeMap<AccountHash, Vec<UnbondingPurse>>;
...
+
+impl Display for BidAddr {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        let tag = self.tag();
+        match tag {
+            BidAddrTag::Unified => {
+                write!(f, "{}", self.0)
+            }
+            BidAddrTag::Validator | BidAddrTag::Delegator => match self.1 {
+                Some(delegator) => write!(f, "{}{}{}", tag, self.0, delegator),
```
Fraser999: Might be worth adding `-` between each `{}` here and in the line below?

types/src/system/auction.rs
```diff
@@ -51,3 +76,683 @@ pub type UnbondingPurses = BTreeMap<AccountHash, Vec<UnbondingPurse>>;
...
+        match self.0.to_bytes() {
+            Ok(bytes) => {
+                let mut ret = vec![];
+                ret.push(KeyTag::Bid as u8);
+                ret.push(BidAddrTag::Delegator as u8);
+                ret.extend(bytes);
+                Ok(ret)
+            }
+            Err(_) => Err(Error::Serialization),
+        }
```
Fraser999: Nit:
```suggestion
        let mut ret = Vec::with_capacity(self.0.serialized_length() + 2);
        ret.push(KeyTag::Bid as u8);
        ret.push(BidAddrTag::Delegator as u8);
        self.0.write_bytes(&mut ret)?;
        Ok(ret)
```

types/src/system/auction.rs
Fraser999, line 1: I think we should have unit tests for some of the more complex methods `BidsExt` and bytesrepr round trip tests for `BidAddr` and `BidKind`.

types/src/system/auction/validator_bid.rs
```diff
@@ -0,0 +1,499 @@
...
+    /// Updates the delegation rate of the provided bid
+    pub fn with_delegation_rate(&mut self, delegation_rate: DelegationRate) -> &mut Self {
+        self.delegation_rate = delegation_rate;
+        self
+    }
+
+    /// Sets given bid's `inactive` field to `false`
+    pub fn activate(&mut self) -> bool {
+        self.inactive = false;
+        false
```
Fraser999, line 220: Seems like this method could just return unit type.  Same for `deactivate` below?

utils/validation/Cargo.toml
```diff
@@ -14,6 +14,7 @@ hex = { version = "0.4.2", features = ["serde"] }
 serde = "1"
 serde_json = "1"
 thiserror = "1.0.18"
+rand = { version = "0.8.3", default-features = false, features = ["small_rng", "std_rng"] }
```
Fraser999: Looks like we don't need this change.

Fraser999 (commented):
Partial review - will complete asap.

types/src/system/auction.rs
EdHastingsCasperAssociation: Bid, ValidatorBid, and Delegator all have proptest and roundtrip tests already; I'll add same for BidKind. 

BidAddr has proptests, but I'll add bytesrepr round trip.

types/src/key.rs
EdHastingsCasperAssociation: the legacy records are migrated on upgrade, accessed by key iter, and on go-forward roothashes no longer exist. afterwards, afaik the only access of historical bids would be on old roothashes by key lookup. but, I don't want to fight about it and will change it to separate variant.

types/src/system/auction.rs
EdHastingsCasperAssociation: sure; all of the existing key payloads are simple types so I was following along. but I would also prefer an enum here. will make this change.

types/src/system/auction.rs
EdHastingsCasperAssociation: I will change it to this in next commit. is there actually a functional difference tho?

types/src/system/auction.rs
EdHastingsCasperAssociation: is also being called from Key to_bytes

types/src/system/auction.rs
EdHastingsCasperAssociation: its not clear to me that it _should_ be, but if you prefer it that way ok. will change.

types/src/system/auction/validator_bid.rs
EdHastingsCasperAssociation: This is straight up copy pasta from the original bid record, looks like Henry set it up this way back in 2021. the return value is used in auction logic, and I don't want to rework that original code over a nitpick unless you feel really strongly about it.

types/src/system/auction.rs
EdHastingsCasperAssociation: that breaks serde_roundtrip_key and serde_roundtrip_trie_leaf

types/src/system/auction/validator_bid.rs
Fraser999: I'm fine to leave as is, since this PR is already pretty meaty.

Fraser999 (approved):

execution_engine/src/engine_state/mod.rs
```diff
@@ -448,7 +478,60 @@ where
...
+                        let validator_public_key = existing_bid.validator_public_key();
+                        let validator_bid_addr = BidAddr::from(validator_public_key.clone());
+                        let validator_bid = ValidatorBid::from(*existing_bid.clone());
+                        borrow.write(
+                            validator_bid_addr.into(),
+                            StoredValue::BidKind(BidKind::Validator(Box::new(validator_bid))),
+                        );
+
+                        let delegators = existing_bid.delegators().clone();
+                        for (_, delegator) in delegators {
```
mpapierski, line 513:
```suggestion
                        for delegator in delegators.into_values() {
```

types/src/execution/transform_kind.rs
```diff
@@ -61,10 +93,14 @@ impl TransformKind {
     /// Applies the transformation on a specified stored value instance.
     ///
     /// This method produces a new `StoredValue` instance based on the `TransformKind` variant.
-    pub fn apply(self, stored_value: StoredValue) -> Result<StoredValue, TransformError> {
+    pub fn apply(self, stored_value: StoredValue) -> Result<TransformInstruction, TransformError> {
+        fn store(sv: StoredValue) -> TransformInstruction {
```
mpapierski, line 97: Is that necessary though? Why not just inline at call sites and do `TransformInstruction::Store(x)`

mpapierski (approved):
LGTM with minor nits,

types/src/execution/transform_kind.rs
EdHastingsCasperAssociation: indeed. initially it was doing more, but got thinned down to its current form of nothing much. i'll inline it.

Issue #4178, opened 2023-07-28
Address remaining TODOs related to block versioning


RitaMAllenCA:
Address remaining TODOs related to the usage of the new versioned block types in the node.

Issue #4176, opened 2023-07-28
Remove JsonBlock and related structs in favor generating JSON serialization and schemas from the Block structs directly


RitaMAllenCA:
We can now use [this crate](https://crates.io/crates/serde-map-to-array) to generate schemas for map type objects. Since we used `JsonBlock` because we had this limitation before, we should remove it and use the new crate for all map like fields for the block. This means that we don’t need to implement new versioned types of `JsonBlock` to work with the versioned `Block` anymore.

PR #4200, opened 2023-08-09
Fix `Block` data extensibility issues

alsrdn:
Address some of the remaining versioning issues with `Block`:
* Remove usage of the JSON representation structs for block data. Since now we can generate JSON schemas directly for the `Block` struct we don't need the additional structures.
* Use versioned block as validation metadata for approvals hashes
* Remove `From` conversion between a versioned block and `BlockV2` in favor of using `TryFrom`
* Register a block as a versioned block in the block synchronizer without doing conversion
* Don't forward blocks that are of an old version to the block accumulator since this component should only deal with gossiped blocks.
* Rename `VersionBlock` to `Block` and `VersionedBlockBody` to `BlockBody`

Fixes: https://github.com/casper-network/casper-node/issues/4176
Fixes: https://github.com/casper-network/casper-node/issues/4178


node/src/components/block_synchronizer.rs
```diff
@@ -478,14 +478,21 @@ impl BlockSynchronizer {
...
-                                        // TODO[RC] .into(), becasue we want to avoid polluting
-                                        // `MetaBlock` with `VersionedBlock`, but maybe we'd have to
-                                        let meta_block = MetaBlock::new(
-                                            Arc::new((*block).into()),
-                                            execution_results,
-                                            MetaBlockState::new_after_historical_sync(),
-                                        );
-                                        effect_builder.announce_meta_block(meta_block).await
+                                        // TODO[RC] .try_into(), becasue we want to avoid polluting
+                                        // `MetaBlock` with `NewBlock`, but maybe we'd have to
```
Fraser999, line 482:
```suggestion
```

node/src/components/block_synchronizer.rs
```diff
@@ -478,14 +478,21 @@ impl BlockSynchronizer {
...
+                                        match <Block as std::convert::TryInto<BlockV2>>::try_into(
+                                            *block,
+                                        ) {
```
Fraser999, line 485:
```suggestion
                                        match BlockV2::try_from(*block) {
```

node/src/effect.rs
```diff
@@ -1058,7 +1059,7 @@ impl<REv> EffectBuilder<REv> {
     }
 
     /// Puts the given block into the linear block store.
-    pub(crate) async fn put_block_to_storage(self, block: Arc<Block>) -> bool
+    pub(crate) async fn put_block_to_storage(self, block: Arc<BlockV2>) -> bool
```
Fraser999, line 1062: I think we should rename the methods here to reflect the renaming of the block types.

types/src/block/block_body.rs
```diff
@@ -1,10 +1,160 @@
...
+    fn eq(&self, other: &BlockBody) -> bool {
+        match (self, other) {
+            (BlockBody::V1(lhs), BlockBody::V1(rhs)) => lhs.eq(rhs),
+            (BlockBody::V2(lhs), BlockBody::V2(rhs)) => lhs.eq(rhs),
+            _ => {
+                panic!("BlockBody structs of different versions should not be compared")
+            }
+        }
+    }
+}
```
Fraser999, line 74: Can't we just derive `PartialEq` and `Eq` rather than this impl with a panic?

types/src/block/json_compatibility.rs
```diff
@@ -3,21 +3,6 @@
 //! JSON, and to allow for the production of a static schema for them.
 
 #![cfg(all(feature = "std", feature = "json-schema"))]
+mod json_block_with_signatures;
```
Fraser999, line 6: Now that we only have a single type in here, we could probably just delete the `json_compatibility` module and move the `json_block_with_signatures` into `block`.

node/src/components/block_synchronizer.rs
rafal-ch: This has been removed in [this commit](https://github.com/casper-network/casper-node/pull/4213/commits/d52746adfb6bbf17c908e869d6767d655e3ddbc0) in the [follow-up PR](https://github.com/casper-network/casper-node/pull/4213).

node/src/components/block_synchronizer.rs
rafal-ch: This has been updated in [this commit](https://github.com/casper-network/casper-node/pull/4213/commits/d52746adfb6bbf17c908e869d6767d655e3ddbc0) in the [follow-up PR](https://github.com/casper-network/casper-node/pull/4213).

node/src/effect.rs
rafal-ch: Function name, variable name and comment clean-up has been done in [this commit](https://github.com/casper-network/casper-node/pull/4213/commits/d52746adfb6bbf17c908e869d6767d655e3ddbc0) in the [follow-up PR](https://github.com/casper-network/casper-node/pull/4213).

types/src/block/block_body.rs
rafal-ch: Fixed in [this commit](https://github.com/casper-network/casper-node/pull/4213/commits/14bcef4cbe729a6359e13189a20c18c4c0149da9) in the [follow-up PR](https://github.com/casper-network/casper-node/pull/4213).

types/src/block/json_compatibility.rs
rafal-ch: Done in [this commit](https://github.com/casper-network/casper-node/pull/4213/commits/e4405a379212371534cf1ce28fa92174c8321f9d#diff-c4a2fcdb17b59c31c2cf823de68c714f9b8675659f2bd9428bb52f94207b0dd2) in the [follow-up PR](https://github.com/casper-network/casper-node/pull/4213).

rafal-ch:
Closed in favor if https://github.com/casper-network/casper-node/pull/4213

Issue #4202, opened 2023-08-09
Post 1.5.2 , dev branch joining nodes issue(ooming)


cspramit:
100+20 network has a change in behavior after 1.5.2 was released, where most of the non-validating joiner nodes were ooming, when using the dev branch to test

Following are the test results
Network: 100  + 20
Name: ajith-dev-big (HIGHWAY)
Branch: dev
Commit: 9b14e264fa660e4fd7492477dca00f40ebcaedfc
Config: {"config":{}, "chainspec":{"era_duration": \"60minutes\", "minimum_block_time": \"32768ms\", "consensus_protocol": \"Highway\" }}
Test: Native Transfer Load Testing
Highway Consensus
running at 32s block time
12k Native Transfers Periodically
Observations:
network behaviour has degraded since 1.5.2 release
Occasionally hitting 'Invalid Deploy' because of timestamp validation checks using casper-client 2.0.0
KeepUp nodes consuming all memory and OOMing
majority of KeepUp nodes OOMing and transitioning to 'CatchUp'
instance becomes unresponsive due to OOMing
node recovers once it re-starts
contract_runtime_run_execute_count metrics on some KeepUp nodes show extremely high count for executed deploys. (729425 (keepUp) vs 9048 (validate))
a lot of Warning/errors observed continuously
{"timestamp":"2023-08-08T06:52:58.585604Z","level":"WARN","fields":{"message":"failed to apply execution results","error":"attempt to apply execution results for already completed block_hash: block hash 6027..c015"},"target":"casper_node::components::block_synchronizer::block_acquisition"}
{"timestamp":"2023-08-08T06:41:40.469691Z","level":"WARN","fields":{"message":"BlockAcquisition: invalid attempt to register deploy for: block hash 6027..c015","deploy_id":"Id { deploy_hash: DeployHash(3d9a7a4c73eb361ec128104b61f3e4b08c9aec1b4ab23da594e03e509bd24893), approvals_hash: ApprovalsHash(7e51b6ee2a97aab4794b1e857704cb8cfc2f4a405bb8593c601b0bf3d1f8002f) }"},"target":"casper_node::components::block_synchronizer::block_acquisition"}
{"timestamp":"2023-08-08T06:41:40.455503Z","level":"WARN","fields":{"message":"invalid attempt to start block execution","block_hash":"block hash 6027..c015"},"target":"casper_node::components::block_synchronizer::block_builder"}
[...]

PR #4205, opened 2023-08-10
Attempt to fix block synchronizer request storm by tweaking the latching.

alsrdn:
Attempt to fix block synchronizer request storm by tweaking the latching.

Don't decrement the latch when responses for old requests that are no longer needed (are received in a state that expects other data).
For example if the synchonizer is the `HaveWeakFinality` state, it will try to fetch the block body from some peers. The latch will be incremented by `num_peers`. If a late finality signature response comes in this state it will decrement the latch. If multiple late responses come, the latch can become 0, and the synchronizer would try to fetch the block body again even if no response for the initial requests were received yet.

Added a test for this also. It's a bit ugly and copy-pasty but I think it showcases the issue.

Attempted fix for: https://github.com/casper-network/casper-node/issues/4202

### DIFF
--- a/node/src/components/block_synchronizer/block_builder.rs
+++ b/node/src/components/block_synchronizer/block_builder.rs
@@ -231,6 +231,11 @@ impl BlockBuilder {
         self.latch.count() > 0
     }
 
+    #[cfg(test)]
+    pub fn latch_count(&self) -> u8 {
+        self.latch.count()
+    }
+
     pub(super) fn check_latch(&mut self, interval: TimeDiff) -> bool {
         self.latch.check_latch(interval, Timestamp::now())
     }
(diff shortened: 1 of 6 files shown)


node/src/components/block_synchronizer/block_builder.rs
```diff
@@ -456,6 +461,23 @@ impl BlockBuilder {
         }
     }
 
+    pub(super) fn waiting_for_block_header(&self) -> bool {
+        match &self.acquisition_state {
+            BlockAcquisitionState::Initialized(..) => true,
```
EdHastingsCasperAssociation, line 466: an if let for these methods would probably be less annoying.

EdHastingsCasperAssociation (approved):

node/src/components/block_synchronizer/block_builder.rs
alsrdn: I've given some thought and while I agree it's annoying using a wildcard will be less helpful if we refactor this logic in the future to add/remove states. Having a match here would help guide future changes (at least force us to reason about it) so I would prefer if we leave it as is. WDYT?

node/src/components/block_synchronizer/block_builder/tests.rs
```diff
@@ -25,38 +25,38 @@ fn handle_acceptance() {
 
     // Builder acceptance for needed signature from ourselves.
     assert!(builder
-        .handle_acceptance(None, Ok(Some(Acceptance::NeededIt)))
+        .handle_acceptance(None, Ok(Some(Acceptance::NeededIt)), true)
```
rafal-ch, line 28: Can you add a single case to this test when the `should_unlatch` is `false`?

rafal-ch (approved):
LGTM :+1: 
No blockers

node/src/components/block_synchronizer/block_builder/tests.rs
alsrdn: Added a new test in https://github.com/casper-network/casper-node/pull/4205/commits/896653b1abdade591bd96cc092faa4199e98fbd0 and renamed the old one for more granular testing.

PR #4212, opened 2023-08-16
Update ed25519-dalek to v2

### LINKED ISSUE #4211: Resolve new rustsec warning
https://rustsec.org/advisories/RUSTSEC-2022-0093 is now affecting casper-types and all crates depending on it.  Ideally we should resolve the issue by upgrading ed25519-dalek, but if this is not feasible in a short time, we can instead verify our code is not vulnerable to the exploit and disable the warning, pending a proper fix when time permits.

Fraser999:
This resolves a new [RUSTSEC warning](https://rustsec.org/advisories/RUSTSEC-2022-0093).

Our own code should be updated in the future so that `pub fn sign` no longer takes a public key, but this is a breaking change and cannot be done in this PR.  In the meantime, the passed public key is ignored in all cases with the changes in this PR.

One unrelated test was affected by these changes (`inactive_validator_limited`), but this turned out to be an issue where random key generation was being done after the start of the test timer.  The test has been refactored to move the random generation to before the start of the timer.

Closes #4211.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -299,6 +299,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a4a4ddaa51a5bc52a6948f74c06d20aaaddb71924eab79b8c97a8c556e942d6a"
 
 [[package]]
+name = "base64ct"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


rafal-ch (approved):
LGTM :+1:

rafal-ch:
Should we request a ticket to modify `pub fn sign` as per description?

Fraser999:
bors merge

Issue #4177, opened 2023-07-28
Restructure MetaBlock to use VersionedBlock or don’t use MetaBlock for announcing historical blocks synced by the Block Synchronizer


RitaMAllenCA:
We use `MetaBlock` to send the synchronized block to the deploy buffer and consensus when a historical block is marked complete by the Block Synchronizer component. Since `MetaBlock` currently keeps track only of blocks created with the latest version, it can’t track a `VersionedBlock`. We can either make `MetaBlock` work with `VersionedBlock` or we can remove the usage of meta block from the Block Synchronizer and announce the block directly to the deploy buffer and consensus. Explore and implement the appropriate solution.

alsrdn:
Implemented in https://github.com/rafal-ch/casper-node/pull/19

PR #4213, opened 2023-08-16
Data extensibility block conversion

rafal-ch:
This is based on the work from @alsrdn here: https://github.com/casper-network/casper-node/pull/4200

It adds few commits to address the original review comments.

Fixes: https://github.com/casper-network/casper-node/issues/4176
Fixes: https://github.com/casper-network/casper-node/issues/4178


Fraser999 (approved):

EdHastingsCasperAssociation (changes requested):
i'd like to discuss this before it merges

rafal-ch:
As per discussion, the requested changes boil down to rename of the functions that deal with `block_v2` structs - these functions should be renamed to **not** include the `v2` in their names.

rafal-ch:
~This waits until the `dev` to `feat-2.0` merge is concluded.~

Alternatively, we might address the comment and merge this to feature branch without the rustsec fix for now.

rafal-ch:
As per the discussion with @Fraser999 and @alsrdn 

1. We plan to leave only a single function related to the extensible datatypes, for example: we'll leave just the `get_block()`
2. This function will always return the enum with version variants
3. While this may create some burden on the callers, there should not be that many different call-sites for this to actually be a huge issue (also, if this eventually becomes an issue, we may always decide to provide additional utility function like `fn get_most_recent_block_version()`, but these should be decided case-by-case).

The above refactor will be delivered as part of https://github.com/casper-network/casper-node/issues/4177, which is the data extensibility work planned for next sprint.

This PR will be merged to the feature branch (`feat-data-extensibility`) as is.
Then we'll merge `feat-data-extensibility` to `feat-2.0` (separate ticket will be requested) because we'd like to unlock work that requires extensible block.
Then, next PRs for data extensibility will go directly into `feat-2.0`.

cc: @EdHastingsCasperAssociation if this sounds ok, please unlock the merge.

EdHastingsCasperAssociation (approved):

PR #4228, opened 2023-08-22
Merge `dev` into `feat-2.0`

rafal-ch:
Bring the recent changes from `dev` (mainly, the private chain support and fix for the slow historical sync) into `feat-2.0`.


Fraser999 (approved):

execution_engine/src/engine_state/mod.rs
```diff
@@ -1611,6 +1653,51 @@ where
...
+        {
+            Some(key) => *key,
+            None => return Ok(ExecutionResult::precondition_failure(Error::Deploy)),
+        };
+
+        let payment_purse_uref = payment_purse_key
+            .into_uref()
+            .ok_or(Error::InvalidKeyVariant)?;
+
+        // let purse_balance_key = match tracking_copy
```
alsrdn: Not sure why we left this block commented out. Is this coming from dev?

alsrdn (approved):

node/src/components/consensus/highway_core/finality_detector/rewards.rs
```diff
@@ -0,0 +1,337 @@
+use tracing::trace;
```
AlexanderLimonov: This file should not be reintroduced, it's only relevant for "old style" rewards that are removed entirely in 2.0 for both protocols

AlexanderLimonov (changes requested):
See comment on reintroducing code slated for removal

node/src/components/consensus/highway_core/finality_detector/rewards.rs
rafal-ch: Thanks, good catch. File and the associated code is removed in [this](https://github.com/casper-network/casper-node/pull/4228/commits/48c9d1e1fa45af727b19df2d6d7c8c9ddf0836c1) commit, **however** I added a `TODO` to actually take care of the private chain support in the 2.0 rewards scheme. Please have a look @AlexanderLimonov as this may require additional ticket related to 2.0 rewards.

execution_engine/src/engine_state/mod.rs
rafal-ch: This check is now moved several lines down the function. It was duplicated after the merge, so I commented it out with the intention to remove it completely after making sure it's not needed. And I just forgot to remove it, which is now fixed in https://github.com/casper-network/casper-node/pull/4228/commits/f0dd96ad29af01e267c23c334d157247203f9b5e.

mpapierski (approved):

AlexanderLimonov (approved):

node/src/components/consensus/highway_core/finality_detector/rewards.rs
AlexanderLimonov: One could use the new rewards just as one would old rewards for private chains

PR #4229, opened 2023-08-23
Debug info about contract execution on `Trap::Unreachable`

### LINKED ISSUE #4002: Use runtime::print in panic handler when test-support and debug is enabled

### LINKED ISSUE #4003: Print current stack height after receiving Unreachable trap when test-support is enabled

rafal-ch:
This PR adds some debug info output in case error is encountered during contract execution:

```
running 1 test
runtime stack overflow, current=501, max=500
test test::stack_overflow::runtime_stack_overflow_should_cause_unreachable_error ... ok
```

Closes https://github.com/casper-network/casper-node/issues/4002
Closes https://github.com/casper-network/casper-node/issues/4003

The scope has been slightly changed due to the discussion, currently, when `Trap::Unreachable` is caught we compare if the runtime stack has overflown and only then output the error message.


rafal-ch:
Changes in this PR made the facet install cost to become variable. Tested on three different machines it gives three different results:
```
PC_1: 84_977_768_930
PC_2: 84_967_688_930
PC_3: 84_957_608_930
```

This is under investigation.

rafal-ch:
The problem with varying build sizes was resolved by remapping the `home` and `pwd` directories [here](https://github.com/casper-network/casper-node/pull/4229/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R18).

execution_engine/src/core/runtime/mod.rs
```diff
@@ -2994,3 +3000,22 @@ where
...
     }
 }
+
+#[cfg(feature = "test-support")]
+fn dump_debug_info(error: &wasmi::Error, instance: wasmi::ModuleRef) {
+    // Currently, we're only interested in the details for the `Unreachable` error.
+
+    if let wasmi::Error::Trap(trap) = error {
+        if let TrapKind::Unreachable = trap.kind() {
+            eprintln!("Wasm execution error: TrapCode::Unreachable");
```
Fraser999: I don't think we need this personally, but if you prefer to keep it, maybe it should say TrapKind rather than TrapCode?

execution_engine/src/core/runtime/mod.rs
```diff
@@ -2994,3 +3000,22 @@ where
...
+            eprintln!(
+                "current instance stack height = {}",
+                if let Some(stack_height) = instance.globals().last() {
+                    format!("{:?}", stack_height.get())
+                } else {
+                    "unknown".to_string()
+                }
+            );
```
Fraser999: As discussed offline, I think it would be less confusing if this function only printed an error referencing the stack height if the stack height exceeded the configured `max_stack_height`.  In that case, I'd prefer if the logged message was more like `stack height limit of {max_stack_height} exceeded`.  Not sure if there's even any point to printing the actual stack height, as I'd expect it should be `max_stack_height + 1`, but maybe worth confirming that?

rafal-ch: Updated in 7f8dcebbde140e9be8b10500c7fb9ac5ff355fe7

execution_engine/src/core/runtime/mod.rs
rafal-ch: Removed in https://github.com/casper-network/casper-node/commit/7f8dcebbde140e9be8b10500c7fb9ac5ff355fe7

alsrdn (approved):

PR #4257, opened 2023-08-31
option to opt out of historical data sync

EdHastingsCasperAssociation:
adds NoSync option, which allows a node to opt out of historical sync via config setting.

### DIFF
--- a/node/src/reactor/main_reactor/keep_up.rs
+++ b/node/src/reactor/main_reactor/keep_up.rs
@@ -37,6 +37,7 @@ pub(super) enum KeepUpInstruction {
     Fatal(String),
 }
 
+#[derive(Debug, Clone, Copy)]
 enum SyncBackInstruction {
     Sync {
         sync_hash: BlockHash,
@@ -45,6 +46,7 @@ enum SyncBackInstruction {
     Syncing,
     TtlSynced,
     GenesisSynced,
+    NoSync,
 }
 
 impl Display for SyncBackInstruction {
@@ -56,6 +58,7 @@ impl Display for SyncBackInstruction {
             SyncBackInstruction::Syncing => write!(f, "syncing"),
             SyncBackInstruction::TtlSynced => write!(f, "ttl reached"),
             SyncBackInstruction::GenesisSynced => write!(f, "genesis reached"),
+            SyncBackInstruction::NoSync => write!(f, "configured to not sync"),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


resources/production/config-example.toml
```diff
@@ -7,7 +7,7 @@
 #trusted_hash = 'HEX-FORMATTED BLOCK HASH'
 
 # Whether to synchronize all data back to genesis, or just [deploys.max_ttl] (chainspec setting) worth of data.
-sync_to_genesis = true
+sync_handling = 'ttl'
```
Fraser999, line 18: It would be good to copy the comment block from `resources/local/config.toml` as it's more likely that will find its way into the operators' config.toml eventually I expect.

Fraser999 (approved):
One nitpick only.

resources/production/config-example.toml
EdHastingsCasperAssociation: good idea; done

node/src/reactor/main_reactor/keep_up.rs
```diff
@@ -129,6 +132,11 @@ impl MainReactor {
             }
         }
 
+        if self.sync_handling.is_no_sync() {
+            // node is not permitted to be a validator with no_sync behavior.
+            return None;
```
alsrdn, line 137: If a node has accumulated TTL worth of blocks in its available block range while keeping up successfully for a while shouldn't we allow it to validate?
I guess it would mean moving part of the TTL synced logic in this function maybe.

alsrdn (approved):
Looks ok to me, but have a question about allowing node to validate.

node/src/reactor/main_reactor/keep_up.rs
EdHastingsCasperAssociation: no, because if a validating node falls behind it drops back to leap again, then back fills gaps using historical sync to regain ttl compliance.

resources/production/config-example.toml
```diff
@@ -6,8 +6,16 @@
 # If set, use this hash as a trust anchor when joining an existing network.
 #trusted_hash = 'HEX-FORMATTED BLOCK HASH'
 
-# Whether to synchronize all data back to genesis, or just [deploys.max_ttl] (chainspec setting) worth of data.
-sync_to_genesis = true
+# Historical sync behavior for this node. Options are:
+#  'genesis'  (node will attempt to acquire all block data back to genesis)
+#  'ttl'      (node will attempt to acquire all block data to comply with time to live enforcement)
+#  'nosync'   (node will only acquire blocks moving forward)
```
KillianH, line 12: @EdHastingsCasperAssociation so a "nosync" node will still acquire blocks etc in the storage.lmdb ? A small improvement would be to clean the storage lmdb on restart to regain previous blocks acquired. (The improvement will be really small tho')

Issue #3231, opened 2022-08-02
Change EE test suite to default to DEFAULT_LOCKED_FUNDS_PERIOD_MILLIS  and DEFAULT_VESTING_SCHEDULE_PERIOD_MILLIS


cspramit:
These values were only chosen for casper’s public chain mainnet and are long obsolete. Tests involving schedule period should change the config to provide 90 days of locked funds period and 13 weeks of vesting schedule explicitly. Other tests can default to zeros.

Refer github discussion - https://github.com/casper-network/casper-node/pull/3219#discussion_r933492149

PR #4258, opened 2023-08-31
Fix after private-chain merge

mpapierski:
After doing the same merge three basically times (feat-1.6, #4109 and #4228) there's two issues fixed in this PR after #4109 :

- Due to conflicts, VFTA tests for locked amounts were incorrect (default should be 0 lockup, tests should exercise 90 days) - this was introduced in #3231, but conflicts in private chain changes made it revert to 90days.
- Payment errors penalty could've been paid to block proposer regardless of `fee_handling`/`refund_handling` option.

Once merged I'll apply this to both feat-1.6 (again) and feat-2.0


Fraser999 (approved):

rafal-ch (approved):
LGTM :+1:

PR #4259, opened 2023-09-01
Merge feat private chain to 1.6 fix

mpapierski:
Similar to #4258, the same fix is applied to feat-1.6.


marc-casperlabs:
I've merged this into 0c805d64a on the fix test failures branch (`4210-fix-failing-test-failures`), which is supposed to turn CI green. Thank you!

Issue #4255, opened 2023-08-30
Fix chain_get_block_transfers RPC so that it returns an empty array of transfers instead of null if the block didn’t have any transfers.


RitaMAllenCA:
Issue was reported [here](https://casper-labs-team.slack.com/archives/CDXBCHB6W/p1693372367760869). When a joiner node is synchronizing historical blocks, if the block doesn’t have any deploys, we don’t write an empty record in the transfers DB. When trying to get the transfers for such block through the chain_get_block_transfers RPC, it will return null since no such record was written while syncing. We need to adjust the RPC (or the storage method for getting transfers) to return an empty array in this case. We also shouldn’t be storing empty records preferably if there are no successful transfers in that block.

PR #4265, opened 2023-09-04
storage: return an empty transfers array on `GetBlockTransfers`

alsrdn:
If there is no record in the transfers DB for a particular block hash and that block is truly empty, we can return an empty array of transfers when calling `GetBlockTransfers` instead of returning `None`.

Fixes: https://github.com/casper-network/casper-node/issues/4255

### DIFF
--- a/node/src/components/storage.rs
+++ b/node/src/components/storage.rs
@@ -853,10 +853,26 @@ impl Storage {
                 block_hash,
                 responder,
             } => {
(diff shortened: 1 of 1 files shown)


alsrdn:
Closing this PR in favor of https://github.com/casper-network/casper-node/pull/4272. Since this has been written, we discovered a new issue with transfer reporting. The PR referenced here fixes both this issue and the new one.

Issue #4240, opened 2023-08-25
[block extensibility] Merge feat-data-extensibility to feat-2.0


RitaMAllenCA:
Once the work for the block data extensibility is complete, we want to merge it in the feat-2.0 branch. This is not trivial because feat-2.0 has advanced quite a bit since the data extensibility work has started on a separate branch. Also, merging to feat-2.0 requires BORS and nightly tests to pass. Because the schemas changed, it’s unlikely that all the NCTL tests will pass out of the box and will require changes to make them compatible with the new schemas, or we will need to have some changes to the client.

PR #4267, opened 2023-09-05
Data extensibility for `Block` and `BlockBody`

rafal-ch:
Merge data extensibility PoC (`Block` and `BlockBody`) into `feat-2.0`.

Fixes: #4240


rafal-ch:
`do-not-merge`, still need to figure out the NCTL tests

**Update:** the `upgrade_scenario_01.sh` (which is currently the only one executed by `nctl_upgrade.sh` is not affected by the changes to block RPC schema and doesn't require client updates. There is a "[temporary hack](https://github.com/rafal-ch/casper-node/commit/72ce59cbfeb8179bc375c2f0be7ec1d8048b84ef)" prepared on the side branch which we can use to move forward, should the refactoring of `MetaBlock` require more work than expected.

With the mentioned commit in place, the NCTL upgrade scenario passes successfully on the `feat-data-extensibility` branch.

cc: @alsrdn

node/src/components/gossiper/provider_impls/block_provider.rs
```diff
@@ -36,11 +36,15 @@ impl ItemProvider<Block> for Gossiper<{ Block::ID_IS_COMPLETE_ITEM }, Block> {
...
         item_id: BlockHash,
-    ) -> Option<Box<Block>> {
+    ) -> Option<Box<BlockV2>> {
         // TODO: Make `get_block_from_storage` return a boxed block instead of boxing here.
-        effect_builder
-            .get_block_from_storage(item_id)
-            .await
-            .map(Box::new)
+        if let Some(block) = effect_builder.get_block_from_storage(item_id).await {
+            match block {
```
alsrdn: nit: we can skip matching here `std::convert::TryInto::<BlockV2>::try_into(block).ok().map(|block| Box::new(block));`

types/src/block.rs
```diff
@@ -3,54 +3,57 @@ mod block_hash;
...
+#[cfg(any(feature = "testing", test))]
+pub mod test_block_builder;
 
-#[cfg(any(all(feature = "std", feature = "testing"), test))]
-use alloc::collections::BTreeMap;
 use alloc::{boxed::Box, vec::Vec};
 use core::fmt::{self, Display, Formatter};
-#[cfg(any(all(feature = "std", feature = "testing"), test))]
-use core::iter;
+#[cfg(feature = "once_cell")]
```
alsrdn: Shouldn't this be `#[cfg(feature = "json-schema")]` instead since we use `Lazy` only for defining `static BLOCK` below?

alsrdn (approved):
One small I think with a feature guarded use, but otherwise LGTM.

node/src/components/gossiper/provider_impls/block_provider.rs
rafal-ch: I used a little more concise form: `block.try_into().ok().map(Box::new)`
Updated here: d0b7db9e4d3a34794f455a209655d4869ee36077

types/src/block.rs
rafal-ch: Yes, good catch. Fixed in 60d1b481e6b1ea24a92343c49ce6b724cb15f6b9 and 06fff783c44ae7a795387e3cc60c63bee9784c93 along with the update to `json_block_with_signatures.rs`.

Fraser999 (approved):

Issue #4247, opened 2023-08-28
Fix validator broadcast efficiency in feat-1.6


RitaMAllenCA:
The validator broadcast based on the ValidatorMatrix in feat-1.6 is currently not fixed and was recently tested with a workaround, causing the broadcast to reach all nodes:
rust
      for peer_id in self.outgoing_manager.connected_peers() {
            total_outgoing_manager_connected_peers += 1;

            if !self.validator_matrix.has_era(&era_id)
                || self.outgoing_limiter.is_validator_in_era(era_id, &peer_id)
            {
                total_connected_validators_in_era += 1;
                self.send_message(peer_id, msg.clone(), None)
            }
        }
(in the workaround, the if condition is prefixed with true ||)
The root cause for this code not working is that it does not check adjacent eras (past and future), causing it to break during switch blocks.
The code needs to be reexamined and reengineered to efficiently send at least to current and previous era validators, as well as potential future validators, all without introducing quadratic copies of validator IDs on every message send.

marc-casperlabs:
The feature (selecting nodes based on their validator status) needs to be cleanly readded to the networking component, quick fixes are not going to cut it here.

Issue #3939, opened 2023-05-08
Improve back-pressure `Ticket` pressure across lifecycle cross components


RitaMAllenCA:
Right now the node eagerly drops tickets potentially too early in the process. To make it reflect actual load more accurately, we should ensure that a `Ticket` from processing a request is dropped exactly when all of its resource use has concluded.

Whether it is feasible to drop the `Ticket` approach entirely in favor of making proper request-response patterns offered by `juliet` fixing this issue remains a matter of scope.

devendran-m:
Housekeeping: Removed the automated sprint allocation and moved it back to triage.

devendran-m:
Triage update: Not Condor related. Therefore removed from Condor/2.0.0 release.

PR #4271, opened 2023-09-05
1.6 feature merge

### LINKED ISSUE #3941: Merge feat-1.6 to dev

marc-casperlabs:
This PR is a summary merge of every new feature and fix that has not made it to `dev` yet and is currently slated for a 1.6 release. Parts of the code have been reviewed already, but this does not apply to the entirety of the changes nor their combination.

The PR contains multiple larger changesets:

* The `juliet` crate and its integration into the node, forming the new base networking for the 1.6 release. There are some open TODOs, feel free to point them out, but follow-ups have been ticketed in most, if not all cases.
* EE changes:
  * Prunes withdraw keys post migration
  * Wasmi is updated to 0.13.2
  * Pruning is exposed as a new transform and could be found in the journal
  * Better error messages when trying to run Wasm  with non-MVP opcodes
* Smaller fixes, expanded test coverage (e.g. around metrics) and utility features and refactorings have been added.

While the review may seem a little daunting, if focusing on a single aspect at once (EE, juliet, integration of juliet and misc changes) should make it more manageable.

History is preserved, keeping some features that never made it to mainline, thus the high number of commits, along with numerous merge commits of `dev` back into the branch for compatibility.

Closes #3941.


Cargo.toml
```diff
@@ -29,15 +33,17 @@ default-members = [
...
 [profile.bench]
 codegen-units = 1
 lto = true
+
+[profile.release-with-debug]
+inherits = "release"
+debug = true
+
+[patch.crates-io]
+datasize = { git = "https://github.com/casperlabs/datasize-rs", rev = "2b980c05af5553522dde5f2751e5a0fd3347d881" }
```
Fraser999: Do we need this?

node/CHANGELOG.md
```diff
@@ -89,9 +93,15 @@ All notable changes to this project will be documented in this file.  The format
   * `execution_queue_size` to report the number of blocks enqueued pending execution
   * `accumulated_(outgoing|incoming)_limiter_delay` to report how much time was spent throttling other peers.
 * Add `testing` feature to casper-node crate to support test-only functionality (random constructors) on blocks and deploys.
+* The network handshake now contains the hash of the chainspec used and will be successful only if they match.
+* Add an `identity` option to load existing network identity certificates signed by a CA.
+* TLS connection keys can now be logged using the `network.keylog_location` setting (similar to `SSLKEYLOGFILE` envvar found in other applications).
+* Add a `lock_status` field to the JSON representation of the `ContractPackage` values.
+* Unit tests can be run with JSON log output by setting a `NODE_TEST_LOG=json` environment variable.
```
Fraser999: Shouldn't this be in the `Unreleased` section?

Also, the change to the formatting (in terms of blank lines) could do with being reverted.  We normally have 3 blank lines between versions, one blank line after each subsection, and no blank line after `###` headers.  This style is (meant to be) consistent across all the changelogs in the repo.

node/Cargo.toml
Fraser999, line 1: Was the change in formatting done via a tool?  If so, can we apply it consistently across all manifests, since they currently have a consistent style afaik (and even add it to CI as we do for rustfmt?)

node/Cargo.toml
```diff
@@ -119,20 +137,60 @@ doctest = false
...
-    ["../resources/maintainer_scripts/logrotate.d/casper-node", "/etc/logrotate.d/casper-node", "644"],
-    ["../resources/maintainer_scripts/pull_genesis.sh", "/etc/casper/pull_genesis.sh", "755"],
-    ["../resources/maintainer_scripts/delete_local_db.sh", "/etc/casper/delete_local_db.sh", "755"],
-    ["../resources/maintainer_scripts/config_from_example.sh", "/etc/casper/config_from_example.sh", "755"],
-    ["../resources/maintainer_scripts/systemd_pre_start.sh", "/etc/casper/systemd_pre_start.sh", "755"],
-    ["../resources/production/README.md", "/etc/casper/README.md", "644"],
-    ["../resources/production/CHANGE_LOG.md", "/etc/casper/CHANGE_LOG.md", "644"],
-    ["../resources/production/config-example.toml", "/etc/casper/config-example.toml", "644"],
-    ["../resources/production/validator_keys/README.md", "/etc/casper/validator_keys/README.md", "644"]
+    [
```
Fraser999: I don't think any of the style changes here are improvements, but this change in particular is horrible imho :)

node/src/components/consensus/highway_core/active_validator.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::arithmetic_side_effects)]
```
Fraser999, line 1: Do we have a ticket to revert this and the other similar `allow` changes?

node/src/components/contract_runtime.rs
```diff
@@ -270,11 +270,16 @@ impl ContractRuntime {
...
+        TrieRequestIncoming {
+            sender,
+            message,
+            ticket,
+        }: TrieRequestIncoming,
     ) -> Effects<Event>
     where
         REv: From<NetworkRequest<Message>> + Send,
     {
+        drop(ticket); // TODO: Properly handle ticket.
```
Fraser999, line 282: Is this ticketed?  I'd hate for us to forget to drop `drop(ticket)` by not ticketing "drop `drop(ticket)`".

node/src/components/network.rs
```diff
@@ -419,7 +403,8 @@ where
 
         for peer_id in self.outgoing_manager.connected_peers() {
             total_outgoing_manager_connected_peers += 1;
-            if self.outgoing_limiter.is_validator_in_era(era_id, &peer_id) {
+
+            if true {
```
Fraser999: I imagine this is ticketed too?

node/src/components/network.rs
```diff
@@ -439,12 +424,14 @@ where
...
         rng: &mut NodeRng,
         msg: Arc<Message<P>>,
-        gossip_target: GossipTarget,
+        _gossip_target: GossipTarget,
         count: usize,
         exclude: HashSet<NodeId>,
     ) -> HashSet<NodeId> {
-        let is_validator_in_era =
-            |era: EraId, peer_id: &NodeId| self.outgoing_limiter.is_validator_in_era(era, peer_id);
+        // TODO: Restore sampling functionality. We currently override with `GossipTarget::All`.
```
Fraser999: Also ticketed?

node/src/components/network.rs
```diff
@@ -488,23 +475,70 @@ where
         &self,
         dest: NodeId,
         msg: Arc<Message<P>>,
-        opt_responder: Option<AutoClosingResponder<()>>,
+        _opt_responder: Option<AutoClosingResponder<()>>, // TODO: Restore functionality or remove?
```
Fraser999: Also ticketed?

node/src/components/network.rs
```diff
@@ -599,24 +633,66 @@ where
...
 
+                // If given a key, determine validator status.
+                let validator_status = peer_consensus_public_key.as_ref().map(|public_key| {
+                    let status = self
+                        .validator_matrix
+                        .is_active_or_upcoming_validator(public_key);
+
+                    // Find the shared `Arc` that holds the validator status for this specific key.
+                    match self.incoming_validator_status.entry((**public_key).clone()) {
+                        // TODO: Use `Arc` for public key-key.
```
Fraser999: Not sure it's worth it, given we plan to make public keys copyable again soon.

node/src/components/network.rs
```diff
@@ -599,24 +633,66 @@ where
...
                     .instrument(span)
-                    .event(move |result| Event::IncomingClosed {
-                        result,
-                        peer_id: Box::new(peer_id),
-                        peer_addr,
-                        span: boxed_span,
+                    .event(move |result| {
+                        // We keep the client around, even though we do not use it, since dropping
+                        // it will cause the connection to be closed from our end.
+                        drop(rpc_client);
```
Fraser999: The comment here doesn't align with the code does it?

node/src/components/network.rs
```diff
@@ -753,14 +838,19 @@ where
             OutgoingConnection::Established {
                 peer_addr,
                 peer_id,
-                peer_consensus_public_key,
-                sink,
-                is_syncing,
+                peer_consensus_public_key: _, // TODO: Use for limiting or remove.
```
Fraser999: Ticketed?

node/src/components/network/connection_id.rs
```diff
@@ -240,6 +129,7 @@ impl ConnectionId {
     ///
     /// The `flag` should be created using the [`Role::in_flag`] or [`Role::out_flag`] method and
     /// must be created accordingly (`out_flag` when serializing, `in_flag` when deserializing).
+    #[allow(dead_code)] // TODO: Re-add if necessary when connection packet tracing is readded.
```
Fraser999: Ticketed?

node/src/components/network/event.rs
```diff
@@ -23,16 +24,20 @@ use crate::{
 };
 
 const _NETWORK_EVENT_SIZE: usize = mem::size_of::<Event<ProtocolMessage>>();
-const_assert!(_NETWORK_EVENT_SIZE < 65);
+const_assert!(_NETWORK_EVENT_SIZE < 999); // TODO: This used to be 65 bytes!
```
Fraser999: Ticketed?

node/src/components/network/message.rs
```diff
@@ -952,6 +892,15 @@ mod tests {
         roundtrip_certificate(false)
     }
 
+    #[test]
+    fn channels_enum_does_not_have_holes() {
+        for idx in 0..Channel::COUNT {
+            let result = Channel::from_repr(idx as u8);
+            eprintln!("idx: {} channel: {:?}", idx, result);
```
Fraser999: Left in by mistake?

node/src/protocol.rs
```diff
@@ -112,53 +112,44 @@ impl Payload for Message {
...
+            Message::DeployGossiper(_) => Channel::BulkGossip,
+            Message::AddressGossiper(_) => Channel::BulkGossip,
+            Message::GetRequest {
+                tag,
+                serialized_id: _,
+            } => match tag {
+                Tag::Deploy => Channel::DataRequests,
+                Tag::LegacyDeploy => Channel::SyncDataRequests,
+                Tag::Block => Channel::DataRequests,
+                Tag::BlockHeader => Channel::DataRequests,
```
Fraser999: I think these could be classed as sync data, since validators always create their own blocks via execution.

Fraser999 (commented):
Mostly nitpicks and questions around todos being ticketed.

I skipped review of the juliet crate as I have already done that on the feature branch and I assume it hasn't changed since then.

We do still need to decide if the crate gets its own repo or not.  If not, then points made in the previous review which should be addressed include:
* decide the crate's name
* specify `edition = "2018"` (or upgrade all manifests in the repo)
* add standard items to manifest's `[package]` section (especially the `license` field)
* update the README.md to include the standard blurb at the top and the license info
* add the crate to the casper-updater tool and `publish_to_crates_io.sh`

Cargo.toml
marc-casperlabs: The patch not, but `datasize` needs a new release.

marc-casperlabs: I've published https://crates.io/crates/datasize/0.2.15

marc-casperlabs: Fixed in 7beeda7cc

node/src/protocol.rs
marc-casperlabs: Let's do that, then!

marc-casperlabs: Fixed in d51a1f396

node/src/components/network/message.rs
marc-casperlabs: :grimacing:

marc-casperlabs: No, I don't see any issues in printing to `stderr` or `stdout` in tests, as one should only see these message when the test fails...

Alas, I think that's not how our CI works. I'll remove it.

marc-casperlabs: 0db3e98a5

node/src/components/network/event.rs
marc-casperlabs: Fixed it instead in 7c7a3ff7c

node/src/components/network/connection_id.rs
marc-casperlabs: I've removed the feature in 7c913ae6d instead.

node/src/components/network.rs
marc-casperlabs: Yes, under https://github.com/casper-network/casper-node/issues/4247

marc-casperlabs: See also: f78d9cc34

node/src/components/network.rs
marc-casperlabs: It's just worded weirdly. "Keep it around" here means moving it into the `.event` callback, which is only run after `task::message_receiver` has concluded, which will happen if the other peer closes the connection, or an error occurs.

As a result, `rpc_client` is only dropped when the connection has been closed anyway -- if it was dropped earlier, the actual act of doing so would cause the connection to be closed.

I'll update the wording on the comment.

marc-casperlabs: See b7f6a04b6

node/src/components/network.rs
marc-casperlabs: That does not change the fact that they are of respectable size (~ 100 bytes? I may be wrong). Multiple copies of 100 bytes for ~ 1000 connections or so does add up.

node/src/components/network.rs
marc-casperlabs: Updated in 04817ff54, although pending an addtional fix.

marc-casperlabs: Finished in 717057ce3, `juliet` to be refactored.

marc-casperlabs: Final fix in 5644e9119.

You may want to look at diff b7f6a04b6..5644e9119 when reviewing.

node/src/components/network.rs
marc-casperlabs: Yes, part of #4247

node/src/components/network.rs
marc-casperlabs: Yes, #4247.

node/CHANGELOG.md
marc-casperlabs: The change in Markdown formatting (blank lines) is unintentional. I realize we have a non-standard way (compare keepachangelog.com) of formatting these, but at one time, I must have missed to hit "save without formatting" in my IDE. Apologies :(

> Shouldn't this be in the Unreleased section?

Probably, must be a merge issue.

node/Cargo.toml
marc-casperlabs: Personally I don't particularly care how stuff is formatted, as long as I am not the one that has to manually reformat :(.

I'll try to rein in the auto-formatter after fixing this by hand.

node/src/components/network.rs
Fraser999: After we update them, they'll be 32 or 33 bytes.

node/src/effect.rs
```diff
@@ -669,8 +669,20 @@ impl<REv> EffectBuilder<REv> {
 
     /// Sends a network message.
     ///
-    /// The message is queued and sent, but no delivery guaranteed. Will return after the message
-    /// has been buffered in the outgoing kernel buffer and thus is subject to backpressure.
+    /// The message is queued and sent, without any delivery guarantees. Will return after the
+    /// message has been buffered by the networking stack and is thus is subject to backpressure
```
Fraser999:
```suggestion
    /// message has been buffered by the networking stack and is thus subject to backpressure
```

node/src/components/network.rs
marc-casperlabs: That sounds much more reasonable - and it would have also affected other corner cases I had to fix where event sizes suddenly grew, often due to the size of `PublicKey`.

I'd leave this comment in still, whoever makes the public key change will then be able to clean it up?

juliet/src/varint.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    // `for` is not stable in `const fn` yet.
+    let mut idx = 0;
+    while idx < input.len() {
+        let c = input[idx];
+        if idx >= 4 && c & 0b1111_0000 != 0 {
+            return Fatal(Overflow);
+        }
+
+        value |= ((c & 0b0111_1111) as u32) << (idx * 7);
```
rafal-ch: Is it a `VARINT_MASK` here?

juliet/src/varint.rs
```diff
@@ -0,0 +1,315 @@
...
+
+        if value < 16384 {
+            return 2;
+        }
+
+        if value < 2097152 {
+            return 3;
+        }
+
+        if value < 268435456 {
```
rafal-ch: values in hex would be less cryptic :)

rafal-ch: I think it might be worth adding a UT for the values at the boundary, such as:
```
    #[test]
    fn foo() {
        assert_eq!(Varint32::length_of(268435455), 4);
        assert_eq!(Varint32::length_of(268435456), 5);
    }
```

juliet/src/header.rs
```diff
@@ -0,0 +1,406 @@
...
+        if header.is_error() {
+            if (header.kind_byte() & Self::KIND_ERR_MASK) >= ErrorKind::COUNT as u8 {
+                return None;
+            }
+        } else {
+            if (header.kind_byte() & Self::KIND_MASK) >= Kind::COUNT as u8 {
+                return None;
+            }
+
+            // Ensure the 4th bit is not set, since the error kind bits are superset of kind bits.
```
rafal-ch: It's the 3rd bit, isn't it?

rafal-ch (commented):
Initial review

rafal-ch:
There were some issue observed during manual run of the `network_soundness.py` (some or all nodes get stuck over time). It'll be good to resolve that before merging to `dev` in order not to break the nightly pipeline. Details have been discussed with @marc-casperlabs .

node/src/components/contract_runtime.rs
marc-casperlabs: Fair enough. I'll ensure something gets added.

marc-casperlabs: Actually, https://github.com/casper-network/casper-node/issues/3939 already exists.

marc-casperlabs: I'll close this discussion, I have updated the linked ticket to be a bit more concrete about this.

node/src/components/consensus/highway_core/active_validator.rs
marc-casperlabs: Why do we want to revert them again? IIRC they are compatible with the current clippy and this is the correct form going forward? Or am I mistaken in this assumption?

node/Cargo.toml
marc-casperlabs: fixed in 7d22ce02e

node/CHANGELOG.md
marc-casperlabs: Fixed in 7d22ce02e

juliet/src/header.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// Peer sent a request cancellation exceeding the cancellation allowance.
+    #[error("cancellation limit exceeded")]
+    CancellationLimitExceeded = 13,
+}
+
+/// Frame kind, from the kind byte.
+#[derive(Copy, Clone, Debug, EnumCount, EnumIter, Eq, FromRepr, PartialEq)]
+#[cfg_attr(test, derive(proptest_derive::Arbitrary))]
+#[repr(u8)]
+
```
mpapierski:
```suggestion
```

juliet/src/header.rs
```diff
@@ -0,0 +1,406 @@
...
+            if (header.kind_byte() & Self::KIND_ERR_MASK) >= ErrorKind::COUNT as u8 {
+                return None;
+            }
+        } else {
+            if (header.kind_byte() & Self::KIND_MASK) >= Kind::COUNT as u8 {
+                return None;
+            }
+
+            // Ensure the 4th bit is not set, since the error kind bits are superset of kind bits.
+            if header.0[0] & Self::KIND_MASK != header.0[0] {
```
mpapierski: Why not

```suggestion
            if header.kind_byte() & Self::KIND_MASK != header.kind_byte() {
```

juliet/src/varint.rs
mpapierski: Or better yet (more readable IMO) `2<<N` syntax i.e. 2<<10, 2<<20, 2<<28 etc.

cspramit:
timeline on or before Monday 9th @SaiProServ

juliet/src/header.rs
marc-casperlabs: The four lower bits of `kind_byte()` are used if `header.is_error()`, but only the lower three bits if the header is NOT an error. This is due to the fact that there are a lot more errors than kind variants. Setting the fourth bit on a non-error (e.g. `0b0000_1000`) is a spec violation.

The check here is supposed to take `kind_byte`, AND-ing it with `KIND_MASK`, which is `0b0000_0111`. Thus only the lowest three bits are preserved. The `!=` check ensures that kind bite is equal to `0b0000_0XXX` by exiting early if it isn't.

We know the highest four are 0, because the highest indicates error status (and `is_error() == false`). The next three (7th, 6th and 5th) are reserved and always zero. So effectively we are checking whether the 4th bit is zero.

juliet/src/varint.rs
marc-casperlabs: Don't you mean `1<<N`? At least `1<<7` makes more sense to me here, since we're using 7 bits to encode the number.

Changed that way in `9cee0fef2`.

marc-casperlabs: > values in hex would be less cryptic :)

Well, in that case I'd like to present exhibit A:

> I think it might be worth adding a UT for the values at the boundary, such as:

```
        assert_eq!(Varint32::length_of(268435455), 4);
        assert_eq!(Varint32::length_of(268435456), 5);
```

`268435455` is `0xfffffff`, while `268435456` is `10000000`. Which is what `decode_known_values_and_crossover_points` tests :)

juliet/src/varint.rs
marc-casperlabs: Fixed in 8d3536a53

marc-casperlabs:
bors try

marc-casperlabs:
bors try

marc-casperlabs:
bors try

marc-casperlabs:
Will be reopened cleanly in a separate PR.

PR #4272, opened 2023-09-05
Fix issues in storage

### LINKED ISSUE #4255: Fix chain_get_block_transfers RPC so that it returns an empty array of transfers instead of null if the block didn’t have any transfers.

### LINKED ISSUE #4268: Investigate and fix issue with missing transfers
As per discussion in https://casper-labs-team.slack.com/archives/CDXBCHB6W/p1693793962701529.

Fraser999:
This PR fixes two separate issues:
* a historical synced block with no deploys would have no entry written for it in the `transfers_db`, causing the JSON-RPC `chain_get_block_transfers` to report the block's transfers as `null` rather than `[]`
* for a block with a non-zero number of successful transfers, if its execution results were put to storage more than once, the initial valid entry in the `transfers_db` would be overwritten with an invalid empty collection, causing the JSON-RPC `chain_get_block_transfers` to report the block's transfers as `[]` rather then the correct value

Th fix is to have `Storage` reconstitute the correct value when reading from `transfers_db` if it encounters a missing value or an empty collection, and store the corrected value before returning it.  The `Storage::write_execution_results` method has also been corrected to not overwrite good values with empty collections.

Closes #4255.
Closes #4268.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,7 @@ lint-smart-contracts:
 
 .PHONY: audit-rs
 audit-rs:
-	$(CARGO) audit
+	$(CARGO) audit --ignore RUSTSEC-2022-0093
 
 .PHONY: audit-as
 audit-as:
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -11,6 +11,14 @@ All notable changes to this project will be documented in this file.  The format
 
 
 
+## 1.5.2-alt
+
+### Fixed
+* Fix issue in `chain_get_block_transfers` JSON-RPC where blocks with no deploys could be reported as having `null` transfers rather than `[]`.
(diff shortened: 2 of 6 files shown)


alsrdn (approved):

rafal-ch (approved):
LGTM :+1:

Fraser999:
bors merge

PR #4280, opened 2023-09-12
Replace `gov96` nctl test

### LINKED ISSUE #4151: Investigate gov96 nctl test failure
The test hung [in this run](https://drone-auto-casper-network.casperlabs.io/casper-network/casper-node/8676/1/4).  The run was killed, so there are no logs available - we will need to attempt to replicate locally.

This is related to the sentinal node testing.

Fraser999:
This PR replaces the nctl `gov96` test with a new `main_reactor` integration test.

The test asserts that after all but one node is killed, the network can continue once the dead nodes are restarted.

I also took the opportunity to slightly refactor the reactor test framework in order to hopefully make it a little more ergonomic and require less boilerplate in the actual tests.  This is in anticipation of other nctl tests being replaced by tests using this framework.  As a side-effect, the total time for the main_reactor tests has reduced from ~13 mins to ~4 mins, despite adding a new test case.

Closes #4151.


ci/nightly-test.sh
```diff
@@ -117,7 +117,6 @@ start_run_teardown "emergency_upgrade_test.sh"
 start_run_teardown "emergency_upgrade_test_balances.sh"
 start_run_teardown "upgrade_after_emergency_upgrade_test.sh"
 start_run_teardown "sync_test.sh timeout=500"
-start_run_teardown "gov96.sh"
```
rafal-ch, line 120: It seems like we can also remove the `itst02.sh` test, which is similar to `gov96.sh`.

rafal-ch (approved):
LGTM :+1:

EdHastingsCasperAssociation (approved):

PR #4286, opened 2023-09-13
Fix query_balance to handle addressable_entity

### LINKED ISSUE #4284: Fix query_balance JSON-RPC
This RPC needs to be updated after the AddressableEntity changes, since the request fails if a public key or account-hash is provided.  This is due to the RPC server expecting an account to be stored under Key::Account whereas for recent global state, it is actually a pointer to an AddressableEntity

darthsiroftardis:
CHANGELOG:

- Fixed the `query_balance` RPC method to handle the introduction of `AddressableEntity`

Manually verified by spinning up an NCTL network and running the casper-client against the 2.0 network

Closes #4284


Fraser999 (approved):

Fraser999:
bors merge

PR #4290, opened 2023-09-15
Add timeouts to `juliet` crate

### LINKED ISSUE #4139: Support timeouts in juliet's rpc layer
Currently there is no concept of time in the juliet crate. It would be great to move all related functionality from the node into the rpc (and potentially io) layer, specifically:
~~[ ] Attempting to send an error reply should eventually timeout to force a connection close~~
[ ] Supporting timeouts on requests. A request timing out should be a regular outcome of the request, similar to a response cancellation. Care must be taken to ensure that timed out requests are cancelled as early as possible (e.g. before they leave the requesting peer).
~~[ ] Built-in ping support. This may be in the form of just a heartbeat, letting the timeout requests do the heavy lifting.~~

marc-casperlabs:
Closes #4139.

Adds support for timeouts to the `juliet` crate. Note that this function is not yet used in the node (`with_timeout` is never called), so there is no direct impact yet.

It also renames one function (hence the node change), adds a lot more traceability and some additional tests, fixing some bugs in the process.

### DIFF
--- a/juliet/src/io.rs
+++ b/juliet/src/io.rs
@@ -116,7 +116,7 @@ impl Display for QueuedItem {
                 if let Some(payload) = payload {
                     write!(f, ", payload: {}", PayloadFormat(payload))?;
                 }
-                f.write_str(" }}")
+                f.write_str(" }")
             }
             QueuedItem::RequestCancellation { io_id } => {
                 write!(f, "RequestCancellation {{ io_id: {} }}", io_id)
@@ -130,7 +130,7 @@ impl Display for QueuedItem {
                 if let Some(payload) = payload {
                     write!(f, ", payload: {}", PayloadFormat(payload))?;
                 }
-                f.write_str(" }}")
+                f.write_str(" }")
             }
             QueuedItem::ResponseCancellation { channel, id } => {
                 write!(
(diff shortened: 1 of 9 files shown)


juliet/src/io.rs
```diff
@@ -268,6 +331,38 @@ pub enum IoEvent {
...
+            }
+            IoEvent::ReceivedResponse { io_id, payload } => {
+                write!(f, "ReceivedResponse {{ io_id: {}", io_id)?;
+                if let Some(ref payload) = payload {
+                    write!(f, ", payload: {}", PayloadFormat(payload))?;
+                }
+                f.write_str(" }")
+            }
+            IoEvent::ReceivedCancellationResponse { io_id } => {
+                write!(f, "RequestCancalled {{ io_id: {} }}", io_id)
```
Fraser999:
```suggestion
                write!(f, "ReceivedCancellationResponse {{ io_id: {} }}", io_id)
```

juliet/src/rpc.rs
```diff
@@ -626,17 +749,67 @@ impl Drop for IncomingRequest {
...
+struct DrainConditional<'a, T, F> {
+    /// Heap to be drained.
+    heap: &'a mut BinaryHeap<T>,
+    /// Predicate function to determine whether or not to drain a specific element.
+    predicate: F,
+}
+
+/// Removes ites from the top of a heap while a given predicate is true.
+///
+/// Will take items from `heap` as long as `predicate` evaluates to `true`.
```
Fraser999:
```suggestion
/// Removes items from the top of a heap while a given predicate is true.
```

Fraser999 (approved):
Couple of nitpicks, but otherwise LGTM.

PR #4291, opened 2023-09-15
Speed up message sending by spawning fewer tasks

marc-casperlabs:
Greatly reduces the number of tasks spawned by attempting to send every message directly at least once. This should be the most common case for 99% of all messages if not under heavy load.

### DIFF
--- a/node/src/components/network.rs
+++ b/node/src/components/network.rs
@@ -490,37 +490,52 @@ where
             };
             trace!(%msg, encoded_size=payload.len(), %channel, "enqueing message for sending");
 
-            if let Some(responder) = message_queued_responder {
-                let client = connection.rpc_client.clone();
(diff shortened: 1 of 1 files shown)


node/src/components/network.rs
```diff
@@ -490,37 +490,52 @@ where
...
+                        None => {
+                            // This should never happen.
+                            error!("payload unexpectedly disappeard");
+                            return;
+                        }
+                        Some(payload) => payload,
+                    };
+
+                    if let Some(responder) = message_queued_responder {
+                        // Reconstruct client the client.
```
Fraser999:
```suggestion
                        // Reconstruct the client.
```

Fraser999 (approved):
One nitpick, otherwise LGTM.

PR #4295, opened 2023-09-18
Add event queue insight to log when there are many events in the queue

rafal-ch:
When the scheduler queue grows too much, the node will dump a basic breakdown of queue content into the log, like so:
```
Current event queue size (26) is above the threshold (25): details [("FinalitySignature", 3), ("FromStorage", 1), ("NetworkIncoming", 21), ("Regular", 1)]
```

The threshold is read from the environment variable: `CL_EVENT_QUEUE_DUMP_THRESHOLD`. If this variable is not set, this function is disabled.

Additionally, in order to prevent log flooding, the next dump will happen only if there are 10% more events in the queue than at the time of creating the previous dump.

This is a performance debug mechanism to support fixing issues with 16s blocks.

### DIFF
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -498,7 +498,7 @@ where
         }
 
         let event_queue_dump_threshold =
-            env::var("EVENT_QUEUE_DUMP_THRESHOLD").map_or(None, |s| s.parse::<usize>().ok());
+            env::var("CL_EVENT_QUEUE_DUMP_THRESHOLD").map_or(None, |s| s.parse::<usize>().ok());
 
         let scheduler = utils::leak(Scheduler::new(
             QueueKind::weights(),
--- a/node/src/reactor.rs
+++ b/node/src/reactor.rs
@@ -497,7 +497,13 @@ where
             );
         }
 
-        let scheduler = utils::leak(Scheduler::new(QueueKind::weights()));
+        let event_queue_dump_threshold =
+            env::var("EVENT_QUEUE_DUMP_THRESHOLD").map_or(None, |s| s.parse::<usize>().ok());
+
+        let scheduler = utils::leak(Scheduler::new(
+            QueueKind::weights(),
+            event_queue_dump_threshold,
+        ));
         let is_shutting_down = SharedFlag::new();
         let event_queue = EventQueueHandle::new(scheduler, is_shutting_down);
         let (reactor, initial_effects) = R::new(
--- a/node/src/utils/round_robin.rs
+++ b/node/src/utils/round_robin.rs
@@ -229,8 +229,8 @@ where
 
         // NOTE: Count may be off by one b/c of the way locking works when elements are popped.
         // It's fine for its purposes.
-        let total = self.queues.iter().map(|q| q.1.event_count()).sum::<usize>();
         if let Some(recent_event_count_peak) = &self.recent_event_count_peak {
+            let total = self.queues.iter().map(|q| q.1.event_count()).sum::<usize>();
             let recent_threshold = recent_event_count_peak.load(Ordering::SeqCst);
             if total > recent_threshold {
                 recent_event_count_peak.store(total, Ordering::SeqCst);
--- a/node/src/utils/round_robin.rs
+++ b/node/src/utils/round_robin.rs
@@ -17,8 +17,6 @@ use serde::Serialize;
 use tokio::sync::{Mutex, MutexGuard, Semaphore};
 use tracing::{debug, warn};
 
-const EVENT_QUEUE_INFO_THRESHOLD: usize = 4;
-
 /// Weighted round-robin scheduler.
 ///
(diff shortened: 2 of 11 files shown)


node/src/utils/round_robin.rs
```diff
@@ -220,6 +227,25 @@ where
             .push_back(item)
             .await;
 
+        // NOTE: Count may be off by one b/c of the way locking works when elements are popped.
+        // It's fine for its purposes.
+        let total = self.queues.iter().map(|q| q.1.event_count()).sum::<usize>();
```
Fraser999: I'd move this inside the `if let Some...` block below so we don't waste cycles calculating the total needlessly.

node/src/utils/round_robin.rs
```diff
@@ -220,6 +227,25 @@ where
...
+        if let Some(recent_event_count_peak) = &self.recent_event_count_peak {
+            let recent_threshold = recent_event_count_peak.load(Ordering::SeqCst);
+            if total > recent_threshold {
```
Fraser999: I wonder if it might be better to do something like
```rust
if total > (recent_threshold * 11) /10
```
so we only print if there's a 10% increase since the last time?

node/src/utils/round_robin.rs
```diff
@@ -220,6 +227,25 @@ where
...
+                recent_event_count_peak.store(total, Ordering::SeqCst);
+                let info: Vec<_> = self
+                    .queues
+                    .iter()
+                    .map(|q| (q.0.to_string(), q.1.event_count()))
+                    .filter(|(_, count)| count > &0)
+                    .collect();
+                warn!(
+                "Current event queue size ({total}) is above the threshold ({recent_threshold}): details {info:?}"
+            );
```
Fraser999: Formatting looks off here?

node/src/utils/round_robin.rs
```diff
@@ -43,6 +43,9 @@ pub struct WeightedRoundRobin<I, K> {
 
     /// Whether or not the queue is sealed (not accepting any more items).
     sealed: AtomicBool,
+
+    /// Only dump event counts when there are more of them than in the previous report.
```
Fraser999: I think it's worth commenting that event counts are only dumped if this member variable is `Some`.

node/src/reactor.rs
```diff
@@ -497,7 +497,13 @@ where
             );
         }
 
-        let scheduler = utils::leak(Scheduler::new(QueueKind::weights()));
+        let event_queue_dump_threshold =
+            env::var("EVENT_QUEUE_DUMP_THRESHOLD").map_or(None, |s| s.parse::<usize>().ok());
```
Fraser999: This env var should be documented in the top-level readme, under "Other environment variables" I guess.  I'd been trying to use a convention of `CL_` prefix for Casper env vars, but I don't think that has stuck, so don't worry about renaming this one :)

node/src/utils/round_robin.rs
rafal-ch: LLVM optimizer is usually good at taking care of such micro-optimizations, but I agree, we should be taking care of such things as well. Good catch!
Fixed here: f8b18ed3b45b978c95908995437ba25d41340596

node/src/utils/round_robin.rs
rafal-ch: Yes, good idea. Even with the basic checking I proposed we may still flood the logs. Update in https://github.com/casper-network/casper-node/pull/4295/commits/be55a4cbe914bf7b95db261e31bf855688dfc89a

node/src/utils/round_robin.rs
rafal-ch: Indeed. Fixed in https://github.com/casper-network/casper-node/pull/4295/commits/04c06d4b1e6913f6b15ab010aa9b2545bd725cd1

node/src/utils/round_robin.rs
rafal-ch: Comment updated in https://github.com/casper-network/casper-node/pull/4295/commits/46ac95c0df9ad2b05c63d22b7ffa959bf655327f

node/src/reactor.rs
rafal-ch: Variable is renamed and readme is updated in https://github.com/casper-network/casper-node/pull/4295/commits/e44daa4b3fa3d3c0b0993f78368fd248c276345c

Fraser999 (approved):

marc-casperlabs (commented):
I'm not a fan of adding this logic to the scheduler itself, as it bloats what is supposed to be a fairly small (🤞 ) type with a single concern. Is there no way of putting this into the reactor itself, or another auxiliary type if the state is too big? That's how it worked for the memory-pressure variant.

We could merge this, as I realize it is needed on short notice, but if you guys agree with my assessment, let's add an issue to clean this up later.

PR #4302, opened 2023-09-21
Use a standard regex in Makefile instead of a GNU specific one

real-felix:
This new regex does not use a GNU extension and also works on MacOS


fizyk20 (approved):

rafal-ch (approved):
LGTM :+1: 

But can you check if there are other places in our scripts that could use such an improvement?

PR #4315, opened 2023-10-03
ee: add support for emitting contract level messages

alsrdn:
Draft work for adding Contract level messages.
Added:
* 2 new FFIs that enable contracts to manage message topics and emit human readable string messages.
* topic names and hashes are added to the addressable entity for easy discovery
* for each topic there is a control record in global state under `Key::message_topic(entity_addr, topic_hash)` that stores the count of messages that were emitted  on a topic and the block time of the block where the deploys that emitted the messages have executed.
* each message checksum is stored in global state under `Key::message(entity_addr, topic_hash, message_index)` and can be discretely queried and verified.
* messages get sent out the SSE as part of `ExecutionResultV2` for each deploy.

Some tests are currently broken because the gas costs were not adjusted in the tests.
Will address these tests in a separate PR that will also implement linear gas cost increase for messages.

### DIFF
--- a/execution_engine/src/execution/executor.rs
+++ b/execution_engine/src/execution/executor.rs
@@ -126,12 +126,14 @@ impl Executor {
                 effects: runtime.context().effects(),
                 transfers: runtime.context().transfers().to_owned(),
                 cost: runtime.context().gas_counter(),
+                messages: runtime.context().messages(),
             },
             Err(error) => ExecutionResult::Failure {
                 error: error.into(),
                 effects: runtime.context().effects(),
                 transfers: runtime.context().transfers().to_owned(),
                 cost: runtime.context().gas_counter(),
+                messages: runtime.context().messages(),
             },
         }
     }
@@ -193,6 +195,7 @@ impl Executor {
         );
 
         let effects = tracking_copy.borrow().effects();
+        let messages = tracking_copy.borrow().messages();
 
         // Standard payment is executed in the calling account's context; the stack already
         // captures that.
@@ -203,12 +206,14 @@ impl Executor {
                 effects: runtime.context().effects(),
                 transfers: runtime.context().transfers().to_owned(),
                 cost: runtime.context().gas_counter(),
+                messages: runtime.context().messages(),
             },
             Err(error) => ExecutionResult::Failure {
                 effects,
                 error: error.into(),
                 transfers: runtime.context().transfers().to_owned(),
                 cost: runtime.context().gas_counter(),
+                messages,
             },
         }
     }
@@ -254,6 +259,7 @@ impl Executor {
         // Snapshot of effects before execution, so in case of error only nonce update
         // can be returned.
         let effects = tracking_copy.borrow().effects();
+        let messages = tracking_copy.borrow().messages();
 
         let entry_point_name = direct_system_contract_call.entry_point_name();
 
@@ -327,6 +333,7 @@ impl Executor {
                     effects: runtime.context().effects(),
                     transfers: runtime.context().transfers().to_owned(),
                     cost: runtime.context().gas_counter(),
+                    messages: runtime.context().messages(),
                 }
                 .take_with_ret(ret),
                 Err(error) => ExecutionResult::Failure {
@@ -334,6 +341,7 @@ impl Executor {
                     error: Error::CLValue(error).into(),
                     transfers: runtime.context().transfers().to_owned(),
                     cost: runtime.context().gas_counter(),
+                    messages,
                 }
                 .take_without_ret(),
             },
@@ -342,6 +350,7 @@ impl Executor {
                 error: error.into(),
                 transfers: runtime.context().transfers().to_owned(),
                 cost: runtime.context().gas_counter(),
+                messages,
             }
             .take_without_ret(),
         }
--- a/execution_engine/src/runtime/externals.rs
+++ b/execution_engine/src/runtime/externals.rs
@@ -8,6 +8,7 @@ use casper_types::{
     addressable_entity::{EntryPoints, NamedKeys},
     api_error,
     bytesrepr::{self, ToBytes},
+    contract_messages::MessageTopicOperation,
     crypto,
     package::{ContractPackageKind, ContractPackageStatus},
     system::auction::EraInfo,
@@ -1104,6 +1105,81 @@ where
 
                 let result = self.enable_contract_version(contract_package_hash, contract_hash)?;
 
+                Ok(Some(RuntimeValue::I32(api_error::i32_from(result))))
+            }
+            FunctionIndex::ManageMessageTopic => {
+                // args(0) = pointer to the topic name in wasm memory
+                // args(1) = size of the topic name string in wasm memory
+                // args(2) = pointer to the operation to be performed for the specified topic
+                // args(3) = size of the operation
+                let (topic_name_ptr, topic_name_size, operation_ptr, operation_size) =
+                    Args::parse(args)?;
+                self.charge_host_function_call(
+                    &host_function_costs.manage_message_topic,
+                    [
+                        topic_name_ptr,
+                        topic_name_size,
+                        operation_ptr,
+                        operation_size,
+                    ],
+                )?;
+
+                self.context
+                    .engine_config()
+                    .wasm_config()
+                    .messages_limits()
+                    .topic_name_size_within_limits(topic_name_size)
+                    .map_err(|e| Trap::from(Error::FailedTopicRegistration(e)))?;
+                let topic_name = self.string_from_mem(topic_name_ptr, topic_name_size)?;
+
+                if operation_size as usize > MessageTopicOperation::max_serialized_len() {
+                    return Err(Trap::from(Error::InvalidMessageTopicOperation));
+                }
+                let topic_operation = self
+                    .t_from_mem(operation_ptr, operation_size)
+                    .map_err(|_e| Trap::from(Error::InvalidMessageTopicOperation))?;
+
+                // don't allow managing topics for accounts.
+                if let Key::Account(_) = self.context.get_entity_address() {
+                    return Err(Trap::from(Error::InvalidContext));
+                }
+
+                let result = match topic_operation {
+                    MessageTopicOperation::Add => self.register_message_topic(topic_name)?,
+                };
+
+                Ok(Some(RuntimeValue::I32(api_error::i32_from(result))))
+            }
+            FunctionIndex::EmitMessage => {
+                // args(0) = pointer to the message kind name in wasm memory
+                // args(1) = size of the name string in wasm memory
+                // args(2) = pointer to the message contents
+                // args(3) = size of the message contents
+                let (topic_name_ptr, topic_name_size, message_ptr, message_size) =
+                    Args::parse(args)?;
+                self.charge_host_function_call(
+                    &host_function_costs.emit_message,
+                    [topic_name_ptr, topic_name_size, message_ptr, message_size],
+                )?;
+
+                self.context
+                    .engine_config()
+                    .wasm_config()
+                    .messages_limits()
+                    .topic_name_size_within_limits(topic_name_size)
+                    .map_err(|e| Trap::from(Error::CannotEmitMessage(e)))?;
+
+                self.context
+                    .engine_config()
+                    .wasm_config()
+                    .messages_limits()
+                    .message_size_within_limits(message_size)
+                    .map_err(|e| Trap::from(Error::CannotEmitMessage(e)))?;
+
+                let topic_name = self.string_from_mem(topic_name_ptr, topic_name_size)?;
+                let message = self.t_from_mem(message_ptr, message_size)?;
+
+                let result = self.emit_message(topic_name, message)?;
                 Ok(Some(RuntimeValue::I32(api_error::i32_from(result))))
             }
         }
--- a/execution_engine/src/runtime/externals.rs
+++ b/execution_engine/src/runtime/externals.rs
@@ -1108,8 +1108,8 @@ where
                 Ok(Some(RuntimeValue::I32(api_error::i32_from(result))))
             }
             FunctionIndex::ManageMessageTopic => {
-                // args(0) = pointer to the topic name in wasm memory
-                // args(1) = size of the topic name string in wasm memory
+                // args(0) = pointer to the serialized topic name string in wasm memory
+                // args(1) = size of the serialized topic name string in wasm memory
                 // args(2) = pointer to the operation to be performed for the specified topic
                 // args(3) = size of the operation
                 let (topic_name_ptr, topic_name_size, operation_ptr, operation_size) =
@@ -1155,10 +1155,10 @@ where
                 Ok(Some(RuntimeValue::I32(api_error::i32_from(result))))
             }
             FunctionIndex::EmitMessage => {
-                // args(0) = pointer to the message kind name in wasm memory
-                // args(1) = size of the name string in wasm memory
-                // args(2) = pointer to the message contents
-                // args(3) = size of the message contents
+                // args(0) = pointer to the serialized topic name string in wasm memory
+                // args(1) = size of the serialized name string in wasm memory
+                // args(2) = pointer to the serialized message payload in wasm memory
+                // args(3) = size of the serialized message payload in wasm memory
                 let (topic_name_ptr, topic_name_size, message_ptr, message_size) =
                     Args::parse(args)?;
                 self.charge_host_function_call(
@@ -1174,12 +1174,11 @@ where
                     )))));
                 }
 
-                self.context
-                    .engine_config()
-                    .wasm_config()
-                    .messages_limits()
-                    .message_size_within_limits(message_size)
-                    .map_err(|e| Trap::from(Error::CannotEmitMessage(e)))?;
+                if message_size > limits.max_message_size() {
+                    return Ok(Some(RuntimeValue::I32(api_error::i32_from(Err(
+                        ApiError::MessageTooLarge,
+                    )))));
+                }
 
                 let topic_name = self.string_from_mem(topic_name_ptr, topic_name_size)?;
                 let message = self.t_from_mem(message_ptr, message_size)?;
--- a/execution_engine/src/runtime/externals.rs
+++ b/execution_engine/src/runtime/externals.rs
@@ -1124,12 +1124,14 @@ where
                     ],
                 )?;
 
-                self.context
-                    .engine_config()
-                    .wasm_config()
-                    .messages_limits()
-                    .topic_name_size_within_limits(topic_name_size)
-                    .map_err(|e| Trap::from(Error::FailedTopicRegistration(e)))?;
+                let limits = self.context.engine_config().wasm_config().messages_limits();
+
+                if topic_name_size > limits.max_topic_name_size() {
+                    return Ok(Some(RuntimeValue::I32(api_error::i32_from(Err(
+                        ApiError::MaxTopicNameSizeExceeded,
+                    )))));
+                }
+
                 let topic_name = self.string_from_mem(topic_name_ptr, topic_name_size)?;
 
                 if operation_size as usize > MessageTopicOperation::max_serialized_len() {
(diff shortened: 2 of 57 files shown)


execution_engine/src/execution/executor.rs
```diff
@@ -203,12 +206,14 @@ impl Executor {
...
                 transfers: runtime.context().transfers().to_owned(),
                 cost: runtime.context().gas_counter(),
+                messages: runtime.context().messages(),
             },
             Err(error) => ExecutionResult::Failure {
                 effects,
                 error: error.into(),
                 transfers: runtime.context().transfers().to_owned(),
                 cost: runtime.context().gas_counter(),
+                messages,
```
jacek-casper, line 216: is there any particular reason this uses `messages` from `TrackingCopy` rather than `Runtime`?

execution_engine/src/runtime/mod.rs
```diff
@@ -3223,4 +3242,72 @@ where
             None => Err(Error::InvalidContract(contract_hash)),
         }
     }
+
+    fn add_message_topic(&mut self, topic_name: String) -> Result<Result<(), ApiError>, Error> {
+        let topic_hash = crypto::blake2b(AsRef::<[u8]>::as_ref(&topic_name)).into();
```
jacek-casper:
```suggestion
        let topic_hash = crypto::blake2b(topic_name.as_bytes()).into();
```

jacek-casper: actually it seems that just borrowing should work 🤔 because `blake2b` is defined as `blake2b<T: AsRef<[u8]>>(data: T)`
```suggestion
        let topic_hash = crypto::blake2b(&topic_name).into();
```

execution_engine/src/runtime/mod.rs
```diff
@@ -3223,4 +3242,72 @@ where
...
+        topic_name: String,
+        message: MessagePayload,
+    ) -> Result<Result<(), ApiError>, Trap> {
+        let entity_addr = self
+            .context
+            .get_entity_address()
+            .into_hash()
+            .ok_or(Error::InvalidContext)?;
+
+        let topic_name_hash = crypto::blake2b(AsRef::<[u8]>::as_ref(&topic_name)).into();
```
jacek-casper: same as above

execution_engine/src/runtime/mod.rs
```diff
@@ -3223,4 +3242,72 @@ where
...
+        let topic_key = Key::Message(MessageAddr::new_topic_addr(entity_addr, topic_name_hash));
+
+        // Check if the topic exists and get the summary.
+        let prev_topic_summary = if let Some(StoredValue::MessageTopic(message_summary)) =
+            self.context.read_gs(&topic_key)?
+        {
+            message_summary
+        } else {
+            return Ok(Err(ApiError::MessageTopicNotRegistered));
+        };
```
jacek-casper: could be simplified a bit with `let-else` syntax
```suggestion
        let Some(StoredValue::MessageTopic(message_summary)) =
            self.context.read_gs(&topic_key)? else {
            return Ok(Err(ApiError::MessageTopicNotRegistered));
        };
```

execution_engine/src/runtime_context/mod.rs
```diff
@@ -945,6 +953,49 @@ where
...
+    pub(crate) fn metered_emit_message<V>(
+        &mut self,
+        topic_key: Key,
+        topic_value: V,
+        message_key: Key,
+        message_value: V,
+        message: Message,
+    ) -> Result<(), Error>
+    where
+        V: Into<StoredValue>,
```
jacek-casper: I find this function signature a bit odd, the function immediately validates that the input is a `MessageTopic` and `MessageValue`, but we allow any `Into<StoredValue>` as an argument
What if we just forced the caller to give as a pair of `MessageTopicSummary` and `MessageChecksum` and then wrap them with `StoredValue` inside the funciton, which can potentially get rid of the need for validation?

execution_engine/src/tracking_copy/mod.rs
```diff
@@ -342,6 +345,24 @@ impl<R: StateReader<Key, StoredValue>> TrackingCopy<R> {
...
+    /// This function does not check the types for the key and the value so the caller should
+    /// correctly set the type. The `message_topic_key` should be of the `Key::MessageTopic`
+    /// variant and the `message_topic_summary` should be of the `StoredValue::Message` variant.
+    pub fn emit_message(
+        &mut self,
+        message_topic_key: Key,
+        message_topic_summary: StoredValue,
+        message_key: Key,
+        message_value: StoredValue,
+        message: Message,
```
jacek-casper, line 359: I'd have a similar comment here to what I mentioned above, I think we could pass in more explicitly typed arguments here, instead of the enums

types/src/contract_messages.rs
```diff
@@ -17,77 +29,195 @@ use rand::{
...
+
+        let (remainder, message_index) = match remainder.strip_prefix(TOPIC_FORMATTED_STRING_PREFIX)
+        {
+            Some(topic_string) => (topic_string, None),
+            None => {
+                let parts = input.rsplitn(2, '-').collect::<Vec<_>>();
+                if parts.len() != 2 {
+                    return Err(FromStrError::MissingMessageIndex);
+                }
+                (parts[1], Some(u32::from_str_radix(parts[0], 16)?))
```
jacek-casper: how about
```suggestion
                let (remainder, message_index_str) = input.rsplit_once('-').ok_or(FromStrError::MissingMessageIndex)?;
                let message_index = u32::from_str_radix(message_index_str, 16)?;
```
same can be done to simplify other usages of `splitn(2, *)` and `rsplitn(2, *)` as well

types/src/contract_messages.rs
```diff
@@ -17,77 +29,195 @@ use rand::{
...
+            }
+        };
+
+        let parts = remainder.splitn(2, '-').collect::<Vec<_>>();
+        if parts.len() != 2 {
+            return Err(FromStrError::MissingMessageIndex);
         }
+
+        let bytes = checksummed_hex::decode(parts[0])?;
```
jacek-casper: same as above

execution_engine/src/execution/executor.rs
alsrdn: In this case not really since the standard payment logic wouldn't emit any messages, but in the future it might. Every time a message is emitted, there is an impact on global state in the form of a write transform to `Key::message(entity_hash, topic_name_hash, message_index)`. In this case if there is a failure for the standard payment, the effects of that call are not included in the execution result (it returns the effects before the call to `call_host_standard_payment`) so it's natural to not include any messages that were produced by that call since the hashes for the messages won't be added to GS and in turn the messages would not be verifiable.

execution_engine/src/tracking_copy/mod.rs
alsrdn: Yeah, makes sense, though I think the tracking copy should probably only deal with `StoredValue`s rather than the inner objects to keep it more abstract.

execution_engine/src/runtime/mod.rs
alsrdn: Changed here: https://github.com/casper-network/casper-node/pull/4315/commits/566b3534a29355dd56ec8375f4c42d0a2ca0f24c

execution_engine/src/runtime/mod.rs
alsrdn: changed here: https://github.com/casper-network/casper-node/pull/4315/commits/566b3534a29355dd56ec8375f4c42d0a2ca0f24c

execution_engine/src/runtime/mod.rs
alsrdn: changed here: https://github.com/casper-network/casper-node/pull/4315/commits/566b3534a29355dd56ec8375f4c42d0a2ca0f24c

execution_engine/src/runtime_context/mod.rs
alsrdn: changed here: https://github.com/casper-network/casper-node/pull/4315/commits/566b3534a29355dd56ec8375f4c42d0a2ca0f24c

types/src/contract_messages.rs
alsrdn: changed here: https://github.com/casper-network/casper-node/pull/4315/commits/566b3534a29355dd56ec8375f4c42d0a2ca0f24c

types/src/contract_messages.rs
alsrdn: changed here: https://github.com/casper-network/casper-node/pull/4315/commits/566b3534a29355dd56ec8375f4c42d0a2ca0f24c

types/src/contract_messages/messages.rs
```diff
@@ -0,0 +1,189 @@
...
+pub const MESSAGE_PAYLOAD_STRING_TAG: u8 = 0;
+
+/// The payload of the message emitted by an addressable entity during execution.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
+#[cfg_attr(feature = "datasize", derive(DataSize))]
+#[cfg_attr(feature = "json-schema", derive(JsonSchema))]
+pub enum MessagePayload {
+    /// Human readable string message.
+    String(String),
+}
```
rafal-ch, line 72: Is such indirection intentional? `pub struct Message` could as well store the `message` as `String`, unless we want to allow different payloads in the _near_ future.

types/src/contract_messages/topics.rs
```diff
@@ -0,0 +1,234 @@
...
+}
+
+impl MessageTopicOperation {
+    /// Maximum serialized length of a message topic operation.
+    pub const fn max_serialized_len() -> usize {
+        OPERATION_MAX_SERIALIZED_LEN
+    }
+}
+
+impl ToBytes for MessageTopicOperation {
```
rafal-ch, line 211: Shall we add the serialization roundtrip test for `MessageTopicOperation`?

execution_engine/src/runtime/externals.rs
```diff
@@ -1106,6 +1107,85 @@ where
...
+                    return Err(Trap::from(Error::InvalidMessageTopicOperation));
+                }
+                let topic_operation = self
+                    .t_from_mem(operation_ptr, operation_size)
+                    .map_err(|_e| Trap::from(Error::InvalidMessageTopicOperation))?;
+
+                // don't allow managing topics for accounts.
+                if let Key::Account(_) = self.context.get_entity_address() {
+                    return Err(Trap::from(Error::InvalidContext));
+                }
```
rafal-ch, line 1147: Please consider extracting the validation part to a function and cover it with UT.

execution_engine/src/runtime/externals.rs
```diff
@@ -1106,6 +1107,85 @@ where
...
+                    return Ok(Some(RuntimeValue::I32(api_error::i32_from(Err(
+                        ApiError::MaxTopicNameSizeExceeded,
+                    )))));
+                }
+
+                if message_size > limits.max_message_size() {
+                    return Ok(Some(RuntimeValue::I32(api_error::i32_from(Err(
+                        ApiError::MessageTooLarge,
+                    )))));
+                }
```
rafal-ch, line 1181: Please consider extracting the validation part to a function and cover it with UT.

types/src/addressable_entity.rs
```diff
@@ -608,6 +609,113 @@ impl KeyValueJsonSchema for EntryPointLabels {
...
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, Debug, Default)]
+#[cfg_attr(feature = "datasize", derive(DataSize))]
+#[cfg_attr(feature = "json-schema", derive(JsonSchema))]
+#[serde(transparent, deny_unknown_fields)]
+pub struct MessageTopics(
+    #[serde(with = "BTreeMapToArray::<String, TopicNameHash, MessageTopicLabels>")]
+    BTreeMap<String, TopicNameHash>,
+);
+
+impl ToBytes for MessageTopics {
```
rafal-ch, line 622: Shall we add the serialization roundtrip test for `MessageTopics`?

types/src/addressable_entity.rs
```diff
@@ -608,6 +609,113 @@ impl KeyValueJsonSchema for EntryPointLabels {
...
+}
+
+impl MessageTopics {
+    /// Adds new message topic by topic name.
+    pub fn add_topic(
+        &mut self,
+        topic_name: String,
+        topic_name_hash: TopicNameHash,
+    ) -> Result<(), MessageTopicError> {
+        if self.0.len() > u32::MAX as usize {
```
rafal-ch: Shouldn't this use `MessagesLimits::max_topics_per_contract` and `>=` instead of `>`?

types/src/addressable_entity.rs
```diff
@@ -608,6 +609,113 @@ impl KeyValueJsonSchema for EntryPointLabels {
...
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Returns true if no message topics are registered.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+
+    /// Returns an iterator over the account hash and the weights.
```
rafal-ch:
```suggestion
    /// Returns an iterator over the topic and its hash.
```

types/src/chainspec/vm_config/host_function_costs.rs
```diff
@@ -295,6 +295,10 @@ pub struct HostFunctionCosts {
     pub random_bytes: HostFunction<[Cost; 2]>,
     /// Cost of calling the `enable_contract_version` host function.
     pub enable_contract_version: HostFunction<[Cost; 4]>,
+    /// Cost of calling the `casper_manage_message_topic` host function.
+    pub manage_message_topic: HostFunction<[Cost; 4]>,
+    /// Cost of calling the `casper_emit_message` host function.
+    pub emit_message: HostFunction<[Cost; 4]>,
```
rafal-ch, line 301: @mpapierski @AlexanderLimonov 
Please note that there are new host functions incoming - this may (or may not) influence you cost re-estimation efforts.

types/src/chainspec/vm_config/messages_limits.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn default() -> Self {
+        Self {
+            max_topic_name_size: 256,
+            max_message_size: 1024,
+            max_topics_per_contract: 128,
+        }
+    }
+}
+
+impl ToBytes for MessagesLimits {
```
rafal-ch: Shall we add the serialization roundtrip test for `MessagesLimits`?

types/src/chainspec/vm_config/messages_limits.rs
```diff
@@ -0,0 +1,113 @@
...
+use rand::{distributions::Standard, prelude::*, Rng};
+use serde::{Deserialize, Serialize};
+
+use crate::bytesrepr::{self, FromBytes, ToBytes};
+
+/// Configuration for messages limits.
+#[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Debug)]
+#[cfg_attr(feature = "datasize", derive(DataSize))]
+#[serde(deny_unknown_fields)]
+pub struct MessagesLimits {
```
rafal-ch: **A nit:** Consider renaming to `MessageLimits`, per analogy to `OpcodeCosts` vs `OpcodesCosts`, `HostFunctionCosts` vs `HostFunctionsCosts`, etc.

types/src/key.rs
```diff
@@ -717,10 +702,20 @@ impl Key {
         Key::Dictionary(addr)
     }
 
-    /// Creates a new [`Key::MessageTopic`] variant based on an `entity_addr` and a hash of the
-    /// topic name.
-    pub fn message_topic(entity_addr: HashAddr, topic_hash: MessageTopicHash) -> Key {
-        Key::MessageTopic(MessageTopicAddr::new(entity_addr, topic_hash))
+    /// Creates a new [`Key::Message`] variant that identifies an indexed message based on an
+    /// `entity_addr` `topic_hash` and message `index`.
```
rafal-ch: Nit:
```suggestion
    /// `entity_addr`, `topic_name_hash` and message `index`.
```

types/src/stored_value.rs
```diff
@@ -648,6 +673,9 @@ mod serde_helpers {
         AddressableEntity(&'a AddressableEntity),
         /// Variant that stores [`BidKind`].
         BidKind(&'a BidKind),
+        /// Variant that stores [`MessageSummary`].
+        MessageTopic(&'a MessageTopicSummary),
+        Message(&'a MessageChecksum),
```
rafal-ch, line 679: Missing comment for `Message`?

types/src/contract_messages/messages.rs
alsrdn: Yes, this is intentional. In the future we may want to add different types of message payloads. One example is having a pure bytes payload that allows full flexibility to create custom message schemas. Another example is to have standardised system level messages (e.g. if we decide that a message will be sent out when some operation on the contract will be performed, like disabling contract or adding a new version).

types/src/addressable_entity.rs
alsrdn: Indeed it should be `>=` but `MessagesLimits::max_topics_per_contract` should be enforced at a higher level (which it is). This is just a hard limit that is enforced.

types/src/contract_messages/topics.rs
alsrdn: added here: https://github.com/casper-network/casper-node/pull/4315/commits/3190647ec0012448907208e3a8406244433c5486

types/src/addressable_entity.rs
alsrdn: added here: https://github.com/casper-network/casper-node/pull/4315/commits/3190647ec0012448907208e3a8406244433c5486

types/src/addressable_entity.rs
alsrdn: changed here: https://github.com/casper-network/casper-node/pull/4315/commits/3190647ec0012448907208e3a8406244433c5486

types/src/addressable_entity.rs
alsrdn: fixed here: https://github.com/casper-network/casper-node/pull/4315/commits/3190647ec0012448907208e3a8406244433c5486

types/src/chainspec/vm_config/messages_limits.rs
alsrdn: I thought it was picked up by a serialization test for `WasmConfig` but it wasn't. Added here: https://github.com/casper-network/casper-node/pull/4315/commits/3190647ec0012448907208e3a8406244433c5486

types/src/chainspec/vm_config/messages_limits.rs
alsrdn: Renamed here: https://github.com/casper-network/casper-node/commit/3190647ec0012448907208e3a8406244433c5486

types/src/key.rs
alsrdn: Fixed here: https://github.com/casper-network/casper-node/commit/3190647ec0012448907208e3a8406244433c5486

types/src/stored_value.rs
alsrdn: Fixed here: https://github.com/casper-network/casper-node/commit/3190647ec0012448907208e3a8406244433c5486

execution_engine/src/runtime/externals.rs
alsrdn: There is already a unit test to check limits in `execution_engine_testing/tests/src/test/contract_messages.rs::should_not_exceed_configured_limits`. I don't think it's necessary to add new functions for checking each parameter.

execution_engine/src/runtime/externals.rs
alsrdn: Added a check for max message size in the UT here: `execution_engine_testing/tests/src/test/contract_messages.rs::should_not_exceed_configured_limits`

types/src/contract_messages/messages.rs
rafal-ch: Makes sense, thanks for explanation.

execution_engine/src/runtime/externals.rs
rafal-ch: I see, you check it on the higher level. That's perfectly fine :+1: 
Is the check for `don't allow managing topics for accounts` also included in some other test?

rafal-ch (approved):
LGTM :+1: 
No blockers.

execution_engine/src/runtime/externals.rs
alsrdn: Good point, I'll make a separate test for that in a different PR.

Issue #4318, opened 2023-10-05
Extend Key variants to accomodate AddressableEntity, Package and ByteCode


SaiProServ:
We should extend the Key enum to include three new variants: AddressableEntity, Package, and ByteCode. This will clearly distinguish between the legacy types and indicate where the new 2.0 records are written in global state.

Issue #4279, opened 2023-09-11
Disable stored session


SaiProServ:
Stored Session has been an unused part of the system and is functionally defunct. 
As part of the Execution Engine work, it needs to be removed as removing it will allow us to then start working on letting Accounts upgrade.

cspramit:
split the large stories @rafal-ch @SaiProServ @darthsiroftardis

Issue #4167, opened 2023-07-25
extend action thresholds to support package upgrade


RitaMAllenCA:
currently action thresholds include key management and deploy. upgrade management action threshold will be added: Do the approvals have enough weight to upgrade the given package.

PR #4320, opened 2023-10-05
Disable stored session and rework standard payment

### LINKED ISSUE #4277: Refactor Contract related types to AddressableEntity
There are many types like ContractHash and ContractPackage that need to be refactored and renamed to be brought in line with the newly AddressableEntity struct. This also includes the renaming of variables within the code to make the code readable and consistent with the changes made by the account/contract merger.

darthsiroftardis:
CHANGELOG:

- Expand the `Key` enum to have three new variants `Key::AddressableEntity` , `Key::Package` and `Key::ByteCode`
- Extend action thresholds to have an extra action for allowing upgrades
- Disable the execution of stored session
- Reworked standard payment to invoke the mint directly 
- Refactored variable names to account for the changes made by introducing the `AddressableEntity` struct

Closes #4277 #4279 #4167 and #4318


EdHastingsCasperAssociation (approved):
there are still some old naming convention jank, but approving given that you plan to fix up the remaining naming issues in a follow on PR as discussed.

PR #4326, opened 2023-10-06
Merge SSE event streams

### LINKED ISSUE #4314: Merge 3 event streams into one
Currently we maintain 3 separate streams: /events/main, /events/deploys and /events/sigs. 

Going forward, these should all be available from the single stream.

rafal-ch:
This PR changes the SSE server in a way that all events are emitted through a single endpoint `<IP:Port>/events`.

Closes #4314

### DIFF
--- a/utils/nctl/sh/scenarios/event_stream.sh
+++ b/utils/nctl/sh/scenarios/event_stream.sh
@@ -8,6 +8,8 @@ source "$NCTL"/sh/scenarios/common/itst.sh
 # Exit if any of the commands fail.
 set -e
 
+TMP_EVENTS_FILE=tmp_events_main.txt
+
 #######################################
 # Runs an integration tests that verifies if events are routed via correct SSE endpoints.
 # Prior to https://github.com/casper-network/casper-node/issues/4314 there were 3 different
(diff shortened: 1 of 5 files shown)


jacek-casper (approved):

utils/nctl/sh/scenarios/event_stream.sh
```diff
@@ -0,0 +1,119 @@
...
+
+# Exit if any of the commands fail.
+set -e
+
+TMP_EVENTS_FILE=tmp_events_main.txt
+
+#######################################
+# Runs an integration tests that verifies if events are routed via correct SSE endpoints.
+# Prior to https://github.com/casper-network/casper-node/issues/4314 there were 3 different
+# endpoints, serving different evvents. The 3 events that this test verifies used to be
```
alsrdn: nit: typo `events` instead of `evvents`

alsrdn (approved):
Looks good to me in this form, but I'm wondering if we actually have any need for the additional `/name` path. Wouldn't it be more clear to subscribe directly to `<host>/events` instead of `<host>/events/main`?

utils/nctl/sh/scenarios/event_stream.sh
rafal-ch: Fixed in d32e1e44754aaa86d750d3a5f84bcc6a29be8c9f

real-felix (approved):

rafal-ch:
> Looks good to me in this form, but I'm wondering if we actually have any need for the additional /name path. Wouldn't it be more clear to subscribe directly to <host>/events instead of <host>/events/main?

The change has been implemented in https://github.com/casper-network/casper-node/pull/4326/commits/ef6a362d7c08171c0e24a3c8bd14b3c1292a74b5

cc: @alsrdn @Fraser999

rafal-ch:
bors try

alsrdn (approved):

PR #4344, opened 2023-10-13
mechanism to increase gas cost for each message emitted within one execution

alsrdn:
Implement a mechanism to increase gas cost for each message emitted within one execution.
Also fix a few test issues and add more tests for contract level messages.


execution_engine/src/runtime/externals.rs
```diff
@@ -1184,6 +1184,9 @@ where
                 let message = self.t_from_mem(message_ptr, message_size)?;
 
                 let result = self.emit_message(topic_name, message)?;
+                if result.is_ok() {
+                    self.charge_emit_message()?;
```
rafal-ch: Are these calls immediate? I.e. if we fail to charge, do we still emit the message?

execution_engine_testing/tests/src/test/contract_messages.rs
```diff
@@ -494,3 +542,241 @@ fn should_not_exceed_configured_limits() {
...
+#[ignore]
+#[test]
+fn should_charge_increasing_gas_cost_for_multiple_messages_emitted() {
+    const FIRST_MESSAGE_EMIT_COST: u32 = 100;
+    const COST_INCREASE_PER_MESSAGE: u32 = 50;
+    const MESSAGES_TO_EMIT: u32 = 4;
+    const EMIT_MULTIPLE_EXPECTED_COST: u32 = FIRST_MESSAGE_EMIT_COST * MESSAGES_TO_EMIT
+        + (MESSAGES_TO_EMIT - 1) * MESSAGES_TO_EMIT / 2 * COST_INCREASE_PER_MESSAGE;
+    const EMIT_MESSAGES_FROM_MULTIPLE_CONTRACTS: u32 =
+        FIRST_MESSAGE_EMIT_COST * 3 + 3 * COST_INCREASE_PER_MESSAGE; // simplification of above
```
rafal-ch: `EMIT_MULTIPLE_EXPECTED_COST` is 700, but `EMIT_MESSAGES_FROM_MULTIPLE_CONTRACTS` is 450.
Simplification suggests that the values stay the same, but formula is simpler.

Also, can we name the magic number `3`?

execution_engine_testing/tests/src/test/contract_messages.rs
```diff
@@ -494,3 +542,241 @@ fn should_not_exceed_configured_limits() {
...
+        EMIT_MULTIPLE_EXPECTED_COST.into()
+    );
+
+    // Try another execution where we emit a single message.
+    // Cost should be `FIRST_MESSAGE_EMIT_COST`
+    emit_message_with_suffix(&builder, "test", &contract_hash, DEFAULT_BLOCK_TIME);
+    let emit_message_gas_cost = builder.borrow().last_exec_gas_cost().value();
+    assert_eq!(emit_message_gas_cost, FIRST_MESSAGE_EMIT_COST.into());
+
+    // Check gas cost when multiple messages are emitted from different contracts.
```
rafal-ch, line 762: Maybe consider moving this test-case to a separate test?

types/src/chainspec/vm_config/host_function_costs.rs
```diff
@@ -301,6 +316,125 @@ pub struct HostFunctionCosts {
...
+            blake2b: HostFunction::<[Cost; 4]>::new(0, [0; 4]),
+            random_bytes: HostFunction::<[Cost; 2]>::new(0, [0; 2]),
+            enable_contract_version: HostFunction::<[Cost; 4]>::new(0, [0; 4]),
+            manage_message_topic: HostFunction::<[Cost; 4]>::new(0, [0; 4]),
+            emit_message: HostFunction::<[Cost; 4]>::new(0, [0; 4]),
+        }
+    }
+
+    fn is_zero(&self) -> bool {
+        todo!()
```
rafal-ch: left on purpose for next PRs?

types/src/chainspec/vm_config/host_function_costs.rs
```diff
@@ -301,6 +316,125 @@ pub struct HostFunctionCosts {
     pub emit_message: HostFunction<[Cost; 4]>,
 }
 
+impl Add for HostFunctionCosts {
+    type Output = HostFunctionCosts;
+
+    fn add(self, rhs: Self) -> Self::Output {
+        Self {
+            read_value: self.read_value + rhs.read_value,
```
rafal-ch: This could use some macro-magic to avoid copy&paste mistakes, or rework of `HostFunctionCosts`.
Not in this PR, though.

execution_engine/src/runtime/externals.rs
alsrdn: The effects should be reverted if we fail to charge since we would trap with a `GasLimit` error.
What I wanted to do is charge if the message was actually emitted. If an `ApiError` occurred, we shouldn't charge for it. Are you suggesting we should charge every attempt to emit a message? Then the next attempt would have increased gas or have the previous gas cost?

execution_engine_testing/tests/src/test/contract_messages.rs
alsrdn: No, the formula is the same, the expression is simplified when `MESSAGES_TO_EMIT == 3`. We can add a `const fn` with the formula and get rid of these expression. WDYT?

types/src/chainspec/vm_config/host_function_costs.rs
alsrdn: No, actually forgot about it. Good catch!

types/src/chainspec/vm_config/host_function_costs.rs
alsrdn: I went with the default aversion to using macros here 😁 but it does seem that we use `derive_more` in some other places in the project so I can switch to that for `Add`.

execution_engine_testing/tests/src/test/contract_messages.rs
alsrdn: Moving it will add a lot of boiler plate which I don't think it's warranted in this situation. I would prefer to leave the check in this test which is related to increasing gas costs.

types/src/chainspec/vm_config/host_function_costs.rs
alsrdn: Changed here: https://github.com/casper-network/casper-node/pull/4344/commits/ffe7e18b9a7139defaa1cebf8c97bde61722a229

types/src/chainspec/vm_config/host_function_costs.rs
alsrdn: Added here: https://github.com/casper-network/casper-node/pull/4344/commits/ffe7e18b9a7139defaa1cebf8c97bde61722a229

execution_engine_testing/tests/src/test/contract_messages.rs
alsrdn: Changed here: https://github.com/casper-network/casper-node/pull/4344/commits/ffe7e18b9a7139defaa1cebf8c97bde61722a229

cspramit:
@alsrdn @SaiProServ whats the associated story?

execution_engine_testing/tests/src/test/contract_messages.rs
rafal-ch: Thanks, it's more clear this way :+1:

execution_engine_testing/tests/src/test/contract_messages.rs
rafal-ch: Ok, fair enough.
Please just confirm if my understanding of this test scenario is correct:
```
Step 1:
1. Emit one message via "emit_message" entry point in the original contract
2. Expect: FIRST_MESSAGE_EMIT_COST

Step 2:
1. Emit 4 messages from single contract via "emit_multiple_messages" entry point in the original contract
2. Expect: emit_cost_per_execution(4)

Step 3:
1. Again emit one message via "emit_message" entry point in the original contract (to assert that costs calculation is reset across executions)
2. Expect: FIRST_MESSAGE_EMIT_COST

Step 4:
1. Emit 3 messages 
    - two via "emit_message_from_each_version" entry point in the upgraded contract
    - one via "emit_message" entry point in the original contract
2. Expect: emit_cost_per_execution(3)
```

alsrdn: yes, this is the expected scenario

execution_engine/src/runtime_context/mod.rs
```diff
@@ -126,6 +127,7 @@ where
             transfers,
             remaining_spending_limit,
             package_kind,
+            last_message_cost: U512::from(0),
```
EdHastingsCasperAssociation: i don't think the `last_` prefix is necessary

EdHastingsCasperAssociation (approved):

execution_engine/src/runtime/externals.rs
mpapierski: As discussed on slack we should always charge, but increase the counter if the message is actually emitted - in this case, the user has a chance to retry the message by (for instance) shrinking the message to fit the limit.

alsrdn: @mpapierski @rafal-ch changed the behaviour here: https://github.com/casper-network/casper-node/pull/4344/commits/8171806b24ac2178a16b6721587d073eeb42a0d5

Also got rid of the `MessageCosts` config and added the parameter for cost increase to `HostFunctionCosts` which I think is more cleaner.

Now within an execution, the first call to emit a message will charge `host_function_cost.emit_message.cost()`.  The second call to emit will have cost `host_function_cost.emit_message.cost() + host_function_cost.cost_increase_per_message`. If for the second call a message was not emitted due to an `ApiError`, the third call will cost the same as the previous one (`host_function_cost.emit_message.cost() + host_function_cost.cost_increase_per_message`).

execution_engine/src/runtime_context/mod.rs
alsrdn: renamed here: https://github.com/casper-network/casper-node/commit/8171806b24ac2178a16b6721587d073eeb42a0d5

rafal-ch (approved):
LGMT :+1:

PR #4346, opened 2023-10-16
Setting URef expiry

jacek-casper:
The main changes here are:
- introduce a new `Key` variant named `Context` for storing keys that are only accessible by the creator of the key (these Context keys are explained [here](https://github.com/casper-network/roadmap/issues/105#issuecomment-1690337391) in more detail)
- introduce a new `StoredValue` variant named `URef` for storing metadata relavant to `URefs` which currently consists of a pair of a `Context` key and a lifetime of the `URef` - this will allow us to prune expired `URef`s
- introduce new ffi operations for creating `Context` keys and creating `URef`s with a lifetime
- swap out the storage implementation of `URef` to leverage the new metadata
  - this means that whenever we read a `URef`, we're actually going through an indirection - we read a `StoredValue::URef` first and check it for expiry and then we read the underlying CLValue using the `Context` key
  - this has impact on genesis, because `GenesisInstaller` directly manipulates the storage and stores a number of important `URef`s that are used for auctions and other system contracts, we've done some changes here to avoid using `URef`s unnecessarily and now most of these keys used here are just `Context` keys owned by contracts that access them
  - this also impacts many tests that use `TrackingCopy` to directly manipulate storage, a number of tests had to be updated to avoid assuming that a `Key::URef` can be used to directly read a `CLValue` out of it
  - the final piece is a migration that converts existing `URef`s to utilize `StoredValue::URef` where appropriate (outside of system contracts)


cspramit:
@jacek-casper  what is the associated story cc: @SaiProServ

jacek-casper:
@cspramit this is the ticket: https://app.zenhub.com/workspaces/core-protocol-60953fafb1945f0011a3592d/issues/gh/casper-network/roadmap/143

jacek-casper:
@EdHastingsCasperAssociation @mpapierski 
What is and is not included here:
- URef share-ability is out of scope of this ticket, but adding it to the storage is going to be pretty easy, it'd just be a new field under `StoredValue::URef`.
- Deleting URefs once expired is also in a separate ticket.
- I've added the new storage implementation for `URef`s, adding it required changing a lot of code, because many places interact directly with storage via `TrackingCopy`. `TrackingCopy` seems like it should not handle the `URef` indirection by default, so I've added some internal helper methods for `URef`s that are used in multiple places.
- I haven't removed the existing `URef` ffi, but I guess that is the goal? As of now both ffi exist, the old ffi just creates `URef`s that last forever and the new ffi accepts a lifetime parameter.
- I haven't added anything to handle the cost of the 2 new FFI functions, I'm not sure what it's supposed to be

PR #4348, opened 2023-10-17
Proposition of changes into Casper Types to support a Rust SDK for 1.6

gRoussac:
Please consider the following when creating a PR:

* Provide a useful description of this PR, suitably verbose, aimed at helping reviewers and contributors

This PR intends to pull changes made for the Rust SDK. It is adding a `std-fs-io` default feature and thus adds several methods or expose them accordingly to be able to compile the SDK into a wasm file.

* Update all relevant changelogs
check
* Provide a link to the GitHub issue relating to this PR

https://github.com/casper-ecosystem/rustSDK/issues/8
https://github.com/casper-ecosystem/rustSDK/issues/10

* Identify if any downstream impact as in to, SDKs, SmartContracts etc

This code is intended to go into 1.6 version for the 1.6 ready version of the SDK.


zacshowa (commented):
Similar to my comments on the related PR against the `casper-client`, I think it makes sense to approve these to get the SDK out for 1.6. However, I would love to see this feature changed in 2.0 to be additive.
Additionally, I would like to see the lints fixed along with whatever is causing the drone to fail on cargo-test-pr.

PR #4351, opened 2023-10-19
Transactions on JSON-RPC server

Fraser999:
This PR upgrades the JSON-RPC server to handle `Transaction`s while leaving the legacy `Deploy` interface unchanged.

The new RPCs are `account_put_transaction` and `info_get_transaction` with similar behaviour and semantics as `account_put_deploy` and `info_get_deploy` respectively.

The speculative execution server was also updated, with a new endpoint `speculative_exec_txn` taking a `Transaction` while leaving the original `speculative_exec` unchanged.

This PR also includes an attempt to unify the approaches to handling versioned data in storage.  I extended the notion of the `VersionedDatabases` to try and make the usage of these less accident-prone, making the actual DBs private and exposing methods which ensure old, pre-existing data in the `legacy` DB is exclusively using legacy types and bincode encoding, while new data is held in the `current` DB and uses bytesrepr-encoding of the keys and values.  In doing this, most of the methods in `storage/write_block.rs` were deleted, so I moved the couple of remaining methods back into `storage.rs` and deleted that file.  I also moved the `Config` and `Event` types unchanged to their own files to try and declutter storage.rs a little.

Closes [roadmap#185](https://github.com/casper-network/roadmap/issues/185).


node/src/components/storage.rs
```diff
@@ -417,132 +326,81 @@ impl Storage {
...
+                }
+                None => true,
+            };
 
-                    cursor.del(WriteFlags::empty())?;
-                    continue;
+            // If we don't already have the block body hash in the collection, insert it with the
+            // value `should_retain_block`.
+            //
+            // If there is an existing value, the updated value should be `false` iff the existing
```
rafal-ch, line 376: `iff`?

I thought it's a typo, but maybe you meant "if and only if"?

node/src/components/rpc_server/rpcs/account.rs
```diff
@@ -95,3 +103,75 @@ impl RpcWithParams for PutDeploy {
...
+        &PUT_TRANSACTION_RESULT
+    }
+}
+
+/// "account_put_transaction" RPC
+pub struct PutTransaction {}
+
+#[async_trait]
+impl RpcWithParams for PutTransaction {
+    const METHOD: &'static str = "account_put_transaction";
```
rafal-ch, line 143: Can you request the ticket to cover these new endpoints with `casper-test`?

rafal-ch (approved):
LGTM :+1: 
Good job!

node/src/components/storage.rs
Fraser999: Yes, I meant if and only if :)

node/src/components/rpc_server/rpcs/account.rs
Fraser999: Good point.  Done now.

Fraser999:
bors merge

Issue #4269, opened 2023-09-05
Integrate Juliet timeouts into the node proper


RitaMAllenCA:
Requires #4139

With native timeout support in Juliet, we want to simplify the node code by removing round-trip pings in favor of making all requests performed use a timeout. Asynchronous connection cleanup should take care of the rest.

PR #4354, opened 2023-10-19
Use juliet's default timeout and timeout bubbling to make missed ACKs fatal.

marc-casperlabs:
Addition to #4269.

* Makes the timeout configurable (from hardcoded 30 seconds).
* Every request has its timeout set through `juliet` means (default timeout), instead of requiring special node logic.
* In the same vein, timeouts bubble up to be a server error as well as a per-request one if configured to do so.

### DIFF
--- a/juliet/src/rpc.rs
+++ b/juliet/src/rpc.rs
@@ -56,6 +56,8 @@ pub struct RpcBuilder<const N: usize> {
     core: IoCoreBuilder<N>,
     /// Whether or not to enable timeout bubbling.
     bubble_timeouts: bool,
+    /// The default timeout for created requests.
+    default_timeout: Option<Duration>,
 }
 
 impl<const N: usize> RpcBuilder<N> {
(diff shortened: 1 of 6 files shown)


juliet/src/rpc.rs
```diff
@@ -345,10 +394,13 @@ where
...
     ///
     /// Returns the duration until the next timeout check needs to take place if timeouts are not
-    /// modified in the interim.
-    fn process_timeouts(&mut self, now: Instant) -> Instant {
+    /// modified in the interim, and the number of actual timeouts.
+    fn process_timeouts(&mut self, now: Instant) -> (Instant, usize) {
         let is_expired = |t: &Reverse<(Instant, IoId)>| t.0 .0 <= now;
 
+        // Track the number of actual timeouts hit.
+        let mut timed_out = 0;
```
Fraser999, line 402: Why bother counting them if any non-zero value triggers the error?

Fraser999 (approved):
Only one non-blocking nitpick.  Otherwise LGTM.

juliet/src/rpc.rs
marc-casperlabs: They are logged. In general, the info is free, so I thought it best to not just discard it in the library.

PR #4361, opened 2023-10-20
Remove the superfluous deploy crypto verification from deploy buffer

rafal-ch:
This PR removes the call to `Deploy::is_valid()` in the `DeployBuffer` as we assume that all deploys coming from `Storage` are already verified.

The following graph shows which part of the flow got removed.

Untitled Diagram (copy) drawio


Fraser999 (approved):

rafal-ch:
@EdHastingsCasperAssociation suggested that the additional `crypto::verify()` check could still be used when we're putting deploys in the `DeployBuffer` at the moment of initialization.

rafal-ch:
@EdHastingsCasperAssociation 
Currently, on initialization of `DeployBuffer` we saturate it with deploy hashes. This means that there's no risk of smuggling in a deploy from other source than our storage, which we assume stores verified data.

EdHastingsCasperAssociation:
> @EdHastingsCasperAssociation Currently, on initialization of `DeployBuffer` we saturate it with deploy hashes. This means that there's no risk of smuggling in a deploy from other source than our storage, which we assume stores verified data.

the deploys could be inserted directly into the db while the node is down

alsrdn:
> > @EdHastingsCasperAssociation Currently, on initialization of `DeployBuffer` we saturate it with deploy hashes. This means that there's no risk of smuggling in a deploy from other source than our storage, which we assume stores verified data.
> 
> the deploys could be inserted directly into the db while the node is down

Yes but the DB is part of the operator's security domain which I think we must assume is safe/sane. Otherwise the same can be said about the in-memory deploy buffer, an operator can insert deploys in memory unrestricted.

rafal-ch:
>>    the deploys could be inserted directly into the db while the node is down
>
> Yes but the DB is part of the operator's security domain which I think we must assume is safe/sane. Otherwise the same can be said about the in-memory deploy buffer, an operator can insert deploys in memory unrestricted.

I leaning towards agreeing with @alsrdn here. If operator can't ensure security he might find that his node binary got replaced with malicious ones or that the keys were tampered with.

rafal-ch:
@EdHastingsCasperAssociation did you have a chance to look at the comments above?

Also, adding more reviewers since this change is related to security.

cc: @marc-casperlabs @fizyk20

rafal-ch:
Before merging we should check how much we gain with this optimization - if the gain is minimal, we might consider leaving the additional crypto check intact just to be on the safe side.

EdHastingsCasperAssociation (approved):
I'm convinced enough by the argumentation to relax this guard.

PR #4370, opened 2023-10-20
Contract level messages

alsrdn:
Contract level messages are a facility that enable a smart contract to emit a message while executing, which eventually gets sent on the event stream and is visible from outside the network.

This PR introduces 2 new FFIs that allow contracts to manage message topics and emit messages.
Messages are sent out through `DeployProcessed` SSE events.


EdHastingsCasperAssociation (approved):

execution_engine/src/runtime/externals.rs
```diff
@@ -1124,6 +1125,100 @@ where
...
+
+                let limits = self.context.engine_config().wasm_config().messages_limits();
+
+                if topic_name_size > limits.max_topic_name_size() {
+                    return Ok(Some(RuntimeValue::I32(api_error::i32_from(Err(
+                        ApiError::MaxTopicNameSizeExceeded,
+                    )))));
+                }
+
+                let topic_name = self.string_from_mem(topic_name_ptr, topic_name_size)?;
```
mpapierski: I know we're not consistent at all with regards to this but I have suggestion that we perhaps don't deserialize a string here, but instead, pass the string directly from Wasm. The host already has access to all the data, and you just need to pass pointer to the string i.e.

```rust
// wasm
fn foo(s: &str) {
  ext_ffi::foo(s.as_ptr(), s.len());
}
// host
self.checked_memory_slice(key_ptr as usize, key_size as usize, |data| {
  std::str::from_utf8(data).map(ToOwned::to_owned)
})?
```

This should make it cheaper for the users as additional serialization, deserialization, etc, adds to the gas costs.

execution_engine/src/runtime/externals.rs
```diff
@@ -1124,6 +1125,100 @@ where
...
+                    )))));
+                }
+
+                if message_size > limits.max_message_size() {
+                    return Ok(Some(RuntimeValue::I32(api_error::i32_from(Err(
+                        ApiError::MessageTooLarge,
+                    )))));
+                }
+
+                let topic_name = self.string_from_mem(topic_name_ptr, topic_name_size)?;
```
mpapierski: Same as above

node/src/components/contract_runtime/operations.rs
```diff
@@ -152,20 +155,21 @@ pub fn execute_finalized_block(
...
-        let (state_hash, execution_result) = commit_execution_results(
+        let (state_hash, execution_result, messages) = commit_execution_results(
             &scratch_state,
             metrics.clone(),
             state_root_hash,
             deploy_hash,
             result,
         )?;
-        execution_results.push((deploy_hash, deploy_header, execution_result));
+        execution_results.push((deploy_hash, deploy_header, execution_result, messages));
```
mpapierski: Could this be changed into a named struct at this point?

node/src/components/contract_runtime/operations.rs
```diff
@@ -152,20 +155,21 @@ pub fn execute_finalized_block(
...
         state_root_hash = state_hash;
     }
 
     // Write the deploy approvals' and execution results' checksums to global state.
-    let execution_results_checksum =
-        compute_execution_results_checksum(execution_results.iter().map(|(_, _, result)| result))?;
+    let execution_results_checksum = compute_execution_results_checksum(
+        execution_results.iter().map(|(_, _, result, _)| result),
```
mpapierski: Then you could hide the complexity

```suggestion
        execution_results.into_iter().map(|execution_result| execution_result.result),
```

smart_contracts/contract/src/contract_api/runtime.rs
```diff
@@ -404,6 +405,44 @@ pub fn get_call_stack() -> Vec<CallStackElement> {
...
+/// Manages a message topic.
+pub fn manage_message_topic(
+    topic_name: &str,
+    operation: MessageTopicOperation,
+) -> Result<(), ApiError> {
+    if topic_name.is_empty() {
+        return Err(ApiError::InvalidArgument);
+    }
+
+    let (topic_name_ptr, topic_name_size, _bytes) = contract_api::to_ptr(topic_name);
```
mpapierski: See my other comment regarding passing strings

smart_contracts/contract/src/contract_api/runtime.rs
```diff
@@ -404,6 +405,44 @@ pub fn get_call_stack() -> Vec<CallStackElement> {
...
+    let (operation_ptr, operation_size, _bytes) = contract_api::to_ptr(operation);
```
mpapierski, line 417: To be fair, looking at the definition this could be passed as a number to the host - unless you plan to add more variants with fields etc. in the future. Adding `#[derive(ToPrimitive, FromPrimitive)]` to `MessageTopicOperation` could help with convenience

types/src/chainspec/vm_config/opcode_costs.rs
```diff
@@ -531,6 +582,48 @@ impl FromBytes for OpcodeCosts {
...
+            unreachable: 0,
+            nop: 0,
+            current_memory: 0,
+            grow_memory: 0,
+            control_flow: ControlFlowCosts::zero(),
+        }
+    }
+
+    fn is_zero(&self) -> bool {
+        self.bit.is_zero()
```
mpapierski: Can you use `let OpcodeCosts { bit, add, .. } = self` etc just so we wouldn't accidentally forget to update this if it grows new fields?

types/src/contract_messages/messages.rs
```diff
@@ -0,0 +1,324 @@
...
+const MESSAGE_PAYLOAD_TAG_LENGTH: usize = U8_SERIALIZED_LENGTH;
+
+/// Tag for a message payload that contains a human readable string.
+pub const MESSAGE_PAYLOAD_STRING_TAG: u8 = 0;
+
+/// The payload of the message emitted by an addressable entity during execution.
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize, Debug)]
+#[cfg_attr(feature = "datasize", derive(DataSize))]
+#[cfg_attr(feature = "json-schema", derive(JsonSchema))]
+pub enum MessagePayload {
```
mpapierski, line 122: What other variants could be added here in the future, and why CLValue wasn't considered for this purpose? If there are no other variants planned, why not just make it a new type over String?

types/src/key.rs
```diff
@@ -902,6 +952,15 @@ impl Key {
...
     }
+
+    /// Return true if the inner Key is of the smart contract type.
+    pub fn is_smart_contract_key(&self) -> bool {
+        if let Self::AddressableEntity(PackageKindTag::SmartContract, _) = self {
+            return true;
+        }
+
+        false
+    }
```
mpapierski, line 963: nit

```suggestion
        matches!(self, Self::AddressableEntity(PackageKindTag::SmartContract, _))
```

mpapierski (approved):
Couple of notes regarding costs, and few nits, but overall looks good 👍

types/src/contract_messages/messages.rs
mpapierski: As discussed offline future variants could have richer structure so it's not really a subset or replacement of a CLValue

smart_contracts/contract/src/contract_api/runtime.rs
alsrdn: Indeed, now we could pass it as a plain number, and it would probably work for a new operation like `Remove`. Though my thinking was that if we're going to support more complex operations in the future like `Rename`, or to allow selection of the checksum algorithm we would need space for some parameters.

types/src/key.rs
alsrdn: The other implementations follow the same (e.g. `is_bid_addr_key`, `is_dictionary_key` etc) follow the same style so I would prefer to keep it like this.

execution_engine/src/runtime/externals.rs
alsrdn: Removed serialization in https://github.com/casper-network/casper-node/pull/4370/commits/53483e16590b67de9be9be4d9ba4e6b64d1f22ee although didn't use `checked_memory_slice` as it's not available yet on the feat-2.0 branch. Will do a follow-up commit after.

execution_engine/src/runtime/externals.rs
alsrdn: Removed serialization in https://github.com/casper-network/casper-node/commit/53483e16590b67de9be9be4d9ba4e6b64d1f22ee although didn't use checked_memory_slice as it's not available yet on the feat-2.0 branch. Will do a follow-up commit after.

node/src/components/contract_runtime/operations.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/commit/53483e16590b67de9be9be4d9ba4e6b64d1f22ee

node/src/components/contract_runtime/operations.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/commit/53483e16590b67de9be9be4d9ba4e6b64d1f22ee

smart_contracts/contract/src/contract_api/runtime.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/commit/53483e16590b67de9be9be4d9ba4e6b64d1f22ee

types/src/chainspec/vm_config/opcode_costs.rs
alsrdn: Changed in https://github.com/casper-network/casper-node/commit/53483e16590b67de9be9be4d9ba4e6b64d1f22ee

alsrdn:
bors merge

PR #4380, opened 2023-10-25
Update refund handling ratio and fix related code

### LINKED ISSUE #4353: 1.5.4 test unspent fund scenarios
( we may need to write unit tests)
For the release of 1.5.4 - we need to ensure via tests that the unspent funds are refunded to users via a setting in the configurations
- Test boundaries of the configuration setting, as in 0 - 50 -100 % refund and assert the correct amount is refunded
- Test the setting when the user does not have any eligible refunds to start with
Scenarios to test are not limited to above

jacek-casper:
Updated the refund_ratio, wich broke some tests that relied on the refund value being zero. These tests and some related code was updated. Also added a prop test for the refund calculation

Closes #4353

### DIFF
--- a/execution_engine/src/system/handle_payment/internal.rs
+++ b/execution_engine/src/system/handle_payment/internal.rs
@@ -49,6 +49,10 @@ pub(crate) fn get_refund_purse<R: RuntimeProvider>(
 }
 
 /// Returns tuple where 1st element is the refund, and 2nd element is the fee.
+///
+/// # Note
+///
+/// Any dust amounts are added to the fee.
 fn calculate_refund_and_fee(
     gas_spent: U512,
     payment_purse_balance: U512,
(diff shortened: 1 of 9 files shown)


node/src/components/network/outgoing.rs
```diff
@@ -27,8 +27,8 @@
 //! * The `Dialer` is expected to produce `DialOutcomes` for every dial [`DialRequest::Dial`]
 //!   eventually. These must be forwarded to the `OutgoingManager` via the `handle_dial_outcome`
 //!   function.
-//! * The `perform_housekeeping` method must be called periodically to give the
-//!   `OutgoingManager` a chance to initiate reconnections and collect garbage.
+//! * The `perform_housekeeping` method must be called periodically to give the `OutgoingManager` a
+//!   chance to initiate reconnections and collect garbage.
```
rafal-ch, line 31: Can you revert this change as unrelated? Or at least put it in a separate commit (assuming the formatter keeps insisting that this is the correct layout).

execution_engine_testing/tests/src/test/explorer/faucet.rs
```diff
@@ -347,7 +348,7 @@ fn should_fund_existing_account() {
 
     assert_eq!(
         user_purse_balance_after,
-        user_purse_balance_before + one_distribution - user_account_initial_balance
+        user_purse_balance_before + one_distribution - user_account_initial_balance + refund
```
rafal-ch, line 351: Looks like we're testing the refund as a side effect in various different tests. Do you think there's a merit in adding a dedicated test like `fn should_refund_proper_amount()` or similar?

node/src/components/network/outgoing.rs
jacek-casper: yeah this was the formatter, I'll move it to another commit 👍

execution_engine_testing/tests/src/test/explorer/faucet.rs
jacek-casper: that's true, we could maybe set the refund to zero percent for these tests and then add a separate dedicated test that uses a non-zero refund, wdyt @rafal-ch?

rafal-ch: I'd be ok with just an additional test dedicated to testing the refund. I think we don't need to revert the other changes you made.

jacek-casper:
address both of your comments now @rafal-ch

rafal-ch (approved):
LGTM :+1:

execution_engine/src/system/handle_payment/internal.rs
```diff
@@ -299,3 +303,40 @@ mod tests {
...
     }
 }
+
+#[cfg(test)]
+mod proptests {
+    use proptest::prelude::*;
+
+    use super::*;
+
+    const DENUM_MAX: u64 = 1000;
```
mpapierski: `DENOM_MAX`, right? as in denominator

mpapierski (approved):

execution_engine/src/system/handle_payment/internal.rs
jacek-casper: fixed

PR #4411, opened 2023-11-13
Host function `generic_hash()` with blake2b/blake3 support.

koxu1996:
_In order to support **Zero Knowledge** software, cryptography related utils should be implemented as a host functions, otherwise execution cost would be too unacceptable. For example running Risc0 VM proof verification for simple **sum of squares** takes more than 6000 CSPR. The underlying hash used: sha256._

This PR adds host function called `generic_hash(input, type)` with support for the following types:
- `HashAlgoType::Blake2b` - used existing blake2b implementation,
- `HashAlgoType::Blake3b` - blake3 library was introduced.

We expect to add more algorithms, i.e. _sha256_, _keccak_, _poseidon_.

### Example usage

Smart contract:

```rust
#![no_std]
#![no_main]

extern crate alloc;

use alloc::format;
use casper_contract::contract_api::{crypto, runtime};
use casper_types::HashAlgoType;

#[no_mangle]
pub extern "C" fn call() {
    let input = "casper".as_bytes();
    runtime::print(&format!("Input: {:?}", input));

    let blake2b = crypto::generic_hash(input, HashAlgoType::Blake2b);
    runtime::print(&format!("Blake2b hash: {:?}", blake2b));

    let blake3 = crypto::generic_hash(input, HashAlgoType::Blake3);
    runtime::print(&format!("Blake3 hash: {:?}", blake3));
}
```

Node output:

```
Input: [99, 97, 115, 112, 101, 114]
Blake2b hash: [163, 40, 63, 203, 149, 170, 125, 252, 231, 127, 90, 198, 44, 250, 234, 42, 1, 199, 20, 158, 150, 188, 25, 205, 21, 106, 255, 40, 204, 65, 45, 196]
Blake3 hash: [122, 158, 116, 51, 175, 213, 20, 50, 237, 63, 251, 230, 53, 2, 233, 86, 36, 159, 195, 109, 254, 108, 189, 210, 233, 1, 93, 219, 226, 1, 50, 217]
```

---

### Considerations

**1. AssemblyScript support**

Is AssemblyScript binding required? I am pretty sure no one uses it.

**2. Gas cost**

For now, I used the same cost for `generic_hash()` as defined for `blake2()` - default fixed value 200. I can do benchmarks for WASM bytecode execution, but is there any conversion ratio between CPU time and execution cost?

**3. Place for core hashes implementation**

Blake2b is implemented in `types` package, which does not seem to be reasonable. I added blake3 to new module `crypto` in execution engine, but I am not fully sure if this is the correct place.

**4. Single vs multiple host function**

Do we want to introduce separate host functions for each hash algorithm?

**5. Place for `generic_hash()` in contract API**

I was thinking of using `runtime`, but finally decided to create new `crypto` module.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,6 +177,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
+name = "arrayref"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+
+[[package]]
 name = "assert-json-diff"
 version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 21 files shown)


koxu1996:
Closed in favor of #4412, that targets `feat-1.6` branch.

devendran-m:
Related Issue (created to take the work forward) #4892

Issue #4088, opened 2021-08-19
Execution Engine 2.0


piotr-dziubecki:
- [Spike]: Explore Wasmer (https://github.com/wasmerio/wasmer) [Michal Papierski](mailto:michal@casperlabs.io)
   - Check out wasmer branch, and extend the code with wasmer engine as extra variant
   - Identify strict wasmer trait requirements known to be blockers and refactor as necessary just to be able to compile without reorganizing everything
   - Hook up host functions to pass meaningful amount of integration tests to measure performance and draw conclusions
   - If successful, identify remaining gaps in VM execution semantics and address them (i.e. fix remaining integration tests)
 - [Spike] Investigate forking wasmi and improving it
 _- [Story/done]: Replace tracking copy entirely w/ in mem scratch based solution. (that means Execution Effects and byte size will go away)_
 - [Story]: Fix EE top level errors coming out of the EE to the node- EE/state/mod.rs errors - not customer facing. 
 - [Story]: Replace Account hashes with Public keys (everywhere). Account hash could become obsolete with the account/contract unification. 
 - [Story]: Get rid of correlation id
 - [Story]: deploy info- do we need it or not -> Remove it from the Global state.
 - [Story]: Simplify or remove FFI Resolver in EE (blocked by the Wasmer investigation).
 - [Story]: Remove protocol version from EE methods.
 - [Story]: Use content-addressed keys for immutable records.
   - ie ‘contract wasm’- hash matches what I expect.
 - [Story]: Alter the FFI methods and remove unnecessary bytes conversion and serialization
 - [Story]: Clean up the ffi, and the contract API
 - [Story]: Introduce the contract ffi API version.
 _- [Spike/triage]: Explore the allocation of data across the wasm boundaries_ 

 - [Story]: Don’t preprocess contracts before storing them and preprocess before executing.	
    - Spreadsheet with preprocessing times measured across integration tests shows that processed wasm code we store is almost 2x as big as original code and takes 5.7ms on avg (with 200ms outlier worth investigating further)
    - Data https://docs.google.com/spreadsheets/d/1qWPrWG1VzU-e3qTaAktDYo489e3g3MZUOWhwSP-Z9s8/edit?usp=sharing

PR #4412, opened 2023-11-13
Host function `generic_hash()` with blake2b/blake3 support.

koxu1996:
_In order to support **Zero Knowledge** software, cryptography related utils should be implemented as a host functions, otherwise execution cost would be too high. For example running Risc0 VM proof verification for simple **sum of squares** takes more than 6000 CSPR. The underlying hash used: sha256._

This PR adds host function called `generic_hash(input, type)` with support for the following types:
- `HashAlgoType::Blake2b` - used existing blake2b implementation,
- `HashAlgoType::Blake3b` - blake3 library was introduced.

We expect to add more algorithms, i.e. _sha256_, _keccak_, _poseidon_.

### Example usage

Smart contract:

```rust
#![no_std]
#![no_main]

extern crate alloc;

use alloc::format;
use casper_contract::contract_api::{crypto, runtime};
use casper_types::HashAlgoType;

#[no_mangle]
pub extern "C" fn call() {
    let input = "casper".as_bytes();
    runtime::print(&format!("Input: {:?}", input));

    let blake2b = crypto::generic_hash(input, HashAlgoType::Blake2b);
    runtime::print(&format!("Blake2b hash: {:?}", blake2b));

    let blake3 = crypto::generic_hash(input, HashAlgoType::Blake3);
    runtime::print(&format!("Blake3 hash: {:?}", blake3));
}
```

Node output:

```
Input: [99, 97, 115, 112, 101, 114]
Blake2b hash: [163, 40, 63, 203, 149, 170, 125, 252, 231, 127, 90, 198, 44, 250, 234, 42, 1, 199, 20, 158, 150, 188, 25, 205, 21, 106, 255, 40, 204, 65, 45, 196]
Blake3 hash: [122, 158, 116, 51, 175, 213, 20, 50, 237, 63, 251, 230, 53, 2, 233, 86, 36, 159, 195, 109, 254, 108, 189, 210, 233, 1, 93, 219, 226, 1, 50, 217]
```

---

### Considerations

~~**1. AssemblyScript support**~~

~~Is AssemblyScript binding required? I am pretty sure no one uses it.~~

_**Update:** AssemblyScript is broken since a long time. I see no reason to include bindings for something that does not work._

~~**2. Gas cost**~~

~~For now, I used the same cost for `generic_hash()` as defined for `blake2()` - default fixed value 200. I can do benchmarks for WASM bytecode execution, but is there any conversion ratio between CPU time and execution cost?~~

_**Update:** There will be [new execution engine](https://github.com/casper-network/casper-node/issues/4088) that will affect (lower) gas usage. It will be the better to make benchmarks then._

**3. Place for core hashes implementation**

Blake2b is implemented in `types` package, which does not seem to be reasonable. I added blake3 to new module `crypto` in execution engine, but I am open to discussing its location.

**4. Single vs multiple host function**

Do we want to introduce separate host functions for each hash algorithm?

**5. Place for `generic_hash()` in contract API**

I was thinking of using `runtime`, but finally decided to create new `crypto` module.

### DIFF
--- a/execution_engine/Cargo.toml
+++ b/execution_engine/Cargo.toml
@@ -14,7 +14,7 @@ license = "Apache-2.0"
 anyhow = "1.0.33"
 base16 = "0.2.1"
 bincode = "1.3.1"
-blake3 = { version = "1.5.0", default-features = false }
+blake3 = "1.5.0"
 casper-hashing = { version = "2.0.0", path = "../hashing" }
 casper-types = { version = "3.0.0", path = "../types", default-features = false, features = ["datasize", "gens", "json-schema"] }
 casper-wasm = { version = "0.46.0", default-features = false }
--- a/execution_engine/Cargo.toml
+++ b/execution_engine/Cargo.toml
@@ -14,6 +14,7 @@ license = "Apache-2.0"
 anyhow = "1.0.33"
 base16 = "0.2.1"
 bincode = "1.3.1"
+blake3 = { version = "1.5.0", default-features = false }
 casper-hashing = { version = "2.0.0", path = "../hashing" }
 casper-types = { version = "3.0.0", path = "../types", default-features = false, features = ["datasize", "gens", "json-schema"] }
 casper-wasm = { version = "0.46.0", default-features = false }
--- a/execution_engine/src/core/runtime/externals.rs
+++ b/execution_engine/src/core/runtime/externals.rs
@@ -1106,8 +1106,19 @@ where
                 // args(3) = size of hash type in Wasm memory
                 // args(4) = pointer to output pointer in Wasm memory
                 // args(5) = size of output
-                let (in_ptr, in_size, hash_algo_type_ptr, hash_algo_type_size, out_ptr, out_size): (u32, u32, u32, u32, u32, u32) =
+                let (in_ptr, in_size, hash_algo_type_ptr, hash_algo_type_size, out_ptr, out_size) =
                     Args::parse(args)?;
+                self.charge_host_function_call(
+                    &host_function_costs.generic_hash,
+                    [
+                        in_ptr,
+                        in_size,
+                        hash_algo_type_ptr,
+                        hash_algo_type_size,
+                        out_ptr,
+                        out_size,
+                    ],
+                )?;
                 let hash_algo_type = self.t_from_mem(hash_algo_type_ptr, hash_algo_type_size)?;
                 let digest =
                     self.checked_memory_slice(in_ptr as usize, in_size as usize, |input| {
--- a/execution_engine/src/core/runtime/externals.rs
+++ b/execution_engine/src/core/runtime/externals.rs
@@ -9,13 +9,14 @@ use casper_types::{
     contracts::{ContractPackageStatus, EntryPoints, NamedKeys},
     crypto,
     system::auction::EraInfo,
-    ApiError, ContractHash, ContractPackageHash, ContractVersion, EraId, Gas, Group, Key,
-    StoredValue, URef, U512, UREF_SERIALIZED_LENGTH,
+    ApiError, ContractHash, ContractPackageHash, ContractVersion, EraId, Gas, Group, HashAlgoType,
+    Key, StoredValue, URef, U512, UREF_SERIALIZED_LENGTH,
 };
 
 use super::{args::Args, Error, Runtime};
 use crate::{
     core::resolvers::v1_function_index::FunctionIndex,
+    core::runtime::crypto as core_crypto,
     shared::host_function_costs::{Cost, HostFunction, DEFAULT_HOST_FUNCTION_NEW_DICTIONARY},
     storage::global_state::StateReader,
 };
@@ -1096,6 +1097,39 @@ where
                 let result = self.enable_contract_version(contract_package_hash, contract_hash)?;
 
                 Ok(Some(RuntimeValue::I32(api_error::i32_from(result))))
+            }
+
+            FunctionIndex::GenericHash => {
+                // args(0) = pointer to input in Wasm memory
+                // args(1) = size of input in Wasm memory
+                // args(2) = pointer to hash type pointer in Wasm memory
+                // args(3) = size of hash type in Wasm memory
+                // args(4) = pointer to output pointer in Wasm memory
+                // args(5) = size of output
+                let (in_ptr, in_size, hash_algo_type_ptr, hash_algo_type_size, out_ptr, out_size): (u32, u32, u32, u32, u32, u32) =
+                    Args::parse(args)?;
+                let hash_algo_type = self.t_from_mem(hash_algo_type_ptr, hash_algo_type_size)?;
+                let digest =
+                    self.checked_memory_slice(in_ptr as usize, in_size as usize, |input| {
+                        match hash_algo_type {
+                            HashAlgoType::Blake2b => crypto::blake2b(input),
+                            HashAlgoType::Blake3 => core_crypto::blake3(input),
+                        }
+                    })?;
+
+                let result = if digest.len() > out_size as usize {
+                    Err(ApiError::BufferTooSmall)
+                } else {
+                    Ok(())
+                };
+                if result.is_err() {
+                    return Ok(Some(RuntimeValue::I32(api_error::i32_from(result))));
+                }
+
+                self.try_get_memory()?
+                    .set(out_ptr, &digest)
+                    .map_err(|error| Error::Interpreter(error.into()))?;
+                Ok(Some(RuntimeValue::I32(0)))
             }
         }
     }
--- a/execution_engine/src/core/runtime/externals.rs
+++ b/execution_engine/src/core/runtime/externals.rs
@@ -1102,24 +1102,17 @@ where
             FunctionIndex::GenericHash => {
                 // args(0) = pointer to input in Wasm memory
                 // args(1) = size of input in Wasm memory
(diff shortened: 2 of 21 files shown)


execution_engine/Cargo.toml
```diff
@@ -14,6 +14,7 @@ license = "Apache-2.0"
 anyhow = "1.0.33"
 base16 = "0.2.1"
 bincode = "1.3.1"
+blake3 = { version = "1.5.0", default-features = false }
```
mpapierski: I think it may actually be better to use default features i.e. 

```suggestion
blake3 = "1.5.0"
```

According to the source here: https://github.com/BLAKE3-team/BLAKE3/blob/92e4cd71be48fdf9a79e88ef37b8f415ec5ac210/Cargo.toml#L21-L26 if you use `--no-default-features` it will not do runtime CPU feature detection and won't use SIMD implementation.

execution_engine/src/core/runtime/externals.rs
```diff
@@ -1097,6 +1098,50 @@ where
 
                 Ok(Some(RuntimeValue::I32(api_error::i32_from(result))))
             }
+
+            FunctionIndex::GenericHash => {
+                // args(0) = pointer to input in Wasm memory
+                // args(1) = size of input in Wasm memory
+                // args(2) = pointer to hash type pointer in Wasm memory
+                // args(3) = size of hash type in Wasm memory
```
mpapierski: `HashAlgoType` is a plain enum represented by u8 - we don't need two parameters to pass this value to the host. You can simply cast the enum variant to an integer and pass it as a single value without involving serialization.

smart_contracts/contract/src/contract_api/crypto.rs
```diff
@@ -0,0 +1,25 @@
...
+    let mut ret = [0; 32];
+
+    let algo_ptr = &algo as *const _ as *const u8;
+
+    let result = unsafe {
+        ext_ffi::casper_generic_hash(
+            input.as_ref().as_ptr(),
+            input.as_ref().len(),
+            algo_ptr,
+            1, // HashAlgoType is just one byte (u8).
```
mpapierski:
```suggestion
            algo as u8,
```

And no need for `algo_ptr` etc. `HashAlgoType` is safe to passed as an integer.

types/src/crypto.rs
```diff
@@ -33,3 +43,53 @@ pub fn blake2b<T: AsRef<[u8]>>(data: T) -> [u8; BLAKE2B_DIGEST_LENGTH] {
...
+#[cfg_attr(feature = "datasize", derive(DataSize))]
+#[cfg_attr(feature = "json-schema", derive(JsonSchema))]
+pub enum HashAlgoType {
+    /// Blake2b
+    Blake2b = 0,
+    /// Blake3
+    Blake3 = 1,
+}
+
+impl ToBytes for HashAlgoType {
```
mpapierski: You can skip ToBytes/FromBytes for this type as it's easier to pass it by value. Same with `impl AsRef` I think.

types/src/crypto.rs
```diff
@@ -33,3 +43,53 @@ pub fn blake2b<T: AsRef<[u8]>>(data: T) -> [u8; BLAKE2B_DIGEST_LENGTH] {
     });
     result
 }
+
+/// HashAlgoType
```
mpapierski:
```suggestion
/// A type of hashing algorithm.
```

execution_engine/Cargo.toml
koxu1996: Good catch! Suggestion approved: https://github.com/casper-network/casper-node/pull/4412/commits/8bd8fbb7893e74846c65468fe1518646eca62071.

types/src/crypto.rs
koxu1996: Thanks, approved: https://github.com/casper-network/casper-node/pull/4412/commits/d38140f3d88806385ae069880735f67dccf57a02.

smart_contracts/contract/src/contract_api/crypto.rs
koxu1996: Done in https://github.com/casper-network/casper-node/pull/4412/commits/2c70942af76e7e46b5c9d1fa74445d8d5131e555.

execution_engine/src/core/runtime/externals.rs
koxu1996: Done in https://github.com/casper-network/casper-node/pull/4412/commits/2c70942af76e7e46b5c9d1fa74445d8d5131e555.

types/src/crypto.rs
koxu1996: I removed both, and introduced `TryFrom<u8>` - see https://github.com/casper-network/casper-node/pull/4412/commits/2c70942af76e7e46b5c9d1fa74445d8d5131e555.

---

Btw. handling error was crazy, as compiler could not determine that `casper_types::bytesrepr::Error` should be converted into `casper_execution_engine::core::execution::error`, and then into `casper_wasmi_core::trap::Trap`. Finally I found working line:
```rs
let hash_algo_type = HashAlgoType::try_from(hash_algo_type as u8).map_err(|e| Error::from(e))?;
```

types/src/crypto.rs
```diff
@@ -33,3 +40,21 @@ pub fn blake2b<T: AsRef<[u8]>>(data: T) -> [u8; BLAKE2B_DIGEST_LENGTH] {
...
+    Blake2b = 0,
+    /// Blake3
+    Blake3 = 1,
+}
+
+impl TryFrom<u8> for HashAlgoType {
+    type Error = bytesrepr::Error;
+
+    fn try_from(value: u8) -> Result<Self, bytesrepr::Error> {
+        FromPrimitive::from_u8(value).ok_or(bytesrepr::Error::Formatting)
```
mpapierski: I'd probably use `ApiError::InvalidArgument` here to make it more obvious, since `Formatting` error may be confusing as there's no serialization involved

execution_engine/src/core/runtime/externals.rs
```diff
@@ -1097,6 +1098,43 @@ where
...
+
+                let digest =
+                    self.checked_memory_slice(in_ptr as usize, in_size as usize, |input| {
+                        match hash_algo_type {
+                            HashAlgoType::Blake2b => crypto::blake2b(input),
+                            HashAlgoType::Blake3 => core_crypto::blake3(input),
+                        }
+                    })?;
+
+                let result = if digest.len() > out_size as usize {
```
mpapierski, line 1129: Something I just noticed - we've done hashing already, but we're checking the size after the fact. Since the output of the hash is constant, we could also move this check above the `checked_memory_slice` line.

mpapierski (approved):
Two remaining comments and looks good otherwise 👍

execution_engine/src/core/runtime/externals.rs
koxu1996: Even though the output is always expected to be 32 bytes, those hashes use different constants for their digest size:
- blake2: `BLAKE2B_DIGEST_LENGTH`
- blake3: `BLAKE3_DIGEST_LENGTH`

Unless we implement all the necessary hash functions (e.g., _sha256_, _keccak_, _poseidon_) and **generalize** it with something like `GENERIC_HASH_DIGEST_LENGTH`, I would prefer to avoid using hardcoded pre-checks.

types/src/crypto.rs
koxu1996: 1. I agree `Formatting` error might be confusing. Why I did that? I based on [`ExecutionResultTag`](https://github.com/casper-network/casper-node/blob/v1.5.3/types/src/execution_result.rs#L44-L57) and few other enums in `execution_result.rs`, but now I see those `TryFrom`'s are always used internally inside `ExecutionResult::from_bytes()`.

2. I think `ApiError` should not be present in base Casper types, as no API is involved here. Therefore I think we can just **keep it simple** and make `()` the error - https://github.com/casper-network/casper-node/pull/4412/commits/85885d0bb5e75842fc4c282317acd5db1ab6a20a:

image

3. We can return `ApiError::InvalidArgument` inside function handler - it makes sense - but it complicates the code a little bit. Nevertheless, done in https://github.com/casper-network/casper-node/pull/4412/commits/d064e2c90148a5e13b3ecaf1f98a1f3646e04ee0.

koxu1996:
I just confirmed that host function is still working correctly after all changes and fixes.

types/src/crypto.rs
koxu1996: Additional, please take in mind that user will not able to face this error, because using `HashAlgoType` enum as an argument forces the correct value.

EdHastingsCasperAssociation:
Michal's approval is sufficient for you to move forward with this at the functional level. As far as inclusion into 1.6, @marc-casperlabs is the principal engineer on that rel; I've had no direct involvement with it.

If it can't go into 1.6 for timeline reasons, we can repoint it to 2.0 and then I'll review it.

EdHastingsCasperAssociation:
You should repoint this to feat-2.0.

EdHastingsCasperAssociation:
@mpapierski should this be repointed to feat-2.0 ?

mpapierski:
> @mpapierski should this be repointed to feat-2.0 ?

I talked to @asladeofgreen and @jonas089 and for now, there's no concrete use case for generic hash functions, but we can revisit this once we have use for this internally or externally from the community. My understanding is that this PR was done as part of research.

devendran-m:
Related issue (created to take this forward) #4892

PR #4439, opened 2023-11-24
Update nctl path

Enflow-io:
According to the instructions:
https://docs.casper.network/developers/dapps/setup-nctl/

we have to fix the NCTL path

Please consider the following when creating a PR:


gRoussac:
@Enflow-io `pwd` should already contain the correct path /utils/nctl so I am closing your pR, if you have issues with that again please contact back

PR #4447, opened 2023-12-01
Fix several JsonSchema impls and update the schema

jacek-casper:
Several types generate incorrect schema due to invalid use of the schemars macro attributes, the code was using the `with` attribute, but it doesn't support container types (structs/enums) ([it's documented here](https://github.com/GREsau/schemars/blob/master/docs/1.1-attributes.md#serdewith--type--schemarswith--type)). I've written the impls by hand, there seems to be no way to do it with attributes out of the box.

### DIFF
--- a/types/src/timestamp.rs
+++ b/types/src/timestamp.rs
@@ -240,7 +240,7 @@ impl From<u64> for Timestamp {
 #[cfg_attr(
     feature = "json-schema",
     derive(JsonSchema),
-    schemars(description = "Human-readable duratio.n")
+    schemars(description = "Human-readable duration.")
 )]
 pub struct TimeDiff(#[cfg_attr(feature = "json-schema", schemars(with = "String"))] u64);
 
--- a/types/src/timestamp.rs
+++ b/types/src/timestamp.rs
@@ -30,7 +30,12 @@ const TIMESTAMP: Timestamp = Timestamp(1_605_573_564_072);
 /// A timestamp type, representing a concrete moment in time.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "datasize", derive(DataSize))]
-pub struct Timestamp(u64);
+#[cfg_attr(
+    feature = "json-schema",
+    derive(JsonSchema),
+    schemars(description = "Timestamp formatted as per RFC 3339")
+)]
+pub struct Timestamp(#[cfg_attr(feature = "json-schema", schemars(with = "String"))] u64);
 
 impl Timestamp {
     /// The maximum value a timestamp can have.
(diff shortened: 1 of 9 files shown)


rafal-ch (approved):
LGTM :+1:

types/src/transaction/transaction_v1/transaction_v1_hash.rs
```diff
@@ -23,13 +23,8 @@ use crate::{
     Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Hash, Serialize, Deserialize, Debug, Default,
 )]
 #[cfg_attr(feature = "datasize", derive(DataSize))]
-#[cfg_attr(
-    feature = "json-schema",
-    derive(JsonSchema),
-    schemars(with = "String", description = "Hex-encoded TransactionV1 hash.")
-)]
```
Fraser999: I think we could reinstate this, but remove only the `with = "String", ` part, as it's misleading.  Having the generated schema as
```json
      "allOf": [
        {
          "$ref": "#/definitions/Digest"
        }
      ]
```
is ok for this type, as `Digest` is correctly identified as a string type in the schema.

types/src/timestamp.rs
```diff
@@ -30,11 +30,6 @@ const TIMESTAMP: Timestamp = Timestamp(1_605_573_564_072);
 /// A timestamp type, representing a concrete moment in time.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "datasize", derive(DataSize))]
-#[cfg_attr(
-    feature = "json-schema",
-    derive(JsonSchema),
-    schemars(with = "String", description = "Timestamp formatted as per RFC 3339")
-)]
 pub struct Timestamp(u64);
```
Fraser999: Seems like the following would produce the same schema:
```rust
#[cfg_attr(
    feature = "json-schema",
    derive(JsonSchema),
    schemars(description = "Timestamp formatted as per RFC 3339")
)]
pub struct Timestamp(#[cfg_attr(feature = "json-schema", schemars(with = "String"))] u64);
```

types/src/transaction/transaction_v1/transaction_v1_hash.rs
```diff
@@ -23,13 +23,8 @@ use crate::{
...
 #[serde(deny_unknown_fields)]
-pub struct TransactionV1Hash(#[cfg_attr(feature = "json-schema", schemars(skip))] Digest);
+pub struct TransactionV1Hash(Digest);
```
Fraser999, line 32: This is a good change, but we should probably apply it similarly to `BlockHash` and `DeployHash`.

Fraser999 (changes requested):
Nice catch on this issue!  However, I have a few suggestions on a slightly different approach for the changes.

Fraser999 (approved):
LGTM - nice work!

Fraser999 (approved):

PR #4461, opened 2023-12-14
Improvements for the contract level messages interface

alsrdn:
This PR adds a few improvements to the contract level messages interface as discussed on the feedback of the Contract level Messages CEP ([here](https://github.com/casper-network/ceps/pull/88#discussion_r1409472986), [here](https://github.com/casper-network/ceps/pull/88#discussion_r1409482000) and [here](https://github.com/casper-network/ceps/pull/88#discussion_r1410061914)):
* Add the ability for `MessagePayload` to hold `Bytes` instead of a string. This would allow users to spend less gas for serializing contents for messages that are not in a human readable format nativelly.
* Don't use a prefix for the `TopicNameHash` string formatting since it is always a part of a `Key` for a topic and prefixing it only makes it harder to create the key.
* Allow message topics to be directly managed when a contract is created or upgraded through the `casper_add_package_version` FFI. This is easier to use for contracts which have a stable number of topics that is known at install time and allows saving some gas since it doesn't require the installer code to use an initializer entry point in the contract to register topics on install.
* ~~Add a new per-transaction log in global state that tracks the sequencing of emitted messages across multiple contracts and topics. The details for this implementation are specified in the CEP (https://github.com/casper-network/ceps/blob/afaf653819ca6db9bf68415586ffd93aec2dd8a8/text/0088-contract-level-messages.md#transaction-message-log-on-chain)~~. Because pruning the per-transaction log entries is rather impractical, we decided not to include that change in the PR. We'll create a new PR to address that particular issue.


Fraser999 (approved):

execution_engine/src/runtime/mod.rs
```diff
@@ -1805,8 +1807,40 @@ where
...
+            .iter()
+            .filter(|(_, operation)| match operation {
+                MessageTopicOperation::Add => true,
+            });
+        // Check if registering the new topics would exceed the limit per contract
+        if previous_message_topics.len() + topics_to_add.clone().count()
+            > max_topics_per_contract as usize
+        {
+            return Ok(Err(ApiError::from(MessageTopicError::MaxTopicsExceeded)));
+        }
```
rafal-ch, line 1835: Right here we return `MaxTopicsExceeded` when we exceed `max_topics_per_contract`, while in `add_topic()` we return the same error when we exceed `u32::MAX`. Is this intended?

execution_engine_testing/tests/src/test/contract_messages.rs
```diff
@@ -579,6 +593,18 @@ fn should_carry_message_topics_on_upgraded_contract() {
     assert_eq!(expected_topic_names, 2);
 }
 
+#[ignore]
+#[test]
+fn should_carry_message_topics_on_upgraded_contract_without_initializer() {
```
rafal-ch:
```suggestion
fn should_carry_message_topics_on_upgraded_contract_with_initializer() {
```

execution_engine_testing/tests/src/test/contract_messages.rs
```diff
@@ -579,6 +593,18 @@ fn should_carry_message_topics_on_upgraded_contract() {
...
+    should_carry_message_topics_on_upgraded_contract(true);
+}
+
+#[ignore]
+#[test]
+fn should_carry_message_topics_on_upgraded_contract_with_initializer() {
```
rafal-ch:
```suggestion
fn should_carry_message_topics_on_upgraded_contract_without_initializer() {
```

execution_engine/src/runtime/mod.rs
alsrdn: Yeah, I think the comparison with `u32::MAX` is unnecessary since it is enforced by this other limit which itself is an `u32`. I'll remove it.

alsrdn: Removed in: https://github.com/casper-network/casper-node/pull/4461/commits/08f12c8722a313d044bfe397297e95a695cb53e2

execution_engine_testing/tests/src/test/contract_messages.rs
alsrdn: Renamed in: https://github.com/casper-network/casper-node/pull/4461/commits/08f12c8722a313d044bfe397297e95a695cb53e2

execution_engine_testing/tests/src/test/contract_messages.rs
alsrdn: Renamed in: https://github.com/casper-network/casper-node/pull/4461/commits/08f12c8722a313d044bfe397297e95a695cb53e2

alsrdn:
bors merge

PR #4483, opened 2024-01-03
Disable banning based on failed pings

marc-casperlabs:
Disables the ban due to failed ping roundtrips in the networking component, turning this feature into a more diagnostic one instead (ping roundtrip times should be visible through the diagnostic port's network info).


rafal-ch (approved):
LGTM :+1:

rafal-ch:
A nit:
The comment in `health.rs` says:
```
/// If `true`, the maximum number of pongs has been exceeded and the peer should be banned.
```
I'd remove the `and the peer should be banned` part, as we're leaving the decision to ban or not to the caller.

EdHastingsCasperAssociation (approved):

Fraser999 (approved):

PR #4484, opened 2024-01-03
config validator shutdown behavior

EdHastingsCasperAssociation:
provides config option to allow a validator node to complete controlled shutdown


Fraser999 (approved):

sacherjj (approved):
My concern was working by default without config change for binary swap and looks like it defaults correctly.

marc-casperlabs:
bors retry

PR #4485, opened 2024-01-03
alt proposer strat: round robin over distinct body_hash

EdHastingsCasperAssociation:
alternate block proposal strategy, buckets all proposable deploys by body_hash (which collates deploys with identical work payload regardless of timestamp or origin), then round robins across the buckets. the intent is to prevent blocks from being monopolized by procedural / iterative type deploys.

this is off the cuff and lightly tested, feedback welcome.

### DIFF
--- a/node/src/components/deploy_buffer.rs
+++ b/node/src/components/deploy_buffer.rs
@@ -327,13 +327,7 @@ impl DeployBuffer {
             .collect()
     }
 
-    /// Returns a right-sized payload of deploys that can be proposed.
-    fn appendable_block(&mut self, timestamp: Timestamp) -> AppendableBlock {
-        let mut ret = AppendableBlock::new(self.deploy_config, timestamp);
-        let mut holds = HashSet::new();
-        let mut have_hit_transfer_limit = false;
-        let mut have_hit_deploy_limit = false;
-
+    fn buckets(&mut self) -> HashMap<Digest, Vec<(DeployHashWithApprovals, DeployFootprint)>> {
         let proposable = self.proposable();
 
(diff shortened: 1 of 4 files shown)


Fraser999 (approved):

node/src/components/deploy_buffer.rs
```diff
@@ -332,7 +333,48 @@ impl DeployBuffer {
...
+            };
+            let (with_approvals, footprint) = match buckets.entry(*body_hash) {
+                Entry::Vacant(_) => continue,
+                Entry::Occupied(entry) => {
+                    let deploys = entry.into_mut();
+                    if deploys.is_empty() {
+                        buckets.remove(body_hash);
+                        continue;
+                    }
+                    deploys.remove(0)
```
marc-casperlabs: This is O(n), making the entire algorithm O(n^2) in theory in the case of having a large queue all under a single body hash. In reality, it will be bound by the deploys-per-block limit, so it'll be O(50n).

Either we should use a `VecDeque` instead of a `Vec`, or, if we can afford it, change the order by calling `deploys.pop()` instead.

marc-casperlabs (approved):
I'd love to see either the proposed changes or the test be added, but given the current time constraints, I'm good with merging this as-is, for now.

Please give the comment a look though, while not a complete deal breaker with current settings, I think it is worth a look.

node/src/components/deploy_buffer.rs
EdHastingsCasperAssociation: tests added, impl'd some of what you suggested

marc-casperlabs (approved):

PR #4486, opened 2024-01-07
Filter for IPv4 addresses

GuybrushX:
Since IPv6 isn't officially supported but caused trouble during multiple upgrades in the past already -> filter it out.

I have NOT tested this PR because I lack the required test network and time and experience.

Reason is that right now there are quite some validators/nodes running on IPv6 because the initial upgrade instructions and scripts didn't filter out IPv6 and the node operators didn't understand that.

So I think it's the best to stop IPv6 on the node level and to force node operators to check and fix their config.

### DIFF
--- a/node/src/utils.rs
+++ b/node/src/utils.rs
@@ -111,6 +111,8 @@ pub(crate) fn resolve_address(address: &str) -> Result<SocketAddr, ResolveAddres
             address: address.to_string(),
             kind: ResolveAddressErrorKind::ErrorResolving(err),
         })?
+        /// Filter for IPv4 addresses
+        .filter(|addr| addr.is_ipv4())
         .next()
         .ok_or_else(|| ResolveAddressError {
             address: address.to_string(),


marc-casperlabs:
I've taken the liberty of rewriting this, instead of fixing the submitted version, and adding the minimum necessary documentation. See #4488.

marc-casperlabs:
We're going to add this to the 1.6 release instead of the 1.5 branch. From our own logs, we assume this is rarely used, while we do see a couple of IPv6 addresses floating around the network, we are currently assuming this is due to old scripts being used that _directly_ write an IP address to the config files, not due to DNS lookups.

GuybrushX:
> We're going to add this to the 1.6 release instead of the 1.5 branch. From our own logs, we assume this is rarely used, while we do see a couple of IPv6 addresses floating around the network, we are currently assuming this is due to old scripts being used that _directly_ write an IP address to the config files, not due to DNS lookups.

Happy to see this fix was adjusted and accepted 🙂

I think the real reason and source of the issue is described in detail here in case someone needs more details:
casper-network/casper-node-launcher#44

TL;DR:
The external IP detection sometimes returned IPv6 addresses in older versions of `node_util.py` when an upgrade was staged -> there was no validation done - neither by the script nor by the node operator so it wrote IPv6 addresses in the `config.toml` file.

Some node operators obviously "don't touch a running system" so the `node_util.py` script is outdated and returns IPv6 addresses in some situations.

marc-casperlabs:
> Happy to see this fix was adjusted and accepted 🙂

Thank you for suggesting it!

> I think the real reason and source of the issue is described in detail here in case someone needs more details: [casper-network/casper-node-launcher#44](https://github.com/casper-network/casper-node-launcher/issues/44)
> 
> TL;DR: The external IP detection sometimes returned IPv6 addresses in older versions of `node_util.py` when an upgrade was staged -> there was no validation done - neither by the script nor by the node operator so it wrote IPv6 addresses in the `config.toml` file.
> 
> Some node operators obviously "don't touch a running system" so the `node_util.py` script is outdated and returns IPv6 addresses in some situations.

That's a known issue - these nodes will never be able to fully participate in the network. I will bring this issue up again internally.

EdHastingsCasperAssociation:
@marc-casperlabs is this PR closeable or is there more to discuss?

GuybrushX:
From my side it can be closed.

PR #4490, opened 2024-01-09
Improve testing and error message for IPv6 rejeciton

marc-casperlabs:
This adds a test and a better error in case of rejection.

### DIFF
--- a/node/src/utils.rs
+++ b/node/src/utils.rs
@@ -612,7 +612,7 @@ mod tests {
 
     #[test]
     fn resolve_address_rejects_ipv6() {
-        let raw = "::1:12345";
+        let raw = "2b02:c307:2042:360::1:0";
         assert!(resolve_address(raw).is_err());
     }
 }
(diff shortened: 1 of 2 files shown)


node/src/utils.rs
```diff
@@ -607,4 +609,10 @@ mod tests {
 
         assert_eq!(extracted, expected);
     }
+
+    #[test]
+    fn resolve_address_rejects_ipv6() {
```
rafal-ch, line 614: Not a blocker, but since I'm paranoid, I'd consider adding:
```
fn resolve_address_accepts_ipv4()
```

rafal-ch (approved):
LGTM :+1:

Fraser999 (approved):

node/src/utils.rs
marc-casperlabs: -.-

marc-casperlabs: See 4aa04b746

